PR #1, opened 2026-06-06
docs: simplify README for getting-started; add canonical AGENTS.md

Lanzelot1:
Simplify the root README to be getting-started first, and add a canonical `AGENTS.md`.

## README.md (105 → 62 lines)
- **Quick start** is now the first substantive section: clone+install on one line, mesh + two peers + console.
- Demoted the dense **wire-contract subjects** block and the **4-tier layout** block to one-line pointers under a new **Learn more** section.
- Kept the delivery-modes table under a tight **How it works**; trimmed Status.

## AGENTS.md (new, canonical)
- The [agents.md](https://agents.md) standard file so any agent (Codex, Cursor, Gemini CLI…) picks up the same repo guidance Claude Code gets.
- Carries the substance of the old `CLAUDE.md`: what Swarl is, layout + one-way dep tiers, core primitives, dev commands, research-first, conventions (ESM, no-fallbacks/throw, keep docs updated).

## CLAUDE.md → symlink to AGENTS.md
- Single source of truth, zero drift. Claude Code reads `CLAUDE.md`, everything else reads `AGENTS.md`, both resolve to the same bytes.

Docs only — no code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3, opened 2026-06-06
License under Apache-2.0

Lanzelot1:
## What

Licenses the whole repo under **Apache-2.0** and records the licensing strategy.

- `LICENSE` — full Apache-2.0 text
- `NOTICE` — copyright + trademark notice
- `LICENSING.md` — the reasoning: why permissive, the trademark reservation, and the deferred commercial path
- `license: "Apache-2.0"` on all 9 `package.json` (root + every package/extension/impl)
- short **License** section in the README

## Why Apache-2.0 (and not GPL like cmux)

cmux is a standalone app, so GPL is painless there. Swarl is **embeddable SDKs + a wire protocol** that other people import into their own (often proprietary) agents. Copyleft on those SDKs would be an adoption killer, which directly fights goal #1: *become the standard*. Apache-2.0 is what the protocols Swarl sits alongside use — MCP, A2A, gRPC, and NATS itself — and it adds a patent grant over MIT.

## Where the business comes from (so we don't need copyleft yet)

The protocol is the funnel, not the company; value accrues at a managed control plane (hosted mesh + observability + registry). So the license doesn't need to be the moat today. We keep the dual-license card in reserve: when **Swarl Cloud** ships, the *server* can move to **AGPL-3.0-or-later + commercial** while the SDKs and spec stay Apache-2.0. Tightening an unreleased server later triggers no community revolt; giving away the SDKs permissively now buys adoption we can't reverse out of. This is documented in `LICENSING.md`.

## Notes

- Base is `cmux-demo` (the active branch, 21 commits ahead of `main`); retarget to `main` if preferred.
- Only licensing files are touched — unrelated working-tree changes were left alone.
- Follow-ups (separate PRs): reposition the docs as a complement to MCP/A2A, and add a DCO/CLA when the commercial server actually lands.


Lanzelot1:
Apache-2.0 is already on main (via 3fb199a); closing this stale PR (base cmux-demo is being removed).

PR #5, opened 2026-06-07
feat: Hermes (Nous Research) as a native mesh peer

Lanzelot1:
## What

Adds **Hermes** ([Nous Research](https://github.com/NousResearch/hermes-agent)) as a native Swarl mesh peer, alongside the existing OpenAI Agents (TS+Py) and Vercel AI peers. A Hermes agent now joins a space and answers DMs, anycasts, and @-mentions on channels, interoperating on the wire with the TS peers.

Hermes is Python-first, so this mirrors `extensions/openai-agents-py`:

- **`extensions/hermes-py`** (`@swarl/hermes-py`) — a `Connector` (`connector.ts`) that `uv run`s a Python peer (`swarl_hermes.peer`). The peer embeds Hermes' `AIAgent` and runs the established serialized loop: each inbound message flips presence to `working`, runs `AIAgent.chat` (off the event loop via `asyncio.to_thread` so heartbeats keep flowing), replies on the same delivery mode, then flips to `idle`. Same wake guards as the other peers (ignore self / `feedback` channel / unmentioned channel traffic). The generic Python wire client is bundled self-contained.
- **`examples/03-hermes-py`** — composition root that registers the connector (+ `swarl` alias).
- **docs** — `agent-frameworks.md` (4th adapter; noted the peer is loop-only, no model-facing mesh tools yet) and `examples.md`.

## Config

Model-agnostic. Default path is **OpenRouter** (`OPENROUTER_API_KEY`, default model `anthropic/claude-sonnet-4.6`, override via `HERMES_MODEL`); `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `NOUS_API_KEY` are also forwarded.

## Verification

- `pnpm typecheck` green across the workspace; `uv run python -m swarl_hermes.peer` imports + constructs cleanly.
- **Live on a real NATS mesh** (mesh layer only): spawned `hermes1`, DM'd it from a tester peer. Confirmed: it appeared in the roster, received the DM, ran the loop, and routed a reply **as a DM back to the sender** on the correct delivery mode.
- **Not yet demonstrated:** a successful model answer. In testing the `AIAgent.chat` call returned an API error (no valid provider key was configured — the test key was invalid and Hermes defaulted its endpoint to Anthropic), and the loop faithfully delivered that error text as the reply. A real answer needs a valid provider key (e.g. `OPENROUTER_API_KEY` with the default model, or `ANTHROPIC_API_KEY` with a `claude-*` model).


Lanzelot1:
Superseded by the feat/connector-hermes rework.

PR #25, opened 2026-06-15
feat(cli): `cotal dm` operator command (non-interactive send)

caffeinum:
## What

A first-class `cotal dm <name> "<message>"` operator command — send a mesh DM to a running agent without attaching to its terminal or proxying through a `me` agent.

```bash
cotal dm team2027 "what's your status?"
cotal dm team2027 "status?" --wait              # wait up to 60s, print the reply
cotal dm team2027 "status?" --wait --timeout 120
```

## Why

`cotal_dm` existed only as an agent-only MCP tool — no CLI counterpart. An operator's only options were (a) `cotal attach` and type raw into the PTY, or (b) spawn a `me` agent to relay. This closes that gap.

## How it works
- connects to NATS like `console`/`watch` (off-roster, presence-watching client; same server/creds resolution)
- resolves target name → instance id off the presence roster (case-insensitive, prefers a live peer, waits briefly for KV replay) — mirrors `MeshAgent.resolvePeer`
- publishes via `ep.unicast(targetId, text)` — the **same** subject (`cotal.<space>.inst.<targetId>.<senderId>`) and envelope the console's operator DM and the agent `cotal_dm` tool emit, so the receiver parses it as a normal DM (no schema drift by construction)
- `--space` / `--server` / `--creds` flags, consistent with other commands

### `--wait` (second commit)
- taps our own inbox (`inst.<me>.*`) **before** sending so a fast reply isn't missed
- registers presence for the wait window so the agent can resolve us as a reply target off the roster (by name or id) — invisible by default otherwise
- prints the first reply from the DMed agent, or a `cotal watch` hint on timeout
- `--timeout <s>` (default 60s)

## Auth mode
A DM is a write, and the self-mintable `observer` cred is read-only — so in auth mode the command requires an explicit admin cred (`cotal mint <n> --profile admin` → `--creds <path>`) and **fails loudly** rather than picking creds that can't publish. Open mesh connects bare. (A narrower publish-only operator profile in `provision.ts` would be a nice follow-up.)

## Test
Verified end-to-end against a live local mesh: `cotal dm team2027 "ping" --wait` resolved the agent, published the DM, and **printed team2027's reply inline** (`← team2027: ...`). Build + typecheck clean.

### DIFF
--- a/implementations/cli/src/commands/dm.ts
+++ b/implementations/cli/src/commands/dm.ts
@@ -1,0 +1,134 @@
+import { parseArgs } from "node:util";
+import { readFileSync } from "node:fs";
+import { userInfo } from "node:os";
+import {
+  CotalEndpoint,
+  isReachable,
+  DEFAULT_SERVER,
+  DEFAULT_SPACE,
+  authDir,
+  loadSpaceAuth,
+  type Presence,
+} from "@cotal-ai/core";
+import { c } from "../ui.js";
+
+/** How long to wait for the presence roster to replay before giving up on a name lookup.
+ *  The observer's KV watch fills the roster asynchronously after start(); a target that's
+ *  truly present shows up well within this window. */
+const ROSTER_WAIT_MS = 3000;
+
+/** Resolve an agent name → its present instance id off the roster, waiting briefly for the
+ *  presence KV to replay. Case-insensitive, prefers a live (non-offline) peer. */
+async function resolveTarget(ep: CotalEndpoint, name: string): Promise<Presence | undefined> {
+  const t = name.toLowerCase();
+  const find = (): Presence | undefined => {
+    const roster = ep.getRoster();
+    const present = roster.filter((p) => p.status !== "offline");
+    return (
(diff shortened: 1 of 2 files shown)


davidfarah2003:
Superseded by #33, which shipped `cotal dm` with the same `ep.unicast` / subject / envelope and the same name→id roster resolution. Closing to avoid a duplicate `dm` command (this branch would now collide with the merged `commands/send.ts` and its `dm` registration).

The one piece #33 doesn't cover is `--wait` / `--timeout` (waiting for and printing the agent's reply). If we want that, we'll add it as a small follow-up on top of the merged `connectSender` so it inherits the `cotalRoot()` auth resolution and self-minted write cred. Thanks @caffeinum!

davidfarah2003:
Decision: we are **not** adding `--wait` for now. The merged `cotal dm` stays a clean fire-and-forget send. To see an agent's reply, observe the mesh with `cotal web` (the browser dashboard shows DMs and anycast live), or `cotal watch` / `cotal console`.

PR #24, opened 2026-06-15
feat(connector): resume a prior claude-code session into the mesh

caffeinum:
## What

Adds an optional `resume` to the connector `LaunchOpts` so a `cotal spawn`/`start` can reattach a prior claude-code conversation instead of always starting a fresh session.

When set, the claude-code connector pushes `--resume <id> --fork-session` into the launch args (fork so the original session id isn't hijacked — we adopt its *context*, not its *identity*). All existing args (dev-channels, MCP config, persona, model) are preserved, so the resumed process still joins the mesh normally.

## Why

Today every spawn is a cold session — an operator doing real work in a claude session can't pull that context into the mesh. This is the minimal change to allow it. `claude --resume`, `--fork-session`, and `--strict-mcp-config` compose cleanly (verified against `claude --help`).

Closes #23.

## Changes
- `core`: `LaunchOpts.resume?` (connector-specific session id)
- `connector-claude-code`: emit `--resume <id> --fork-session` when set
- `manager`: `opStart` threads `args.resume` into `buildLaunch`; `--resume` flag on `start`
- `cli`: `--resume <id>` flag on `spawn`
- codex/opencode connectors ignore `resume` (claude-specific no-op)

## Test
Build + typecheck clean (`pnpm -r build`, `pnpm -r typecheck`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31, opened 2026-06-15
feat(resume): late-join an existing claude session into the mesh

### LINKED ISSUE #23: Feature: resume an existing claude-code session into the mesh instead of always spawning fresh
Today the claude-code connector's `buildLaunch` (`@cotal-ai/connector-claude-code/dist/extension.js`) always spawns a **fresh** `claude` on every `cotal start`/`spawn` — there's no `--resume`/`--continue`/`--session-id`, and `LaunchOpts` has no `resume` field, so an existing session's conversation, context, and in-flight task are lost. I'd like to pull a session I'm already working in (deep context, long transcript) into the mesh as a managed agent. A truly running process can't be retrofitted (MCP is fixed at launch and the connector isolates with `--strict-mcp-config`), so "resume" means replaying the *conversation* into a fresh, cotal-equipped process — which `claude` already supports composing with strict MCP:

```bash
# what buildLaunch could emit when a resume id is provided:
claude --resume <session-id> --fork-session \
  --strict-mcp-config --mcp-config '{cotal}' \
  --append-system-prompt <persona> --model <model>
```

Concretely: add an optional `resume?: string` to `LaunchOpts`, thread it through `opStart` + a `--resume` flag on `cotal spawn`, and have `buildLaunch` push `--resume <id>` (with `--fork-session` so the original id isn't hijacked). Small, well-scoped, and composes with the existing isolation model.

```bash
cotal spawn canary --resume <session-id>
```

Lanzelot1:
## What
Late-join an existing Claude Code session onto the mesh. A live `claude` can't be hot-attached (its MCP, hooks and wake channel bind at launch), so late-join means relaunching the session's history wired to the mesh.

It's a `--resume <id>` pass-through on **`cotal spawn`** (and `cotal start`) — claude-specific, no new command:
- **fork (default)** — `--resume <id> --fork-session`: replays the transcript into a fresh, mesh-equipped process under a **new** id, so the original keeps its identity and runs untouched.
- **`--in-place`** (spawn only) — plain `--resume <id>`: the **same** id/transcript continues, now mesh-wired (exit the original first — two live processes writing one transcript corrupts it).

Resume is claude-only; the other connectors ignore the flag (like `prompt`).

## Changes
- `core`: `LaunchOpts.resume?` + `LaunchOpts.fork?` (fork defaults on)
- `connector-claude-code`: emit `--resume <id>`, add `--fork-session` unless `fork === false`; skip the greeting on resume
- `cli`: `--resume` / `--in-place` on `spawn`
- `manager`: `--resume` on `start`; threaded into `buildLaunch`
- tests: `resume.smoke.ts` — claude argv composition + fork toggle
- docs: `claude-code-integration.md`, `getting-started.md`, `CLAUDE.md`

## Notes
- Based on `split/04-setup` (composes with the `--strict-mcp-config` isolation from #27). Rebase to `main` after #26-29 merge.
- Supersedes #24. Closes #23.


davidfarah2003:
Instead of writing a new resume command, we should have a pass-through for resume with spawn. This way we can just wire it to the claude code command and be specific to claude code, so that we don't interfere with other implementations and connectors.

davidfarah2003:
Closing as superseded: late-joining an existing Claude session landed on main via the connector's `--resume` path (in production use today: `cotal spawn --agent claude --resume <session-id>`). Thanks — the capability shipped, just through a later branch.

PR #35, opened 2026-06-16
fix(manager): unhang managed agents on Claude ≥2.1.178 (clear all startup prompts)

### LINKED ISSUE #34: Managed agents hang at `starting…` on Claude 2.1.178 — new --dangerously-load-development-channels warning isn't auto-cleared
## Summary

Managed agents never join the mesh on **Claude Code ≥ 2.1.178** — they sit at `starting…` forever and never register presence. Root cause: Claude changed the `--dangerously-load-development-channels` confirmation screen, so the manager's pty auto-confirm no longer clears it and the session hangs *before* the cotal channel/MCP ever loads.

## What changed in Claude

The connector launches `claude --dangerously-load-development-channels <ref>` and the manager's pty runtime auto-clears the one-time prompt by matching the text **`"Enter to confirm"`** (`connector-claude-code` sets `confirm: "Enter to confirm"`; `manager/runtime/pty.ts` watches for it for `CONFIRM_WINDOW_MS` and presses Enter).

Claude **2.1.178** no longer shows "Enter to confirm". It now renders a blocking warning and waits:

```
  WARNING: Loading development channels

  --dangerously-load-development-channels is for local channel development only. Do not use this option
  to run channels you have downloaded off the internet.

  Please use --channels to run a list of approved channels.

  Channels: server:cotal
```

The manager's matcher never matches this new text → Enter is never sent → claude stays on the warning screen → the cotal channel/MCP never initializes → no presence → the agent is stuck `starting…`. (Confirmed: `mcp-logs-cotal` is never written for a managed agent; `cotal ps` shows `starting…` indefinitely; the connector run standalone — `node dist/mcp.cjs` with `COTAL_*` env — connects and registers fine, so the mesh/connector are healthy and the break is purely the launch-time prompt.)

## Repro

1. Claude Code ≥ 2.1.178
2. `cotal up` + a manager, then `cotal start --name <agent>`
3. `cotal ps` → agent stays `starting…`, never `idle`

## Impact

All managed (`cotal start` / supervised pty) agents are dead on arrival on current Claude. `spawn` in a real interactive terminal still works because a human can clear the warning.

## Notes toward a fix

- The warning text itself hints at the intended path: **`Please use --channels to run a list of approved channels.`** Claude appears to want an *approved channels* list (`--channels`) rather than the dev-only `--dangerously-load-development-channels`. Migrating the connector to `--channels <approved-ref>` (if it accepts the installed `plugin:cotal@cotal-mesh`) would avoid the warning entirely.
- Short term, the pty auto-confirm matcher could be widened to also match the new warning (e.g. `"Loading development channels"` / `"--channels"`) and send Enter — but that's brittle against further Claude changes; the `--channels` migration is the durable fix.
- `upstream/main`'s connector already moved the channel ref to `plugin:cotal@cotal-mesh` but still sets `confirm: "Enter to confirm"`, so it is **also affected** until the prompt handling is updated.

Environment: Claude Code 2.1.178, macOS (darwin arm64), cotal pty runtime, open mesh.

Lanzelot1:
## Problem (#34)
Managed (pty) agents hang at `starting…` on Claude Code ≥2.1.178 and never join the mesh.

## Root cause
On a fresh/untrusted workspace, Claude ≥2.1.178 shows **two** back-to-back confirm screens at startup, both footered `Enter to confirm · Esc to cancel`:
1. the workspace **trust** prompt, then
2. the **dev-channels warning** (`--dangerously-load-development-channels`).

The pty runtime's auto-confirm was **one-shot**: its single Enter cleared the trust prompt and the channels warning then blocked forever — so the cotal MCP never loaded, presence never registered, and `cotal ps` stayed `starting…`.

## Fix
`implementations/manager/src/runtime/pty.ts`: make the auto-confirm **timer-driven, firing once per prompt** instead of one-shot. While the `Enter to confirm` footer is on screen during the startup window, press Enter, clear the buffer so the same screen isn't pressed twice, and re-arm for the next — capped so it can never loop. No connector change.

## Verified on real Claude 2.1.179
- **Captured startup**: a fresh workspace shows trust → channels-warning, both with `Enter to confirm`; two Enters reach the live UI (channel injected). No further approval gate.
- **A/B via `pnpm smoke:confirm`**: the **old** runtime stalls on the channels warning; the **fixed** runtime clears both and reaches the live UI. ✅
- `pnpm build` + `pnpm typecheck` green.

`smoke:confirm` skips cleanly when no `claude` is available; point it at a build with `COTAL_TEST_CLAUDE_BIN=/path/to/claude`.

Closes #34.

### DIFF
--- a/implementations/manager/channels-confirm.smoke.ts
+++ b/implementations/manager/channels-confirm.smoke.ts
@@ -1,0 +1,68 @@
+/**
+ * Issue #34 regression: a managed (pty) Claude session hangs at "starting…" on Claude Code
+ * ≥ 2.1.178 because startup now shows TWO back-to-back "Enter to confirm" screens (workspace
+ * trust, then the dev-channels warning) and the old one-shot auto-confirm cleared only the first.
+ *
+ * This drives the REAL PtyRuntime with the connector's LaunchSpec against a real `claude`, in a
+ * fresh (untrusted) workspace so both prompts appear, and asserts the session gets past them to
+ * the live UI instead of stalling on the channels warning.
+ *
+ * Run: COTAL_TEST_CLAUDE_BIN=/path/to/claude-2.1.178 pnpm smoke:channels
+ * Skips cleanly if no claude binary is available (e.g. CI without auth).
+ */
+import { spawnSync } from "node:child_process";
+import { mkdtempSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { PtyRuntime } from "./src/runtime/pty.js";
+
+const claudeBin = process.env.COTAL_TEST_CLAUDE_BIN ?? "claude";
+const probe = spawnSync(claudeBin, ["--version"], { encoding: "utf8" });
+if (probe.status !== 0) {
+  console.log(`SKIP channels-confirm smoke — no runnable claude at "${claudeBin}"`);
+  process.exit(0);
+}
+console.log(`claude: ${probe.stdout.trim()}`);
+
+const strip = (s: string): string =>
+  s
+    .replace(/\x1b\][^\x07\x1b]*(?:\x07|\x1b\\)/g, "")
+    .replace(/\x1b\[[0-9;?]*[ -/]*[@-~]/g, "")
+    .replace(/\x1b[@-Z\\-_]/g, "");
+
+// Mirrors the claude connector's launch recipe (origin/main): the dev-channels flag (which on a
+// fresh workspace triggers trust + warning prompts) and the confirm token the runtime auto-clears.
+const spec = {
+  command: claudeBin,
+  args: ["--dangerously-load-development-channels", "plugin:cotal@cotal-mesh"],
+  env: { COTAL_SPACE: "smoke34", COTAL_NAME: "smoke34", COTAL_CHANNEL: "1" },
+  confirm: "Enter to confirm",
+};
+const cwd = mkdtempSync(join(tmpdir(), "cotal-ch34-")); // fresh dir ⇒ untrusted ⇒ trust prompt fires
+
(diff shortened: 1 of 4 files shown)


davidfarah2003:
Reviewed this against real Claude 2.1.181 and opened #45 as a simpler alternative for the same bug.

**Reproduced the root cause** by driving the real `PtyRuntime` in a fresh untrusted dir: the one-shot auto-confirm clears the trust gate then sits frozen on the dev-channels warning (identical screen t=2s→t=16s). Both gates are 2-option menus that default-highlight "proceed", and the channels screen is **static** once rendered (no further output), which is why an output-driven matcher can't re-fire on it.

**Where this PR is fragile:**
- It matches the literal `"Enter to confirm"` footer. If Anthropic changes that wording, it silently hangs again — the worst failure mode (this is exactly what upstream [anthropics/claude-code#58152](https://github.com/anthropics/claude-code/issues/58152) warns about).
- It clears its match buffer after each press, so a dropped first press on a static screen is never retried.

**#45 instead** presses Enter on a short blind timer during startup (capped), mirroring what the cmux runtime already does: wording-independent, self-retries a dropped press, handles the variable gate count (a re-launch on a trusted folder shows only the channels gate), and deletes ~40 lines of match/cooldown/buffer machinery. Verified end-to-end: cleared both gates, reached the live UI with the cotal channel injected, extra Enters were no-ops.

Proposing we close this in favor of #45. Thanks @Lanzelot1 for the precise root-cause writeup — it made validating the fix fast.

davidfarah2003:
Closing in favor of #45, which fixes the same #34 hang with a simpler, wording-independent blind-timer approach (verified end-to-end on Claude 2.1.181) and brings the pty runtime to parity with cmux. Thanks again @Lanzelot1 for the root-cause writeup.

PR #101, opened 2026-06-23
refactor(cli): make all mesh commands work from any directory

davidfarah2003:
## What

**Stacked on #98.** Migrates the remaining CLI commands that connect to a running mesh off the old cwd-walk resolution onto the shared `resolveMeshTarget` + preflight — so `send`/`dm`/`msg`/`ask`, `console`, `join`, `web`, `channels`, `history`, and `personas --running` all work from **any** directory, not just inside the project that ran `cotal up`.

Before, each re-derived creds from `loadSpaceAuth(authDir(cotalRoot()))` + `DEFAULT_SERVER`; from outside the project that walk-up lands on `$HOME/.cotal` and the command crashes with a raw NATS Authorization Violation — the exact bug `cotal spawn` fixed in #98, still live in every other connecting command ("spawn works, watch doesn't").

## How

- **New `lib/connect.ts`** — `resolveTargetOrExit` + `preflightOrExit`, extracted from `spawn.ts` so every mesh-touching command resolves the same way and emits the same one-sentence preflight instead of a raw trace. `preflightOrExit` takes optional probe creds (for the `--creds` escape hatch).
- **`lib/transient.ts`** (`resolveConnect`/`openTransient`) now uses the shared helper → migrates `send`, `history`, `channels`, `personas --running` in one move.
- **`web` / `console` / `join`** migrated, each keeping its escape hatches: `web`/`console` keep `--creds` and `console`'s open-mode space overview (space stays undefined with no `--space`); `join` keeps the join-link / `--token` / `--creds` paths and only the bare case resolves + self-mints.
- **`spawn.ts`** now imports the shared helper (de-duplicated; the local copies are gone).

`up` / `down` / `setup` / `mint` / `demo` stay on `cotalRoot()` — they operate on the local project by design.

## Hygiene (deferred review notes, folded in)

- **`targetFromEntry`** (mitnick): prune + fail loud when an AUTH entry's recorded space no longer matches its root's on-disk `auth.space` — never mint space-Y creds against the entry for space X.
- **`resolveTargetOrExit`** (fowler): a dim `default "X" is down — using "Y"` note when a dangling `current` is silently bypassed (the N>1 case already errors loudly; this was the one quiet redirect).
- **`cotal meshes` / `cotal down`** (norman): a dangling-`current` note in the listing; `down` summary clarified as local-only.
- **Not folded:** "`removeMesh` self-clears `current`" (fowler) — socrates showed it would *widen* the exit-handler clobber guarded in #98 (round 8 `f4de2fa`), so `clearCurrent` stays at identity-aware call sites.

## Tests

- Hermetic smoke now **24 checks** (+ the space-mismatch prune); live-broker e2e still **9/9**; `pnpm typecheck` green across all 16 projects.

## Note

Targets `feat/spawn-from-anywhere` (#98) since it builds on that PR's resolver + the round-8 exit-handler guard. Rebase onto `main` once #98 merges.

### DIFF
--- a/implementations/cli/smoke/connect.smoke.ts
+++ b/implementations/cli/smoke/connect.smoke.ts
@@ -1,0 +1,71 @@
+/**
+ * The preflight decision tree — `classifyPreflightFailure` (lib/connect.ts). This is the riskiest
+ * new logic in the command migration: a wrong branch PRUNES A LIVE registry entry. Pure + offline,
+ * so every (source × reason × has-auth) combination is exercised here without a broker. The
+ * load-bearing invariant: a NON-registry source (flag-server / local-space, or a raw `--creds`
+ * connection) is never pruned — only the registry owns its entries. Run: pnpm smoke:connect
+ */
+import { strict as assert } from "node:assert";
+import { classifyPreflightFailure } from "../src/lib/connect.js";
+
+let pass = 0;
+const check = (name: string, cond: boolean, extra?: unknown) => {
+  assert.ok(cond, `${name}${extra !== undefined ? ` — ${JSON.stringify(extra)}` : ""}`);
+  pass++;
+  console.log(`  ✓ ${name}`);
+};
+
+const REGISTRY = ["registry", "current", "flag-space", "local-recorded"] as const;
+const NON_REGISTRY = ["flag-server", "local-space"] as const;
+
+// unreachable: a registry-owned target prunes (broker gone); a non-registry one never does.
+for (const s of REGISTRY)
+  check(
+    `unreachable + ${s} → prune + 'unreachable'`,
+    (() => {
+      const r = classifyPreflightFailure(s, "unreachable", true);
+      return r.prune === true && r.kind === "unreachable";
+    })(),
+  );
+for (const s of NON_REGISTRY)
+  check(
+    `unreachable + ${s} → NO prune + 'unreachable'`,
+    (() => {
+      const r = classifyPreflightFailure(s, "unreachable", false);
+      return r.prune === false && r.kind === "unreachable";
+    })(),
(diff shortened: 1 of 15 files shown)


davidfarah2003:
## Review fix round (all six lenses addressed)

Pushed `b1d5292` + `f1681e3`:

- **channels + history now genuinely migrated** (were claimed but missed) — both route through a new shared `connectOrExit`; `channels` gains the `--creds` option it lacked. Destructive `history clear` / privileged `channels set` no longer blindly target `DEFAULT_SERVER` + the cwd space from outside the project.
- **Escape hatch restored** (review-general-2 #1/#4): explicit `--creds` → a RAW off-registry connection (bypasses `resolveMeshTarget`), so `--server remote --space remote --creds x` works for an unregistered/remote mesh again. This *also* fixes **prune-on-bad-creds** (#2): caller creds never reach the registry-prune path.
- **join's raw crash fixed** (socrates): the explicit path preflights with `reachableOrExit` (probeConnect) instead of `isReachable` (which returned true on an auth reject → raw `ep.start()` trace).
- **console --plain** streams the resolved mesh's space, not `resolveSpace(cwd)` (norman).
- **Tests** (truthium — the preflight tree had zero coverage): extracted the branch logic to a pure `classifyPreflightFailure`; new `smoke:connect` exercises every (source × reason × has-auth) combo incl. the "non-registry never prunes" invariant (18 checks). Live e2e adds raw-path reachability (now 10).

Deferred per scope: the `web` account-seed exposure (socrates, pre-existing) → filed as #102.

typecheck green (16 projects), `smoke:spawn-from-anywhere` 24, `smoke:connect` 18, live e2e 10.

PR #106, opened 2026-06-24
feat(cli): re-export lifecycle helpers for external composition roots

caffeinum:
## Why
`@cotal-ai/cli` restricts its package `exports` to `.`, so a downstream composition root (e.g. one that wants cotal to own its mesh + manager lifecycle) can't import the detached-daemon helpers via a subpath like `@cotal-ai/cli/commands/up.js` — the import is blocked.

## Change
Re-export the lifecycle helpers from the package index (additive only):
- `startMeshDetached` (from `./commands/up`)
- `ensureManager`, `managerUp` (from `./lib/manager-proc`)

`isReachable` (core) is already exported from `@cotal-ai/core`. No behavior change.

### DIFF
--- a/implementations/cli/src/index.ts
+++ b/implementations/cli/src/index.ts
@@ -175,3 +175,9 @@ registry.register(...baseCommands);
 
 export { runCli } from "./command.js";
 export { c, statusBadge } from "./ui.js";
+
+// Lifecycle helpers for an external composition root (e.g. paw) — this package's exports are
+// restricted to ".", so subpath imports are blocked; re-export from the index instead.
+// `isReachable` (core) is already exported from @cotal-ai/core.
+export { startMeshDetached } from "./commands/up.js";
+export { ensureManager, managerUp } from "./lib/manager-proc.js";


davidfarah2003:
Thanks for this, and for the clear writeup of the motivation.

We're going to pass on this one — not for the mechanics (the re-export itself is correct), but for the direction. Three concerns, the first being the real blocker:

**1. `ensureManager`/`managerUp` don't actually work outside the `cotal` binary.** `ensureManager` → `startManagerDetached` re-execs `selfArgv()` (`[execPath, …, process.argv[1]]`) with `supervise`. `argv[1]` is the *host* entry script, and `supervise` is only registered by the composed `cotal` binary. Imported into an external composition root, this re-execs that root's own entry with a subcommand it doesn't handle — so it exports a contract that misfires for the very consumer it's meant to serve.

**2. Tier / dependency direction.** Per AGENTS.md, `implementations/*` are self-contained surfaces and reuse flows through `core` — which is exactly why `isReachable` is importable from `@cotal-ai/core`. An external composition root reaching into `@cotal-ai/cli`'s internals inverts that. And these aren't primitives: `startMeshDetached` is the `up --detach` body (it writes the mesh registry, provisions `.cotal/` auth+creds, resolves the bundled nats binary, and boots the delivery daemon). Re-exporting it freezes the `.cotal/` on-disk layout, the detach opts/return shape, and the delivery-daemon coupling as public API — and under the fixed lockstep version group that's a one-way door: changing *or even removing* it later is a major bump across every `@cotal-ai/*` package.

**3. (Minor)** the comment names an out-of-repo project, and there's no changeset for the new public surface.

Recommended path: an external composition root should drive the public CLI directly — `cotal up --detach`, and the manager via the registered `supervise` command. That needs no new API, works today, and gives you the fully-composed behavior `ensureManager` already depends on. If an in-process library seam is genuinely wanted, the right move is to lift the generic primitive (boot nats + provision streams) into `@cotal-ai/core`, where the tier permits reuse, and keep the registry/`.cotal` bookkeeping private — happy to look at a PR along those lines.

Closing for now; thanks again for the contribution.

Issue #159, opened 2026-07-01
Harden detached launch-failure handling (resume mint-leak + async early-exit surfacing)


davidfarah2003:
Surfaced during the mesh review of #23 (session resume). These are **pre-existing manager gaps**, but session-resume is the first feature to make them both operator-reachable *and* retry-amplified, so tracking them as a scoped follow-up.

Two related gaps, one root: on a detached `cotal start` / MCP `cotal_spawn`, `startAgent` returns `{ok:true}` the moment `runtime.spawn()` hands back a PTY handle — **before** the child (`claude`) actually runs.

### 1. Unsupported-connector resume mints before it fails (early-fixable)
`buildLaunch` throws for `opencode`/`hermes` resume *inside* `startAgent`'s try — i.e. **after** `provisionAgent` mints scoped creds + broker durables (auth mode). So `cotal start --agent opencode --resume x` fails loud but late, orphaning the minted DM/DLV/role-TASK durables + a creds file for a peer that never launched.

**Fix:** a declared `supportsResume` capability on the `Connector` SPI + a central early preflight check (before the reserve/mint, beside the binary/transcript prechecks). This also makes the default **fail-loud for any future connector** that neither declares support nor reads `opts.resume` — closing the fallback-by-omission hole that `hermes` exemplified (it references `opts.resume`/`opts.mcpServers` nowhere; the guard is hand-plugged today). The `LaunchOpts.resume` doc-comment is the type-level contract; the capability is the runtime-enforcement half TS can't provide.

### 2. Supported-path async early-exit is neither surfaced nor cleaned up (the dominant leak)
On the claude path a bad id / cwd-mismatch / old-CLI-rejecting-`--fork-session` makes `claude` exit *inside its PTY*, and:
- **(i) not surfaced:** the caller already saw `✓ started`; the peer just never joins the roster. Diagnosable only via `cotal attach`.
- **(ii) leaks under auth:** `identity = newIdentity()` is a fresh nkey **per attempt**, and `onAgentExit → freeSlot` does `agents.delete` + a cooling stamp only — **no deprovision, no creds unlink**. The creds *file* is per-name (bounded/overwritten), but the broker durables are per-identity, so each failed attempt orphans a distinct durable set. Retry-amplified: debugging a typo'd id retries 5–10× → 5–10 orphaned durable sets.

[...]

PR #180, opened 2026-07-03
fix(connector-core,cli): launch clients outlive the manager's readiness wait

davidfarah2003:
## What

#159 B1 made the manager reply to `start`/`launch` only on a **real outcome** — presence join, process exit, or its ~30s readiness backstop — but two launch clients still used `requestControl`'s 5s default, so a real spawn of a slow-booting agent died client-side (`no manager reachable (timeout)`) while the launch actually proceeded:

- **`MeshAgent.spawn`** (connector-core) — the MCP `cotal_spawn` door.
- **`launchAgent`** (cli, `lib/manifest/live`) — `cotal spawn -f` deploying agents onto a running mesh (`launch` funnels into the same `startAgent` readiness wait).

The CLI's `spawn --detach` was already fixed in the rework train; this closes the remaining doors and turns the relation into an enforced contract.

## How

- **connector-core**: exported `SPAWN_TIMEOUT_MS` (40s), passed on the `start` op.
- **cli**: exported `START_TIMEOUT_MS` (40s) in `lib/control` — `spawn --detach`'s literal moves onto it, `launchAgent` now passes it.
- **manager**: exported `READINESS_TIMEOUT_MS` (30s) so the relation is testable — the tier rule forbids the clients importing it directly.
- **launch-parity smoke** (new check 4): every launch client's window must **outlive** the manager's readiness wait, and `launchAgent` must actually pass it (fake ep, no NATS).
- **spawn-args smoke**: `MeshAgent.spawn` must carry `SPAWN_TIMEOUT_MS` — and the smoke now gates in `smoke:ci` (it previously ran nowhere).

## Testing

Red/green proven: with the call sites unwired, the new spawn-args check and the launch-parity `launchAgent` assert both fail; wired, all green. Locally: `pnpm typecheck`, `pnpm test`, cli-kernel / flag-inventory / start-overrides / launch-parity / spawn-args smokes, and the `spawn-detach` (20), `up-manifest` (15), `spawn-manifest` (31) live e2e suites.

Found while validating (NOT addressed here, pre-existing on main): `smoke:manifest-launch` is broken and gated nowhere — its auth-mode `startAgent` provisions against the default `127.0.0.1:4222` (`servers: undefined`), so it dies with `Authorization Violation` next to a live broker (connection-refused elsewhere).


davidfarah2003:
Added a live e2e (`smoke:readiness:live`, gated in `check` + the CI live step): a real broker + real manager + an agent that deliberately takes ~7s to join the mesh — past the 5s op default, inside the ~30s readiness backstop. Both fixed doors (`MeshAgent.spawn` and `launchAgent`) now get the manager's real join-outcome reply at ~7.1s; red-proven live — with the timeouts unwired, both die client-side with `TIMEOUT` at 5s while the launch proceeds.

PR #182, opened 2026-07-04
feat(pi): TUI launch mode + persona/model forwarding

TheReaperJay:
feat(pi): TUI launch mode + persona/model forwarding

Stacked on #126 — merge #126 first, then retarget this PR's base to `main`.

This adds a robust persona/model forwarding for an interactive TUI mode, allowing each terminal to be watched and manually interacted with as opposed to just the default headless behavior. This is useful for micromanaging agents, or for when the user wants more control or oversight of what is produced. These do not overwrite the existing #126 but instead extend the available control options. 

Any pi extension that uses ctx.ui is now able to be natively surfaced in Cotal, allowing for wider compatibility of workflows and easier onboarding for devs.

The original headless mode from #126 is not edited at all, the TUI mode is a synergistic extension.

It also introduces a robust fix for model and prompt injection that was overlooked in the original PR.

## Usage

Using this is very simple, just update your agent with the peerMode:

```markdown                                                                                                                                                                                       
   ---                                                                                                                                                                                             
   name: worker                                                                                                                                                                                    
   role: builder                                                                                                                                                                                   
   allowPublish: [builds]                                                                                                                                                                          
   peerMode: headless    ## Run headless as usual                                                                                                                                                                          
   ---                                                                                                                                                                                             
   Runs unattended; no UI gates.                                                                                                                                                                   
 ```                                                                                                                                                                                               
                                                                                                                                                                                                   
 ```markdown                                                                                                                                                                                       
   ---                                                                                                                                                                                             
   name: helper                                                                                                                                                                                    
   role: pair                                                                                                                                                                                      
   allowPublish: [general]                                                                                                                                                                         
   peerMode: tui    ## Run with interactive TUI                                                                                                                                                                               
   ---                                                                                                                                                                                             
   Edit-approval gates render live in your tmux pane.                                                                                                                                              
 ```

You can also run it ad hoc without a configuration change:
```bash                                                                                                                                                                                           
   cotal up --detach                                # mesh + delivery + manager (no env needed here for a foreground spawn)                                                                        
   PI_PEER_MODE=tui cotal spawn pi1      # foreground TUI pi peer in THIS terminal, assuming "pi1" is a pi agent
   cotal console                                                                                                                                                                                   
   cotal down                                                                                                                                                                                      
 ```

If you do not activate TUI mode, it simply falls back to using headless Pi, making the original PR the default behavior.

## What

Two additions to the `@cotal-ai/pi` connector introduced in #126, both closing gaps where a spawned pi peer ran as a generic pi instead of its declared persona/model, with no operator-answerable surface.

### 1. TUI launch mode (`PI_PEER_MODE=tui`)

The headless peer in #126 embeds pi in-process and never binds a UI context, so `ctx.hasUI` is `false` and any installed extension that calls `ctx.ui.*` (approval gates, prompts, selects, editors) silently no-ops. This adds a second entry point that acts as a host process: it spawns stock `pi --mode rpc` as a child and bridges to it over the NDJSON stdin/stdout protocol pi already exposes, rendering every `ctx.ui.*` call live in the peer's own tmux pane. Any pi extension becomes operator-answerable per-pane (an edit-approval gate is the canonical use case). Zero coupling to `@earendil-works/pi-tui` — the renderer uses Node builtins only.

Three Esc behaviors: in-dialog = cancel, while streaming = `{type:"abort"}`, idle = no-op. Persistent raw mode so Esc is captured by the keypress handler rather than eaten by cooked-mode line discipline. The connector uses `node` (not `tsx`) for built `dist/` so `cotal stop`'s SIGTERM reaches the peer and triggers graceful shutdown + cooked-mode restore.

### 2. Persona + model forwarding (both paths)

The pi connector was the lone holdout among the harness connectors — claude-code, opencode, and hermes all translate `AgentDef.persona` → system prompt and `AgentDef.model` → the host's `--model`. Now `buildLaunch` is the single parse point (`loadAgentFile` at launch → fail-loud on a malformed persona; `cotal start --model` beats the file's `model:`); it forwards only the short resolved model string (`COTAL_MODEL`), and each runtime path reads the persona body from the already-forwarded `COTAL_AGENT_FILE`:

- **TUI:** body → 0600 tempfile → `--append-system-prompt <path>` (pi reads the file via `resolvePromptInput`, so a large persona never hits an argv ARG_MAX ceiling); `--model COTAL_MODEL`; tempfile unlinked on every exit path.
- **Headless:** `DefaultResourceLoader({appendSystemPrompt:[body]})` (pi's in-process equivalent, preserving extension/skill/theme loading) + a `resolveModel` helper on public `ModelRegistry.getAvailable()` / `Model.provider` / `Model.id`, mirroring pi's internal `findExactModelReferenceMatch` contract. Throws on not-found/ambiguous — no fuzzy, no silent fallback.

Intentional asymmetry, documented in code: TUI accepts fuzzy model patterns (pi's CLI resolves them); headless requires an exact reference and throws on fuzzy/ambiguous (a declarative persona `model:` pin should be exact; pi doesn't export a public fuzzy resolver).

`docs/agent-frameworks.md` is updated in the same change — corrects the now-stale "pi has no permission gate" and "launches via tsx" lines from #126, and adds the persona/model forwarding.

## Verification

End-to-end through the real supervisor→connector→spawned-peer path: a haiku persona produces haiku replies in the TUI pane and lands in the headless session's system prompt; the pinned model is forwarded (`COTAL_MODEL` in the child env) and honored by pi; a bogus or ambiguous model makes headless throw at startup (no fallback to pi's default); persona tempfiles cleaned on peer exit; Esc aborts a streaming turn; SIGTERM (`cotal stop`) triggers graceful shutdown + cooked-mode restore; `pnpm smoke:ci` green; `pnpm typecheck` + `pnpm build` clean; no `@earendil-works/pi-tui` coupling; no deep import of pi internals.

## Commits

- `feat(pi): PI_PEER_MODE dispatch seam (tui/headless)`
- `feat(pi): rpc bridge + child spawn for the TUI launch mode`
- `feat(pi): add TUI launch mode for operator-answerable mesh peers`
- `feat(pi): forward persona + model into spawned pi peers (TUI + headless)`
- `feat(pi): read peerMode from agent file frontmatter]`


davidfarah2003:
Closing as superseded by the landed Pi connector. Current main's `extensions/pi` already forwards persona and model in its launch builder (with smoke coverage for precedence and `--append-system-prompt`/`--model` placement), and ships the Pi TUI/wrapping integration. This branch predates that implementation and now spans the repository/control-surface rewrites; any residual Pi launch-mode gap should be a focused PR against current `extensions/pi`.

PR #197, opened 2026-07-07
feat(console): json-render views lens + activity sparkline

Lanzelot1:
## What

Consolidates the console-TUI work (previously scattered across `feat/console-json-render` and uncommitted local iterations) into one reviewable branch, rebased onto current main.

- **View parts on the wire** (`@cotal-ai/core`): a peer can publish a renderable view — a json-render flat spec — as a `{ kind: "view", spec }` part on an ordinary message via `endpoint.publishView(spec, { channel | to | toService })`. No new subject; rides all three delivery modes with a plain-text label part alongside. SPEC.md gains the third core `Part` shape; `spec/cotal.schema.json` regenerated.
- **Views lens in the console** (`V`): renders peer-published views through a fixed `@json-render/ink` component catalog — declared components only, validated props, never code; an invalid spec shows its rejection reason. The golden-signal tiles strip now renders through the same catalog (dogfoods the guardrail).
- **Activity sparkline**: MeshView keeps a 15-bucket / 60s message-volume series; the status bar draws it next to msgs/s.

## Verified

- `pnpm typecheck` + `pnpm test` green; view/connect/spawn-from-anywhere smokes pass.
- Live TUI against a running mesh: `publishView` → V lens renders the panel; a spec naming an unknown component (`Script`) is rejected with the reason; sparkline animates with traffic.

## Notes

- Supersedes branch `feat/console-json-render` (same feature, pre-drift base) — that branch can be deleted after merge.
- Overlaps `feat/tmux-extension` only on the identical `Sparkline.tsx`; merges cleanly either order.

(The example-02 demo assets that previously rode along were split into #212 to keep this PR focused on the feature.)


davidfarah2003:

PR #198, opened 2026-07-07
feat(console): registry-first run-anywhere overview with a mesh picker

Lanzelot1:
## What

A bare `cotal console` now resolves from the mesh registry **only** — never the cwd — so a stray local `.cotal/` can neither hijack the target nor block it with `default-occupied` ("another mesh is running at …"). With several meshes registered, a TTY opens a mesh overview of every broker on the machine (space · server · mode · root); pick one to drill into its spaces, `b` walks back up. Piped/`--plain` still resolves a single target (`current`, else a clear ambiguous error), and `--space`/`--server`/`--creds` behave exactly as before.

- `@cotal-ai/workspace`: new `resolveRegistryTarget()` (registry-only, no cwd) and `registryTarget(entry)`; `resolveMeshTarget` shares the registry tail.
- `@cotal-ai/cli`: new `MeshPicker` level above the space picker; per-mesh connect failures render inline in the picker instead of exiting the TUI; the same admin-mint + preflight + stale-prune rules apply on entry.

## Why

The console is the read-only run-anywhere view. Local-project-wins target precedence makes sense for write commands (`spawn`), but for the console it meant a leftover `.cotal/` dir anywhere up the tree could block the whole overview.

## Verified

- `pnpm typecheck` + `pnpm test` green; preflight/connect/view smokes pass; spawn-from-anywhere smoke extended with registry-only resolution checks (34 checks).
- Live: from a directory with a stray `.cotal/`, `cotal console` joins the registered mesh instead of erroring; with two registered meshes the picker → space overview → space console → back chain works against a live broker.

### DIFF
--- a/docs/watch-a-mesh.md
+++ b/docs/watch-a-mesh.md
@@ -28,11 +28,14 @@ cotal console                    # no --space on an open mesh → the admin over
 
 ![The cotal console: a live roster of agents and their all-activity feed in a terminal TUI](../assets/quickstart.gif)
 
-**Admin overview.** On an open mesh, `cotal console` with **no `--space`** opens a space picker:
-every space on the server (enumerated from its `CHAT_*` streams and presence buckets) with its
-agents, channels, and message counts. Pick one to drop into its console; `b` returns to the
-overview. `--space X` skips the picker. Under auth a server hosts a single space, so the console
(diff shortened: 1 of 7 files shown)


davidfarah2003:

PR #200, opened 2026-07-07
feat(console): full operator control — fix kill, add pause/resume, spawn/purge/status

Lanzelot1:
## What

Makes the console a real "lazygit for Cotal" operator surface. **Stacked on #197** (`feat/console-views`) — retarget to `main` once #197 merges.

### The bug this fixes
`D`-kill sent `requestControl("manager", …)` — the *privileged* tier, where the manager only stops your own spawned child — using the console's self-minted god-view `admin` cred, which has **no `ctl.*` publish grant**. So on any auth mesh, kill silently did nothing. Now every control action mints a **per-action, tier-scoped caller cred** from held trust material (open mesh: bare; `--creds`: as-is) and targets the right tier — the same pattern the CLI (`askManager`) and web (channel-purger) already use. Gated on a new `canControl`, distinct from chat-write.

### New capability: pause/resume (new wire ops)
Freeze an agent in place — `SIGSTOP`/`SIGCONT` to the pty process group (POSIX only; tmux/cmux have no owned pid and win32 has no job-control signals, so both error loud, no fallback). Same authority as a named stop. Paused-ness lives in the manager's **managed-state** (`ps`/`status`), not presence — a stopped agent can't heartbeat, so its presence would TTL to "offline" and lie; the console merges `ps` by id and shows a `⏸ paused` badge. A graceful stop of a paused agent resumes it first (SIGTERM isn't delivered to a stopped process).

### Console affordances (graduated destructiveness)
- `p` on a roster agent → pause/resume, no confirm (recoverable).
- `D` → kill behind a confirm: `y` graceful stop, `f` force-kill.
- `:` palette: `spawn <persona> [name]`, `status <agent>`, `ps`, `purge` (type-the-name confirm).
- Every action flashes its result; StatusBar/Help updated.

## Verified
- `pnpm typecheck` + `pnpm test` clean; view/connect/lifecycle-e2e smokes pass.
- **New `pause-resume-live` smoke (15 checks)** against a real broker + manager + process: real OS freeze (state `T`), tier denial for a non-spawner (regression guard for the old kill bug), `control-caller-admin` cred scoped to `ctl.admin` only, unsupported-runtime error, and graceful-stop-of-paused (SIGCONT-first).
- Live TUI: kill confirm renders the graceful/force choice; `p` and `:ps` reach the manager end-to-end (an old v0.10 manager on the running mesh returns `unknown op: pause`, confirming the console wire works — the new op is proven by the smoke).

## Notes
- No SPEC/schema change: control op vocabulary is implementation-defined (SPEC §5 defines only the envelope).
- `docs/architecture.md` + `docs/protocol-view.md` updated in-change.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -138,10 +138,19 @@ and configures them.
   isn't loaded, never a silent fallback); **byo** is the floor (a human's own terminal,
   tracked via presence); **host** (Agent SDK, true mid-turn interrupt) is the documented
   upgrade path ([roadmap](roadmap.md)).
+- **Pause and resume.** Optional `pause`/`resume` on the Runtime contract freeze an agent
+  in place (POSIX `SIGSTOP`/`SIGCONT` to its process group). Only the `pty` backend owns a
+  real process, so only it implements them (POSIX only; Windows/ConPTY errors loud);
+  `tmux`/`cmux` reply "not supported by the `<kind>` runtime" rather than degrading.
+  Authority matches a named `stop` (privileged tier reaches your own child, admin any).
+  Paused-ness lives in the manager's managed state (`ps`/`status` report `paused`), not
+  presence: a SIGSTOPped agent cannot heartbeat, so its presence key TTLs out to `offline`;
+  surfaces merge `ps` by id instead (the console shows a `⏸ paused` badge). A graceful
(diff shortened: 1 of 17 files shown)


davidfarah2003:

PR #206, opened 2026-07-07
feat(console): operator participant mode — agents can reply to you

Lanzelot1:
## What

Sending from `cotal console` was **one-way**: the console joins as an invisible observer (`registerPresence:false`) with no inbox (`consume:false`), so an agent's DM reply to the operator had no consumer to land in and expired unread. Channel posts already round-tripped (agents reply in-channel, the tap shows it); DMs/`:call`/`:ask` didn't.

Now, on the operator's **first send**, the console upgrades the observer **in place** (no rebuild, no provisioning, no manager):
- **Presence** — a new `CotalEndpoint.startPresence()` registers the operator as a `role:"operator"` peer, so agents see it on the roster and can reply. Clean offline leave on exit (reuses the existing stop path).
- **Receive** — `MeshView` reveals the DM lens and captures the operator's own inbox. On an auth mesh (chat-narrowed tap) it opens a second tap scoped to `inst.<selfId>.>`; on an open mesh the whole-space tap already covers it (so no second tap — avoids a double-count). A best-effort one-shot backfill catches a reply that raced activation.
- Outbound DMs are recorded locally (auth path) so a thread shows **both sides**.

**Gating / reach:** `canWrite`-gated, so a pure-watch session never registers (stays invisible). It works where the console can actually send — an **open mesh** (bare) or an auth mesh with a capable **`--creds`**. Under the auth-default read-only god-view cred the console can't send at all, so participant mode never activates there; full auth-default participation would need a dedicated `participant` cred profile (future work, deliberately out of scope to keep this minimal and not touch the core cred matrix).

**Stacked on #200** (`feat/console-control`) — retarget to main after it lands.

## Verified
- `pnpm typecheck` + `pnpm test` + `smoke:view` clean.
- **Live end-to-end** (open broker + a stub agent that acks any DM): before activation the operator is off the roster; after a `:dm`, it registers (the stub sees it), the agent's `ack` reaches the operator's DM lens as a two-sided thread, and there's exactly one distinct outbound (no double-count). 5/5.
- **tmux click-through**: `:dm @stub hi there` → operator joins the roster as an endpoint peer, the feed shows the reply, the DM lens shows `user → stub` / `stub → user: ack`; on `q` the operator deregisters cleanly (only the agent remains live).

## Core change
One small, generic method — `CotalEndpoint.startPresence()` (post-construction presence + heartbeat, idempotent) + making `doRegister` mutable + a scoped `inboxHistory(selfId)`. Everything else is console/MeshView.

### DIFF
--- a/docs/mesh-view.md
+++ b/docs/mesh-view.md
@@ -16,7 +16,9 @@ presence) lives in the [SPEC](../SPEC.md).
 Every surface is built on one **read-only observer**: a `CotalEndpoint` started with
 `consume: false, registerPresence: false, watchPresence: true`, invisible to peers, binding no
 durables, reading the space through the live tap plus history and presence-watch. No surface opens
-its own NATS connection, and none re-implements the wire semantics.
+its own NATS connection, and none re-implements the wire semantics. (The console can upgrade its
+observer into a **participant** the moment the operator sends, so agents can reply — see
+[watch a mesh](watch-a-mesh.md); a pure-watch session stays the invisible observer.)
 
 ## The model: `MeshView` (`@cotal-ai/cli`)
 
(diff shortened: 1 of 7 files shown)


davidfarah2003:

PR #208, opened 2026-07-07
feat(console): authoritative membership in the topology lens

Lanzelot1:
## What

Closes the one real gap where the web dashboard beat the console. The topology lens (`t`) was **traffic-derived only** — a fold of the last ~120s of messages — so it couldn't show a **subscribed-but-silent** agent, couldn't distinguish **live** vs **durable-offline** membership, and had no freshness signal. It now overlays the **broker-authoritative membership feed** (`readMembership`/`watchMembership` — the exact source the web `/graph` uses: delivery-daemon CONNZ ∪ durable registry).

- **MeshView** reads the feed once + watches it (150ms-debounced re-read), best-effort: on an open mesh or a space with no delivery daemon the read throws and is caught → the lens stays exactly as before (`prefillDms`-style degrade; never surfaced as a connection error). Exposed as `MeshSnapshot.membership`.
- **`foldTopo`** merges it onto the traffic graph: silent subscribers become member nodes, subscriptions become resting spokes (live / durable), wide readers (`>`/`*`) get a `≫` badge (no spoke-per-hub), bounded wildcards (`team.>`) expand against the channel registry via core `subjectMatches`, and ids resolve through the roster so a member that **also** has traffic isn't double-counted. Membership links are kept **separate** from traffic edges (they have no rate).
- **Renderers:** the **map** (③) draws the full membership skeleton under traffic (live solid-faint, durable/offline dashed) + the wide badge; the **matrix** (②) shows a light `∘`/`◌` marker for member-without-traffic cells; the **sequence** (①) stays a traffic timeline. A header pill reads `membership: live / stale / traffic-only`.

No core change (the feed + methods already exist); no SPEC change.

**Stacked on #206** (`feat/console-participant`) — retarget to main as the chain merges.

## Verified
- `pnpm typecheck` clean; `smoke:view` (MeshView) still green (no regression).
- **New hermetic `smoke:topo-membership` (15 checks):** silent subscriber → member node + live link; durable-offline → durable link; `>` → wide, no spoke; `team.>` → expands to both concretes; member-with-traffic → single node (no double-count) with its edge preserved; freshness live/stale; **degrade** (no feed → empty memberships, `available:false`, node set identical to a plain fold).
- **Live end-to-end** against a sandboxed **auth mesh + delivery daemon**: an admin observer's MeshView reads the feed (with a daemon `asOf` heartbeat), a provisioned **subscribed-but-silent** agent appears in it, and `foldTopo` surfaces it as a member node with a live spoke to `#general` — matching the web graph. Plus the real TUI on the open demo mesh shows `membership: traffic-only` and degrades cleanly.

### DIFF
--- a/docs/mesh-view.md
+++ b/docs/mesh-view.md
@@ -60,6 +60,7 @@ interface MeshSnapshot {
   channels:  { channel: string; messages: number }[];
   feed:      FeedEntry[];       // classified + coalesced + windowed
   views:     ViewItem[];        // peer-published renderable views (json-render specs), newest last
+  membership?: MembershipSnapshot; // broker-authoritative channel membership (needs the delivery daemon)
   rates:     { msgsPerSec: number };
   status:    { connected: boolean; space: string; dmVisible: boolean; error?: string };
   signals:   MeshSignals;       // derived operator signals (below)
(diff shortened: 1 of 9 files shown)


davidfarah2003:

PR #209, opened 2026-07-07
feat(console): attach a live agent terminal

Lanzelot1:
## What

Closes the #1 control gap — the console couldn't open or drive a running agent's terminal (the classic lazygit/k9s "look inside the process" move). Now **`a`** on a roster agent (or **`:attach <agent>`**) suspends the console and hands the terminal to the agent's live PTY over the manager's WS attach endpoint; **`Ctrl-]`** (or `COTAL_DETACH_KEY`) returns and repaints.

- **Reuses everything that already exists** — the manager's `attach` op (returns a loopback ws after own-child/admin authz, rejects non-pty runtimes) and the CLI's `attach-client`. The console resolves the ws via a per-action `control-caller-admin` mint (the exact path kill/pause use) and streams it.
- **Suspends Ink in place, no rebuild:** on attach the App renders `null` and gates its top-level `useInput`, so Ink releases stdin (ref-counted raw mode) *before* `attachClient` grabs it; the observer endpoint + MeshView keep running in the background (App stays mounted), so the console repaints current state on return. On detach the effect re-asserts alt-screen/alt-scroll/cursor (attach-client leaves the main buffer after a full-screen child).
- `canControl`-gated, pty-runtime only (tmux/cmux are watched natively and flash their own hint), same-host only (the ws is loopback). Keyed by name, so it works for a managed agent even if it isn't a mesh roster peer (like `cotal attach --name`).

**Stacked on #208** (`feat/console-membership`) — retarget to main as the chain merges.

## Verified
- `pnpm typecheck` clean; existing `smoke:attach` (pty-vs-tmux gating) still green.
- **New `smoke:attach-control:live` (4 checks):** against a real broker + Manager + pty agent, the console's exact control-op path (`control-caller-admin` mint → `requestControl(ADMIN, {op:"attach"})`) returns a valid `ws://127.0.0.1:…/attach/<name>`, and a missing agent errors (not crashes). The attachClient data path + terminal hygiene is already covered by `attach-repaint.smoke`.
- **Live tmux end-to-end:** broker + Manager + a pty echo-agent → console `:attach echo` suspends the UI (Ink → null), the agent terminal takes over, typing `HELLO-FROM-CONSOLE` reaches the agent and echoes back (bidirectional I/O), and the console process/MeshView stay alive in the background. (Detach byte delivery through `tmux send-keys` is a harness quirk; the detach+repaint path is smoke-covered.)

## Notes
No core change — reuses  + . No SPEC change.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -133,7 +133,8 @@ and configures them.
   `ps`.
 - **Pluggable runtimes.** Spawning is abstracted behind a `Runtime` contract (like pm2 or
   docker for agent TUIs): **`pty`** ships built-in (the manager owns a pseudo-terminal;
-  watch or type via `cotal attach`); **`tmux`** and **`cmux`** are extensions that put
+  watch or type via `cotal attach`, or in-process from the console's `a` key, which reuses
+  the same attach client); **`tmux`** and **`cmux`** are extensions that put
   each teammate in its own window/tab (explicit opt-ins that throw when the extension
   isn't loaded, never a silent fallback); **byo** is the floor (a human's own terminal,
   tracked via presence); **host** (Agent SDK, true mid-turn interrupt) is the documented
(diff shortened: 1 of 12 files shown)


davidfarah2003:

PR #211, opened 2026-07-08
feat(console): channel delete, unread badges, harness tag, skills in detail

Lanzelot1:
Closes the remaining web-dashboard-parity gaps in the console TUI (all but the graph rendering, deliberately deferred), plus the harness marker.

## What

- **`:delchan <channel>`** — delete one channel's history + registry entry. Extends the manager's admin `purge` op with an optional `channel` arg (one purge grammar): a filtered `STREAM.PURGE` across all senders plus the channel-registry key delete via core `clearChannel` — the web dashboard's delete path — under a per-op `channel-purger` mint (the role holding exactly the scoped purge + channel-KV grants). Wildcards rejected. Console side: palette verb behind a type-the-channel-name confirm (same pattern as `purge`); the active tab falls back to `all` if it was the one deleted.
- **Per-channel unread badges** — inactive channel tabs show a yellow `+N` (messages since that channel was last viewed), like the web sidebar's pill. Viewer-local watermarks over the cumulative channel counts: baseline at startup (history is not unread), viewing pins, stream rewinds (purge/delchan) clamp.
- **Harness tag + `runs`/`model`/`skills` in detail** — the roster tags each agent with what it *is* (`cc` = Claude Code, `oc` = OpenCode, `hm` = Hermes, else the connector's first two letters) from the card's self-published `meta.connector`, falling back to the manager's launch record via the existing `ps` merge (widened from paused-only). Agent detail (Enter) adds `runs` (connector · runtime), `model`, and the card's `skills` list. Omitted when unknown — never faked.

## Verification

- `smoke:purge-channel:live` (new): real auth broker + Manager — narrow purge leaves other channels intact, wildcard rejected, arg-less op still space-purges (6/6).
- `smoke:console-gaps:live` (new): the real console TUI under node-pty — no fake unread at startup, badges accrue and viewing pins the watermark, harness tag renders, detail shows runs/model/skills, the full `:delchan` confirm flow, clean quit (14/14).
- `smoke:console-attach:live` (13/13) and `smoke:attach-control:live` (4/4) stay green; `pnpm typecheck` clean.

Docs: `protocol-view.md` control section + feature map updated; `web.md`'s stale "dashboard is read-only" reconciled with its existing channel-delete write path.

Stacked on #209.

### DIFF
--- a/docs/mesh-view.md
+++ b/docs/mesh-view.md
@@ -100,13 +100,13 @@ visible (god-view / open mode); a chat-only observer leaves it empty.
 |---|---|---|---|---|
 | roster (status, activity, age) | `agents` / `endpoints` | ✓ panel | ✓ presence lines | ✓ sidebar |
 | all-activity feed | `feed` | ✓ feed panel | ✓ log | ✓ Monitor view |
-| channels plus counts | `channels` | ✓ tabs (`1`–`9`) |  | ✓ sidebar + Channel view |
+| channels plus counts, unread badges | `channels` (+ client state) | ✓ tabs (`1`–`9`, `+N`) |  | ✓ sidebar + Channel view |
 | golden-signal counts | `signals.counts` | ✓ tiles strip |  | ✓ tiles |
 | needs-you / blocked | `signals.waiting` | ✓ rail (`n`) |  | ✓ NEEDS-YOU rail |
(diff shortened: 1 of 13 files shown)


davidfarah2003:

PR #212, opened 2026-07-08
docs(examples): self-improving-console — opencode reviewer, demo replay + launch assets

Lanzelot1:
Split out of #197 to keep that PR focused on the views-lens feature — this is the example/demo half, all under `examples/02-self-improving-console/` (private, never published).

Adds the OpenCode cross-vendor reviewer persona, a deterministic `demo-script.json` + `replay.ts` for reproducible demo runs, LAUNCH.md posting assets, and refreshed agent personas/GOAL for the recorded demo. No code outside the example; no dependency on the views API — this can merge independently of the console stack.

Verification: `pnpm typecheck` (the example package compiles); `replay.ts` is exercised by the demo runbook in LAUNCH.md.


davidfarah2003:

PR #278, opened 2026-07-22
fix(renewal): prove broker acceptance before reporting a daemon cred adopted (W3 3a)

davidfarah2003:
## What

`cotal doctor auth` could report a credential renewal **"adopted" that the broker never accepted** - a false green. This makes the daemon's credential-reload path *prove* broker acceptance before it adopts, and narrows every surface so it only ever claims what was proven.

This is W3 increment **3a**, shipped standalone: it's a correctness fix, deliberately not bundled with the 3b store migration so bisect and review stay clean.

## The fix

- **Broker-acceptance proof.** `reloadCreds` re-signs the candidate and proves it on a throwaway preflight connection; a refused cred yields a structured failure *before* any live-cache mutation. The reply narrows to `brokerAccepted` (the `identity`/`iat`/`exp` the broker actually accepted) plus a best-effort `residentSwap` - never an unwitnessed `"adopted"`.
- **Reply-before-swap.** The resident wire swap (`nc.reconnect()`) is armed only *after* both components' proofs settle, so a slow co-component proof can't strand the admin reply.
- **Single-flight + passive preflight.** The passive 75% renewal timer and the explicit reload share one transaction (`runCredsTxn`), and both preflight before installing a candidate - so a rejected cred sitting in the store can never strand the live connection.
- **Deadline-bounded.** The whole daemon-side transaction (store fetch + preflight) is bounded at 12s, under the manager's 15s request bound, with a late-commit fence: a hung store fails loud instead of a silent "no responder".
- **Doctor honesty.** `cotal doctor auth` now exits non-zero and says so when the last renewal was refused by the broker, instead of letting cred-file health alone read as `healthy`.
- **Fingerprint redaction.** The ephemeral generation fingerprint (SHA-256 of the re-signed JWT) that binds the expected generation is redacted at the single persistence boundary, so it never lands in `.cotal/renewal.json` or logs - covering the manager *and* `doctor --fix`.

## Regression gates (all wired into `smoke:ci`)

- `adoption-false-green` (18/18) - inverted regression gate: a refused cred yields a structured failure, no adopted window, the rail survives.
- `adoption-swap-strand` (3/3) - the old "swap armed before reply" strands at the request bound; the new "swap after reply" delivers at proof time.
- `adoption-passive-preflight` (4/4) - a rogue cred in the store is rejected by the passive timer at preflight; the daemon stays up, rail responsive.
- `renewal-record-honesty` (9/9) - no fingerprint/64-hex ever reaches disk; `doctor` exits 1 on a broker-refused renewal, 0 when broker-accepted.

Also verified locally: typecheck, `standing-renewal` (8/8), `delivery-renewal` (18/18), plus the W3-seam neighbors (`secret-store` 41/41, `delivery-store-boundary`, `delivery-teardown`, `cred-lifetime`, `auth-store`).

## Scope / follow-up

The **membership** feed's conn-B passive re-read is the one path still un-preflighted; it's deferred to **3b** (needs the 75% timer/proven-value cache the store migration builds). A conn-B strand only degrades the graph feed - Plane-3 delivery is fail-soft and unaffected.


davidfarah2003:
Round-2 fixes from a freelance cold re-read, folded in `36040524`:

- **Reload deadline is now absolute from entry.** The 12s deadline was captured when the queued transaction *started*, so a reload queued behind a hung-store 75% timer tick got a fresh 12s and could commit/swap at ~24s, past the manager's 15s bound. Now captured at `reloadCreds`/`renewCredsOnTimer` entry (queue wait included) and fenced before any source I/O. New gate: `reload-deadline-queue.smoke.ts` (6/6).
- **`doctor --fix` no longer erases a broker refusal to green.** It overwrote a broker-refused record with an adoption-absent one and could exit 0 on file health alone. It now carries a prior refusal forward as an explicit unproven state (exit 1 + 'start the manager to prove'); a non-refusal is left absent. `renewal-record-honesty` extended (14 checks).

Panel: all three lanes (engineer/security/critic) APPROVE at `c3afdaa`; the delta above is re-posted for re-confirmation. Blocker 2 is closed for the endpoint resident only; the membership conn-B passive re-read remains an explicit 3b residual (a conn-B strand degrades only the graph feed, not delivery).

PR #289, opened 2026-07-23
examples: quorum code review (resample-and-vote over v0.4 endpoints)

Lanzelot1:
## Summary

A new self-contained example (`examples/05-quorum-code-review`) that runs a resample-and-vote code
review live on the v0.4 endpoint surface: one three-persona reviewer team, run **N times in strict
isolation** on a PR, keeping only the findings that appear in **≥ k of N** runs.

Point it at a public GitHub PR:

```
(cd examples/05-quorum-code-review && pnpm run review -- https://github.com/OWNER/REPO/pull/N)
```

The transport is the endpoint surface end to end: a reviewer service (`ai.cotal.reviewer`) with N
instances each serving three persona commands, and an orchestrator (a plain core-client) that
`freezeExpectedSet` → `scatter`s each persona command to every instance → groups the schema-clean,
attributed replies by `instanceId` into runs → merges the three personas per run → clusters findings
across runs → keeps support ≥ k. The k-of-N vote is application code; Cotal has no quorum primitive.

## What it demonstrates

- **N instances ARE the N runs.** One endpoint, N instances, three persona commands scattered on the
  `all` rail. Two instances with the same persona are two independent samples.
- **Structural isolation.** `scatter` fans one request to each instance and gathers one reply each;
  replies ride the caller's own nonce-scoped rail — no shared channel, so no run can condition on
  another run's output (which is what keeps the vote valid).
- **Schema-clean replies** (ajv at the serving boundary + caller-side), so the orchestrator never
  sees a malformed finding — no JSON-repair pass on the wire.
- **Broker-derived attribution** — every reply's `(instanceId, epoch)` comes from the reply subject.
- **Deadlines / partial results for free** — `missing` / `churn` / `duplicate` from the scatter.
- A **`--mock`** mode returns canned findings, so the whole transport + vote runs with no model
  quota and no network.

## How to run / smoke evidence

```
$ (cd examples/05-quorum-code-review && pnpm run smoke)
QUORUM REVIEW SMOKE OK  (14 passed, 0 failed)
```

The smoke brings up a broker, the reviewer daemon with three instances, and the orchestrator on a
fixture PR, and asserts: each of the three persona scatters is complete over three attributed
instances (0 missing / churn / duplicate / unexpected / invalid), the runs reconstruct, the 3-of-3
vote keeps the two shared findings and votes out the per-instance noise (support 1), and a foreign
caller's reply rail receives nothing.

```
$ (cd examples/05-quorum-code-review && pnpm run review -- --mock)
scatters:
  bughunter  complete=true runs=3 missing=0 churn=0 duplicate=0
  keeper     complete=true runs=3 missing=0 churn=0 duplicate=0
  sweeper    complete=true runs=3 missing=0 churn=0 duplicate=0

survived the 3-of-3 vote: 2 finding(s)
  [HIGH] src/auth.ts:42  (support 3/3)  off-by-one on the token loop bound ...
  [MED]  src/cache.ts:17 (support 3/3)  cache entry written before the DB commit ...
```

`pnpm --filter @cotal-ai/example-05-quorum-code-review typecheck` and `build` both pass. The GitHub
fetch path is verified against a real public PR; a real model run is wired but needs `opencode` + a
configured model, so it is not exercised in CI.

## Isolation fidelity (called out honestly)

The example runs on a **single local broker** and enforces each instance's command surface **in
code** — through its authorized serve grant — exactly as `@cotal-ai/core`'s own endpoint smoke tests
do. That gives real structural isolation (no shared channel, per-caller nonce-scoped reply rails,
broker-derived attribution). The stronger wire-level default-deny — a *malicious* instance provably
cannot subscribe to another instance's rail — is a credential-scoping property minted by the auth
layer for a secured space, and is deliberately out of scope for a self-contained example. Both the
README and the smoke are explicit about which property is demonstrated.

## Scope

In: reviewer daemon (register N instances, serve the three persona commands, serialize model calls
per instance for OpenCode's SQLite lock), the two JSON-Schema-2020-12 contracts with closure-digest
pinning, the orchestrator (freeze → 3 scatters → group → merge → cluster → k-of-N vote), a `--mock`
path, the GitHub fetch, a README, and a mock smoke.

Out: the Martian benchmark harness / judge / F1 scoring, the durable 50-PR work-pool queue, and any
quorum composite in core — none of that is here. The only core touch is importing existing exports.

One deviation worth noting: the contracts use an in-memory content-addressed store (a map keyed by
`contractDigest`, as the core smoke tests do) rather than the KV-backed EPC store, because the
digest-pinning guarantee is identical and it avoids provisioning an EPC stream the runnable path does
not otherwise need.

## Real e2e evidence (2026-07-23)

`pnpm run review -- https://github.com/Cotal-AI/Cotal/pull/282` — 3 instances, k=3, model `openai/gpt-5.5`, wall clock 1m37s:

- model canary ok; all three persona scatters **complete** over 3 instances (`missing`/`churn`/`duplicate` = 0)
- 2 of 9 review calls returned unparseable model output and degraded to zero findings with a warning (the designed partial-run path); the vote still converged
- the 3-of-3 vote kept exactly **one** finding, a real defect in the reviewed PR:

> [MED] `implementations/cli/src/commands/update.ts:122` (support 3/3)
> The new array parser silently drops invalid entries as long as at least one valid semver remains. That contradicts the existing loud-failure contract for malformed npm output: `["0.13.2","not-a-version"]` now returns `0.13.2` instead of throwing.

Follow-up hardening in the second commit: fail-fast model canary before broker start, clean error for a missing `nats-server` binary, an opencode-failure guard in the reviewer path, and README documents `COTAL_REVIEW_TIMEOUT_MS`.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -16,6 +16,7 @@ README.
 | [02: Self-improving Console](../examples/02-self-improving-console/README.md) | A swarm of Claude Code agents (with an OpenCode/GPT agent reviewing their work) ships a live activity-pulse sparkline into Cotal's own console, settling the data↔UI contract peer-to-peer over the mesh. Agents improving the system that coordinates them. |
 | [03: Personas](../examples/03-personas/README.md) | Ten character personas join one space and talk in real time: the same primitives (presence, channels, DMs) as the worker examples, but the peers are personalities, not roles. Research drops and derived personas are gitignored; only the READMEs and the template are committed. |
 | [04: Frontier Faces](../examples/04-frontier-faces/README.md) | Panelist personas as animated 32×32 pixel-art OpenCode agents: each thinks, lip-syncs its streamed reply, and steers its own expression. Two front-ends onto the *same* live mesh (a browser studio and a tmux wall), both spawning real agents that coordinate as lateral peers. |
+| [05: Quorum Code Review](../examples/05-quorum-code-review/README.md) | Resample-and-vote code review over the v0.4 endpoint surface: a reviewer service with N instances, three persona commands, and an orchestrator that `scatter`s a PR to every instance, groups the schema-clean replies by instance into runs, and keeps only findings that appear in ≥ k runs. The `scatter` verb gives isolation, attribution, and partial-result handling; the k-of-N vote stays application code. |
 
 Example 02 running, a Claude Code swarm with the live console beside it:
 
--- a/examples/05-quorum-code-review/.gitignore
+++ b/examples/05-quorum-code-review/.gitignore
@@ -1,2 +1,1 @@
 dist/
-.runs/
--- a/examples/05-quorum-code-review/.gitignore
+++ b/examples/05-quorum-code-review/.gitignore
@@ -1,0 +1,2 @@
+dist/
+.runs/
--- a/examples/05-quorum-code-review/README.md
+++ b/examples/05-quorum-code-review/README.md
@@ -65,7 +65,9 @@ minted by the auth layer for a secured space, and is out of scope here.
 - Node ≥ 20, pnpm, and `nats-server` on PATH (macOS: `brew install nats-server`).
 - Build core once, from the repo root: `pnpm install && pnpm -r build`.
 - For a **real** run (not `--mock`): `opencode` on PATH with a configured model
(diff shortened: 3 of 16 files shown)


davidfarah2003:
Closing as stale: this example targets the v0.4 endpoint surface, which no longer exists (current line is 0.29.x). The resample-and-vote pattern is still a good example idea if rebuilt against the current API.

PR #294, opened 2026-07-24
fix(connector-opencode): dead agents go dark — idle clears activity, shim lifeline, honest offline rows

### LINKED ISSUE #286: Stale presence entries retain a dead agent's last-known status indefinitely, masking that it died
## Problem

A dead agent's presence/roster entry keeps showing its **last-known status** (e.g. mid-action text
like "retrying: Rate limit exceeded...") for hours after the underlying process has actually died,
instead of being marked dead or expiring in a reasonable time.

Confirmed directly: an opencode-connector worker (`worker-qwen`) crashed and was replaced (killed +
respawned under the same base name, which collided and registered as `worker-qwen-2` per the existing
name-collision behavior). Hours later, `cotal endpoints`/`cotal status` still showed the **original,
dead** `worker-qwen` entry as `○ idle` with a trailing status of `retrying: Rate limit exceeded for
api_key: ...`. I verified the underlying process was genuinely dead — the `serve.pid` file for that
specific opencode session pointed to a PID that no longer existed (`ps -p <pid>` returned nothing).
The live, working instance was registered under `worker-qwen-2` the whole time.

## Real-world impact this caused

In a mesh where several `claude`-connector peers delegate work to shared opencode workers by name
(a common pattern — see also #260), this produced genuine operational confusion:

- Multiple peers, each independently checking roster/status for `worker-qwen` before delegating, saw
  the stale "retrying rate limit" status and concluded the worker was busy/unavailable — not that it
  had died and a differently-named instance was actually the live one.
- A peer that DMed the dead `worker-qwen` name got no response and no error — consistent with (though
  not independently isolated/proven in my testing) the message being silently dropped since there was
  no live subscriber on that identity. No delivery-failure feedback reached the sender either way.
- Combined with the name-collision behavior (dead name stays "canonical", live instance gets the
  suffix), any persona/instruction that says "delegate to `<worker-name>`" using the clean base name
  can end up silently talking to nothing after a single crash+respawn cycle, with zero signal that
  anything is wrong.

## Suggested directions (not prescriptive — happy to hear what fits the existing design better)

- Health-check/heartbeat-based expiry that marks a presence entry dead (or removes it) promptly once
  the underlying connection is actually gone, rather than only expiring on a long TTL tied to last
  reported activity.
- Distinguish "last known status of a possibly-dead entry" from "confirmed live, currently doing X" in
  `cotal endpoints`/`cotal status` output, so a caller (human or agent) doesn't have to independently
  verify process liveness out-of-band the way I had to here.
- If a `dm`/`send` targets a name with no live subscriber, some signal back to the sender (even just
  "no live listener for this identity") would help agents self-correct instead of retrying/pinging a
  name that will never answer.

## Context

Same overall thread as #260 (auth), but this is a separate mechanism (mesh presence/roster, not
`~/.claude/.credentials.json`). Hit this running a small mesh (2 machines, ~30 agents/workers) with
several restarts over a working session — the effect compounds with any workflow that respawns agents
somewhat frequently.

Wally-Ahmed:
# fix(connector-opencode): dead agents go dark — idle clears activity, shim lifeline, honest offline rows

Fixes #286

## What the ghost actually is

#286 reports a dead opencode worker that sat in the roster for hours as `○ idle` with activity
`retrying: Rate limit exceeded…`, and a respawn that collided with `worker-qwen-2`. We chased it
expecting a core presence bug and found none: the presence physics on `main` are correct — the KV
bucket is created with the 6s TTL, observers materialize stale-`ts` records as offline, the sweep
flips silent peers, and `toOffline` retains `activity`/`ts` by design. The ghost is manufactured
upstream of all that, in the opencode connector, in two independent ways — plus a display gap that
makes the retained record indistinguishable from a live one:

1. **Stale activity survives into idle (plugin.ts).** `session.status: retry` publishes
   `retrying: <message>` as presence activity. Every idle transition then called
   `agent.setStatus("idle")` **bare** — and `setStatus(status, activity?)` only touches activity
   when one is passed — so the retry text stayed in the presence record forever: through idle,
   and (because offline retains activity on purpose) through death. The issue's exact screenshot.

2. **A SIGKILLed shim orphans an immortal serve (serve.ts + plugin.ts).** The serve shim spawns
   `opencode serve` — and the Cotal plugin (mesh endpoint, presence heartbeat) lives **inside**
   that serve process. The shim forwards only the signals it can catch (`SIGINT`/`SIGTERM`); a
   SIGKILL of the shim — which is exactly what the manager's own hard stop sends at the end of its
   kill ladder — reaches no handler, so the serve survives, and its plugin keeps heartbeating the
   mesh indefinitely: a dead agent that never goes offline, whose pidfile/SQLite also wedge the
   next same-name spawn ("already running — kill it first"). Killing the *serve* pid itself dies
   clean (verified against npm `opencode-ai` on darwin: the POSIX bin is a hardlinked native
   binary, single process, no wrapper to orphan) — the orphan is strictly shim-level.

3. **Display: retained activity renders as if live.** `cotal endpoints` and the `cotal status`
   roster printed the retained activity unqualified next to `⨯ offline` — a dead agent's last
   `retrying: …` looked identical to a live one's. The issue explicitly asks for dead-ness to be
   distinguishable.

## The fix

- **plugin.ts — idle clears activity.** The three idle sites (`session.idle`, `session.status:
  idle`, `session.error` recovery) now call `safeStatus("idle", "")`. The empty string rides the
  existing `setStatus → setActivity` path onto the wire, and every renderer treats falsy activity
  as none. No connector-core change needed. A fourth site is the session-replacement path:
  adopting a `/new` session publishes `("idle", "")` too — the replaced session's own events no
  longer pass the `ours()` filter, so nothing else could ever clear its leftover activity, and the
  heartbeat would republish it (fresh `ts`, immune to TTL/stale sweeps) forever.
- **serve.ts + plugin.ts — shim lifeline.** The shim plumbs its own pid into the serve child env
  (`COTAL_OPENCODE_SHIM_PID`); the plugin polls it every 2s with `process.kill(pid, 0)` (a
  platform-neutral liveness probe — signal 0 delivers nothing). When the shim is gone it takes the
  same cooperative-shutdown path the control plane uses: publish offline, stop the agent, exit the
  serve process — so the ghost dies *and* the pidfile/SQLite wedge can't strand the next spawn.
  Arming the lifeline also installs a stderr `'error'` no-op: the dead shim was the only reader of
  the serve's stderr pipe, so the shutdown log line would otherwise surface as an async `EPIPE`
  and kill the serve *before* the offline put reaches the broker (death by TTL instead of the
  clean leave). `shutdown()` itself is single-flight (the control plane and the lifeline
  can race into it) and caps its tail with an unref'd 10s hard-exit, so a wedged broker drain on a
  half-open connection can't hold the dying serve open for minutes.
- **cli — offline rows read as dead.** Offline rows now render `last seen <relative>` from the
  retained heartbeat `ts`, and qualify retained activity as `was: …`. Live rows are unchanged. New
  pure `relativeTime(ts, now?)` helper in `ui.ts` (renders `unknown` for a non-finite/non-positive
  ts rather than `NaNd ago`), and both rosters render the one shared `offlineDetail()` qualifier
  so the two surfaces can't drift; `printEndpoints` stays exported/pure, so the rendering is
  unit-asserted.

## Repro / test evidence (all three new smokes fail on the unpatched tree)

**`pnpm smoke:opencode-idle-activity`** — real plugin + real broker, raw presence-KV read (the
wire, not the observer roster), covering every idle transition plus the `/new`
session-replacement adoption. Unpatched:

```
AssertionError [ERR_ASSERTION]: session.status idle clears the stale retry activity on the wire —
{"card":{"id":"local.otto_idleact","name":"Otto","role":"generalist","kind":"agent",...},
 "status":"idle","activity":"retrying: Rate limit exceeded, retrying in 27 seconds (attempt 5)",
 "ts":1784854486353}
```

**`pnpm smoke:opencode-shim-orphan`** — the real headless shim, SIGKILLed; hermetic (the
`opencode serve` host process is a stub node HTTP server that loads the *real* plugin from
`OPENCODE_CONFIG_CONTENT`, the same real-plugin/fake-host idiom as `cooperative-stop.smoke.ts` —
CI has no opencode binary). The poll is strict: it runs until the presence record is *observed*
as offline on the raw KV **and** the serve has exited (or the deadline lapses), asserts each
separately — an either/or would also pass a crash-only death, where the record silently drops by
KV TTL without offline ever being published — and then respawns the same base name to prove the
#286 wedge heals. Unpatched — the immortal ghost, `ts` still refreshing 8–9s after the shim died,
serve alive past the whole 10s window (transcript from the pre-split check, whose failure mode is
unchanged):

```
  ✓ the shim died to SIGKILL
  ✗ FAIL: a dead shim takes the agent down within 10s (serve exits and/or presence goes offline) {
  servePid: 32680,
  ps: '32680 S    /opt/homebrew/.../bin/node serve --hostname 127.0.0.1 --port 50034',
  reads: [
    { t: 6015, status: 'idle', ts: 1784854633915, serveAlive: true },
    { t: 7017, status: 'idle', ts: 1784854633915, serveAlive: true },
    { t: 8020, status: 'idle', ts: 1784854635916, serveAlive: true },
    { t: 9024, status: 'idle', ts: 1784854635916, serveAlive: true } ] }
  ✗ FAIL: the orphaned serve process exits
```

The same scenario was also run against the **real npm `opencode-ai` binary** (darwin, outside CI)
with the built shim + built plugin bundle. Unpatched — the real serve orphaned and heartbeating
(handshake password redacted):

```
shimErr: 'opencode server listening on http://127.0.0.1:50135
[cotal-connector] opencode plugin ready — space="ocorphan" name="Otto" role="worker"
[cotal-session] ses_06e5c3271ffeeW5Bllw6h7IOz5'
  ✗ FAIL: a dead shim takes the agent down within 10s ... {
  servePid: 33662,
  ps: '33662 S    .../node_modules/.bin/opencode serve --hostname 127.0.0.1 --port 50135',
  reads: [ { t: 6016, status: 'idle', ts: 1784854865297, serveAlive: true },
           ...
           { t: 9023, status: 'idle', ts: 1784854867307, serveAlive: true } ] }
```

Patched, same real binary (fresh run of the final tree — built shim, built plugin bundle, raw KV
poll): the orphaned serve **publishes offline and exits** — both *observed*, asserted separately
under the strict poll, one lifeline interval after the kill:

```
  ✓ the dead shim's REAL serve published offline (observed on the raw presence KV)
  ✓ the REAL orphaned serve process exits
  [reads] … { "t": 1509, "status": "idle",    "ts": 1784876883814, "serveAlive": true  },
          { "t": 2012, "status": "offline", "ts": 1784876885709, "serveAlive": false }
```

The presence bucket's 6s TTL stays as a backstop — if the clean leave were ever preempted, the
crash still stops the heartbeat and the record expires — but it is exactly that, a backstop: the
smokes assert the cooperative path itself (offline observed on the wire, then the exit), never
the either/or.

**`pnpm smoke:offline-display`** — pure, broker-free: `relativeTime` buckets + `printEndpoints`
offline/live row rendering. Unpatched it fails on the missing qualification.

Full local gates on the patched tree: `pnpm typecheck` ✅, `pnpm build` ✅, `pnpm test` ✅, the
three new smokes ✅, plus the nearest neighbors `smoke:presence-scrub`, `smoke:opencode`,
`smoke:opencode-coop`, `smoke:opencode-transcript` ✅. The new smokes are wired as root `smoke:*`
scripts and appended to the `smoke:ci` chain (so the shards pick them up automatically).

Merge note: open PR #285 touches the same root `package.json` `smoke:ci` mega-line, so whichever
lands second has a trivial one-line rebase conflict there — the fix files themselves are
disjoint.

## Out of scope (follow-ups, called out on purpose)

- **Presence bucket TTL reconcile.** `@nats-io/kv`'s `Kvm.create` binds an *existing* stream as-is
  (`streams.info` hit → bind; only "stream not found" → `streams.add` with the configured
  `max_age`), so a pre-release or externally minted presence bucket lacking the 6s TTL is
  silently reused and never retrofitted. Deserves a small reconcile/doctor check; separate change.
- **DM-to-dead-name behavior** (what a peer sees when it DMs the ghost's name) is delivery-plane,
  not presence — untouched here.
- **Manager hard-stop group-kill.** The manager's kill ladder SIGKILLs the shim pid, not the
  process group; a group kill (or child-aware kill) would close the same orphan from the other
  side and would also cover non-plugin children. The lifeline in this PR heals it from inside the
  serve regardless of who killed the shim, which also covers a crashed/OOM-killed shim; the
  manager-side hardening is still worth doing separately.
- **Narrower residual surfaces of the same display class.** Declared scope here is the
  `endpoints` + `status` rosters; retained activity still renders un-qualified next to offline on
  the join console's presence-event lines (`render.ts:51`) and its event/`/who` surfaces
  (`join.ts:246`, `join.ts:318`), the web `Roster.tsx:12,32` (softened by its own `ago(ts)`
  column), and the agent-facing `cotal_roster` tool (`tool-specs.ts:224`, `offline: <activity>`).
  Same class, narrower exposure — one follow-up sweep. Related residual on the process side: the
  lifeline's error no-op shields **stderr only** (the stream this plugin writes); a host-level
  *stdout* write in the post-shim-death window would still crash the serve into the TTL backstop.
  Shielding host-wide stdout from inside a plugin is deliberately not done.
- **The same idle-leak class in the sibling connectors.**
  `connector-claude-code/src/mcp.ts:65,:100` and `connector-hermes/src/hermes-hooks.ts:34,:49`
  call `setStatus("idle")` bare after activity-carrying states (`waiting` + permission title,
  `working` + pending tool) — the same one-arg fix, kept out of this PR so it stays a reviewable
  single-connector change; a follow-up PR covers the four sites. The retry→busy transition in
  this connector (busy re-entry keeps a stale `retrying: …` while genuinely working) rides with
  that follow-up, with one caution attached: clearing it via a bare `("working", "")` could stomp
  a live tool name if the host re-emits busy mid-tool.
- Observation while wiring the chain: `smoke:presence-scrub` exists as a root script but is not in
  the `smoke:ci` chain today. Left as-is here (not this PR's call) — flagging in case it's an
  accidental omission.

Two `devDependencies` were added to `@cotal-ai/connector-opencode` (`@nats-io/kv`,
`@nats-io/transport-node`, mirroring core's versions) so the new smokes can read the raw presence
KV record wire-level, the same way `presence-offline-scrub.smoke.ts` does from core.

---

This came out of a public benchmark study of the mesh under harness failure
(https://hermes-handoff-overview.fly.dev/) — the ghost in #286 is one of the failure modes we kept
hitting while measuring recovery. An interactive deep-dive of this exact bug and fix — the
ghost's anatomy, each repair, and the test evidence — is on the site.


davidfarah2003:
Closing as stale: the wake/turn/settlement region this patches was reworked repeatedly after this branch (966903cb, 15e8314f, 4a116613), presence gained TTL-based expiry with open-mesh honesty (smoke:presence-ttl-*), and failed turns now publish as run errors (a71fbd3a). The connector this diff applies to no longer exists in this shape. If a dead agent still shows a live row on current main, please open a focused issue with the repro — that would be a real, new bug.

PR #305, opened 2026-07-26
fix(connector-claude-code): forward CLAUDE_CODE_OAUTH_TOKEN so concurrent claude agents stop racing on one login

### LINKED ISSUE #260: claude-connector: concurrent spawns share one ~/.claude/.credentials.json, race on OAuth refresh (cascading logouts)
## Problem

Every `claude`-connector agent spawned by Cotal shares the same `~/.claude/.credentials.json` on its host machine (confirmed in `connector-claude-code`: comment says "claude auths via macOS Keychain / an OAuth token... The OS allow-list (PATH/HOME/TERM/...) is the only thing inherited" — `HOME` is passed through unmodified, so every spawn reads/writes the same Keychain entry / credentials file as every other spawn on that machine, and as any manually-run `claude`).

Claude Code's OAuth refresh token is single-use/rotating. With no file locking, when N concurrent `claude` processes share one credentials file, whichever refreshes first invalidates the token for the rest — they get `invalid_grant` and drop to "Not logged in · Please run /login". This is a known upstream Claude Code bug (anthropics/claude-code#24317, #56339, #27933, #48786, #43392), reported to become reliably reproducible with as few as 3-4 concurrent sessions.

In a Cotal mesh with more than a handful of `claude`-connector agents on one machine (common once you're running several peers + workers), this becomes constant: logging into one agent to fix it cascades into knocking out the others sharing the same file, sometimes repeatedly through a session.

## Proposed fix

Give each spawned `claude`-connector agent an isolated `CLAUDE_CONFIG_DIR` (and thus its own `.credentials.json` / Keychain-scoped entry — Claude Code 2.1+ scopes Keychain services by `CLAUDE_CONFIG_DIR`), so agents stop sharing one file and stop racing.

This is exactly what `stablyai/orca` (MIT, github.com/stablyai/orca) already ships for its own multi-agent case — see `src/main/claude-accounts/`: each "account" gets its own managed auth dir via a real `claude login` captured into isolated storage, selected per session/target. Might be useful prior art for the design (isolation strategy, Keychain-scoping-by-CLAUDE_CONFIG_DIR handling, WSL edge cases).

Concretely for Cotal, this would need:
- `CLAUDE_CONFIG_DIR` added to the connector's per-agent env (today it's not in `OS_ENV_ALLOW`, and there's no per-agent override path in `buildLaunch`)
- Some way to specify/generate a distinct config dir per agent (persona frontmatter field, or auto-derived from agent name under e.g. `~/.cotal/claude-auth/<name>/`)
- A bootstrap flow to get each agent logged in once into its own dir (could piggyback on an existing valid session rather than requiring N manual browser logins, if Anthropic's OAuth allows copying a session as a one-time seed into an isolated dir without it still being "shared" going forward — worth checking)

## Context

Hit this running ~15-19 concurrent `claude`-connector agents across 3 machines (2 Macs + 1 LXC) in a single mesh — logins were expiring constantly, cascading across agents on the same box every time one was manually re-authenticated.

davidfarah2003:
Fixes #260. Supersedes the approach in draft #287.

## Problem

Every `claude`-connector spawn inherits the operator's `HOME` and sets no per-agent auth env, so all claude agents on a host share one Claude login. That login's subscription-OAuth refresh token is single-use and **rotates**: once several sessions refresh at the same time, the first wins and the rest get `invalid_grant` → "Not logged in", cascading across the box. Reliable at a handful of concurrent agents.

Reproduced: two distinct agents both resolve `CLAUDE_CONFIG_DIR` unset, the same `HOME`, and the same `~/.claude/.credentials.json`.

## Why not per-agent `CLAUDE_CONFIG_DIR` (the #287 premise)

Research against the Claude Code docs (verified on claude 2.1.218):
- On **macOS** — the reporter's platform — the OAuth token lives in a **fixed shared Keychain item** that `CLAUDE_CONFIG_DIR` does **not** scope (docs relocate `.credentials.json` only on Linux/Windows). So per-agent `CLAUDE_CONFIG_DIR` would not fix the race there, and pointing a spawn at an empty dir also triggers the first-run onboarding/trust block.
- Copying `.credentials.json` into isolated dirs doesn't help — all copies share one server-side rotating refresh-token lineage; the first refresh invalidates the rest.
- `claude setup-token` mints a ~1-year **non-rotating** token used via `CLAUDE_CODE_OAUTH_TOKEN` (a static env bearer). Sharing one across many concurrent processes causes **no** cascade. It takes precedence over `~/.claude`; `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN` override it. Cotal's own `deploy/docker/compose.yaml` already injects this token for multi-agent containers — but the connector never forwarded it to local spawns.

## Fix

Forward `CLAUDE_CODE_OAUTH_TOKEN` (by name, only when present) on the same `launchEnv` provider-key rail opencode/hermes already use for `ANTHROPIC_API_KEY`. An operator runs `claude setup-token` once, exports the token in the env that starts the mesh, and every spawn authenticates off the single static bearer. Absent, behavior is unchanged.

- `extensions/connector-claude-code/src/extension.ts`: pass `providerKeys: CLAUDE_CRED_KEYS` to `launchEnv`; correct the stale "claude forwards no provider key" comment.
- `extensions/connector-core/src/launch.ts`: update the `MODEL_PROVIDER_KEYS` doc comment.
- `docs/connect-claude.md`: new "Running many claude agents at once" section documenting the workflow, precedence, trust posture (the token is the whole-account bearer and reaches each agent's env — rendered into an owner-only launcher by the tmux/cmux runtimes, the same rail every forwarded secret takes today), and the restart-to-rotate / opt-in caveats. Docs bundle regenerated.
- `smoke:claude-auth-env` (wired into `smoke:ci`): asserts the token is forwarded when set, absent when unset, never on argv, and that an unrelated operator secret does not bleed.

## Scope / honesty

This is **opt-in** mitigation. With no token set, subscription-OAuth concurrency still cascades — there is no host-side auto-fix for that on macOS (Anthropic's shared-Keychain limitation); the `setup-token` bearer is the supported path, and this change makes it actually reach local spawns. Deliberately not included: per-agent `CLAUDE_CONFIG_DIR` (ineffective on macOS, regression risk), auto-running `setup-token` (interactive browser OAuth), forwarding `ANTHROPIC_API_KEY` (billing scope creep).

## Verification

`smoke:claude-auth-env` green; `smoke:docs` and `smoke:spawn-args` green; `@cotal-ai/connector-core` + `@cotal-ai/connector-claude-code` typecheck clean; docs bundle regenerated and idempotent.

Note: the regenerated docs bundle also repairs a pre-existing drift (an earlier `setup-internals.md` edit whose bundle regen was skipped when CI was bypassed).


davidfarah2003:
Closing as superseded: a087c2b1 "feat(connectors)!: inherit the operator's environment, reset only COTAL_*" landed a stronger version of this — connectors now inherit the full operator environment (CLAUDE_CODE_OAUTH_TOKEN included) instead of forwarding an allowlist, so concurrent agents share the login without a per-variable forward.

PR #307, opened 2026-07-27
fix(connector-claude-code): forward CLAUDE_CODE_OAUTH_TOKEN to stop concurrent-spawn logout cascades

### LINKED ISSUE #260: claude-connector: concurrent spawns share one ~/.claude/.credentials.json, race on OAuth refresh (cascading logouts)

davidfarah2003:
Fixes #260. (Supersedes the empty draft #287, whose per-`CLAUDE_CONFIG_DIR` premise the research below overturns.)

## Problem

Every `claude`-connector spawn shares the operator's one Claude credential store (macOS Keychain / `~/.claude/.credentials.json`): the connector forwards `HOME` but never a claude model credential. Claude's subscription login is a **single-use, rotating** OAuth refresh token, so several concurrent `claude` agents race on refresh — the first to rotate invalidates the token for the rest, which drop to `invalid_grant` / "Not logged in", cascading across every agent on the host.

Reproduced: two distinct agents both resolve `CLAUDE_CONFIG_DIR` unset, same `HOME`, same `~/.claude/.credentials.json`.

## Why not per-agent `CLAUDE_CONFIG_DIR` (issue #287's approach)

Grounded in the Anthropic docs + the installed `claude` 2.1.218:

- On **macOS** the OAuth token lives in a **fixed shared Keychain item** (`Claude Code-credentials`), **not** scoped by `CLAUDE_CONFIG_DIR` (docs relocate `.credentials.json` only on Linux/Windows). So per-agent config dirs would **not** fix the race on the platform the bug was reported on (2 Macs + 1 LXC).
- Copying `.credentials.json` into per-agent dirs does **not** create independent lineages — same rotating token; first refresh invalidates the rest.
- Pointing a spawn at a fresh empty dir would also trip the first-run onboarding/trust block.

## Fix

Forward `CLAUDE_CODE_OAUTH_TOKEN` (the long-lived, **non-rotating** token from `claude setup-token`) into each spawn, via the connector-core `providerKeys` rail — by name, only when the operator has it set. One static bearer shared across all agents sidesteps refresh-token rotation entirely (precedence rank 5, beats `.credentials.json` rank 6), so concurrent agents no longer log each other out. This is the same credential the containerized `deploy/` already used; it now works for local spawns too. A single-agent subscription login is unaffected (token absent → nothing changes). Same trust posture as opencode/hermes forwarding `ANTHROPIC_API_KEY` (connector-core P6).

## Changes

- `extensions/connector-claude-code/src/extension.ts`: define `CLAUDE_CRED_KEYS`, pass as `providerKeys` to `launchEnv`; fix the stale "forward NO provider key" comment.
- `extensions/connector-core/src/launch.ts`: clarify the `MODEL_PROVIDER_KEYS` comment (no code change — the `providerKeys` seam already existed).
- `smoke:claude-oauth` (new, in `smoke:ci`): token forwarded when set / absent when unset / env-only (never argv) / no unrelated-secret bleed.
- `docs/connect-claude.md`: "Running many claude agents at once" — the one-time `claude setup-token` workflow.

## Verification

- New smoke green (7/7); `pnpm typecheck` green across all packages; changeset (`@cotal-ai/connector-claude-code` patch).
- The actual OAuth refresh-token rotation is upstream Claude behavior; the Cotal-side defect (no credential forwarded → shared rotating store) is what's reproduced and fixed here.


davidfarah2003:
Closing without merging: the cascade this targets could not be reproduced. On the default macOS subscription login it only fires when several processes hit the rotating-refresh boundary at once, and forcing that would rotate live credentials. On current claude (2.1.220, past the v2.1.211 changelog fix for parallel-session logout) and in real multi-agent use it does not manifest. Per the repo rule (do not ship a fix for an unreproduced issue), discarded. Reopen if the cascade is observed on a current version.

PR #331, opened 2026-08-09
feat(cli): register, list and unregister meshes with `cotal meshes add|rm`

davidfarah2003:
## What

The mesh registry (`~/.cotal/meshes/space.<hex>.json`) was written only by `cotal up` and cleared only by `cotal down` and the liveness sweep. Two consequences:

1. **A mesh running on another machine could not be registered at all.** No `cotal up` here means `--space`, `cotal use` and a bare `cotal spawn` never see it — the only workaround was hand-writing a JSON file whose filename is a hex encoding of the space name.
2. **The sweep deleted records nothing here could write back.** `pruneStaleMeshes` drops any record whose broker fails two liveness probes, so a sleeping laptop or a VPN blip silently unregistered a healthy remote mesh, permanently.

## The commands

```
cotal meshes                                                     list (unchanged default)
cotal meshes add <space> --server <url> [--root <dir>] [--mode auth|open] [--force]
cotal meshes rm  <space> [<space> …] [--force]
```

**`add`** — `--root` is the folder whose `.cotal/auth` holds that mesh's credentials and whose `.cotal/agents` holds its personas (default: the project you run it in; required outside one). The registry stores that path, never a secret, exactly as `up` does. `--mode` is inferred from the root's on-disk account inventory and can be stated; `--mode auth` without that account fails loud, and `--mode user` is refused because a user-auth space's IdP pins are trust that only `cotal up --user-auth` establishes. The broker is **probed before anything is recorded**, so a wrong address, a broker that wants auth, rejected creds, or an expired cred are all caught at registration with a command-local sentence instead of surfacing at the first `spawn`. `--force` records unverified (and replaces an existing record). Adopts `current` only when there is no usable one, same policy as `up`.

**`rm`** — drops records; it never stops a mesh. Several names at a time, non-zero exit if any failed. A mesh **running on this machine** is refused in favour of `cotal down` (`--force` drops the record only, leaving the broker up). Releases `current` when it pointed at a removed mesh.

**`list`** — unchanged as the bare form, now a header + a `registered` / `offline` tag.

## The load-bearing half: records have an origin

`MeshEntry.origin?: "up" | "manual"` (absent = `up`, so every existing record behaves as before), and every automatic prune now goes through one helper, `pruneMesh(space): boolean`, so the rule cannot be forgotten at one of the three sites (`pruneStaleMeshes`, `preflightOrExit`'s classified prune, `targetFromEntry`'s stale-auth-root prune):

- `up` records stay auto-prunable — `cotal up` writes them straight back.
- `manual` records are never auto-pruned. A dead broker under one is a **state, not a deletion**: `pruneStaleMeshes` returns `{ pruned, offline }`, the listing shows `offline`, and the preflight sentence says the mesh is registered but not up (`cotal meshes rm` to unregister) instead of "run `cotal up`".
- The stale-auth-root copy stops claiming a removal that did not happen (`MeshTargetErrorDetails.removed`).

`MeshTarget` gains `origin`; `personaDir` is exported from workspace so registration resolves the persona root through the same function target resolution does. `cotal status` names the new origin and mentions `meshes add` when the registry is empty.

## Verification

- New `implementations/cli/smoke/meshes-registry.smoke.ts`, wired into `smoke:ci` — **42 checks** against a real broker + a sandboxed `COTAL_HOME`: verified add, failed add records nothing, `--force` unverified/replace, duplicate refusal, `--mode auth`/`--mode user`/junk-mode guards, the prune split (`up` dropped, `manual` kept and reported `offline`, still kept on a second sweep, absent-origin treated as `up`), listing tags, `rm` (multi-name, current release, unknown name, running-mesh refusal, `--force`), unknown subcommand, and completion.
- `preflight.smoke.ts` gains the workspace-layer exemption + copy assertions (72 checks).
- `flag-inventory` golden updated for the new flags/positionals.
- `pnpm typecheck` clean; `smoke:preflight`, `smoke:connect`, `smoke:spawn-from-anywhere`, `smoke:multi-space`, `smoke:clean`, `smoke:down-target`, `smoke:cli-kernel`, `smoke:flag-inventory` all green.
- Exercised through the real binary against a live `nats-server`: verified add, dead-address failure, `--force`, listing, `rm` (all paths) — and the regression itself: a command that preflights an offline registered mesh now **keeps** the record and prints `cotal meshes rm` as the remedy.

Docs updated in the same change: `docs/cli.md` (the `meshes, use, status` section) and `docs/run-a-mesh.md` (a new "Meshes you did not start here").


davidfarah2003:
## Review fold — two real registry-deletion paths closed

Cold review (grok-4.5) found the origin exemption incomplete: the liveness sweep and the classified preflight honored it, but two paths that delete by **root** or by **name** did not. Both reproduced live before any fix.

**1. `removeMeshesByRoot` — `cotal down` / `cotal clean all` destroyed a hand-registered record.** `meshes add` defaults `--root` to the project you run it in, so registering a remote mesh from inside your own project files both records under one root; tearing down the local mesh took the remote registration with it. Repro: record `local` (up) + `remote` (manual) on one root → `removeMeshesByRoot(root)` returned `['local','remote']`, registry empty. Now skips `origin === "manual"`, and `clean`'s "is this root's mesh still live" refusal asks the narrowed question through a new `localMeshesForRoot` — a reachable *remote* broker is not the operator's to stop and must not block a local wipe forever with an instruction nobody here can carry out.

**2. `claimSpace` — `cotal up --space <name>` reclaimed a registered holder by deleting its record.** Repro through the real binary: manual record for `team` at an unreachable address, `cotal up --space team` against an occupied port → record gone, `up` then failed, operator left with neither mesh and no way to rebuild the record (the reclaim runs *before* the broker starts). Now refuses for a manual holder and names `cotal meshes rm <space>`; a dead `up` holder is still reclaimed unchanged.

**3.** `registry-creds-rejected` / `registry-open-now-auth` are now origin-aware — they were still prescribing `re-run cotal up from <root>`, which for a mesh this machine only registered would start a *different*, local mesh under that name.

**4.** JSDoc and both docs pages retightened. The old claim ("only `cotal meshes rm` removes it") was false; the honest rule is that nothing which merely *infers* a record is dead touches it, and the deliberate replacements (`meshes rm`, `add --force`, a `cotal up` for that space) are named.

**5.** Both paths are now covered in `meshes-registry.smoke.ts` (50 checks), proven red before the fix and green after — with a workspace rebuild in between, since the smoke imports the built package and the first red/green attempt was silently testing stale `dist`.

Gate after the fold: `pnpm typecheck` clean; `meshes-registry` 50, `preflight` 72, `connect` 18, `clean` 58, `down-target` 11, `multi-space` 158, `spawn-from-anywhere` 34, `cli-kernel`, `flag-inventory` — all green.

davidfarah2003:
## Second review fold (gpt-5.6-sol-fast, max)

The second cold review found the verification itself dishonest in places. Everything below was reproduced live before being fixed.

**Mode was never actually proved.** A NATS broker with no auth configured accepts a credential-bearing CONNECT and ignores it — so "the creds were accepted" was not evidence of enforcement. Reproduced: `meshes add trustme --mode auth` against a wide-open broker recorded `auth` and reported success, promising JWT/ACL protection that did not exist. `add` now probes **credlessly first** — an auth broker refuses, an open one lets us in — and refuses the mismatch in both directions.

**Auth without composable trust.** An account record on disk selected mode `auth` while `loadSpaceAuth` returned undefined, so the "verification" connected credlessly and every later resolution failed `stale-auth-root`. Composed trust is now required before the probe.

**Credentials in `--server`.** Reproduced with `nats://alice:swordfish@…`: accepted, persisted to the registry file, and echoed verbatim by `add` and `meshes`. Now validated the way a manifest broker URL is (broker schemes only, no userinfo, no query/hash), and the refusal never repeats the secret.

**The outside-a-project `--root` guard did nothing.** `findCotalRoot` returns its starting directory when no `.cotal` is found up-tree, so running `add` from `/` recorded `root: "/"`. It now requires a real project directory.

**`up` downgraded provenance.** `recordOurMesh` stamped `origin: "up"` on every write — including the "a broker is already on this port" refresh, which concludes a mesh is up from reachability alone — silently converting a record only the operator can rebuild into one the next sweep may delete. A refresh now preserves the origin. The foreground exit callback likewise unrecorded on server+root alone, deleting a concurrent `meshes add --force` replacement; it now checks origin too.

**`claimSpace` ordering.** The reachability branch ran first, so a *live* registered holder was told to `cotal down` it — impossible for a mesh this machine does not run. Origin is now decided before liveness, and the smoke covers both live and dead holders instead of only dead.

**`rm` refused on reachability, not ownership.** Any broker answering on the address produced a refusal plus a `cotal down` instruction that would stop nothing. It now refuses on a live recorded pid under that root, which is what the message claims — and the smoke supplies a real process rather than asserting against the state it hand-wrote.

**Declined, with reasoning:** persisting an `unverified` flag on the record. Verification is point-in-time — a verified record loses it the moment a port is reused — so a stored flag would decay into a false claim. The reviewer's own alternative is taken instead: the line now says "recorded without verifying", describing the action rather than the record.

Smoke is at **64 checks** and now states its own limit (the live broker is open and JetStream-less, so auth admission is covered by refusals only; a positive auth registration belongs with the provisioning smokes). It also imports the CLI composition root — without it the local-process surface is empty and the ownership check silently has nothing to read.

Gate: `pnpm typecheck` clean; `meshes-registry` 64, `preflight` 72, `connect` 18, `clean` 58, `down-target` 11, `multi-space` 158, `spawn-from-anywhere` 34, `cli-kernel`, `flag-inventory` — all green.

davidfarah2003:
## Third review fold — provenance, and not trusting it

The second fold's "preserve a manual origin on any re-record" rule was too broad and reintroduced the same failure from the other side. Found by the second reviewer, verified, fixed.

**Provenance is now a call-site fact.** `claimSpace` returns early when server+root match, so a same-endpoint manual record whose broker is *down* survived into a run that then genuinely started a local broker — and the blanket rule kept it `manual`. Consequences: `meshes rm` would unregister a **running** local broker without `--force`, and `cotal down` would leave its record behind. `recordOurMesh` now takes `"started" | "refresh"`; the four paths that spawned the broker or proved a listener they own claim the record, and only the "a broker is already on this port" branch — which starts nothing — preserves.

**The safety checks no longer trust provenance at all.** `rm`'s live-process guard is no longer origin-gated: a live local broker is a fact about this machine whatever the record says about who wrote it, so a mis-stamped record cannot unregister a running mesh. The guard is also skipped entirely under `--force` — the probe itself throws on a multi-tenant or unreadable root, which was defeating the documented override — and is keyed on the entry's own space rather than one re-resolved from the root (which on a multi-tenant root could name a different tenant). `liveMeshProcess(root, space?)` takes the space explicitly.

**`--server` validation was leaking on its own error paths.** A malformed credential URL (`nats://alice:swordfish@`) took the parse branch and printed raw; the query-string refusal echoed `?token=…` verbatim. Neither repeats input now, and a hostless `nats://` is refused rather than persisted as structurally unusable.

**The project-root guard** accepted a regular *file* named `.cotal`; it now requires a directory.

**The mode-honesty smoke check was circular.** Its root held a marker file rather than trust, so `add` failed at the compose guard and never reached the mode probe — the branch could have been deleted with the check still green. It now mints real composed trust (`createSpaceAuth` + `saveSpaceAuth`), requires the specific "accepts unauthenticated connections" refusal, and asserts the trust really composes so the probe is proven reached. The compose guard gets its own case.

**Declined again, same reasoning:** persisting an `unverified` marker on the record. The wording is now consistent in all four places instead (docs ×2, changeset, success line): `--force` "records without verifying" — an action, not a durable property that would decay the moment a port is reused.

Smoke: **69 checks — 46 with these fixes reverted, 69 with them.** Typecheck clean; `preflight` 72, `connect` 18, `clean` 58, `down-target` 11, `multi-space` 158, `spawn-from-anywhere` 34, `cli-kernel`, `flag-inventory`, `backup-cli` all green.

davidfarah2003:
## Fourth fold — the origin gate goes back, for a better reason

Removing it last round overcorrected, and the reviewer caught the regression.

**Pidfiles are root-scoped, and the root is shared by design.** `add` defaults `--root` to the project you run it in, so a local mesh and a registration for a remote one routinely share one root — which makes the local mesh's live pid visible under it. `meshes rm <remote-registration>` therefore refused and claimed a mesh on another machine was running here; passing the entry's own space cannot disambiguate a root-scoped pidfile. That broke exactly the co-rooted scenario this feature exists to support.

The gate is safe now for the reason it wasn't before: **provenance is decided at the call site that starts the broker**, so anything this machine actually runs is stamped `up` and still reaches the ownership check. Only a hand-registered record skips it, and for that record a pid under the shared root is not evidence either way. The `--force` skip and the explicit-space keying stay.

**Contracts caught up with the previous fold.** The JSDoc and both docs still said a manual record changes only via `meshes rm` / `add --force`, which stopped being true when `started`-over-manual landed: a `cotal up` that actually starts the broker for that same space, server and root *becomes* that mesh, claims the record, and `cotal down` then clears it — while a refresh that merely found a broker answering starts nothing and preserves. Both cases are now written down.

Two stale "unverified" strings folded into the action wording (a failure hint and a comment); `grep` for the old wording is now empty across src, docs and the changeset.

Smoke at **71 checks** — the co-rooted case is covered from both sides: the registration drops despite the local mesh's live pid, and the local mesh sharing that root is still protected. Red before this fix, green after. Typecheck and the full affected suite green.

davidfarah2003:
## Fifth fold — first reviewer's re-verification at head

Both findings real, both fixed.

**An explicit `--root` was never checked.** Only the *inferred* path required a directory; `if (flag) return resolve(flag)` recorded whatever it was handed, so a regular file or a typo'd path became a mesh root whose `.cotal/auth` and `.cotal/agents` can never exist. It now fails loud like the inferred path. It still does not require an existing `.cotal` — naming the folder is the operator saying "here", and an auth registration is separately gated on trust that composes.

**My "outside a project" smoke case depended on the machine it ran on.** `findCotalRoot` walks *up*, and the temp dir sits under the OS temp root, which the test does not own — so a stray `.cotal` there makes the walk find a genuine project, the guard correctly accepts, and the check goes red on a dirty box while passing on a clean one. The reviewer hit exactly that with a leftover `/tmp/.cotal`. The case now **asserts its own precondition** and names the offending ancestor, so a polluted workstation gets an actionable failure rather than a misleading verdict.

Worth noting how that was verified: the reviewer's Linux reproduction (`/tmp/.cotal`) does not apply on this machine, where `tmpdir()` is under `/var/folders`. Planting `.cotal` in the *real* tmp ancestor reproduces it, and the new assertion reports that exact path.

**LOW folded:** `--server` now rejects a URL path, which the "bare broker URL" contract already claimed but enforced only for userinfo, query and fragment.

**LOW kept, with reasoning:** the credentials refusal echoes the username while redacting the password (`"alice:***@…"`). That is the existing manifest validator's wording, and the identifying half is what tells the operator which URL was rejected.

Smoke at **74 checks**; typecheck and the full affected suite green.

PR #362, opened 2026-08-12
feat(herdr): add Herdr runtime extension

Lanzelot1:
Adds `@cotal-ai/herdr`: a new extension that lets the Cotal manager run agents inside [Herdr](https://herdr.dev) panes.

## Why

Herdr is a terminal workspace manager whose background server owns the panes. Spawning Cotal agents there means they keep running when the manager's terminal goes away, and they show up in Herdr's agent sidebar.

## What's included

- `extensions/herdr` — a thin driver over the herdr CLI plus a self-registering `herdr` Runtime provider (same pattern as tmux/cmux/orca).
- Agents spawn into a dedicated Herdr session (`cotal-<space>`) with its own server and socket — never the operator's default session. Watch them with `herdr session attach cotal-<space>`; each pane carries a `cotal` metadata token.
- CLI: `herdr` is listed as an official runtime (`cotal runtimes`; `cotal ext add @cotal-ai/herdr`, which also accepts a local path pre-publish).
- Docs: AGENTS.md package row and the runtime lists in cli, architecture, glossary, manifest, and run-a-mesh.
- Changeset: lockstep minor; `@cotal-ai/herdr` joins the fixed group.

## Design notes

- Lifecycle is keyed by Herdr's stable `terminal_id`; the public `pane_id` changes when a pane moves, so it is re-resolved before every pane operation and never cached.
- Exit proofs fail closed: only a successful pane inventory missing the terminal proves exit; any failed inventory throws. `waitForExit` stays strict; the `status()` probe reads uncertainty as "running" (tmux precedent).
- Secrets never touch herdr's command line or records: connector env (creds, control token) rides an owner-only, self-deleting launcher script.
- Server provisioning (`ensureServer`) surfaces the server's own stderr on a failed start. A dead child is detected early where POSIX `ps` exists, and never assumed dead where it doesn't (Windows, minimal containers) — the bounded startup window fails loud either way.
- Fail-loud guards: nonexistent or non-directory cwd (herdr would silently substitute `$HOME`), duplicate agent names (`agent_name_taken`), and unknown error codes all throw; launcher files are cleaned up on failed starts.

## Testing

- `pnpm smoke:herdr` — 56 checks against live herdr 0.7.4 in an isolated named session: spawn/stop/interrupt, pane-move re-resolution, secret-leak canaries, error classification, provisioning failure (with and without `ps`), session isolation. Wired into `pnpm smoke:runtime`; skips cleanly when herdr is absent, so CI is unaffected.
- Full lifecycle e2e run locally: a real JWT-auth broker, a real Manager with runtime `herdr`, and a real agent process that joined mesh presence from inside a Herdr pane. Despawn tore the pane down; herdr's records contain no creds; the pane is parented by the herdr server, not the manager — which is what makes agents survive the manager going away.
- `pnpm build`, `pnpm -r typecheck`, and `pnpm changeset status` all clean.


davidfarah2003:
Reviewed this end to end against live herdr, with a panel covering the secret surface, correctness, and repo-fidelity. Summary up front: **the design is sound and largely survived adversarial review — the lifecycle model and the credential handling are both right.** The blockers are a target-version mismatch and a set of gaps where nothing in the pipeline can observe whether the code works.

Six blockers, each reproduced.

### B1 — the driver cannot spawn against current stable herdr

```
herdr --session cotal-herdr-smoke agent start probe --cwd /tmp --no-focus -- sleep 30
→ unknown option: --cwd
```

`pnpm smoke:herdr` dies at `smoke.ts:82`, exit 1. herdr 0.8.0's contract is `agent start <NAME> --kind <KIND> --pane <ID>` — it attaches a *recognized agent kind* to an **existing** pane, and cannot create a pane or run arbitrary argv. `agent send` (graceful stop's `/exit`, `driver.ts:295`) is gone too.

herdr v0.8.0 published 2026-08-03; this branch's last feature commit is 2026-08-13. A user following the README's own `brew install` gets a runtime that cannot spawn.

**The fix is bounded.** A fresh headless server has no workspace, tab, or pane, so `pane split` fails `pane_not_found`. `workspace create` is the bootstrap, and it verifies clean:

```
workspace create --cwd PATH --label <name> --no-focus
  → root_pane { pane_id: "w1:p1", terminal_id: "term_658f0af52b1f51" }, cwd honored
pane run w1:p1 "<command>"
  → process-info: argv ["sleep","45"], cwd /private/tmp, pid 92633
```

`terminal_id` still comes back at creation, so the terminal-id-vs-pane-id re-resolution design is untouched. One-tab-per-agent comes **free** — each workspace is its own tab, so `paneMoveNewTab` and its garbled-move handling become unnecessary rather than broken.

Two consequences to plan for:
- `agent list` stays empty for a pane-run command, so **agents will not appear in herdr's Agents sidebar** — that claim in the description isn't achievable on 0.8.0. It also means `agentInfo()` (`driver.ts:240`) can't resolve a pane-run pane; the lifecycle lookup has to move to `pane get`/`pane list`.
- `available()` (`driver.ts:26-33`) probes only that the binary runs `--version`, never the version itself — a fake `herdr --version` printing `herdr incompatible` still returns true. So `cotal runtimes` currently advertises herdr as ready on a version where every spawn fails. A version floor belongs with the retarget.

### B2 — the 56 checks never run in CI

`ci.yml:118` runs `pnpm smoke:runtime` (which this PR extends with `smoke:herdr`). The job installs tmux explicitly; nothing in `.github/workflows/` installs herdr, and `smoke.ts:64` exits 0 when it's absent. The step is still labelled "Runtime smoke tests (tmux)". "Skips cleanly when herdr is absent, so CI is unaffected" is accurate, and is why B1 shipped green.

### B3 — `smoke.ts` claims cleanup it doesn't perform

Header says "cleans up on pass or fail", but `cleanup()` at line 367 is a bare call with no try/finally and no exit hook, so any uncaught throw skips it. Reproduced by B1's crash: session `cotal-herdr-smoke` left `running`, server pid 24815 alive.

### B4 — `ensureServer()` leaks everything it claims when the poll loop throws

```
{"harnessStillAliveAfterMain":true,"detachedServerStillAlive":true,
 "leakedTempDirs":["cotal-herdr-srv-DL56jX"],
 "harnessOpenServerLogFds":1,"serverOpenServerLogFds":1}
```

Forced with a fake herdr whose second `status server` exits 23. One throw from `serverRunning()` — the 2s timeout, a nonzero exit, the binary vanishing — leaves an open log fd in both processes, a leaked temp dir, and a detached server child never `unref()`d or killed. `finish()` and `fail()` exist at `driver.ts:179-221`, but the poll loop has no catch/finally so a throw bypasses both; `serverRunning()` (`driver.ts:121`) wraps `execFileSync` with no try/catch, which is what makes the throw reachable. Killing the exact server pid produced `harnessExit:{code:0}`, confirming the detached child — not the unref'd timers — held the process open.

Under `--runtime herdr` this means a manager that hits a transient herdr hiccup can't exit cleanly.

### B5 — the release will fail on the new package

`npm view @cotal-ai/herdr version` → E404 (positive controls: `@cotal-ai/tmux` and `@cotal-ai/orca` both 0.16.0). `.github/workflows/changesets.yml` publishes via OIDC trusted publishing only — `id-token: write`, no `NPM_TOKEN`/`NODE_AUTH_TOKEN`/`registry-url` anywhere. Trusted publishing is configured per package on npmjs.com and requires the package to exist first; initial OIDC publish of a new package isn't supported (npm/cli#8544).

`docs/release.md:16-24` states the requirement and enumerates the packages; this PR adds `@cotal-ai/herdr` to the `fixed` group but not to that list. Since the group versions in lockstep, the next release bumps herdr and fails publishing it. Needs a one-time manual bootstrap publish plus the docs entry.

### B6 — a checked test is deterministically stale

`bin/smoke/orca-extension-live.smoke.ts:193` pins `(known: pty, orca, tmux, cmux)` with the closing paren right after `cmux`. This PR adds `herdr` at `ext-loader.ts:32`, so the CLI emits `(known: pty, orca, tmux, cmux, herdr)` and that regex can never match again. It didn't redden CI because `smoke:orca-e2e:live` appears only in `package.json:114` and in no workflow.

---

### What came back clean

Worth stating explicitly, since the panel tried hard to break these and couldn't:

- **Secret handling: approved.** `launcherSource` breakout attempts all failed — U+2028/U+2029 in env, quotes and newlines in cwd, lone surrogates, string-breakout in command/args, each built as a file and run under node. Dir `0700` and script `0600` verified. `reportMetadata` value injection is impossible (`execFileSync`, single argv element): `"val --token pwned=yes --title HIJACKED"` created no key and hijacked no title. Keeping creds off herdr's command line was the **right call and it's load-bearing** — herdr's native `--env KEY=VALUE` puts values straight into pane scrollback (verified with `pane read`). Any retarget must not "simplify" to `--env`.
- **The lifecycle model is correct.** `pane list` defaults to full-session scope (`--workspace` only narrows), so a moved pane does not read as a false exit and `terminalState`'s fail-closed proof holds. Tested with a real pane move against the shipped function.
- **`stopSession` is correct** — `session stop` takes a positional `<NAME>`; it downs only the named session.
- **`status()` returning `running` on a failed inventory is a justified preserving read**, not a no-fallbacks violation: `waitForExit()` calls `terminalState` directly and still throws, so uncertainty never suppresses a real exit proof. Matches tmux/cmux/orca.

Residuals, non-blocking: a same-uid window on the launcher file between write and self-delete (inherent to the pattern; `0600` + per-user tmp blocks other uids — unlinking right after `agentStart` succeeds would close it); `smoke.ts:355` asserts emptiness rather than session scoping; `smoke.ts:328/346` are wall-clock assertions that will flake under load; `countLauncherDirs()` globs `cotal-herdr-`, which also matches the `srv-`/`proxybin-`/`fakebin-` dirs the suite itself creates.

The smoke suite is well above average — it drives real herdr, uses proxy binaries to force partial failures, and includes leak canaries. Its problem is that CI never runs it.

Gates at `14551061`: `install` 0, `build` 0, `typecheck` 0, `changeset status` 0 (lockstep minor, 18 packages, correct), `smoke:herdr` **1**. The red Windows check is inherited, not caused — `auth: right token → handler reply returned` and `DoS: server still serves a valid frame after the flood` fail identically on `main`.

I'm working through fixes for B1–B4 and B6 now and will follow up.

PR #377, opened 2026-08-13
refactor(manager): derive the right to settle a goal from winning its claim

davidfarah2003:
Follow-up to #370, which fixed two instances of a problem without fixing its shape.

## The shape

`serveSpawnGoal` used one boolean, `terminalEntered`, to answer two different questions:

| question | kind |
|---|---|
| has this goal already been settled, or claimed by someone? | bookkeeping |
| **may this attempt settle it at all?** | **authority** |

The second is an authority question and the flag starts in the permissive state. The post-accept
fallback exists to settle a goal that nobody answered, so **every way of leaving the accept path was
opted INTO committing a terminal** unless someone remembered to claim it by hand.

That is how a duplicate-goal loser came to commit `failed` on the winner's goal, using its own abort
message as the caller-visible outcome. #370 fixed it by adding two more hand-placed claims, bringing
the total to four. A rule that has to be re-applied at every exit will eventually miss one, which is
what happened.

## The change

An attempt earns `ownsGoal` by winning the create-only `bindGoal` CAS, and the single commit path
refuses anything else:

```ts
if (!ownsGoal) return;
```

Both loser branches drop their hand-placed claims. A losing attempt now cannot commit down **any**
unwind path, including ones added later by someone who never read this code. `terminalEntered` keeps
its own separate job: stopping a second settle behind a despawn that owns the outcome.

Default-closed, and enforced in one place rather than maintained in four.

## This closes a coverage gap rather than arguing it away

#370 shipped with an honest caveat: it guarded the sibling-instance branch (a foreign manager already
recorded the goal) by symmetry, and **mutation-testing that guard killed no check**, because the
duplicate-goal test races a single manager. Two guards, one of them unproven.

There is now one enforced check covering both branches, and it is proven by an existing test.
Removing it reddens the duplicate-goal case with the same stolen terminal in the diagnostic. The
safety property no longer rests on a line nothing exercises.

A dedicated two-manager test would still be worth having, to prove that branch is *reached* the way
I think it is. It is not in this change.

## Verification

Run on darwin at `463d5976`:

```
smoke:manager-spawn-action        exit 0    35 passed, 0 failed
smoke:manager-spawn-action-auth   exit 0
smoke:goal-epoch-fence            exit 0
smoke:ep-action                   exit 0
smoke:manager-restart-live        exit 0
smoke:manager-multi-live          exit 0
smoke:manager-coexist             exit 0
smoke:static-lifecycle            exit 0
build / typecheck                 exit 0
```

Mutation proof, predicted before running: removing the single fence must redden exactly the two
duplicate-goal terminal cells and nothing else. Observed **33 passed / 2 failed**, those two cells,
with the loser's abort text as the committed outcome.

The multi-instance and restart suites are included deliberately: this touches who may write a
terminal, and those are the suites where a second incarnation settles goals.


davidfarah2003:
## The two-manager test is in, so the caveat below is withdrawn

I opened this saying a dedicated sibling-instance test "would still be worth having, to prove that
branch is *reached* the way I think it is. It is not in this change." It is now: `smoke:goal-sibling-race`.

**Deterministic by order, not by simultaneity**, which is all the branch needs. A is addressed on its
own `inst` rail and allowed to finish accepting, then the identical body (same `id`, so the same
goalId and fingerprint) is published on B's `inst` rail. B therefore *necessarily* finds the goal
index already carrying A's instance id. No queue-group coin flip, no fault injection into either
manager.

The spawned agent never joins presence, so A's goal is still **in flight with no terminal** when B
arrives. Without that the suite would assert against an already-settled goal and prove nothing.

**Four positive controls**, because every assertion here is about something *not* happening and would
pass just as well if the frame never reached B at all: the frame was captured, B answered on its own
reply rail, B served A's acceptance rather than allocating its own, and B provisioned no second agent.

Mutation proof, predicted before running: removing the ownership fence must redden the two terminal
cells and leave all four controls green. Observed **7 passed / 2 failed**, and the attribution is
unambiguous in the artifact itself:

```
state: "failed"
data.error:  goal "…" was accepted by instance "avjl95bp…"   <- A
committer.instanceId: "2cje44cc…"                            <- B committed it
```

Two different instance ids in one record. That is the sibling branch committing a terminal on
another manager's goal, which is exactly what #370 could only argue about.

Verification at `59c90b49`: `goal-sibling-race` 9/9, `manager-spawn-action` 35/35, plus
`manager-coexist`, `manager-multi-live`, `gate-inventory`, build and typecheck, all exit 0. The suite
is appended to the `smoke:ci` chain, not inserted, since shard assignment is positional.

Issue #424, opened 2026-08-14
fix(manager): agent seats export COTAL_SERVERS and COTAL_CREDS into every child process


davidfarah2003:
## Summary

A manager-hosted agent seat exports `COTAL_SERVERS` — and `COTAL_CREDS` — into every child process
it spawns. Any script, test, or smoke suite that defaults its broker to the environment therefore
connects to whatever broker the seat was launched against, **authenticated**, without ever naming
it. When the seat is running against a production broker, a local test run silently becomes a
production client.

## Measured

Read directly from `/proc/<pid>/environ` on a Linux box, walking the process chain from a child
shell up through the manager:

```
pid 1342637  (child shell spawned by the seat)
    COTAL_SERVERS=nats://<production-host>:4222     <-- INHERITED
    COTAL_CREDS=/…/.cotal/auth/creds/<name>.<uid>.creds   <-- INHERITED
    COTAL_SPACE=main
    COTAL_ID=…            COTAL_LIFECYCLE_UID=…
    COTAL_CONTROL_SOCKET=/tmp/cotal-….sock
    COTAL_CONTROL_TOKEN=…

pid 1245774  (the seat process itself)
    COTAL_NAME=<seat>     COTAL_SERVERS=nats://<production-host>:4222

pid 1236231 / 1236215 / 1236195  (the manager, node)
    COTAL_SERVERS  ABSENT
```

The injection boundary is visible: the manager's own processes do **not** carry `COTAL_SERVERS`;
the seat does; and every descendant of the seat inherits it.

## Why this is worse than a stray variable

`COTAL_SERVERS` alone would be a misrouted connection. `COTAL_CREDS` travels with it, so an
inheriting process gets a **live, authenticated** connection rather than one that fails at the auth
boundary. `COTAL_CONTROL_SOCKET` and `COTAL_CONTROL_TOKEN` are inherited too, which hands any
descendant the seat's control-plane handle.

The failure is silent and green: a suite that connects "to the default" passes, and nothing in its
output records which broker it used.

## Reproduce

From any manager-hosted seat:

```sh
tr '\0' '\n' < /proc/$$/environ | grep -E '^COTAL_(SERVERS|CREDS|CONTROL_TOKEN)='
```

## Suggested direction

Two things, and they are separable:

1. **Do not export connection material into arbitrary children.** The seat needs `COTAL_SERVERS` /
   `COTAL_CREDS` for its own client; a shell it spawns to run a build does not. Passing them
[...]

PR #426, opened 2026-08-14
test(core): measure whether live eviction can revoke a seed-minted credential

davidfarah2003:
## What this is

A self-contained probe suite measuring whether broker-side live eviction can **revoke** a
seed-minted credential, plus the observer publish fence carried onto a main-based branch so its
green is re-derivable on the tree everyone actually has.

**No product source is touched.** The whole branch is two smoke files and their `package.json`
script entries.

## The finding

**Eviction reaches a seed-minted credential but does not revoke it.** It is attributable to the
CONNZ scan and it *is* kicked — `verifiedGone: true` — and the holder reconnects immediately with
the same material under a new cid. So `verifiedGone: true` describes a **gap, not a departure**.

It is **a disconnect button, not a revocation lever**, for this credential shape.

`evictDeniedPrincipal` names a committed deny-new as its precondition
(`packages/core/src/evict.ts:226`: *"ledger revoke / cred expiry / ACL removal"*). For a still
unexpired raw `operator` mint there is **no immediate, per-credential, ledger-backed** deny-new —
`mintCreds` records no ledger row for that profile; only `endpoint-serve` enters durable
finalization (`packages/core/src/provision.ts:819`). **Expiry and loaded data-signing-key rotation
are real deny-new boundaries**, with rotation coarse and account-wide.

The contrast arm is the point: a bearer-exchanged credential *is* refused a fresh connection,
because the actor ledger's deny-new boundary reads it. **That boundary has no seed-mode arm.**

## How it is verified

Nine named cells. Both directions are controlled, so the arms can differ:

| Cell | What it establishes |
|---|---|
| `A1` | control — the seed connection **survives** an unrelated eviction (the probe can observe survival) |
| `A2` | the seed principal is attributable and kicked (`kicked >= 1`, guarding against an empty scan) |
| `A3` | it is live again under a **new cid** — the kick landed and the client redialled |
| `B1` | control — the bearer connection **dies** through the same path (the probe can observe death) |
| `B2` | after revoke, the bearer is refused a fresh connection |
| `B3` | control — a granted, **never-revoked** bearer through the same path **does** reconnect |

`B3` exists so `B2` is falsifiable: without it, "the revoked bearer is refused" is equally
consistent with *no* bearer being able to reconnect at that point, which would make the contrast an
artefact.

**A registered prediction was falsified and is kept falsified, not rewritten.** `A3` was predicted
as "drops and stays dropped". It dropped and was already back. The cell now asserts the measured
behaviour and is **green while the defect exists** — it reddens if a seed-mode deny-new boundary
ever lands, which is intended and stated in the file.

### Mutation proof

Predictions were registered in a commit **before** mutating, naming cells rather than a count.

- **M1 — valid and killed.** Neutering the CONNZ target filter reddened exactly `A2`, `A3`, `B1`
  and left `A0`, `B0`, `A4`, `B2`, `A1` green, as predicted. Non-equivalent by observation:
  `kicked` goes `>=1 → 0` and the connections stay live.
- **`A1` staying green is the result worth keeping.** A scan matching *nothing* returns exactly
  `A1`'s shape, so the control proving the probe can observe survival is also the control a **blind**
  probe passes. `A2`'s `kicked >= 1` and `A3`'s cid comparison are the clauses carrying the weight.
- **M2 — invalid / not performed, and recorded as such rather than as "survived".** It never
  executed, and even loaded it could not have produced its registered outcome, because a second
  fresh ACL resolver read independently refuses the deleted row. An arm that can only ever report
  one outcome is not a control.

M2's deadness was caught by a **positive control on the mutation itself**: an unconditional `throw`
at the top of the mutated function still left the suite green. In that worktree the package resolved
outside the branch, so the edit was never loaded. **This is not a property of worktrees in general**
— measured on the same machine at the same time, a second worktree resolved the same package
locally. The portable lesson is the remedy: **resolve the module path and plant a positive control
before grading any cross-package mutation**, or rebuild and prove the changed artifact.

## Review

Independently reviewed by a separate agent on its own worktree and its own ephemeral broker, tasked
to **refute** rather than confirm. It re-derived the suite green independently, attacked `A2`, `A3`,
`A4` and the mint path and landed none of them, and **corrected two claims that were wrong**: the
header's description of what the mint records, and an over-strong statement that the credential was
not revocable *at all*.

Its narrowing of `B3` is carried in its own words, written into the cell rather than only into the
review, so the claim cannot drift upward later:

> B3 validates the actor-ledger **system boundary** — the refusal is revocation-caused rather than a
> constant bearer-reconnect failure — but it is **not** mutation proof of `ledgerAuthorizeConnect`
> specifically, and it does not identify **which** boundary refuses.

**Stated scope limit:** the suite does **not** replay the production bearer issuance chain. It wires
the file-row gate and resolver directly, while production also reads the credential ledger.

## Safety

Ephemeral broker on a free port, from a scratch directory. The suite **asserts the URL it actually
dials is not the live host as its first action**, and deletes inherited `COTAL_*` connection
variables so no code path can pick a live default up. It kills only the `nats-server` it started, by
exact pid, and awaits its exit before removing the scratch directory.

The inherited-environment hazard this guards against is filed separately as #424.

## Not in this PR

- **No send surface.** None is implemented, on any design. That remains ruled out.
- **No replacement revocation lever.** Named as a follow-up and owned elsewhere — see below.
- No fix for the request-handler seed reachability; that is scoped to its own change.

> **Follow-up owned outside this PR.** Once the account seed is properly unreachable from the
> request handler, the only on-demand coarse deny-new reachable from that path goes with it, leaving
> a still-unexpired `operator` credential with nothing but its expiry ceiling. That loss is real and
> must not be rediscovered by someone assuming rotation is still available.


davidfarah2003:
Closing: this is a measurement, not a fix, and it has no owner. If the eviction-vs-seed-minted-credential question comes back, re-cut it from this branch on current main.

PR #437, opened 2026-08-15
fix(core)!: no part renders as nothing — name the kind instead of stringifying it away

davidfarah2003:
## What was broken

A message part whose kind the renderer did not know rendered as nothing at all. The text pipeline stringified it away, so a reader saw an empty message rather than a message it could not display — the failure mode where the reader cannot tell "there was nothing" from "there was something I could not show you".

## The change

An unrecognised part now renders as a named placeholder carrying its kind, so the reader learns that content exists and what sort it is. Marked breaking because it changes what existing consumers see for parts they previously received as empty.

## Scope

- `packages/core/src/parts.ts` — the rendering path
- `packages/core/smoke/artifact-contract.smoke.ts` — contract coverage for the new behaviour
- a changeset (0.x, so a breaking change is a minor)

The branch's own docs commits record two corrections to its scoping: the affected instrument class is 23 sites across 20 files rather than four, and the shared renderer closes three of seven rather than five.

## Review note

This branch was authored on the box and has not been reviewed by me beyond reading its diff and history. Opening it so it is reviewable rather than sitting unmerged on a disk; it should get a normal review before landing.


davidfarah2003:
Closing as superseded: main's packages/core/src/parts.ts landed this consolidation via e08fa84e — one shared renderer for the three surfaces, `[unrenderable part kind …]` and `[empty data part]` markers instead of silent empty strings, plus the generic part-renderer extension seam (which this branch predates). The defect this PR names is fixed on main in a more evolved form.

PR #438, opened 2026-08-15
feat(core): artifact attachment index and the confirmAttach control rail

davidfarah2003:
## Scope

The artifact plane's attachment path: a possession index for attachments, `confirmAttach` served on the delivery control rail, and the boundary work that keeps the index single-writer.

32 files, ~3.9k lines added.

## The shape of it

- **The index is bound inside the attach module, not at the rail.** Binding at the rail would let any rail caller reach the mutators.
- **The attachment index mutators are not public API** (breaking, hence a minor on 0.x). They were reachable from outside the module that owns them, which is what makes a single-writer claim unenforceable.
- **`fitChunk` refuses a sub-one ceiling** rather than returning a zero, which livelocked its caller — a zero-size chunk means the loop never advances.

## Test work worth noting

The branch's own history records the single-writer sweep failing twice to see what it claimed to guard: once because it could not see the raw mutators, once because it did not know about the delete path. Both are recorded rather than quietly fixed, and the guard was written against the possession index before a write grant for it exists.

## Review note

Authored on the box; I have read its diff and history but not reviewed it line by line, and I have not run the suite on it. Opening it so it becomes reviewable instead of sitting unmerged on a disk. It needs a normal review — particularly the single-writer argument — before landing.


davidfarah2003:

PR #439, opened 2026-08-15
feat(core,cli): a health surface the ready card asks instead of infers

davidfarah2003:
## Scope

A health module in core plus the CLI ready card that consumes it: the card now **asks the manager** for its row rather than inferring the manager's state from local evidence.

33 files changed, 4820 insertions, 27 deletions, measured against the merge base.

## Why asking beats inferring

An inferred row is right until the thing it infers from stops tracking the thing it describes, and then it is confidently wrong with no signal. The manager is the authority on its own liveness, so the card asks it. `packages/core/src/health.ts` is the shared surface.

## Supporting fixes

- **The pid-record read is injected**, so both read orders can be constructed in a test — a race you cannot construct is a race you cannot prove you fixed.
- **The build guard could not see `.tsx` sources**, so part of the tree it claimed to cover was never scanned.

## Test work worth noting

The branch records a mutation prediction before applying the mutant, and separately records a review pre-probe mutant that **stayed green** — an instrument that failed to notice, written down rather than dropped. It also states what the decision table does NOT discriminate, which is the half most tables leave out.

## Review note

Authored on the box; read but not line-by-line reviewed by me, and I have not run the suite on it. Opening it for review rather than leaving it unmerged. Needs a normal review before landing.

## Scope correction

This branch previously carried 197 per-lane working files — predictions, results, captured suite
output, traces and shell instruments — under `.lane/` and `.observations/`. They are working notes,
not product: they bury the change a reviewer is meant to read, and they put absolute host paths and
machine-specific traces on a public remote. They are now untracked and ignored, so a later commit
cannot put them back, and they remain on disk for whoever works in the tree.

**This changes what the PR adds, not what the repository has already seen.** Those files were pushed
to this branch earlier and that history is unchanged; nothing here rewrites or unpublishes it.
Whether the earlier commits should be dealt with separately is a repository decision and is not
attempted in this PR.

Twelve citations in nine tracked files pointed into those directories, four of them in product
source under `implementations/cli/src`. Removing the files without fixing the citations would have
left shipped source referring to evidence that is not in the repository, so each now states its
claim without the path. No tracked file reads such a path at runtime — that was checked first,
because a working note a test depends on is not a working note.


davidfarah2003:

PR #440, opened 2026-08-15
feat(connector-core,connector-claude-code): the AG-UI event plane

davidfarah2003:
## Scope

The AG-UI event plane across the shared connector runtime and the Claude Code adapter, plus the frame contract that makes it checkable.

67 files, ~11k lines.

## The shape of it

- **A consumer-side frame validator ships in `connector-core`**, because the contract is one types cannot enforce: the producer is a separate process and the frames arrive as data. A type says what the code believes; the validator says what actually came over the wire.
- **A run opens on an SDK-submitted prompt**, and a session that produced zero runs now says *why* rather than reporting an empty result indistinguishable from "nothing happened".
- **The absent-origin case enters the §3.1 system class** instead of throwing, so a frame with no origin is classified rather than fatal.
- **`examples/02` stopped losing non-text parts** from its transcript.

## Test work worth noting

The branch grades the absent-origin table and **states the denominator every count rests on**, which is the part that makes a coverage number mean anything. It records M20 — a row that was graded against a corpus that no longer exists — rather than deleting the embarrassment. And it declares `smoke:agui-map` **ungated, with the reason it cannot be gated**, which is the honest form: a suite that runs below the CI door should say so out loud instead of being counted as coverage.

## Review note

Authored on the box; read but not line-by-line reviewed by me, and I have not run the suite. Opening it for review rather than leaving 90 commits unmerged. The ungated-suite declaration in particular deserves a reviewer's eye before landing.


davidfarah2003:

PR #441, opened 2026-08-15
feat(core,connector-core): agent connection control and the peer-reach fences

davidfarah2003:
## Scope

Agent connection control in the core endpoint, with the fences that bound peer reach and the probes that prove they hold.

33 files, ~7.6k lines, most of it in `packages/core/src/endpoint.ts`.

## Test work worth noting

This branch's evidence discipline is the reason to read it rather than skim it.

- **The DM and anycast reach are DRIVEN, not read.** An assertion that reads a value proves the value; sending the message proves the path.
- **A fence assertion now names WHICH fence refused**, after the loose form survived a mutation. A test that passes when any refusal occurs cannot tell you the right one did.
- **The outage probe was written unrun, with its refutation criterion first**, and all four outcomes pre-declared before it was executed — so the result could not be read to fit whatever came back.
- **A term list that made CLEAN unreachable is refused**, which is the class of defect where a scan reports success because it can never fail.
- Two fixture bugs are recorded in place and one refuted claim is struck rather than quietly dropped.

## Review note

Authored on the box; read but not line-by-line reviewed by me, and I have not run the suite. Opening it so 93 commits become reviewable instead of sitting on a disk. Needs a normal review before landing.


davidfarah2003:

PR #446, opened 2026-08-15
fix(web): route on the channel the broker policed, and stop rendering parts as nothing

davidfarah2003:
The browser dashboard decided which channel a message belonged to by reading `msg.channel` off the
payload. That field is written by the publisher. The broker polices **subjects**, not payload
fields, so a sender could put any channel name in a message body and have the dashboard file it
into that channel's transcript — including a channel the sender could not publish to.

The verified channel was already available and was being discarded: the observer parses the subject
to get the authenticated sender, then dropped the rest of it. Routing now uses the channel derived
from the subject the broker actually enforced, and where no authoritative channel exists — direct
messages and anycast, which carry none — the publisher's claim is cleared rather than trusted, so a
forged value cannot reach a transcript, a channel list, or an unread badge.

### Also in this branch

- **Undrawable message parts no longer render as nothing.** A part kind the surface has no renderer
  for produced an empty body, so a message with content appeared as a blank line. It now renders a
  marker naming the kind. A part carrying data keeps that data instead of having it replaced by the
  marker; a surface that prints a marker while dropping the content looks like successful rendering,
  which is the failure this is meant to prevent.
- **A shared parts renderer** (`parts.js`) so the two dashboard surfaces cannot drift apart on what
  a part looks like, which is how the original defect reached both of them.
- **Four fixes to the `setup-pure:live` smoke.** Its project directory is now anchored explicitly
  rather than resolved by walking upward, so it cannot silently bind to a shared root; its absence
  cells drive the real resolver instead of a stand-in; its control detects wrong-root drift instead
  of passing through it; and its comments name the statements they point at rather than line numbers
  that move, and claim only what the suite actually proves. That smoke also builds the workspace
  package before running, because it was reading a stale build.

### Tests

Two new suites, both wired into `smoke:ci`:

- `smoke:web-channel-authority` — extracts the shipped handler and backfill functions from the
  served JavaScript and **executes** them, rather than matching their text, then asserts message
  content and destination for chat, direct, and anycast messages on both surfaces. Hostile payloads
  carry a forged channel, and each run gets its own payload because the code under test mutates what
  it is handed.
- `smoke:web-parts` — asserts each surface's own renderer call is driven, not a substring of it.

The channel-authority suite locates its targets through the TypeScript parser rather than by regex
over text, so a commented-out copy is invisible by construction, and it executes the whole
containing function rather than an extracted fragment — identifying a statement is not identifying a
statement that runs, and a fragment lifted out of its control flow will execute happily whether or
not the real page would ever reach it.

### What this does not cover

- No protocol or spec surface changes; `docs/` and `SPEC.md` are untouched.
- The suites drive the served JavaScript directly. **No cell opens a browser or asserts rendered
  HTML**, so this proves the routing and the renderer's return value, not that either survives to
  the pixels.
- Rendering of external observer/UI event frames, and the event filter that selects them, are
  separate work and are deliberately untouched here.
- The dashboard's HTTP surface is unauthenticated on loopback. That predates this branch and is not
  addressed by it; it is named here because the routing fix should not be read as making that
  surface safe.
- A failed membership read still renders as a successful empty result, so a viewer cannot
  distinguish "nobody is subscribed" from "the read failed". Also predates this branch, also not
  fixed here.

### Release

Changeset included. `@cotal-ai/web` and `cotal-ai` carry the first-party changes; the `fixed` group
versions every package in lockstep, so the plan bumps all of them at patch.


davidfarah2003:
Rebased onto `main` (`183f9582`). The branch is mergeable again; no change to the fix itself.

### What conflicted, and what was chosen

One file, one key, twice — `package.json` → `scripts.smoke:ci`, once in each of the two commits that add a suite to it. Both sides were **pure additions with no removals**, so the resolution is the **union**, not a side:

| # | commit | on `main` | in the replayed commit | resolution |
|---|---|---|---|---|
| 1 | `stop the dashboard rendering an undrawable message part as nothing` | +`smoke:mesh-root-identity`, +`smoke:canonical-root-missing`, +`smoke:mesh-target-rebind` | +`smoke:web-parts` | 218 + 1 = **219**, `web-parts` inserted after `web-harness` |
| 2 | `route on the channel the broker policed…` | (same three) | +`smoke:web-channel-authority` | 219 + 1 = **220**, inserted after `web-parts` |

Neither `--ours` nor `--theirs` is correct here: each drops the other side's entries with nothing in the diff to show it. A one-line conflict is exactly the shape where that goes unnoticed.

### Checks on the resolution

```
smoke:ci   base 215 · this branch before 217 (+2) · main 218 (+3) · after merge 220
           215 + 2 + 3 = 220                     duplicates 0
           entries on main missing from the result   0
           entries on this branch missing            0
           entries in neither                        0
           relative order of both sides preserved as subsequences: yes
```

`package.json` re-parses as JSON.

### Checks on the rebase itself

`git range-diff` over the old and new series: **28 entries — 25 identical, 2 changed, 1 added.** The 2 changed are exactly the two commits above, and the only file appearing in either interdiff is `package.json`. The added one advances the internal notes pointer, which was a pending change on this branch before the rebase and is now committed as part of it.

Scope against `main` is unchanged at 12 files. `git rebase` throughout; nothing was reset or cleaned.

PR #450, opened 2026-08-15
fix(web): tell the browser a membership read failed instead of serving it as empty

davidfarah2003:
The dashboard's `/api/membership` route answered a failed read with `{asOf: undefined, members: []}`
and a `200`. `JSON.stringify` drops a key whose value is `undefined`, so those bytes are
`{"members":[]}`, byte-identical to a successful read of a space where nobody is subscribed. The one
field that might have separated the two never reached the wire.

Downstream, the graph's freshness pill rendered that as `membership: traffic-only`, which is a claim
about the **mesh** (no membership feed is being published) standing in for a fact about the **viewer**
(it could not find out). A UI reported traffic-only against a mesh that had a feed, and nothing on the
page could have revealed the difference.

### What changed

Four silences fed that one display, and each is now a named condition:

- `/api/membership` answers a failed read with a `503` and a body naming the condition
  (`membership-read-failed`), instead of a `200` and an empty snapshot.
- The two server-sent-event paths (the debounced push and the per-client seed) emit a named event
  instead of swallowing the rejection with `.catch(() => {})`. A swallowed rejection on the push path
  left the graph showing its last good snapshot indefinitely: not merely empty, but stale and
  confident.
- The page stops manufacturing an empty snapshot, both from a non-200 (whose body parses fine as
  JSON, so the status is checked before the body is trusted) and from a fetch that never answers.
- The pill gains an `unreadable` state, tested **before** `traffic-only` so a refusal cannot borrow
  that phrase. `live`, `stale` and the honest use of `traffic-only` are unchanged and each keep a cell.

### Verification

`smoke:web-membership-refusal`: **36 cells, `rc=0`**. The shipped route block, the `json` helper, the
event encoder and broadcaster, the boot fetch and dispatch, and the pill are lifted out of the files
that ship them and executed. Status and response bytes are read off a recording `ServerResponse`;
event frames are compared **whole**, so a malformed frame is a red rather than a substring hit.

Every mutation below was predicted to a **named** cell before it ran, and confirmed present in the
file by a grep of the mutated line before the suite was run. A mutation whose application is not
shown is not evidence, and a "survival" from a file that was never mutated is the most convincing
false green available (it happened once here, caught and re-run).

| # | mutation | mutant confirmed | predicted first red cell | result |
| 1 | restore the silent-empty catch | grep 1 | `a FAILED read is not served 200` | KILLED |
| 2 | keep the 503, drop the condition name | grep 1 | `a FAILED read NAMES its condition…` | KILLED |
| 3 | pill tests `available` before `unreadable` | grep 1 | `an unreadable feed says so` | KILLED |
| 4 | drop the status gate on the boot fetch | grep 1 | `M4: a 503 refusal … renders 'unreadable'` | KILLED |
| 5 | restore `.catch(() => {})` on the event push | grep 1 | `M5: the SSE push writes EXACTLY the refusal frame` | KILLED |
| 6 | rename the client's listener literal | grep 1 | `the listener's event name IS the server's exported constant` | KILLED |
| 7 | restore `.catch(() => ({members: []}))` at boot | grep 1 | `M7: a REJECTED boot fetch renders 'unreadable'` | KILLED |
| 8 | encoder returns early for the refusal event | grep 1 | `M5: the SSE push writes EXACTLY the refusal frame` | KILLED |
| 9 | prefix refusal frames with one character | grep 1 | `M5: the SSE push writes EXACTLY the refusal frame` | KILLED |
| 10 | drop the frame terminator | grep 1, original 0 | `M5: the SSE push writes EXACTLY the refusal frame` | KILLED |

Mutation 2 is the non-equivalence proof for 1: under it both status cells still pass, so the naming
cell is not a restatement of them. Every mutated file was restored from a copy and the restore
verified **against `HEAD`** (not against the copy, which would prove only that the backup was intact),
and the suite is green again at the restored bytes.

Non-vacuity is asserted before anything executes, and both lifters carry positive controls (an anchor
that is not in the file must come back `null`), so their successes are matches rather than a lifter
that cannot miss. The event cells carry a success-frame control, so the refusal cells are not passing
because the harness writes nothing at all.

### Review

Five rounds of adversarial review, all five findings confirmed by execution here before any fix was
written. Mutations 4, 5, 7, 8 and 9 each survived a suite that was green at the time: 18, 18, 31, 32
and 36 cells respectively. The recurring shape, recorded in the suite's own header:

> the fixture, or the assertion, reached one step less far than the sentence describing it

first the boot path was undriven while the header claimed coverage; then the event paths; then a stub
that could only resolve, leaving the no-answer arm unexecuted; then a stub whose *argument* was
asserted rather than what it *did*; then a substring match where a whole frame was meant.

### Registration and scope

Appended to `smoke:ci` rather than inserted: suites are assigned to runners by list position, and the
measurement is **0 of 221** suites changing runner label and **0 of 221** changing their on-runner
predecessor. `smoke:gate-inventory` re-run as a postcondition, `rc=0`. `tsc --noEmit` passes on the
web project and on the suite; note that project excludes `src/web`, so the browser files are covered
only by the executable cells.

**Where the pinned surface stops**, stated rather than implied: native `EventSource` parsing and
dispatch over a real HTTP stream, and a live broker actually causing the rejection. Both need
infrastructure this change does not stand up, and no cell here implies either. The cells stand on the
narrower fact that when the read rejects, the refusal is distinguishable from data at every surface
that carries it. Full gate not run.


davidfarah2003:
### Review verdict attached

Reviewed by an independent reviewer over five rounds, on its own detached worktree, reading only —
it could not run the suite (that tree has no `node_modules`, and it declined to install rather than
disturb a shared checkout). Every one of its findings was a static reading, and every one was
confirmed here by execution before a fix was written.

Its verdict at `d8a436e4` was **REQUEST-CHANGES** on one finding: the wire assertions used
`.includes(...)`, which is satisfied by `xevent: membership-read-failed` — one character that turns a
named event into the default `message` event, which the page's `onMessage` drops for having no `msg`.
The refusal would have gone silent again with every cell green. Confirmed: that mutation survived all
36 cells. Fixed in `d58b07cd`, which compares each frame whole; the mutation now reddens
`M5: the SSE push writes EXACTLY the refusal frame`, as does a second mutation added off the back of
it (a dropped frame terminator).

**The first unreachable boundary, in the reviewer's own scoping:** native `EventSource` parsing and
dispatch over a real HTTP stream, and then a live broker actually causing the rejection. Both need
infrastructure not stood up here, and no cell in this PR implies either.

PR #451, opened 2026-08-15
feat(core)!: refuse a class-queue split before the command runs, and say what a failure proved

davidfarah2003:
An endpoint call resolves one incarnation and then invokes through a class-anycast queue that may
pick another. Until now the caller was the only party that noticed: the mismatch was detected on
the reply, by which time the responder had already handled the request. That is why the error had
to say it proved nothing about whether the command ran. A check that runs after the effect is a
report, not a guard. In a multi-manager space it is how one spawn becomes several: the effect lands
on A, the caller bound to B is told the call failed, and the retry duplicates it.

A request now carries the incarnation the caller resolved against (`bind`, a new optional
`EndpointRequest` field), and a responder that is not that incarnation refuses at the pre-effect
seam: before args validation, before target resolution, and before the governed gate that can
consume a one-use payment proof. The refusal carries `ai.cotal.ep.bind-refused` and states that the
command did not run, so re-resolving and re-issuing is safe.

## What earns the `!`

**Not the new field.** `bind` is optional, absent on older callers, ignored by older responders
(§5), refused where it has no reading (`describe`, the scatter rail), and it confers nothing. It
can only make a responder the subject already reached refuse, so it narrows and never widens
(§13.3). Attribution still comes from the reply subject. No existing peer breaks on it.

**The breaking change is `fix(core)!: key retry safety by endpoint, and drop models from the safe
set`**, which predates the fence in this branch: `invokeService`'s automatic retry is now keyed by
`(endpoint, command)` rather than by command name, an unclassified command on an unclassified
endpoint fails closed, and `models` left the repeat-safe set because its effect depends on an
argument the classifier cannot see.

## The pair this does not fix, which is now tested rather than argued about

A responder that predates the fence ignores `bind` and **executes**. Its reply carries no refusal
marker, so it proves nothing about whether the command ran, and a caller that re-issued on it would
duplicate the effect. **That risk is unchanged by this PR**, and "unchanged" must not be read as
"covered", because until now nothing drove the case at all: every responder in every fixture was
fenced, so the allowlist protecting that path was graded on a path no test touched.

`smoke:unfenced-responder` drives it, with a hand-rolled responder answering the class rail without
a fence. `serveEndpoint` cannot produce the case, because its fence refuses a mismatched bind before the
handler, so a request it executes is one whose bind matched, and a matching bind is not a split.
The re-issue is withheld from every reply that does not carry the refusal, and the refusal itself is
checked rather than believed: the bind it was computed against must be the one this request carried,
and the incarnation it claims to be must be the one the reply subject attributes it to. Both halves
are derivable by the caller; a refusal failing either is `internal`, not a licence to try again.

## The instrument changed, and it had to

The suites now count executions **at the responder** rather than publishes at the caller. "One
publish" meant "one execution" only while a split was caught after the responder had handled the
request. Under the fence the second publish carries the *first* execution, so the old instrument
reports a correctly repaired call and a duplicated one identically.

Three defects were found by that change and by nothing else, all fixed here:

1. A refusal was honored on the marker alone, though a caller acts on it by re-issuing a command it
   would otherwise never repeat.
2. A re-issue that could not be resolved discarded the refusal and surfaced the resolve's own
   timeout: a retired endpoint answered a stale handle with a describe deadline ten seconds later
   instead of the immediate account saying the command had not run.
3. **That repair then fell into the unmarked-`failed-precondition` re-resolve below it and made a
   third attempt.** The handler ran twice more. No publish count could have caught it; the execution
   count did, immediately.

Where a cell's claim narrowed it was re-cut with the narrowing named, not replaced, including
three that had *silently changed which mechanism they graded*: a read that used to heal through the
allowlist now heals through the refusal, same green, different code defended. Those now assert which
path healed them. Two cells are inverted rather than narrowed and say so: "exactly one publish" was
the signature of a guard holding and is now the signature of a client stranded on a stale bind.

"Exactly once" is asserted as a biconditional rather than literally, since a reply carrying the refusal
means the command never ran, any other reply means it ran once. A re-issue goes back through the
same class queue and can be refused a second time, so a correct run may legitimately end having run
zero times; `=== 1` would be a flaky assertion of a true property.

## Recovery is counted

Handling a split makes it invisible, and the routing event is the only evidence the split exists at
all. `CotalEndpoint.splitRecoveryCount` is always on and never behind a flag, and a `split-recovered`
event carries the same fact for anyone listening; the event can be missed, the count cannot. On a
live two-manager mesh, 5 of 6 unpinned class-anycast reads split, so this is not a rare-event
counter.

## A re-issue that ran must not be reported as one that did not

The fence authorizes exactly one re-issue, and that re-issue can fail on its own terms. The
recovery wrapped both the re-resolve and the re-issued invoke in one `try`, so a second hop that
published, ran, and then threw on the caller's own post-reply currency check was re-wrapped as the
first hop's refusal. The caller was handed `WAS NOT RUN` plus `ai.cotal.ep.bind-refused` for a
command that had executed, and would make its next attempt believing it was the first. Measured on
a two-hop fixture: attempts 2, executions 1, wrapped as not-run.

Only the resolve is wrapped now. A re-issue that fails on its own terms propagates its own error
unchanged, so `respondedButUnbound` reaches the caller as itself: it is the marker that says a
responder answered and the effect may have landed, which is the opposite of what the fence's marker
asserts.

The suite that claims to grade this pair never composed the two populations. Every arm held one
responder at one mode for a whole call, so none of them could produce the case where the re-issue
is the thing that goes wrong. `smoke:unfenced-responder` now composes a fenced first hop with an
unfenced second, one responder with its mode switched between them. Every pre-existing cell stayed
green under the pre-fix code while three assertions in the new cell went red, which is the measured
statement that this was a gap rather than decoration.

## The reply now says whether the effect occurred

§13.3 defines `error.outcome` (`executed`, `not-executed`, `unknown`) and requires a responder
refusing before dispatching to the handler to carry `not-executed`. Both fence arms now do. An
omitted outcome must be read as `unknown`, so absence is never evidence of non-execution.

The parse side mattered more and was nearly missed. `pickError` rebuilds the error rather than
passing it through, so a field it does not name is discarded, and discarding this one silently
downgrades a conformant peer's `not-executed` to an omitted outcome. That turns a proof of
non-execution into an absence of evidence, at the parser, for a caller that did everything right.
It is now parsed and validated; an unrecognised value is refused rather than coerced, and an
omitted one stays omitted, because manufacturing `unknown` would make "the peer said unknown" and
"the peer said nothing" the same bytes.

`ai.cotal.ep.bind-refused` is this implementation's vocabulary and `outcome` is the spec's. A
conformant peer that has never heard of the detail kind reads only the second, so emitting one
without the other is what makes two implementations disagree about whether a command ran.

Both directions are graded, and separately. `smoke:ep-envelope` builds replies by hand, which
isolates the parser and `toEpError` but grades no wire; `smoke:bind-fence` drives a real
`serveEndpoint` over a real broker and asserts the outcome off the reply on both fence arms, which
is emit to wire to parse. Hand-built inputs grade a parser, not a wire, and the second pair exists
because of that distinction.

Not done here, and stated because a reader will ask: the recovery still authorizes its re-issue on
the detail kind rather than on `error.outcome`. A conformant peer sending only the spec field is
parsed correctly and then not retried, which fails closed. Other pre-effect refusals (args,
class-mismatch, target-mismatch) still omit the field, which §13.3 also requires. Both are the
caller-side half of the same question and are tracked separately from this change.

`REPEAT_SAFE_COMMANDS` stays, and is now documented as a stand-in for a field no responder in this
tree can emit: §13.7 `effect` rides `protocol.v: 2`, the descriptor schema pins `v` to the constant
1, and the registry refuses a non-1. The allowlist derives nothing from a descriptor, which is what
§13.7 forbids; it is first-party knowledge of first-party endpoints and fails closed. A tripwire in
`smoke:unfenced-responder` asserts the pin, so the pin cannot be lifted without the allowlist being
named.

## Measured against a live duplicate-effect reproduction

An independent reproduction of the duplicate-effect defect, written by another contributor to
characterise it and not to grade this branch, was run unchanged against this branch. Two symmetric
managers on one loopback broker, effects counted at both responders' own workspace roots rather
than inferred from the caller.

**Zero duplicate effects in 168 trials, no exceptions.** One counter, one denominator. On the
branch it was written against, the same harness produced duplicates in every run.

The mechanism is that the responder-side fence refuses before the handler, so the caller-side
currency path is never reached and there is nothing left to retry. The remedy is not a better
caller-side guard; it is that the state the guard fires on stops occurring.

The residual, from the 96 instrumented trials:

| quantity | value |
| --- | --- |
| calls that split, repair attempted | 43/96 = 44.8% |
| repairs that healed into a success | 25/43 = 58.1% |
| caller-visible failure | 18/96 = 18.8% |

Quoted against the instrumented runs only, and not for tidiness: the two populations do not share a
rate (18.8% against 34.7% uninstrumented), so pooling them averages populations that disagree and
then sets that average beside a split rate and a heal rate drawn from one of them. No mechanism is
offered for the gap; run length is confounded with the counter and it may be ordinary variance.

**So the split is not removed and this PR does not claim it is.** A re-resolve re-enters the same
anycast and re-rolls the same dice, healing a little over half the time. What changes is what a
lost roll costs: a silent duplicate write becomes an honest refusal that states no effect occurred.
Claiming the split is handled would be an overstatement. Claiming the silence is gone is exact.

Two limits travel with those numbers. The reproduction covers two symmetric local managers on one
box, so zero duplicates shows the fence closes the case that reproduction reaches, not that the
case is unreachable under an asymmetry this setup cannot produce. And the reporting contributor
corrected their own strongest claim without being asked: "every refusal carried the marker" was 30
of 43, since two runs predate that counter. 30 of 30 checked carried it. The duplicate count is
measured complete; marker coverage is measured partial.

## What this change trades

The re-issue on a bind refusal is deliberately not gated by `REPEAT_SAFE_COMMANDS`, because the
responder states the command did not run, which makes the next attempt a first attempt rather than
a second. The caller does not take that claim on faith: the reply subject must not attribute the
refusal to the bound instance, the `boundTo` must be the block this request carried, and the
`servedBy` must be what the reply subject attributes it to. A refusal failing any of the three is
`internal`.

Derivable is not verifiable. A responder that ran the command and then sent a well-formed refusal
would pass all three checks. That cannot happen for a `serveEndpoint` responder, where the marker
is set ahead of every effect by construction; it can for a third-party implementation. The change
removes duplicates arising from the split and adds a dependency on a responder's pre-effect claim,
which is the same trust §13.3 already asks for, and is new in this tree.

## Gates

`smoke:bind-fence` (20 cells), `smoke:unfenced-responder` (33), `smoke:bind-fence:live` (13),
`smoke:manager-restart-fence` (22), `smoke:instrument-instance-pin` (55, run five consecutive times
because its re-issue can legitimately split again and the flake surface had to be measured rather
than assumed), `smoke:ep-envelope` (70). Mutation sets: 5 killed on the unfenced driver, 3 on the
pin re-cut, 4 on the fence, 3 on the recovery, and 7 on the re-issue and outcome work, each with
its expected red cell registered in the config **before** the run.

Those last 7, named because each grades a claim the body makes: widening the inner catch back over
the re-issue; lifting the descriptor's `v` pin to admit 2; `pickError` dropping the outcome;
`pickError` passing an unrecognised outcome through; `toEpError` not emitting it; and each fence
arm separately failing to set `not-executed`. One of them found something before it ran: the parse
side had **no coverage anywhere in the tree**, so the harness refused to mutate code no suite reads
and reported an error rather than a pass. A conformance fix with no cell is indistinguishable from
no fix, and the interop half cannot go red on the author's own side by construction, because the
party being wronged is the peer.

One mutation is expected to survive and is run outside the configs deliberately: removing the fence
leaves `smoke:unfenced-responder` green, because that suite's responder is hand-rolled and never
loads the fence. Its partner run kills the same mutation on `smoke:bind-fence`. A survivor reported
without its partner reads as a hole in the guard.

SPEC §13.2 and §13.3 carry the normative rules; `docs/control-surface.md` is updated.
`spec/cotal.schema.json` needs no entry for `bind`: it is generated from the `CotalMessage` root and
models no endpoint envelope at all, so §13.3's tables are their normative shape. Bringing §13
envelopes under the schema is real work and belongs in its own change.

## Rebase notes

Rebased onto `5ded759d`. Two things in that replay are worth stating, because both would have
passed silently.

**The `smoke:ci` chain was merged as a union, not by taking a side.** Both branches append to one
long `a && b && c` line, so either side taken whole drops the other's suites while the file still
looks complete, and a dropped gate stops running with nothing red to say so. Verified by set
comparison in both directions: nothing lost from either side.

That check also caught the opposite failure, which is the one worth remembering: **a union merge
can add but cannot express a deletion.** An intermediate commit here added a suite that a later
commit removed, and replaying the union resurrected it, leaving a chain entry that was defined
nowhere, on either side. `pnpm smoke:ci` would have died partway on a missing script. So the
chain is now also checked for the property that catches that: every entry resolves to a defined
script.

A later replay produced the twin of that trap, and it is worth stating because it is invisible for
the opposite reason. The same script name appeared **twice** in `package.json`, both lines
byte-identical: a union merge cannot express "this line already exists" any more than it can
express a deletion. `JSON.parse` keeps the last, so behaviour was unaffected, the file was valid,
every tool read it correctly, and nothing anywhere could go red. A dropped gate still looks like a
gate and never runs; a duplicated one runs twice and no parser objects. Removed here, and the check
for it has to read the raw text rather than the parsed object, since the parser is what hides it.

The chain at this tip: 235 entries, no duplicates, every name resolving to a defined script whose
file exists.

**The line-anchored SPEC citations were replaced with phrase anchors rather than shifted.** They
were already wrong before this rebase. The `13.2` anchor sat about 180 lines from its subject and
the `13.5` one landed on a `replyExpected` table row, so remapping them by the rebase delta would
have preserved a wrong pointer precisely. A citation that resolves to the wrong sentence is worse
than a dead one, because the reader gets a confident wrong answer. Phrase anchors do not rot when
the document grows.


davidfarah2003:
---

## Review round 1: five findings, all accepted

A focused pass over the concurrency behaviour turned up five real issues, one of them more serious
than anything in the original change. All are fixed; the tip moved to `df00a496`.

### The one that mattered: the retry contradicted the message

The reworded split error tells the operator not to retry because the queue winner may have executed
the command. `Endpoint.invokeService` catches that exact `failed-precondition`, drops the cached
resolve, and **invokes again**, automatically, one frame below the warning.

That recovery was written for one reading of the code ("the describe-bound incarnation is gone, so
the first invoke never ran, and a second is a repair"). The describe-bound currency check also
raises it when a *different live instance* wins the class queue and **replies**, which is an ordinary split,
where the request was received and executed and the error is raised afterwards. Re-invoking there
runs a mutating command twice: the "one spawn, several seats" shape.

The refusal now carries an `ai.cotal.ep.unbound-responder` detail marking the one fact the code
alone cannot express: a responder handled it.

**The first fix was wrong in the other direction, and the suite caught it.** A blanket rethrow made
its own warm-up `ps` go red: in a two-manager space roughly half of all class-queue calls split, so
surfacing every one breaks ordinary reads for nothing. The retry is therefore withheld only for
`GOAL_BEARING_COMMANDS`, where a repeat *creates* something. That is a conservative approximation
and is documented as one: commands carry no idempotency declaration and every manager command is
`class: "ephemeral"`, so the resolved surface cannot distinguish a read from a mutation. `despawn`
and `define-persona` still retry and are convergent by construction. A contract-level safety
annotation is the general fix and is a SPEC change, not made here.

### The cleanup that wasn't

The previous commit claimed the permission watch was released "on every exit". It was not.
The transport's status stream is a `QueuedIterator` whose generator parks on an internal signal
await, so a queued `return()` does not run until the next status event, which on a healthy
connection may never arrive. Measured: 100 parked `return()` calls left 100 listeners, which a
single dispatched status event then collected. Release is now `stop()` (declared on `QueuedIterator`
in `core.d.ts`; `status()` is narrowed to `AsyncIterable`, so it is reached structurally and its
absence fails loud rather than leaking).

### Also fixed

- Both split messages conditioned the mutation caveat on `if "<command>" mutates`, so they fired
  unconditionally, including for reads, where nothing can be duplicated and re-issuing is safe.
- The `cli-on-instance` coverage note claimed `spawn`/`despawn` were "covered transitively". They
  are not: each reaches the seam through its own forwarding site, and dropping the argument at
  either leaves every cell green. The note now says so and names them as needing their own cells.
- The full-instance-id behaviour had no regression assertion (the cell checked an 8-char prefix,
  which the old abbreviated header also satisfied). It now asserts the full token.

### Proof

| mutation | expected red | result |
| --- | --- | --- |
| retry guard → `if (false)` | `a CREATING command's split SURFACES instead of being swallowed by a second invoke` | KILLED |
| `statusStream?.stop()` → `void 0` | `12 resolves leave the listener count where they found it` | KILLED |

`instrument-instance-pin` is now 25 cells, stable over five consecutive runs. Green: `cli-on-instance`,
`instrument-instance-pin`, `ep-invoke`, `ep-serve`, `manager-on-route`, `inst-route-grant`,
`inst-route-enforce`, `manager-scatter`, `manager-multi-live`, `queue-win`, `ps-operator-path`,
`manager-service-invoke`, `manager-service-ops`, `manager-split`, `control-reply`,
`goal-sibling-race`, `goal-epoch-fence`.

### Reported, not fixed

`invokeService` absorbs a class-queue split with exactly **one** re-resolve, and that re-resolve can
split again, so an ordinary read surfaces roughly one time in four with two managers. This is
pre-existing (the single retry predates this branch and the second failure has always propagated),
so it is recorded rather than changed here; the read cells retry with an explicit bound and say why.




---

**Correction (posted after `main` moved).** The paragraph above says the general fix is a spec change not made here. That is no longer true of this tree: `SPEC.md` on `main` now carries `effect` (§13.7) and `error.outcome` (§13.3), landed in `cf05bfb6` and dated 2026-08-16, after the text above was written. This branch does not yet implement either: `EpError` is `{code, message, details?}`, and `pickError` in `endpoint-envelope.ts` rebuilds exactly those three fields, so an `outcome` sent by a conformant peer is silently dropped. The bind fence in `endpoint-serve.ts` refuses before the handler, which is the case §13.3 says MUST carry `not-executed`, and it sets no outcome. Treat the sentence above as describing the state when it was written, not the state of the spec now.

davidfarah2003:
## Review round 2: the retry guard was still wrong, and `purge` is the proof

The previous round withheld the split auto-retry only for `GOAL_BEARING_COMMANDS` (spawn/launch), on the theory that a duplicated effect only hurts when it **creates** something. Closer review broke that with a measured counterexample.

**`purge` is neither goal-bearing nor convergent.** Its handler reaches `clearSpaceHistory`, a real `STREAM.PURGE`, so "the first one already deleted everything" is false the moment anyone is still publishing. Measured on a live two-manager mesh: the second execution deleted a message published *after the first purge completed*. The operator retried nothing; the client did it.

The boundary is not "creates something". It is "changes anything", and core cannot tell which is which, because a command carries no idempotency declaration and every manager command shares `class: "ephemeral"`.

### The fix is an allowlist, and the polarity is the substance

`REPEAT_SAFE_COMMANDS` names the commands whose second execution is observably indistinguishable from one (the reads plus `describe`). Everything else surfaces the split to its caller, carrying the marker that says a responder *did* handle the request.

A denylist was the smaller change and was rejected deliberately: it **fails open**. Classifying only `purge` would leave the other seven `manager.admin` commands auto-retried, and the next command anyone adds would be auto-retried by default with nothing going red. The allowlist fails closed: an unclassified command is surfaced, not repeated.

Membership is written out literally rather than derived from `MANAGER_READ_COMMANDS`: retry-safety and grant class answer different questions and must be able to disagree. Deriving it would silently widen retry-safety the day a mutating command joins the read tier for grant reasons.

Scope, stated plainly: every non-read command on every endpoint now surfaces a split instead of self-healing. That can only bite in a multi-instance space, since a single responder never splits, so single-instance deployments see no behavioural change. Non-split `failed-precondition` paths are untouched: the marker is attached at exactly one site (the describe-bound currency check), so a genuinely superseded incarnation still retries exactly as before.

### The test gap was real, and correcting the prose did not close it

`spawn --detach`, `stop`, and `attach` each forward `--on` at an **independent** call site. Dropping any one of them type-checks, because the parameter is optional by design, and left every existing cell green. The previous round corrected an over-read in the coverage comment and stopped there; that protects nobody.

Cell 5 now aims each command at a valid-but-absent instance and asserts the pinned-describe deadline. Non-mutating by construction: an honoured pin has no responder on that rail, so the command deadlines before it can reach an execute. `attach` is covered rather than scoped out.

The coverage boundary is rewritten to say how the chain is actually split: the shared tail (`resolveControlTarget` → `connectOrExit` → mint) is graded once by cell 3; the per-command heads are graded separately by cells 4 and 5. Neither covers the path alone. Still not graded: that a pinned `spawn`/`stop`/`attach` **completes** against a real instance. Those mutate, and that is a different fixture.

### Mutation-proofs

| mutation | reddened assertion | marks |
| --- | --- | --- |
| add `purge` back to the repeat-safe allowlist | `a DESTRUCTIVE non-creating command's split SURFACES` | 30 / 32 |
| `spawn.ts`: `values.on` → `undefined` | `spawn --detach HONOURS --on` | 12 / 13 |
| neutralize `pinForTarget`'s `v.on` | `stop HONOURS --on` (attach too) | 11 / 13 |

Seventeen suites green.

### Confirmed, not fixed here

A read fails outright roughly **1 time in 4** in a two-manager space: the client absorbs a split with exactly one re-resolve, and that re-resolve can split again. Verified pre-existing at the merge base (`b33aae7b`); this branch does not own it. Operators seeing intermittent `ps`/`inspect` failures are seeing the split, not flaky agents.

The general remedy remains open and is deliberately not attempted here: a safety/idempotency annotation on the command contract plus an effect outcome in the reply, so a caller can *know* whether the first execution landed instead of a client-side heuristic guessing. That is a wire change.





---

**Correction (posted after `main` moved).** The paragraph above says the general fix is a spec change not made here. That is no longer true of this tree: `SPEC.md` on `main` now carries `effect` (§13.7) and `error.outcome` (§13.3), landed in `cf05bfb6` and dated 2026-08-16, after the text above was written. This branch does not yet implement either: `EpError` is `{code, message, details?}`, and `pickError` in `endpoint-envelope.ts` rebuilds exactly those three fields, so an `outcome` sent by a conformant peer is silently dropped. The bind fence in `endpoint-serve.ts` refuses before the handler, which is the case §13.3 says MUST carry `not-executed`, and it sets no outcome. Treat the sentence above as describing the state when it was written, not the state of the spec now.

davidfarah2003:
## Review round 3: retry safety is keyed by endpoint, and `models` comes off the safe list

Two holes in the name-keyed allowlist, both found on a second read of the allowlist, both verified before acting.

**Endpoint.** `invokeService` is endpoint-agnostic, so a flat list of command names handed every endpoint in the mesh a judgement derived from the manager's vocabulary, so a third-party endpoint offering `list` or `ps` inherited "safe to run twice" from a table that had never seen it. The allowlist is now keyed by endpoint, read through a private module-load snapshot, and an **unknown endpoint has no repeat-safe commands at all**. Same fail-closed default, one level up.

**Arguments.** `models` reads a catalog, unless called with `{refresh: true}`, which reaches the connector's `listModels({refresh})` and, for OpenCode, shells out to `opencode models --refresh`: a provider round-trip that rewrites a cache. Same command name, same grant class, opposite answer, decided by an argument the classifier cannot see. Encoding per-command argument rules would reintroduce exactly the fail-open shape this replaced, so `models` is simply not repeat-safe. The cost is that a plain `cotal models` surfaces a split instead of absorbing it in a multi-instance space.

That exclusion is the clearest statement of what this table is: a client-side stand-in for something the wire does not carry. The general fix, a safety annotation on the command contract plus an effect outcome in the reply, remains open and is deliberately not attempted here.

`describe` stays exempt without an endpoint, and the source now records why, with two independent fences: a cluster document declaring the name is invalid at registration, **and** `serveEndpoint` throws on any def whose command is `describe`, at the top of the defs loop before the grant, class, and duplicate checks. No handler can be attached. The residual is written down rather than assumed. `describe` is authorization-scoped, so the exemption rests on transport and telemetry not being state the endpoint owns; a future `describe` that consumed a grant or decremented a quota would be a write, and the exemption would be wrong for a reason having nothing to do with handlers.

### A test of mine passed green while proving nothing

Worth recording, because mutation-proof alone could not have told me.

The cell that sweeps every `manager.admin` command through the guard asserted only that a doctored resolve cache survived each call. A command that never reaches a responder leaves that cache untouched too, so a vacuous pass and a real pass were indistinguishable. **All seven commands were vacuous**: they were invoked with `{}`, which fails `required` on every admin input schema, so the envelope refused them client-side and none was ever published.

It surfaced only because a mutation replacing the guard with a hardcoded `command === "spawn" || command === "purge"` reported SURVIVED at the full baseline, twice. The suite was not testing what its assertion names claimed.

Each command now carries schema-valid but deliberately nonsense arguments, and the responder-answered marker is required *before* the cache assertion means anything, so a command that cannot reach the guard is red, not absorbed. The same mutation is now KILLED.

A cell in the same file also inherited the pre-existing 1-in-4 read failure by calling `ps` once; the bounded tolerance that a neighbouring cell already had is now shared and documented, with 5 of 5 clean consecutive runs.

### Mutation-proofs

| mutation | reddened assertion | marks |
| --- | --- | --- |
| hardcode the guard to `spawn`/`purge` | `none of them is quietly retried` | 36 / 38 |
| make the classifier endpoint-blind | `an UNKNOWN endpoint has no repeat-safe commands` | 36 / 38 |
| add `purge` back to the repeat-safe set | `a DESTRUCTIVE non-creating command's split SURFACES` | 30 / 32 |
| `spawn.ts`: `values.on` → `undefined` | `spawn --detach HONOURS --on` | 12 / 13 |
| neutralize `pinForTarget`'s `v.on` | `stop HONOURS --on` (attach too) | 11 / 13 |

Seventeen suites green; pin suite 38/38 across five consecutive runs.





---

**Correction (posted after `main` moved).** The paragraph above says the general fix is a spec change not made here. That is no longer true of this tree: `SPEC.md` on `main` now carries `effect` (§13.7) and `error.outcome` (§13.3), landed in `cf05bfb6` and dated 2026-08-16, after the text above was written. This branch does not yet implement either: `EpError` is `{code, message, details?}`, and `pickError` in `endpoint-envelope.ts` rebuilds exactly those three fields, so an `outcome` sent by a conformant peer is silently dropped. The bind fence in `endpoint-serve.ts` refuses before the handler, which is the case §13.3 says MUST carry `not-executed`, and it sets no outcome. Treat the sentence above as describing the state when it was written, not the state of the spec now.

davidfarah2003:
## Correction: "single-instance deployments see no behavioural change" was too broad

I claimed twice in this thread that withholding the retry can only bite in a multi-instance space, because a single responder never splits. That is not right, and the correction is worth stating plainly since it changes who is affected.

**One responder live at a time is not the same as one logical instance over time.** A *new logical manager started from a different root* registers a new instance id, and it can answer a handle whose cached resolve was bound to the previous one. The describe-bound currency check fires on that reply, because a different instance answered than the one the handle resolved against, and the command now surfaces rather than being silently re-invoked. So a deployment that never runs two managers concurrently can still reach this path, through logical replacement.

A **same-root restart is unaffected**: it preserves the instance id, so the bound handle stays current and the guard is never consulted.

The behaviour in the replacement case is still the safe one, since the caller is told a responder handled the request instead of having it executed a second time against the new instance, so nothing about the change follows from this. Only my characterisation of the blast radius was wrong, and anyone reading "single-instance deployments are unaffected" as a reason not to look would have been misled.

Nothing else in the earlier comments is affected: the `purge` counterexample, the endpoint-qualified allowlist, the `models` exclusion, and the vacuous-sweep report all stand as written.

PR #452, opened 2026-08-15
test(core): grade the split-retry guard's allowlist per ENDPOINT, not per name

davidfarah2003:
Rebased onto `bb5086e6` and re-cut. The guard this suite grades was replaced while the suite was open: it now consults `isRepeatSafeCommand(endpoint, command)` rather than a command-name allowlist, so the earlier cells graded a boundary that no longer exists.

## What this measures that nothing else can

`invokeService` is endpoint-agnostic. The classifier is keyed on the **pair**, and this is the one place a `list` served by somebody else's endpoint could inherit the delivery endpoint's safety judgement for free, purely because the names match. A manager-side suite can only assert that direction against an endpoint string it invented; this fixture serves its own endpoints, so it can measure it against one that really serves and really answers.

## The discriminating pair

Two endpoints, **one command name**, one forced split:

| addressed to | `list` | effect |
| --- | --- | --- |
| an endpoint named `delivery` — where the table lists it | retried | executes **twice** |
| an endpoint the table has never heard of | surfaces | executes **once** |

Same contracts, same handler body, same instant, same split. The only difference between the two outcomes is the endpoint the command was addressed to. A guard that consulted only the command name would make these identical, and a suite driving a single endpoint could not tell them apart.

`append` on `manager` then shows the pairing is not merely per-endpoint either: being on the table does not launder an endpoint's unlisted commands. It uses a second listed endpoint rather than reusing `delivery`, whose instances from the first measurement are still live — sharing would stage that measurement with the previous one's outcome rather than with the fixture.

## The describe exemption's second fence, executed

`describe` is repeat-safe on **every** endpoint. That is only safe because no endpoint can serve its own: registration refuses a cluster command by that name, and `serveEndpoint` refuses a def by that name unconditionally, at the top of its defs loop — before the grant, class, and duplicate checks, so no ordering slips past it. This attempts the registration for real and asserts the throw, because a source comment citing two files rots silently if either stops throwing.

## Absence is never the evidence

Every claim resting on something *not* happening is preceded by a positive control proving the request reached a responder at all — the priming cell asserts the ledger grew, attributed to the instance that served it. A command that never leaves the caller also leaves the ledger untouched, and without the control that is indistinguishable from the guard holding. A sibling suite shipped green twice on exactly this: its arguments failed `required` client-side, so nothing was ever published and the empty result read as success. Every invoke here carries schema-valid arguments.

## What was removed

The previous KNOWN GAP cells are **gone rather than inverted**. They asserted that a mutating command off the goal-bearing list is retried and executes twice; that gap is closed, since unclassified now defaults to unsafe, so the behaviour they characterized no longer exists to characterize.

## Gates

11 cells, green. `pnpm typecheck` clean.

Mutation, predicted before the run: make the guard ignore the endpoint (`isRepeatSafeCommand("delivery", command)`) → the three `ledger/list` cells redden and nothing else. **KILLED, 8 marks against a baseline of 11** — exactly the three predicted.

One correction worth recording. My first attempt at that mutation substituted a name-only lookup over `REPEAT_SAFE_COMMANDS`, and mutation-proof reported KILLED. It was worthless: that symbol is not imported into `endpoint.ts`, so the suite reddened on a `ReferenceError`, not on the guard's semantics. It surfaced only because the kill set was 4 cells where I had predicted 3, and enumerating the difference showed every failure carrying the same `ReferenceError`. Red is not proof that the test noticed the *behaviour*; a mutation that crashes reddens everything downstream of it.


davidfarah2003:
Superseded by #451, which grades the split-retry allowlist per (endpoint, command) and adds the responder-side fence. The branch stays for reference.

PR #454, opened 2026-08-15
feat(web)!: authenticate the console's HTTP surface with a single-use launch link

davidfarah2003:
Slice 1 of the console's send work. **It contains no send route** — it closes the precondition that
blocked one, and it is worth landing on its own because the hole it closes exists today.

### The measurement this starts from

    req.headers read in implementations/web/src/web.ts    0 times
    control: req.url read in the same file                3 times

Not weak authentication — none. Nine matches for `authorization|bearer|cookie|token` exist in that
file and not one is a check on an incoming request; they are the process authenticating **itself** to
the broker. A grep for auth words returns hits and the surface is still unauthenticated.

Binding loopback defends against other *hosts*. It does not defend against another *process* on the
same machine, and it does not defend against a page in the operator's own browser issuing requests to
`http://127.0.0.1:7799`. What that reaches today is the whole mesh read path plus a channel-delete
POST. A send route on this surface would have handed the operator's voice to anything that can open a
socket.

### What changed

- A one-time launch token is minted per process and printed in the URL that is opened. It is
  exchanged **exactly once** for an `HttpOnly; SameSite=Strict` session cookie, and the exchange
  answers `302` so the spent secret leaves the address bar and the browser history.
- **Every route runs behind the gate**, asserted positionally: a gate placed below a route leaves
  that route unauthenticated and nothing in its own behaviour would say so.
- Refusals name which condition failed — `unauthenticated` (401), `launch-token-already-used` (401),
  `cross-origin` (403) — in the body, so a caller reading only the body still learns which. Four
  different failures reported as one is the same defect as a failure reported as success.
- **Origin is checked before the session.** A cross-site request arrives without the cookie anyway, so
  checking the session first would report every one of them as merely `unauthenticated` and the
  operator would never learn another site was talking to their console.
- `--detach` keeps working through a **second, separate credential**: the child writes a `0600`
  session file after `listen()` succeeds, and the parent presents that nonce as a header when polling
  `/api/meta` for readiness. A second credential rather than an exempt route, because **an exempt
  route is unauthenticated for everyone forever**, and the next person to add a field to it will not
  know that.

### Verification

`smoke:web-console-auth` — **128 cells, `rc=0`**, in two layers. The exported gate is driven directly;
the gate block is then **lifted out of the shipped `handleRequest` and executed** over a recording
`ServerResponse`, because a correct gate wired to a handler that ignores its verdict is an
unauthenticated surface with a passing test.

The suite is run under the default umask **and under `umask 077`**, both `rc=0`. That is not
belt-and-braces: one fixture created its "permissive" file with `mode:` at creation, which is masked
by the process umask, so under a restrictive umask the file was born `0600` and the suite failed its
own non-vacuity check **before reaching any shipped code**. A green that cannot be re-derived in
another operator's shell is not a green.

This branch went through **ten adversarial review rounds**. Every finding was re-run here before any
fix and again after; each mutant was grep-confirmed present in the file, and each restore was verified
with `git diff` against `HEAD` rather than against a backup. Where a mutation survived, it was also
proved **non-equivalent** — that the surface was genuinely broken — rather than only that a cell
stayed green.

| mutation | before the fix | predicted first red cell | after |
| --- | --- | --- | --- |
| refusal branch loses its `return` (keeps `res.end`) | **SURVIVED** | `…and EXECUTION STOPS: the routes below the gate are never reached` | KILLED |
| `/api/meta` moved above the gate | **SURVIVED** | `the gate runs BEFORE EVERY route dispatch` | KILLED |
| `secretEquals` reduced to length equality | **SURVIVED** | `a same-length readiness nonce differing in ONE character is refused` | KILLED |
| readiness check moved above the origin check | **SURVIVED** | `a cross-origin request carrying a VALID readiness nonce is still refused` | KILLED |
| static dispatch (`PAGE[path]`) moved above the gate | **SURVIVED** | `the handler's first THREE STATEMENTS are …` | KILLED |
| refusal predicate conjoined with `path === "/api/roster"` | **SURVIVED** | the path sweep, incl. `/api/channel/delete` | KILLED |
| refusal predicate conjoined with `req.method !== "POST"` | **SURVIVED** | the method sweep | KILLED |
| chmod moved to AFTER the write | **SURVIVED** | `the secret is written into an ALREADY-0600 descriptor` | KILLED |
| `secretEquals` reduced to `===` | **SURVIVED** | `secretEquals compares with timingSafeEqual` | KILLED |
| exchange conjoined with `path === "/graph"` | **SURVIVED** | `an exchange on / sets the cookie and redirects THERE` | KILLED |
| route dispatch appended to the END of the query line | **SURVIVED** | `the handler's first THREE STATEMENTS are …` | KILLED |
| `readiness` dropped from the session file's JSON | **SURVIVED** | `…AND the readiness nonce` | KILLED |
| dummy compare deleted from the length-mismatch branch | **SURVIVED** | `…and the LENGTH-MISMATCH branch still does the work before failing` | KILLED |
| a route inserted BETWEEN the gate call and the refusal branch | **SURVIVED** | `…and statements 4 and 5 are the REFUSAL and EXCHANGE handlers` | KILLED |
| a side effect inside the still-named `path` initializer | **SURVIVED** | `…and the path parse is EXACTLY a parse` | KILLED |
| refusal conjoined with a bracket-notation header read | **SURVIVED** | `the refusal condition is EXACTLY verdict-only` | KILLED |
| the timing-safe call kept but made unreachable | **SURVIVED** | `secretEquals RETURNS the timing-safe comparison` | KILLED |

Each fix is pinned by its own named cell; the predictions above were written down **before** each
mutation was applied, and a mutation is only counted as killed when it reddens the cell that was
predicted for it.

**Two of these are worth reading, because they are defects in what a *test* could see rather than in
what the code did.**

The first is the missing `return`. The harness lifts the gate block and runs it in isolation — so
nothing follows it, and *"did execution stop?"* was a question the fixture could not ask. Response
*content* was asserted because content was all it could observe, and a 401 that then falls through
into the routes writes byte-identical output to one that returns: an unauthenticated `POST` still
reached the channel-delete. There is now a route sentinel after the lifted block, with a **positive
control** — an allowed verdict must *reach* it, so the refusals' non-reach means something.

The second is the cell that asserts nothing precedes the gate. It has been beaten four times, each
time by a shape the previous version did not model: first it compared the gate against **one named
route**; then against an **enumeration of route syntaxes**, which silently omitted the shipped static
dispatch; then it measured from the wrong anchor, leaving the space *between* the path and query
declarations unguarded; and finally it counted **trimmed source lines**, so a route dispatch appended
after the query declaration's semicolon — same physical line — left it green. That last one was
measured to answer `200 {}` to an unauthenticated `GET /api/meta` with `gate.check` **never called at
all**, against a control showing `401` on the unmutated file.

A fifth shape sat **below** the gate rather than above it: a route inserted between the gate call
and the refusal branch lets the gate run, compute a refusal, and have that verdict discarded by a
route that returns first — measured at `200` for an unauthenticated request with the gate consulted,
against `401` on the clean file. A verdict that is computed and ignored is not a gate, so the
handler's first *five* statements are now pinned: the two parses, the gate call, then the refusal and
exchange handlers.

A list of route shapes can always omit one, and the omission is invisible. So the cell is now written
in the **grammar of the property**: the shipped file is parsed with the TypeScript compiler and the
handler's first three *statements* are required to be the path parse, the query parse and the gate
call. Lines are not statements, and only one of the two is the property being claimed.

Three product fixes came out of these rounds. The origin check compared only `URL.host`, so
`https://<allowed host>` — a different origin to every browser — passed it. `writeFileSync`'s `mode`
applies at **creation only**, so a stale `web.session` kept whatever mode it had while holding the
launch URL and the readiness nonce; the file is now opened and `fchmod`ed on the **descriptor**, and
narrowed **before** the bytes go in rather than after, because a credential exposed for an instant is
exposed. And the origin allow-list was built from hand-written strings while the incoming header is
normalized through `new URL(...).origin`, which **drops the default port** — so on port 80 the console
would have refused its own page.

The two layers are now **causally joined**. For most of this branch's life the gate was driven with
real inputs and the handler was driven with hand-built verdicts, so the suite could hand the handler a
refusal the gate itself could never mint — and a divergence between them was invisible. A real gate
now drives the real handler prefix end to end: a bare request refused `401`, the real launch token
exchanged for a cookie and redirected to `/`, that same spent token replayed and refused as
`launch-token-already-used`, the real readiness nonce accepted, and a cross-origin request refused
`403` by the condition the gate itself chose.

Four of the survivors above share one cause worth naming: in each case the assertion was **weaker than
the property** it claimed. Statements were pinned by their declared *name* while an initializer is
code; the refusal condition was checked against a *list of forbidden spellings* rather than for
equality, and one more spelling walks through; and the constant-time comparison was pinned by its
*presence in the file*, which an unreachable copy satisfies while the live return path uses a
byte-by-byte compare that exits early. Presence is not dataflow, a name is not a body, and an
enumeration is not an equality.

`tsc --noEmit` on the web project: `rc=0`, no errors, the exit code read from an EXIT-trap artifact
rather than from a pipe. Repo-wide `typecheck`: `rc=0`, zero `error TS`. Dependencies were installed
inside a dedicated worktree with `--frozen-lockfile --prefer-offline`; no other checkout was touched.

### Registration and scope

Appended to `smoke:ci` rather than inserted — suites are assigned to runners by list position, and
the measurement is **0 of 221** changing runner label and **0 of 221** changing on-runner
predecessor. `smoke:gate-inventory` re-run as a postcondition, `rc=0`, with a control confirming it
prints only the *ungated* set (so a new entry appearing zero times is correct rather than invisible).

### Known limitations, stated rather than left for review

- The cookie **cannot be `Secure`** over `http://127.0.0.1` — a Secure cookie would simply never be
  sent. It is `HttpOnly` + `SameSite=Strict`, and the origin check carries the rest.
- `SameSite=Strict` is asserted as **emitted**, not as **enforced**; enforcement belongs to the
  browser and would need a browser to test.
- A session is bound to the hostname it was opened on, so one started on `cotal.localhost` does not
  carry to `127.0.0.1`. Documented.
- Once the token is spent, any `?k=` value answers `launch-token-already-used`, which reveals that a
  token existed and was used. That is information the operator wants (a replayed link) and it does
  not distinguish a correct guess from a wrong one.
- Not covered by any cell here: a real browser's cookie handling, and a real cross-site request.
- **The four known survivors of this suite, each with its measured boundary.** A further independent
  adversarial review reproduced the interim guard's documented boundary and found three more
  bypasses. All four are recorded in the suite itself, beside the guard each one walks past, so a
  reader meets the limit in the file rather than only here. Each was applied to a clean tree, ran
  **128 passed / 0 failed with `rc=0` from an EXIT-trap artifact**, **passed the package typecheck**,
  and was restored with the source blob verified by hash — so each is a *compiling, test-passing*
  bypass and the compiler is not a backstop for any of them.
  1. Replacing the gate object the handler closes over, below the handler, opens the surface and is
     **not caught** — every cell here builds its own gate. Proved non-equivalent rather than merely
     surviving: the shipped `check` answers `{refuse:"unauthenticated"}` for a bare request and
     `{refuse:"cross-origin"}` for a foreign Origin, while the mutant answers `undefined` for every
     input, turning each named refusal into a silent allow.
  2. A nested declaration that **shadows the gate factory** — same name, same shape, always allows —
     survives while the pinned text and the gate-to-handler adjacency stay **byte-identical**. The
     cells resolve the test module's binding; the composition function resolves the nested one. The
     general point is worth more than the case: exact text is fail-closed against harmless
     reformatting, but it does not establish **identifier identity**.
  3. Ending a refusal's body **before** its status write survives, because the response recorders here
     retain only final field values and do not model **call order**. On a live Node HTTP server the
     shipped ordering answers `401` and the swapped ordering answers **`200`** with body
     `{"error":"unauthenticated"}` — the later `writeHead` throws after the body is already on the
     wire. The condition stays named in the body but stops being client-distinguishable from success
     by status.
  4. An early successful empty response inside the **server callback**, before the handler runs,
     survives because no cell here drives that callback. TypeScript did not flag the now-unreachable
     handler call.

  **What the suite does still prove**, so the above is not read as "the tests are worthless":
  deleting the handler's refusal response entirely **does** die on the named `401` cell. The
  handler-prefix cells are sound *for the code they reach*, and no survivor shows a covered assertion
  to be wrong — each walks around that prefix or exploits ordering the fakes do not model.

  **The durable fix**, named and deliberately deferred: drive a real request through the actual server
  wiring, or refactor the server/handler composition into a directly executable factory a test can
  invoke end-to-end. A cheaper substitute that cannot drive the running server would be a fixture
  wearing an integration test's name — reproducing defect (3) above. Filed as a follow-up rather than
  held as a blocker, because every survivor lives in what the suite can **see**, not in what the
  shipped code **does**.

- The wrong-scheme refusal is **defence in depth, not a demonstrated exploit**. A page would have to
  be served from `https://<allowed host>:7799`; the console binds `127.0.0.1` only, so `[::1]:7799`
  could host one. Ordinary hostname cookie scoping plus `SameSite=Strict` still stops the
  session-bearing request, so this closes a correctness gap rather than a live attack.

### The reviewed hash, and what moved after it

The commit an independent adversarial review graded is **`150a302f`**. This branch's tip has moved
since that grade, and the delta is stated here rather than left silent — a tip that moves after a
grade is fine when the diff is named, and dishonest only when it is not:

```
$ git diff --stat 150a302f..HEAD
 .changeset/web-console-auth.md                  |  1 -
 implementations/web/smoke/console-auth.smoke.ts | 48 +++++++++++++++++++++++++
 2 files changed, 48 insertions(+), 1 deletion(-)
```

**At that grade, comment and changeset only — zero product lines.**
`implementations/web/src/web.ts` was byte-identical at both hashes (blob `96c32b61`).

**That is no longer true, and saying so here is the point of this section.** A second, independent
review of the auth model at `c295f2aa` returned approve-with-findings, and its two substantive
findings are folded in below. Since `c295f2aa`:

```
$ git diff --stat c295f2aa..HEAD
 bin/smoke/dogfood-live.smoke.ts                 | 39 ++++++++++++++++++-------
 implementations/web/smoke/console-auth.smoke.ts | 27 ++++++++++++-----
 implementations/web/src/web.ts                  | 25 ++++++++++++++--
 3 files changed, 70 insertions(+), 21 deletions(-)
```

`web.ts` moves from blob `96c32b61` to `9006f958`. At the new tip the scoped auth suite is
**129 passed, 0 failed**, and `pnpm smoke:dogfood:live` is green (**31 checks**).

### What changed after that review

**The readiness nonce is scoped to `/api/meta`.** The launch token is burned before its session is
minted, but the nonce was neither consumed nor scoped: it lasted the life of the process, was
accepted on every path, and is written into `web.session` beside the link. A reader of that file
held a full-surface credential after the single-use link had been spent. It now opens the one path
its only caller polls, and `web.session` joins the process artifacts so a launch that dies without
running its exit handler does not leave the credential behind.

**The live dogfood suite stopped asserting an open surface.** It fetched `/`, `/app.js` and
`/api/meta` unauthenticated and expected 200. Each site now uses the credential its caller would:
the browser path spends the launch link once and carries the session cookie it mints; the detached
path presents the readiness nonce. It also asserts the unauthenticated refusal, so it fails in both
directions rather than passing if the gate were removed.

### Unproven: the artifact sweep on a crash

`web.session` is declared in `webProcess.artifacts`, so `cotal down` removes it once the process is
confirmed gone. **The suite proves the entry is declared and correct. It does not prove the sweep
runs.**

The entry only matters when the exit handler does not — a `SIGKILL`, say — and every path the tests
exercise is a clean exit, where the handler already removes the file. So every cell here passes with
the entry reverted, and saying "the suite is green" would be evidence of nothing. What the structural
cell does kill is the failure that is otherwise silent: an absent entry, a typo, or a path matching
nothing, each of which leaves the rest of the file green while `down` sweeps nothing. It was checked
by removing the entry and watching the cell go red on its own assertion.

The end-to-end case — kill the process, run `down`, assert the file is gone — is a process-lifecycle
suite. It is tracked as a follow-up rather than folded into this change.

### Known and accepted: the spend-state oracle

After the link is spent, any request carrying `?k=` is answered `launch-token-already-used` rather
than `unauthenticated`. That reveals **that the link was opened** — it does not reveal whether a
guessed token matched, since a wrong token gets the same answer as no token once the real one is
gone. The alternative is a refusal that cannot tell an operator their link was replayed, which is
the more useful failure to be able to see. Stated here rather than left for a reader to notice.

### Running the live suite

`pnpm smoke:dogfood:live` resolves its sandbox through the nearest mesh root. On a machine where a
root has ever been recorded at `/tmp`, `up --detach` writes into that root instead of the suite's
sandbox, and the run fails at a missing `nats.pid` — an error that names a pidfile rather than the
cause, well before it reaches anything this PR changes. It is machine state, not a defect in the
suite; run with `TMPDIR` pointed somewhere whose parents hold no `.cotal`.

### The readiness nonce is a standing credential, not a single-use one

Stated plainly because the rest of this description talks about a single-use link, and the two are
different things. **"Single-use" describes the launch link (`?k=`), which is burned on first
exchange. It does not describe the readiness nonce.** That nonce is written to `<root>/.cotal/web.session`
at 0600, is never consumed, and stays valid for the life of the process — precisely because the
detached parent may poll many times. It is a **standing bearer credential with a process lifetime**,
and anything able to read that file holds it until the console exits.

This PR narrows it from a full-surface credential to the single path its only caller polls
(`/api/meta`), which is what makes the file worth less to an attacker than it was. It does not make
it single-use, and it is not claimed to.

**This is a follow-up, not a reason to hold the PR:** the surface it guards was previously
**unauthenticated and reachable**, so this strictly reduces exposure. Turning the standing nonce
into something short-lived or consumable is a separate change against a surface that is by then
already gated.

### Rebase note

Rebased onto `b6d6d799`. The `smoke:ci` chain moved out of `package.json` into
`bin/smoke/ci-suites.txt`, so the suite is registered by **one appended line** there rather than by
editing the chain string; `package.json` now carries only the script definition. Verified: the
branch adds exactly `smoke:web-console-auth` and removes nothing (set difference checked in both
directions), and all **243** names in the chain file resolve to defined scripts. Re-run on the
rebased tip: console-auth **131 passed / 0 failed**, the chain-file suite **19 / 0**, typecheck clean.


davidfarah2003:
Rebased this onto current `main` (tip `c295f2aa`) to get CI running on it. Every other job is green, but `live` now fails, and it needs a decision from someone who owns this surface, so I'm handing it back rather than patching it.

**The failure**

`bin/smoke/dogfood-live.smoke.ts:117`:

```
Error: FAIL: the extension `web` command serves the dashboard — ""
```

That cell does a bare `fetch("http://127.0.0.1:<port>/")` and asserts `status === 200`. This PR puts the console's HTTP surface behind a single-use launch link, so an unauthenticated `GET /` is *supposed* to stop returning 200. The smoke is asserting the contract this PR deliberately replaces. It's not a flake and it's not rebase drift. It is the intended behaviour change meeting a test written against the old one.

The same run also fetches `/app.js` and `/api/meta` bare, so those cells sit behind the same problem.

**Why I stopped**

The mechanical-looking fix is to drive the real operator path: `web.ts` already writes `{launchUrl, readiness}` to the state file for "a detached parent (and an operator who lost the printed link)", so the smoke can read the launch URL, fetch it, keep the `set-cookie`, and send that cookie on the subsequent requests. Roughly:

```ts
// after the server is up, instead of a bare GET:
const state = JSON.parse(readFileSync(join(root, ".cotal", "web.session"), "utf8")); // 0600, beside the pidfile
const launch = await fetch(state.launchUrl, { redirect: "manual" });
ok("the launch link serves the dashboard", launch.status === 200, ...);
const session = (launch.headers.get("set-cookie") ?? "").split(";")[0];
ok("the launch link establishes a session", session.startsWith("cotal_web_session="), ...);
// then thread `headers: { cookie: session }` through /app.js and /api/meta
```

But `implementations/web/smoke/console-auth.smoke.ts:21` says the two things **not** driven, with no cell implying them, are "a real browser's cookie jar, and a real cross-site request". The gate is tested as a unit via `gate.check(req({ cookie: ... }))`, never over real HTTP. So the sketch above isn't completing an existing contract, it's adding the first real-HTTP session coverage of an auth path that was scoped out on purpose.

That's worth doing deliberately or not at all. If the handshake is subtly wrong, `dogfood-live` goes green while proving nothing about the surface it appears to cover, which is worse than the current honest red, because a passing live suite reads as end-to-end proof.

**What's needed**

Either extend `dogfood-live` to the authenticated path as its own considered change (the sketch is a starting point, not a recommendation to paste), or decide the live dogfood cell should assert the refusal instead and the authenticated path stays unit-level. Both are calls about how much of this surface the live suite should own.

Everything else here is ready: merged cleanly onto `main`, `smoke:ci` chain resolved as a union with both directions asserted (no gate dropped from either side), docs bundle regenerated, and the branch is a fast-forward from `d70b7897`.

davidfarah2003:
REJECT @ 4fedcd6f1e3ebe7a3896078bca37ed648b2a78d8

Blocking: the launch token can leak into the detached dashboard's persistent `web.log`. The token is a query parameter, and the request error frame logs raw `req.url` at `implementations/web/src/web.ts:1032`. After exchange, a valid session request such as `/api/activity?k=<launch-token>&limit=bad` passes because the session is checked before `k`, then throws `BadRequest`; the catch logs the complete URL including `k`. This violates the PR's authentication boundary even though the ordinary exchange redirects before serving content. Log only the parsed path or explicitly redact secret query keys.

Security review notes:
- Single use is enforced by setting the closure-held token to `undefined` before session minting, so Node's run-to-completion prevents concurrent double exchange. Replay is 401 `launch-token-already-used`. Unused tokens live until process exit. Restart mints new launch/session secrets and invalidates old sessions. State is intentionally process-local, while `web.session` only persists the launch URL/readiness nonce.
- Launch, readiness, and session secrets use `randomBytes(32)` and launch/readiness comparison uses `timingSafeEqual`; session authorization uses exact `Set` membership.
- The gate precedes SSE `/feed`, all API routes, the destructive POST, and the static allow-list. No WebSocket upgrade handler exists.
- The breaking unauthenticated flow returns a named 401/403 body, and `docs/watch-a-mesh.md` accurately describes exchange, replay, origin binding, and restart.
- The 131-cell auth smoke does construct unauthenticated requests and discriminates route/method/static/SSE/destructive cases, but it has no token-in-error-log leakage cell.

Ran: `gh pr view 454 --json headRefOid` (head matched), `pnpm build`, `pnpm smoke:web-console-auth` (131 passed), source/diff inspection and route/WebSocket/logging searches. Did not run any live, stack-starting, full CI, browser, or prohibited suite.

davidfarah2003:
Follow-up strengthening the blocker with the real detached acceptance path:

Detached mode redirects the child’s stdout to `<mesh-root>/.cotal/web.log` (`web.ts:1142-1152`), and the child prints the complete secret-bearing `launchUrl` (`web.ts:1107-1110`). Unlike `web.session`, `web.log` is opened with mode `0600` only at creation and is never `fchmod`ed. A stale pre-existing permissive log retains its old mode, so launching detached writes a currently valid launch token into a potentially group/world-readable persistent file. The dedicated smoke explicitly tests narrowing a stale permissive `web.session`, but has no equivalent test for `web.log`. This is direct end-user behavior, independent of the earlier raw-`req.url` error-path example.

Required fix: enforce `0600` on the opened log descriptor before spawning the child, and add a stale-permissive-log acceptance cell proving the emitted launch URL is not exposed.

davidfarah2003:
REJECT @ cb1dcdb25ef980a1fecb64e72a4c7f4fe0196cd2

Blocking: the request-log fix redacts only decoded query keys exactly equal to lowercase `k`; it does not prevent a live launch token from entering the persistent log under another query key. I reproduced this through the real process and malformed real route by changing the new cell's request to `/api/activity?note=<percent-encoded-live-token>&limit=bad`. The named assertion failed, and the emitted diagnostic contained the complete encoded token:

`~ GET /api/activity?note=%45%4c...&limit=bad refused: ...`

The same survives under mixed-case `K`. Repeated lowercase `k` and encoded `%6b` are covered because `URLSearchParams` decodes and `set()` removes duplicates, but key-specific redaction is not a guarantee that the launch credential never enters logs. Redact query values wholesale for request diagnostics, or redact occurrences of the actual process launch token regardless of parameter name, and add the adversarial non-`k` cell.

The detached-log half clears focused review: `openDetachedLog()` opens then synchronously `fchmod`s/hardens before spawn, closes on failure, and no child can inherit or write through the descriptor before that. The normal detached launcher always sets `COTAL_WEB_DETACHED_LOG`; the real-child cell confirms startup output remains while the launch URL is absent. The stale-log cell genuinely pre-creates a 0644 file before calling the shipped opener.

Ran: confirmed and re-pinned PR/worktree head; `pnpm build`; `pnpm smoke:web-console-auth` (139 passed); direct variant of the new real-process leakage cell with the live token under `note` (failed on the named assertion as expected); focused source/delta/env/ordering inspection. Did not run live or stack-starting suites, full CI, browser automation, or prohibited commands.

davidfarah2003:
REJECT @ c48c33a2ab044683532f1d0101f925035aec7a07

Blocking: the value-based fix only examines complete query values. The live launch token still enters the persistent diagnostic log when placed in the request path. I reproduced this through the real process by changing the author's malformed-route cell to request:

`/api/channels/<live-launch-token>/history?limit=bad`

The named no-leak assertion failed, and the real diagnostic contained the complete raw token in the logged path. `requestTargetForLog()` receives the live token but compares only `URLSearchParams` values; `target.slice(0, q)` is copied verbatim. This remains a secret-leak blocking class.

The suite also overstates its invariant for split query values. A real-process variant carrying the token halves under `a` and `b` completed 141/141 while the diagnostic retained both halves, because no individual value equals the token and the assertion only searches for the contiguous token. That is weaker than the path leak but confirms the cell does not prove “never logs the launch token” across the whole target.

The final query-value delta itself works for complete values under encoded `k`, `note`, and `K`; preserves `limit=bad`; and uses the existing timing-safe comparison. The `name === "k" || ...` short-circuit does not reveal token content: key `k` is unconditionally classified secret, while every other value reaches `secretEquals`. Attacker-controlled lengths add linear work but request-target size is already bounded by Node's HTTP parser, and this review found no forge/replay implication.

Residual: the security helper docstring still describes only encoded spellings of key `k`, not the value-based invariant.

Ran: confirmed/re-pinned head; `pnpm build`; official `pnpm smoke:web-console-auth` (141 passed); real-process path-token variant (failed the named assertion with raw token in diagnostic); real-process split-token variant (141 passed despite both halves remaining logged); focused delta/timing/test inspection. Did not run prohibited live/stack commands, full CI, or browser automation.

davidfarah2003:
APPROVE @ b78a5da94ffe1badac6d6f1d330fd6ed9948df8e

The four closed-scope requirements hold:

- Whole-target contiguous redaction runs before query parsing and covers path/query tokens in raw, percent-encoded, or mixed raw/encoded form. `k` remains independently belted. The real malformed-route diagnostics preserve `limit=bad` while removing the credential.
- The helper docstring now states the contiguous whole-target guarantee and honestly distinguishes diagnostic substring scanning from the constant-time comparison at the authentication boundary.
- The query cell is renamed to the guarantee it actually proves.
- The new real-process path cell poses the previously missing case and passes; my additional alternating raw/percent-encoded path variant also passed.

The case-insensitive raw branch can over-redact a non-secret case variant. That is a small diagnostic-fidelity residual which errs safely and does not block. The secret-derived regex has a fixed linear sequence of 43 character alternatives, no quantified/nested ambiguous groups, so I found no practical ReDoS shape.

Ran: confirmed/re-pinned PR head; inspected the two-file delta; `pnpm build`; official `pnpm smoke:web-console-auth` (142 passed); real-process mixed raw/encoded path variant (142 passed). Did not run prohibited live/stack commands, full CI, or browser automation.

PR #458, opened 2026-08-16
spec(11): roll a non-additive discovery change out caller-first

davidfarah2003:
**Companion to #453, and stacked on it.** #453 (§13.7) adds `effect` and `error.outcome` and moves the discovery protocol to `protocol.v: 2`. This adds the deployment rule that makes that move safe to perform. #453 names this amendment normatively as required, so neither reads complete alone; they should land together or adjacent, in this order.

Split out on a ruling rather than folded into #453: a change-process rule addressed to **deployers** is a different normative object from a runtime rule addressed to **callers**, with a different audience and a different blast radius, and #453 was already at its third review round.

> **Why this PR's base is `spec/effect-and-outcome` and not `main`** — it is deliberate, not a mistake. Two reasons. First, based on `main` this branch would conflict with #453 on the Appendix D table head and the snapshot date every time either moves, and resolving a change-log table by taking one side silently drops a whole revision row. Second, and more important: `spec/discovery-rollout-order` contains #453's commits, so a PR based on `main` would show them in its diff and **merging this one alone would sweep #453 into `main` without its own review closing**. Targeting #453's branch makes the stack enforced rather than advisory — the diff below is this amendment only (+20/−2), and GitHub retargets this PR to `main` automatically once #453 merges.

## The gap this closes

§13.7 can require that a responder not emit `v:2` into a deployment whose callers do not all implement it. **It cannot make that achievable**, and the §11 bullet immediately above this one is why: v0 has no in-band capability negotiation, and no request carries a caller version. `describeEndpoint` (`packages/core/src/endpoint-invoke.ts:129`) casts the reply's `data` straight to `DescribeAnswer` without validating it, and that shape carries no `protocol` field at all — so an unamended caller resolves whatever it is handed, and a responder has no way to tell it apart from an amended one.

The obligation therefore rests on the deployment, which is exactly where §11 already puts version agreement.

## What it says

A non-additive discovery change is an out-of-band deployment cutover and rolls out **caller-first**: every caller implements the new rules before any responder registers or describes at the new version. The two halves **SHOULD ship in separate releases**, and shipping them in one release does not make a deployment safe — **a release is not a deployment**, and an already-running caller is unchanged by whatever a new artifact contains.

The observable marker is `protocol.v` on the registered service record (§13.7): "has any responder cut over" is a checkable registry property, while "has every caller adopted" is the out-of-band agreement §11 already requires.

## A withdrawn rule, recorded rather than quietly replaced

An earlier draft of #453 carried this as a §13.7 *ordering* clause — add the caller check before widening the pinned output constant, in the same release, and separate releases are non-conforming. That is withdrawn, not relocated. It was unenforceable in principle for the reason above, and it was self-contradicting: it permitted a release that "already carries" the check while declaring separate releases non-conforming, which are the same thing a beat apart — so it forbade the one rollout that actually works. Review caught all of that; the record is in #453's body.

## What protects callers today, measured

The fence that currently makes this cutover impossible to perform accidentally is on the **responder** side, and that was measured rather than assumed. Mutating `deriveDescriptor` to publish `protocol: { v: 2 }` and running `pnpm smoke:ep-serve` went **127 passed / 8 failed**, every failure reading `/descriptor/protocol/v must be equal to constant` — `describe`'s pinned output schema (`packages/core/src/endpoint-serve.ts:559`) refuses to let a conforming responder publish a v:2 descriptor. Baseline 135/135 green; mutation reverted. The registry read path fails closed the same way, refusing a service record whose `protocol.v` is not `1` (`packages/core/src/endpoint-service.ts:130`).

Both fences are on the responder side. Neither is in the caller. That asymmetry is the whole reason this rule has to bind the deployment.

## Residual

A deployment that cuts a responder over early exposes its unamended callers to exactly what the new version prevents, and within v0 nothing in band detects it. That is stated in the text rather than engineered around — closing it needs negotiation v0 does not have, and the v1 marker in the same section owns it. A rule that read as a guarantee here would repeat the mistake the withdrawn clause made.

## Blast radius

SPEC §11 prose, one Appendix D row, the snapshot date, and the regenerated offline docs bundle. **No schema, no wire field, no code** — the rule binds deployers and the release process, not a running participant, which is precisely why it does not belong in §13.7.


davidfarah2003:
**This PR did not land on `main`. GitHub marked it merged in error, and the amendment is now #460.**

It was opened with base `spec/effect-and-outcome` to enforce a stack. Review then showed the dependency ran the other way — the §13.7 amendment declares itself incomplete without this rule, so landing it first would have put an admittedly incomplete normative document on `main`. The fix was to invert the stack: this rule goes to `main` first, and §13.7 rebases on top of it.

Rebasing `spec/effect-and-outcome` onto this branch made this PR's head an ancestor of its own base, which GitHub reads as merged and closes. Nothing reached `main`.

The amendment is open as **#460** (same commit, `cacb8335`, base `main`), with **#453** stacked on it. Merge order is #460, then #453.

PR #462, opened 2026-08-16
test(herdr): the soak's positive control waits for the payload instead of sampling once

### LINKED ISSUE #459: test(herdr): the soak's positive control samples ps once, right after spawn returns
## What happened

CI run https://github.com/Cotal-AI/Cotal/actions/runs/31926505707 (live job, `pnpm smoke:herdr:soak`) went red on the first cell of the soak, on a branch that touches no herdr file:

```
✗ FAIL: positive control: the survivor instrument can see a LIVE agent — 0 seen
```

Every other cell in the run was green (25 cycles, no panes/workspaces/tabs/launcher dirs/scratch accumulated, no agent processes survive, server healthy). The same job was green on the previous commit of that branch (9b64f176, run 31924891360) and on main (0e4cffb0).

## Why it is a finding about the suite, not noise

That cell is the one that proves the survivor instrument (`ourProcs()`: `ps -eo args=` filtered by `sleep <nonce>`) can see anything at all. If it can flake red while an agent is running, then "no agent processes survive" is green with exactly the probability that the instrument was looking at the right moment; the control was added in f6dc8a0d precisely so that green would mean something.

## Diagnosis

`extensions/herdr/soak.ts:123-131`: the control runs `ourProcs()` immediately after `runtime.spawn(...)` returns with `handle.status() === "running"`. `HerdrRuntime.spawn` returns once `herdr agent start` has a pane running `node <launcher>`; the launcher then loads, deletes its private dir, and only then `spawn`s the payload `sleep <nonce>`. By design the nonce never appears on the launcher's own command line (the env and args ride the 0600 script), so nothing matches `ps` until the CHILD exists. A spawn returning is not the process being visible: a single sample there is a race by construction, and on a loaded runner it loses.

## Proposed fix (not in any open PR; out of scope for the branch that hit it)

Poll `ourProcs() >= 1` until visible with a bounded deadline (a few seconds, well under the cycle timing the soak already asserts), and fail the cell explicitly on expiry with the last count and the elapsed time. Never a single sample. The negative direction (`noSurvivors` after the churn) is already a settled state and can stay a single read.

davidfarah2003:
Fixes #459.

The soak's positive control is the cell the other eleven depend on: it proves the survivor instrument can see a live agent, so that "no agent processes survive" means something. It sampled `ps` once, immediately after `runtime.spawn` returned `running`.

## What the measurement showed

At that moment the pane is running the **launcher**, which has not yet spawned the payload — and the nonce is on the payload's command line only, by design, since the args and env ride an owner-only script precisely to keep them off every command line.

Driving the real `privateLauncher` with its exact argv and sampling at the instant spawn returns:

```
payload visible AT SPAWN-RETURN: 0/6      <- what the control samples
became visible later:            6/6
visibility delay ms:             min 70, max 155   [84, 134, 70, 155, 116, 142]
```

Which forces the question the issue does not ask: **why does CI usually pass?** Because the cell samples after `herdr agent start` returns, and that round trip is itself slow enough to usually cover the launcher's 70–155 ms. The green is bought by the latency of the very call the test is trying not to depend on. That margin belongs to an unrelated subsystem, nobody chose it, nothing documents it, and nothing fails when it shrinks. Under load it shrinks.

So this is not an ordinary race, and that is why the obvious repair is wrong: **a longer wait keeps the shape and merely buys a bigger borrowed margin.** The control has to become a statement about whether the instrument *can* see, not about when it happened to look.

**Scope of that measurement, stated plainly:** mechanism reproduced with the real launcher; the pane is not exercised; `herdr` was not available locally. Omitting the pane is exactly what makes the repro read 0/6 rather than intermittent — removing the accidental cover is what exposes the defect. These numbers establish the **cause** and say nothing about the **rate**.

## The fix

`waitUntilVisible(probe, {deadlineMs, intervalMs})` waits on the named event — the payload becoming visible to the instrument, which is the very property the control asserts. The deadline is a failure outcome, never a warning and never a quiet pass.

Reaching the deadline and being unable to sample at all are reported as **different outcomes**. A `ps` that cannot run says nothing about whether the process exists, and reporting it as "saw nothing" sends whoever reads the red hunting a process that was never looked for. A throw that later reads successfully is tolerated, since transient fork failure under load is the very condition that produced this.

Also: `ourProcs()` appeared twice on one line — once to decide, once to report — so a red could print a count from a later sweep than the one that failed, and during a race those two legitimately disagree. Both failure messages now bind the deciding sample. Same defect at reporting scale.

## Second finding, independent: the nonce could match another run's payload

Separated into its own cells so it can be judged apart. `12${process.pid % 1000}` scoped the count to this run only by accident, and fails two ways — the second far likelier than the first:

- **Birthday:** 1000 distinct values, so two concurrent soaks on one host can share one.
- **Prefix:** the tag is matched with `includes` and the values are variable-width. `pid % 1000 === 1` tags `sleep 121`; `pid % 1000 === 10` tags `sleep 1210`; and `"sleep 1210".includes("sleep 121")` is true. One soak counts another's payloads with **no value collision at all**.

The damage is two-directional inside a single run: a neighbour's live payload passes this run's positive control, and a neighbour's surviving payload fails its leak check. A green bought by another process, and a red that belongs to one.

Fixed-width with 64 bits of entropy in a zero-padded fractional part. Value-uniqueness alone would not have sufficed — the old form already had 1000 distinct values and still let one tag match another's command line.

A unique string that `sleep` *rejects* would be strictly worse than the collision, so that is proven by running it rather than assumed: cell 7 spawns the payload, confirms it is alive, and confirms the full nonce reached argv.

## Does this make past greens trustworthy?

**No.** The fix makes future greens meaningful; it does not make past greens trustworthy. Every soak green behind a run where the control could flake is exactly as weak as it always was.

## Evidence

New suite `pnpm smoke:herdr:probe` — 31 cells, deterministic (injected probe, fake clock), 5/5 consecutive clean runs. A race has no deterministic red through the live path; through an injected probe it does.

Cells 5 and 8 assert the **seam**: cells 1–4 and 6–7 grade the extracted functions, and none of them would fail if the soak kept its old inline sample and the new code sat unused beside it.

| mutation | reddened assertion | marks (baseline 31) |
| --- | --- | --- |
| wait reverts to a single sample | `it is SEEN, not missed` | 16 / 21¹ |
| collapse `unavailable` into `deadline` | `it is UNAVAILABLE, not deadline` | 18 / 21¹ |
| reinstate the bare sample at the call site | `no longer decides that control on a bare sample` | 19 / 21¹ |
| revert to the pid-derived nonce | `200 nonces are 200 distinct values` | 29 |
| drop the fixed width only | `every tag is the same length` | 30 |

¹ run against the 21-cell baseline, before the nonce cells were added.

Kill sets were predicted by name and count before running. Four of five matched exactly. The third predicted 20 marks and produced 19 — **more kills than predicted, not fewer**, since a mark is a *surviving* assertion and an extra red subtracts one. Re-enumerated against the current tree rather than reasoned about: `29 passed, 2 failed`, both reds in cell 5. The prediction missed one because I counted the seam assertions naming `waitUntilVisible` and not the one naming `visibilityDetail`; both key on the statement the mutation replaces, so both correctly notice the seam was cut.

Two rules came out of that, and the second is the one that bit: a mutation report has two independent halves, the count and the account, and when they disagree the finding is that you do not yet know what happened. And **a count quoted without its baseline is not a fact about anything** — the same mutation reads 19/21 and 29/31 across this PR's history, one result wearing two numbers because the nonce cells landed in between. Every count above is quoted `n / baseline`.

`extensions/herdr/probe.ts` is a sibling of `soak.ts`, not under `src/`, and the package ships `dist` only — verified it does not reach the published artifact.

**Gating.** `pnpm smoke:herdr:probe` runs in the live job immediately before `pnpm smoke:herdr:soak`. It was ungated when first pushed and `smoke:gate-inventory` refused the branch for it — correctly: a suite nothing runs is not partial coverage, it is zero coverage with a maintenance cost. Where a suite is gated is an argument, not a detail: this one grades the soak's positive control and its nonce, so if either is wrong the soak greens regardless, and a deterministic sub-second suite belongs in front of the expensive one it can invalidate. It is deliberately not in `smoke:ci` — that chain is the protocol/security gate and it also shards onto Windows, where the cell proving the nonce is a duration `sleep` accepts must really run `sleep`, `ps` and `pkill`. A `win32` skip has precedent here and would have passed; siting a suite where its dependencies are real beats a skip that reads green. That fix (`744956c7`) is **7 lines of YAML and no source change**, so every mutation result and baseline above stands exactly as reported — there is nothing here to re-run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,13 @@ jobs:
         run: pnpm smoke:lifecycle-e2e
       - name: Runtime smoke tests (tmux, orca, herdr)
         run: pnpm smoke:runtime
+      # Runs BEFORE the soak because it grades the soak's own instrument: the wait behind the
+      # positive control, and the nonce the survivor count matches on. If either is wrong the soak
+      # still goes green — a control that can pass while the payload is absent, or a count that
+      # includes a neighbouring run's processes. Deterministic (fake clock, injected probe), so it
+      # is a cheap gate on the expensive one, not a second copy of it.
+      - name: herdr soak instrument (visibility wait + nonce)
+        run: pnpm smoke:herdr:probe
       # Accumulation is invisible to a single-shot suite: a per-cycle leak of a pane, workspace,
       # launcher dir or process only shows up over many spawn/stop cycles, which is what a
       # long-lived manager actually does.
--- a/extensions/herdr/probe.smoke.ts
+++ b/extensions/herdr/probe.smoke.ts
@@ -1,0 +1,151 @@
+/**
+ * THE SOAK'S POSITIVE CONTROL, GRADED WITHOUT A RACE (#459).
+ *
+ * The control is the cell the other eleven depend on: it proves the survivor instrument can see a
+ * live agent, so that "no agent processes survive" means something. It could pass or fail on
+ * timing, which made every green behind it weaker evidence than it looked.
+ *
+ * The fix is a wait on a named event; this suite is what makes that wait provable. The live soak
+ * cannot grade it — a race has no deterministic red through the real path, and reaching for one
+ * would just be the flake in a different costume. So the probe is injected and the clock is fake,
+ * and every branch is exercised on purpose.
+ *
+ * WHAT THIS DOES AND DOES NOT COVER, stated so it is not over-read. It grades `waitUntilVisible`
+ * and the fact that the soak CALLS it. It does not spawn a herdr agent, does not exercise the
+ * pane, and does not prove `ourProcs()` matches a real process — that is the live soak's job and
+ * it needs a herdr binary. Cell 5 is a structural assertion for exactly that reason, and it is
+ * explained where it sits.
+ *
+ * Run: pnpm smoke:herdr:probe
+ */
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { waitUntilVisible, visibilityDetail, type VisibilityOutcome } from "./probe.js";
+
+let pass = 0, fail = 0;
+const check = (name: string, cond: boolean, extra?: unknown) => {
+  if (cond) { pass++; console.log(`  ✓ ${name}`); }
+  else { fail++; console.log(`  ✗ FAIL: ${name}`, extra !== undefined ? JSON.stringify(extra) : ""); }
+};
+
+/** A fake clock. Time moves ONLY when the wait sleeps, so these cells are instant and identical on
+ *  every machine — the whole point, given what is being fixed. */
+function fakeClock(): { now: () => number; sleep: (ms: number) => Promise<void>; elapsed: () => number } {
+  let t = 0;
+  return { now: () => t, sleep: async (ms: number) => { t += ms; }, elapsed: () => t };
+}
+
+/** A probe that returns 0 for the first `n` calls and then 1 — the real shape: the payload does not
+ *  exist yet, then it does. */
+const visibleAfter = (n: number): (() => number) => {
+  let calls = 0;
+  return () => (++calls > n ? 1 : 0);
+};
+
+console.log("\n── herdr positive-control probe ────────────────\n");
+
+// ---- 1. THE DEFECT: A SINGLE SAMPLE IS NOT AN ANSWER -----------------------------------------
+// Measured against the real launcher, the payload was visible 0 times out of 6 at spawn-return and
+// appeared 70-155ms later. So the case that matters is "not yet, then yes" — and the ONLY thing
+// that distinguishes a wait from the old sample is that this case comes back `seen`.
+console.log("1. a process that becomes visible LATE is still seen");
+{
+  const c = fakeClock();
+  const out = await waitUntilVisible(visibleAfter(3), { deadlineMs: 5_000, intervalMs: 50, now: c.now, sleep: c.sleep });
+  check("it is SEEN, not missed", out.kind === "seen", out);
+  check("...on the try that actually saw it, not the first", out.kind === "seen" && out.tries === 4, out);
+  check("...and it reports when, so a slow appearance is visible in the log rather than silent",
+    out.kind === "seen" && out.elapsedMs === 150, out);
+  // The old implementation, exactly: one sample. Stated as a cell so the regression has a name.
+  const single = visibleAfter(3);
+  check("...whereas a SINGLE sample at that instant reports absent — the defect, in one line",
+    single() === 0);
+}
+
+// ---- 2. IT CANNOT PASS WHILE THE PAYLOAD IS ABSENT --------------------------------------------
+// The first of the two directions required of this fix. A wait that gave up quietly, or returned
+// something a caller could read as success, would be the widened window in a different costume.
+console.log("\n2. an absent payload can NEVER be reported as seen");
+{
+  const c = fakeClock();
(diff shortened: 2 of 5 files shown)


davidfarah2003:
## Reconciliation of the third mutation's count

The table records the third mutation as `19 / 21`, against a prediction of `20 / 21`. Re-enumerated on the current tree, where the baseline is 31 rather than 21:

```
unmutated                       PASS — 31 passed, 0 failed
reinstate the bare sample       FAIL — 29 passed, 2 failed
  ✗ ...and no longer decides that control on a bare sample
  ✗ ...and the control's evidence line is the outcome's, not a second measurement
```

**Direction: more kills than predicted, not fewer.** I predicted one reddened assertion and got two, so the surviving-mark count came in one *below* the prediction — 19 where I expected 20 at the 21-cell baseline, and 29 where 30 would correspond at the 31-cell baseline. The two numbers are the same result measured against two different baselines; nothing about the mutation changed between them.

Both reds are the intended kind. The mutation replaces the whole control statement, and that statement is referenced by two separate seam assertions in cell 5 — one checking that the control is no longer decided by a bare sample, one checking that its evidence line is the outcome's rather than a second measurement. Neither is incidental and neither is a crash: both are assertions about the seam, and both correctly notice that the seam was cut. What was wrong was my prediction, which counted the assertions that mention `waitUntilVisible` and missed the one that mentions `visibilityDetail`.

Stated as the general point, since it is the reason this paragraph exists: a mutation report has two independent halves, the count and the account, and when they disagree the finding is that you do not yet know what happened. The count is authoritative; the account has to be re-derived from an enumeration rather than reasoned backwards from the number.

davidfarah2003:
### CI red on `c646fd5a`, and what it was

`smoke (shard 1/4)` failed at `pnpm smoke:gate-inventory`:

```
✗ FAIL: 1 suite(s) exist but nothing runs them, and they are not in UNGATED:
    smoke:herdr:probe
  Gate it in smoke:ci, or add it to UNGATED with the reason it is excluded.
```

Mine, and worth naming rather than quietly fixing. This PR adds a suite whose whole subject is
an instrument that reported a number without observing what the number claimed to be about —
and I left that suite reachable only by hand. A suite nothing runs is the same defect one layer
up. The repo already had a gate for it.

**Fix (`744956c7`)** — seven lines of YAML, no source change: `pnpm smoke:herdr:probe` runs in
the live job immediately **before** `pnpm smoke:herdr:soak`.

The placement is the argument. This suite grades the soak's positive control and the nonce the
survivor count matches on; if either is wrong the soak goes green regardless. A deterministic
sub-second suite gating an expensive one belongs in front of it, not beside it.

Not `smoke:ci`, for two reasons. That chain is the protocol/security gate and this is neither.
And it is sharded onto Windows, where the cell that proves the nonce is a duration `sleep`
accepts must actually run `sleep`, `ps` and `pkill`. There is precedent in this repo for a
`process.platform !== "win32"` guard with a printed skip line, and it would have passed CI — but
siting a suite where its dependencies are real is better than a skip that reads green.

No new `UNGATED` entry either. That list's own header says adding a line is a decision, not a
fix, and that a suite proving shipped behaviour should be gated. This one does.

After: `gate inventory: 297 smoke scripts, 261 reached, 36 not run by anything — OK`, and the
probe suite still 31/31. The diff touches no source, so the five mutation proofs and their
baselines stand exactly as reported above.

PR #468, opened 2026-08-16
fix(core): let a class scatter stop waiting on an instance the broker says is gone

davidfarah2003:
A class scatter freezes its expected set from the service registry, which records **registration,
not liveness**. An instance that crashes never deregisters, so its `svc….status` stays `ready`
indefinitely. The gather has exactly two exits (every frozen slot answered, or the deadline), and a
slot that can never answer makes the first unreachable. So the deadline is paid in full, on every
scatter in the space, for as long as the dead record survives. `attach` and `stop` route through the
same scatter, so they pay it too.

Measured on a live mesh with three registered managers, at the merge-base:

```
resolveService=1230ms  scatterCommand=8372ms  replies=1 missing=2   cotal ps TOTAL 14.62s
resolveService=1095ms  scatterCommand=8378ms  replies=1 missing=2   cotal ps TOTAL 13.80s
```

`scatterCommand` pays its 8000ms budget to the tens of milliseconds, every run.

## What moves, and what does not

`epScatter` takes an optional `probeLiveness` hook and ends the gather once every frozen slot has
either produced a valid reply or been affirmed gone. It moves the classification point and nothing
else: an affirmed-gone slot is still `missing`, still surfaced, still not `complete`, and a straggler
arriving after an early finish is still reported `late` rather than dropped.

The check runs on **both** events that can make it true: a verdict landing, and a reply landing.
That is not symmetry for its own sake: the last thing to settle is as often the live peer's slow
answer as it is the corpse's verdict, and a version that re-checked only on verdicts was inert in
exactly the shape it was built for (one corpse affirmed immediately, one live instance answering a
second later). That version existed, and is `mutations[0]`.

## Only one thing licenses an early classification

`gone` and nothing else. `live`, `unknown`, a hook that throws, and any value outside the closed set
all leave the full deadline standing. The asymmetry is the safety argument, not a defensive default:
a broken or silent probe degrades to exactly the previous behaviour, never to a fast wrong answer.

The obvious cheaper design, cross-checking the frozen set against presence and giving a
presence-absent instance a short grace, was **measured and rejected**. On the mesh this was built
against, the manager actually running the space both drops out of presence periodically (its liveness
lease times out and it re-registers) and needs about a second to answer:

```
deadline=1000ms → replies=0 missing=3      deadline=1200ms → replies=1 missing=2
```

A grace keyed on a lapsed presence entry would have reported that manager unreachable, *faster* than
the current code produces the right answer. A missing presence key is absence of evidence, and this
repo already says what that is worth: "a timeout is the canonical 'absence of evidence', and reading
it as death is the exact defect this command exists to avoid."

What does count is the broker's own no-responders 503 on an instance's own rail.
`epProbeInstanceInterest` publishes a `describe` **cast** there with the reserved sentinel as its
reply-to. That is the same primitive and the same trust rule `epCall` already relies on, since no responder
holds a publish grant for that subject. A serving incarnation subscribes its instance rail for every
command it serves and every endpoint must serve `describe`, so silence there is evidence of absence.
The reply is never read, so an instance whose describe is broken still reads as present.

## A negative result, kept rather than buried

The CLI wiring that would make this reach `cotal ps` was built and then **removed**, because it was
measured and it made the command *slower*: **13.8s → 17.0s**.

`ps` mints a one-shot instrument before the freeze, so it holds no instance rows and the probe's
publish is refused. Rather than mint a wildcard instance row, a widening this repo's grant tests
assert against by shape, the CLI froze the class first and re-minted against the exact frozen ids.
That costs an extra connection and freeze (a connection's permissions are fixed at authentication, so
the open one cannot gain a rail), about 3.2s against this broker, and it bought nothing:

```
[PROBE] <instance A> -> gone    (123ms)
[PROBE] <instance B> -> unknown (7087ms)
[PROBE] <instance C> -> unknown (7087ms)
```

Of the three registered managers only **one** is a true corpse; the probe finds it in 123ms. Another
answers nothing for 8s while still **holding its subscriptions**. It is hung, not gone, and no
liveness probe may shortcut that, because "connected but slow" and "connected but wedged" are the
same observation. A hung instance still costs the full deadline, and that is the correct result.

So the mechanism is proven and its cost is not yet worth paying on that path. The remaining slowness
there is a hung manager, which is a different defect. Shipping a measured regression to fix a
hypothesis is the wrong way round, and `instancePinnedInstrumentCapabilities` accepting a list of ids
is kept so the wiring is a small step whenever the extra round trip can be made free.

## A cell that passed on the defect

`manager-scatter`'s severed-manager case runs a **3000ms** budget and asserts `elapsed < 8_000`. The
defect yields ~3050ms, so it passed, for as long as it existed. It proves the gather is *bounded*;
its name says "a dead instance does not hang the gather", which reads as *fast*. Kept as-is, and the
claim it was read as making is now asserted next to it.

## Gates

`smoke:scatter-liveness`, 34 cells, wired into `smoke:ci`. Its predictions were written and committed
**before the first run**. One came back wrong and is recorded rather than re-cut into agreement: a
cell claiming "a gone verdict never un-counts a reply already in hand" was in fact grading a race in
which no reply was ever in hand. It is now two cells: one with the verdict delayed (the property
meant), one asserting what the racy form actually grades, namely that a lying `gone` fails toward
**unreachable**, never toward a false success.

Kill set of 5, all killed on their named assertion (2 / 5 / 13 / 2 / 1 red against a 34 baseline). The
third is the trap: make every verdict license the early finish and the slow-but-alive instance becomes
a fast wrong "unreachable".

The `pnpm mutation-proof` caveat applies: that suite hands `epScatter` its hook, so it proves the gather
depends on a verdict, not that any caller produces one. `manager-scatter` proves the entry point
separately: a real `scatterCommand` over a real broker against a manager severed exactly as a crash
severs it.

Also green: `ep-verbs`, `ep-grants`, `inst-route-grant`, `frozen-exports`.


davidfarah2003:
Closing as superseded: main landed this behavior independently — 2752fe79 "fix(ps): stop scattering into dead registrations, and give them an exit" plus follow-up fcb2e871, including a scatter-liveness smoke. The branch here predates the smoke-shard refactor and would regress main's current endpoint-verbs implementation (-36/+8 on the landed code).

PR #469, opened 2026-08-16
test(docs): run the docs smoke, and stop it asserting a fact about the corpus

davidfarah2003:
`pnpm smoke:docs` was red on `main`, and no automated path ran it.

It is in the `check` root only: absent from `smoke:ci`, and absent from every workflow. The
docs-adjacent CI job runs typecheck, build, `check:docsbundle`, test and `smoke:view`, then stops.
So a contributor running `smoke:ci` locally saw green and CI saw green, while the suite sat red.
That is the shape worth naming. Not "local misses what CI runs", not the reverse, but a suite
wired into the repo's own `check` root that nothing executes. `gate-inventory` grades that an entry
resolves and is reachable from a root, and reachability is not execution.

## What was actually red

The cell required the SPEC section documenting `$SYS` to be among the default five hits for the
query `$SYS.>`. Replayed against the generated bundle either side of the commit that added `$SYS`
credential-rotation prose to `docs/cli.md` and `docs/identity-and-auth.md`: before, the spec ranks
4th; after, two new sections appear and it ranks 6th. Nothing about the code or about `$SYS`
changed. A docs edit that broke nothing turned a suite red, and nobody saw it.

The leading explanation on offer, that a very large changelog row in `SPEC.md` diluted the spec
page's term frequencies, is false, and cheap to falsify: dropping both giant rows moves every score
by under 0.06 and changes no ordering at all. Worth stating, because a search-quality regression
and a docs defect have completely different repairs.

So the assertion was a fact about the corpus, pinned as if it were a property of the code. It was
never held up by any mechanism, only by no other page happening to say `$SYS`, and it was going to
break on someone else's docs edit.

## A ranking mechanism was tried and rejected on measurement

Emitting the namespace prefix a wildcard names (`$SYS.>` to `$SYS.`) does rank the spec first,
because exactly one place in the corpus writes `` `$SYS.` ``. It also scores punctuation: every
`mesh.` in the docs, across 14 pages, is the word "mesh" ending a sentence, so `mesh.*` traded its
useful hits for the pages with the most sentence-final periods. Reverted. There is no signal here
to rank on, so the cells now assert what the code guarantees, which is that the wildcard-stripped
base is what reaches the `$SYS` sections at all and that the spec is among them, and they stop
claiming rank.

## Changes

* **The suite runs.** Added to the `unit` job, after `Build`, because it reads the built connector
  through a package specifier. It is broker-free and network-free.
* **The suite is legible.** It reported one word for some thirty assertions; each of its five
  sections now prints a mark. That is also what makes it gradable: a failure inside the search
  section stopped before the terminal line, so `mutation-proof` could not distinguish a real
  failure from a run that died at the first cell. It graded both mutations INCONCLUSIVE until the
  marker moved upstream of the region under test.
* **The guard is proved.** `extensions/connector-core/smoke/mutations/docs-wildcard-base.json`, 4
  of 4 killed. The command carries the package build, because the suite loads `dist` and a change
  made in `src` would otherwise be void on arrival, grading module resolution rather than the
  guard.

No package behaviour changes, so no changeset.

## The first repair was the same defect with more slack, and review caught it

Replacing "the spec is in the default five" with "the spec is in the first ten" reads as
reachability and is not. The spec ranks sixth, so the cell was carrying four spare slots against one
afternoon's corpus: a docs wave adding four higher-scoring `$SYS` sections would have reddened it
exactly the way the top-five version did, and any hardcoded cut of six or more left every cell
green. The mutation aimed at it was killing on the corpus rather than on the code, because five
happens to sit below the spec's current rank. A corpus fact pinned as a code property does not stop
being one because the window got wider.

The reachability claim is now unwindowed and asserts no rank at all, which is the claim the code is
responsible for: the wildcard-stripped base is what makes the section reachable. The limit argument
is graded separately against a count, which nothing in the corpus can move. The two are a pair
rather than a duplicate: an unwindowed query cannot see a limit that is never applied, and a count
cannot see a base subject that is never indexed.

## Reported, not fixed

Disabling the dotted-segment expansion outright leaves the suite green. The cell named "a dotted
path matches on its segments" only asserts `searchDocs("cotal.schema.json").length > 0`, which the
raw whole-token match satisfies on its own, so it does not test segment matching. Out of scope
here.


davidfarah2003:
Independent corroboration, recorded here so the next reader does not re-derive it.

**The bisect is confirmed from the opposite end.** A direct `searchDocs` measurement puts `cli` and
`identity-and-auth` in the top five for `$SYS`, which are the same two pages, by name, that the
bisect identifies as gaining `$SYS` prose in the credential-rotation commit. One approach found the
commit and read off the pages; the other found the pages without knowing the commit. They agree.

**Three separate working trees, one failure.** The red survives swapping in `main`'s
`docs-bundle.generated.ts` alongside `main`'s `docs/control-surface.md`, and separately alongside
`main`'s `SPEC.md`. It does not belong to any one branch.

**The dilution explanation is falsified twice, by different methods.** Truncating the roughly
19,000-character changelog row to 88 characters and regenerating the bundle reproduces the identical
assertion; dropping both giant rows moves every score by under 0.06 and changes no ordering.
Recorded because a plausible untested cause is exactly the kind that sticks.

**And two measurements that bound the defect more tightly than the cell's own name did:**

* `$SYS.>` and `$SYS` return byte-identical results, and `team.>` passes, so wildcard resolution is
  not broken. The defect is narrower than "surfaces the spec section".
* `$SYS.REQ.SERVER.PING.VARZ` ranks the spec page first. Not a missing page, not a broken index,
  not tokenization. It is purely rank on a single-token query, which is precisely the property a
  docs edit can move, and precisely what a test should not have been pinning.

That is the case for asserting reachability rather than rank, made from evidence this change did
not have when it was opened.

davidfarah2003:
**Superseded by the comment below.** The occurrence counts in this table were line counts, and the
corrected figures with the method stated are in the final comment. Kept so the correction has
something to point at. The conclusion did not change.

The original text asserted that specificity was the discriminator, based on `endpoint` and
`lifecycleUid`. The final comment shows the actual account, which is ordinary term frequency.

davidfarah2003:
**Final account, and it is stronger than "an accident of the corpus".**

*Corrected: an earlier comment carried line counts labelled as occurrences. These are
case-insensitive substring occurrences, re-measured directly from the generated bundle. The
ordering and the conclusion are unchanged.*

| term | occurrences in SPEC.md | docs pages containing it | densest docs page | ranks first |
|---|---|---|---|---|
| `endpoint` | 333 | 25 / 36 | `cli` = 28 | spec |
| `credential` | 232 | 22 / 36 | `cli` = 26 | spec |
| `lifecycleUid` | 70 | 5 / 36 | `build-a-client` = 2 | spec |
| `presence` | 37 | 29 / 36 | `architecture` = 16 | `presence-and-delivery` |
| `anycast` | 18 | 18 / 36 | `build-a-client` = 7 | `presence-and-delivery` |
| `mappingRevision` | 13 | 0 / 36 | none | spec |
| `$SYS` | 6 | 4 / 36 | `cli` = 8 | `cli` |

`endpoint` is on 25 of 36 pages, so its IDF is near zero, and the spec still wins. That rules out
specificity as the discriminator. The rare-term rows rule out length normalization: the spec is the
largest document in the bundle and wins outright on every term it actually discusses. What is left
is ordinary BM25.

**`$SYS` is the only row where a docs page says the term more often than `SPEC.md` does.**
`docs/cli.md` says it eight times against six, on a page a fortieth of the size. So the spec is not
being penalized. It is losing a query it deserves to lose, and the test was asserting that the
largest document would outrank every other page on a token it barely discusses. That was wrong when
it was written and passed only because the competition did not exist yet.

Three proposed mechanisms are now dead: the giant `SPEC.md` changelog row, falsified two ways; BM25
length normalization, falsified by the rare-term rows; and a namespace-prefix scoring term tried in
this branch, which ranked the spec first and scored punctuation, because every `mesh.` in the docs
is the word "mesh" ending a sentence. The third is the instructive one. It passed the query it was
written for and failed only on queries nobody was fixing, which is how a wrong mechanism earns
enough credibility to reach a pull request. Diffing 36 queries instead of re-running the one under
repair is the check that caught it.

PR #491, opened 2026-08-17
fix(manager): the forced-split probe accepts both correct fence endings

davidfarah2003:
The forced arm of `describe-split-duplicate-effect` asserted `forcedEffects === 1 && forcedOk`, which accepts only the ending where the repair healed. Core repairs a bind refusal exactly once and surfaces a second one, so the re-issue meets the same contended queue and splits again at roughly the rate the probe prints in its own section 3. The cell was a coin flip on live behaviour, and it graded the documented re-split as a failure.

That is not hypothetical. It reddened `smoke (shard 3/4)` on #489, which touches only the browser dashboard. On the failing run the forced arm behaved correctly at every step (`refusals repaired: 1`, `effects: 0`, `caller saw ok: false`, refusal carrying the marker and `outcome=not-executed`) and was still graded FAIL, while section 3 of the same run printed `re-issue attempted: 7, HEALED: 4, split AGAIN and surfaced: 3`. A gate that reddens unrelated pull requests at that rate is worse than no gate: it gets ignored, then reverted, and it takes the cells that do work with it.

What the fence promises is not that the repair succeeds. It is that the caller is never lied to. This asserts that instead: the write landed exactly once and was reported so, or it landed nowhere and the caller was told conclusively, with the marker and an explicit `not-executed` outcome read exactly as the trial loop already reads it. A duplicate effect, or a silent failure, satisfies neither arm. The fence cell above it is unchanged, and it is what still catches the fence being removed.

Verification. Eight local runs pass, two of them on the surfaced ending the previous cell rejected, so the case this widens for is observed rather than argued. Because widening an assertion is the direction that can silently accept everything, the loosened cell was mutation-proved: breaking `define-persona` so it reports success without writing produces `effects: 0` with `ok: true`, which neither arm covers. Predicted before the run, and KILLED on the named cell, baseline green.

### DIFF
--- a/implementations/manager/smoke/describe-split-duplicate-effect.smoke.ts
+++ b/implementations/manager/smoke/describe-split-duplicate-effect.smoke.ts
@@ -191,7 +191,7 @@ try {
   const bound = cache.get(MANAGER_ENDPOINT);
   check("the caller holds a resolved handle to rewrite (the forced case can be set up at all)",
     bound !== undefined, { cached: [...cache.keys()] });
-  let forcedRecoveries = 0, forcedEffects = 0, forcedOk = false;
+  let forcedRecoveries = 0, forcedEffects = 0, forcedOk = false, forcedConclusive = false;
   if (bound !== undefined) {
     bound.responder.instanceId = mintLifecycleUid();
     const forcedName = `epsplit-forced-${randomUUID().slice(0, 6)}`;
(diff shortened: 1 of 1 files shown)


davidfarah2003:
Closing as a duplicate: #487 already carries this change, with the same disjunction and the same stated invariant, and it is further along. It also fixes the swallowed `catch` and adds the arming control, neither of which is here.

The evidence gathered here transfers and is recorded on #487 rather than lost: eight local runs with two on the surfaced ending, and a mutation proof that the widened cell still has teeth.

PR #528, opened 2026-08-17
feat(core)!: an agent reads only the channels it lists

davidfarah2003:
## What changes

An agent now reads only the channels it lists. An omitted or empty read set means **no channels**, where it previously meant `general`.

Four places defaulted an absent read set to `["general"]`, independently of each other: the agent-file loader, the provisioner, the credential mint, and the endpoint. A persona that simply did not mention channels was therefore subscribed to `general` by code, and had the matching channel read row baked into its credential. Every DM-only reviewer and every probe seat landed on a channel nobody chose.

An agent on no channel is still a full mesh peer. It appears on the roster and it sends and receives DMs and anycasts; only the channel rows are gone. The default-deny that already governed `allowPublish` now governs reads as well, and for both an empty list means the same thing as an omitted one: channels denied, everything else intact.

With no concrete channel there is no default broadcast target, so a send with no explicit channel is refused with a message that says so, instead of resolving to `general` (which, for a scoped agent, the broker would deny anyway).

`cotal join` keeps its `general` default. That is a human typing a command whose whole point is to land somewhere and talk, not a persona-driven agent.

## Scope

Behaviour:

- `packages/core/src/agent-file.ts`: an omitted or empty `subscribe` is the empty set. The `subscribe` subset-of `allowSubscribe` invariant is unchanged. Note the direction of the old bug here: a persona with `allowSubscribe: [ops]` and no `subscribe` used to be **rejected at load**, because the invented `general` read set was not within its own ACL. Such a file now loads.
- `packages/core/src/provision.ts`: `provisionAgentDurables` and `permissionsFor` no longer invent a read set. An omitted or empty `allowSubscribe` mints no per-channel history-consumer create grant and no `chat.*.<ch>` sub row. The DM, presence, anycast and control rows are unconditional and unchanged.
- `packages/core/src/endpoint.ts`: `channels` defaults to none, and `multicast` with no explicit channel throws instead of picking `general`.
- `extensions/connector-core/src/config.ts`: the launcher then persona then link chain keeps its order; only its tail default is removed.
- `implementations/manager/src/manager.ts`, `implementations/cli/src/commands/spawn.ts`: the same default removed on the spawn and mint paths.
- `extensions/pi/src/tools.ts`: the pi send-call render carried the same fallback and now renders the absence.

Tool descriptions:

- `cotal_leave` said "you can't leave your only channel". That was false and there is no such guard in `leaveChannel`. It now states that leaving your last channel is allowed, that you stay on the mesh and DM-reachable, and that you have no default send channel until you join one.
- `cotal_send` renders a concrete default when the agent has one, and says the argument is required when it does not.

Docs, in the same change: `docs/agent-files.md`, `docs/channels-and-permissions.md` (the verb table plus a paragraph on the rule and one on leaving your last channel), `docs/config.md`, and the regenerated `docs/mcp-tools.md` and docs bundle. `SPEC.md` states no channel default anywhere and is untouched.

Stale comments describing the old fallback are corrected in `packages/core/src/connector.ts`, `extensions/connector-core/src/launch.ts`, `implementations/manager/src/manager.ts` and `implementations/cli/src/commands/mint.ts`.

## Tests

`packages/core/smoke/no-implicit-general.smoke.ts`, 24 cells against its own JWT-auth broker:

- a persona with no read set joins nothing, and its credential carries no channel row. Proven **at the broker**: a native subscribe on `chat.*.*.general` under those creds is denied. A mint that lists no row and a server that refuses the subscribe are different claims, and only the second one is the boundary.
- a persona that lists `general` still gets it, and still receives `#general` traffic that the channel-less agent does not.
- an agent minted with `allowPublish: []` still sends and receives DMs and is on the roster.
- a send with no channel and no subscription refuses, with the refusal text asserted and asserted not to name `general`.
- leaving your only channel is allowed, the send then refuses, and re-joining restores it.

`extensions/connector-core/smoke/session-channels.smoke.ts`, 7 cells, no broker: the connector's resolution chain and its order, with an empty tail.

Both suites are appended at the tail of `bin/smoke/ci-suites.txt`. Both import their target by relative specifier, so the bytes graded are the source this PR edits rather than a build of it.

`packages/core/smoke/fixtures/no-implicit-general.mutations.json` carries three named mutations, one per default: restoring the loader default, restoring the mint default, and restoring the endpoint fallback. Each names the cell it must redden.

## Migration

List `general` explicitly if you want it. The personas `cotal setup` seeds already do, so a fresh install is unchanged. A hand-written persona that relied on the old default needs `subscribe: [general]` added.


davidfarah2003:
Brought this up to date with `main` and swept it for prose the change invalidated.

**Merge (`630eff81`).** The branch was 943 commits behind. Three conflicts:

- `implementations/cli/src/commands/spawn.ts` — one hunk, and the only resolution here that a
  mechanical ours/theirs would have got wrong in either direction. Both sides had edited the same
  `grantAgent` call: this branch replaced the implicit `["general"]` read default with the boot set,
  and `main` had added an events publish grant derived where the owner first resolves. Taking either
  side whole silently drops the other feature, so the resolution keeps this branch's
  `allowSubscribe` and `main`'s `allowPublish: publish`.
- `bin/smoke/ci-suites.txt` — took `main`'s list verbatim and appended this branch's two suites at
  the end, so `main` stays a byte prefix. That drops `smoke:opencode-transcript` and
  `smoke:transcript-grant`, which this branch still listed: `main` retired both in `993810a9` and
  neither script exists in the merged `package.json`.
- `extensions/connector-core/src/docs-bundle.generated.ts` — regenerated rather than hand-resolved,
  after `pnpm gen:tooldocs` refreshed `docs/mcp-tools.md` through this branch's edited generator.

Gates at the merge, each read off the bare command: `pnpm build` 0, `pnpm typecheck` 0,
`pnpm check:docsbundle` 0, `pnpm smoke:mutation-fixtures` 0, `pnpm changeset status` 0.

**Stale prose (`a4933167`).** `ActorRow.allowSubscribe`'s docblock in
`implementations/auth/src/ledger.ts` told the reader that the spawn paths "fall back to
`["general"]` for the read set" — the exact fallback this branch removes from both files that
docblock names. The file was never part of this change, which is where prose about an invariant
tends to go stale: you do not proofread a file you never open. Comment only, no behaviour, so no
changeset line.

Three other `general` mentions in shipped comments were checked and deliberately left: the
`agent-file.ts` docblock examples list `general` explicitly, which is correct under the new rule;
`join.ts` keeps a `general` landing default for a person typing `cotal join`, with a comment above
it saying that is deliberate and scoped to that command; and `manager.ts:3092` describes the old
default in the past tense as behaviour already removed.

davidfarah2003:
REJECT @ 5f1666d60c84445f5756cc9107f5d8169a022f06

**Blocker: the breaking-change migration note is false for the default fresh-install path.** `.changeset/no-implicit-general.md` says “The personas `cotal setup` seeds already do [list `general`], so a fresh install is unchanged.” At this head, `implementations/cli/src/commands/setup.ts` seeds the common non-demo `default_agent` with `subscribe: []`; at the pre-change base it seeded `subscribe: [general]`. Therefore a fresh minimal setup changes from auto-reading `general` to reading no channel. The implementation is defensible and now matches the existing “isn't firehosed” comment, but the npm-facing migration claim describes the opposite behavior. The generated package changelogs already state the correct behavior (“`default_agent` no longer auto-subscribes to `general`”), which makes the source changeset internally contradictory rather than curing it. For a `!` wire-default change, fix the changeset before release.

**Confirmed, not blockers:**
- Omitted and empty `subscribe` both resolve to `[]`. With both read keys omitted, `allowSubscribe` also resolves to `[]`, so runtime `cotal_join` is denied. Loader, manager spawn, provisioner, ACL registry, and minted credential agree. The changeset does mention that omitted/empty applies to both and that reads are default-deny, though “only the channel rows are gone” understates the runtime-join consequence.
- Explicit `allowSubscribe: ["> "]` typo aside, the actual seeded default uses `[">"]`, so it boots on no channel but may join one. A persona omitting both keys may not join any channel.
- No surviving production implicit-`general` default was found. The remaining `general` in `join` is the explicit landing-channel default for that command, not an agent read default. The apparent survivor in `packages/core/smoke/e2e-acl.smoke.ts` is test helper code.
- No loader/credential divergence on new spawns: the same resolved ACL reaches durable provisioning, the ACL row, and `mintCreds`. Existing running/static credentials are not retroactively narrowed; retained inventory and renewal preserve their previously resolved `launch.allowSubscribe`, so an old seat may keep `general` until respawn. That is an upgrade residual worth documenting, not a new internal divergence.
- The no-default-send refusal is actionable and the tool schema changes dynamically to say the channel is required when none exists. No production caller relying on an omitted multicast target was found.

**Suites run:** `pnpm build`; `pnpm smoke:no-implicit-general` (24 checks); `pnpm smoke:session-channels` (7 checks); mutation proof for `no-implicit-general.mutations.json` (3/3 killed on named cells); mutation proof for `events-grant-acl.json` (8/8 killed on named cells, including G8 manager default). Tree restored clean after mutations. I did not run full `smoke:ci`, any `*-live` suite, or any stack start/stop command.

**Named residuals:** the changeset should explicitly say that omitting both keys removes runtime join authority, and should state that already-running/preserved seats retain previously minted/resolved ACLs until respawn rather than being retroactively narrowed. The reported 24/59 omitted-persona fleet shape is consistent with a material migration blast radius, but is not treated as user-population evidence.

davidfarah2003:
APPROVE @ ff91a7148d1213ccf4dd5a4afc3dbea022dbf39a

Bounded re-review of the one-file delta from `5f1666d60`:

- The false fresh-install claim is removed. The changeset now accurately says the seeded `default_agent` has `subscribe: []`, no longer boots on `general`, and retains `allowSubscribe: [">"]` so runtime join remains available.
- The omitted-both-keys consequence is now explicit and accurate: the read ACL resolves empty too, so `cotal_join` is unavailable unless `allowSubscribe` is declared.
- The upgrade statement is accurate: retained live static seats renew from their recorded `launch.allowSubscribe`, so their previously minted read ACL persists until respawn rather than being narrowed retroactively.
- No new inaccuracy found. No duplicate “fresh install is unchanged” / “seeds already do” claim remains in Markdown.

Only `.changeset/no-implicit-general.md` moved. Per the bounded request, I did not rerun suites or mutation fixtures already cleared at `5f1666d60`.

davidfarah2003:
APPROVE @ 0dd51d03a

Bounded merge-delta re-grade:
- Merge is honest: `ff91a7148..HEAD` changes only the two web smoke files, and both exactly match the merged main parent. All other #528 files are byte-identical to the approved first parent; `bin/smoke/ci-suites.txt` is unchanged.
- No #528/#888 interaction blocker: the shipped web endpoint explicitly uses `channels: []` as a non-consuming observer and connects with the resolved admin credential/view. Its HTTP routes read registry/history directly, so #528’s removal of implicit channel membership does not remove the web read authority. The two fixtures use an unauthenticated local broker, where the new credential ACL defaults do not apply. This conclusion is from the code paths; I did not re-run the settled suites.
- `pnpm typecheck` passed.

No blockers found.

Issue #373, opened 2026-08-13
security: manager spawn contract exposes events/transcript to any spawn-capability caller with no admin-tier check


davidfarah2003:
## Summary

The manager's typed spawn contract admits a caller-controlled `events: boolean`, and `opStart` arms it — mutating the child's publish grant and setting `COTAL_EVENTS` — with no operator/admin-tier check. Any holder of an ordinary spawn capability can therefore cause a spawned agent's harness session output to be published to a channel other principals can read.

This is an **authority** question rather than a behaviour one: it changes *who* can cause host-session disclosure, not what happens under existing authority.

## Path

| Step | Location |
| --- | --- |
| `events: boolean` admitted on the spawn input schema | `implementations/manager/src/manager-service-contract.ts:110-135` |
| ordinary spawn-capability handler reaches `opStart` | `implementations/manager/src/manager.ts:1800-1803` |
| `opStart` copies the bit through | `implementations/manager/src/manager.ts:2495-2545` |
| child grant mutated and `COTAL_EVENTS` armed | `implementations/manager/src/manager.ts:2887-2902` |

No admin-tier signal is consulted before connector resolution or grant mutation.

The broker rollup gives ordinary `spawn` holders untargeted manager creation (`packages/core/src/endpoint-grants.ts:214-227`), so the capability that reaches this path is not itself operator-scoped.

The public `cotal_spawn` MCP tool does **not** expose the option (`extensions/connector-core/src/tool-specs.ts:517-555`), which limits exposure in practice — but a raw typed endpoint caller can submit it, so the tool's omission is a convention rather than an enforcement point.

## Pre-existing, not newly introduced

Worth stating plainly so the remedy is scoped correctly: this shape **pre-exists** under the previous spelling, where the same field was `transcript: boolean` on the same contract, reachable by the same callers, armed at the same place with the same absence of a check. A rename of that field changes the spelling, not the reachability. Independently confirmed by two reviewers.

So this is not a regression introduced by any in-flight rename — it is an inherited property of the contract that a rename makes newly visible.

## Why it is filed rather than carried as an accepted residual

Authority-widening is not the kind of finding a single lane should be able to classify down, because the cost is not paid by whoever makes that call. Filing it keeps it tracked and owned by the surface's maintainer instead of being absorbed as a footnote in an unrelated change.

## Remedy shapes

Both are viable and the choice belongs to the owner of this surface; this issue deliberately recommends neither.

[...]

PR #614, opened 2026-08-18
fix(connectors)!: deliver seat connection material by private file, not ambient env

### LINKED ISSUE #396: agent sessions export live credentials and control tokens into the ambient env of every descendant process
## Summary

A manager-hosted agent session receives its connection material and control-plane access as **environment variables**, and the process environment is inherited by **every descendant process** — gates, `pnpm`/`npm` scripts, test suites, linters, editors, and any tool the session shells out to. Nothing in that chain asked for the credentials, and nothing scopes them back down.

Confirmed independently by four operators on one space, on separate sessions.

## What is exported

Observed in the environment of an ordinary shell command run from an agent session (values redacted; the point is presence, not content):

```
COTAL_SERVERS=nats://broker.cotal.ai:4222     <- the LIVE broker
COTAL_CREDS=<path to the session's lifecycle credential file>
COTAL_SPACE=main
COTAL_CONTROL_SOCKET=/tmp/cotal-<...>.sock
COTAL_CONTROL_TOKEN=<bearer for the control socket>
COTAL_ALLOW_PUBLISH=>
COTAL_ALLOW_SUBSCRIBE=>
COTAL_ID=... COTAL_LIFECYCLE_UID=... COTAL_AGENT_FILE=... COTAL_CAPABILITIES=spawn
```

Two distinct kinds of authority are in there:

1. **Broker identity** — `COTAL_CREDS` plus `COTAL_SERVERS` is everything needed to connect to the live mesh *as that agent*. `COTAL_ALLOW_PUBLISH` / `COTAL_ALLOW_SUBSCRIBE` are exported as `>` here, so the grant being described is not narrow.
2. **Control-plane access** — `COTAL_CONTROL_SOCKET` + `COTAL_CONTROL_TOKEN` reach the session's control surface. That is a different and larger blast radius than "can send a message".

## Why it is worth fixing rather than documenting

The inheritance is **transitive and invisible**. A session that runs a build has just handed the build its mesh identity and control token. A session that runs a third-party CLI has handed it the same. There is no point in that chain where a human sees a credential being passed, so there is no natural moment to object — and the usual instinct ("I didn't pass any secrets to that command") is wrong here in a way that is hard to notice.

It also breaks test and CI isolation in a specific, biting way. `COTAL_SERVERS` points at the **live broker**, so any suite that reads its server from the ambient environment — a perfectly ordinary thing for a smoke to do — silently targets production instead of its own ephemeral fixture. One operator's probe caught exactly this: it refused to start because `COTAL_SERVERS` named the live host, on a run that was supposed to be entirely local. A suite without that guard would simply have connected.

Separately, one operator found a shipped `||=`-style inheritor on the launch path that propagates these values onward rather than requiring them explicitly, so the reach extends past plain `fork`/`exec` inheritance.

## Operator mitigation that works today

Not a fix, but it holds the line, and the negative control is the part that matters:

- **Scrub, then assert** — in-process, delete `COTAL_SERVERS` / `COTAL_SERVER` / `COTAL_CREDS` / `COTAL_SPACE` (and the control pair) as the **first action**, *then* assert that no remaining environment variable mentions the live host and that the URL actually in use is an ephemeral loopback address. Neutralize first and verify the neutralization; asserting alone leaves a suite that reads the variable before the check.
- **Prove the guard fires.** A guard that has never refused anything is not known to work. Drive it once against a deliberately bad value and confirm the refusal.

## Suggested direction, not a prescription

- **Do not inherit by default.** Pass connection material to the session process without exporting it into the environment children see — a file descriptor, a scoped config read at startup, or an explicit opt-in for children that genuinely need mesh access.
- **Scope the control token to the session process, not its descendants.** The control socket is the sharper half of this; a build script has no business holding a control-plane bearer.
- At minimum, if the variables must exist, **strip them from spawned child processes** unless the caller opts in, so "run a command" stops meaning "delegate my identity".

## Reproduce

1. From a manager-hosted agent session, run any shell command that prints its environment.
2. Observe `COTAL_CREDS`, `COTAL_CONTROL_TOKEN`, `COTAL_CONTROL_SOCKET`, and a live `COTAL_SERVERS` in the output.
3. Note that step 1 involved no credential handling of any kind.

### LINKED ISSUE #424: fix(manager): agent seats export COTAL_SERVERS and COTAL_CREDS into every child process

davidfarah2003:
Closes #396. Closes #424.

## The defect

A manager-hosted agent seat received its credential, its broker URL and its local control token as
**environment variables**. A process environment is inherited transitively and invisibly, so the seat
handed all three to every build, linter, test suite and third-party CLI it ever shelled out to.
Nothing in that chain asked for them, and there is no point where a human sees a credential being
passed, so there is no natural moment to object.

#396 and #424 are the same mechanism at two surfaces: one construction site, one commit. Every
connector built the seat's environment as `{ ...launchEnv(), ...aclEnv(), ...userAuthEnv(),
COTAL_CREDS, COTAL_SERVERS, COTAL_CONTROL_SOCKET, COTAL_CONTROL_TOKEN }` and the runtimes handed
exactly that to the seat process. `extensions/connector-core/src/launch.ts` was already the audited
chokepoint for the other direction of bleed (the operator's environment into the child) and documents
its scope honestly. What nothing covered was Cotal's own material flowing out of the seat and into
its descendants.

## Reproduced live before anything was written

On a sandbox mesh with its own project root, its own `XDG_CONFIG_HOME` and its own `COTAL_HOME`, on a
private broker port. Never the operator's mesh, never operator credentials. Full transcripts are on
both issues.

A real seat, spawned through the manager, ran one shell command. That command was an ordinary `node`
process, a direct child of the seat, doing no credential handling of any kind. It reported the whole
`COTAL_*` set including `COTAL_CREDS`, `COTAL_SERVERS`, `COTAL_CONTROL_SOCKET` and
`COTAL_CONTROL_TOKEN`; it could read the creds file; and it opened an **authenticated broker
connection as the seat** using only what it had inherited. The negative control, the same probe with
an emptied environment, could not connect at all, which is what establishes that the authentication
came from inheritance and nothing else.

After this change, re-measured on the same sandbox shape at the tip of this branch: a real seat
spawned through the manager, joined, and asked to run one shell command. That command is an ordinary
`node` process whose `PPID` is the seat, and it reports `COTAL_SERVERS` absent, `COTAL_CREDS` absent,
`COTAL_TOKEN` absent, `COTAL_CONTROL_TOKEN` absent and `COTAL_LAUNCH_MATERIAL` absent. The seat
itself joined the mesh, registered presence and executed the instruction, so the working path is
intact rather than merely quiet, and the material file the seat read at startup, along with the
private directory that held it, was already gone from disk by then. Full output is on both issues.
Box load average at the probe: 1.77 / 3.85 / 5.27.

One measurement worth keeping, because it is a trap for the next person: `/proc/<seat>/environ` still
shows `COTAL_LAUNCH_MATERIAL` for a seat whose session dropped it. That file is the environment as it
was at `exec` and never updates. What a descendant actually receives is built from the parent's live
environment, which is what the probe measured and what matters.

## The fix

The material rides a **file**, never argv (visible in a process listing) and never the ambient
environment (inherited). That is not a new invention: it is the shape `cotal agent-bearer` already
uses for its spawn-time secret, and the reasoning is recorded on that function.

- `packages/core/src/launch-material.ts` writes one 0600 JSON file inside a 0700 `mkdtemp` directory
  and reads it back. It refuses an empty material file, refuses malformed content, and on POSIX
  refuses a file readable by group or other, because a material file other local users can open is
  the disclosure this carrier exists to prevent wearing a different hat.
- `materialEnv()` in `connector-core/src/launch.ts` replaces `userAuthEnv` and the per-connector
  `COTAL_CREDS` / `COTAL_SERVERS` / `COTAL_CONTROL_TOKEN` assignments. It exports one variable,
  `COTAL_LAUNCH_MATERIAL`, carrying a path.
- `configFromEnv` reads the file when that variable is set. Both carriers at once is refused rather
  than resolved by precedence: two answers to "who is this session" is how a seat ends up connected
  as something nobody chose. An unreadable file throws rather than falling back to the environment,
  which would turn a broken launch into a quietly different one. The refusal covers every direct
  carrier, `COTAL_LINK` included, because a join link carries the server, the auth and the space in
  one string; left off the list it did not conflict, it lost quietly to material precedence for the
  server while still contributing the rest of the identity, which is worse than either answer.
- `controlFromEnv` gives every consumer the socket path from the environment and the first-frame
  token from the material. It returns nothing when neither half is present, which is a session with
  no control plane and a normal launch, and it **throws on half a pair in either direction**.
  Returning `undefined` there made each caller's own check the real contract, and the callers did
  not agree: a session could run believing it had a control plane it did not have. One exception,
  and it is deliberate: the lifecycle hook relay catches that refusal, because a hook that throws is
  a hook that blocked the session. It no longer catches it silently (see below).
- pi, OpenCode and codex drop the pointer once both readers have used it, so the shells and tools
  those seats run inherit no reference at all. Stated precisely, because "the seat process" is not
  the same process in all three: for pi and codex the session runs in the seat process itself; for
  OpenCode the seat process is a shim that starts `opencode serve`, the plugin runs inside that
  server, and the server is also what executes the session's tool calls. The shim keeps the
  reference and runs no tools of its own. Those three also **discard the file** at that
  moment. There is provably no later reader (`serve.ts` does not respawn its server), so a file that
  still exists is exposure with no purpose, and this is the one place where the residual boundary
  gets smaller rather than better documented.
- `discardLaunchMaterial` lives beside the writer, sharing its filename and directory-prefix
  constants, so the rules for what is safe to delete are stated once by the module that created the
  thing. The named file is removed unconditionally, because the variable names material for one
  launch that the session has already read. The private directory is removed only on structural
  proof: the exact filename, the writer's prefix, a parent that resolves through symlinks to the
  resolved OS temp root, and `rmdir`, which cannot recurse. Anything else inside makes the removal
  fail and the directory is left as it was found, which is also what closes the window between the
  two removals. Deleting the named file is the operation the caller asked for; deleting a directory
  is an inference about something the caller never named, and an inference needs proof.

One design note for whoever picks up the neighbouring problem: `materialEnv()` is now the single
place a launch decides what material a seat may hold, which makes it the natural chokepoint for a
future per-tier check on that material. #373 is a different mechanism (a grant armed at spawn rather
than material inherited through the environment) and is deliberately not folded in here, but the two
should compose rather than collide: this change adds a gate on WHAT rides the launch, and leaves WHO
may ask for it exactly where it was.

Two connectors keep the pointer, and the PR should be read as claiming less for them. **Claude**'s
readers are short-lived children (the MCP server, one process per lifecycle hook) that start after
the session is running; **Hermes**' launcher starts a gateway child that needs the control token.
For both, a shell the seat runs still inherits a path to the material file, though not the material
itself. The three that drop it are pi, codex and OpenCode.

**File lifetime, exactly.** One file per launch, in a 0700 directory of its own. For pi, codex and
OpenCode it is deleted as soon as the session has read it. For Claude and Hermes it lives as long as
the seat, in the OS temp tree, and is reaped by the OS rather than by us, because the processes that
read it start later and there is no moment we can name as the last read. The Hermes launcher writes
a merged file (its own minted control token folded into whatever it inherited) and **deletes the
inherited copy**, so it holds one file and not two. A launch that fails before the session starts
leaves its file for the OS reaper; that is the retention this paragraph is accepting rather than
claiming away.

The control **socket path** stays in the environment on purpose. It is a path, the socket refuses an
unauthenticated frame, and the short-lived lifecycle hook processes need it. Only the token moved.

Driving a connector by hand is unchanged: set `COTAL_CREDS` / `COTAL_SERVERS` yourself and no
material file is involved. Docs updated in the same change (`docs/config.md` gains a Launch material
section; `docs/identity-and-auth.md` follows).

## What depends on the ambient export, measured

Counted on `main` before the patch, because a fix that turns every connector suite red is the wrong
fix. The greps are stated rather than summarised, because a review of the first version of this
section corrected two of the numbers and the only way to settle that is to publish the query.

**25 files** match a read of either variable off the ambient environment:

```
git grep -lE 'process\.env\.COTAL_(CREDS|SERVERS)|process\.env\["COTAL_(CREDS|SERVERS)"\]|\be\.COTAL_(CREDS|SERVERS)\b' \
  origin/main -- '*.ts' '*.mjs' | wc -l
```

That figure is a match count, not a count of things that break, and it is worth reading as such: the
manager's live suites are the large group inside it and they are **unaffected**, because they
register their own stub connectors that build the child environment themselves. They exercise the
manager, not the connector contract.

**8 paths** match the `||=` / `??=` defaulting query:

```
git grep -lE 'process\.env(\.COTAL_SERVERS|\["COTAL_SERVERS"\])\s*(\|\||\?\?)=' origin/main -- '*.ts' '*.mjs'
```

**Seven of those are real suites; the eighth is `bin/smoke/broker-disclosure.smoke.ts`, the grader
itself, matching on its own fixtures.** They tolerate the export rather than depend on it, and each
discloses which broker it resolved. They pass unchanged, and after this change a seat no longer
supplies the value they were silently inheriting. That is the test-isolation half of #424.

**Three suites asserted the old contract and all three are updated here; a fourth needed a change
for a different reason:**

- `extensions/connector-codex/smoke/codex-args.smoke.ts`, `extensions/pi/pi.smoke.ts` and
  `extensions/connector-opencode/smoke/cooperative-stop.smoke.ts` asserted the material variables
  were present in the launch env. All three now assert BOTH halves: absent from the environment,
  recoverable through the material.
- `extensions/connector-claude-code/smoke/wake-path.smoke.ts` is not an old-contract assertion. It
  sets `COTAL_CONTROL_TOKEN` as a harness input to drive the relay, and it needed a strip because the
  two-carrier refusal now covers that variable.

One more instance is worth naming because it will otherwise be misread as a failure of this change.
Run the smoke chain from inside a launcher-spawned agent session on a RELEASED build, and
`implementations/manager/smoke/gate-reconcile-auth.smoke.ts` refuses at its first action, before doing
any work, because the ambient `COTAL_SERVERS` it inherited names the live broker. That suite
provisions credentials, revokes rows and kicks connections, so refusing is exactly right, and it is
the only reason the inheritance was visible at all. It is this PR's subject arriving in the runner
rather than in the code: the file is untouched here, the same run with the ambient `COTAL_` variables
unset passes 33 checks, and once this change ships such a session's descendants do not receive the
variable in the first place. The suites in the frozen census have no comparable guard, which is the
difference the census exists to name.

The last two were found the hard way rather than by the grep, and the way they were found is the
point. `cooperative-stop` booted its probe with `{ ...process.env, ...spec.env }`. Where that suite
runs inside a managed agent session the ambient environment carries a live credential and a live
broker URL, and the connector env used to overwrite exactly the two variables that would have made
the inheritance visible. It no longer does, so the two-carrier refusal fires and the child dies with
a sentence naming both carriers. Found by the full suite going red at 133 of 338. `wake-path` had
the same spread and was found by reading, after the first one.

## Proof

`bin/smoke/seat-env-scope.smoke.ts`, wired into `bin/smoke/ci-suites.txt`:

- **A2** starts a real `node` process with each connector's launch environment and has THAT process
  start a child with no environment argument, exactly as a coding agent's shell tool does. What the
  child reports is what the operating system actually inherited. It runs first, so it is the
  assertion a reintroduced export reddens.
- **A1** sweeps all five production connectors for the forbidden variables.
- **A3** asserts the working path for each: broker URL, creds content, identity and the control token
  round-trip through the material file to the exact values the manager holds. Without A3 a connector
  that exported nothing at all would read as fixed.
- **A5** pins the ordering hazard: dropping the pointer makes the control token unreachable, so every
  reader must run before the scrub. It also asserts the file and its private directory are gone, not
  merely unreferenced, and that what is left of a scrubbed environment refuses instead of
  half-working.
- **A4** proves the permissive-mode refusal fires.
- **A6** proves a material file that says nothing is refused on READ, and that a valid one still
  reads, so the four refusals are not satisfied by a reader that refuses everything.
- **A7** proves a material file alongside a direct carrier (`COTAL_LINK`) is refused by a message
  naming both, rather than ranked silently.
- **A8** proves half a control pair throws in both directions, and that no control plane at all is
  still a normal launch.

Two things this suite does **not** prove, written into its own header so a reader finds them there
rather than assuming coverage that is not present. A2's seat is a bare `node` process, not a live
harness, so it never runs a connector's own scrub: A2 proves the VALUES are not inherited, and the
live spawn is what evidences the pointer being dropped. And A3 skips the control token for hermes,
whose endpoint is minted inside its own launcher; that connector's control path is covered by
`smoke:hermes-hooks-control` instead. The second gap is not theoretical, as the section below
records.

`pnpm mutation-proof` against `bin/smoke/mutations/seat-env-scope.mutations.json`: **10 of 10 killed,
each on its named assertion.** M1 puts the broker URL back into the pi seat environment and reddens
A2. M2 puts the control token back into the claude environment and reddens A1. M3 makes the material
silently lose the broker URL and reddens A3, which is the failure the absence assertions cannot see.
M4 makes the reader accept a material file that says nothing and reddens A6. M5 drops `COTAL_LINK`
back out of the direct-carrier list and reddens A7. M6 restores the half-pair `undefined` and reddens
A8. M7 restores a scrub that drops the pointer without unlinking and reddens A5. M8 removes the file
but leaves its private directory, and reddens A5's second assertion. M9 restores a recursive
directory removal and reddens A9's assertion about a bystander file. M10 drops the requirement that
the directory sit in the OS temp root while leaving the name check in place, which is the version
that looks careful and is not. Every mutation builds first,
because `@cotal-ai/core` and `@cotal-ai/connector-core` resolve to `dist/`, and each restores by
rebuilding. M5 through M8 each restore a real prior state of this branch rather than an invented one.

One of them was graded WRONG-RED before it was killed, and the reason is worth recording because it
generalises. M6 made the suite exit 1, but at A5's half-pair assertion rather than at A8, which
exists for it: the suite is fail-fast and A5 ran first. A red run plus a plausible story would have
been reported as a kill and would have been wrong about which assertion guards what. The fix is
ordering (A8 now runs before A5) and the constraint is written at both legs, including why A5 has to
stay last: its scrub unlinks the file A4 needs on disk.

The tool's own caveat applies and is answered separately: a killed mutation shows the suite depends
on the mutated code, not that a real entry point reaches it. The live sandbox spawn is that second
proof, and it earned its place. It caught a real ordering bug this suite did not see: the pi
extension dropped the pointer before asking for the control token, and every pi launch refused with
the contract's own error message. Fixed here, and pinned by A5.

Box load average during the mutation proof run: 5.30 / 6.11 / 7.35. A local green is not a CI green;
these are answers to different questions.

## A defect this change introduced, found in review, fixed here

Worth its own section because it is the most useful thing the review produced and because the shape
generalises.

The hermes lifecycle hooks are **Python** (`extensions/connector-hermes/plugin/cotal/hooks.py`).
They read `COTAL_CONTROL_TOKEN` out of the environment and return early, silently, when it is
absent. When the token moved into the launch material, every TypeScript consumer was updated and
proved, and this one was not. The result was a hermes seat that joins the mesh, runs its sidecar and
bridge normally, and never moves off its first presence value for the life of the session. No error,
no exit, nothing in a log.

This is the reason the cell suite alone was not sufficient evidence for this change, and it is worth
stating rather than leaving implied: a seat that joins the mesh, runs its sidecar, and silently never
advances was invisible to every assertion in the tree, because they were all written in the language
that was already correct. It was found by reading, not by running.

Nothing in the tree could have caught it. The seat-environment cell reads launch specs; its A3 leg
skips the hermes control token because that connector mints its endpoint inside its own launcher and
`buildLaunch` returns none; and no suite loaded the Python plugin at all. Every assertion was written
in the language that was already correct.

The hooks now read the material file, keep the direct variable as the standalone path, refuse a file
readable beyond its owner the way the TypeScript reader does, and **warn once on stderr** when a
control socket is configured but no token can be resolved. That last part is the actual repair: the
defect's whole character was that it produced no output.

`smoke:hermes-hooks-control` drives the real `hooks.py` through a real `python3` against a real
control listener, and is gated in `ci-suites.txt`. `pnpm mutation-proof` against it kills two: the
shipped defect restored (reddens the managed-launch leg) and the silent return restored (reddens the
assertion that refuses to accept silence as success). The second matters more than the first.

## The same defect one connector over: silent fail-open in the hook relay

Found by the security lens reading `relay.ts` after the Python finding above, and confirmed by the
engineering lens, which owns having walked past it on its own first read.

`runHookRelay` resolved the control endpoint inside a `try/catch` and returned an empty reply on any
failure. Fail-open is correct and stays: a lifecycle hook that throws is a hook that blocked a human's
session, and no configuration fault is worth that. Silent fail-open is not correct. A material file
that is missing, permissive, malformed, or contradicted by a direct carrier produced a hook that did
nothing and said nothing, which is the same shape as the Python defect: the symptom is dead presence
and no injected messages, with nothing anywhere to read.

The relay now writes **one bounded warning to stderr, with no interpolated values**, and then fails
open exactly as before. It warns on the refusal only, not on the ordinary no-control case: a
hand-driven session that sets `COTAL_NAME` and never had a control socket is a legitimate launch, and
warning there would fire on every hook of a working session and teach the operator to ignore the
channel. Bounded means once per hook process, and the runtime starts one process per lifecycle
event, so a genuinely broken session keeps saying so. That is the intended trade: the alternative is
one line early in a session that scrolls away, for a fault that persists until somebody fixes the
launch.

Cells R1 to R3 in `wake-path.smoke.ts` drive the real hook entry point as its own process: one
warning on the refusal, still exit 0 with an empty reply, no value anywhere in the text, and silence
on both the resolvable case and the no-control-plane case. Two mutations, both killed on their named
assertions: restoring the silent catch reddens R1, and moving the warning to the no-control branch
reddens R3. The second is the one that matters as much as the first, because over-warning destroys
the value of the repair without breaking a single functional assertion.

## A second defect this change introduced, found by reading, fixed here

The cleanup that removes the material's private directory was, in its first form, the most
destructive thing in the diff.

It removed the parent **recursively** whenever the parent's basename began with the writer's prefix,
under a comment claiming that proved the directory came from the writer. It proved that a string
starts with another string. A pointer set by hand at any path whose parent happened to be named
`cotal-launch-something` would have taken that parent and every sibling file with it, and the gap
between removing the file and removing the directory turned a concurrent create into collateral
deletion. A cleanup added to a change about reducing harm must not be the most dangerous thing in it.

The root cause was placement as much as logic: deletion logic sat in a file that did not know how the
thing had been created, and then asserted knowledge the code did not have. The removal now lives
beside the writer as `discardLaunchMaterial`, shares its constants, and is used by every caller
including the Hermes launcher, so no connector can grow its own idea of what is safe to delete.

A9 is the cell, and it is shaped so that the obvious version of it would not have worked: the
assertion that catches the recursive case watches a **bystander file**, not the directory, because
the directory disappearing is the correct outcome in the ordinary case and an assertion watching only
the directory would stay green under the dangerous version.

## Two more review findings, fixed here

**The material was refused when empty only at WRITE.** The reader is handed a path by an environment
variable and never sees the writer, so a file that is empty, truncated or shaped wrong reached it
unchecked. `{}` parsed cleanly into a material with every field undefined, `configFromEnv` filled in
its defaults, and a launch that REFERENCED material and supplied none resolved to the default broker
with no credential: open mode, silently, on a launch that asked for the opposite. Every field is now
graded on read, and a material with nothing the reader recognises is refused. Covered by A6 and by a
mutation that removes the refusal and reddens it.

**Nothing could see a suite capturing an identity from its own runner.** That is how both of the
ambient-spread defects above reached this branch, and a runtime cell structurally cannot catch it:
`seat-env-scope` watches what a connector hands a real descendant, not what another suite's harness
inherits. `smoke:suite-ambient-env` is a static census over the suite sources requiring a `COTAL_`
strip before any `...process.env` spread. It ships as a **ratchet and says so in its own header**:
the 44 pre-existing spreads are frozen and explicitly NOT cleared, the set cannot grow, and an entry
that stops spreading must leave the list. Several of the frozen files spawn the `cotal` CLI, which
does read connection material, so some are likely real instances of the same defect; auditing them is
its own piece of work and does not belong in a credential change. Two mutations: deleting one strip
reddens the offenders assertion, and freezing a file that no longer spreads reddens the ratchet.

The audit is **filed as #619**, not deferred to a comment, and the census header cites that number.
A frozen baseline pointing at an open issue is a deferral; one pointing at nothing is a waiver that
outlives everyone who agreed to it. #619 carries the 44 names and two measurements taken while
filing: 15 of them spawn or reference the `cotal` CLI, and 11 have a spread site that sets no
`COTAL_` variable of its own, so an inherited value reaches the child unopposed. Four are on both
lists. The second measurement is a fixed-window text scan and is recorded there as a lead, not a
finding.

The two in-tree instances fixed here are worth naming as what they are: this class of defect was
already present in this repository's own tests, and it was invisible for the same reason the seat
leak was invisible, because the connector environment overwrote the variables that would have shown
it.

One behaviour change worth naming because the "driving by hand is unchanged" sentence does not fully
cover it: hand-driving the **hermes launcher** with direct `COTAL_CREDS` / `COTAL_SERVERS` now hits
the two-carrier refusal, because that launcher introduces a material file itself. It fails loud, it
is not a shipped path, and the standalone hermes plugin is unaffected.

## Scope, stated plainly

The accurate claim, and everything else here has to live up to it: the file carrier is an honest
reduction of transitive environment inheritance, not same-uid isolation.

This removes the values from every descendant's environment. An `env` dump, a CI log capture, a suite
that reads its broker from the environment, a tool handed a credential it never asked for: all of
them stop seeing them. It does **not** hide the material from a process running as the same user that
deliberately opens the referenced file, and no environment-level control can. The same is already
true of `~/.cotal/auth/creds`, and `launch.ts` already documents filesystem secret access as out of
scope for env-level controls. What changes is that reaching the material becomes a deliberate act
instead of an inheritance nobody chose.

One narrowing is **explicitly staged rather than guessed at**. The Claude connector is the one that
cannot drop the pointer after startup: its readers are short-lived children, the MCP server and one
process per lifecycle hook, that start after the session is already running. So a shell `claude` runs
still inherits a path to the material file. Narrowing that means handing the MCP server its material
through the `--mcp-config` `env` block and giving the hooks a separate control-only file, which
depends on host behaviour that has to be verified against a live `claude` first. It is reasoned about
in the connector rather than half-wired.


davidfarah2003:
## A conditional in `materialEnv` that is safe here and becomes unsafe the moment the env boundary relaxes

Surfaced while reviewing a proposed change to the launcher's environment policy, verified first-hand
against this branch's `extensions/connector-core/src/launch.ts`. **This is not a defect in this PR as
it ships.** It is a property of the shape that is worth stating in the source, because the safety
depends on something outside this file.

`materialEnv` is conditional in the same way the family it replaces was:

```ts
// Nothing to hand over (an open mesh launched with no control endpoint) → no file and no env
// entry, rather than a file that says nothing.
if (Object.keys(material).length === 0) return {};
return { [LAUNCH_MATERIAL_ENV]: writeLaunchMaterial(material) };
```

So a launch with nothing to hand over assigns nothing. Today that is harmless, because the launcher
forwards a fixed allow-list and `COTAL_LAUNCH_MATERIAL` is not on it, so the variable cannot cross
into a child at all.

What makes it worth a note is what the variable now holds. Before this PR, an unset conditional var
fell through to a role, a space, an ACL. After it, `COTAL_LAUNCH_MATERIAL` names **the path to the
0600 file holding the credential and the control token**. If the launcher ever inherits the operator
environment instead of forwarding an allow-list, a child launched on the empty-material path (the
open-mesh case this very comment names) would inherit the parent's variable and be handed a path to
the parent's material.

That is a real prospect rather than a hypothetical: relaxing the env boundary is under active
discussion, on the grounds that Cotal should not be opinionated about the operator's environment.

**The point is not that this PR should defend against a change that has not landed.** It is that this
PR's own framing is hardening: it moves connection material out of the environment precisely so that
an environment leak stops mattering. That argument holds only while the one variable that survives
cannot itself leak, and nothing in this file says so. A reader who later relaxes the boundary has no
signal here that this variable is different in kind from the identity vars beside it.

Two cheap ways to close it, either is enough:

1. **Say it where it is decided.** One line in the docblock: this variable must never be inheritable,
   because it dereferences to the credential rather than describing the seat. A future contributor
   relaxing the boundary then meets the constraint at the place they would have to violate it.
2. **Grade it.** A cell asserting that a child launched with empty material has no
   `COTAL_LAUNCH_MATERIAL` in its environment, whatever the parent's environment held. Under the
   current allow-list that passes trivially; its value is that it fails loudly the day the boundary
   changes, which is exactly when nobody will be thinking about this function.

The prefix-strip design in the proposed change happens to cover this variable automatically, without
its author having known the variable existed. That is an argument for that design rather than a
reason to leave this unstated here: the guarantee should not depend on which change lands first.

PR #672, opened 2026-08-19
fix(connector-core): an empty id is never a dedup key, so two id-less messages both deliver

### LINKED ISSUE #624: fix: two messages with an empty id collapse to one at ingest, and the second is never delivered
Found while reviewing #607. Not caused by it and not fixed by it.

The inbox's receive-time dedup keys on the message id. Two messages that arrive with `id: ""` collapse
to the first: the second is treated as a duplicate of the first and its body is never delivered.

Measured in process, driving the agent's own message path with two distinct bodies and an empty id on
both: one item buffered, the first body, the second silently gone.

Official publishers set `randomUUID()`, so this is not reachable from the first-party client. It is
reachable from a raw publisher on the same subject, and the failure mode is the one this repo cares
about most: a message accepted, acked, and never handed over, with no reply saying anything was
withheld.

## Acceptance

- A message with no id is either refused at ingest with a named reason, or given an id that cannot
  collide, whichever the wire contract prefers. Silently folding it into an unrelated message is the
  one outcome to rule out.
- A cell driving two empty-id messages with distinct bodies, and a mutation that fails without the
  repair.
- If the wire contract already requires a non-empty id, the refusal belongs where that requirement is
  enforced, and the cell should show the refusal rather than the delivery.

davidfarah2003:
`MeshAgent.ingest` (`extensions/connector-core/src/agent.ts`) coalesces deliveries by message id on three read sites (already-handled, protected disposition, still-pending). An id of `""` satisfies all of them as a key, so two DISTINCT messages that each carry an empty id read as duplicates: the second is silently dropped, and after the first is drained the empty id sits in `handledIds` so every later empty-id message drops on arrival. The only route to an empty id is a foreign client on the wire (first-party publish APIs mint ids), so this is silent, permanent message loss under a conformant wire client.

## Fix

An empty id is never a dedup key, in both directions. One derivation guard (`ingestDedupKey`) gates all three ingest reads; `markHandled` and `protectDisposition` decline to record an empty key, so no read can ever consult it. Two distinct empty-id messages now both deliver, a third after drain delivers, and dedup for real ids is untouched. Refusing empty ids at ingest was rejected because the requirement is that both messages deliver.

Also folded (same defect class, deliberate): `drainInboxIds` used the empty id as a selection key and swept every pending empty-id item in one drain, acking unsurfaced messages. The same guard applies safely: the empty id is filtered out of the requestable set, the items stay for a scope drain. Cell pair plus mutation K4 prove it.

Stated cost, not hidden: no id means no coalescing, so a redelivered id-less copy may surface twice. That is the wire's at-least-once stance, and it is the correct trade against dropping distinct messages.

## Spec and docs

SPEC §4, §7 item 5, and Appendix B item 12 said receivers MUST dedupe by id with no empty carve-out, so a spec-built client reproduces the collapse. All three now state that an empty id is never a dedup key, with the absent-id case staying malformed under §5. `docs/presence-and-delivery.md` and `docs/build-a-client.md` are updated; the docs bundle is regenerated. Because this edits root `SPEC.md`, it is Tier B for David's list.

## Proof

`pnpm smoke:empty-id-ingest` (new suite, own broker, raw-wire entry so no hand-built unit inputs; appended to `ci-suites.txt` at strict EOF, main's line count an exact prefix of the new one) 9/9 green. Mutation-proof 4/4 KILLED on named assertions: K1 raw-id keying restored, K2 pending-lookup-only relapse, K3 dedup disabled entirely (proves no weakening for real ids), K4 empty-requestable drain sweep. Neighbor suites green (cross-path-dedup, inbox-window, attention, channel-attention, durable-source, subject-frontier); typecheck clean.

## Deliberately not built, and disclosed follow-ups

The focus-recall paths (`focusExcludedIds`, `aheadDelivered`) are NOT folded: there the empty id is a disposition carrier, and the same blanket guard would fix one lane by breaking two others (double-surfacing quiet-channel and leaking muted traffic). That needs per-message keying plus per-disposition policy and is its own issue. Two further residuals are disclosed rather than overstated as fixed here: the adapter raw-id surface / commit seam, and the Plane-3 derived-msgID publish-dedupe on durables.

Closes #624.


davidfarah2003:
Rebased onto current main `76952256` and integrated the two later changes that overlapped this branch instead of silently reverting them:

- preserved #809's bounded overflow give-up semantics, re-keyed on per-delivery `recvKey` so id-less deliveries do not share an eviction tally;
- preserved #810's Jcode retry pacing/waiting-status policy and migrated only its successful-turn drain seam from wire ids to receive keys;
- preserved every current-main smoke registration and regenerated the connector docs bundle.

Measured on the rebased head:
- `pnpm typecheck` — rc 0
- empty-id ingest, Hermes empty-id, overflow churn, directed overflow, cross-path dedup, inbox window, Claude wake, and Jcode retry-policy smokes — all rc 0
- empty-id mutation fixture — 5/5 mutations killed on named cells
- `smoke:jcode-host` reaches the same 11 cells as current main and then both artifacts hang at the merged #843 JSON-own `__proto__` live-relay cell; untouched main was independently run at `76952256` and timed out at that identical cell, so this is a landed baseline, not a #672 regression.

No main-side files or smoke registrations are dropped by the rebased diff.

PR #728, opened 2026-08-19
fix(cli): announce the operator-global seed-store write with its path

### LINKED ISSUE #593: cotal up rewrites the operator-global extension seed store without saying so
## What happens

`cotal up` provisions the extensions a mesh needs by writing an operator-global store, not a
per-space or per-checkout one:

- `~/.config/cotal/seed/store/<version>/*` is re-seeded from the packages carried by the binary that
  is running.
- `~/.config/cotal/extensions` is then installed from that seed.

Both paths are shared by every space, every project directory, and every checkout on the machine.
`XDG_CONFIG_HOME` moves them, and nothing else does: a fresh project directory isolates `.cotal/`
and does not isolate this.

The output does not say so. It reads as provisioning for the mesh you just asked for:

```
✓ added @cotal-ai/connector-claude-code@0.22.0 - provides: connector:claude
✓ added @cotal-ai/connector-opencode@0.22.0   - provides: connector:opencode
...
✓ running in the background: nats-server (pid …), delivery daemon, manager
```

Nothing in those lines names a path outside the current directory.

## Why it matters

Run `cotal up` from a checkout that is not a released tree, for example a feature branch or a local
build, and the seed store for that version key now holds those packages. Every later mesh on the
machine that materializes an extension gets them, including meshes started from a different
directory by a different user of the same account, until something re-seeds the key again. The
version number does not change, so `cotal ext ls` and the extensions manifest both report the
published version while the bytes on disk are the local build.

It is easy to hit without noticing. Standing up a throwaway mesh in a temporary directory to try
something out looks like a fully local action, and the isolation a temporary directory gives you is
real for `.cotal/` and absent for this.

## Suggested remedies

1. **Announce the write.** When `up` seeds or installs into the global store, print the path it is
   writing, so the line reads as a machine-wide action rather than a mesh-local one.
2. **Document the isolation mechanism.** Say in the docs that `XDG_CONFIG_HOME` relocates the seed
   store and the extension prefix, and that this is what to set when running `up` from a tree that
   is not a released install.
3. **Decide what `up` should do when its own packages differ from the published ones at that version
   key.** Overwriting silently is the current behaviour. Refusing, or requiring a flag, or seeding
   under a distinct key would each avoid the mismatch between the reported version and the bytes.
   This one is a design question and this issue does not try to answer it.

## Repro

```
mkdir -p /tmp/scratch/.cotal && cd /tmp/scratch
<path-to-a-non-released-checkout>/bin/cotal.ts up --detach --server nats://127.0.0.1:PORT --space scratch
# then, outside that directory:
ls -l ~/.config/cotal/seed/store/<version>/opencode
# the contents are the checkout's packages, and the version key is unchanged
```

Verified on macOS with a 0.22.0 binary. The seed store's `opencode` directory carried source that
exists only in the local checkout, while the extensions manifest reported `0.22.0`.

davidfarah2003:
## What breaks today

`cotal up`, and every command that reconciles the built-in connectors, materializes each shipped
connector payload into `~/.config/cotal/seed/store/<generation>/<name>`. That store is
operator-global: a sibling of the shared `extensions/` prefix, shared by every space, every project
directory and every checkout on the machine, and moved only by `$XDG_CONFIG_HOME`. A fresh project
directory isolates `.cotal/` and does not isolate this.

Both directions of that store were silent. Nothing in the output named a path outside the current
directory when a payload was written, so running `up` from a tree that is not a released install
re-seeded the machine-wide payload for that generation key while `cotal ext ls` kept reporting the
published version. And nothing named the DELETE either: a forced reconcile drops an unreferenced
generation with a recursive removal under the same root, exits 0, and said nothing about it.

## What changes

The store now announces both of its machine-wide acts on the provenance channel, each naming its
path:

```
→ wrote operator-global seed store payload (opencode): ~/.config/cotal/seed/store/0.23.0/opencode
→ removed operator-global seed store generation (0.22.0): ~/.config/cotal/seed/store/0.22.0
```

The write fires on a real materialization only. The idempotent early return, taken when an intact
prior copy is already in the store, copies nothing and stays silent, so the line tracks writes rather
than boots and a repeated `up` adds no noise that means nothing. `provenance` gains a `removed` verb
for the destructive half, because a silent delete is worse than a silent write: the write at least
leaves the thing it made.

`docs/config.md` states the isolation mechanism in the same change: both `extensions/` and
`seed/store/` are operator-global, `$XDG_CONFIG_HOME` (or `%APPDATA%` on Windows) is what relocates
them, and that is what to set when running from a checkout that is not a released install.

## The bound on the announce, stated where the claim is made

Both lines ride the provenance channel, which is a plain stderr write with no failure policy, and
they are emitted after the act. If stderr is closed or erroring, the payload is still written and the
line is lost. So the guarantee is "an act that reaches a working stderr is named", not "no act is
ever unannounced".

Announcing before the rename would trade that for a worse lie, a named write that then failed.
Throwing on a lost line would fail a seed over a disclosure fault, after its durable act. The
channel-wide policy is not this call site's to invent, so the bound is stated in the source and in
the docs, and the fix surface is filed as #730.

## What this does not do

It does not decide what `up` should do when its own packages differ from the published ones at that
generation key. Overwriting is still the behaviour, now visible rather than silent. That is remedy 3
of the issue, which the issue itself declines to answer, and so does this change.

It does not make the extensions-prefix install visible on the reconcile path. That write IS announced,
but inside the `ext add` seed child, whose stderr the parent drops on success. Pre-existing, unchanged
here, and named so it is not mistaken for covered.

## Evidence

- `pnpm smoke:seed`: 68 passed, 0 failed, rc=0. Four cells carry the change: the first boot announces
  the write, the announced destination is the payload directory itself rather than the store root, a
  boot that copies nothing announces nothing, a re-seed that reuses an intact payload announces
  nothing, and a forced reconcile that drops an unreferenced generation names the removal.
- `pnpm mutation-proof` over `implementations/cli/smoke/fixtures/seed-store-announce.mutations.json`:
  five mutations, each stating which kind it is. Four break a guard: delete the write announce, name
  the store root instead of the payload directory, move the announce onto the idempotent early return
  so a reuse announces, delete the removal announce. The fifth models the steady-state failure no
  single edit can produce, because that path is held by two independent layers and breaking either
  alone still copies no bytes.
- `pnpm typecheck`: rc=0, stated precisely because it is weaker than it sounds. `implementations/cli`
  includes only `src`, so the smoke file is not in that program; it was compiled directly instead.
- `pnpm check:docsbundle`: rc=0 (the generated bundle embeds `docs/config.md`, so it is resynced in
  the same PR).

Closes #593


davidfarah2003:
Verification status, stated plainly because one lens did not finish cleanly.

**Security lens: APPROVE** at `ce86991b48bc6566ea5fb5e8075a3d843e79bf88`, model `cliproxy/grok-4.6:high`. It drove the change rather than reading it: isolated run under a temporary `XDG_CONFIG_HOME`, seven write announces observed with destinations confirmed present on disk, file modes tabled, `provenance.{wrote,removed,read}` traced to confirm only the two supplied arguments are interpolated. `seed.smoke.ts`: 68 passed, 0 failed. No token, account id, or secret rides the announced line, and it is not broadcast.

**Independent operator reproduction** at the same sha, separate checkout:

```
seed smoke: 68 passed, 0 failed
mutation proof: All 5 mutation(s) killed. The suite discriminates.
  KILLED  the write announce names the store root instead of the payload directory it wrote
  KILLED  the announce moves onto the idempotent early return
  KILLED  the machine-wide generation DELETE is silent again
  KILLED  the announce is emitted from the steady-state NOOP boot
```

Each mutation went red at its own named assertion, so the suite discriminates rather than merely passing.

**Engineering lens: did not produce a substantive verdict.** Its artifact still carries an interim procedural BLOCK written early in the session, which it labelled itself as a fallback record and explicitly not a rejection of the substance. Its static review graded all five grading questions CLEAN. The reason it stalled is visible in its process tree: it launched the mutation proof, then parked on a `sleep 900` polling loop and stopped advancing. Two follow-ups asking it to collect its own results did not move the artifact.

That failure is about the lens, not the change: it never reported a contradicting measurement, and the two runs that did complete agree exactly. Filed separately as a supervision defect - a seat that starts long work and stops waiting for it looks identical to one that is thinking.

Residuals from the security review are filed as #795 (`cliVersion()` path traversal now also printed, plus three smaller pre-existing ones). CI green 17/17.

PR #750, opened 2026-08-20
fix(connectors): contain spawned environment by default

davidfarah2003:
## Summary

Restores the declared environment boundary for managed agent launches. Spawned children receive the fixed OS execution allow-list, connector-declared provider inputs, explicitly shared MCP references, and only names deliberately added through `spawn.env`.

`SSH_AUTH_SOCK` and other ambient operator capabilities no longer cross the default spawn boundary. `spawn.env` remains an explicit opt-in for a required extra name.

## Verification

- Reproduced the prior default through the shipped detached manager path with a fresh ephemeral ssh-agent. The child completed an ssh-agent signing challenge after the private-key file was deleted.
- Added a shipped-path two-arm regression:
  - D1-D3: default child has no `SSH_AUTH_SOCK` and cannot sign after key deletion.
  - O1-O3: `spawn.env: ["SSH_AUTH_SOCK"]` explicitly forwards the socket and the child signs.
- Mutation proof killed default ambient inheritance on D2 and removed explicit opt-in on O2.
- Passed build, typecheck, targeted connector/runtime smokes, mutation fixtures, docs bundle, changeset status, and the sanitized `smoke:ci` suite: 341 smokes.

## Scope

This changes the managed spawned-child environment boundary only. It does not claim filesystem confinement for files accessible through the supplied OS roots.


davidfarah2003:
Closing this in favour of #754.

The security finding behind it is real and was independently reproduced: default inheritance forwards `SSH_AUTH_SOCK`, and a spawned child completed an agent-backed signing challenge for a key whose private file had been deleted. The verification work here is good, and the two-arm shipped-path regression is the right shape.

The disposition is what differs. Inheriting the operator's environment by default is the intended behaviour, not a defect: an agent should resolve its harness, model and provider exactly as it does when the operator runs it themselves, and containment is opt-in through `spawn.env`. This branch removes `OPERATOR_ENV_KEEP`, drops the `envAllow` mode branch so no inherit mode remains, and makes the allow-list unconditional. That reinstates the default that was deliberately removed.

#754 keeps the behaviour and fixes what was actually wrong: `docs/config.md` claimed `spawn.env` protects only env-resident secret values while reassuring the reader that a shell reads `~/.ssh` regardless. Neither holds for a capability handle. It also documents the `codex`/`opencode` model-discovery exception, makes a vacuous env-boundary cell live, and adds the two ungated env suites to the sharded gate.

Anyone wanting containment gets it with one line of config, which is the trade this project chose.

davidfarah2003:
Superseded by #754; see the comment above for the reasoning.

PR #759, opened 2026-08-21
fix(cli): suggest supervise for a registered remote mesh

### LINKED ISSUE #757: up: refusal for a hand-registered space suggests `cotal meshes rm`, a destructive de-registration, where supervise is the right verb
`cotal up` on a workspace whose space is hand-registered to a remote broker refuses (correctly), but its remedy text is:

    space "main" is registered to a mesh at nats://broker.cotal.ai:4222 (<root>) - it was registered by hand, so `cotal up` neither takes it over nor reclaims the name: `cotal meshes rm main` to drop that record first, or start this one under a different `--space`

Both suggestions are wrong for the common case, where the mesh is real and the operator wants its control plane back:

- `cotal meshes rm main` drops the registry record a live remote mesh is addressed by (spawn-from-anywhere, attach, and every read verb hang off it);
- a different `--space` forks the space identity.

The correct next command was `cotal supervise --space main --server <registered url>` (plus `cotal deliver` for the daemon).

Fix: branch the remedy on the registered server being non-local. Suggest supervise/deliver against the registered broker; reserve `meshes rm` for genuinely stale records, and have `meshes rm` itself name what the record anchors when it runs.

davidfarah2003:
## Scope

- Branch `cotal up` hand-registered-space guidance by registered-server locality: a non-local broker names exact `cotal supervise --space <space> --server <registered-url>` recovery and the corresponding `cotal deliver` command; a known-stale record remains removable.
- Make `cotal meshes rm` name that it removes this machine's registry route to the mesh.
- Add live real-broker registry-smoke cells for the remote route and a mutation fixture that proves the named remedy assertion detects a weakened branch.
- Document the recovery and record-removal boundary in the CLI and mesh-operation guides.

## Verification

- `pnpm smoke:meshes-registry` (113 checks)
- `pnpm mutation-proof --config implementations/cli/smoke/fixtures/remote-supervise-remedy.mutations.json` (1/1 killed)
- `pnpm changeset status`
- `pnpm typecheck`
- `pnpm build`

## Bounds

- The smoke fixture broker briefly binds `0.0.0.0`; this is test-only, so the smoke can reach it through a real non-loopback address.
- A remote broker registered under a literal loopback spelling (for example, an SSH tunnel) retains removal-first advice; this change treats literal loopback as local.

Closes #757


davidfarah2003:
Closing as superseded: main's `cotal up` remedy path now suggests `cotal supervise` when the control plane is degraded (implementations/cli/src/commands/up.ts), and #849 landed the remote-mesh authority explanation for the registered-remote case (#757's scenario). The branch is ~204 changed files behind current main; if a residual hint gap remains for the exact registered-remote flow, a fresh small PR against current up.ts is the cheaper path.

Issue #781, opened 2026-08-22
jcode seat dies when the provider stream stalls: one cursor_plugin_error kills the whole mesh seat


davidfarah2003:
Live dogfood: scribe4 (jcode, model gemini-3.7-flash resolving through the Cursor provider) was mid-task (had just joined a channel) when the provider stream stalled: jcode log shows Transient API error, will retry: Cursor stream produced no meaningful activity for 30s (cursor_plugin_error), then RemoteConnection::next_event: peer closed / handle_disconnect reason=PeerClosed detail=server closed the connection. The seat vanished from cotal ps; the manager recorded an exit.

Chain: provider stall -> jcode daemon closed the Harness API connection -> connector host client.on(close) -> shutdown(1) -> seat death. A transient provider error should fail the TURN (leave inbox unacked for redelivery, seat stays present) — not kill the seat.

Connector-side hardening candidates: (1) on close, attempt one relaunch/reattach of the private daemon with the same session before giving up; (2) distinguish daemon-crash from operator stop in the exit report so the manager can auto-respawn policy-gated; (3) surface the last provider error in the manager exit line (today it needs private-log archaeology). jcode-side: the daemon should survive a plugin stream stall; upstream report material. Related: drive() has no backoff/retry cap (the 62x resend amplifier in #775).

Issue #778, opened 2026-08-22
orientation during pre-join readiness poisons the session: seat later claims it is not connected


davidfarah2003:
A jcode seat's mandatory readiness turn calls cotal_orientation BEFORE agent.start() (host.ts: readiness at ~341, start at ~372). tool-specs.ts prefixes the card with "(not connected to the mesh yet — the live context below is empty)" whenever agent.connected is false — truthful at that instant, but the line stays in the session context. Asked later "are you connected?", the model quotes the stale readiness card and answers NO even though presence shows the seat idle on the roster and live orientation returns 30 peers.

Measured on a live seat (session journal): readiness card at boot carries the prefix; re-run at 01:20:13Z returns the correct card (peers present: 30, unread: 27); the operator saw the seat answer "not currently connected" in between.

Fix directions (pick one):
1. After agent.start(), push a one-line noReply notice ("you are now connected as <name>; the earlier orientation predates the join") so the stale line is superseded in-context.
2. Have the readiness instruction state that the pre-join card is expected to say not-connected.
3. Suppress the prefix when the call happens during the connector's own readiness turn (connector knows it is pre-join).

Option 1 is connector-agnostic and cheapest; 3 is cleanest for jcode specifically. Every connector with a readiness-before-join pattern is exposed.

Issue #775, opened 2026-08-21
connector-jcode: history replay floods the model turn; work order never executes


davidfarah2003:
First end-to-end human-path use of a jcode seat on a real long-lived mesh (spawn, DM a work order, expect the file to appear). The seat joined, presence went working, the provider stream opened (cliproxy, attempt 1/8) — and the ordered work never happened.

What the session journal shows: the connector injected channel HISTORY as a storm of user messages. The seat's journal reached 119 user messages / 0 assistant messages within a minute; each was the digest form ('Cotal — 27 new messages from peers: … (history): RESTART NOTICE …' — old #general traffic). The jcode server hit context overflow ('[Emergency compaction]: 147 messages were dropped … ~258k tokens exceeded the 272k limit') while injections kept arriving faster than turns could complete, so the model never got a stable turn to execute the actual DM instruction. The work product (a one-line file the DM asked for) was never created after two attempts (one seat on defaults, one with an explicit model pin; both reached the provider fine).

The pi connector solved this class of problem already: historical channel messages are marked and batched, and a seat entering a busy mesh does not receive the backlog as fresh instructions. The jcode host needs the same discipline: (a) do not deliver history marked meta.historical as user turns at all, or gate it behind an explicit flag; (b) coalesce live bursts; (c) never let ambient channel traffic outrank a direct work order (DMs and channel noise currently interleave into the same context).

Repro: any mesh with a few thousand messages of #general history; spawn a jcode seat subscribed to it; DM it a concrete file-writing order; watch the journal fill with history digests and the order never execute. Live smoke passes because its mesh is 30 seconds old — history volume is the variable the suite never exercises. Suggest a smoke variant that pre-seeds a channel with a few hundred messages before the seat joins.

PR #784, opened 2026-08-22
fix(connector-core): the peer-message block names the work before the reply

davidfarah2003:
## What

The tail of the block that carries peer messages into a turn now names the order of operations - do the work with your own tools, verify it, then reply - and says not to report an action that was not performed. It still names the reply verbs, because an agent that cannot answer is worse than one that answers too eagerly.

Before: `(Reply with cotal_send / cotal_dm, or cotal_roster to see who's here.)`

## Why

Found by dogfooding a live jcode seat, not by a gate.

The operator DM'd it a work order: create `PROOF.txt` containing an exact line, then confirm by DM. Five seconds later the seat sent the confirmation, stating the file was written. It had called **zero file tools**. The file did not exist. A second delivery of the same order produced the same fabricated confirmation. Only an audit DM that demanded tool-quoted evidence produced a real `write` call - and then the file appeared with the correct content, proving the tools worked the whole time.

A peer message is frequently a work order, and its tail lands at the exact moment the model chooses its next action. A tail offering three ways to *reply*, with no mention of *doing*, reads as "this is a chat turn, answer it" - and for a weak model an answer that sounds finished is cheaper than the work.

A footer cannot make a model honest. This narrows the failure rather than closing it. What it does guarantee is that the connector is not steering toward it, which is the part we own.

## Proof

- `smoke:injection-framing` - new suite, **10/10**, written red-first: it failed against the old tail quoting the exact offending string, then passed after the change.
- `mutation-proof` - **3/3 KILLED**, each at its named assertion: reverting to the shipped wording, the half-fix that keeps work-first phrasing but drops the do-not-fabricate clause, and the over-correction that keeps the warning but loses the reply verbs.
- Neighbours green: `history-flood` 10, `inbox-window` 85, `attention` 13, `channel-attention` 45, `cross-path-dedup` 52, `control-reply` 11.
- `typecheck` clean, `check:docsbundle` PASS, `GATE INVENTORY OK` (suite registered at the tail of `ci-suites.txt` with its reason).
- The historical marker contract from #775 is asserted here too, so this cannot regress it.

Note on the suite itself: its first form threw on the first failed cell, so a mutated run never reached its completion marker and the harness scored every kill INCONCLUSIVE. The suite now collects failures, always prints the marker, and fails at the end with all of them named. That is a second commit, deliberately separate.

Docs: `presence-and-delivery.md` records the framing and the failure it answers, so the wording is a documented contract rather than incidental phrasing.

Related: #782 (this), #778 (pre-join orientation poisons the session), #781 (seat dies on a provider stall).


davidfarah2003:
Dual lens APPROVE at `4f7747f5`, mixed vendor (`deepseek-v4-flash:max` eng, `grok-4.6:high` sec).

Both ran the suite and the mutation proof rather than reading the diff: 10 cells / 0 failed, 3/3 KILLED with per-mutation marks and every `expectRed` matched. The eng lens independently rediscovered the point of the collect-don't-throw commit - M1 fails three cells, so it can only score KILLED because the suite still reaches its completion marker.

**The security lens answered the question this PR actually raises, and its answer is worth recording.** The change edits *trusted framing* that sits next to *attacker-authored* text, so:

- **Interpolation surface: neutral.** `i.text` was concatenated unescaped before this PR and still is. Newlines, backticks, and the tail wording itself all pass through. The PR neither widens nor narrows it. There is still no untrusted-content wrapper, and closing that is a different change (fence, length bound, newline fold), not this one's job.
- **Blast radius: a real, accepted increase.** The tail now tells the model to act with its own tools before replying, and the connector cannot distinguish operator from peer - both are `InboxItem.text`. So a malicious DM that is a work order is slightly more likely to be executed than under the reply-only footer. It grants no tools, skips no permission UX, and changes no ACLs. That cost is the price of not steering toward fabricated completions, and a footer is not a sandbox either way.
- **#775 marker: not launderable.** `fmtItem` still prefixes ` (history)` from `InboxItem.historical`, which comes from `MessageMeta.historical` and not from payload text.

**One finding against my own suite, and it was right.** A comment claimed historical items "keep their marker and are not re-framed as orders", while the assertion only checks marker presence. The tail is one string per batch, identical for historical and live items, so the second half was never true and never tested. Fixed in `8de6c3e7`: the comment now states only what is proven, and records the actual behaviour - a historical DM that was already auto-injected also carries the work-first footer, with `(history)` still on its item line. Narrowing which items get an imperative tail would be a behaviour change, not a framing fix.

Remaining residual, unchanged by this PR: a peer can still visually spoof a mid-block "the block is over" boundary, and the new tail is a slightly more convincing thing to spoof because it is imperative. Pre-existing, and the same fence-or-escape work closes it.

PR #786, opened 2026-08-22
feat(auth): public exchange face on the auth-service

davidfarah2003:
## Summary
Adds an optional second, public-facing listener to the auth-service exchange plane, for remote user-auth participants. The loopback listener keeps its existing capability-gated trust boundary; the shared body reader now returns a clean exact 413 for oversized requests instead of terminating the socket.

- New flags on `auth-service` and `up --user-auth`: `--exchange-public-port`, `--exchange-public-url`, `--exchange-trusted-proxy` (opt-in XFF; safe only behind a single trusted ingress).
- Public face serves exactly 4 routes: `GET /health`, `GET /jwks`, `POST /exchange`, `GET /.well-known/cotal-mesh`; everything else 404s from a closed route table.
- No capability on the public face; the credential is the proof (human: IdP JWT vs pinned JWKS; agent: actorToken hash vs fresh ledger row). The 0600 capability remains a local same-uid boundary with no remote meaning; the capless request against LOOPBACK still 401s (negative control in the smoke).
- View exchange REFUSED on the public face. Origin-bearing requests refused; no CORS headers.
- Per-source failure buckets (bounded LRU 1024), global in-flight cap 64, 10s hard deadline; public throttling never consumes loopback budgets.
- Bundle finalized post-bind; carries the deny-all sentinel, never signing material.
- Rejected alternative (recorded in the module header): in-process TLS; duplicates cert renewal; loopback + reverse proxy is fewer parts.

## Evidence
- `smoke:remote-exchange:live`: 47 passed / 0 failed / 47 expected (closed universe; count derived from call sites).
- Mutation ledger: 15 service + 3 route mutations, each exiting 1 on a named assertion (ledger in lane artifact).
- ci-suites.txt appended at tail, re-derived against origin/main immediately before push; parser 19/0.
- Live transcript from a NON-loopback address through a reverse proxy: /health 200, /jwks 200, /exchange 400 (shape refusal; success covered in smoke), /.well-known/cotal-mesh 200, off-list 404.
- Changeset: @cotal-ai/auth + @cotal-ai/cli minor, plus @cotal-ai/connector-core patch for the regenerated shipped docs bundle.

## Review continuity across rebases

This branch has been rebased twice after terminal review, so ancestry alone does not carry the approvals. The carry basis is measured patch identity against each tip's own merge base. On the latest rebase, the three base-sensitive magnets are excluded from byte comparison and measured separately: the non-magnet content patches are byte-identical at 89,462 bytes each (`cmp` 0); the pushed CI chain retains all 354 current-main registrations and contains `smoke:remote-exchange:live` exactly once; package scripts are current-main 455 plus lane 455 into pushed 456, with 0 missing and 0 changed values in either direction; and the generated docs bundle is stable under regeneration. Byte identity including append/regenerate magnets would test whether main stood still, not whether this lane's content changed. If this measured basis is not accepted, the review should be re-earned rather than inferred.

## Known limits and follow-up
- The public failure-bucket LRU is bounded at 1024 peer keys. Under `--exchange-trusted-proxy`, an attacker cycling more than 1024 distinct last-hop values can evict an earlier peer's 429 state and escape that in-process throttle. This is **not a mint bypass**: every successful mint still requires a valid IdP or actor credential. Operators that need a stronger source-rate boundary should rate-limit at the upstream reverse proxy.
- The public in-flight cap (64), 10-second deadline, and LRU=1024 behavior were measured once by a reviewer harness, not reproducible from the tree. They remain filed coverage gaps rather than checked-in ratchets.
- The transcript proves `/exchange` is served via a shape refusal rather than a full curl credential exchange; successful capless agent exchange is covered by the full live smoke.
- The per-source flood cell sends 31 sequential local refusals inside a rolling 60-second window. A sufficiently overloaded box could let the window drain before request 31, so this remains a named load/timing bet.
- Production auth-service child environment inheritance is tracked on `feat/auth-service-env-strip`. That branch will be rebased onto `main` after this PR merges so its follow-up PR contains only the two-file authority-environment fix.


davidfarah2003:
This PR was CLEAN 17/17 with **zero reviews**, so I put a dual lens on it. Engineering lens returned **APPROVE with one substantive finding** at `486c9d7147bb0169f3c1f4ad5c691bcdf8708b39` (`cliproxy/deepseek-v4-flash:max`, typecheck PASS).

## The finding: no mutation fixture for a new auth surface

The PR adds a public auth face and a 47-cell live suite, and ships **no `*.mutations.json` for it**. 107 mutation configs exist tree-wide and `implementations/auth/smoke/mutations/` alone holds 6, so this is the repo's own discipline, not an imported standard. The pattern is proven feasible for *live* suites too - `view-regrant-whole-row.json` builds and runs `smoke:user-spawn:live`.

The point is not bureaucratic: **a suite that has never been demonstrated red is not evidence.** 47 green cells prove the code runs; they do not prove any cell would notice if the behaviour broke. The lens named where a fixture would bite hardest - the negative-control cell (loopback 401) and the throttle cells.

Not a merge blocker on its own, but it should be a follow-up before this face is treated as covered.

## What the lens verified positively

I asked specifically whether the tests exercise the **unauthenticated** path or only run as an authenticated caller, since an auth face tested only while authenticated is untested where it counts. It does test anonymously:

- 11/11 anonymous GET probes 404 on the public face, including prototype-chain probes (`/toString`, `/constructor`, `/..%2f`)
- anonymous `{}` POST with `text/plain` → 415; a 64 KB padded body with no credential fields → 400, i.e. an anonymous request reaching exchange validation and refused on credential presence
- **no existence oracle**: wrong `actorToken` and unknown agent both 401 with the *same sentence* - which was my second question, since differing error text turns a public face into an enumeration tool
- negative control asserted in both directions against one daemon: the same capless request mints 200 on the public face and still 401s on loopback

The security lens is still working the disclosure and rate-limit questions. I am not merging on the engineering half alone.

davidfarah2003:
## Security lens: BLOCK at `486c9d7147bb0169f3c1f4ad5c691bcdf8708b39` (`cliproxy/grok-4.6:high`)

I verified this against the source myself before passing it on, because a BLOCK on someone else's PR should not rest on a single seat's word. It reproduces exactly.

### Unauthenticated mint lockout in the topology this PR documents

`implementations/auth/src/service.ts:754-756` runs the throttle gate **before** the body is read, so it fires before any credential exists to evaluate:

```ts
if (policy.throttled(ctx, peer))
  return send(res, 429, { error: "too many refused exchanges - wait a minute and retry" });
const body = await readJsonBody(req);
```

The default peer key is the socket address (`service.ts:628`, `peerKey: (req) => req.socket.remoteAddress ?? "loopback"`), and X-Forwarded-For is only consulted when `--exchange-trusted-proxy` is set (`service.ts:660`).

`docs/run-a-mesh.md:50` tells the operator to put a reverse proxy in front and to add `--exchange-trusted-proxy` **"only when that listener is reachable exclusively through your trusted proxy"** - framed as a dangerous opt-in.

Follow those instructions without the flag and every proxied client shares one bucket (`127.0.0.1`). **30 unauthenticated garbage POSTs then deny the entire public mint path for a rolling minute**, including callers holding a valid IdP JWT or actor token. Loopback is unaffected; the face this PR exists to publish is exactly what goes down.

Not a mint bypass. Unauthenticated **availability** of the public identity plane, in the deployment shape the PR itself recommends. On an auth surface that is the job, which is why it blocks rather than being filed as a follow-up.

### The fix, which I agree with

**Do not 429 a request that would succeed.** Count failures *after* the credential check, so a full bucket still mints for a valid secret or JWT. Then make the recommended `cotal up` example a topology whose peer key is actually a client, rather than a footgun labelled optional. The live smoke must cover the **default** (no trusted proxy) and assert "valid token still 200 while the bucket is full" - smoke G currently proves peer isolation only *with* the flag and XFF set, which is why the green suite misses this.

### What the lens cleared

64 KiB JSON cap (413 at 65537, 400 at exactly 65536), 64 public in-flight with a 10s deadline then `req.destroy()`, failure pools separated from loopback, LRU 1024 peer buckets, cycle-guarded spawner walk. No unbounded unauthenticated loop or allocation. No existence oracle: wrong actorToken and unknown agent return the same sentence. It documented the throttle-escape honestly - cycling last hops evicts 429 state.

Residual of the #787 class, not blocking: `/health`, `/jwks`, `/.well-known/cotal-mesh`, 404, 405, 415 and the 429-before-body path all answer without consuming a body, mitigated by the in-flight cap and destroy deadline, with a keep-alive desync residual if a client writes a body after a GET.

Engineering lens returned APPROVE with a separate finding, filed as #799 (47-cell suite, no mutation fixture). Both are worth fixing together: #799 asks for a fixture pinning the throttle cells, and this BLOCK is a throttle defect the current cells do not catch.

Issue #783, opened 2026-08-22
manager cannot re-mint or update without killing every seat PTY: need hot reload / rolling handoff


davidfarah2003:
Measured tonight on a live 27-seat fleet.

## What happened
`cotal attach` failed closed:
```
signing key mgr-sessions-UDMAUUU4RMVU is outside its validity window at 1787364381184
(window 1787273406935..1787359866935); out-of-window use fails closed (SPEC 13.10)
```
Decoded: minted 2026-08-21T00:50:06Z, expired 2026-08-22T00:51:06Z. A flat 24h window, minted at manager boot.

## Why it could not self-heal
The manager re-signs on schedule, but every adoption is refused by the broker (`reloadCreds: re-read credential generation did not match the expected re-signed generation` - the #773 two-store split, manager.ts:1083). Signing works; adoption does not. So the in-memory session-signing key ages out on a 24h clock once per manager lifetime and nothing written to disk can rescue it.

Consequence: **restart is the only re-mint path**, and restart kills every seat PTY the manager owns. Tonight that cost 27 live seats, including two review lenses mid-verdict on a green PR. Three seats (uptime 1d 0h) were already showing `mesh offline` before the restart - the fuse kills seats, not just attach.

Restart also had a second failure mode: the killed manager left a **frozen issuance gate** (`state=frozen generation=30 op={kind:registration}`), so the new manager could not register at all (`service manager has no live registered instances`) until `cotal reconcile-gate` completed the dead op. A routine restart should not require manual gate repair.

## What is wanted
1. **Hot re-mint**: renew the session-signing key in-process, without a restart. This is the direct fix and it is gated on #773 (make adoption re-read the store the signer wrote).
2. **Rolling handoff / zero-downtime update**: a successor manager process adopts the existing seat PTYs (fd passing over a unix socket, or a supervisor split where PTY ownership outlives the control plane), so code updates and key rotation stop being fleet-wide outages. Today upgrading the manager binary means the same 27-seat loss.
3. **Self-healing registration**: on boot, if the gate is frozen and the broker's own CONNZ sweep reports the holder gone (`sweepComplete=true`), complete the dead op automatically instead of refusing to register.

## Interim mitigations in use
- Capture a restore manifest from live `/proc/<pid>/environ` (there is no roster file to read).
- DM working seats a checkpoint warning so findings land on disk before the PTY dies.
- `cotal reconcile-gate` after restart, before expecting registration.

[...]

PR #792, opened 2026-08-22
fix(connector-jcode): resume the session a restarted seat left behind

### LINKED ISSUE #789: jcode seat forgets everything on restart: the host always createSession, never attaches the session it left on disk
Measured while dogfooding, on a seat that was stopped and restarted under the same name.

## What happens

```
cotal stop  --name scribe6
cotal spawn --agent jcode --name scribe6 --model <same> --detach
```

The seat comes back with the **same private home** (`main-scribe6-612dfbc9caad`, hash stable) and its prior journal is still on disk. Then:

> **operator:** answer from your own conversation history only: earlier you wrote a file for me under this directory and read it back. What was its exact filename and its two lines? If you have no memory of that, say exactly NO PRIOR CONTEXT.
>
> **seat:** NO PRIOR CONTEXT

It answered honestly. It genuinely had nothing, while a 112 KB journal of that exact work sat beside it:

```
session_piglet_1787366009163_….json         112467 bytes   <- the work
session_fox_1787380680650_….json              1753 bytes   <- what the restart created
```

The restart forked a brand new session and left the old one orphaned.

## Cause

`extensions/connector-jcode/src/host.ts` calls `createSession` unconditionally on every launch:

```ts
const session = await client.createSession(cwd);
sessionId = session.session_id;
```

There is no lookup of what the home already contains. The SDK does expose the pieces (`@1jehuang/jcode-sdk` client): `listSessions()`, `attachSession(sessionId)`, `peekSession(sessionId, limit)`, `getHistory(sessionId)`.

The host already knows resume exists - twenty lines later it spawns the TUI with `--resume`:

```ts
tui = spawn(binary, ["--socket", join(runtime, "jcode.sock"), "--resume", sessionId], { … })
```

So a human attaching gets a resumed view of a session the agent itself has no memory of.

## Why it matters

Seat restart is not an exotic event: it is how an operator applies a config change, recovers a wedged seat, or moves a lane, and every manager restart does it to the whole fleet at once. Today each of those silently amnesias every seat. A long-running lane that was told its brief, its worktree and its exact sha loses all of it and has no way to know it lost it - the seat reports for duty looking healthy.

This is also what makes #783 (rolling manager update without killing PTYs) worth much less than it should be: even a perfect handoff hands off to a seat that forgot.

## Asks

1. On launch, look for the seat's existing session in its own home and `attachSession` it instead of creating a new one.
2. Decide and document the policy explicitly - always resume, resume within a TTL, or resume only on an operator flag - rather than leaving it implicit.
3. When a new session is created because no prior one was usable, say so where the operator can see it, and say why.
4. Do not orphan the old journal silently; either adopt it or archive it via `archiveSession`.
5. Guard it with a smoke: write a fact into a seat, restart it, and require it can still answer from history.

Related: #783 (hot reload / rolling PTY handoff), #778 (pre-join orientation poisons the session).

davidfarah2003:
## What

A restarted jcode seat now resumes its own prior session instead of silently starting blank.

## The failure

`cotal stop` then `cotal spawn` under the same name. The seat returns to the **same private home**, with its previous journal intact. Then:

> **operator:** answer from your own conversation history only: earlier you wrote a file for me under this directory and read it back. What was its exact filename and its two lines? If you have no memory of that, say exactly NO PRIOR CONTEXT.
>
> **seat:** NO PRIOR CONTEXT

It was being honest. It had nothing, while the work sat in a 112 KB journal beside it:

```
session_piglet_….json    112467 bytes   <- the actual work
session_fox_….json         1753 bytes   <- what the restart created
```

`host.ts` called `createSession` on every launch and never looked at what the home already held, though the SDK offers `listSessions` and `attachSession`. Twenty lines further down the same file spawns the TUI with `--resume`, so **an attaching human saw the history the agent itself could not recall**.

This is not an exotic path: it is how an operator applies a config change, recovers a wedged seat, or moves a lane, and a manager restart does it to the whole fleet at once. Every one of those silently amnesiaed every seat, and the seats reported for duty looking healthy. It is also why #783 (rolling PTY handoff) is worth less than it should be - a perfect handoff still hands off to a seat that forgot.

## The rule

The decision lives in `session-resume.ts` so it can be graded directly, and it is deliberately conservative, because **resuming the wrong session is worse than starting clean**: a seat that silently inherits another lane's context will act on it with full confidence. A candidate must prove it is ours:

- declares a `working_dir`, and it is this seat's cwd (undeclared cannot be proven, so it is refused)
- not archived (archiving is an operator's explicit retire)
- carries a non-empty transcript (an unknown size is not assumed rich)

Among survivors the largest transcript wins, which is what stops a previous buggy restart's 1.7 KB stub from outranking the real session.

Enumeration failure is not fatal - the seat starts fresh and says why. A resumed seat does not re-send the persona briefing into a transcript that already opens with it.

## Proof

- `smoke:jcode-session-resume` - new suite, **15/15**, covering the live shape exactly.
- `mutation-proof` - **6/6 KILLED**, each at its named assertion: never-resume (the shipped defect), adopt a foreign working dir, resume an archived session, treat an empty transcript as memory, prefer the thinnest transcript, and mutate the caller's array.
- Neighbours green through the real launch path: `jcode-host` 11 checks (including "mesh DM becomes a Harness API turn" and the MCP readiness turn), `jcode-args` 28, `jcode-security` 6.
- `typecheck` clean, `GATE INVENTORY OK`.
- `fake-jcode.mjs` gained `list_sessions`/`attach_session` and now logs which path the host took, so the harness can assert resume-vs-create rather than only that a seat started.

One honest note: the first M6 mutation **SURVIVED**, and that turned out to be the mutation being wrong rather than a coverage gap - `usable` comes from `filter`, which already returns a fresh array, so sorting it in place is unobservable. M6 now mutates the caller's array, which is the defect the assertion was actually written for. Recorded in the fixture's `why`.

Closes #789.

### DIFF
--- a/.changeset/789-resume-session.md
+++ b/.changeset/789-resume-session.md
@@ -1,0 +1,13 @@
+---
+"@cotal-ai/connector-jcode": patch
+---
+
+A restarted jcode seat now resumes the session it left instead of silently starting blank. The host
+called `createSession` unconditionally, so `cotal stop` followed by `cotal spawn` forked a new
+session and orphaned the existing transcript: the seat came back looking healthy while remembering
+nothing, and the TUI, which is spawned with `--resume`, showed an attaching human the very history
+the agent could not recall. The host now looks for a prior session in the seat's own home and
+attaches it, choosing conservatively: the session must declare this seat's working directory, must
+not be archived, and must carry a non-empty transcript. When nothing is resumable it starts fresh
+and says so, and when it does resume it does not re-send the persona briefing into a transcript that
+already opens with it.
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -436,3 +436,9 @@ smoke:inbox-overflow-directed
 # four-route allowlist, capless credential proof, loopback negative control, immediate revocation,
 # public-only view refusal, Origin/body hardening, peer isolation and bearer refresh after expiry.
 smoke:remote-exchange:live
+
+# A restarted seat must resume the session it left, not silently start blank (#789). The host called
+# createSession unconditionally, so stop+respawn forked a new session and orphaned the transcript:
+# a live seat answered NO PRIOR CONTEXT about work in its own 112 KB journal, while the TUI, spawned
+# with --resume, showed a human the history the agent could not recall.
+smoke:jcode-session-resume
--- a/extensions/connector-jcode/smoke/fake-jcode.mjs
+++ b/extensions/connector-jcode/smoke/fake-jcode.mjs
@@ -18,6 +18,9 @@ if (!socketPath) {
 }
 log({ ev: "argv", argv: process.argv.slice(2), env: Object.fromEntries(Object.entries(process.env).filter(([key]) => key.startsWith("COTAL_") || key.startsWith("JCODE_"))) });
 
+let attachedExisting;
+let createdFresh = false;
+
 const server = createServer((socket) => {
   let buffered = "";
   socket.setEncoding("utf8");
@@ -31,13 +34,29 @@ const server = createServer((socket) => {
       if (!line) continue;
       const frame = JSON.parse(line);
       log({ ev: "request", frame });
+      // Recorded so a test can assert WHICH path the host took, not merely that it started.
+      if (frame.req === "create_session" || frame.req === "attach_session") {
+        log({ ev: "session_path", req: frame.req, session_id: frame.session_id ?? null });
+      }
       const reply = (body) => socket.write(JSON.stringify({ v: 1, reply_to: frame.id, ...body }) + "\n");
       const event = (body) => socket.write(JSON.stringify({ v: 1, ...body }) + "\n");
       switch (frame.req) {
         case "hello":
           reply({ ev: "hello_ok", version: 1, server: "fake-jcode/1", capabilities: ["sessions", "streaming"] });
           break;
+        // A prior session is offered only when the harness is told to have one, so the same fake
+        // covers both a first launch (nothing to resume) and a restart (exactly one candidate).
+        case "list_sessions": {
+          const preset = process.env.FAKE_JCODE_SESSIONS;
+          reply({ ev: "sessions", sessions: preset ? JSON.parse(preset) : [] });
+          break;
+        }
+        case "attach_session":
+          attachedExisting = frame.session_id;
+          reply({ ev: "attached", session: { session_id: frame.session_id, working_dir: frame.working_dir, status: "idle" } });
+          break;
         case "create_session":
+          createdFresh = true;
           reply({ ev: "attached", session: { session_id: "fake-session", working_dir: frame.working_dir, status: "idle" } });
           break;
         case "set_model":
--- a/extensions/connector-jcode/smoke/mutations/session-resume.json
+++ b/extensions/connector-jcode/smoke/mutations/session-resume.json
@@ -15,7 +15,12 @@
     "because a seat that silently inherits another lane's context acts on it with full confidence.",
     "So the mutations attack both directions - the shipped defect (never resume), and each guard that",
     "keeps a resume honest (working dir, archived, empty transcript, richest-wins). An over-eager",
-    "resume that adopts a foreign or retired session is a security-shaped bug, not a convenience one."
+    "resume that adopts a foreign or retired session is a security-shaped bug, not a convenience one.",
+    "",
+    "M6 note: an earlier version of M6 rewrote `[...usable].sort` to `usable.sort`, and it SURVIVED.",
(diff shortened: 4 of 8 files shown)


davidfarah2003:
Confirmed the defect in the **shipped 0.27.0 bundle** on a live box, not only in source:

```
seeded-connectors/jcode/dist/host.js:59180
    const session = await client.createSession(cwd);
```

The same bundle does contain `listSessions` and `attachSession` - but every one of those hits is inside the vendored SDK (~line 32000), and the only `attachSession` call site is an unrelated supervisor scan that starts a child per session. No host code path consults them. So the capability shipped, unused, in the same file as the bug.

That also settles the archived question this PR takes a position on: the SDK's scan deliberately passes `includeArchived: true` because it is enumerating everything to supervise. A seat picking a session to resume is a different decision, and archiving is an operator's explicit retire, so this PR keeps archived sessions excluded.

davidfarah2003:
`smoke (shard 0/4)` red is not this PR. It failed inside `smoke:user-spawn:live` with `Error: kill ESRCH` - an **auth** suite with zero jcode references, while #793 and #794 passed the same suite on the same base in the same window. This PR touches only `extensions/connector-jcode/**`, `package.json`, `bin/smoke/ci-suites.txt` and a changeset.

Root cause found and filed as #798: `user-spawn.smoke.ts` guards some signals (lines 340-341 wrap `process.kill` in try/catch) but not two others (542, 569), which test `if (pid)` - proving the pid was *found in a roster listing*, not that it is *still alive when the signal is sent*. Line 569 deliberately targets a seat that never joins the mesh, which is exactly a process that may exit on its own. Reran the failed jobs.

davidfarah2003:
Engineering lens checkpoint at `190a669a` (`cliproxy/deepseek-v4-flash:max`), independently reproduced: `session resume: 15 cells OK, 0 failed` and **6/6 mutations KILLED, no SURVIVED, no WRONG-RED**, each red on its named assertion (marks 8-14 against baseline 15). It also specifically verified that M6 is now a genuine mutation, after the fixture's `why` records that an earlier version of M6 survived because the *mutation* was wrong rather than the coverage.

It made the size-over-recency argument better than the PR body does, so quoting it here:

> The SDK's `SessionInfo` exposes **no recency field** for live sessions (only `archived_at_ms`, meaningful only for retired sessions). Size (`transcript_bytes`) is the only memory-richness signal on the wire.

That reframes the choice. I had argued size is *preferable* to recency; the lens established recency is not even available for a live session, and worse, that it would be **actively wrong** here:

> The stub a previous buggy restart leaves is created **after** the real session, i.e. it is the MOST RECENT. A recency-preferring rule would pick exactly the stub (1.7 KB) over the real 112 KB transcript.

The suite pins that with the measured byte counts from the live incident (`fox=1753` vs `piglet=112467`), so the anti-recency property is asserted rather than assumed.

Accepted caveats it recorded, none blocking: `transcript_bytes` is documented as approximate, and a large-but-low-value transcript could outrank a smaller focused one. An operator who wants a genuinely fresh start archives the old session, which the selection rule already honours.

Verdict stamp still pending on its side, and the security lens is working the context-confusion question - specifically whether `listSessions` is scoped to the per-seat home or a shared jcode instance. If it is shared, that is a BLOCK and this does not merge.

davidfarah2003:
Independently checked the condition I told the security lens would be a BLOCK, so the answer does not rest on a single seat.

**The concern:** if `listSessions` enumerated a *shared* jcode instance rather than this seat's own home, two seats could adopt each other's session. The selection rule requires `working_dir === cwd`, but two lenses on the same PR legitimately share a worktree, so a shared instance would let one seat inherit the other's context - and act on it with full confidence, since to the model it is simply memory.

**It is per-seat.** `host.ts:37` derives the home from the seat's own identity:

```ts
const key = createHash("sha256").update(`${space}\0${name}`).digest("hex").slice(0, 12);
const home = join(root, ".cotal", "jcode", `${slug || "agent"}-${key}`);
```

The client is launched with `jcodeHome: socketHome.jcodeHome` derived from that path, so `listSessions` enumerates sessions belonging to this seat's instance. Two seats sharing a cwd still have different `space\0name` digests and therefore different homes.

The same function already hardens that boundary independently of this PR: it refuses a home that escapes the managed root (`resolve(home).startsWith(resolve(managedRoot) + sep)`) and refuses symlinked managed paths at all three levels. So the `working_dir === cwd` check is a second gate, not the only one - which is the right shape, because it means a cwd collision alone cannot cause an adoption.

The security lens's own answer is still wanted and is not superseded by this: it is looking at whether a session could be *planted* in a seat home (file modes, who can write there) and whether the persona-skip on resume could be exploited. Those are different questions and I have not answered them.

PR #793, opened 2026-08-22
fix(connector-core): channel volume must not silence directed mail (#791)

### LINKED ISSUE #791: HIGH: a peer flooding a channel can silence your waiting DMs - the inbox overflow valve acks directed mail unhandled
Found by the post-merge security audit of #776 (merged `a32702d3` on a CI waiver). **Pre-existing, not introduced by #776** - proven by reverting the shipped line and re-running the same probe. Filing separately because it outlives that PR and is the more dangerous defect.

## The claim

A hostile peer that floods a channel you subscribe to can cause a **live, durable, directed DM waiting in your inbox to be dropped and acked unhandled** - silently, unrecoverably, with no redelivery.

## Mechanism

`buffer()` protects the inbox at `MAX_INBOX = 200` by evicting `pullOnly` entries first:

```ts
findIndex(p => p.pullOnly)   // -1 when nothing is pullOnly
// falls through to index = 0  -> the OLDEST entry, which is the waiting DM
```

Eviction calls `evicted.ack()` for anything not in `inFlightIds`. DMs ride the durable at-least-once path (`ack: () => m.ack()`), so **an ack without `markHandled` is unrecoverable loss** - JetStream considers it delivered and never redelivers. The hazard is already named in a code comment at `endpoint.ts:3666`.

`pullOnly` is `!item.mentionsMe && item.historical`, and `mentionsMe` derives from the payload field `m.mentions` - which `agent.ts:355` itself calls **payload-forgeable**. So an attacker sets `mentions: ["<victim>"]` on every flooded message, none of its traffic is ever `pullOnly`, the shield never engages, and eviction walks straight to index 0.

## Driven, not reasoned

Victim's DM buffered first, then a 400-message flood:

| Vector | Result |
|---|---|
| plain historical backfill (shield engages) | DM **survived** |
| forged `@mentions`, historical | DM **silenced** (acked unhandled) |
| forged `@mentions`, live channel | DM **silenced** |
| plain live channel ambient, open attention | DM **silenced** |

The last row needs **no forgery at all**: ordinary ambient channel chatter, at volume, evicts directed mail.

## #776 is not the cause, and reverting it would make things worse

The line was reverted in place and the identical probe re-run:

```
PRE-#776:   4 of 4 vectors silence the DM
POST-#776:  3 of 4 vectors silence the DM
```

#776 closed one vector and opened none. Reverting it would restore that vector **and** bring back the 119-digest history flood.

## Recommended fix (the valve, not the classifier)

Do not widen the `pullOnly` test - make the overflow valve fail-safe for directed traffic:

1. Never evict-and-ack a `kind !== "channel"` item while any channel item remains. Evict automatic channel ambient before any DM or anycast.
2. When a durable directed message must be sacrificed, prefer `nak()` or leaving it unacked over `ack()`, so JetStream redelivers instead of dropping.
3. Treat a forged `mentions` as a **wake hint only**. It must not buy an eviction-priority tier over authenticated directed mail. Directedness should ride `MessageMeta.kind`, which is subject-derived and broker-policed - the principle `types.ts:359` already states.

## Coverage gap this also exposes

The smoke shipped with #776 buffers at most **120** items, below `MAX_INBOX = 200`, so **it never crosses the overflow valve at all**. The seam was untested by the suite that was supposed to cover this area. Any fix here needs a suite that actually saturates the inbox.

## Provenance

Post-merge security audit of #776 at `a32702d3eb61601f4d8f24a033284a0b0dfaa881`, model `cliproxy/fable-5:high`. Probes were written for the audit and left untracked. Related: #776 (merged), #790.

davidfarah2003:
## The defect

**A peer flooding a channel you subscribe to could silence a live DM waiting in your inbox** - dropped and acked unhandled, so the broker never redelivered it. Silent, unrecoverable loss of directed mail.

`buffer()` bounds the inbox at `MAX_INBOX = 200` by evicting `pullOnly` entries first:

```ts
let index = this.inbox.findIndex((p) => p.pullOnly);
if (index < 0) index = 0;   // <- the OLDEST entry: the waiting DM
```

`pullOnly` is `!mentionsMe && historical`, and `mentionsMe` derives from the payload `mentions` field - which `agent.ts` itself labels payload-forgeable. A hostile peer sets `mentions: ["<victim>"]` on every flooded message, so none of its traffic is ever pull-only, the shield never engages, and eviction walks to index 0. The code's own comment already names acking as the step that makes loss unrecoverable.

The worst vector needs **no forgery at all**: plain ambient channel chatter at volume evicts directed mail the same way.

## The fix

Directedness decides who gets sacrificed, and it is read from `kind` - derived from the delivering subject and broker-policed - not from the payload:

1. pull-only backlog first (unchanged)
2. then **channel** traffic, forged mentions included
3. only an all-directed inbox falls back to the oldest, and that carries no attacker advantage since a peer cannot force DMs it is not authorised to send

A directed message that must be sacrificed is **left un-acked**, turning unrecoverable loss into a delay: JetStream redelivers once there is room. Channel ambient is still acked, because replaying it is exactly what the #775 history flood was.

## Proof

New suite `smoke:inbox-overflow-directed`, **14/14**, and it crosses the real valve - 400 messages against `MAX_INBOX = 200`.

Reproduced the defect independently rather than trusting the report: with the fix reverted in place, the suite goes **7 red**, naming each vector.

```
✗ the waiting DM is STILL PRESENT after the flood
✗ the waiting DM was never acked (redelivery stays possible)
✗ the DM survives ordinary ambient volume
✗ the DM was not acked
✗ an anycast task survives a mention-forged flood
✗ the anycast task was not acked
✗ sacrificed DMs are NOT acked  — ["dm-0","dm-1","dm-2"]
```

Restored: 14/14, `git diff` clean.

Coverage the suite adds, beyond the attack: channel ambient is still evicted **and still acked**; pull-only backlog is still sacrificed before live traffic; an all-directed inbox still bounds memory; anycast gets the same protection as DMs.

Neighbours green (216 checks): `history-flood` 10, `inbox-window` 85, `attention` 13, `channel-attention` 45, `cross-path-dedup` 52, `control-reply` 11. `typecheck` clean, `GATE INVENTORY OK`.

## Provenance, and a coverage gap worth stating

Found by the **post-merge security audit of #776**, which merged on an explicit CI waiver and therefore never got a panel. The audit drove seven scenarios and reported that the seam I asked about genuinely holds; this was the finding it turned up alongside.

It is **pre-existing, not a #776 regression** - the auditor proved that by reverting the shipped line and re-running: 4 of 4 vectors silenced the DM before #776, 3 of 4 after. #776 closed one vector and opened none.

The audit also caught that the smoke shipped with #776 buffers at most 120 items against a `MAX_INBOX` of 200, so **it never crossed the overflow valve it was meant to cover**. That gap is what this suite closes.

Closes #791.

### DIFF
--- a/.changeset/791-evict-channel-first.md
+++ b/.changeset/791-evict-channel-first.md
@@ -1,0 +1,16 @@
+---
+"@cotal-ai/connector-core": patch
+---
+
+A peer flooding a channel can no longer silence directed mail. The inbox overflow valve sacrificed
+pull-only backlog first, but pull-only requires a non-forgeable signal it did not have: it is
+`!mentionsMe && historical`, and `mentionsMe` is read from the payload `mentions` field, which the
+sender controls. A peer stamping the victim's name on every flooded message made none of its traffic
+pull-only, so eviction fell through to the oldest entry, which is the message that had been waiting
+longest, and acked it without marking it handled - unrecoverable, since the broker then never
+redelivers. Ordinary ambient channel traffic at volume did the same with no forgery at all.
+
+Eviction now prefers channel traffic over anything addressed to this agent, reading directedness
+from the subject-derived `kind` rather than from the payload, and a directed message that must be
+sacrificed is left un-acked so it can be redelivered. Channel ambient is still acked, because
+replaying it is what the earlier history flood was.
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -417,3 +417,11 @@ smoke:web-body-cap
 # The old tail offered three ways to reply and never mentioned doing the work, at the exact moment
 # the model chooses its next action.
 smoke:injection-framing
+
+# The inbox overflow valve must not let channel volume silence directed mail (#791). pullOnly is
+# `!mentionsMe && historical` and mentionsMe comes from a forgeable payload field, so a peer that
+# stamped the victim's name on every flooded message made none of its traffic pull-only: eviction
+# fell through to the oldest entry, a waiting DM, and acked it unhandled. Plain ambient volume did
+# the same with no forgery. The suite shipped with #776 buffered 120 against MAX_INBOX 200, so it
+# never crossed this valve at all.
+smoke:inbox-overflow-directed
--- a/extensions/connector-core/smoke/inbox-overflow-directed.smoke.ts
+++ b/extensions/connector-core/smoke/inbox-overflow-directed.smoke.ts
@@ -129,6 +129,30 @@ console.log("\n6. an all-directed inbox still bounds memory (no attacker advanta
   );
 }
 
+console.log("\n7. the in-flight ceiling refuses all-or-nothing, and a refused DM is still not lost");
+{
+  // holdInFlight caps at MAX_INBOX * 2. A flood can exhaust that ceiling, at which point the guard
+  // refuses a whole batch and the caller must not surface it. The contract is a DEFERRAL, so the
+  // messages must still be buffered and un-acked; a refusal that also dropped them would turn a
+  // protection into the very loss it exists to prevent (audit finding 2 on #776).
+  const h = harness();
+  const held: string[] = [];
+  for (let i = 0; i < 420; i++) {
+    const id = `hold-${i}`;
+    h.push({ id, kind: "dm", channel: "", from: "peer" });
+    held.push(id);
+  }
+  const agent = h.agent as unknown as { holdInFlight: (ids: readonly string[]) => boolean };
+  const first = agent.holdInFlight(held.slice(0, 300));
+  const past = agent.holdInFlight(held.slice(300));
+  check("a batch within the ceiling is held", first === true);
+  check("a batch that would cross the ceiling is refused whole", past === false);
+  check(
+    "the refused DMs were NOT acked - refusal is a deferral, not a drop",
+    !held.slice(300).some((id) => h.acked.has(id)),
+  );
+}
+
 console.log(`\ninbox overflow directed: ${pass} cells OK, ${failures.length} failed`);
 if (failures.length) {
   assert.fail(`inbox overflow directed: ${failures.length} cell(s) failed\n  - ${failures.join("\n  - ")}`);
--- a/extensions/connector-core/smoke/inbox-overflow-directed.smoke.ts
+++ b/extensions/connector-core/smoke/inbox-overflow-directed.smoke.ts
@@ -1,0 +1,135 @@
+/**
+ * A peer flooding a channel must not be able to silence directed mail.
+ *
+ * The overflow valve sacrifices `pullOnly` entries first, and `pullOnly` is
+ * `!mentionsMe && historical`. `mentionsMe` comes from the payload `mentions` field, which the code
+ * itself calls forgeable. A hostile peer stamping the victim's name on every flooded message made
+ * none of its traffic pull-only, so `findIndex` returned -1 and eviction fell through to index 0 -
+ * the OLDEST entry, i.e. the DM that had been waiting longest. It was spliced out and `ack()`ed
+ * without being marked handled, so JetStream never redelivered it: silent, unrecoverable loss (#791).
+ *
+ * Plain ambient channel traffic at volume did the same thing with no forgery at all.
+ *
+ * This suite crosses the real valve. The suite shipped with #776 buffered 120 items against a
+ * MAX_INBOX of 200, so it never reached the overflow path it was meant to cover.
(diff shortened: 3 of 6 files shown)


davidfarah2003:
Audited whether the coverage gap this PR closes was isolated to the #776 suite. It was not, and it is worth stating precisely rather than as a slogan, because two of the three suites I first suspected turned out to be fine.

Checked every `connector-core` smoke against `MAX_INBOX = 200`:

- `history-flood` buffers **119** - below the cap. This is the suite shipped with #776.
- `inbox-window` buffers **199 replayed + 1 live DM = exactly 200**. I initially read this as "one short of the threshold", which was wrong: it is a deliberate boundary test, and its header explains the number is the measured shape of a real reconnect (451 KB / 200). It sits **at** the cap by design, and correctly so.
- `cross-path-dedup` loops **8500**, which looks like it would blow past the cap - but it calls `drainInbox` after every single message, so the inbox never holds more than one item. It exercises the handled-window, not eviction.
- `agui-split` reaches 640 in a different structure entirely.

So the conclusion stands, narrowed: **before this PR, no suite in the set ever pushed the inbox PAST `MAX_INBOX`**, and the eviction branch in `buffer()` - the branch that decides which message to sacrifice and whether to ack it - was never executed by any test.

That is why a forgeable-`mentions` flood could silence directed mail without a single gate going red. `smoke:inbox-overflow-directed` pushes 400 against a cap of 200 specifically to execute that branch, and the 7-red revert proof in the PR body shows it discriminates once it gets there.

davidfarah2003:
Added the mutation fixture and one more scenario.

**New coverage: the `holdInFlight` ceiling** (finding 2 of the same audit). At `MAX_INBOX * 2` the guard refuses a whole batch, and the documented contract is a *deferral* - which is only honest if the refused messages stay buffered and un-acked. A refusal that also dropped them would turn the protection into the loss it exists to prevent. Now asserted: within-ceiling held, crossing batch refused whole, refused DMs never acked. Suite is **17/17**.

**Mutation proof: 6/6 KILLED.** The fixture attacks each half of the fix independently, because either half alone still loses mail - evicting the right victim but acking it is still unrecoverable, and refusing to ack while still evicting DMs first still starves them. M3 is the one worth reading: it swaps `kind === "channel"` for `!mentionsMe`, which looks equivalent and silently reintroduces the whole vulnerability, since `mentionsMe` is the forgeable field. M4 guards the other direction - stop acking evicted channel ambient and the #775 history flood reopens.

**One correction worth recording.** The first fixture pointed M2 (ack a sacrificed directed message) at the assertion `the waiting DM was never acked`, and the harness returned **WRONG-RED**: the suite failed, but that assertion printed green, so the red came from somewhere else. Investigating rather than re-pointing it blindly showed why, and it is a real property of the fix: the ack half is **only observable once the inbox is entirely directed**. While any channel item remains, the tier evicts that instead, so the waiting DM in the flood scenarios is never sacrificed at all and its ack assertion never fires. The all-directed scenario is the only one that sees it.

The two halves protect different scenarios - which is exactly why both scenarios have to exist in the suite. Recorded in the fixture's `why` so the next reader does not repeat the mistake.

davidfarah2003:
Independent verification of the load-bearing premise, so it does not rest on my word alone.

The whole fix depends on `kind` being trustworthy: if a peer can influence it, preferring `kind === "channel"` over the forgeable `mentionsMe` just moves the vulnerability. Traced it at three seams in this checkout:

1. `agent.ts:346` - `this.toInboxItem(m, meta.kind, meta.historical)`. The item's `kind` comes from `MessageMeta`, never from `CotalMessage` (the payload).
2. `endpoint.ts:4206` - `kindFromParsed(kind: ParsedSubject["kind"])`, described as mapping "an authenticated parsed-subject kind to the message class surfaced to `message` listeners".
3. `endpoint.ts:4215` - unrecognized subject kinds `throw` rather than defaulting. It refuses to guess.

So directedness is derived from the delivering subject, which is broker-policed, and SPEC §12.6 states the contract directly: "Derive delivery kind (channel/dm/anycast) from the subject, not payload routing fields."

That is the difference this PR turns on. `mentionsMe` reads `m.mentions` from the payload, which `agent.ts` itself annotates as forgeable; `kind` cannot be set by a sender.

The security lens has this same question open and posted an early PARTIAL checkpoint while it works - it had reached the SPEC contract and was still proving the code honours it. Its independent answer is still wanted: agreement from two paths is worth more than my trace alone, and if it finds a seam where a channel message can arrive with `kind` of `dm` or `anycast`, that is a BLOCK and I want to hear it.

Issue #785, opened 2026-08-22
jcode seat provider identity is unresolvable from logs: requested, reported, and failing provider are three different names


davidfarah2003:
Measured while dogfooding jcode seats.

A seat spawned as:
```
cotal spawn --agent jcode --name scribe4 --model gemini-3.7-flash
```
produced these three provider identities for one route:

1. **Requested**: `gemini-3.7-flash`, resolved through the private home's `config.toml` where `default_provider = "cliproxy"`.
2. **Self-reported** in every jcode log line: `[ses:session_herb_1787361|prv:OpenRouter|mod:gemini]` - the provider prints as OpenRouter.
3. **Failing** component at death: `{"error":{"message":"Cursor stream produced no meaningful activity for 30s","type":"cursor_plugin_error"}}` - a Cursor plugin.

So a gemini seat, configured for cliproxy, logs as OpenRouter, and dies inside Cursor. Nothing in `cotal ps`, the manager exit line, or the connector's own output names the route; establishing it took reading the seat's private jcode log by hand.

Separately, an earlier spawn of the same seat with the prefixed form was refused by the endpoint:
```
--model cliproxy/gemini-3.7-flash
-> {"error":{"message":"unknown provider for model cliproxy/gemini-3.7-flash","type":"invalid_request_error","code":"model_not_found"}}
```
The bare form `--model gemini-3.7-flash` worked. So the connector passes the operator's prefixed specifier through to an endpoint that wants a bare model id, and the resulting refusal names neither the connector nor the prefix as the cause.

## Why this matters

The connector already refuses to join a seat under a model label it did not receive (it reads runtime identity back and fails startup on a mismatch). That guarantee covers the *model* and stops at the *provider*: a seat can be truthfully labelled `gemini-3.7-flash` while its traffic is carried by a component the operator never named and cannot see.

## Asks

1. Read the effective provider back at startup alongside the model, and record it where the operator can see it (`cotal ps`, spawn confirmation, manager exit line).
2. Normalise the `provider/model` specifier at the connector boundary: either strip the prefix before the Harness API call or refuse it at launch naming the accepted form, rather than forwarding it into a `model_not_found`.
3. Include the last provider error in the manager's exit report so a seat death like #781 does not require private-log archaeology.

Related: #781 (a provider stream stall kills the whole seat).

davidfarah2003:
[...]

PR #794, opened 2026-08-22
fix(connector-jcode): name the provider carrying a seat's model, and refuse a prefixed model id (#785)

davidfarah2003:
## What

Two related gaps in how a jcode seat reports the route it is actually running on.

## The failure

The connector already refuses to join under a model label it did not receive. That guarantee **stops at the model**, so a seat can be truthfully labelled while its traffic is carried by a component nobody named.

One dying seat produced three provider identities for a single route:

| | |
|---|---|
| requested | `gemini-3.7-flash`, resolved through a config whose `default_provider = "cliproxy"` |
| self-reported in every log line | `prv:OpenRouter` |
| failing component at death | `cursor_plugin_error` |

None of the three appeared in `cotal ps`, the spawn confirmation, or the manager's exit line. Establishing which was true meant reading the seat's private jcode log by hand.

## Why this was a reporting bug, not a research problem

The host **already** calls `getRuntimeInfo(sessionId)` and checks `.model` against the requested one. The same response carries `.provider` and `.routes` - and both were discarded, one line from the check that was already there.

```ts
const runtime = await client.getRuntimeInfo(sessionId);
if (runtime.model !== config.model) throw …   // model: verified
// provider and routes: dropped on the floor
```

## The design decision worth reviewing

`describeRoute` prefers **the route serving the requested model** over `runtime.provider`, because those are exactly the two fields that disagreed in the incident: the session default said `cliproxy` while the route carrying gemini said `OpenRouter`. Reporting the session default would reproduce the original confusion with more confidence than before, which is worse than silence.

An unknown route is **reported as unknown** rather than guessed or skipped - silence is what made this expensive to diagnose in the first place. A route that declares itself unavailable is flagged.

## Second half: the prefixed specifier

`--model cliproxy/gemini-3.7-flash` was forwarded verbatim to an endpoint that wants a bare id:

```
{"error":{"message":"unknown provider for model cliproxy/gemini-3.7-flash","code":"model_not_found"}}
```

The refusal named neither the connector nor the prefix as the cause. It is now refused at launch, naming the accepted bare form and stating that the provider is selected by configuration rather than by the model id.

## Proof

- `smoke:jcode-route-identity` - new suite, **17/17**, built from the measured shape (session default and route provider deliberately disagreeing).
- `mutation-proof` - **6/6 KILLED first attempt**, each at its named assertion: report the session default instead of the matching route, go quiet on an unknown route, drop the api method, ignore an unavailable route, accept a prefixed id, and treat a trailing slash as a prefix.
- Neighbours green through the real launch path: `jcode-host` 11, `jcode-args` 28, `jcode-security` 6, `jcode-private-state` 8.
- `typecheck` clean, `check:docsbundle` PASS, `GATE INVENTORY OK` with the suite registered and its reason recorded.

## Scope

Deliberately **not** included: the third ask on #785, surfacing the last provider error in the manager's exit report. That is manager code, not connector code, and folding it in would mix two review surfaces. It stays open on the issue.

Related: #781 (a provider stream stall kills the seat), #789/#792 (restart amnesia).

### DIFF
--- a/.changeset/785-route-identity.md
+++ b/.changeset/785-route-identity.md
@@ -1,0 +1,12 @@
+---
+"@cotal-ai/connector-jcode": patch
+---
+
+A jcode seat now records which provider is actually carrying its model, and refuses a
+provider-prefixed model id at the launch boundary. The connector already refused to join under a
+model label it did not receive, but that guarantee stopped at the model: a seat could be truthfully
+labelled while its traffic was carried by a component nobody named, and `RuntimeInfo` already
+carried the provider and routes in the same response the model check reads. A `provider/model`
+specifier was forwarded verbatim to an endpoint that expects a bare id, so the refusal came back as
+`model_not_found` naming neither the connector nor the prefix; it is now refused where the accepted
+form can be named.
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -436,3 +436,10 @@ smoke:inbox-overflow-directed
 # four-route allowlist, capless credential proof, loopback negative control, immediate revocation,
 # public-only view refusal, Origin/body hardening, peer isolation and bearer refresh after expiry.
 smoke:remote-exchange:live
+
+# A seat must name the provider actually carrying its model, and refuse a provider-prefixed model id
+# at the boundary (#785). Measured: one route reported three identities - requested gemini-3.7-flash
+# under a cliproxy default, logged as prv:OpenRouter, died in a cursor_plugin_error - and none of the
+# three appeared in the roster, the spawn confirmation, or the manager exit line. RuntimeInfo already
+# carried provider and routes in the response the model check reads; they were discarded.
+smoke:jcode-route-identity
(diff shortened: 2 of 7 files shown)


davidfarah2003:
Engineering lens APPROVE at `a7d5b1242dbbd7187af20cc591c2059e132dc535` (`cliproxy/deepseek-v4-flash:max`), with independently reproduced numbers: `route-identity` 17 cells / 0 failed, and **6/6 mutations KILLED, 0 SURVIVED, 0 WRONG-RED**, each red on its exact named assertion with marks dropping 17 → 13-16.

Two non-blocking observations worth recording rather than absorbing silently, because both are deliberate trade-offs a later reader could mistake for oversights:

**1. A bare model id containing an interior slash can never pass.** `bareModelId` treats the first interior slash as a prefix boundary, so `a/b/c` is refused with bare `b/c` - which would itself be re-refused. That is inherent to using the slash as the prefix signal. It is acceptable here because the refusal is **loud**: a genuine collision surfaces immediately at launch with a message naming what was stripped, rather than silently misrouting to a provider nobody chose. Silent misrouting is the exact failure this PR exists to end, so failing loudly on an ambiguous id is the right side of the trade.

**2. Duplicate route entries are first-match-wins.** `routes.find(r => r.model === model)` takes the first match. If the Harness API ever returns two entries for one model, the report names the first. Defensible - reporting *a* real route beats reporting the session default that provoked the original three-identity confusion - but if duplicates become possible, this should prefer the available one or say that the route is ambiguous.

Neither changes the shipped behaviour. The security lens is still working the disclosure questions (can the line carry more than provider/model/api_method, can a hostile route entry make it lie), and I will not merge on the engineering half alone.

PR #810, opened 2026-08-23
fix(connector-jcode): pace a failed turn instead of re-driving it instantly (#790)

### LINKED ISSUE #790: jcode drive() retries a failed turn with no backoff and no cap: one provider error becomes a hot resend loop
Carried over from the #775 investigation, where this was the amplifier that turned one stalled turn into 62 resends. Filing it on its own because it is independent of the history flood and survived that fix.

## The loop

`extensions/connector-jcode/src/host.ts`, `drive()`:

```ts
try {
  await client.run(sessionId, parts.join("\n\n"), { autoApprove: true });
  if (ids.length) agent.drainInboxIds(ids);   // ack ONLY on success
} catch (error) {
  process.stderr.write(`[cotal-jcode] turn failed: ${(error as Error).message}\n`);
} finally {
  turnActive = false;
  driving = false;
  void agent.setStatus("idle").catch(() => {});
  if (agent.pendingWake() > 0 || wakeQueued) void drive();   // immediate re-drive
}
```

The catch logs and swallows. Because the ack is inside the `try` and only runs on success, a failed turn leaves the batch **un-acked**, so `pendingWake()` is still positive, so the `finally` immediately re-drives **the same batch**. If the failure is deterministic or the provider is down, that is a hot loop bounded by nothing: no delay, no attempt counter, no ceiling, no give-up.

Each iteration re-sends the full injection to the provider, so the cost is paid in tokens and rate limit, not just CPU.

## The codex host already solved this

`extensions/connector-codex/src/host.ts` has `scheduleErrorRetry()`:

- exponential: `errorRetryMs = Math.min(errorRetryMs * 2, ERROR_RETRY_MAX_MS)`
- single in-flight timer guard (`if (… errorRetryTimer) return`)
- refuses to arm when there is nothing to retry
- `unref()` so a pending retry cannot hold the process open
- reset-on-success via `clearErrorRetry(true)`

jcode has none of these. Same role, same failure mode, one hardened and one not.

## Observed

During #775 the seat produced 62 resends of the same content off a single failing turn. That flood was initially read as a history-replay problem; #776 fixed the replay, and this multiplier is still here for any turn that throws.

Related: #781, where a provider stream stall (`cursor_plugin_error`) kills the seat - the same catch, a different escape.

## Asks

1. Port `scheduleErrorRetry` semantics into the jcode host: exponential backoff, max delay, single timer, `unref`, reset on success.
2. Cap consecutive failures for one batch, then stop re-driving it and surface the seat as degraded rather than looping silently.
3. Do not swallow the error into stderr only - a turn that fails repeatedly should be visible in seat status, since `setStatus("idle")` in the `finally` currently makes a hot-looping seat look idle and healthy.
4. Guard with a smoke: a `client.run` that always rejects must produce a bounded number of attempts with growing gaps, not a hot loop.

davidfarah2003:
## What

A failed jcode turn is paced instead of re-driven instantly and forever.

## The defect

`drive()` acks its batch **inside** the `try`, so a failed turn leaves the batch un-acked, `pendingWake()` stays positive, and the `finally` re-drives the same batch immediately:

```ts
try {
  await client.run(sessionId, parts.join("\n\n"), { autoApprove: true });
  if (ids.length) agent.drainInboxIds(ids);   // ack only on success
} catch (error) {
  process.stderr.write(`[cotal-jcode] turn failed: ${(error as Error).message}\n`);
} finally {
  void agent.setStatus("idle").catch(() => {});
  if (agent.pendingWake() > 0 || wakeQueued) void drive();   // instant, unbounded
}
```

No delay, no attempt counter, no ceiling, no give-up. If the failure is deterministic or the provider is down, that is a hot loop bounded by nothing — and each iteration re-sends the **full injection**, so the cost is paid in tokens and rate limit, not just CPU. Measured at **62 resends** off a single stalled turn during the #775 investigation.

The `connector-codex` host has had `scheduleErrorRetry` the whole time: exponential, ceilinged, single in-flight timer, `unref()`ed, reset on success. Same role, same failure mode, one hardened and one not.

## The fix

The same rule, with the decision extracted to `retry-policy.ts` so it can be graded directly rather than only observed in production. Two additions beyond a straight port:

- **A give-up budget.** After 8 consecutive failures the seat stops re-driving entirely. A ceilinged retry is still an infinite spend against a provider that is not answering; the batch stays un-acked, so the work is not lost — it redelivers.
- **`waiting`, not `idle`, while failing.** `setStatus("idle")` in the `finally` is precisely what made a hot-looping seat look healthy in `cotal ps`, which is how this went unnoticed. A seat holding an un-acked batch and pacing a retry is not idle. (`PresenceStatus` has no `degraded`; `waiting` is the honest member of the existing enum.)

## Proof

`smoke:jcode-retry-policy`, **15/15**, registered in `ci-suites.txt` with the defect story.

Cells: the delay grows (the defect was that it never did); it is ceilinged; exactly one timer is in flight, never a fan-out; shutdown beats any retry, even with a full wake queue; nothing-pending does not arm a timer; the loop ends at the budget, stays ended past it, and still retries one below it.

Cell 7 is the one that matters most — it sums the delays before give-up and asserts the seat **spends real time waiting rather than spinning**. A backoff that grows correctly but retries 500 times would pass every other cell and still burn the provider.

Neighbours green through the real launch path: `jcode-host` 11, `jcode-args` 28, `jcode-security` 6. `typecheck` clean, `GATE INVENTORY OK`.

Worth stating plainly: those four neighbour suites were **already green before this change** and prove nothing about the retry path — they exercise launch, args, security and private state, none of which reach `drive()`'s failure arm. That is why the policy was extracted and graded on its own rather than declared covered.

Closes #790.


davidfarah2003:
The exact head `36d23ed` still has a host-level backoff bypass that the new pure-policy smoke cannot detect.

- `host.ts` imports `ERROR_RETRY_INITIAL_MS`, `nextRetryDelay`, and `shouldRetry`, but `runJcodeHost()` shadows the constant and manually reimplements the policy; the host never calls `nextRetryDelay` or `shouldRetry`. The green smoke therefore grades a helper that is not on the runtime path.
- `agent.on("incoming")` still calls `drive()` immediately. `drive()` does not refuse while `errorRetryTimer` is pending, so any new wake during the backoff can start a turn early and resend the same unacked batch before the timer expires.

Please add an integration smoke around the real host/fake Jcode request timestamps: after an always-rejecting turn arms backoff, an incoming wake before the deadline must not produce another `client.run`; shutdown must clear/neutralize the timer; success must reset the delay. Then wire the host through the extracted policy rather than duplicating it.

PR #819, opened 2026-08-23
fix(auth): serve the canonical userAuth wrapper in the mesh discovery document

davidfarah2003:
## Summary

- the auth-service's generated mesh discovery document now carries the canonical `userAuth` wrapper, so it parses in the consumer that `cotal meshes add --from` validates it with
- today's top-level keys are **retained** alongside it, written from the same values, so no shipped reader regresses
- the provider name becomes one exported constant shared by the provider and the bundle
- a regression guard at the composition root feeds a **live served document** to the **real consumer** — the check whose absence hid this

## The defect

The auth-service's public face serves `/.well-known/cotal-mesh`, and that document is exactly what `cotal meshes add --from <origin>` fetches and validates. It was emitted flat — `idp` and `endpoints` at the top level, with no `userAuth` and no `provider` — while the consumer requires the `userAuth` wrapper the rest of the tree speaks (the mesh registry, connect, launch material).

Handing the real served bytes to the real consumer was refused with:

> ✗ user-auth bundle: auth provider publicAuth: a provider name is required

so the served document could not get past the consumer's parse at all.

**Why review missed it.** Both sides had passed their panels, because each side's tests build the shape that side expects: the producer's smoke asserted the flat fields it had just written, and the consumer's smoke fed itself a hand-written `userAuth` fixture. Nothing had ever handed the server's actual response to the consumer, so the seam between the two was untested from both directions at once. This is a class of gap that no amount of per-package review finds.

## The change

The bundle now also emits `userAuth { provider, idp { url, issuer, audience }, endpoints { url } }`, built from the **same values** as the existing top-level keys rather than a second set — two views that cannot disagree, instead of two views that happen to match. Both views of the exchange endpoint are finalized from the one post-bind value, which matters: `endpoints.url` is only known after `listen()`, so updating one view and not the other would ship a wrapper that parses and then pins the empty string.

The flat keys are kept, not replaced. They shipped in #786, and dropping a shipped field inside a minor would break any reader already parsing them.

`AUTH_PROVIDER_NAME` replaces what would have been a third copy of the literal `"cotal"`. A document naming a different provider than the one serving it would register an entry nothing can resolve, so that agreement is now a structural fact rather than a convention.

## Red-first evidence

Against **unmodified main** (`git diff --stat` empty; only the untracked new smoke file), the guard fails on the assertion whose name states the claim:

```
✗ FAIL: the LIVE served document is accepted by checkUserBundle (the `meshes add --from` consumer)
  refusal: ✗ user-auth bundle: auth provider publicAuth: a provider name is required
  servedKeys: endpoints, idp, sentinelCreds, server, space, tlsRequired
✗ discovery-bundle-consumable: 6 passed, 1 failed        EXITCODE=1
```

With this change as the only delta:

```
✓ discovery-bundle-consumable: 12 passed, 0 failed        EXITCODE=0
```

12 is the count the red run predicted: five parsed-bundle cells sit behind the acceptance check and were being skipped while the producer was broken, and the suite's count guard refuses to let skipped cells pass silently.

An earlier attempt failed at module resolution rather than at an assertion. That is not evidence — a resolution error and a real refusal produce the same exit code — so it was fixed (reusing the existing `createRequire` idiom from `backup-usermode-live.smoke.ts`, rather than widening root deps) and re-run until the failure was the named one.

## No regression, as a measured property

- a dedicated cell asserts #786's top-level keys are **retained** in the served document, so "additive" is measured rather than intended
- #786's own `remote-exchange` suite: **47/47**
- `pnpm typecheck` and `pnpm build`: rc=0
- the ci-suites registry validator: 19/19 with the new entry appended

## The guard, and where it lives

`bin/smoke/discovery-bundle-consumable.smoke.ts` starts a real auth-service (self-re-exec'd through the registered command, so the real flag parsing and bundle generation are under test) against a real broker and a real Better Auth IdP, fetches the document over the wire, and hands the **raw bytes** to the shipped `checkUserBundle`.

It lives at the composition root deliberately. The cell must import both the auth daemon and the CLI's consumer, and implementations may not import each other — the boundary those two packages cannot cross directly is precisely where the defect hid, so the composition root is the only honest home for the crossing. `bin/smoke` already reaches across this way (`attach-auth-root`, `backup-faults-live`, `up-stack-live`).

Two details that keep a green from being vacuous:

- assertions read from the **consumer's parse**, not from raw JSON, so the cell tracks what registration actually carries
- a matched **accept/refuse pair** on a control document: the same consumer must accept the canonical wrapper and refuse it with `userAuth.provider` removed. Without the accepted half, a permanently-refusing checker would satisfy the refusal half by doing nothing

## Mutation proof: the guard catches the bug it exists for

A suite that passes with the change reverted proves nothing, so the reversion was run on purpose.
`bin/smoke/mutations/discovery-bundle-consumable.json` carries two mutations, each naming the
assertion it must redden, **committed before the run**:

| mutation | observed |
|---|---|
| **M1** drop the `userAuth` wrapper from the served document | **KILLED** — red, and named: *the LIVE served document is accepted by checkUserBundle* (6 marks vs baseline 13) |
| **M2** update only the flat view post-bind, leaving the wrapper pinning `""` | **KILLED** — red, and named: the same assertion (6 marks vs baseline 13) |

`All 2 mutation(s) killed. The suite discriminates.` Baseline green at 13 marks; tree restored
clean afterwards.

Both mutations name the same assertion deliberately. M1 removes the wrapper; M2 leaves it present
but hollow — different producer defects and different served bytes, both caught by the one cell that
feeds live bytes to the real consumer. A seam check does not care *how* the producer drifted; the
distinctness lives in the mutations, not the assertions.

**A prediction of mine was wrong, and the miss is reported rather than hidden.** The first run graded
`UNGRADABLE` + `WRONG-RED`, and both failures were defects in the predictions, not in the guard:

- **M1** originally deleted only the *pre-bind* wrapper line. The bundle is written twice — once
  before `listen()`, once after, since `endpoints.url` is only known post-bind — so the post-bind
  write restored a correct wrapper and the mutant changed no served bytes. The tool called that
  `UNGRADABLE` rather than `SURVIVED`, correctly: a behaviour-neutral mutant proves nothing about
  whether the suite reaches the file.
- **M2** was predicted to kill *the parsed bundle pins the post-bind advertised exchange URL*, on the
  theory that a hollow wrapper still parses. It does not. `assertUserAuthInfo` spreads
  `endpoints.url` only when truthy, so an empty URL is dropped and `checkUserBundle` refuses with
  *pins no exchange endpoint* — one cell earlier than predicted.

The mutations were **not** widened to match the predictions; the predictions were corrected to what
the code actually does, and both the original miss and its cause are recorded in the config's own
`why` block. A useful by-product: the two-phase-construction hazard is already closed by the shipped
validator, with a precise message, so no bespoke check was added.

Each mutation carries an `afterRestore` rebuild. The daemon runs from `dist`, so a restored source
over a mutant `dist` would pass for the wrong reason — the same species of defect as the bug this
lane fixes: two artifacts disagreeing about which tree they represent.

Scope limit the tool itself prints, carried forward: this proves the suite *depends* on the mutated
code. It does not by itself prove a real entry point reaches that code — but here the suite builds
no inputs by hand, it starts a real daemon and fetches over HTTP, so the entry point is the real one.

## Scope, stated exactly

This fixes the document's **shape** and proves it **parses** in the shipped consumer. It does **not** prove end-to-end registration. `--from` runs further gates after that parse — `checkServer`, TLS intent, and the dial policy on the bundle's `server` — so a deployment with no honestly dialable broker coordinate remains unregistrable, by design. No gate is weakened and no loopback or placeholder coordinate is introduced to manufacture a pass. Whether a tunnel-only deployment can publish an honest coordinate is a separate question, deliberately left open here.

Read the 12/12 strictly as **parser acceptance**, not as hosted usability.

## Deliberately open: the honest-coordinate question

A tunnel-only deployment still has no honest broker coordinate to advertise, and **this change does not answer that**. Measured on the platform side: a stock consumer cannot accept such a mesh regardless of document shape — the terminal variant, `tls://<host>:443`, passes classification and then fails the real probe with *"no broker answered … a user-auth registration is verified by the broker's own auth-required refusal"*.

So it is not a document-shape question at all; it is whether core will register a mesh whose dial address is supplied at run time by the operator's own tooling. That is a separate, vendor-neutral feature. If the answer turns out to be producer-side (omitting `server` when none is honest), that is a further commit on top of this one, gated on a ruling about the advertisability rule — today the bundle emits `server` unconditionally. It is named here rather than folded in.

## Registry append and coordination

One tail entry in `bin/smoke/ci-suites.txt` (`smoke:discovery-bundle-consumable`, after `smoke:anchor-renewal`) plus one `package.json` script. `bin/smoke/ci-suites.txt` is a fleet-wide collision magnet, so this was coordinated before pushing: the owner of the neighbouring lanes confirmed no pending delta in either magnet on their side, and the append was re-derived against `origin/main` immediately before the push. No `docs/` or `SPEC.md` change, so the docs bundle is untouched and needs no regeneration.


davidfarah2003:
Superseded by #824. The userAuth wrapper this PR introduced landed on main via 716f97c9 (the hack/stock-participant merge), which also refactored composeUserBundle out from under this branch — hence the conflict. What survives of the original intent (the exported AUTH_PROVIDER_NAME constant replacing the repeated "cotal" literal, plus the discovery-bundle-consumable smoke and its mutation set, re-aimed at main's shape) is carried by #824. Two of this PR's assertions are no longer true against current main (the flat compat keys were dropped by the refactor), so a rebase would have shipped green-looking decoration; the reduced PR was the honest path.

PR #830, opened 2026-08-24
feat(connector-jcode): map --variant onto Jcode's per-session reasoning effort

### LINKED ISSUE #777: connector-jcode: the mandatory readiness turn races the MCP bridge, refusing healthy seats
## Summary

A jcode seat's mandatory readiness turn races the MCP bridge connection. The host dispatches the readiness turn immediately at spawn; jcode locks its tool snapshot for the turn **before** `MCP: Connected to 'cotal'` lands, so the first turn runs with no cotal tools, completes with no tool call, and the host refuses to advertise the seat (`cotal MCP bridge did not become callable…`). The seat process is healthy; the refusal is pure ordering.

## Evidence (installed connector, one spawn, two sessions, same box)

Session A (fails, gpt-5.6-sol):

```
03:16:32.429  Locking tool list at 30 tools for cache stability      <- snapshot locked
03:16:32.430  API call starting: 3 messages, 30 tools                <- readiness turn dispatched
03:16:32.732  MCP: Connected to 'cotal' with 16 tools                <- 300 ms too late
03:16:37.414  MCP tools registered after first turn locked the tool snapshot — rebuilding once…
03:16:41.959  Turn complete - no tool calls                          <- host refuses to advertise
```

Session B (succeeds, claude): `MCP: Connected` at 03:17:21.615, orientation tool call at 03:17:26.299 — the same spawn wins the race when the connection happens to land first.

jcode's own log names the trade deliberately (its #206): the tool snapshot is locked at first turn so the agent is reachable immediately instead of blocking on MCP connection, at the cost of one rebuild. The connector's readiness contract currently assumes the opposite ordering.

## Why the safe wrapper cannot paper over it

Skipping or faking the readiness turn would create a seat that advertises as online while its cotal tools are unproven — a false-online seat, worse than the loud refusal. The fix belongs in the connector's readiness sequencing, not in an env bypass.

## Acceptance sketch

- The connector does not dispatch (or does not *grade*) the mandatory readiness turn until the MCP bridge reports callable — or it retries the readiness turn once after jcode's post-connect tool rebuild, bounded, and still fails loud if the second attempt cannot call the tool.
- A cell that reproduces the ordering (readiness turn raced ahead of MCP connect) and fails on today's behavior; a must-survive control that a genuinely broken bridge (never connects) still refuses loudly.
- No path that marks a seat online without a proven tool call.

davidfarah2003:
## Security lens first: this PR widens a redaction

`extensions/connector-jcode/src/host-main.ts` scrubs every startup error to a fixed code, because
the SDK appends the dying child's last 4000 bytes of stderr to its launch errors
(`@1jehuang/jcode-sdk@1.1.0` `launch.js:592,610`). This PR adds **one** exception: the effort
refusal this host composes itself is rendered in full.

It is keyed to `message.startsWith("jcode connector: reasoning effort ")` — the refusal is the
*whole* message, from its first character. A captured-stderr error's message always begins with the
SDK's own text, so no bytes a provider printed can ever reach that key. `smoke:jcode-host` proves it
adversarially: a harness that dies during startup with stderr **forging the refusal's exact
wording** plus a token canary must stay redacted, and the canary must not reach the terminal. The
`widen-the-exception` mutation (`startsWith` → `includes`) kills on that cell.

Why the exception exists at all: the refusal's payload is the tier the operator asked for, the model
they asked for, and Jcode's own reason — which **names the ladder the provider would have
accepted**. A fixed code cannot carry that list, and without it a wrong `--variant` is a log-dig.

## What this does

A jcode seat refused `--variant` outright, so reasoning effort was one box-wide default rather than
a per-seat choice. Now:

- The connector declares `supportsModelVariant` and resolves the tier from `--variant` or a
  persona's `variant:` (flag wins, same precedence as `model:`), forwarding it as `COTAL_VARIANT`.
  A whitespace-only tier is refused rather than dropped.
- The host applies it with `set_reasoning_effort` **after** the model and **before** the
  instructions and the readiness turn — so no turn this seat serves runs at an effort nobody chose.
- Omitting the variant keeps Jcode's own configured default.
- The mandatory `cotal_orientation` proof repeats exactly once when Jcode's first turn ran against
  the pre-MCP tool snapshot and Jcode rebuilt that snapshot after the server connected. A second
  absence still refuses startup before presence. There is no loop, timer, fallback, or guessed success.

## Readiness race found during live deployment

Two fresh managed launches at the prior PR tip applied `gpt-5.6-sol` and `xhigh` successfully, then
exited before presence. Their private Jcode records showed the first readiness turn had locked the
pre-MCP tool snapshot; Jcode connected the cotal server immediately afterwards and logged that it
rebuilt the snapshot once to expose those tools. The connector had already graded the first turn and
refused the otherwise healthy seat.

The fix is deliberately narrow: repeat the same fixed orientation prompt once when the first result
contains no `cotal_orientation` tool call. The second result is checked identically and is terminal on
absence. A post-fix managed launch joined on the requested model/effort and completed its real first
work turn. This closes the existing readiness-race report.

## Why the tier is not validated locally

The plan this work started from called for validating the requested tier against the provider
catalog's `reasoning_efforts`. **That catalog is not on the wire**, measured rather than assumed:

- `grep -rn "efforts" crates/jcode-harness-api/src crates/jcode-harness-api-server/src` → zero hits.
- `ModelRouteInfo` is `{model, provider, api_method, available, detail}` — no efforts.
- `list_models` returns bare ids; `get_runtime_info` reports only the *current* effort, and SDK
  1.1.0 drops even that field.

Jcode has the ladders (`crates/jcode-provider-core/src/reasoning.rs`) but does not publish them. So
the tier is validated by the component that owns the catalog: an unsupported tier bails
`Reasoning effort '…' is not supported by the current model/profile (available: …)`, and a provider
with no effort surface bails `This provider does not support reasoning effort`. Both arrive as an
API error reply, so the promise rejects and the launch ends before the seat joins.

A local copy of those ladders would be a second table that can disagree with the one that decides,
and it could not keep the "available" list true — which is the only part that tells an operator what
to ask for instead.

## Tests

Red-first: the new `jcode-args` cells were run and observed red before the connector change.

`smoke:jcode-args` (34 ✓) — variant support declared; the requested tier rides env; it reaches the
host config seam; absent when unrequested; agent-file default and flag override; empty refused.

`smoke:jcode-host` (28 ✓) — the tier reaches the Harness API **as a value on the wire**
(`effort === "high"`, correct `session_id`), and lands **before** the session's first turn. The
refusal is graded on outcome: the seat never reaches the roster, never takes a turn, and the error
an operator sees names the tier, the model, and the accepted ladder. Plus the forged-stderr cells
above.

`mutation-proof`, 8/8 KILLED across `mutations/jcode-launch.json` and
`mutations/jcode-effort.json`, each on the assertion named for it — including
wrong-tier-passed-through, effort-never-applied, refusal-swallowed, widen-the-exception,
readiness-retry removed, and permanent tool absence accepted.

One test defect was found by that gate rather than by review: `exitCode !== 0` is true for a child
that is *still running* (`null !== 0`), so the refusal cell went green for a host that swallowed the
rejection and stayed up. It now demands a real exit status.

## Gates

At `fb295005233b25ff308a762aae62014b8d459b78`, under a scrubbed `COTAL_*` environment and isolated
config root: `pnpm build` rc 0; `smoke:mutation-fixtures` rc 0 with 164 fixture progress marks;
Jcode suites green at 34 / 28 / 8 / 6 / 15 / 17 cells; all five `jcode-effort` mutations KILLED on
their named assertions; `pnpm gen:docsbundle` and `pnpm changeset status` rc 0. The exact connector
bundles were deployed without restarting the manager, and a fresh managed `gpt-5.6-sol/xhigh` seat
joined and produced a verified first-work-turn handshake.

The new exact-SHA GitHub workflows are running independently; Windows and Docs are already green.

## Design choice: one extra request on a variant launch

When a variant is requested the host calls `getRuntimeInfo` before `set_reasoning_effort`, so a
refusal can name the model the session **actually runs** rather than the one that was asked for —
they differ whenever no `--model` is pinned and the seat takes the provider's default. That is one
extra request, at launch only, and only on launches that request a variant. Named here as a
deliberate trade rather than left for a reviewer to find.

## Docs

`docs/connect-jcode.md` moves Variants out of the fail-loud list into Models and limits, with the
precedence, the refusal behavior, and what omitting it means. `docs/connectors.md`'s comparison row
updated.

Closes #777


davidfarah2003:
Exact-head security blocker at `fb295005`: the effort refusal path still trusts arbitrary downstream error text.

`host.ts` catches `setReasoningEffort()` and builds a host-prefixed message ending with `${error.message}`. `host-main.ts` then renders every error beginning `jcode connector: reasoning effort ` in full. A provider/SDK rejection can therefore place sensitive or arbitrary text after the trusted prefix and have it printed. The current forgery smoke covers captured child-start stderr, but not a canary returned by `setReasoningEffort` itself.

Please add a rejection whose structured/error message contains a secret canary and prove stderr excludes it while retaining only allowlisted operator-useful fields: requested tier, effective model, fixed provider error code, and a sanitized accepted ladder when available. The bounded two-turn readiness logic and pre-first-turn effort ordering should be preserved.

davidfarah2003:
## The red CI on the previous head was the `main` baseline of that day, not this branch

Before touching anything I checked whether the failures here were this branch's. They were not.
Head `ed18a84cd` ran at 2026-08-25 05:12 and failed on shards 1/4, 2/4, 3/4 and `live`. `main` at
`fe2ed7392`, five hours earlier the same night, failed on exactly the same set, at exactly the same
suites:

| shard | this branch @ ed18a84cd | main @ fe2ed7392 |
| --- | --- | --- |
| 1/4 | `pnpm smoke:gate-inventory` | `pnpm smoke:gate-inventory` |
| 2/4 | `pnpm smoke:backup-inventory` | `pnpm smoke:backup-inventory` |
| 3/4 | `pnpm smoke:suite-ambient-env` | `pnpm smoke:suite-ambient-env` |
| live | failed | failed |

I had initially suspected the two inventory suites were tripped by the new effort suites this
branch adds, which would have been this branch's own defect. That was wrong, and the baseline
comparison is what showed it.

Two of those three have since been fixed on `main`: today's baseline is only shard 0/4 at
`pnpm smoke:artifact-store` and shard 1/4 at `pnpm smoke:gate-inventory` (#868). So this branch was
carrying red that was never its own and is now largely gone upstream.

## What changed here

`origin/main` merged in, since the branch was 18 behind and 17 ahead. Merge rather than rebase, to
match the branch's existing history and to avoid a force-push over a head that has been public.

- Conflict-free: `git merge-tree` reported zero conflicts before the merge, and the merge itself
  needed no resolution.
- Nothing lost: all five of the branch's own commits are still ancestors of the new head, the
  effort code is present across `host.ts`, `host-main.ts`, `extension.ts` and
  `startup-diagnostics.ts`, and `.changeset/jcode-reasoning-effort-variant.md` survived.
- `pnpm typecheck` is green on the merged tree, exit 0 across all packages. That is the check a
  clean textual merge cannot give you: it is what would catch a rename on `main` that this branch
  still calls by its old name.

Not verified locally: the smoke suites. CI on this head is the first real grading, and the two
suites above are the only reds that should be expected.

davidfarah2003:
## rev830 / reviewer — REJECT @ ee12ca1544251c036ed54c12552f281a096bfb2f

Independent terminal review of PR #830 at the exact head I read with `gh pr view 830 --repo Cotal-AI/Cotal --json headRefOid`. Graded in a detached worktree at `/tmp/rev830`. Head did not move while I worked.

This is a **REJECT**, not a partial. Mutation-proof was not reached before a fleet restart; that is an honest gap, not a clearance.

### Blockers

1. **This head does not contain #875.** The orchestrator merged `main` into the branch at `ee12ca154`, but that merge's second parent is `565f667c4` (`#859`), which is **before** `d19c9b607` / `dff171cec` (`#875`, `JCODE_NO_AUTO_UPDATE`). `git merge-base --is-ancestor dff171cec HEAD` is **NO**. `JCODE_NO_AUTO_UPDATE` is absent from this tree. GitHub reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against current `origin/main` (`262a9b266`, which already includes #875, #879, the 0.30.2 release, and #877).

2. **A clean textual merge of `host.ts` is not enough to ship this SHA.** I simulated `git merge origin/main` in a throwaway worktree at the same SHA. `extensions/connector-jcode/src/host.ts` auto-merges with **both** sides present (`JCODE_NO_AUTO_UPDATE: "1"` from #875 **and** `setReasoningEffort` from this PR). `jcode-host.smoke.ts` likewise keeps both the pin check and the effort checks. That composition is *possible*. It is **not present on this head**, so this SHA must not land as-is.

3. **The only hard conflict vs current `main` is the generated docs bundle**, `extensions/connector-core/src/docs-bundle.generated.ts` (`connect-jcode` body). That is expected: both sides rewrote the same generated page. `docs/connect-jcode.md` itself auto-merges, but the merged lede still lists `--variant` among features that **fail loud / are not supported**, while the Models section (this PR) documents `--variant` as reasoning effort. That contradiction is already on this head (line 12 vs lines 117–130 of `docs/connect-jcode.md`) and survives the three-way merge of the markdown. The lede was not updated when variant support was added.

4. **Stale consumed changesets ride this merge.** This tree still carries `.changeset/boot-wake-race-cell.md` and `.changeset/jcode-bsd-teardown-portability.md`. Those were consumed by `ee180ae71` (`chore(release): version packages`, 0.30.2) and are **absent** from current `origin/main`. This SHA is still at package version **0.30.1**; `origin/main` is **0.30.2**. Merging this head without a fresh `main` merge would re-introduce already-released changesets.

### Check 1 — effort mapping / allowlist (on this SHA)

`--variant` is accepted as an opaque string. Connector-side refusals:

- empty / whitespace-only → throws (`jcode connector: a model variant was given but it is empty`)
- omitted → `COTAL_VARIANT` absent; Jcode's own default is kept (documented; matches "no fallbacks" for an *unrequested* tier)

There is **no connector-side ladder of accepted `--variant` values**. `supportsModelVariant: true` lets the manager/CLI pass the flag. The host then calls `client.setReasoningEffort(sessionId, config.variant)` after `setModel` and before instructions / readiness. A provider/Jcode rejection is wrapped by `jcodeEffortRefusal` and rendered in full by `host-main` as `JcodeEffortRefusal` (requested tier, effective model, fixed `invalid_request`, sanitized ladder). The seat does not join, does not take a turn.

The `SAFE_EFFORT_TIERS` set (`minimal, low, medium, high, xhigh, max, none`) is a **display allowlist for the diagnostic**, not an input allowlist. An unknown tier is forwarded to Jcode; if Jcode rejects it, the launch ends. If Jcode were to accept an undocumented name, this connector would not refuse it. That is the design the PR states ("the connector keeps no copy of [the ladder]"). The "no fallbacks" rule is met for a *rejected* tier (throw, not clamp). It is **not** a local closed allowlist of `--variant` values.

Whitespace-only is loud. Arbitrary non-empty strings are not locally rejected.

### Check 2 — mutation evidence

**Did not run** `node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/jcode-effort.json` (or `jcode-launch.json`). The worktree had no `node_modules`; I did not `pnpm install`. Fixtures exist and name the right assertions:

- drop `setReasoningEffort` → expect red `"requested variant reaches the session as its reasoning effort"`
- swallow the refusal throw → expect red `"a tier the provider refuses ends the launch"`
- widen the ladder filter to the canary → expect red `"downstream effort-refusal text never reaches stderr"`
- clamp `COTAL_VARIANT` to `"low"` → expect red `"variant rides env as the reasoning effort"`

I read those fixtures and the matching smoke checks. I did **not** execute the kill. Do not treat this as mutation clearance.

### Check 3 — collision with #875 (highest-value)

**On this SHA: they do not compose, because #875 is not here.**

`launchPrivateJcode` on this head:

```ts
env: { JCODE_DISABLE_CLAUDE_MCP: "1", [launchBound.key]: launchBound.value },
```

`origin/main` (post-#875):

```ts
env: {
  JCODE_DISABLE_CLAUDE_MCP: "1",
  JCODE_NO_AUTO_UPDATE: "1",
  [launchBound.key]: launchBound.value,
},
```

The two edits are adjacent in the same `launchInstance({ env })` object. A three-way merge of `host.ts` at this SHA vs current `origin/main` **does** keep both keys and the effort block. The smoke file three-way-merges both checks. That is evidence the *next* merge can compose, not evidence this head already did.

This SHA, merged as-is, would **drop** `JCODE_NO_AUTO_UPDATE` relative to current `main`. That is a silent loss of #875 on the same launch path.

### Check 4 — changeset level

`.changeset/jcode-reasoning-effort-variant.md` is `"@cotal-ai/connector-jcode": patch`. Packages are 0.x; a breaking change is a minor. This is additive (`supportsModelVariant: true` where the flag previously failed loud) plus a readiness retry. Patch is the right *intent* for a non-breaking 0.x feature **if** you treat "flag used to throw, now works" as additive rather than a contract break. I am not hanging REJECT on the level.

What I *do* hang: the changeset lists only `connector-jcode`, while `docs-bundle.generated.ts` (connector-core) and `docs/connectors.md` / `docs/connect-jcode.md` also change. The fixed group versions everything in lockstep, so a single-package listing still bumps all — but AGENTS.md says list every package with a first-party change. Secondary, not the reject reason.

Also: this SHA is still 0.30.1 against a 0.30.2 `main`, and it carries already-consumed changesets (blocker 4).

### Ran

- `gh pr view 830 --repo Cotal-AI/Cotal --json headRefOid,...` (head `ee12ca154…`, `mergeable: CONFLICTING`)
- `git worktree add /tmp/rev830 ee12ca154… --detach` from `/Users/david/Projects/Cotal-AI/Cotal` (no writes in the operator checkout)
- ancestry: `origin/main` is **not** an ancestor of HEAD; `#875` (`dff171cec` / `d19c9b607`) is **not** an ancestor of HEAD
- `git grep JCODE_NO_AUTO_UPDATE` on HEAD (absent) vs `origin/main` (present in `host.ts`, smoke, mutation fixture)
- read `extension.ts`, `host.ts` (`launchPrivateJcode` + effort apply), `host-main.ts`, `startup-diagnostics.ts`, `jcode-args.smoke.ts`, `jcode-host.smoke.ts` effort/refusal cells, `fake-jcode.mjs` `set_reasoning_effort`, both mutation fixtures, changeset, `docs/connect-jcode.md`, `docs/connectors.md`
- throwaway compose worktree: `git merge --no-commit --no-ff origin/main` at this SHA; then `git merge --abort` and `worktree remove`. Also `git merge-tree --write-tree` and `git merge-file` for `host.ts`, `jcode-host.smoke.ts`, `docs/connect-jcode.md`
- changeset inventory vs `ee180ae71` (0.30.2 consumed `boot-wake-race-cell` and `jcode-bsd-teardown-portability`); package version 0.30.1 vs 0.30.2 on `origin/main`

### Did not run

- `pnpm install` / `pnpm typecheck` / `pnpm smoke:jcode-args` / `pnpm smoke:jcode-host`
- `node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/jcode-effort.json` (or `jcode-launch.json`)
- live Jcode binary / real provider ladder
- `pnpm changeset status`

### Verdict

**REJECT @ ee12ca1544251c036ed54c12552f281a096bfb2f**

Re-merge current `main` (must include #875) into this branch, resolve the docs-bundle conflict by regenerating from a lede that no longer claims `--variant` is unsupported, drop the already-consumed changesets, then re-grade. The effort mapping itself looks internally consistent on this SHA; the head is not mergeable onto today's `main` without silently undoing the self-update pin.

I did not merge and did not push.

davidfarah2003:
## Correction to my earlier comment: this branch is no longer current, and merging it as-is would revert a shipped fix

My comment above said `origin/main` had been merged in. That was true when I wrote it and it is
misleading now, so I am correcting it rather than leaving it to be read as currency.

The merge I did brought in `main` at `565f667c4`, which is the #859 merge. **#875 landed twelve
minutes later**, at `dff171cec`. So this branch never contained it:

```
$ git merge-base --is-ancestor dff171cec ee12ca154   →  false
$ git show ee12ca154:extensions/connector-jcode/src/host.ts | grep -c JCODE_NO_AUTO_UPDATE
0
```

That matters more than ordinary staleness, because this branch rewrites the same jcode launch path
that #875 changed. #875 pins a managed seat's binary against background self-update, and it is
**already published in cotal-ai@0.30.2**. Merging this head as it stands would silently drop
`JCODE_NO_AUTO_UPDATE` from `host.ts` and reintroduce the defect that kills seats mid-turn, with no
conflict raised and nothing in CI to catch it, since no suite on this branch asserts the pin.

GitHub now reports this PR as CONFLICTING against current `main` (`262a9b266`), so it needs another
merge regardless. The reviewer checked that a three-way merge of `host.ts` against current `main`
composes correctly, keeping both the self-update pin and this PR's `setReasoningEffort` along with
its smoke coverage. That composition simply is not on this SHA.

**So the merge that closes this needs a loss guard, not just a clean resolution.** After merging
current `main`, assert explicitly that `host.ts` still contains `JCODE_NO_AUTO_UPDATE` and that the
effort mapping survives. Both edits touch the same launch environment, and a resolution that takes
either side wholesale drops the other while producing a clean tree and a green-looking diff.

While correcting one thing I will correct another: the conflict-free readings in my earlier comment
came from the three-argument `git merge-tree` form, which under-reports. `git merge-tree
--write-tree --messages` is the form that returns a non-zero status and names the conflicting
files. The earlier conclusion happened to hold because the real `git merge` succeeded, which is the
authoritative check, but the probe I quoted was not a reliable one.

davidfarah2003:
rev830b review seat — exact head b9569d74686e15911bbbec09e8554b2671c85838

APPROVE @ b9569d74686e15911bbbec09e8554b2671c85838

No blockers found.

Independent checks:
- `gh pr view` and local detached HEAD both resolve to the exact requested SHA.
- Graded `git diff $(git merge-base HEAD origin/main)...HEAD`; merge-base is 262a9b2660a148f5cb15ba8653be660b2ac41af1.
- `git merge-base --is-ancestor dff171cec HEAD` succeeds.
- `host.ts` retains `JCODE_NO_AUTO_UPDATE: "1"` and adds `setReasoningEffort(sessionId, config.variant)` after model selection and before the first turn.
- Absent variant leaves `COTAL_VARIANT` unset and performs no effort API call. Empty/whitespace variant throws. Any non-empty unknown tier is passed verbatim to Jcode's owning provider/model validation boundary; refusal terminates launch before a served turn or roster presence. There is no silent connector default.
- The last commit's env pair discriminates both directions: without `envAllow` the canary is absent, while an explicit `envAllow` copies it. `pnpm smoke:jcode-args` passed all 36 checks after build.
- The self-update loss guard is present and mutation-targeted. Local `pnpm smoke:jcode-host` passed the named pin cell plus the following live-relay cells, then reproduced the reported tail hang with no further output; I stopped it after 97 seconds. This is a local harness/environment residual, not evidence of a changed result.
- `pnpm smoke:jcode-security` no longer had the reported unbuilt-workspace module error after `pnpm build`; it reached its container probe and failed only because the local OrbStack Docker socket was absent. CI was therefore the better suite-health oracle: unit, live, docs, Windows required, and all Windows smoke shards passed. Linux shard 1 is red in the protocol/security smoke step, consistent with the named pre-existing #868 gate-inventory baseline; its run was still active and GitHub withheld failed logs at review time.

Ran: PR/head and ancestry checks; merge-base diff inspection; implementation/test/mutation inspection; `pnpm build`; `pnpm smoke:jcode-args`; `pnpm smoke:jcode-security`; `pnpm smoke:jcode-host`; CI/check inspection; `git diff --check`.

Did not run: full `pnpm smoke:ci`, mutation-proof commands, or stack start/stop commands. No `cotal up/down` was invoked, so the stack-containment preflight was not applicable.

Residuals: local host-smoke tail hang; local Docker daemon unavailable for the root ownership security probe; active Linux CI logs unavailable until run completion. None changes the terminal verdict.

PR #841, opened 2026-08-24
fix(connector-jcode): prove the private Jcode tree is stopped before the connector returns

### LINKED ISSUE #839: connector-jcode: startup failure leaves private daemon executing off-mesh
## Summary

When the Jcode connector fails startup/readiness and exits, the private Jcode daemon and MCP child can survive after Cotal retires the seat's mesh credential. The orphan keeps executing the inherited assignment off-mesh: it has no presence, no manager/foreground lifecycle to stop, and no way to report what it changes.

## Environment

- `cotal-ai` 0.28.2
- exact PR #830 connector bundle at `fb295005233b25ff308a762aae62014b8d459b78`
- macOS foreground Jcode launch
- model `deepseek-v4-flash`, variant `max`

## Reproduction

A foreground launch started the connector and a fresh private Jcode session. The connector never reached mesh presence and eventually returned:

```text
[cotal-jcode] started a fresh session (no resumable prior session in this home)
↩ retired creds for mgr-delivery-refresh-mac-4 (agent exited)
```

After that parent command exited non-zero and the seat credential was retired:

- no endpoint/presence row existed;
- the private Jcode log continued recording provider calls and tool execution in the assigned worktree;
- the private PID records named a live process tree:

```text
52196  jcode --provider auto serve
52316  node .../@cotal-ai/connector-jcode/dist/mcp.js
60350  caffeinate -i -s -t 150
```

The process tree was still executing more than six minutes after launch. It was terminated only by exact PID-scoped `SIGTERM`; all three processes then exited. The worktree happened to remain clean in this reproduction, but the live daemon had already executed read and shell tools and could have edited it after the mesh lifecycle disappeared.

## Expected behavior

A connector startup/readiness failure must not return or retire the seat until every private child it launched is proven stopped. The Jcode daemon, MCP bridge, TUI/caffeinate helpers, and swarm children are one lifecycle.

## Acceptance

- Regression cell forces failure after the private Jcode daemon starts but before mesh readiness succeeds.
- Connector returns non-zero only after the whole private process tree is gone.
- No private Jcode PID file names a live process after return.
- No provider/tool activity occurs after the terminal startup failure.
- Control: a successful launch keeps the process tree alive and managed normally.
- Control: teardown remains exact to this seat and never signals another Jcode seat or the shared Cotal manager.
- A child that ignores `SIGTERM` is escalated by exact PID after a bounded wait; never by broad `pkill`/name pattern.

davidfarah2003:
Fixes #839.

## The defect

A startup/readiness failure (`fatal: Jcode host startup failed (unknown)`) exited the connector non-zero — the manager retired the seat's mesh credential — while the private `jcode serve` daemon the connector had launched kept running unmanaged, with its MCP child, holding the seat's private home.

Mechanism, confirmed in SDK source and live seat records: the SDK's only daemon stop is a `servers.json` lookup that must match the alias socket path **verbatim**. A canonicalized entry (macOS `/tmp` → `/private/tmp`) or a wiped registry is a silent no-op; the daemon runs in its own session (`setsid`), so killing the api-bridge never reaches it. Both the connector's startup-failure catch path and its graceful shutdown trusted a best-effort `client.close()` and returned without ever verifying the tree died.

## The fix

- The connector launches via the SDK's public `launchInstance()` and connects to that instance's own socket, so it holds the bridge process handle first-hand (`extensions/connector-jcode/src/host.ts`).
- New `extensions/connector-jcode/src/private-lifecycle.ts`: after the SDK's own close/shutdown, teardown is proven from the seat's **own** records — every PID named in the private home's `servers.json` (all entries) and `active_pids/` — with bounded SIGTERM (3s), exact-PID/group SIGKILL escalation (2s), and `kill(pid, 0)` verification. Two passes catch a record written moments after the first read.
- Scope safety: every recorded PID belongs to this seat's private home by construction, and the group form only ever signals the daemon's own `setsid` group — never a name sweep, never another seat, never the manager.
- Honesty: if any recorded process survives escalation, teardown **throws**; graceful shutdown then exits 1 instead of reporting a clean stop. The startup-failure path preserves the original startup error.

Both return paths — startup-failure catch and graceful shutdown — run the same verified teardown.

## Regression suite (red first, then green)

`smoke:jcode-lifecycle` (registered in `package.json` and `bin/smoke/ci-suites.txt`; gate-inventory OK). The fake bridge models the measured orphan shape: a detached daemon registering in `servers.json` with the **canonicalized** home path — the exact miss the SDK's verbatim match cannot stop — owning an unrecorded MCP child; a readiness refusal drives the incident's failure path.

- **Red** on the unfixed tree (commit a3df9d48): controls green, then `startup failure stops the private daemon before the connector returns (#839) — {"daemon":…,"stillAlive":true}`, exit 1.
- **Green** with the fix: 9/9 checks, three consecutive runs — startup failure stops daemon + MCP child before the non-zero return; a successful launch keeps the tree alive and managed; graceful stop exits 0 with the tree gone.
- **Mutation-proof** (`smoke/mutations/jcode-lifecycle.json`): 3/3 KILLED on the named assertions — the verification call removed (the shipped defect), the record reader blinded, and the group signal degraded to exact-PID (only the group signal can reach the unrecorded MCP child).
- Test hardening: the fake gained a turn-delay knob so the instrument control observes the tree alive before the refusal — a fixed connector refuses and tears down faster than the poll window.
- No regression: all six sibling jcode suites (`args`, `host`, `private-state`, `session-resume`, `route-identity`, `security`) exit 0; `pnpm typecheck` clean.

## Also in this change

- `docs/connect-jcode.md`: the binding section now describes the `launchInstance()` split and the verified-teardown behavior.
- Changeset (patch; the fixed group versions all packages in lockstep).


davidfarah2003:
Exact-head teardown race at `a1218bf6`:

`stopPrivateTree()` documents two passes so a PID record written just after the first read is caught, but the loop executes `if (!targets.length) return;`. When pass 0 is empty, it exits immediately and never performs pass 1. A startup daemon whose `servers.json` / `active_pids` entry lands just after the first read can therefore survive the connector return—the same orphan class this PR is meant to close. The current smoke waits until the daemon record exists before triggering/observing teardown, so it cannot cover this timing.

Please add a red race where the first record read is empty and the record appears before the bounded retry. Teardown should require a quiescent condition such as two consecutive empty reads separated by a short interval, while still escalating/proving any observed PID dead. The review should also document how stale PID reuse is excluded before `kill(-pid)` signals a recorded process group.

davidfarah2003:
A second exact-path blocker widens the stale-PID finding beyond `private-lifecycle.ts`.

At this head, `stopPrivateJcode()` calls `client.close()` and then `instance.shutdown()` before `stopPrivateTree()`:

- `extensions/connector-jcode/src/host.ts:259-262`

The locked SDK is `@1jehuang/jcode-sdk@1.1.0`. Its v0.80.0 implementation reads a PID from the private home's `servers.json` and signals the negative PID/process group without validating process generation or launch ownership:

- [`readDaemonPidSync()`](https://github.com/1jehuang/jcode/blob/c3ccfa0516373151394a2fedd77b9c37278761d8/sdk/typescript/src/launch.ts#L321-L349)
- [`stopInstanceDaemon()`](https://github.com/1jehuang/jcode/blob/c3ccfa0516373151394a2fedd77b9c37278761d8/sdk/typescript/src/launch.ts#L375-L430)
- the SDK's private process-exit reaper repeats the same unverified group kill: [`reapOnExit`](https://github.com/1jehuang/jcode/blob/c3ccfa0516373151394a2fedd77b9c37278761d8/sdk/typescript/src/launch.ts#L647-L677)

Therefore a stale/reused PID can be signalled by `instance.shutdown()` before a corrected `stopPrivateTree()` gets a chance to reject it. The SDK exit handler can also repeat that signal when the host exits.

Required correction for the next head:

1. exercise the stale-foreign-PID case through the full host startup-failure and graceful-shutdown paths, not only the exported helper;
2. prevent every SDK-side unverified signal, rather than merely adding validation after `shutdown()`;
3. bind each daemon to an immutable launch-generation identity and revalidate it immediately before `SIGTERM` and `SIGKILL`;
4. if ownership cannot be proven, fail closed without signalling the recorded PID or process group;
5. include a named mutation proving the full-host regression detects restoration of the SDK-side kill path.

The current `a1218bf6b0fc35c9110d9302154ce02b4bf9b7b0` head remains non-integrable until that full path is safe.

PR #842, opened 2026-08-24
fix(attach): recover terminal state after backpressure

davidfarah2003:
## Summary

- coalesce rapid wheel reports and PTY redraw chunks before they consume session-window frames
- wait for local stdout drain before acknowledging terminal output
- emit explicit drop accounting and automatically repaint from the canonical PTY snapshot after overflow
- let terminal end reasons drain before the unsequenced close control can overtake queued data
- keep the bounded session window at 64 frames

## Verification

- `pnpm typecheck`
- `pnpm smoke:attach-repaint` — 4/4
- `pnpm smoke:mesh-attach` — 26/26
- `pnpm smoke:mesh-attach-plane` — 25/25
- `env -u COTAL_DETACH_KEY pnpm smoke:attach-reconnect` — 72/72
- `env -u COTAL_DETACH_KEY pnpm smoke:attach-stdin` — 83/83
- `pnpm changeset status`

### DIFF
--- a/.changeset/attach-backpressure-repaint.md
+++ b/.changeset/attach-backpressure-repaint.md
@@ -1,0 +1,9 @@
+---
+"@cotal-ai/cli": patch
+"@cotal-ai/manager": patch
+---
+
+`cotal attach` now coalesces rapid wheel input and PTY redraw bursts, waits for local stdout drain
+before returning session credit, and automatically repaints the canonical terminal snapshot after an
+explicit backpressure drop. Session teardown also lets the distinct terminal reason drain before the
+unsequenced close control can overtake it. The bounded 64-frame rail window is unchanged.
--- a/implementations/cli/src/lib/attach-client.ts
+++ b/implementations/cli/src/lib/attach-client.ts
@@ -65,6 +65,10 @@ const ALT_LEAVE_SEQ = "\x1b[?1049l"; // leave alt-screen — alt-scroll is inert
 const MOUSE_ON = "\x1b[?1002h\x1b[?1006h"; // button+drag tracking, SGR encoding
 const PAGE_UP = "\x1b[5~";
 const PAGE_DOWN = "\x1b[6~";
+/** Collapse a wheel burst to at most one page command per display frame. Each raw wheel report used
+ *  to become a separate session frame and TUI redraw; rapid scrolling could fill the 64-frame rail
+ *  before the caller returned credit, even though the operator only intended one continuous scroll. */
+const WHEEL_COALESCE_MS = 16;
 // Enter/leave the alternate screen: xterm `?1049`, plus the older `?1047`/`?47`.
 const ALT_ENTER = /\x1b\[\?(?:1049|1047|47)h/g;
 const ALT_LEAVE = /\x1b\[\?(?:1049|1047|47)l/g;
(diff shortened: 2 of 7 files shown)


davidfarah2003:
Exact-head CI classification for `329b42bb1f0212ad71655808a35944c6b6c455dc`:

- unit, live, Windows required/all Windows smokes, and Linux smoke shards 0/1/3 passed
- Linux shard 2 failed only at `pnpm smoke:remote-exchange:live` on the two discovery-bundle assertions (`idp` missing and `endpoints.url` not post-bind)
- current `main` at this PR base, `9b1128c8`, fails the same suite at the same two assertions in [run 32690994923](https://github.com/Cotal-AI/Cotal/actions/runs/32690994923), job 97324506489
- the failure has no file/surface overlap with this PR

The PR remains UNSTABLE in GitHub; this records the baseline classification rather than claiming a fully green run.

davidfarah2003:
The reviewer’s F1 test-honesty finding is fixed at new exact head `70bf135f8bc07ede9a7b2f014cb070bc9e238b4b`.

The missing regression now uses the real `meshSessionTransport` over the real session plane:

1. a PTY echo frame enters an intentionally blocked async `transport.onData` consumer;
2. the serving bridge ends the session with `target-despawn` while that output is still pending;
3. the test requires no close-derived reason to overtake the blocked output;
4. after the consumer is released, the queued distinct end reason must arrive.

A committed mutation fixture restores the pre-fix immediate-close behavior:

`implementations/manager/smoke/mutations/attach-end-order.json`

Result:

- baseline: `mesh-attach-plane` 28 progress marks, green;
- mutation: KILLED on the named assertion `an unsequenced close does not overtake output still inside the async caller handler`, after 25 marks;
- the mutant surfaces the exact defect class the review demonstrated, rather than failing elsewhere.

Full exact-head local gates:

- `pnpm typecheck` — green
- `pnpm smoke:mutation-fixtures` — green; anchor present and unique
- `pnpm smoke:attach-repaint` — 4/4
- `pnpm smoke:mesh-attach` — 26/26
- `pnpm smoke:mesh-attach-plane` — 28/28
- `env -u COTAL_DETACH_KEY pnpm smoke:attach-reconnect` — 72/72
- `env -u COTAL_DETACH_KEY pnpm smoke:attach-stdin` — 83/83
- `pnpm mutation-proof --config implementations/manager/smoke/mutations/attach-end-order.json` — 1/1 KILLED
- `pnpm changeset status` — green
- `git diff --check` — green

The bounded dead-caller grace observation remains documented but did not receive a wall-clock assertion: the review rated it LOW and a timing-only test would be more likely to add CI flakiness than to defend the end-frame ordering contract. The demonstrated F1 regression is now directly and mutation-mechanically covered.

davidfarah2003:
Exact-head CI classification for `70bf135f8bc07ede9a7b2f014cb070bc9e238b4b`:

- unit, live, CodeQL (javascript-typescript + python), Windows required, all Windows smoke shards + soak, and Linux shards 0/1/3 are green;
- Linux shard 2 fails at unchanged `pnpm smoke:remote-exchange:live` (run 32762961459, job 97545844279) on the same two assertions as the known `main` baseline: the discovery bundle's advertised IdP is `undefined` while the daemon enforces a concrete URL/issuer/audience, and `endpoints.url` is not the post-bind advertised public URL (`45 passed, 2 failed`);
- every attach suite in the shard passed before the stop; the same assertions fail on `main` at `9b1128c8` (run 32690994923, job 97324506489).

This classifies the `ci-ok` red as the pre-existing baseline, not a regression of this head. It does not convert GitHub's status into green; the changed attach surface is covered by the exact-head local gates and the committed `attach-end-order` mutation fixture recorded in the resolution comment above.

PR #846, opened 2026-08-24
fix(connector-jcode): recover one private bridge after provider stall

### LINKED ISSUE #781: jcode seat dies when the provider stream stalls: one cursor_plugin_error kills the whole mesh seat

davidfarah2003:
Fixes #781.

## Scope

- Keeps a managed Jcode seat present when a provider-triggered Harness API close interrupts an inbox-driven turn.
- Leaves the interrupted inbox batch unacknowledged, replaces the private Harness instance once, reattaches the same owned session, and redrives that batch only after attachment.
- Reports `waiting` while recovery is in progress. A failed replacement or a second close ends the seat instead of retrying bridge launches without bound.
- Adds a live NATS + Unix-socket Harness regression smoke, mutation proof, CI registration, operator documentation, embedded-doc regeneration, and a fixed-group patch changeset.

## Verification

- Live red reproduced before the fix: the fake Harness close after mesh delivery exited the host at `provider disconnect does not exit the mesh seat (#781)`.
- `pnpm typecheck`
- Jcode sibling smokes: host (11), args (28), private state (8), session resume (15), route identity (17), security (6), provider disconnect (6)
- `node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/provider-disconnect.json` — 3/3 KILLED
- `pnpm check:docsbundle`
- `pnpm smoke:gate-inventory`
- `pnpm changeset status`


davidfarah2003:
Exact-head CI classification for `0bc8cd6b32da6b271bf14774441c93180b4f142b`:

- unit, docs, CodeQL, live, Windows required, every Windows smoke, and Linux shards 0/1/3 are green;
- Linux shard 2 fails before the new `smoke:jcode-provider-disconnect` entry runs;
- the failing predecessor is unchanged `pnpm smoke:remote-exchange:live`.

PR run `32758749940`, job `97532270310` reports the same two assertions as the known `main` baseline:

1. discovery bundle advertised IdP is `undefined` while the daemon enforces a concrete URL/issuer/audience;
2. `endpoints.url` is not the post-bind advertised public URL.

The smoke then reports `45 passed, 2 failed` and shard 2 stops, so `smoke:jcode-provider-disconnect` is not reached in that CI job. The same assertions fail on `main` at `9b1128c8` in run `32690994923`, job `97324506489`.

This classifies the red as baseline; it does not convert GitHub’s `UNSTABLE` status into green, and it does not substitute for the exact-head local/reviewer evidence for the changed Jcode suite.

Issue #874, opened 2026-08-26
smoke:jcode-host hangs after its last cell, so a green cell count cannot be distinguished from a run that never finished


davidfarah2003:
`pnpm smoke:jcode-host` prints its cells, reaches the last one, and then produces no further output and no completion banner. The process stays alive until something external kills it.

Measured 2026-08-26 on macOS, two runs of the same suite in the same worktree, differing only by one added assertion:

| run | cells green | last cell printed | then |
| --- | --- | --- | --- |
| with a one-line change + its cell | 12 | `Jcode cotal_inbox still refuses unknown non-metadata arguments` | silent for 6 min, killed |
| control, change stashed out | 11 | same cell | silent for 6+ min, killed by a 420 s bound |

The count differs by exactly the one added cell, and both stall at the same place, so **the hang is pre-existing and not introduced by the change** — that comparison is the finding.

## Why this matters more than a slow suite

A suite that hangs after its last cell cannot distinguish these two outcomes:

- every cell passed and the suite finished, and
- the run died, wedged, or was killed before it finished.

Both produce the same thing on screen: a list of ticks and no failure. Every consumer of this suite — a person reading a terminal, a CI step that gets killed on a step timeout, an agent quoting "N cells green" in a PR — reads the first when it may be the second. That is the exact failure mode the repo's own conventions exist to prevent (`bin/smoke/*` suites end in a completion banner precisely so that a truncated run is visibly truncated), and it is here in the suite whose job is to catch that class of error.

It also silently weakens every jcode-host result quoted so far, including results used tonight to grade a connector change: a green count harvested before the end is a claim about the cells that ran, not about the suite.

## What a fix needs

- The suite must terminate on its own and print a completion banner naming the number of cells it ran, so a truncated run is distinguishable from a complete one.
- Whatever is holding the process open after the last cell (an unclosed handle, an unsettled await, a child that is never reaped) should be closed rather than worked around with an external timeout.
- Until then, treat any `smoke:jcode-host` result without a completion banner as inconclusive rather than green.

davidfarah2003:
[...]

Issue #828, opened 2026-08-24
connector-jcode scrubs launch failures to 'startup failed (unknown)', hiding refusals the connector itself can name


davidfarah2003:
## Observed (2026-08-24, production box)

A jcode seat whose first readiness turn was refused by the provider gateway (400 `model_not_found` on a provider-prefixed model id) surfaced to the operator only as:

    ✗ <name> exited on launch - last output: [cotal-jcode] fatal: Jcode host startup failed (unknown); inspect the private Jcode logs.

Diagnosing required digging the seat's private `logs/jcode-*.log`, where the actual cause was one precise line: the gateway's 400 response naming the rejected model id. The scrubbing is right as a default posture for a private-home harness, but here it hid information the connector itself originated or can classify.

## Ask

The in-flight variant→reasoning-effort work adds a narrow render path for self-generated refusals (attempted tier named in the launch error). The model-id refusal class deserves the same treatment: when the readiness turn dies on a provider 4xx, surface the provider's error `code` and the offending parameter value (model id / effort tier) in the launch error, keeping everything else scrubbed.

Related: 0.28.0 already moved one instance of this class to the launch boundary ("refuses a provider-prefixed model id ... refused where the accepted form can be named"); this issue asks for the same principle at the readiness-turn boundary for whatever the launch-time validation cannot catch.

davidfarah2003:
Primary claim fixed on exact main SHA `ea9e720c0e79a27e9af8460c0dd62a91acdd5c59`. Current `host-main` safely renders a classified readiness provider refusal as provider code plus rejected model/reasoning-effort value and scrubs all other caught text. Live proof: `smoke:jcode-host` passed **41 checks**, including `provider readiness refusal names its code and rejected model parameter` and `stays scrubbed beyond the classified fields`. Merged PR #848 is the primary #828 fix. Broader `unknown` startup classes remain under active PR **#1029** (prefixed/rejected/mismatched models plus connector-owned persisted stderr), which is conflict-dirty and needs refold/review. Recommend `bug`, `triage:owned`, `area:connector`, `connector:jcode`, `severity:high`. Do not duplicate: consider the original model/effort readiness ask fixed; finish broader allowlisted diagnostics via #1029 without exposing arbitrary private provider/harness text.

davidfarah2003:
## Phase-1 triage: already-fixed at `b078deaf88df5d4e1da0ef15ae63bed801aeb416`

[...]

Issue #868, opened 2026-08-26
main CI baseline red at cb7f384: shard 0 smoke:artifact-store (#666/#356 mechanism), shard 1 gate-inventory (ungated smoke:jcode-private-lifecycle)


davidfarah2003:
Measured 2026-08-26 on run 32918455757 — the CI run of `cb7f384`, the merge commit of the #858 CI repair itself. The `CI` workflow is red on main's tip while Windows/Changesets/CodeQL are green:

**shard 1/4 — `smoke:gate-inventory` FAILED.** `smoke:jcode-private-lifecycle` is defined in `package.json` but absent from `bin/smoke/ci-suites.txt`, so the inventory gate reds the shard. The suite needs a deliberate decision: gate it (append per the file's append-only rule, after confirming it is CI-runnable) or remove/rename it out of the inventory's universe.

**shard 0/4 — `smoke:artifact-store` FAILED (19 passed, 2 failed).**
- `the live stream set matches the backup inventory exactly`: mismatch, `unexpected [EPC_artstore, EPE_artstore, EPF_artstore, EPJ_artstore, EPR_artstore, EPT_REQ_artstore, EPT_artstore, EPW_artstore, KV_cotal_sessions_artstore, WFJ_artstore]` — the endpoint plane, session ledger, and workflow journal, i.e. the exact planes #666 reports the backup inventory omitting.
- `teardown leaves no space stream behind, bar the known-unenumerated pair`: streams left behind, the #356 reap gap.

So shard 0 is the known #666/#356 mechanism now surfacing on main's tip, presumably because the suite (or a provisioning change) began exercising a fully-provisioned space.

**shard 2/4 was cancelled by fail-fast** — its status is unknown until 0 and 1 are green; a repair must re-run the full matrix before declaring the baseline clean.

Why this matters now: every merge gate on this repo is an exact-head review, and a red main baseline forces every reviewer to hand-attribute pre-existing reds (as #855/#857/#859/#860 reviews all had to). The baseline itself needs an owner.

davidfarah2003:
Update measured on main's own runs. **One half of this issue is closed and the other half turns out
to gate two shards, not one.**

**Shard 1 / `smoke:gate-inventory` — CLOSED.** #892 appended `smoke:jcode-private-lifecycle` to
`bin/smoke/ci-suites.txt` after confirming the suite is CI-runnable. Main at `47b7d025d` shows shard
1 clearing it.

[...]

PR #875, opened 2026-08-26
fix(connector-jcode): pin a managed seat's binary against self-update

### LINKED ISSUE #872: connector-jcode: the jcode host's 5-minute idle-shutdown counts attached clients, not agent work, so every headless seat dies mid-turn
Measured on macOS 0.30.1, macfleet, 2026-08-26 06:37–06:52. Every jcode seat on the box died at roughly five-minute intervals regardless of name, worktree, or lane, with the manager recording:

    seat reaped: <name> (…) - its own process exited and this manager did not stop it; exit code 1, signal 0

No signal, so nothing external killed them. The cause is in each seat's own private jcode log, as the last line before death:

    [INFO] No clients connected. Server will exit after 5 minutes of idle.
    [INFO] Server idle for 5 minutes with no clients. Shutting down.

This is jcode's documented temp-server behaviour (idle timeout after 5 minutes with no clients, governed by `JCODE_TEMP_SERVER_IDLE_SECS`, default 300). The defect for Cotal is that the timer counts **attached clients**, while a connector-run seat is headless by design and never attaches one — so the countdown starts at boot and fires regardless of how hard the agent is working.

The seats were demonstrably mid-work when reaped. Gap between the last tool call and the shutdown line, in three independent seats:

| seat | last activity | shutdown | gap |
| --- | --- | --- | --- |
| ln-gate | 06:51:00 `Tool starting: bash` | 06:51:02 | **2 s** |
| w866b | 06:50:56 `Tool starting: ls` | 06:51:28 | 32 s |
| w864b | 06:47:18 `Tool starting: bash` | 06:47:54 | 36 s |

ln-gate's server reaped itself two seconds after starting a bash call. One seat had already committed real work to its branch when it died.

Ruled out: resource pressure (74% memory free, zero swap), credential expiry (`cotal doctor auth` healthy across daemon, `$SYS` and agent creds), and the manager restart / frozen issuance gate seen earlier the same night (#871) — those are real but the deaths continue independently of them.

## Where a fix goes

`extensions/connector-jcode/src/extension.ts:37` builds the seat's environment (`launchEnv({ envAllow })` plus the explicit `COTAL_*` set). Nothing sets `JCODE_TEMP_SERVER_IDLE_SECS`, and the existing comment at the `COTAL_JCODE_TUI` passthrough notes that anything needed must cross explicitly because `launchEnv` resets the per-session namespace.

Two candidate shapes, and the choice is a design call rather than a spelling:

1. The connector holds a client connection for the life of the seat, so the host's own liveness rule is satisfied honestly rather than suppressed.
2. The connector pins the idle timeout off for connector-managed hosts, since "no attached client" is the normal, permanent state of a headless seat and the timer can never mean what it is measuring.

Option 1 is the truer statement of the invariant; option 2 is smaller. Either way it must be the default and not an operator opt-in: a seat that dies after five minutes is unusable for real work, and the failure is silent to the operator until they notice the lane stopped.

## Related, but not duplicates

- **#828** — the launch-failure message is scrubbed to `Jcode host startup failed (unknown)`. Separate defect (error fidelity), separate fix; it cost several attempts to isolate this one because the real refusal never surfaced.
- **#866** — the manager's seat environment should come from an explicit allowlist. Adjacent surface: whatever this fix passes to the seat crosses the same boundary #866 is redesigning, so the two must not be designed in ignorance of each other.
- **#871** — manager restart leaves the issuance gate frozen. Same night, different mechanism; not the cause of these deaths.

Reproduction logs preserved outside the mesh root at `~/mesh-logs/idle-timeout-evidence-2026-08-26/` (three seat logs plus the manager log).

davidfarah2003:
A managed Jcode seat updated its own binary mid-session and died five minutes later, mid-turn. This pins the seat's version at spawn time.

## The defect

Jcode's background updater restarts the process tree when it lands a release. That restart SIGTERMs the seat's TUI — which is the only connection the Jcode server counts as a client, by the server's own accounting — and nothing re-attaches afterwards. The server then declares itself idle and reaps itself 300 seconds later, taking the seat with it: `exit code 1, signal 0`, no signal from the manager, and no message the operator can act on.

Measured 2026-08-26 on one mesh, three independent seats, identical chain:

| step | seat A | seat B | seat C |
| --- | --- | --- | --- |
| `Client connected. Idle timer cancelled.` | 06:39:12 | 06:36:34 | 06:39:58 |
| `Updated to v0.81.1. Restarting...` | 06:45:49 | 06:42:38 | 06:46:19 |
| `Received SIGTERM in TUI process` | 06:45:51 | 06:42:40 | 06:46:20 |
| `No clients connected. Server will exit after 5 minutes of idle.` | 06:46:02 | 06:42:44 | 06:46:28 |
| `Server idle for 5 minutes with no clients. Shutting down.` | 06:51:02 | 06:47:54 | 06:51:28 |

One seat died two seconds after starting a tool call. All three updated within seven minutes of each other, which is why a whole generation of seats died together while an earlier generation had run 35+ minutes untouched. Reviewer seats on other connectors ran through the same window untouched.

## The change

The launch environment now carries `JCODE_NO_AUTO_UPDATE=1` alongside the `JCODE_DISABLE_CLAUDE_MCP` it already set. A managed seat's version is whatever is on `PATH` when the operator spawns it, and it stays that version for the seat's life. Beyond this crash, a seat that silently swaps its runtime version mid-review is not a reproducible artifact.

Covered by a named cell in `smoke:jcode-host` (`host pins the seat binary against background self-update`) and a mutation fixture that deletes the flag and requires that cell to redden. Docs updated in the same change. Changeset: patch.

## What this does NOT prove, stated plainly

- **The suite proves the flag reaches the child, and nothing more.** End-to-end suppression of a real self-update needs an outdated binary and a published release, which no hermetic suite can stage — and the box this was measured on is already on the latest version. The field evidence above is what ties the flag to the crash; the cell ties the flag to the launch.
- **This removes the trigger, not the mechanism.** A restart still leaves a live Jcode session with no client, and the reaper still kills it. That behaviour is upstream Jcode's: a server whose session is mid-tool-call should not consider itself idle. Worth an upstream report; not fixed here.
- **`smoke:jcode-host` currently hangs after its last cell** and prints no completion banner (#874, pre-existing — verified by running the suite with this change stashed out: 11 cells versus 12, both stalling at the same cell). So the cell count quoted for this PR is a count of cells that ran, not a completed suite. That is a defect in the instrument, filed separately, and it should not be read as this change being unverified — but it should not be read as a completed green suite either.

Verified directly (exit codes read, not piped): `pnpm typecheck` 0, `pnpm changeset status` 0, the new cell green.

Related: #872 (this defect, with the corrected root cause), #828 (the connector scrubbing refusals it can name), #874 (the hanging suite).

Fixes #872


davidfarah2003:
Two updates on verification, one of which corrects a caveat in the description.

## The cell discriminates — proven by a manual mutation, not by the runner

`scripts/mutation-proof.mjs` **cannot grade this suite** right now: its per-command default is a 900 s timeout, and `smoke:jcode-host` never exits (#874), so the baseline is killed and scored as an unknown rather than a pass. The fixture in this PR is therefore correct but currently ungradeable by the tool. That is a consequence of #874 worth knowing about beyond this PR: the hang does not just make results ambiguous, it makes mutation-proof unusable for every cell in this suite.

The cell prints early in the run, well before the hang, so the same mutation was applied by hand instead:

- **flag removed** — suite dies at that cell, exit 1:
  ```
  AssertionError [ERR_ASSERTION]: host pins the seat binary against background self-update —
  {"JCODE_RUNTIME_DIR":"…","JCODE_HOME":"…","JCODE_SOCKET":"…",
   "JCODE_DISABLE_CLAUDE_MCP":"1","JCODE_COTAL_LAUNCH_IDENTITY":"…","JCODE_API_SOCKET":"…"}
  ```
  The dumped environment is the finding: `JCODE_NO_AUTO_UPDATE` is simply absent.
- **flag restored** — that cell passes and the run continues to 12 cells before hitting the #874 hang.

So the assertion depends on the shipped code and fails loudly without it. The working tree was verified clean before and after.

## Severity is higher than the description states

A fourth seat reproduced this while the PR was open, and its log carries a detail the first three did not make obvious: it logged `Updated to v0.81.1. Restarting...` **on a box already running v0.81.1**. The updater re-fires and restarts the process tree with nothing to update.

Consequences:

- This is not a version-transition artifact that drains away. Every managed jcode seat dies roughly five to six minutes after spawn, indefinitely, until this lands.
- It retires the explanation offered in #872 that the original three seats died together because they updated within seven minutes of each other. The truer statement is that each seat updates a few minutes after *its own* start, so deaths are rolling rather than clustered — which also removes "wait for the version to settle" as a workaround, because there is nothing to settle.
- Four of four jcode seats on this machine have now died this way.

An end-to-end check that the flag actually suppresses the update — two isolated bridges, one flagged, one not — is running now and its result will be posted here either way, including if it shows the flag does not work.

davidfarah2003:
## End-to-end result: the flag does suppress the self-update

The description originally said end-to-end verification was unstageable, because it would need an outdated binary and a published release. That turned out to be wrong for a reason that is itself part of the defect: **the updater re-fires on a binary already at the target version**, so the update can be provoked on demand.

Two isolated `jcode api-bridge` instances, each with its own `JCODE_HOME` and `JCODE_RUNTIME_DIR`, launched three seconds apart and left alone. One with `JCODE_NO_AUTO_UPDATE=1`, one without.

| | flagged | control |
| --- | --- | --- |
| `jcode starting` | 07:33:31 | 07:33:34 |
| `Server idle for 5 minutes with no clients. Shutting down.` | 07:38:32 (5m01s) | 07:38:35 (5m01s) |
| `Verified SHA256 checksum for jcode-macos-aarch64.tar.gz` | — | 07:39:03 |
| `Updated to v0.81.1. Restarting...` | — | 07:39:04 |
| `jcode starting` (restarted by the updater) | — | 07:39:06 |
| `builds/` in its home | **no such directory** | `current`, `current-version`, `shared-server` |

Both instances lived exactly 5m01s and were reaped by the same idle timer, so the comparison is like-for-like. The control then ran its update check as part of that shutdown, downloaded a binary, verified it and restarted itself. The flagged instance took the same path and did none of it — it has no `builds/` directory at all, so it never even began a download.

Three further unflagged homes left over from earlier probes in the same session all carry `builds/current` as well, which makes the ratio four unflagged homes that self-updated to one flagged home that did not.

**Stated limit, so this is not read as more than it is:** both instances were observed for a single ~5-minute cycle, and the control's update fired during its shutdown sequence. This shows the flag suppresses the update check on the path a seat actually takes; it is not a multi-hour soak.

A negative result here would have meant this PR is decorative, and it was set up to be able to produce one. It did not.

davidfarah2003:
## CI on this head: both red shards are the main baseline, not this diff

The run on `4fd3fa439` finished with `smoke (shard 0/4)` and `smoke (shard 1/4)` red. Both fail at
the same suites that already fail on `origin/main`:

| shard | this PR @ 4fd3fa439 | origin/main @ 565f667c4 | main @ cb7f38473 |
| --- | --- | --- | --- |
| 0/4 | `pnpm smoke:artifact-store` | `pnpm smoke:artifact-store` | red |
| 1/4 | `pnpm smoke:gate-inventory` | `pnpm smoke:gate-inventory` | red |

Green on this head: `smoke (shard 2/4)`, `smoke (shard 3/4)`, `unit`, `docs`, `live`, CodeQL, and
every Windows job including all four non-blocking shards.

Both failing suites are the known #868 baseline. Neither touches the connector this PR changes.

One caveat stated rather than glossed: CI grades the merge commit, not the tip, and this is
attribution by matching the failing suite name on both sides, not a bisect. It is not a claim that
`smoke:artifact-store` and `smoke:gate-inventory` are healthy, only that this diff is not what
broke them.

Separately, while reading the CI logs: `package.json` on `origin/main` carries
`"smoke:jcode-retry-policy"` twice, at lines 85 and 90, with identical values. Harmless to shard
membership, since `bin/smoke/shard.mjs` reads `bin/smoke/ci-suites.txt` rather than the script map,
but it emits a duplicate-key warning on every esbuild pass. Noted here rather than fixed, since it
is outside this PR's scope.

davidfarah2003:
## Review record, transcribed after the fact

**Provenance, stated plainly: this is not a first-hand post.** The reviewer (`rev875b`, an
independent pi seat) delivered this verdict by direct message before the merge, and I have
transcribed it here afterwards from that message. It was not posted by the reviewer and it is not a
GitHub review object. The seat was stood down after the merge, so the original delivery no longer
exists anywhere but this transcription.

It is being recorded because the merge record otherwise cannot distinguish "independently reviewed
at the exact sha" from "merged unreviewed", and those are precisely the two cases the review
requirement exists to separate. A check that leaves an artifact indistinguishable from the check
not happening is not much of a check.

---

**Verdict: APPROVE @ `4fd3fa43944cf8613c6b21a3b690dbfa50fce949`. No blockers.**

The reviewer read the head itself, graded in its own detached worktree at that sha, and re-read the
head afterwards to confirm it had not moved. I re-read it a third time immediately before merging.

**What it verified, rather than accepted:**

1. **The flag reaches the child.** The named cell `host pins the seat binary against background
   self-update` is green with the change and red without it: the reviewer deleted
   `JCODE_NO_AUTO_UPDATE: "1"` in its own worktree and the assertion failed on the launch env, with
   `JCODE_DISABLE_CLAUDE_MCP` present and `JCODE_NO_AUTO_UPDATE` absent. Restored afterwards.
2. **The env key is the right one**, checked against the installed binary (v0.80.1 / `d37ffd3dc`,
   the generation that died updating to v0.81.1) and against upstream v0.80.1 source, not against
   this PR's claims: `JCODE_NO_AUTO_UPDATE` appears in the binary and in
   `crates/jcode-app-core/src/update.rs::should_auto_update()`, where any value disables;
   `check_and_maybe_update` returns `NoUpdate` immediately when set; and the release-build
   background check in `src/cli/startup.rs` is the path that logs `Updated to {}. Restarting...`
   then calls `replace_process`, which is the measured crash line.
3. **`api-bridge` is not in the update-skip list** (only Update / Serve / Acp / `--resume` /
   `--quiet` / `--no-update`), and the SDK forwards `options.env` onto the api-bridge child, so the
   flag lands on the process that would otherwise have self-updated.
4. **The TUI launch path** (`host.ts` `launchTui()`) does not set the flag, and the reviewer
   established independently that this is not a hole on this binary:
   `should_spawn_background_update_check_with_config` requires `args.resume.is_none()`, and the TUI
   argv is `["--socket", …, "--resume", sessionId]`. It states explicitly that it did not take this
   from the previous reviewer.

**What it explicitly did not verify**, in its own words: the official
`scripts/mutation-proof.mjs` run (unrunnable while `smoke:jcode-host` hangs, #874 — the hand A/B
above was used instead); end-to-end suppression against an outdated binary and a published release;
a live TUI process with an outdated binary (source-level `--resume` skip only); `smoke:jcode-live`.
`smoke:jcode-security` failed on that box for a missing `docker.sock`, and
`smoke:jcode-provider-disconnect` timed out on a reconnect path this one-line env change does not
touch; neither was treated as a blocker and neither is attributed to this diff.

**Suites it ran green:** `pnpm typecheck`, the connector's own typecheck, `pnpm changeset status`,
`smoke:jcode-args`, `smoke:jcode-private-state`, `smoke:jcode-lifecycle`,
`smoke:jcode-session-resume`, `smoke:jcode-retry-policy`, `smoke:jcode-route-identity` (17 cells).

It independently confirmed the CI attribution posted above: shard 0/4 at `smoke:artifact-store` and
shard 1/4 at `smoke:gate-inventory`, both #868 baseline, by reading the job logs itself.

Issue #871, opened 2026-08-26
up: delivery and manager boot concurrently, and a lost manager startup race freezes the issuance gate silently


davidfarah2003:
Field report from a remote user-auth mesh deployment (relayed, not yet reproduced locally — filing so it isn't lost; needs a repro before any fix per repo convention).

## Reported chain

1. `cotal up --detach` starts the delivery daemon and the manager concurrently. On the reporting box the manager loses the startup race and crashes mid-registration on every start.
2. The half-done registration freezes the issuance gate (SPEC 13.8, "another barrier holds it"), and every subsequent manager start refuses with the same frozen-gate error.
3. The failure is silent for ~24h: the mesh looks up, but no issuance happens. It becomes visible only when the delivery daemon dies of JWT expiry ("User Authentication Expired" → lease exit) and membership freezes mesh-wide.
4. Recovery required the reconcile-gate procedure (holder proven gone via a CONNZ sweep, 3 holders verify-evicted, gate reopened at generation 2).

## Asks

- `cotal up` should sequence or retry the manager start so a lost boot race cannot leave the gate frozen. A crash mid-registration on a cold boot should either be retried into a clean registration or fail the whole `up` loudly, not strand a half-registration.
- The delivery daemon logged renewal-failure warnings for hours before the expiry exit; that condition needs a louder operator surface than a log file.

## Related

- #725 tracks the frozen-gate error not naming reconcile-gate (the fallback half of the first ask).
- #856 tracks reconcile-gate itself being unusable in production; this issue is about not needing it on a cold boot at all.
- #817 is a different route into the same frozen-gate state (seats outliving their manager).
- #864 is restructuring the `up` launch path (detach + persisted record); whoever fixes this should coordinate with that change rather than layering sequencing onto the pre-#864 code.

davidfarah2003:
Local evidence for the frozen-gate half of this report, observed on a macOS 0.30.1 mesh tonight (the original report was remote and unreproduced).

The manager was restarted several times in a row. Its own log shows the re-registration failing to evict the superseded family, then refusing every subsequent attempt:

    ✗ re-registration could not revoke + verify-evict the superseded serve family; the gate is
      left frozen for reconciliation, no new spec published (SPEC 13.1): principal
      "local.U…" is not verified evicted
    ✗ the issuance gate for "<instance>" is frozen; another barrier holds it; re-read and
      re-decide (SPEC 13.8)

[...]

Issue #864, opened 2026-08-26
cotal up leaves no persisted mesh record — a dead stack erases the mesh's existence


davidfarah2003:
Two coupled behaviors, measured 2026-08-26 on macOS with 0.30.0→0.30.1:

**1. An `up`-started mesh is only discoverable while its processes live.** `cotal meshes` rows for meshes registered with `meshes add` persist in `~/.cotal/meshes/` and carry the `registered` tag. A mesh started by `cotal up` shows a row with no tag, derived from runtime state. When the stack's processes die, the row vanishes: the operator gets `✗ no mesh named "X" is running - see cotal meshes`, and `cotal meshes` shows nothing — no record the mesh ever existed, no root path, no restart affordance. To someone who did not start the stack themselves this reads as data loss ("my mesh is not even registered, wtf" — verbatim operator reaction).

**2. `cotal up` runs the whole stack as children of the invoking process, with no daemonize option.** Broker, delivery daemon and manager all die with the invoking process tree. When the invoker is an agent session — increasingly the normal case, an agent setting a mesh up on a person's behalf — a session restart silently kills the broker, the manager, and every seat mid-work.

Together they compound: the session dies → the mesh dies with it → and the CLI then denies the mesh was ever there, so the person has nothing to restart from except memory of a directory path.

**Proposal:**
- Persist the mesh record at provision time (same store as `meshes add`, marked self-hosted with its root), so a dead stack yields `mesh "X" is recorded at <root> but not running — run cotal up there to restart` instead of an existence denial.
- Give `cotal up` a `--daemon` mode (setsid/reparent + pid files), or make detaching the default with the current behavior behind `--foreground`. An agent-invoked `up` should produce a stack that outlives the agent.

davidfarah2003:
Operator directive raising the bar on this: **the right path should be the only way** — cotal must be resilient to a naive agent launching it. Not a `--daemon` flag an agent can forget, but the default and only behavior: `up` always detaches the stack from the invoking process (setsid/reparent), always persists the mesh record, and treats the invoking environment as hostile (an agent session's env, an ssh hangup, a CI runner — none of them may become a load-bearing parent). `--foreground` can exist for debugging; nothing else should depend on the invoker's lifetime.

davidfarah2003:
[...]

Issue #813, opened 2026-08-23
Every manager with >24h uptime loses its session plane: the session signing key is minted once at boot and never renewed


davidfarah2003:
Hit three times in one day on a single host. Reported by the operator:

```
cotal attach --name fm-orch-cotal
✗ signing key mgr-sessions-UCZ2XFX524KR is outside its validity window at 1787489502074
  (window 1787365061492..1787451521492); out-of-window use fails closed (SPEC 13.10)
```

Decoded: the window is **24.02 hours**, and the failure landed **10.55 hours after it closed**.

## Root cause

`implementations/manager/src/manager.ts`, in `standUpSessionPlane()`:

```ts
const signer = newArtifactSigner();
const keyId = `mgr-sessions-${identity.id.slice(0, 12)}`;
const anchor: SignerAnchor = {
  keyId, publicKey: signer.publicKey, owner: MANAGER_ENDPOINT, roles: ["sessions"],
  scope: { sessions: [MANAGER_ENDPOINT] },
  validFrom: Date.now() - 60_000,
  validTo: Date.now() + SESSION_GRANT_MAX_TTL_MS,   // 24h, frozen at boot
};
…
resolveAnchor: (id) => (id === keyId ? anchor : undefined),
```

The anchor is a `const` in a local scope, computed once. `resolveAnchor` returns that same object for the life of the process, and **nothing anywhere re-mints it**. `grep` for `validTo` across `implementations/` finds exactly one non-test assignment: this one.

So this is not a race, a clock skew, or a credential problem. **Every manager that stays up past 24 hours loses its session plane, permanently, by construction.**

## Why it is severe

- **It is guaranteed, not probabilistic.** Uptime > 24h means the session plane is dead.
- **The client cannot recover it.** The expired window lives in the *manager's* anchor, so restarting the CLI, re-authing, or rotating credentials all fail identically.
- **The only known recovery kills everything.** Restarting the manager re-mints the key and takes down every live session with it. On this host that meant a full fleet restart.
- **It is silent until it bites.** Nothing warns as the window closes; the first symptom is a hard refusal at attach time.

## What is NOT the bug

Failing closed on an out-of-window key is correct and required (SPEC 13.10). `resolveAnchorForUse` in `packages/core/src/endpoint-signing.ts` is doing exactly its job. The defect is that a **correct refusal became permanent** because no renewal path existed at all.

## Fix

`packages/core/src/signing-key-rotation.ts` — a `RotatingSigner` that:

[...]

PR #879, opened 2026-08-26
fix(manager): complete a frozen registration gate on boot

davidfarah2003:
## Summary

A manager that dies mid-registration leaves the issuance gate frozen under that registration op. The freeze is correct (it stops two incarnations serving at once), but the successor then refused SPEC 13.8 ("the issuance gate is frozen; another barrier holds it") until an operator ran `cotal reconcile-gate`. A routine restart should not require that.

Boot now completes the dead registration itself, using the same composition as `cotal reconcile-gate`, then continues the normal takeover.

This is **#783 item 3 only** (same surface as #871). It is not #773 (daemon `reloadCreds` two-store split) and not rolling PTY handoff.

**#783 item 1 is already shipped.** Session-signing-key rotation landed as #813 (`RotatingSigner`, `d6a75c139`). The issue text conflates that with the daemon `reloadCreds` generation mismatch; those are different stores and different keys.

## Shape

- Before `registerServiceInstance` demands `open`, if the gate is `frozen` under a **registration** op, run `reconcileEndpointGate` with the same probe/evict seams the CLI uses, on the dead op's own `opId` and freeze token.
- Admit only `gone` **and** `sweepComplete=true`. Live / unknown / unestablishable / wrong-op-kind stay loud refusals. No TTL (request-ingress has no process-epoch fence).
- Abort-reopen (generation+1, processEpoch unchanged), then the successor's own freeze and takeover. Not a mid-phase resume of the interrupted barrier.
- Concurrent CLI reopen that leaves the gate `open` continues boot; raced-but-still-frozen stays loud.
- Open meshes skip this path (no delivery-admin CONNZ oracle).

`cotal reconcile-gate` remains for when boot cannot run (daemon down, non-manager endpoint, or lifting the freeze without starting a manager).

## Evidence

Reproduced first against a private broker and `COTAL_HOME` (never the live macfleet stack). Before the fix, successor start failed with the exact SPEC 13.8 copy. After:

- `pnpm smoke:boot-self-heal-gate` — 12/12. Dead holder + complete sweep: successor starts, instanceId preserved, epoch advanced. Live holder: `holder-alive`, gate and family untouched. No oracle: `liveness-unestablishable`, gate stays frozen.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/boot-self-heal-gate.json` — both mutants KILLED (skip the heal; forge `gone` at the boot probe).
- `pnpm smoke:gate-reconcile:auth` — 33/33.
- `pnpm smoke:manager-restart-live` — 7/7 (ordinary AUTH restart still succeeds).
- `pnpm --filter @cotal-ai/manager typecheck` — clean.

Docs updated in the same change (`docs/cli.md`, `docs/run-a-mesh.md`, `docs/control-surface.md`). Suite appended to `bin/smoke/ci-suites.txt`. Changeset included.

## Out of scope

- #773 two-root daemon credential adoption.
- Rolling PTY / fd-passing handoff (#783 item 2).
- `cotal up` sequencing (#864 / #871 ask 1).


davidfarah2003:
CI attribution for reviewers:

- **smoke shard 1/4 red** is the known main baseline (#868): `smoke:gate-inventory` fails on ungated `smoke:jcode-private-lifecycle`. Confirmed on this run (`FAILED at: pnpm smoke:gate-inventory`). Not this change. That suite is index 1 of `ci-suites.txt`, so it always lands in shard 1.
- **This PR's suite** is `smoke:boot-self-heal-gate`, appended at the end of `ci-suites.txt`, so it runs in **shard 3/4**. Local: 12/12, both mutation-proof mutants killed, `smoke:gate-reconcile:auth` 33/33, `smoke:manager-restart-live` 7/7.
- **smoke:jcode-host** is also shard 1 (#874 hang). This PR does not touch it.

davidfarah2003:
## Review record, transcribed after the fact

**Provenance, stated plainly: this is not a first-hand post.** The reviewer (`rev879`, an
independent pi seat) delivered this verdict by direct message before the merge, and I have
transcribed it here afterwards from that message. It was not posted by the reviewer and it is not a
GitHub review object. The seat was stood down after the merge, so the original delivery no longer
exists anywhere but this transcription.

It is being recorded because the merge record otherwise cannot distinguish "independently reviewed
at the exact sha" from "merged unreviewed", and those are precisely the two cases the review
requirement exists to separate.

---

**Verdict: APPROVE @ `4914da2af5eedcbc35aadaaf1f6bf0d721e245b2`. No blockers, one named residual.**

The head moved under this reviewer mid-review, from `a4a418a1c` to `4914da2af`. It was re-pinned,
and it independently confirmed the delta was one file and one insertion (the changeset line listing
`@cotal-ai/connector-core`) before carrying its findings forward rather than re-grading from
scratch. I re-read the head again at merge time and it had not moved further.

**The admission predicate, which is the whole safety of this change.** The reviewer was asked to
attack one question: can a freeze holder that is still running ever read as `gone`, letting two
managers proceed under one gate? It traced the derivation rather than reading the PR body:

- `gone` is produced only by `observePrincipalLiveness` on a live CONNZ sweep that is complete,
  single-server-proven (SPEC 13.13: one replier, no cluster declaration), with no row attributing
  to the freeze holder. Incomplete, lost-page, empty-mid-page, malformed, no-reply, cluster and
  multi-replier all withhold it as `unknown` or `sweepComplete=false`.
- `holderLivenessFromReply` is the AND: `gone && sweepComplete !== true` collapses to
  `unestablishable`, as do garbled, foreign, no-echo, timeout, unreachable rail and daemon refusal.
- The reconciler then refuses `live` / `unknown` / `unestablishable` by name before any mutation,
  and `wrong-op-kind` refuses before the probe runs at all. The `HolderLiveness` union is closed,
  so after those refusals only `gone` proceeds. There is no TTL.
- Boot swallows only `not-frozen` / `raced`, and only when the gate is now open; raced-but-still-
  frozen stays loud.
- The freeze holder is the persist serve principal, not the registrar process. Same-instance dual
  start is blocked earlier by the lease, acquired before the heal. A wedged registrar that has
  already verify-evicted its serve family can read CONNZ-`gone` while its process still runs, but
  it cannot mint serve after the successor abort-reopens, because the token-pinned CAS loses. That
  is not two managers serving under one gate.
- False-`gone` via a wrong-account sweep is pinned at daemon start (root and account versus daemon
  cred); a mismatch refuses. The presence roster is never consulted. An open mesh skips the path.

**Named residual, not a blocker.** The AND lives in the probe mapping and is not re-checked inside
the reconciler, so a forged `{state:"gone"}` probe would admit. The PR's own mutation 2 is exactly
that forge and the live-holder cell kills it, which is why this is a residual. But the belt against
"process alive, no CONNZ row" is the lease, not the eviction, because eviction reads the same CONNZ
family and is therefore not independent. Worth remembering if anyone later proposes relaxing the
lease. The live-holder cell also uses a minted agent cred attributed as the serve principal rather
than a live Manager process.

**Proof it ran itself** (in its own worktree, nats-server v2.14.5): `pnpm smoke:boot-self-heal-gate`
12/12, covering dead holder (real CONNZ `gone` with `sweepComplete=true`, successor starts,
instanceId preserved, epoch 0→1), live holder (`holder-alive`, gate and family untouched), and no
oracle (`liveness-unestablishable`, gate stays frozen). It confirmed the residue is built by the
shipped barrier after a real register-and-stop rather than by a hand-written gate row, and it ran
`node scripts/mutation-proof.mjs --config bin/smoke/mutations/boot-self-heal-gate.json` with both
mutations KILLED.

**What it explicitly did not run**, in its own words: `smoke:gate-reconcile:auth`,
`smoke:manager-restart-live`, and the manager typecheck — those numbers are the author's, not its
own. It noted that the `holder-unknown` and `wrong-op-kind` cells live in `gate-reconcile-auth`
rather than in this suite, though boot uses the same function.

It confirmed the scope is item 3 of #783 only (no #773, no PTY handoff, no TTL) and that the
changeset correctly lists the two first-party packages as patch. It attributed the CI reds to the
#868 baseline itself.

PR #888, opened 2026-08-26
test(web): repair authenticated console smokes

davidfarah2003:
## Summary

- strip inherited `COTAL_*` connection material before the console-auth smoke starts its child
- exchange the bounded-aggregation console's single-use launch link once and carry its session cookie across all route assertions
- preserve every bounded aggregation status, body, deadline, partial-page, and refusal assertion

## Verification

- `pnpm smoke:suite-ambient-env`
- `pnpm smoke:web-console-auth` (142 passed, 0 failed)
- `pnpm smoke:web-bounded-aggregation` (40 passed, 0 failed)
- `pnpm typecheck`

### DIFF
--- a/implementations/web/smoke/bounded-aggregation.smoke.ts
+++ b/implementations/web/smoke/bounded-aggregation.smoke.ts
@@ -362,27 +362,36 @@ try {
     webChild.stdout?.on("data", (d: Buffer) => { log += d.toString(); });
     webChild.stderr?.on("data", (d: Buffer) => { log += d.toString(); });
 
-    // Readiness is a real 200 from a CHEAP route, so "the server is up" is never inferred from a
-    // process that started and then failed to connect. No skip if it never comes up: the cells below
-    // fail, loudly, which is the correct reading of a dashboard that cannot serve.
-    let up = false;
-    for (let i = 0; i < 200; i++) {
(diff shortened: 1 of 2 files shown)


davidfarah2003:
APPROVE @ 04fc2a85ca228ef351c95464a4457ab75a5ee9b4

Verified at the pinned head:
- `pnpm build`: exit 0.
- With `COTAL_*` scrubbed, `pnpm smoke:suite-ambient-env`: PASS.
- With `COTAL_*` scrubbed, `pnpm smoke:web-bounded-aggregation`: 40 passed / 0 failed.
- Deliberately broke partial marking, the DM deadline status/message, and the server partial log. The named route cells 5.3, 5.6, 5.7, and 5.10 all reddened. After restoring, the suite returned to 40 passed / 0 failed.
- The launch URL is spent once, the minted session cookie is reused, and cell 5.0 fails loudly unless exchange is 302, a session exists, and authenticated readiness is 200.
- The added environment strip removes every inherited `COTAL_*` key. `COTAL_WEB_DETACHED_LOG=1` is only a behavior switch suppressing the launch URL in persistent output. The other child in the file receives explicit `--server` and `--space` values and web.ts does not read inherited connection settings.
- Diff remains test-only: two smoke files, no census waiver and no product change.

PR #890, opened 2026-08-26
test(web): authenticate guarded smoke suites

davidfarah2003:
## Summary

- spend each web console launch link once in the four remaining guarded smoke suites
- reuse the minted `cotal_web_session` cookie for fetch and raw-socket requests
- preserve every existing behavior predicate while restoring the routes each suite measures

## Verification

- `pnpm build`
- `pnpm typecheck`
- `pnpm smoke:web-history-limit` (39 cells)
- `pnpm smoke:web-echo-escaping` (29 cells)
- `pnpm smoke:web-channel-alias` (11 cells)
- `pnpm smoke:web-body-cap` (30 cells)

All commands ran with `COTAL_*` variables scrubbed.

## Discrimination proof

Each named cell was deliberately made red by changing only its request, then restored and rerun green:

- history limit: `3.4 control: the identical read with no link cost returns the full page, so 3.3 is the link`
- echo escaping: `3.3 a channel name that DECODES but the wire would rewrite is refused 400 with the offending codepoint escaped, never raw`
- channel alias: `2.1 a name the wire would rewrite is REFUSED at the history route, rather than quietly reading the channel it aliases onto`
- body cap: `5.2 a body with NO declared length is still refused, so the cap is enforced on the bytes as they arrive and not only on what the caller admits to`

### DIFF
--- a/implementations/web/smoke/body-cap.smoke.ts
+++ b/implementations/web/smoke/body-cap.smoke.ts
@@ -99,21 +99,26 @@ try {
   webChild.stdout?.on("data", (d: Buffer) => { log += d.toString(); });
   webChild.stderr?.on("data", (d: Buffer) => { log += d.toString(); });
 
-  let served = false;
-  for (let i = 0; i < 200; i++) {
-    const r = await fetch(`http://127.0.0.1:${WEB_PORT}/api/roster`).catch(() => undefined);
-    if (r?.status === 200) { served = true; break; }
+  let launchUrl: string | undefined;
+  for (let i = 0; i < 200 && launchUrl === undefined; i++) {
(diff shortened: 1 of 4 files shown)


davidfarah2003:
APPROVE @ 035188e67

All four diffs preserve the original predicates and change only request authentication/readiness. Body-cap authenticates every fetch and every raw request construction, including declared, chunked, keep-alive refusal, and reusable keep-alive controls. Launch/session failure is non-vacuous: history-limit throws immediately, while the other suites fail their served control and their subject predicates rather than accepting auth responses.

Independent discrimination proofs: bypassing canonical channel validation reddened named cell channel-alias 2.1 with a 200 alias read, then restoration returned 11/11 green. Disabling invisible-codepoint escaping reddened named echo-escaping cell 3.3 for RLO, ALM, and TAG, then restoration returned 29/29 green. Build passed before smoke execution.

Issue #887, opened 2026-08-26
main CI: shards 2 and 3 red since the #454 merge (console-auth spreads ambient COTAL_ env; bounded-aggregation drives gated routes with no session)


davidfarah2003:
Main's `CI` workflow is red on all four smoke shards. Shards 0 and 1 are the known baseline in #868.
Shards 2 and 3 are new, and both trace to the #454 merge.

## When it broke

Shard conclusions across main's recent `CI` runs:

| run | sha | sh0 | sh1 | sh2 | sh3 |
| --- | --- | --- | --- | --- | --- |
| 08:53 | 7466e6a3e | fail | fail | **fail** | **fail** |
| 08:46 | 9f7815e98 | fail | fail | **fail** | **fail** |
| 08:04 | 24e9fb29e | fail | fail | success | success |
| 07:42 | a11d663ff | fail | fail | cancelled | success |
| 05:32 | d19c9b607 | fail | fail | cancelled | success |

`9f7815e98` is the #454 merge commit. The only other change in that window is the release version
bump. Both failures are reproduced locally at main's tip with `COTAL_*` scrubbed from the
environment, and match CI exactly.

## Shard 2/4 — `smoke:suite-ambient-env`

```
these suite files spread the ambient environment into a child without stripping COTAL_ first:
  implementations/web/smoke/console-auth.smoke.ts
actual: [ 'implementations/web/smoke/console-auth.smoke.ts' ]   expected: []
```

`bin/smoke/suite-ambient-env.smoke.ts` censuses suite sources for a child spawn that spreads
`process.env` without stripping `COTAL_` first. #454 added `console-auth.smoke.ts`, which does, and
never registered it.

The census is doing its job here rather than being in the way: whatever runs a suite may be a
managed agent session, so the spread hands the child a live credential and a live broker URL. The
fix belongs in the suite — strip the keys — not in a waiver.

## Shard 3/4 — `smoke:web-bounded-aggregation` (30 passed, 10 failed)

Every failing cell, 5.0 through 5.10, returns HTTP 401 `unauthenticated`:

```
x FAIL  5.1 CONTROL: a small route answers 200 across this link: 401
x FAIL  5.2 `/api/activity` ANSWERS 200 where the shipped route answered 500: {"status":401,"aMs":2}
x FAIL  5.6 `/api/dms` REFUSES at its deadline rather than answering long after the reader left: {"dStatus":401,"dMs":1}
[...]

Issue #666, opened 2026-08-19
Backup inventory omits the endpoint plane, the session ledger, and the workflow journal: backups fail closed on fully-provisioned spaces


davidfarah2003:
## Summary

`spaceBackupInventory()` enumerates 13 streams per space (8 backed up + 5 excluded). A fully
provisioned user-auth space holds 25. The extra twelve families are:

```
EPC_, EPE_, EPF_, EPJ_, EPR_, EPT_REQ_, EPT_, EPW_, WFJ_,
KV_cotal_auth_, KV_cotal_records_, KV_cotal_sessions_
```

(names elided to family prefixes; each carries the space token)

`validateSpaceBackupInventory` is exact set-equality, so `cotal backup create` on such a space
fails closed:

```
space backup inventory mismatch: missing [], unexpected [EPC_, EPE_, EPF_, EPJ_, EPR_, EPT_REQ_,
EPT_, EPW_, KV_cotal_auth_, KV_cotal_records_, KV_cotal_sessions_, WFJ_]
```

and the failed run writes no artifact, so the caller then ENOENTs on the manifest.

Two of the twelve (`KV_cotal_auth_`, `KV_cotal_records_`) are the pair already tracked in #356.
The new unlisted set is ten: the seven endpoint-plane streams (EPJ/EPF/EPE/EPT_REQ/EPT/EPR/EPW),
the endpoint contract store (EPC), the session ledger (KV_cotal_sessions_), and the workflow
journal (WFJ_).

## Evidence

On origin/main, after `pnpm build`:

```
pnpm smoke:backup-usermode:live
```

exits 1 at the quoted mismatch (step D of that suite, `cotal backup create` against a real
`up --user-auth` mesh), then fails reading the manifest the failed backup never wrote. The
failure reproduces identically on content that predates unrelated in-flight changes, so it is
pre-existing on main.

## Why the CI gate does not see it

The de-circularized gated check from #643 (PR #658) validates the inventory against what a real
broker holds for a space created by `setupSpaceStreams`, which provisions exactly the 13
enumerated streams and none of the twelve. The check is green there, correctly: within that
corpus the inventory is exact. A user-auth space is a larger universe because the provisioner
(`provisionerPermissions` in `packages/core/src/provision.ts`) creates the seven endpoint
streams, WFJ, EPC, and the records/auth/sessions buckets. The corpus root is part of the claim,
so this gap closes only when the families are enumerated and the live suites join a CI shard
[...]

Issue #356, opened 2026-08-12
§13.12 authority stores are created but never reaped: backup validation throws on real spaces, and teardown leaks both KV buckets permanently


davidfarah2003:
Found by the artifact-plane lane's S2 live suite (which enumerates what the broker actually holds rather than a synthesized list) and verified on origin/main:

`ensureAuthorityStores` creates `KV_cotal_auth_<space>` and `KV_cotal_records_<space>` from `setupSpaceStreams`, but neither appears in `spaceBackupInventory()` nor in `deleteSpace`'s explicit array. Two live consequences today:

- `validateSpaceBackupInventory` against a real provisioned space **throws** `unexpected [KV_cotal_auth_*, KV_cotal_records_*]` — space backups fail closed on every space with authority stores.
- `cotal down -f` **leaks both buckets permanently**: `deleteSpace` works from its explicit list and its own comment names teardown as the sole `STREAM.DELETE` holder, so nothing else can ever reap them.

This is the exact resource-lifecycle defect class where a resource must appear in every list that governs it (create / delete / grants / backup inventory / restore) — here a resource is in *create* only.

The artifact lane deliberately did NOT fix this cross-lane: the authority stores belong to the control-surface (§13.12) work, and a quiet edit to teardown grants from a file-transfer branch would be the wrong provenance. Its new inventory suite subtracts these two stores BY NAME with the reason and removal instruction written beside the constant, so the suite still guards its own membership without normalizing the hole. Whoever owns §13.12 should add both stores to the delete array, the backup inventory (with a deliberate class), the teardown grants, and the restore path — and then delete the named subtraction from `smoke:artifact-store`.

davidfarah2003:
A second unenumerated stream family, surfaced by a lane typechecking the smoke trees, which corroborates this defect class and extends it beyond the auth/records buckets.

Repairing the type errors in `backup-conservation-live` advanced it to the actual backup, where it stops at:

```
space backup inventory mismatch: missing [], unexpected [EPC_..., EPE_..., EPF_..., EPJ_..., EPR_...]
```

That suite was dead, not merely stale: `provisionAgent` has required a `lifecycleUid` since 2026-07-14 and the suite never passed one, so it threw before its first line. It is in `pnpm check` but in no CI shard, so nothing caught it for a month.

[...]

PR #892, opened 2026-08-26
test(ci): gate smoke:jcode-private-lifecycle

davidfarah2003:
`smoke:jcode-private-lifecycle` is defined in `package.json` but absent from `bin/smoke/ci-suites.txt`,
so `smoke:gate-inventory` reds shard 1 on every run. This is half of #868.

#868 deliberately left the choice open — gate the suite, or remove it from the inventory's universe —
because it depended on whether the suite is actually CI-runnable. Some connector-jcode suites are
not: `smoke:jcode-host` hangs after its last cell (#874).

**This one is.** Run at this tree with `COTAL_*` scrubbed, it exits 0 in under a second:

```
$ pnpm smoke:jcode-private-lifecycle
PRIVATE LIFECYCLE SMOKE PASSED (status-1 failure for live PID stays loud)
rc=0
```

So the answer is to gate it, and `smoke:gate-inventory` passes with the entry appended
(`GATE INVENTORY OK`).

## Append-only, verified rather than assumed

`ci-suites.txt` is positional and auto-merges without a conflict marker, so a plain diff is not a
valid check on it. Verified by prefix comparison instead:

```
prefix preserved: True
appended only   : ['smoke:jcode-private-lifecycle']
no reordering   : True
```

384 real suites become 385. No existing index moves, so no suite changes shard.

## What this does NOT claim

It closes the inventory gap. It does **not** necessarily green shard 1, and I want that stated
rather than discovered.

`smoke:gate-inventory` sits at index 1, the second suite in shard 1, and a shard aborts at its first
failure. So it has been masking roughly 95 downstream shard-1 suites that have never been observed
on this baseline. Fixing it lets shard 1 run much further; whether it runs all the way green is
unknown, and some of those suites may fail for unrelated reasons.

This exact mistake was made once already this week: #888 repaired a shard-3 suite, shard 3 stayed
red, and the fix looked like a failure when it had actually worked and merely let the shard advance
to the next masked failure (#887). Treat any new shard-1 red after this as newly *revealed*, not
newly *caused*, and check it against the suite list before attributing it here.

The newly gated suite itself lands at index 384, which is shard 0. Shard 0 currently aborts earlier
at `smoke:artifact-store`, so it will not actually execute in CI until the other half of #868 is
fixed. That is a pre-existing condition, not a regression introduced here.

## Not addressed

The other half of #868, shard 0's `smoke:artifact-store`, is untouched. That one is not CI hygiene:
the suite is correctly reporting the #666 backup-inventory omissions (endpoint plane, session
ledger, workflow journal) and the #356 reap gap. It needs a product decision, not an append.

### DIFF
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -542,3 +542,4 @@ smoke:claude-launch-env
 smoke:web-console-auth
 smoke:no-implicit-general
 smoke:session-channels
+smoke:jcode-private-lifecycle


davidfarah2003:
APPROVE @ 5ee5d554

Verified independently:
- `origin/main`'s 384-suite parsed sequence is an exact prefix of the PR's 385-suite sequence. The sole parsed delta is `smoke:jcode-private-lifecycle`; every existing index remains stable.
- `smoke:jcode-private-lifecycle` passes with all `COTAL_*` state absent in a minimal CI-like environment. It is not vacuous: its strict assertion exercises the production status-1/live-PID branch and requires the original operational error to remain loud. It uses injected probes, not machine-local process or Cotal state.
- `smoke:gate-inventory` passes and reports 385 suites with no duplicates.
- Parsed indices and round-robin placement are as claimed: `smoke:gate-inventory` is index 1 / shard 1; the appended suite is index 384 / shard 0. Shard 1 therefore had about 95 later suites masked, while shard 0 reaches `smoke:artifact-store` at index 232 before the new index 384 entry.
- The newly exposed `smoke:eps-grant-sweep` failure is not caused or relocated by this append: it remains at index 45 / shard 1, the PR changes neither its smoke nor `provision.ts`, and `origin/main` already has `remote-manager` in the profile source but absent from the sweep.

The PR body accurately limits the claimed effect: it fixes the inventory gap without claiming either shard becomes green or that the newly appended suite can execute before shard 0's existing earlier failure is repaired.

PR #895, opened 2026-08-26
test(core): sweep remote manager EPS grants

### LINKED ISSUE #894: smoke:eps-grant-sweep has no producer for the remote-manager profile, and its documented compile-time backstop does not run
`smoke:eps-grant-sweep` fails on main: the `remote-manager` profile is in the frozen D5 lifetime
table but the sweep has no producer for it.

```
✗ FAIL: the sweep covers EXACTLY the profiles of the frozen D5 lifetime table
        (a new profile fails here until it is swept) {"missing":["remote-manager"],"extra":[]}
✗ shard 1/4 FAILED at: pnpm smoke:eps-grant-sweep (exit 1)
```

## Where it comes from

`remote-manager` was added by `f721478b4` (`feat(auth,manager)!: add remote manager authority`),
which put it in the profile union at `packages/core/src/provision.ts:152` and in
`CREDENTIAL_LIFETIMES` at `packages/core/src/provision.ts:221`, without adding a corresponding
entry to `PRODUCERS` in `packages/core/smoke/eps-grant-sweep.smoke.ts`.

The suite is behaving exactly as designed. Its own header states the rule it is enforcing:

> A profile the sweep cannot construct FAILS; there is no skip list, because a skipped profile is
> an escaped profile.

So this is not a false positive to silence. `remote-manager` is a standing-renewable credential
that provisions descendants of its owner, which is precisely the class the sweep exists to check
for foreign-session-rail leakage. It needs a real producer, not an exemption.

Building one is not a one-liner: `permissionsFor("remote-manager", …)` throws unless given
`opts.remoteManager` with the `{instanceId, owner}` of an authorized manager lifecycle
(`provision.ts:988-990`), and validates the owner against the authenticated principal
(`:1395`) and the actor against the server-selected `manager_<iid>` / `manager_exec_<iid>`
(`:1399`). `implementations/auth/smoke/manager-service-authority.smoke.ts:71,78` already
constructs it and is the obvious reference.

## Why nobody saw it at merge time — two independent reasons

**1. The compile-time backstop the suite documents does not exist for this file.**

The header says:

> the compiler refuses a missing profile where this file is typechecked

`PRODUCERS` is a `Record<Profile, …>`, so a missing key genuinely would be a compile error — if
anything compiled it. Nothing does. `packages/core/tsconfig.json` has `"include": ["src"]`.
Asking tsc directly:

```
$ npx tsc -p packages/core/tsconfig.json --listFiles --noEmit | grep -c "packages/core/smoke"
0
$ npx tsc -p packages/core/tsconfig.json --listFiles --noEmit | wc -l
412
```

412 files in the program, none of them the suite. The clause is conditional and the condition is
false, so the documented two-layer defence is one layer. Whoever added the profile would
reasonably have expected a red compiler and got a green one.

This is worth fixing beyond this instance: either bring `packages/core/smoke` into a typecheck
project, or delete the clause. A guarantee with nothing behind it is worse than no guarantee,
because the next author trusts it.

**2. Shard abort masking hid the runtime cell.**

A shard stops at its first failing suite, so at most one casualty per shard is visible per run.
On main at `a67e704b7`, shard 1 aborted at `smoke:gate-inventory` (an unrelated ungated-suite
complaint) and never reached `smoke:eps-grant-sweep`. The sweep has been red for as long as
`remote-manager` has existed; the run simply never got to it.

It became visible only once the earlier failure was fixed on the #892 branch, at which point
shard 1 advanced and hit this. That is the mechanism working as intended, but it means the
observed shard-1 failure on main has been reporting the wrong defect since `f721478b4`, and
anyone reading "shard 1 red" as a single known baseline has been under-counting.

## Suggested scope

1. Add a `remote-manager` producer to `PRODUCERS`, modelled on the existing construction in
   `manager-service-authority.smoke.ts`, so section B actually checks it for foreign-rail leakage.
2. Make the file's compile-time claim true or remove it.

Item 2 is the one that prevents the recurrence; item 1 only fixes today's instance.

davidfarah2003:
## Summary

- add a valid `remote-manager` producer to the exhaustive EPS grant sweep
- construct the producer with its authorized owner, instance, and server-selected manager actor
- exercise its real permission rows against foreign session rails

## Verification

- `pnpm smoke:eps-grant-sweep`
- mutation proof: injected a foreign `remote-manager` EPS `in` rail and confirmed only the section B foreign-rail cell failed, then restored and reran green

Closes #894 (item 1 only).

### DIFF
--- a/packages/core/smoke/eps-grant-sweep.smoke.ts
+++ b/packages/core/smoke/eps-grant-sweep.smoke.ts
@@ -75,6 +75,11 @@ const PRODUCERS: Record<Profile, () => string[]> = {
   observer: via("observer"),
   admin: via("admin"),
   supervisor: via("supervisor"),
+  "remote-manager": () => {
+    const actor = `manager_${IID}`;
+    return rowsOf(permissionsFor("remote-manager", SPACE,
+      { ...pr, actor }, { remoteManager: { instanceId: IID, owner: DEV_OWNER, actor } }));
+  },
   provisioner: via("provisioner"),
   deprovisioner: via("deprovisioner", { deprovisionTarget: { principal: `${DEV_OWNER}.worker`, lifecycleUid: UID } }),
   "retirement-requester": via("retirement-requester", { retirementRequester: { owner: DEV_OWNER, actor: "manager", uid: UID, target: { owner: DEV_OWNER, actor: "worker", lifecycleUid: UID } } }),
--- a/packages/core/smoke/eps-grant-sweep.smoke.ts
+++ b/packages/core/smoke/eps-grant-sweep.smoke.ts
@@ -75,11 +75,9 @@ const PRODUCERS: Record<Profile, () => string[]> = {
   observer: via("observer"),
   admin: via("admin"),
   supervisor: via("supervisor"),
-  "remote-manager": () => {
-    const actor = `manager_${IID}`;
-    return rowsOf(permissionsFor("remote-manager", SPACE,
-      { ...pr, actor }, { remoteManager: { instanceId: IID, owner: DEV_OWNER, actor } }));
-  },
+  "remote-manager": () => [`manager_${IID}`, `manager_exec_${IID}`].flatMap((actor) =>
+    rowsOf(permissionsFor("remote-manager", SPACE,
+      { ...pr, actor }, { remoteManager: { instanceId: IID, owner: DEV_OWNER, actor } }))),
   provisioner: via("provisioner"),
   deprovisioner: via("deprovisioner", { deprovisionTarget: { principal: `${DEV_OWNER}.worker`, lifecycleUid: UID } }),
   "retirement-requester": via("retirement-requester", { retirementRequester: { owner: DEV_OWNER, actor: "manager", uid: UID, target: { owner: DEV_OWNER, actor: "worker", lifecycleUid: UID } } }),


davidfarah2003:
BLOCKER on `609ff4f60be778c6ede27127046868cfdfb4a79a` (verified locally with `git rev-parse HEAD`).

The new producer sweeps only `manager_<iid>`, while `permissionsFor("remote-manager", ...)` also accepts `manager_exec_<iid>`. I measured the two real constructions: their rows are identical except for the actor-scoped presence key today. However, this still leaves the second accepted arm outside the claimed exhaustive guard.

Mutation proof:
- Clean isolated `pnpm smoke:eps-grant-sweep`: 11 passed, 0 failed.
- Injecting a standing `cotal.<space>.eps.>` row into the shared remote-manager rows produced the predicted sole failure in section B: 10 passed, 1 failed, with both foreign rails attributed to `remote-manager`.
- Positive control for the unswept arm: injecting that same genuine EPS leak only when `pin.actor === manager_exec_<iid>` left the suite green at 11 passed, 0 failed, while a direct construction of `manager_exec_<iid>` confirmed the leaked row was present.

This is blocking class 2: a genuine leak in one permitted remote-manager credential can remain green. Sweep both server-selected actors, or otherwise make the cell fail for an actor-specific leak in either arm.

davidfarah2003:
Updated the sweep to construct and combine rows for both permitted remote-manager actors: `manager_<iid>` and `manager_exec_<iid>`.

Mutation proof, run independently for each arm by adding a genuine `cotal.<space>.eps.>` permission row only when that actor was selected:

- `manager_<iid>` only: section B failed with 10 passed / 1 failed and reported both foreign rails.
- `manager_exec_<iid>` only: section B failed with 10 passed / 1 failed and reported both foreign rails.
- restored implementation: 11 passed / 0 failed.

Not checked: full `smoke:ci`, full repository typecheck/build, any `:live` suite, or any live stack/broker/manager/daemon workflow. Item 2 of #894 remains untouched.

davidfarah2003:
APPROVE at `1fd740d3b3044dc4d9b5607b8a0628cab3ef14de` (verified locally with `git rev-parse HEAD`).

Delta-only re-grade against `609ff4f60`: the producer now constructs and concatenates the real rows for both permitted actors, `manager_<iid>` and `manager_exec_<iid>`. Concatenation cannot mask a leak because section B examines every returned row independently.

The original blocker is closed. Clean isolated sweep: 11 passed, 0 failed. I then injected a genuine standing `cotal.<space>.eps.>` row conditioned only on `pin.actor === manager_exec_<iid>`. Direct construction of that exec actor confirmed the injected row existed, and the sweep now produced the required sole section B failure at 10 passed, 1 failed, reporting both foreign rails under `remote-manager`. After restoration, the source diff was clean and the sweep returned to 11 passed, 0 failed.

PR #898, opened 2026-08-26
fix(web): name rejected history reads

### LINKED ISSUE #896: web-bounded-aggregation 5.6/5.7: a competing timeout pre-empts AGGREGATION_DEADLINE_MS and restores the illegible 500 {"error":"timeout"} refusal
`smoke:web-bounded-aggregation` cells 5.6/5.7 fail intermittently, and the shape of the failure is
the pre-fix defect returning rather than a slow runner.

Observed on main at `47b7d025d`, run 32964962658, shard 3, job 98165339092:

```
x FAIL  5.6 `/api/dms` REFUSES at its deadline rather than answering long after the reader left:
        {"dStatus":500,"dMs":7845}
x FAIL  5.7 and the refusal names the deadline it exceeded: {"error":"timeout"}
```

The cells expect (`implementations/web/smoke/bounded-aggregation.smoke.ts:415-416`):

```ts
ok("5.6 `/api/dms` REFUSES at its deadline rather than answering long after the reader left",
   dStatus === 503, { dStatus, dMs });
ok("5.7 and the refusal names the deadline it exceeded",
   /did not finish within 8000ms/.test(String(dBody?.error)), dBody);
```

## Why this is not simply "the runner was slow"

`AGGREGATION_DEADLINE_MS` is `8_000` (`implementations/web/src/web.ts:348`). The response came back
at **7845ms — 155ms before that deadline**. So the aggregation deadline did not fire and then get
mis-reported; it never fired at all. Something upstream produced a generic timeout first and won
the race.

The body it produced is `{"error":"timeout"}` with HTTP 500. That is exactly the shape the code
itself documents as the defect the legible-refusal work replaced:

- `web.ts:587` — "`{\"error\":\"timeout\"}` indistinguishable from data is exactly this mistake one layer up."
- `web.ts:609` — "the first produced `500 {\"error\":\"timeout\"}` after 15.94s"
- `web.ts:636` — "`{\"error\":\"timeout\"}` - five characters of cause for a panel that went blank."

So under this timing the endpoint reverts to the old illegible refusal, and a reader gets five
characters of cause instead of a 503 naming the budget it exceeded. That is the exact user-visible
harm the cells were written to prevent, which is why this is worth fixing rather than retrying.

## Evidence that it is intermittent rather than newly broken

It passed on main at `a67e704b7` and failed at `47b7d025d`. The complete diff between those two
commits is one appended line in `bin/smoke/ci-suites.txt` at index 384, which round-robins to
shard 0. Shard 3's membership and ordering are byte-identical across the two runs. Nothing in the
change can reach this suite, so the same code produced both results.

## A candidate mechanism, offered as a hypothesis and not as a finding

`packages/core/src/endpoint.ts:3281` carries a measured note: "200 x ~6 KB timed out at 5s with the
message \"timeout\"." A ~5s transport-level timeout producing the literal message `timeout` is the
right shape to be the competing deadline, though it does not by itself explain 7845ms, so I have
not established it. Whoever picks this up should confirm which timer actually fired rather than
adopting this.

## Suggested framing for the fix

The bug is not that a timeout happened. It is that **two independent budgets can expire for the
same request and only one of them produces a legible refusal.** Whichever inner timeout wins should
either surface through the same 503-with-named-deadline path, or be given a budget that provably
cannot pre-empt `AGGREGATION_DEADLINE_MS`. A guarantee that only holds when the faster timer loses
the race is the recurring defect class here: the name claims more than the code delivers.

Related: #887 (the #454 auth fallout in the same suite, a separate and already-repaired failure
mode — the ten HTTP 401 cells are gone).

davidfarah2003:
## Summary
- return named 503 refusals when direct-message history rejects before the aggregation deadline
- return named 503 refusals when channel history rejects before the aggregation deadline
- add deterministic route-level rejection coverage for both sibling endpoints and mutation cases for each

## Verification
- pre-fix: `/api/dms` returned `500 {"error":"timeout"}`
- pre-fix: `/api/channels/team00/history` returned `500 {"error":"timeout"}`
- post-fix: `pnpm smoke:web-bounded-aggregation` passed 43/43 cells
- `pnpm typecheck`
- targeted mutation proof killed both route mutations on their predicted isolated cells

Fixes #896

### DIFF
--- a/.changeset/tidy-reads-refuse.md
+++ b/.changeset/tidy-reads-refuse.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/web": patch
+---
+
+Name failed direct-message and channel-history reads in dashboard 503 responses.
--- a/implementations/web/smoke/bounded-aggregation.smoke.ts
+++ b/implementations/web/smoke/bounded-aggregation.smoke.ts
@@ -168,6 +168,7 @@ const releaseBroker = teardownOnSignal(broker, store);
 let link: { close(): void } | undefined;
 let link2: { close(): void } | undefined;
 let webChild: ReturnType<typeof spawn> | undefined;
+let rejectingWebChild: ReturnType<typeof spawn> | undefined;
 try {
   let up = false;
   for (let i = 0; i < 80; i++) { if (await isReachable(SERVER)) { up = true; break; } await wait(150); }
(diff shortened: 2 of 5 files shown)


davidfarah2003:
APPROVE

Graded exactly `ed041bf59ff0640b6e17a4c9fdb47dd0f284f66f` in the detached review tree. PR #898's head also resolved to that SHA at verdict time, and the tree was clean after mutation restoration.

Evidence:
- Fresh `pnpm install --frozen-lockfile`, `pnpm build`, and `pnpm typecheck` passed.
- `pnpm smoke:web-bounded-aggregation` passed 43/43 before mutation and again after restoration.
- B10, removing only the `/api/dms` rejection naming, was killed by cell 5.1. Cell 5.2 remained green.
- B11, removing only the channel-history rejection naming, was killed by cell 5.2. Cell 5.1 remained green.
- The demonstrations reach the shipped `web()` entry point through `run-web.mts`, not a hand-built handler.
- The shared harness injection is opt-in through `COTAL_WEB_SMOKE_REJECT_HISTORY=1`. With it unset, `smoke:web-body-cap` passed 30 cells, `smoke:web-channel-alias` passed 11, and `smoke:web-history-limit` passed 39.

Blocking class 3 is closed. Inside each new `try`, the only awaited operation is `within()` over `dmHistory()` or `channelHistory()`, followed by response serialization. Repository-wide enumeration shows `BadRequest` is created only by channel-name and limit parsing, both before these `try` blocks. `PayloadTooLarge` is created only by request-body reading on the POST delete path. The endpoint history methods and their `streamHistory` path do not create either web-local error type. Therefore the unconditional catches do not make a reachable 400 or 413 become 503.

Residual: `patch` is the correct changeset level. This corrects the documented failure response from an accidental generic 500 to the intended named 503 and does not break the successful response contract.

Not checked: any `:live` suite, stack verb, persistent broker/manager/delivery/web stack, full `pnpm smoke:ci`, browser UI behavior, or platforms other than this macOS arm64 host.

davidfarah2003:
Non-blocking residual: section 6 does not deterministically select the deadline ending it names. Cells 6.6 and 6.7 cannot both pass on the rejection path: 6.6 accepts any 503, but 6.7 requires the mutually exclusive `did not finish within 8000ms` body, while an inner rejection produces `direct messages: the read failed: <cause>`. Both local review runs passed 43/43, so this host reached `LATE`; reported CI wall times below 8000ms are consistent with the inner read rejecting first there. The selected ending is therefore environment/load-sensitive. This does not weaken cells 5.1/5.2 or block this fix, but it is worth a separate issue so the deadline-path arm cannot turn a correct rejection-path fix into a shard failure.

PR #899, opened 2026-08-26
docs(core): mark the load-bearing conjunction in the EPS grant sweep

davidfarah2003:
`packages/core/smoke/eps-grant-sweep.smoke.ts` is correct, and it is correct as a **conjunction**
that nothing in the file marks. This records the dependency. Comments only; no assertion changes.

## What is unmarked

Section C's first four cells are named for a plane and do not test one:

```ts
c("session-caller PUBLISHES its own `in` rail",  caller.some((r) => subjectMatches(r, inMine)),  caller);
c("session-caller SUBSCRIBES its own `out` rail", caller.some((r) => subjectMatches(r, outMine)), caller);
```

`caller` comes from `rowsOf`, which returns `[...pub.allow, ...sub.allow]`. Read alone, each of those
four accepts the opposite plane — "PUBLISHES its own `in`" is satisfied by a subscribe grant on `in`.

They mean what their names say only because the asymmetry cell below them asserts
`!callerPub.includes(outMine)` and `!servingPub.includes(inMine)` against a pub-only accessor. Those
exclusions leave exactly one plane able to supply each match above.

So loosening or deleting that one cell would take the meaning out of four others without touching
them and without reddening anything. There is no signal in the file that it carries that weight, and
this is the kind of thing that is free to write down today and not reconstructible later.

## Also recorded

- **On `rowsOf`**: merging the planes is *required* for section B, which asserts the ABSENCE of a
  grant reaching a foreign session rail — a leak is a leak whichever plane carries it. It is unsound
  for a positive capability claim, because a JetStream API subject is a request the client
  *publishes*, so a merged check can pass on an unrelated subscribe grant.
- **Why the exact `.includes()` in the asymmetry cell is deliberate** rather than an inconsistency
  with `subjectMatches` elsewhere. It is stricter, not weaker: a wildcard pub grant fails its
  positive halves (`callerPub.includes(inMine)`) instead of slipping past its negative ones. That is
  the specific regression worth fearing, because `sessionServingPermissions` in `provision.ts`
  records that this design replaced a standing writer holding `eps.<endpoint>.*.<epoch>.{in,out}`.

## Verification

`eps-grant-sweep`: **11 passed, 0 failed** on this branch. The change adds no assertion and removes
none, so the suite's behaviour is unchanged by construction; the run confirms the file still parses
and executes.

### DIFF
--- a/packages/core/smoke/eps-grant-sweep.smoke.ts
+++ b/packages/core/smoke/eps-grant-sweep.smoke.ts
@@ -58,6 +58,15 @@ function subjectMatches(pattern: string, subject: string): boolean {
   return p.length === s.length;
 }
 
+/** BOTH PLANES, DELIBERATELY, and only sound for the question §B asks. §B asserts the ABSENCE of any
+ *  grant reaching a foreign session rail, and a leak is a leak whichever plane carries it, so merging
+ *  is required there rather than a shortcut.
+ *
+ *  It is NOT sound for a question about a specific plane. A JetStream API subject is a request the
+ *  client PUBLISHES, so a check that merges the planes and then claims "X may call Y" can pass on an
+ *  unrelated subscribe grant. §C's first four cells read against this merged list and are named for a
+ *  plane; what actually pins them is the exclusion in the cell below them — see the note there before
+ *  reusing this helper for a positive capability claim. */
 const rowsOf = (perm: unknown): string[] => {
   const p = perm as { pub?: { allow?: string[] }; sub?: { allow?: string[] } };
   return [...(p.pub?.allow ?? []), ...(p.sub?.allow ?? [])];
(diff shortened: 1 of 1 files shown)


davidfarah2003:
APPROVE — graded 2a92e3beded168051f6b9681e4f9538d7be821f1.

Verified the diff is +25 comment lines only with no assertion change. The baseline `pnpm smoke:eps-grant-sweep` result is 11 passed / 0 failed after install and core build with COTAL_* unset.

I falsified the new comments against the real smoke entry point: swapping both session profiles onto the opposite planes left all four merged `rowsOf` cells green while the asymmetry cell failed 10/1; weakening that cell then made the wrong-plane fixture pass 11/0. A wildcard caller pub grant matched the positive rail under `subjectMatches`, but failed the exact `.includes(inMine)` positive half; the suite failed 9/2, including both the foreign-rail sweep and asymmetry cell. The comments accurately describe the load-bearing relationship and the stricter exact-match guard.

PR #921, opened 2026-08-27

davidfarah2003:
## Summary

This branch makes every persona-facing CLI path resolve from the target mesh root instead of silently mixing it with the current directory.

- `cotal personas` list, show, edit, new, remove, and completion read and write the resolved mesh catalog.
- spawn and send completion use that same catalog.
- `cotal setup` seeds the catalog a following spawn reads, prints the absolute destination, uses a named cwd fallback only when no mesh exists, and refuses ambiguity.
- `cotal status` names the folder and spawn catalogs when they differ. A green `default` now means the file passes the same shipped loader spawn uses. A present but invalid file is reported as invalid.
- agent-profile `cotal mint` resolves one root for the persona ACL, signing authority, and default credential storage. If cwd trust and target trust name different spaces or accounts, mint refuses before writing and names both roots.
- docs and changesets cover the target-selected catalog, setup behavior, status validation, and mint authority rule.

## Security behavior

Before the final mint guard, a caller standing under trust root A could select mesh root B, read B's persona ACL and role, then receive a credential signed or stored under A. That combined policy and authority from roots the operator had not chosen as one trust boundary.

Mint now resolves the target once. Persona policy, signing material, default storage, and optional provisioning all use that target root. Competing cwd trust is never silently preferred or ignored. A disagreement is a refusal that names both roots.

## Tests

All commands below passed on the final local tree:

```text
pnpm build
pnpm typecheck
pnpm smoke:setup-destination       20 checks passed
pnpm smoke:persona-root            35 checks passed
pnpm smoke:status-persona-root     18 checks passed
pnpm check:docsbundle
pnpm changeset status
pnpm check:shard-stability 66d3f882 HEAD
  suites 420 -> 423
  pre-existing suites changing shard: 0 of 420
git diff --check
```

Mutation proof runs were red on the required named cells:

```text
status-persona-root: 2 of 2 KILLED
persona-root:        5 of 5 KILLED
setup-destination:  2 of 2 KILLED
```

The status fixture's positive control previously asserted that an existence-only `# default` file was green even though `loadAgentFile` rejects it. It now uses valid frontmatter and proves that a loader-valid default is green, while a separate cell proves the old invalid shape is not green.

The three lane smokes are offline and sandbox their own `COTAL_HOME`. No live broker or stack lifecycle command was used for this fold round.

## Release

The first-party packages are currently at 0.36.0. The CLI minor changesets participate in the fixed group, so the next versioning run advances that group on the 0.37.0 line. The mint changeset names the target-root ACL and signing rule and the new conflicting-trust refusal.

## Accepted residual

Commit `5be4b34e` contains non-ASCII punctuation in its published commit message. It is left unchanged because correcting it would require rewriting pushed history and force-pushing, which this repository forbids. All new commit messages and this PR body use plain ASCII.


davidfarah2003 (commented):

- ci-suites is clean: the three new names (`smoke:setup-destination`, `smoke:persona-root`, `smoke:status-persona-root`) sit at the absolute tail of `bin/smoke/ci-suites.txt` with the rationale comment attached to them. The only mid-file edits are two blank lines (183, 548), which `bin/smoke/ci-suites.mjs:29` strips before `shard.mjs` indexes, so shard assignments are untouched. The body's own `suites 420 -> 423` matches.
- The fallback removal is structural, not disciplinary. `implementations/cli/src/lib/personas.ts` dropped `root = cotalRoot()` from all five exports *and* stopped importing `cotalRoot` entirely, so a forgotten root is TS2554 at every call site rather than a silent cwd read. That is the right shape for rail (d).
- The one surviving cwd path is narrow and loud: `seedDestinationFor` in `implementations/cli/src/commands/setup.ts` catches only `code === "no-meshes"` and rethrows everything else, so `ambiguous-target` still refuses; `announceSeedDestination` prints both roots when the mesh root differs from the cwd root. `personasComplete` fails closed to `{ items: [], directive: "nofiles" }`.
- mint now holds ACL read, signing key and secret store to one resolved root, and refuses with an explicit both-roots message when the target and cwd authorities disagree (`implementations/cli/src/commands/mint.ts`). Three of the five `persona-root` mutation cells drive exactly that refusal — including the `&& process.pid < 0` cell that disables it. `--signer` deliberately keeping cwd semantics is consistent with the body.
- Evidence reconciles against the shipped artifacts exactly: 5 + 2 + 2 = 9 mutations in `persona-root.mutations.json`, `setup-destination.mutations.json`, `status-persona-root.mutations.json`, matching the claimed 9 kills; check counts 35 / 18 hold statically and setup-destination's 17 static `check(` sites reach the claimed 20 through the two-iteration `[["alpha",alpha],["beta",beta]]` loop. Docs updated across `agent-files.md`, `cli.md`, `identity-and-auth.md`, `setup-internals.md` with the bundle regenerated.
- One gap worth naming, not a blocker: `announceSeedDestination` is driven by no cell. Every smoke call site is `seedDefaultForSetup(here, false)`, so the three announce shapes are asserted on the returned `SeedDestination` fields (`dest.dir`, `fell.reason`), never on printed output. The claim "says it fell back and why" is verified one layer below where the user reads it.

davidfarah2003:
Superseded by #921: this PR's pull_request events stopped minting CI runs after the branch history rewrite (close/reopen and an empty-commit retrigger both produced only CodeQL), so it cannot be validated here. Same branch and tree over there; the APPROVE-quality review in this thread applies to that tree.

davidfarah2003:
Correction: the successor PR is #1085 (the earlier comment's number was scraped from an error and is self-referential).

Issue #967, opened 2026-08-29
the live job has not finished in twelve consecutive runs: it times out at 25 minutes and GitHub badges that as cancelled


davidfarah2003:
The `live` job has not finished on any of the last twelve CI runs. It is not being superseded: it
runs to `timeout-minutes: 25` and stops. GitHub badges a timed-out job `cancelled`, which reads as
"a newer run replaced this one" rather than "this job did not finish", so the failure is invisible
in the checks list and in `gh pr checks`.

This is the same defect class as #897, on a different job, and unlike #897 it is still current.

## The measurement

`live` conclusion and wall time across the last twelve completed CI runs:

| run | conclusion | duration |
|---|---|---|
| 33220057022 | cancelled | 25m15s |
| 33219897171 | cancelled | 25m15s |
| 33219527255 | cancelled | 22m23s |
| 33219443753 | cancelled | 25m16s |
| 33217394490 | cancelled | 13m36s |
| 33216727065 | cancelled | 25m14s |
| 33216717392 | cancelled | 25m15s |
| 33216636261 | cancelled | 25m16s |
| 33216366010 | cancelled | 25m15s |
| 33215950948 | cancelled | 6m20s |
| 33205995896 | cancelled | 25m16s |
| 33204835100 | cancelled | 25m17s |

Nine of twelve land in a three second band at 25m14s to 25m17s, against `timeout-minutes: 25` at
`.github/workflows/ci.yml:148`. That is the limit firing, not a coincidence of scheduling. The three
short ones (6m20s, 13m36s, 22m23s) are the genuine supersessions, and they are the only ones whose
duration varies.

Twelve for twelve is the number that matters.

**Correction, see the comment below:** I originally added "whatever the live suite would have caught,
it has not been catching it" here, and that is wrong. Step-level evidence shows the job gets through
almost everything and dies on `smoke:herdr-e2e:live`, so the uncovered set is three suites rather
than the whole job. The legibility problem this issue is about is unaffected.

## Why the badge makes this worse

A red job gets read. A cancelled job gets skipped, because cancellation is normally something the
system did on purpose when a newer push arrived. So the checks list says nothing is failing while a
whole job never ran, and a reviewer scanning for red sees none. Two merge decisions on this
repository in the last day treated a cancelled `live` as fail-fast collateral rather than as a job
that never finished, which is the exact misreading the badge invites.

## What would fix it

Two separable things, and the first matters more than the second.

[...]

PR #970, opened 2026-08-29
fix(smoke): release the manager's pipes so herdr e2e can exit

davidfarah2003:
The `live` job has not finished in twelve consecutive CI runs, including on the current `main` tip. It
is not being superseded and it is not short of time. `smoke:herdr-e2e:live` passes and then refuses to
exit.

## The measurement

Step timings for `live` on `main` at `ae9f6164` (run 33220279945, job 99014587114). Everything before
step 17 costs seven minutes in total; step 17 consumes the remaining eighteen and is killed.

| step | | elapsed |
|---|---|---|
| 1-11 | set up, install, build | 48s |
| 12 | core boundary, mesh preflight & resolution | 5m29s |
| 13 | manager lifecycle e2e | 14s |
| 14 | runtime smokes (tmux, orca, herdr) | 23s |
| 15-16 | herdr soak instrument, endurance soak | 5s |
| 17 | **herdr e2e** | **18m13s, killed** |
| 18-19 | cmux secret-hygiene, OpenCode cooperative-stop | never ran |

From that job's log:

```
23:43:12.977  $ tsx bin/smoke/herdr-e2e-live.smoke.ts
23:43:19.336    ✓ the agent process exits with its pane
23:43:19.336  25 passed, 0 failed  (25 cells ran)
00:01:25.647  ##[error]The operation was canceled.
```

All 25 cells in 6.7 seconds, zero failures, then 18 minutes of nothing. The same job's teardown reports
`Terminate orphan process: pid (11877) (node)`.

## Cause

herdr's server is spawned `detached` by the manager. That is the claim this suite exists to prove: the
agent outlives the manager. The detached server also inherits the manager's stdout/stderr, which are
this suite's pipes, so SIGKILLing the manager at L174 does not close the write ends. The read ends stay
open, the event loop never drains, and the process hangs.

Only a **passing** run hangs. Every failure path exits explicitly (L54, L60, L79, L139, L192) and the
success path falls off the end. That asymmetry is why the step read as slow rather than as leaked:
there was no failing observation to compare it against.

## Fix

Record the pipes at creation time and release them in teardown, by that exact identity, which is the
convention the suite's own header states and which it already follows for the broker pid, the manager
pid and the scratch dir.

## Proof

The real suite, same machine and same broker, one difference between the two runs:

```
=== CONTROL: pristine suite ===
25 passed, 0 failed  (25 cells ran)
exit=124        # killed by `timeout 150`
=== FIXED ===
25 passed, 0 failed  (25 cells ran)
exit=0          # 6 seconds
```

The control reproduces the CI failure locally, so this is not a fix for a symptom read off a log.

## The `timeout` on the step

A hang inside a step is badged `cancelled` on the job, which reads as "a newer run replaced this one"
rather than "this did not finish". That is how a passing suite held the gate red for twelve runs
without anyone reading the log. Bounding the step makes the next occurrence fail red and name itself.

The general form of that problem is not fixed here and stays open on #967: any live step can still hang
invisibly, and this only bounds the one that demonstrably does.

## Scope

Two files. `bin/smoke` is not in the published `files` of `bin/package.json`, so no shipped code
changes and no changeset is needed.

Refs #967.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,7 +208,10 @@ jobs:
       # The integration the unit suite cannot reach: real broker + real Manager + a real agent in
       # presence, then SIGKILL the manager to prove the agent outlives it.
       - name: herdr e2e (real broker + manager + agent, manager-kill survival)
-        run: pnpm smoke:herdr-e2e:live
+        # A hang here reads as `cancelled` on the job, which looks like a supersession rather
+        # than a failure — that is how a PASSING suite held this gate red for twelve runs. Bound
+        # the step so the next one fails red and names itself.
+        run: timeout 600 pnpm smoke:herdr-e2e:live
       - name: cmux launcher secret-hygiene smoke
         run: pnpm smoke:cmux
       - name: OpenCode cooperative-stop smoke test
--- a/bin/smoke/herdr-e2e-live.smoke.ts
+++ b/bin/smoke/herdr-e2e-live.smoke.ts
@@ -24,6 +24,7 @@ import { randomUUID } from "node:crypto";
 import { execFileSync, spawn } from "node:child_process";
 import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync, existsSync } from "node:fs";
 import { createServer, type AddressInfo } from "node:net";
+import type { Readable } from "node:stream";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { createSpaceAuth, serverConfig, setupSpaceStreams, mintCreds, newIdentity, isReachable } from "@cotal-ai/core";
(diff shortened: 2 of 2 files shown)


davidfarah2003:
Closing this. #968 fixed the same hang while this was in CI, and its diagnosis is better than mine in a
way that matters, so this should not be merged even rebased.

**What I got wrong.** I said the detached herdr server inherited the manager's stdout and stderr, so
SIGKILLing the manager never closed the write ends. The actual cause is one level earlier: `child.pid`
in that suite is the **tsx wrapper**, not the Manager. The Manager runs in a separate process that the
child reports as `managerPid`, which the suite never read. So the suite was SIGKILLing a wrapper and
leaving the real manager alive and orphaned, and that orphan held the pipes.

**Why that difference is not cosmetic.** The suite's headline cells are "the manager process is really
dead" and "the agent SURVIVED the manager being killed". Both were measuring a wrapper's death, so the
one claim this suite exists to prove was never actually being tested. That is a false green, which is
a worse defect than the hang. My change would have made the suite exit cleanly while leaving that false
green in place, and by removing the hang it would have removed the only symptom that led anyone to
look at this suite at all.

**And it explains the timing, which I never asked about.** #968 identifies the trigger: #925 changed an
orphaned manager to keep serving rather than end its own process over a lost liveness lease. Before
that, the orphan died on its own, the pipes closed, and the suite's reliance on natural event-loop exit
went unnoticed. That is why the hang starts on 2026-08-28 and not earlier. I documented "twelve
consecutive runs" without ever asking what changed at run one, which is the question that would have
led to the real cause.

**The methodological lesson, since it generalises.** I ran a control pair on the real suite: pristine
hung and was killed at 150s, patched exited 0 in six seconds, same machine and same broker. I read that
as confirming the diagnosis. It does not. A control pair proves the fix removes the **symptom**. A wrong
fix that removes the same symptom produces an identical control pair. #968 did the control I skipped,
running pre-#925 main (`2dc2c9063`) on the same machine and observing a natural exit, which is what
actually isolates the cause. Its handle inventory also names `ProcessWrap` alongside the two
`PipeWrap`s; my standalone reproduction only ever produced the two pipes, because the model I built had
no equivalent of a surviving real manager, so it reproduced my hypothesis rather than the bug.

**One piece here that #968 does not cover**, offered as a note rather than a reopened PR: this branch
also bounded the CI step with `timeout 600`. #968's explicit `process.exit` means this particular suite
can no longer hang, so the bound is redundant for it. The general problem stays open on #967: a hang in
any other step still consumes the job's remaining budget and is still badged `cancelled`, which reads
as a supersession rather than as a job that did not finish. That wants a job-level fix, not a
per-step timeout, and #967 is the right place for it.

No objection to #968. It is the correct fix.

PR #980, opened 2026-08-29
ci: queue main runs instead of evicting them

### LINKED ISSUE #976: 45% of merges to main get no CI: the concurrency queue evicts pending runs, and they badge as cancelled with zero jobs
Thirteen of the last twenty-nine completed CI runs on `main` were created with **zero jobs** and
badged `cancelled`. Those commits have no CI result of any kind. This is the opposite of what the
concurrency block was configured to guarantee, and its own comment states the goal it is missing.

## The configuration and its stated intent

`.github/workflows/ci.yml:19`

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

The comment directly above it says why `cancel-in-progress` is conditional:

> Deliberately not unconditional: a main commit's CI result is the record for that commit, not a
> preview of the next, and a cancelled run leaves a merge ungradable, neither red nor green.

The reasoning is right. The outcome is not. Leaving `cancel-in-progress` false on `main` does not
serialize main's runs so each one eventually gets its turn. GitHub holds at most one *pending* run
per concurrency group, so when a third push arrives while one run is in flight and one is queued,
the queued one is cancelled rather than the new arrival waiting behind it.

Under a rapid sequence of merges, only the first and the last commit of a burst get CI. Everything
between is evicted, gets zero jobs, and is badged `cancelled`.

## The measurement

One burst, fully instrumented:

| run | head | created | jobs | outcome |
|---|---|---|---|---|
| 33225555092 | `2d3ffa3f7` | 01:07:49 | 7 | ran, last job finished 01:45:06 |
| 33225849491 | `e3de2919c` | 01:14:07 | **0** | cancelled at **01:31:01** |
| 33226611986 | `d58d507c3` | 01:31:00 | 6 | acquired jobs 01:45:09 |

`e3de2919c` was cancelled one second after `d58d507c3`'s run was created. It was not superseded by a
newer push to the same ref in the sense the comment has in mind, and no human cancelled it: the
triggering actor is the pusher and the event is `push`. It was evicted from the queue by the arrival
behind it.

`d58d507c3` then started three seconds after `2d3ffa3f7`'s final job completed, which confirms the
group is what was holding both of them.

Frequency over the window 2026-08-28T06:47Z to 2026-08-29T01:31Z, twenty-nine completed `main` runs:

```
33225849491  e3de2919c
33204050462  81b4e1f99
33199345820  1f7c79f11
33197239261  110d93b0b
33196619129  8b4edf391
33196359923  96edee12f
33194952595  18af66cc6
33160406600  a1382a66d
33157447948  7635c47d8
33149497750  1da187267
33149202253  7c30d1610
33149156101  f1e60d2c3
33149138785  8e92380fa
```

Thirteen of twenty-nine, about 45% of merges to `main`, tested by nothing.

## Why it landed on the worst possible commit

`e3de2919c` is the merge of #968, whose entire content is a fix to the `smoke:herdr-e2e:live` suite
that #967 tracks. The one commit whose CI result most needed reading is the one that got none. The
fix reached `main` unverified and stayed unverified until a later, unrelated merge happened to carry
it into a run that survived.

## Three causes, one badge

This compounds #967. A `cancelled` badge on a CI run now means any of:

1. a job hit its own `timeout-minutes` (#967)
2. a run never acquired jobs because the queue evicted it (this issue)
3. someone cancelled it

Nothing in the checks list separates them, and only the third is what a reader assumes. A reviewer
scanning a merge sees `cancelled`, reads it as housekeeping, and moves on. Case 2 is the worst of
the three because the commit was never built, so there is no log to look at even if someone did.

## What would fix it

Give each `main` commit its own concurrency group, so no main run ever queues behind another and
none can be evicted:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name == 'pull_request' && 'pr' || github.sha }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

PR behaviour is unchanged: `github.ref` is `refs/pull/N/merge`, the suffix is constant, so a new
push to a PR still supersedes the run in flight. Push events get a per-sha group, which is what "the
record for that commit" requires.

The cost is real and should be stated: main merges can then run concurrently, so a burst of merges
uses more runners at once instead of serializing. That is the price of the guarantee the comment
already claims to provide.

A second, independent improvement, worth doing whether or not the group changes: a run with zero
jobs should not be reportable as `cancelled` alongside runs that actually executed. Whatever
mechanism #967 grows for distinguishing a timeout from a supersession should cover this case too.

Same defect family as #967, #897, #881, #874, #868, #833: a green or a neutral badge standing in for
evidence that does not exist.

### LINKED ISSUE #990: #980 fixes pending-run eviction in ci.yml only: windows.yml, changesets.yml and installer.yml still evict
#980 fixes the pending-run eviction described in #976, but only for `ci.yml`. Three other workflows fire on pushes to `main` and carry the same exposure, in two different shapes. Surfaced while reviewing #980, then confirmed by sweeping every workflow at `origin/main`.

## The sweep

| workflow | fires on push to main | concurrency form | evicts? |
| --- | --- | --- | --- |
| `ci.yml` | yes | long form, `queue` conditional | no, fixed by #980 |
| `windows.yml` | yes | long form, no `queue` | **yes** |
| `changesets.yml` | yes | short form (bare string) | **yes** |
| `installer.yml` | yes, paths-filtered | short form (bare string) | **yes** |
| `docs.yml` | yes, paths-filtered | none | no, runs are independent |
| `github-releases-to-discord.yml` | no, `release` event | none | no |

## windows.yml

`.github/workflows/windows.yml:33-35` is the exact pre-#980 shape:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Its own comment two lines above says "Same rule as ci.yml, and for the same reason". That lockstep was deliberate, and #980 breaks it without updating either side to say so. The comment now describes an arrangement that no longer exists.

This lane is marked REQUIRED, so a main commit whose Windows run is evicted has no Windows verdict while still showing a green `ci-ok`.

## changesets.yml and installer.yml

Both use the short form, a bare string rather than a mapping:

```yaml
concurrency: ${{ github.workflow }}-${{ github.ref }}
```

The short form takes the default `cancel-in-progress: false` and the default `queue: single`, so it behaves exactly like the `ci.yml` case #976 measured: one in-flight run, one pending slot, and a third arrival cancels the pending one rather than joining the line.

`changesets.yml` is the release workflow. An evicted run there means a merge to `main` that should have staged a version PR silently did not, and the badge reads `cancelled`, which is the same badge a human tidying up produces. That is the most expensive instance of this class in the repo, not the least.

## Suggested fix

Apply the same conditional `queue` to `windows.yml`, which keeps the stated lockstep with `ci.yml`:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
  queue: ${{ github.event_name == 'pull_request' && 'single' || 'max' }}
```

For `changesets.yml` and `installer.yml`, expand the short form to a mapping and set `queue: max`. Neither sets `cancel-in-progress: true`, so the validation error that forces the conditional in `ci.yml` does not apply and `queue` can be a literal.

Worth deciding at the same time: whether the lockstep between `ci.yml` and `windows.yml` should be asserted by something other than a comment, since it has now drifted once.

## Known limit of the fix, from the #980 review

`queue: max` caps at 100 pending. The 101st pending run is cancelled, so this raises the ceiling rather than removing the eviction class. Worth recording as accepted rather than discovered later. A burst of N merges also delays HEAD's result by roughly (N-1) run durations, and the smoke job's timeout is 40 minutes.

davidfarah2003:
Fixes #976. Fixes #990.

A concurrency group holds at most ONE pending run. With a run in flight and a run queued, a third push CANCELS the queued one rather than making it wait, so `cancel-in-progress: false` does not serialize merges to `main`, it evicts the middle of a burst. The evicted run finishes with zero jobs and the commit ends up with no CI result of any kind, badged `cancelled`, which is also what a run somebody tidied up looks like.

Measured on this repo: **13 of 29 consecutive `main` runs had zero jobs.** The timing is unambiguous, the middle run died one second after the third run was created and the third started three seconds after the first finished, on `push` events with the pusher as actor, so nothing human cancelled them.

`queue: max` is the part that makes runs wait, FIFO, up to 100 deep.

## Scope

All four workflows a push to `main` can start, not just CI. The three siblings were missed originally because two of them spell the defect differently.

| workflow | before | after |
| --- | --- | --- |
| `ci.yml` | long form, no `queue` | conditional `cancel` + conditional `queue` |
| `windows.yml` | long form, no `queue` | conditional `cancel` + conditional `queue` |
| `changesets.yml` | short form (bare string) | `cancel-in-progress: false` + `queue: max` |
| `installer.yml` | short form (bare string) | conditional `cancel` + conditional `queue` |
| `docs.yml` | no concurrency block | unchanged, no group means nothing can evict |

`windows.yml` carried the identical shape while its own comment claimed to follow ci.yml's rule, and it is the REQUIRED lane, so an evicted run leaves a commit with no Windows verdict while `ci-ok` still reports green. `changesets.yml` and `installer.yml` used the short `concurrency: <string>` form, which is not a neutral minimal choice: it selects `cancel-in-progress: false` and `queue: single`, the evicting defaults. `changesets.yml` is the release workflow, so an evicted run there is a merge that silently staged no version PR.

`changesets.yml` takes literals rather than the conditional expression because it has no `pull_request` trigger, and an expression there would carry a branch no event can reach.

## One deliberate behaviour change beyond the queue fix

`installer.yml`'s short form gave `cancel-in-progress: false` on BOTH events, so a PR revision did not supersede its predecessor and the container matrix ran twice over. It now supersedes on `pull_request`, matching every other workflow. Called out because it is scope beyond the eviction defect.

## Why `queue` is conditional rather than a literal

`queue: max` together with `cancel-in-progress: true` is a workflow validation error, so `queue` has to track `cancel-in-progress` wherever that is event-conditional.

`queue` accepting an expression is not documented either way. It was verified by making this PR the test: a `pull_request` event resolves `cancel-in-progress` to `true`, which is exactly the illegal combination, so an invalid form would have failed here and never reached `main`.

## New gate: `smoke:workflow-concurrency`

The property this protects is only observable on `main` after a merge, so it cannot be tested by the pull request that changes it. That is precisely why it needs a static guard, and why one workflow being fixed while three siblings drifted went unnoticed.

The check asserts the population first, so it cannot pass by finding nothing; evaluates both concurrency keys per event rather than per file, since a per-file glance cannot see a combination that only arises for one event; and refuses an expression form it does not recognise instead of skipping it.

Its first run failed `docs.yml` for having no concurrency block. The rule was wrong, not the workflow: a workflow with no group cannot evict, because nothing is ever pending behind anything. The rule now distinguishes the two cases and asserts both counts, so a group removed elsewhere shows up as a moved number rather than as one fewer cell.

### Mutation proof

Committed as `bin/smoke/mutations/workflow-concurrency.json` and reproducible:

```
pnpm mutation-proof --config bin/smoke/mutations/workflow-concurrency.json
```

```
baseline green (27 progress marks)

KILLED       changesets.yml goes back to the short concurrency form
  red, and named: changesets.yml: uses the mapping form . 21 marks (baseline 27)
KILLED       windows.yml queue becomes a literal max, pairing with cancel-in-progress true on pull_request
  red, and named: windows.yml: on pull_request, not (queue=max AND cancel-in-progress=true) . 26 marks (baseline 27)
KILLED       installer.yml queue uses an expression form the reader does not recognise
  red, and named: installer.yml: `queue` is a form this check understands . 23 marks (baseline 27)

```

The harness requires a clean tree, baselines the command green before mutating since an already-red suite grades every mutation as killed for the wrong reason, requires each red to be NAMED by the expected cell, and verifies its own restore.

The first run of this fixture returned INCONCLUSIVE on all three: red and correctly named, and refused for having stopped early. The suite printed its completion marker only on success, so the only evidence it had reached the end was the line it withheld on failure. The marker means the run finished, not that it passed, and it now prints on both outcomes.

`mutation-proof` prints its own scope caveat, that killing a mutation shows the suite DEPENDS on the mutated code rather than that a real entry point reaches it. Here the two coincide: the suite reads the actual `.github/workflows/*.yml` files from disk, and the mutations edit those same shipped files, so there is no hand-built input standing between the test and the artifact.

### Limits, named in the suite rather than left to be rediscovered

Job-level `concurrency:` is unread, only the top-level block. `on:` detection is indentation-driven, so an unfamiliar trigger shape is misclassified. The PR-supersede half of the rule is unasserted. None can produce a pass while a workflow is broken: the first two land on the population counts and fail loudly, and the third is a scope limit of the test.

## ci-suites ordering

The entry was appended AFTER merging current `main`, not before. `bin/smoke/ci-suites.txt` carries a union merge driver, which preserves both sides of an append but not their order, so appending from a branch that is behind `main` reshuffles the tail and moves existing suites between shards. This branch was five lines behind. `check:shard-stability` reports 411 to 412 suites with **0 of 411 pre-existing suites changing shard**.

## What the green check here does NOT prove

A `pull_request` run evaluates the PR branch of every expression, so it never exercises the `main` grouping. It establishes that `queue` accepts an expression, because the run would not have started otherwise. It cannot establish the eviction fix, which is only observable after merge. Do not read the green check as evidence that runs now queue.

## Limit, accepted rather than discovered later

`queue: max` caps at 100 pending. The 101st pending run is cancelled, so this raises the ceiling rather than removing the eviction class. A burst of N merges also delays HEAD's result by roughly (N-1) run durations, and the smoke job's timeout is 40 minutes.

## The one workflow this PR cannot exercise

`ci.yml`, `windows.yml` and `installer.yml` all fire on `pull_request`, and all three started jobs on this PR, so their new concurrency blocks are proven to parse. `changesets.yml` fires only on `push` and `workflow_dispatch`, so its block is not exercised until it lands on `main`, and it is the release workflow.

What that risk is bounded by: the mapping-with-`queue` shape is the same one the three exercised workflows use and is therefore proven valid, and `changesets.yml` fills it with plain literals rather than expressions, which is a strict subset of what those three prove. All six workflow files were additionally YAML-parsed with a positive control confirming the parser rejects malformed input. What none of that covers is GitHub's own schema validation of that specific file, which only a push to `main` performs.


davidfarah2003:
CI at `37eb117a3`: 14 pass, 3 pending, 2 fail. **The two failures are `installer-ok` and `matrix`, and they are not caused by this PR.** Filed separately as #994.

## What fails

The onboarding matrix case `pinned-version`, which installs `cotal-ai@0.14.1` and asserts the result reports its version. It reports none.

```
  upgrade        node:22-bookworm-slim PASS
  pinned-version node:22-bookworm-slim FAIL
    installed cotal did not report a version:

  29 passed, 1 failed
```

`upgrade` installs the current version and then upgrades, and it passes. So `install.sh` works. What has rotted is the specific old version pinned as a literal at `install/test-matrix.sh:61`.

## Why it is not this PR

This PR changes eight files:

```
.github/workflows/changesets.yml
.github/workflows/ci.yml
.github/workflows/installer.yml
.github/workflows/windows.yml
bin/smoke/ci-suites.txt
bin/smoke/mutations/workflow-concurrency.json
bin/smoke/workflow-concurrency.smoke.ts
package.json
```

Neither `install.sh` nor `install/test-matrix.sh` is among them, and the only change to `installer.yml` is its `concurrency` block. The failing case runs `install.sh` inside a Docker container against a package pulled from the npm registry, so nothing in this diff is reachable from it.

## Why it appeared now

The Installer workflow fires only on changes to `install.sh`, `install/**`, or its own file. Its previous run was **2026-08-01** and passed. This PR edits `installer.yml`, which matched that path filter for the first time in twenty-eight days.

That is the same shape as the defect this PR fixes: the badge was green because nothing ran, not because anything passed. #994 argues the dormancy is the more important half.

## One unplanned piece of evidence in this PR's favour

Three heads were pushed in quick succession. The middle one's Installer run (`5e3f04df5`) came back **`cancelled`** rather than running to completion, which is the `cancel-in-progress` change working on a real `pull_request` event. Before this PR, `installer.yml` used the short concurrency form and would not have superseded.

It is the first empirical confirmation of any behaviour in this PR beyond the fact that the expressions parse. It still says nothing about the `main` queueing half, which remains unobservable until after merge.

davidfarah2003:
Head moved to `ad9c75e2968e070981072aab53b1ac5b48c79133`. Recording it here because the mesh is currently unreachable, so this is the durable channel for anyone mid-review.

**Why it moved, and it was not optional.** `main` advanced to `a4ba5f96c` and appended `smoke:sys-injection-evict` to `bin/smoke/ci-suites.txt`, after the same line this branch appends after. Merging `main` in produced no conflict and required no resolution, and the union merge driver placed this branch's entry **before** main's.

Measured at the merge commit, before correction:

```
pnpm check:shard-stability a4ba5f96c e552bc1c5
  pre-existing suites CHANGING SHARD: 1 of 412
  first few: smoke:sys-injection-evict
  RESHARD DETECTED.
```

After moving this branch's block to the true tail:

```
pnpm check:shard-stability a4ba5f96c ad9c75e29
  suites: 412 -> 413 · added 1 · removed 0
  STABLE
```

The displaced suite belongs to a change this branch never touched. Merging without the correction would have moved a third party's suite to a different runner, invisibly, with no conflict and nothing in either PR's diff to show it. Written up as the third instance on #983, and the first caught in the act.

**Re-verified at the new head**, since the merge pulled in 15 files of main's changes:

- `smoke:mutation-fixtures`, `smoke:workflow-concurrency`, `smoke:gate-inventory`, `smoke:ci-suites` all pass
- `pnpm mutation-proof --config bin/smoke/mutations/workflow-concurrency.json`: baseline green, **3 of 3 killed**, each named by its expected cell
- tree clean after mutation-proof's own restore

**One new gate this head has to satisfy that earlier heads did not.** `main` added `bin/tsconfig.smoke.json` with `include: ["**/*.ts"]`, so `bin/smoke` is typechecked for the first time and this PR's new suite lands inside it. Checked: `tsc -p bin/tsconfig.smoke.json` reports no error mentioning `workflow-concurrency.smoke.ts`. It imports only `node:fs`, `node:path` and `node:url`. The 31 errors that config does report are all one class, `Cannot find module '@cotal-ai/workspace'`, in other suites, from the workspace package being unbuilt in that checkout rather than from anything in this branch.

**Delta for a reviewer holding a verdict at `37eb117a3`:** a merge of `main`, plus one commit moving the ci-suites block to the tail. No workflow file, no fixture, and no line of the guard changed.

davidfarah2003:
One fact for whoever rules on whether this waits for the two red checks, since the mesh is still unreachable and this is the durable channel.

**Nothing mechanical is blocking this PR.** GitHub reports it `MERGEABLE` with `mergeStateStatus: UNSTABLE`, which is the state for "mergeable, checks failing, none of them required". A failing or pending required check reports `BLOCKED` instead.

`main` has no branch protection (`404 Branch not protected`) and no ruleset (`[]`), so `ci-ok` is not enforced and neither are `installer-ok` and `matrix`. Filed separately as #996, because `ci.yml:223` describes `ci-ok` as "Single required status" and it is not one.

So the question is not "are we waiting on CI". It is "do we merge with two reds that belong to #994", and that is a judgement call rather than a gate. The two reds are a rotted `--version 0.14.1` pin in `install/test-matrix.sh:61`, on a workflow whose previous run was 2026-08-01, and this branch touches neither `install.sh` nor `install/`.

Not arguing for either answer here. Recording the measurement so the decision is made against it.

Current state at `ad9c75e2968e070981072aab53b1ac5b48c79133`: 14 pass, 3 smoke shards still running, the 2 above red.

davidfarah2003:
Merge-ordering note for whoever presses the button. Not an objection to this PR.

**Merging this will leave a red Installer run on `main`.**

`installer.yml` triggers on a push to `main` that touches its own file:

```yaml
push:
  branches: [main]
  paths:
    - 'install.sh'
    - 'install/**'
    - '.github/workflows/installer.yml'
```

This PR edits `.github/workflows/installer.yml`, so the merge commit matches that filter and the workflow runs on `main`. Its `matrix` job will fail the same way it fails here, on the `pinned-version` case, because that failure has nothing to do with this branch: `cotal-ai@0.14.1` is permanently broken on fresh install through a floating `json-canonicalize` range that now resolves to a tarball missing its own `main` file. Full trace on #994.

So the red follows this PR onto `main` while being caused by neither.

**Two ways to avoid it, both cheap:**

- Re-point `install/test-matrix.sh:61` from `0.14.1` to `0.19.0` first. 0.19.0 is the first release carrying the exact pin that fixes this, and I verified it installs and reports its version correctly in the same container image the matrix uses. That is a separate one-line change and it needs a comment recording why the version must not go older.
- Or merge this and accept a red Installer on `main` until #994 is closed.

Worth stating plainly because nothing here is enforced: `main` has no required status checks (#996), so neither option blocks anything. The choice is about whether `main`'s Installer badge tells the truth in the meantime.

Everything else on this PR is unchanged: 18 checks passing, the two reds being the pair described above.

PR #981, opened 2026-08-29
feat(core,connector-core): report transport liveness separately from bind readiness

### LINKED ISSUE #975: stop() during an initial start() leaks a live connection, heartbeat and supervisor: only the rebuild path checks for a stop that landed mid-bind
`CotalEndpoint.stop()` called while the **initial** `start()` is still binding does not stop anything.
The bind completes, the supervisor arms, and the endpoint ends up `stopped` while still owning a live
connection, a heartbeat and a supervisor.

## Where

`packages/core/src/endpoint.ts` on current `main` (`d58d507c3`):

```ts
async start(): Promise<void> {
  await this.connectAndBind();
  // ... Armed only after a successful first connect ...
  this.superviseConnection();
}
```

`tearDownIfStopped()` is defined at `endpoint.ts:1049` and has **exactly one caller**, `endpoint.ts:1134`,
which is inside `doRebuild()`:

```
$ grep -n "tearDownIfStopped" packages/core/src/endpoint.ts
1049:  private async tearDownIfStopped(): Promise<boolean> {
1134:      if (await this.tearDownIfStopped()) return;
```

So the rebuild path honours a stop that lands mid-flight and the initial start path does not. The
asymmetry looks like an oversight rather than a decision: the two paths do the same thing and only one
of them checks.

## Observed

Delaying a controlled initial bind, calling `stop()`, then releasing the bind produces:

```
{ stopped: true, hasNc: true, hasHeartbeat: true, supervised: 1 }
```

A stopped endpoint holding a connection, a heartbeat timer and an armed supervisor. Nothing tears them
down afterwards, because `stop()` has already run and the only teardown-on-stopped check lives in a
path this sequence never enters.

## Why the window is real rather than theoretical

`connectAndBind()` is not fast and not bounded by anything the caller controls: it dials the broker,
opens streams and KVs, binds durables, arms membership watches and writes presence. Any host that can
decide to shut down while a first connect is in progress can hit this, which includes every ordinary
case of a session that starts and is cancelled: a spawn that is stopped during startup, a readiness
gate that gives up, a broker that is slow or briefly unreachable at exactly the wrong moment. It does
not need a race that is hard to arrange, only a stop during a slow first connect.

## Consequences

- A leaked connection, heartbeat and supervisor per occurrence. The supervisor is the worse half,
  because on a terminal close it will run `reestablishLoop()` and rebuild a connection for an endpoint
  that was stopped, which is a leak that actively recreates itself.
- Presence for a session that no longer exists.
- `stop()` returning to its caller having done nothing, while reporting nothing.

## The reason this is being filed now, and it generalises

It surfaced during review of an unrelated change that makes `MeshAgent` clear its local flags on stop
and refuse late endpoint events. Those guards are correct on their own terms, and they have a side
effect worth stating: **before them, a stopped session reporting itself connected was a visible symptom
pointing at this leak. After them the session reports correctly and the leak is unchanged.**

So the observable evidence for this defect is about to disappear while the defect stays. That is the
argument for fixing the lifecycle here rather than only fixing what is reported, and it is the same
shape as #967, where a hang was the only visible symptom of a suite that had stopped testing its own
headline claim.

## Ask

`start()` should honour a stop that landed during its initial bind, the way `doRebuild()` already does.
Whether that is a `tearDownIfStopped()` call after `connectAndBind()`, a check before arming the
supervisor, or a rejection of `stop()` during startup with a defined wait, is a design call for whoever
owns the lifecycle. What is not defensible is the current asymmetry, where the same race is handled on
one path and ignored on the other.

Accepted when a `stop()` issued during an in-flight initial bind leaves no connection, no heartbeat and
no armed supervisor, proved by a test that fails if the check is removed.

Refuted if a stop during initial start is considered the caller's error and is supposed to throw or be
refused. In that case the fix is to say so and make it do that, because today it silently succeeds and
leaks. Either outcome closes the asymmetry; only silence leaves a `stop()` that does not stop.

davidfarah2003:
A connector session had one notion of "connected": the endpoint's `connection` event, which means a
full Cotal bind has completed. That is the right signal for readiness gates and for the self-heal
supervisor, and it is the wrong signal for "is the socket up right now". A transient NATS
disconnect and reconnect never moved it, so a session could sit with a dead socket while every
consumer read it as connected.

This adds raw transport liveness as a separate, additive signal and leaves `connection` alone. The
readiness gates and the self-heal supervisor consume `connection`, and making it track socket
liveness would flap them on every transient reconnect.

## What changed

`CotalEndpoint` gains a `transport` event carrying `{ connected, server? }`, driven from the
client's status stream. The status union was checked against the pinned client rather than assumed:
`disconnect` and `reconnect` are the authoritative pair, and `reconnecting` is telemetry that must
not move state.

`watchStatus` needed the epoch guard `superviseConnection` already had. An old connection's status
iterator keeps running after a rebuild replaces it, so without the guard a dying epoch can overwrite
the state of the healthy connection that replaced it.

`MeshAgent` tracks the new state alongside readiness, suppresses duplicate edges so a repeated
confirmation cannot look like a fresh outage, and ignores late endpoint events after `stop()`, since
an in-flight bind can complete after shutdown and must not resurrect a stopped session.

## Behavior changes worth calling out

`stop()` now clears both states locally rather than waiting on an endpoint event that deliberately
ignores its own stopped close, and it fences the initial `connectAndBind` the way the rebuild path
already did. Without that fence, a `stop()` racing the *initial* bind left a live connection,
heartbeat, consumers and presence behind. That is #975, and only the rebuild path had the check.

The same race had a second half that the state fence did not cover. `connectAndBind` emitted its
`connection: true` as its last act, before either caller could reach `tearDownIfStopped`, so a
`stop()` landing mid-bind left every endpoint listener holding a connected edge on a connection
the caller had already discarded, with nothing following to correct it. The emit is now guarded on
`stopped` as well. `MeshAgent` was never the exposed consumer, since it carries its own `stopping`
guard and drops the late edge; that is why the guard belongs at the emitter, where one check covers
every listener, rather than in a consumer.

Review then found the sibling edge on the same race, and it reaches further back. `watchStatus`
seeds `transport: true` as soon as the dial returns, and `connectAndBind` calls it right after the
dial, long before the bind completes. A `stop()` landing while the dial was still in flight had
that seed fire on an endpoint already stopped, so the listener saw transport true after stop,
followed by false. Reproduced through a real pending dial before it was fixed. Both edges are now
guarded on `stopped`.

`stop()` no longer clears the last connection issue. A cleanly stopped session keeps it for
post-mortem diagnosis instead of discarding the reason it stopped. Post-stop errors cannot overwrite
that retained value. The issue remains scoped to pre-bind readiness failures and still clears on a
successful bind.

## Testing

Two suites, both appended to the end of the CI suite list so every existing shard assignment is
unchanged.

`smoke:transport-liveness` (20 cells) grades the state contracts deterministically with controlled
status queues: epoch staleness, disconnect and reconnect edges, duplicate suppression, the stop
races, terminal-close diagnostic ordering, and issue scoping.

`smoke:transport-liveness:broker` (9 cells) owns a throwaway broker on an OS-assigned port and
proves the real paths: initial transport before full-bind readiness, a real broker loss lowering
transport without flapping readiness, a real client reconnect restoring it, and a real terminal
close through `nc.closed()` reaching the diagnostic ordering rather than a hand-built double.

An 18-mutation fixture pins the deterministic suite and each mutation names the cell expected to
fail. The broker suite carries its own two-mutation fixture, one for the readiness-event fence and
one for the stopped guard on the transport seed that fires first, because those guards are
unreachable from the deterministic suite: all three of its stop-versus-bind
cells replace `connectAndBind` wholesale, so nothing inside the method is under test there. The
broker cell gates `armPlane3` instead, the last await `connectAndBind` makes before reporting the
endpoint live and a no-op for an endpoint hosting no Plane 3, which holds a real bind open at its
final step while `stop()` lands and leaves the method itself real.

## What the proof does not claim

One mutation covers the epoch guard on `watchStatus`'s catch handler, and its cell manufactures the
rejection rather than reaching it from a real client. Rather than dress that up, it was measured:
temporary instrumentation around the real catch, across five complete broker-companion runs covering
loss, reconnect, manual epoch replacement and terminal close, recorded zero fires. Every status
iterator ended normally.

The guard is kept, because removing it would leave the catch path disagreeing with the reachable
in-loop epoch guard beside it, and the claim is bounded to what was measured. The cell is named as a
*controlled* throw, and the suite header, the fixture and the endpoint comment all state that real
reachability is intentionally unclaimed on the pinned client. This is deliberately not a claim of
impossibility across other runtimes or future client versions.

The stopped guard is measured for the `start()` caller only. `connectAndBind` has a second caller,
`doRebuild`, and no cell races a stop against a rebuild's bind. That path is covered by the guard
being one shared unbranched statement both callers await, with `tearDownIfStopped` beside it in
both places, rather than by measurement. A cell there would re-prove control flow at the cost of a
timing-heavy terminal-close race on a CI shard. The reasoning is written at the guard itself along
with the condition that voids it: if that tail ever becomes caller-aware, or the emit splits per
path, the rebuild race needs its own cell.

Closes #975


davidfarah2003:
Head moved to `653aedd84ae26f2b01e2bbb88736694bdd9aa452`. Flagging it rather than leaving it to be discovered, since a verdict may be held at `0e043457c`. The mesh is unreachable, so this is the durable channel.

**Why it had to move.** This PR was reported `CONFLICTING` / `DIRTY`, and a conflicting PR runs no CI, so its checks had been stale since 04:22 and it could neither merge nor get a fresh verdict. The conflict was not real. At the two shas GitHub was comparing, local git merged clean:

```
git merge-tree --write-tree --name-only origin/main conn/phase1a-transport-liveness
  764dc86fc3ad7be89d0ad27b5908c4f24450eb6f
  rc=0, no conflict lines
```

Refs verified identical to the remote first, so the disagreement was not a stale checkout. The cause is `bin/smoke/ci-suites.txt merge=union`: local git applies the attribute and GitHub does not appear to. Same three blobs under plain 3-way merge give `rc=1`, one hunk, and that is the only file involved. Written up as #998.

So the fix was to merge `main` in and resolve that file by hand, which is what moved the head.

**What changed, in full:**

1. A merge of `origin/main` at `a4ba5f96c` (69 files from main, no conflict after the union resolution).
2. One commit moving this branch's `ci-suites.txt` block to the true tail.

No line of this branch's own code changed. The diff of `packages/core` and `extensions/connector-core` against `0e043457c` is empty.

**The second commit is not cosmetic, and it is #983 for the fourth time.** Union kept both sides of the append but placed this branch's two entries before main's, displacing a suite this branch never touched:

```
pnpm check:shard-stability a4ba5f96c 45b046d45
  pre-existing suites CHANGING SHARD: 1 of 412
  first few: smoke:sys-injection-evict
  RE-SHARD DETECTED.
```

After moving the block to the tail:

```
pnpm check:shard-stability a4ba5f96c 653aedd84
  suites: 412 -> 414 · added 2 · removed 0
  STABLE
```

**Re-verified at the new head**, since the merge brought in 69 files including `bin/tsconfig.smoke.json`, which typechecks `bin/smoke` for the first time:

- `pnpm typecheck`: clean across every package, `bin` included
- `pnpm build`: clean
- `pnpm smoke:transport-liveness`: 20 passed, 0 failed
- `pnpm smoke:transport-liveness:broker`: 7 passed, 0 failed
- no broker left behind

`mergeStateStatus` is now `UNSTABLE` rather than `DIRTY`, so CI can run again.

davidfarah2003:
I pushed the merge that unstuck this PR without having read its diff, which was the wrong order. Doing that read now, and recording it here so the review has a durable home rather than living on a mesh channel that is currently down.

## The core change holds up

`packages/core/src/endpoint.ts` does what E2 asked and no more. `connection` is untouched, and the new `transport` event is a separate, lower-level fact, so the readiness gates and the self-heal supervisor cannot be made to flap by a transient socket blip. The epoch guard is the part I would have most expected to be missing and it is there:

```ts
// A rebuild can replace `this.nc` before the old iterator finishes. Late disconnect/close
// from that old epoch says nothing about the replacement and must not flip its liveness.
if (this.nc !== nc) continue;
```

Capturing `nc` locally rather than reading `this.nc!` inside the loop is what makes that guard possible, and the same guard is repeated on the catch. The `nc.closed()` handler's ordering comment ("ORDER IS PART OF THE DIAGNOSTIC CONTRACT") explains a constraint that is otherwise invisible and easy to undo.

## One gap, and it is a missing assertion rather than a defect

The terminal-close cell asserts readiness only:

```ts
check(
  "a REAL terminal close marks readiness false before exposing its user-visible reason",
  await until(() => !agent.connected && /mesh connection closed/.test(terminalIssueAtError ?? ""), 30_000),
  { ready: agent.connected, terminalIssueAtError, latestIssue: agent.connectionIssue, transport },
);
```

`transport` is passed into the diagnostic payload but never asserted on. The only cell that asserts `transportConnected === false` is the next one, and it runs after `await agent.stop()`, so it proves `stop()` clears the flag rather than that a terminal close does.

That matters because of what consumes it. Phase 1 derives:

```ts
return this._transportConnected ? "connecting" : "disconnected";
```

so `connected === false` with `transportConnected === true` renders as **"connecting"**. If a terminal close ever left transport true, a permanently dead session would report itself as coming up, and nothing here would fail.

**Measured, not assumed.** I added `&& agent.transportConnected === false` to that cell and ran the broker suite: **7 passed, 0 failed**. The behaviour is correct today. Then I reverted the edit; the tree is clean at `6a69df027`.

So this is not a blocker and nothing needs fixing. It is one clause missing from an existing assertion, protecting a fact a downstream phase renders to a model. Worth adding before phase 1 lands, since phase 1 is what makes the fact user-visible.

## Scope of this read

Open-brief, not a lens. I read the `packages/core` and `connector-core` diffs and the two suites. I did not audit the 460-line unit-shaped suite cell by cell, and this is my own lane, so it does not substitute for independent review.

davidfarah2003:
## The correctness block is addressed at 99c84147

`connectAndBind` emitted `connection: true` as its last act, before either caller could reach `tearDownIfStopped`. A `stop()` arriving during the bind left every endpoint listener holding a connected edge on a connection the caller had already discarded, with nothing following it to correct the record.

The state teardown was already right on both paths, so only the event was unfenced. The emit is now guarded on `stopped`.

## Two things the fix turned up that change the finding's shape

**MeshAgent was never the exposed consumer.** It carries its own `stopping` guard at `extensions/connector-core/src/agent.ts:328` and drops the late edge, so its flag stayed correct throughout. The exposure was to any other listener on the endpoint, which had no such guard. That is also why the fix had to sit at the emitter rather than in a consumer: one guard there covers every listener instead of asking each to carry its own.

**The unit suite structurally cannot prove this.** All three of its stop-versus-bind cells replace `connectAndBind` wholesale, so nothing inside the method is under test there. Adding an assertion to that suite would have produced a cell that cannot fail.

## How it is proven

The broker suite gets the cell. It gates `armPlane3`, the last await `connectAndBind` makes before reporting the endpoint live, which holds a genuine bind open at its final step while `stop()` lands. `armPlane3` is a no-op for an endpoint that hosts no Plane 3, so gating it holds the clock without changing any behaviour under test.

Removing the guard turns that cell red with the late edge in its own diagnostic:

```
✗ FAIL: stop during a REAL initial bind never announces the connection it then tears down {
  reachedFinalStep: true,
  raceEdges: [ { connected: true } ],
  ready: false
}
```

`reachedFinalStep: true` is what rules out a vacuous pass, and `ready: false` is the MeshAgent guard above showing its work. Recorded as a mutation fixture so it stays proven rather than being a one-off:

```
KILLED  a stop landing mid-bind still announces the connection it then tears down
  red, and named: stop during a REAL initial bind never announces the connection
  it then tears down - 7 marks (baseline 8)
```

## Limits worth stating

The timing is forced rather than observed. The gate makes the race deterministic instead of waiting for it to happen, which is the only way to test it repeatably; the race itself is not hypothetical, and `start()` already carried a comment describing it before this change.

The entry point is real: `start()` goes through `connectLoop` to `ep.start()` to the real `connectAndBind`, against a real dial and a real broker. Only the one no-op step is held. So this is not a test that builds its inputs by hand and proves nothing about reachability.

Verified at 99c84147: broker suite 8/8, unit suite 20/20, `pnpm typecheck` clean repo wide, all 216 mutation fixtures present and unique. Both suites are already in the CI list at lines 663 and 664, so the new cell runs in the gate.

davidfarah2003:
Repaired against current `origin/main` without changing feature behavior.

Exact head: `ed3091317482e121bf64078c98cb8d5e82e39027`.

The real local merge was conflict-free, matching `git merge-tree`; GitHub's prior `CONFLICTING / DIRTY` state did not correspond to a content conflict at the pinned refs. The merge did expose a real positional registry repair: the transport suites landed before main's newer `smoke:workspace-import-exports`, and the shard checker correctly returned `RESHARD=1`. I moved only the transport block to the true tail and committed that ordering repair.

Validation at start load `21.03, 18.01, 24.53`:

- `pnpm check:shard-stability origin/main HEAD`: `421 -> 423`, added 2, removed 0, `0 of 421` pre-existing suites changing shard, `STABLE=0`.
- `pnpm smoke:transport-liveness`: 20 passed, 0 failed, completion banner present.
- `pnpm smoke:transport-liveness:broker`: 9 passed, 0 failed, completion banner present.
- `pnpm typecheck`: green.
- `pnpm changeset status`: valid fixed-group plan.
- `git diff --check`: green.

Stacked PR #1002: its four shared files were not manually resolved or behavior-edited during this repair. A tree merge of #1002's exact approved head `b6179d712cb9cf8ddcc3c4cd2629f8c550332a25` onto this repaired head succeeds cleanly (`merge-tree` rc 0). GitHub is recomputing its stacked mergeability after the base moved.

davidfarah2003:
Repair update at exact head `ed3091317482e121bf64078c98cb8d5e82e39027`: reproduced a real local merge of current main `bb54b7fcb8771f094e1e18d4e68d672f6f8a591d` into prior head `b5805ca2cfa86d1da3a99761b49165b682369ad1`. Git merged source cleanly, but the suite registry required one mechanical re-tail: the two transport suites initially displaced `smoke:workspace-import-exports`. After re-tailing, `pnpm check:shard-stability origin/main HEAD` reports 421 -> 423 suites, added 2, 0 of 421 pre-existing suites changing shard, STABLE. `smoke:transport-liveness` is 20/20 and broker companion 9/9. Full `pnpm typecheck` is green. PR #1002 still merge-trees cleanly atop this head (rc 0); current main altered only its generated docs among the four shared files, while #1002 retains its expected deltas on all four. PR #1052 also merge-trees cleanly with this head, but a synthetic combined commit cannot be graded by either branch checker because each pins a different verifier blob; landing order remains an orchestration decision, not a conflict resolution performed here. No behavior change beyond current-main integration and suite re-tail. Load at final exact-head check: 35.98, 26.94, 26.28.

davidfarah2003:
Final stack state after PR #1002 merged into this branch: exact head is now `46acf2cbb0df63ada21178093f6b7e1655dcfdca`. This is the GitHub merge commit for approved #1002 atop repaired head `ed309131...`, not an unreviewed manual rewrite. Re-ran shard stability against current main `bb54b7fc`: 421 -> 424 suites, added 3, removed 0, 0 of 421 pre-existing suites changing shard, STABLE. GitHub reports this PR MERGEABLE / UNSTABLE. The foundation plus approved connection-status layer are now one landable branch; no merge into main was performed here.

davidfarah2003 (commented):

- **The body no longer describes what this branch ships.** #1002 was merged into `conn/phase1a-transport-liveness` (`46acf2cbb`, 2026-08-30T03:17), so this PR now carries an entire second feature the body never mentions: the `cotal_connection_status` tool at `extensions/connector-core/src/tool-specs.ts:551`, new public API `ConnectionState` (`extensions/connector-core/src/agent.ts:211`) plus the `transportConnected` / `lastInboxDrainedAt` / `stopping` / `connectionState` getters (`agent.ts:361,371,378,385`), a `_lastInboxDrainedAt` write on both drain paths, `docs/mcp-tools.md`, `scripts/generate-tool-docs.mjs`, a 15-cell smoke with its own 7-mutation fixture, and a second `minor` changeset `.changeset/connection-status-tool.md`. #1002's own body says "It does not merge before #981"; the stack collapsed the other way and this body was not updated to follow. For a landing queue the body is the record of what lands, and right now it under-describes the diff by a whole feature.
- **Testing section counts disagree with the shipped artifacts.** "Two suites, both appended to the end of the CI suite list" — three are appended: `smoke:transport-liveness`, `smoke:transport-liveness:broker`, `smoke:connection-status`. "`smoke:transport-liveness:broker` (7 cells)" — the shipped fixture pins it at `"completionMarker": "SUITE COMPLETE: 9 cells"` (`extensions/connector-core/smoke/fixtures/transport-liveness-broker.mutations.json:6`), and the suite has 9 `check(` calls. "The broker suite carries its own single-mutation fixture" — that fixture ships **two** mutations (same file, `mutations` array at line 21), the second being the transport-seed guard your own "Behavior changes worth calling out" section describes. Under-claiming, but these are the numbers a reviewer reconciles against and none of the three match.
- **No reported evidence run.** There is no Evidence block: no pass counts for the three suites on the tree that actually ships, no kill line for the 18-mutation, 2-mutation and 7-mutation fixtures, no typecheck/build, no `check:shard-stability`. The suites and fixtures are described, never reported as executed. #1002 carries those numbers, but for a merged-in child, not for this PR's current head — and rail (b) wants the mutation proof named on what lands.
- Verified good, and it is genuinely good work: the epoch guard in `watchStatus` captures `const nc = this.nc` and re-checks it both in-loop and in the catch; the two stop-race fences (`endpoint.ts` `if (this.stopped) return;` before the `connection:true` emit, and before the `transport` seed) are each pinned by a broker-suite mutation on real code; `disconnect`/`reconnect`/`close` are handled and `reconnecting` deliberately is not; the terminal-close emit ordering is asserted as a contract rather than left implicit. The catch-handler guard is honestly declared unreachable-as-measured rather than dressed up as proven — that is the right way to write that down.
- Rails and hygiene are clean: three suite names at the absolute tail of `bin/smoke/ci-suites.txt` (the only mid-file edits are two blank-line deletions, which `bin/smoke/ci-suites.mjs` strips before `shard.mjs` indexes, so shard assignments hold); no fallbacks — the `server?` field is omitted when the runtime supplies none rather than invented, and status comes straight off `nc.status()` with no polling substitute; `connectionIssue` narrowing to pre-bind is documented in `docs/connectors.md`; both changesets are `minor`; no AI/tool attribution anywhere.
- (i) Interactions: #1002 is inside this branch, so the two land together or not at all — its review is posted there. `bin/smoke/ci-suites.txt` tail conflicts with #921, #880, #1053 and #1054, and #1052/#1034/#1033 restructure that file and `shard.mjs` underneath all of them.

PR #982, opened 2026-08-29
fix(connector): retain unreplayable focus traffic

### LINKED ISSUE #977: focus mode ack-drops @mention bodies that recall then cannot return, and reports no loss
A `focus`-mode agent ack-drops channel ambient and `@mention` bodies at ingest, on the promise that
they stay recallable. On a `replay=off` channel and on a wildcard subscription that promise does not
hold, and the loss is not reported: `cotal_inbox` answers as though there were no chatter.

This is not the recall window aging out, which the code already reports. It is recall returning
empty and also returning "nothing was dropped".

## The path

Ingest, `extensions/connector-core/src/agent.ts:496-501`:

```ts
if (cm !== "quiet" && !snapshottedPullOnly && this._attention === "focus") {
  this.protectDisposition(item.id, "drop");
  delivery.ack();
  if (item.mentionsMe) this.emit("mention-wake", item);
  return;
}
```

The delivery is acked and the item is never buffered, so the JetStream copy will not redeliver and
there is no local copy. The comment three lines above states the compensating promise: "they stay
recallable via cotal_inbox (recallAmbient)".

Recall, `extensions/connector-core/src/agent.ts:1009`:

```ts
for (const channel of this.ep.joinedChannels()) {
  if (!isConcreteChannel(channel)) continue;          // <- silent, no droppedChannels entry
  if (this.focusRecallUnsafeChannels.has(channel)) { droppedChannels.push(channel); continue; }
  const { messages, dropped } = await this.ep.recallChannel(channel, this.focusSince);
  ...
  if (dropped) droppedChannels.push(channel);
}
```

And `packages/core/src/endpoint.ts:4163`:

```ts
if (!isConcreteChannel(channel)) return { messages: [], dropped: false };
const policy = await this.joinPolicyFresh(channel);
if (!policy.replay) return { messages: [], dropped: false };   // <- silent
```

So there are two ways for a focus-dropped body to become unrecoverable, and both return
`dropped: false`, which is the value that means "the window was complete".

1. **The channel has `replay=off`.** The gate exists for a good reason, that focus must not become a
   history bypass for a channel that denies replay to everyone else. The bug is not the gate. It is
   that ingest acked and dropped the message on a promise recall then declines to keep, without
   saying so.
2. **The subscription is a wildcard** such as `team.>`. `recallAmbient` skips it before it reaches
   `recallChannel`, because a wildcard cannot be Direct-Get read, and adds nothing to
   `droppedChannels`.

`replay=off` is not exotic. It is settable as a whole-space default, and the suites exercise exactly
that shape: `packages/core/smoke/channels-auth.smoke.ts:60`,
`packages/core/smoke/no-implicit-general.smoke.ts:160`,
`extensions/connector-opencode/smoke/turn-wedge.smoke.ts:125`, among others. A space configured with
`defaults: { replay: false }` loses every focus-mode mention body, silently, today.

## Why silence is the defect and not the loss

The surrounding code is deliberately never-silent about this exact class of thing. The
`droppedChannels` docstring says it names channels "whose earliest retained message postdates the
focus-watermark, older ambient may have aged out of the per-channel window (never-silent)". The
aged-out case is reported. The two cases above are strictly worse, since the message was destroyed
rather than merely aged, and they are the two that report nothing.

For a mention specifically there is no second chance anywhere. `createWakePolicy` in
`extensions/connector-claude-code/src/hooks.ts` already carries the reasoning: "A focus-mode @mention
is ack-dropped at ingest, not buffered, so `pendingWake()` is 0 for it and JetStream will never
redeliver it either." It solved the missed-push half by holding `pendingMentionWake` and retrying.
The missed-recall half is unsolved.

## Repro shape

1. Seed a channel registry with `defaults: { replay: false }` and one channel.
2. Bring up two agents, both joined.
3. Put agent A in `focus`.
4. From agent B, post to the channel with `mentions: ["A"]`.
5. On A, call `cotal_inbox`.

Expected: either the mention body, or a statement that a body was dropped and cannot be recovered.
Actual: no items, and `droppedChannels` empty.

Repeat with A subscribed to `team.>` and B posting to `team.x` for the wildcard variant.

## What a fix has to decide

Not obvious, and worth deciding rather than defaulting:

- **Do not ack-drop what cannot be recalled.** At ingest, consult the same replay policy recall will
  use, and buffer rather than drop when recall could not return it. Costs a policy read on the
  ingest path, and the read is already cached by `joinPolicyFresh`'s callers.
- **Or report it.** Keep the drop and have recall name these channels in `droppedChannels`, so the
  reply says a body was destroyed. Cheaper, and honest, but it still loses the message.

The first is the one that matches the repo's no-fallbacks stance: a promise the code cannot keep
should not be made at ingest.

## Why it is worth fixing before the hosted work

This blocks a pull-only MCP session, which cannot be woken at all and for which recall is the only
delivery path. But it is a live defect on push sessions too, so it is not a prerequisite of an
unbuilt feature. It is current behaviour.

davidfarah2003:
## Summary

- preserve focus-mode channel bodies locally when stream recall cannot return them
- keep replay-enabled concrete channels on the existing stream-recall path
- cover replay-off and wildcard focus loss through a real-broker `cotal_inbox` acceptance smoke

## Decision

Buffer at ingest rather than only report the loss. A hosted pull-only client has no push fallback, so naming a destroyed body would leave the delivery defect intact. The connector now ack-drops only after an authoritative replay-policy read proves that this exact concrete subscription is recallable.

## Verification

- independent real-broker reproduction at `7c0b24a9`: both replay-off concrete and wildcard cases lost ambient + mention bodies silently
- `pnpm smoke:focus-recall-loss`: real broker, real `MeshAgent`, real `cotal_inbox`; target bodies delivered exactly once and not repeated
- mutation proof: 3/3 killed
- attention: open 14/14, auth 14/14
- per-channel attention: 45/45, auth 14/14
- cross-path dedup: 54/54
- inbox window: 85/85
- Claude wake: 63/63
- wildcard backfill: 5/5
- channel registry: 29/29
- full `pnpm typecheck` green

Named gap: `smoke:jcode-live` skipped under its own `COTAL_E2E_JCODE=1` authentication gate.

Closes #977.


davidfarah2003:
Marking this draft as a hold. It is not a judgement on the change, which is well evidenced.

Fixes in this issue-triage program need an explicit go-ahead before they land, and that has not been
given yet. The reproduction, the acceptance runs and the mutation proof all stay exactly as they
are. Draft only stops an accidental merge while that is sorted out, and it can be undone in one
click. Reviewers can still read, review and comment normally.

Worth stating for whoever reviews: the decision here is to buffer at ingest rather than to report
the loss, and the ack-drop path now does a registry read before deciding. Whether that read is per
message or cached, and what happens when it fails, is the first thing to check.

davidfarah2003:
Head moved to `1d4036f08`, fixing this PR's red.

## What was failing

`smoke (shard 1/4)` failed in 67 seconds, with `ci-ok` red behind it. Not a suite failure, and not the known shard-1 flake:

```
===== pnpm smoke:gate-inventory =====
  ✗ FAIL: 1 suite(s) exist but nothing runs them, and they are not in UNGATED:
      smoke:focus-recall-loss
    Gate it in smoke:ci, or add it to UNGATED with the reason it is excluded.
GATE INVENTORY FAILED
```

`smoke:focus-recall-loss` was defined at `package.json:135` and absent from `bin/smoke/ci-suites.txt`, so nothing ran it.

## Why this one matters more than a missing line

The summary above cites that suite as verification:

> `pnpm smoke:focus-recall-loss`: real broker, real `MeshAgent`, real `cotal_inbox`; target bodies delivered exactly once and not repeated

So a reader of this PR sees a pass count for a suite CI never runs, and the acceptance test for #977 would have merged ungated. That is the exact case `gate-inventory` names in its own output as the worst of the categories it tracks: cited in the record by pass count, and run by nothing. The guard did its job.

## The fix

One entry, appended at the tail of `ci-suites.txt` with the comment block the file's convention asks for.

Order mattered. I merged `origin/main` first and appended afterwards, rather than the reverse, because the union merge driver on that file keeps both sides of an append but places the incoming branch's entry first, which silently moves a third party's suite to another shard (#983, and #998 for why that same driver makes these PRs read as conflicting).

Verified:

```
pnpm smoke:gate-inventory
  gate inventory: 470 smoke scripts, 450 reached, 20 not run by anything
  GATE INVENTORY OK

pnpm check:shard-stability a4ba5f96c 1d4036f08
  suites: 412 -> 413 · added 1 · removed 0
  STABLE
```

Still a draft, and still not mine to lift.

davidfarah2003:
Closing: superseded by #1058, which fixes the same issue (#977) and merged first with a fully green 16/16 run. Both branches were reviewed approve-quality; the twin rule here is whichever greens first lands, and under today's CI queue this branch never minted a full check list. No unique changes remain on this branch.

Issue #992, opened 2026-08-29
user-mode `up --restore` degrades at resume-preserved: operator-reach refuses a caller whose grant carries admin


davidfarah2003:
Found while closing the bin/smoke typecheck gap: `smoke:backup-usermode-live` is run by nothing in CI (reached only via the `check` composite). At current main it fails 9 of 38 cells, and the first failure is the root — everything after cascades from the degraded maintenance state.

First failure (cell E, "restore the user-mode artifact"): the restore reaches
```
resume-preserved is operator reach; the caller's current ledger grant does not carry "admin" (SPEC 13.2)
```
and the journal lands `degraded` with a repair recourse.

What the suite establishes before the backup (all ledger writes precede the cut, per its own comments):

- device login established (`sub === userId` checked);
- `actor grant cli --sub <sub> --scope "spawn,role:default,admin"` — the suite's comment notes the preserve cut "speaks ctl.admin", and that preserve cut **succeeds** with this grant;
- the backup artifact then binds to the USER authority fingerprint (cells pass).

At restore time the identity plane is up before the failure (the cell logs the user-auth service URL), and `epAdminReach` (implementations/manager/src/manager.ts:2048) does a fresh `actorScope` read from `userAuthStateDir(workspaceRoot, space)` and finds no `admin`.

Open questions I did not settle, in either direction:

- which principal (`owner`/`actor`) the restoring `up` presents to the serve gate at that moment — the pre-cut preserve succeeded with the same grant, so a plain actor-name mismatch alone does not explain it;
- whether the ledger directory `epAdminReach` reads during `resume-preserved` is the restored one, or a pre-restore state that legitimately has no rows yet (the manager's own comment says the resume family deliberately keeps the admin gate while bypassing serveGated).

Repro: `pnpm smoke:backup-usermode:live` at main → `BACKUP USER-MODE SMOKE FAILED (29 passed, 9 failed)`.

davidfarah2003:
[...]

Issue #991, opened 2026-08-29
backup create refuses a healthy stamp while a kv-scan ordered consumer is resident on a KV stream


davidfarah2003:
Found while closing the bin/smoke typecheck gap: `smoke:backup-faults:live` is run by nothing in CI (reached only via the `check` composite), and once its unrelated startup defect is fixed (mintCreds without a lifecycleUid at backup-faults-live.smoke.ts:181 — fix in an incoming PR), the suite fails deterministically at "replay verification backup":

```
KV_cotal_channels_backup_faults_replay/oc_..._1 has an unsupported stopped KV watcher config:
{"name":"oc_..._1","deliver_policy":"all","ack_policy":"none","max_deliver":1,
 "filter_subject":"$KV.cotal_channels_backup_faults_replay.>","replay_policy":"instant",
 "flow_control":true,"deliver_subject":"_INBOX_...","idle_heartbeat":30000000000,
 "inactive_threshold":300000000000,"num_replicas":1,"metadata":{"_nats.req.level":"0"}}
```

Evidence, both sides named:

- `implementations/cli/src/commands/backup.ts` (`isStoppedKvWatcher`) accepts a stopped `oc_*` KV watcher only with `deliver_policy: "last_per_subject"`; everything else about the refused consumer matches the allowlist.
- `packages/core/src/kv-scan.ts` deliberately creates its ordered consumer with `DeliverPolicy.All` (documented there: All chases a moving tail so tombstones collapse correctly), and it is imported by `channels.ts`, `members.ts`, `acls.ts`, `endpoint.ts` — i.e. ordinary registry reads (`channels set`, `send`, endpoint connects) mint exactly this consumer shape.
- The consumer carries `inactive_threshold: 300s`, so it outlives its client by up to five minutes.

Consequence: any `backup create` within ~5 minutes of any registry read can refuse a healthy mesh — the refusal window is invisible operationally because it depends on tar-time consumer residue, and the one suite that witnesses it runs nowhere.

Not proposing which side should move (accepting All-policy stopped watchers in the allowlist, skipping them, or quiescing before the cut each have different custody implications) — flagging the mismatch with the repro so the call can be made deliberately.

davidfarah2003:
[...]

PR #993, opened 2026-08-29
test(ci): typecheck bin/smoke and repair the broken live backup suites

davidfarah2003:
bin/smoke held 469 runnable scripts that no tsc ever read: bin/tsconfig.json includes only the two published entrypoints, and nothing else compiles the directory. Two of the five live backup suites had rotted unnoticed behind that gap — they are reached only via the `check` composite, so CI never runs them either.

- **fix(smoke): restore the env destructure in seven backup-restore live scenarios** — `sandboxRun` returns `env`, but seven scenarios destructured only `{ run }` and then referenced `env` in their spawn calls. Seven TS2304/TS18004 errors; at runtime the suite dies at the first such scenario. With the fix, `smoke:backup-restore:live` passes end to end.
- **fix(smoke): mint the replay sender with the lifecycleUid its grant is keyed by** — `mintCreds(..., "agent", ...)` without `lifecycleUid` throws at provision.ts (SPEC 13.1); the replay scenario never started. With the fix the suite proceeds to a deeper, product-level refusal now filed as #991.
- **test(ci): typecheck everything under bin, not just the two entrypoints** — adds `bin/tsconfig.smoke.json` (noEmit companion; the build config keeps its emit layout untouched) and extends the package's `typecheck` script to run both, so `pnpm -r typecheck` covers bin/smoke from now on. Verified green end to end.

Suite status after this branch: `backup-restore:live` passes; `backup-conservation:live` and `backup-perms:live` pass; `backup-faults:live` and `backup-usermode:live` reach real product-level failures filed as #991 and #992. A follow-up commit on this PR will gate the three green suites in `bin/smoke/ci-suites.txt` once #988's append lands (avoiding a queued end-of-file conflict).

### DIFF
--- a/bin/package.json
+++ b/bin/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cotal-ai",
   "version": "0.35.0",
-  "description": "Cotal — lateral agent coordination over NATS. Run `npx cotal-ai` for the CLI.",
+  "description": "Cotal \u2014 lateral agent coordination over NATS. Run `npx cotal-ai` for the CLI.",
   "license": "Apache-2.0",
   "keywords": [
     "cotal",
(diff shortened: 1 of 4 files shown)


davidfarah2003:
Correction to the follow-up plan stated in the description: the gating commit is not coming. Measured runtime changed the call — smoke:backup-restore:live alone runs ~35 minutes wall-clock (twelve broker-booting scenarios), which no CI shard budget absorbs, and its siblings are the same class. The suites stay reached via the `check` composite like the other full-stack live suites (the UNGATED map's own category), now actually runnable since this PR repaired them. #991 and #992 track the two that reach real product-level failures.

PR #1002, opened 2026-08-29
feat(connector-core): report connection state through cotal_connection_status

davidfarah2003:
Adds `cotal_connection_status`, so an agent can tell whether silence means no messages or no connection. Today it cannot: `cotal_inbox` returning nothing looks identical whether the mesh is healthy or the session has been unbound for an hour.

**Stacked on #981.** The base is `conn/phase1a-transport-liveness`, so the diff here is this change alone. GitHub retargets to `main` when #981 merges. It does not merge before #981 and is not useful before it, because the transport fact it reports does not exist until then.

## Why the shape is what it is

An earlier attempt at this returned `connected`, `bufferedCount`, `connectionIssue` and a derived `degraded: !connected && issue !== undefined`, under a description claiming "live mesh connection state". It was held, and correctly, because that boolean is wrong in both directions once transport liveness exists as a separate fact:

- false negative: the genuinely degraded state is bound with the socket down, `connected === true && transportConnected === false`, which a formula requiring `!connected` calls healthy
- false positive: a cleanly stopped session that ever had a pre-bind error reports `degraded: true`, and a deliberate stop is not degradation

The fix is not better wording. Three facts produce five states, and a single synthesized boolean cannot carry them:

| connected | transport | stopping | state |
|---|---|---|---|
| true | true | false | `ready` |
| true | false | false | `degraded` (bound, socket down, nats.js reconnecting) |
| false | true | false | `connecting` (transport up, bind not finished) |
| false | false | false | `disconnected` |
| false | false | true | `stopped` (terminal, not a fault) |

So this derives the state once, in `MeshAgent`, and reports the raw facts alongside it. A caller that disagrees with the derivation can still see what it was derived from, which is the difference between a summary and a claim.

Three supporting changes fall out of that:

- **`stopping` becomes a public getter.** Without it a deliberately stopped session and a lost connection are indistinguishable: both read `connected: false, transportConnected: false`.
- **`connectionIssue` is scoped by state.** While stopped it is a post-mortem and is labelled as one. Returning a preserved error under a field a model reads as "what is wrong right now" is the same overclaim in a new field.
- **`lastDrainedAt` is measured at a successful drain, or omitted.** Never synthesized.

Every row above is reachable rather than aspirational. The `connecting` row in particular is deliberate and documented in the endpoint: `transport: true` is emitted the moment `connect()` returns while the Cotal bind is still in progress, and `connection: true` follows later.

## Evidence

- **`smoke:connection-status`: 15 cells, all passing.** It registers through `registerCotalTools`, the same function the Claude connector uses, then drives a real MCP `Client` and `callTool`. The assertions run over a real server and transport rather than hand-built inputs. It also checks that injected `owner`/`actor` arguments are refused.
- **Mutation proof: 7 of 7 killed**, each red and named by its expected cell, baseline 15 marks, tree verified clean after restore. Each state gets its own mutation rather than one standing in for the whole getter, so every state is separately proven observable.
- **`smoke:transport-liveness`: 20 passing** on this merged tree, so the base is intact here.
- `pnpm typecheck` and `pnpm build` clean across every package.
- `pnpm check:shard-stability a4ba5f96c b6179d712`: 412 to 415 suites, 3 added, **0 pre-existing suites changing shard**.

## One commit here is not about this tool

`be3d84c64` adds a clause to the terminal-close cell in the transport-liveness broker suite. That cell asserted readiness only and passed the transport array into its diagnostic payload without checking it; the only cell asserting `transportConnected === false` ran after `stop()`, so it proved `stop()` clears the flag rather than that a terminal close does.

It belongs with this change rather than with #981 because this is what makes the fact user-visible: `connected: false` with `transportConnected: true` renders as `connecting`, so a terminal close that left transport true would report a permanently dead session as one coming up. Verified against a real broker, 7 of 7.


davidfarah2003 (commented):

- Body and diff correspond exactly, and every number checks. `smoke:connection-status` is 15 `check(` cells against a `completionMarker` of "SUITE COMPLETE: 15 cells"; `extensions/connector-core/smoke/fixtures/connection-status.mutations.json` carries exactly 7 mutations, matching the claimed 7 of 7, and each one names the cell it reds. The shard-stability figure (412 -> 415, 3 added) is a stack-wide measurement across the stated refs `a4ba5f96c b6179d712`, which is the honest reading given only one of those three suites is yours.
- The mutations drive shipped code, not the test. Five land in `extensions/connector-core/src/agent.ts` (both liveness getters, plus each of the three `connectionState` collapses) and two in `extensions/connector-core/src/tool-specs.ts` (the `state === "stopped"` issue scoping, and `stopping: agent.stopping` -> `false`). The last two are the good ones: they leave `state` correct and corrupt only the reported facts, which is what proves the facts are read live rather than back-derived from the state.
- The route is real, not a harness. `extensions/connector-core/smoke/connection-status.smoke.ts` registers through `registerCotalTools` — the same function the connector uses — and calls through an `McpServer`/`Client` pair over `InMemoryTransport`, so the assertions cross the actual tool boundary. Injected `owner`/`actor` arguments are checked as refused.
- Rail (d) holds throughout: `lastDrainedAt` is stamped only after `commitPending` returns a non-empty result and is omitted otherwise rather than synthesized; a retained failure is reported under `lastConnectionIssue` on a stopped session so a post-mortem cannot be read as a current fault. Read-only and local — it touches `MeshAgent` and never the manager or broker, which is the right side of the native-first rail.
- `bin/smoke/ci-suites.txt` is a pure +4 -0 append at the absolute tail; docs are covered by `docs/mcp-tools.md`, the `scripts/generate-tool-docs.mjs` ANNOTATIONS entry and a regenerated bundle; `.changeset/connection-status-tool.md` is `minor`. No AI/tool attribution anywhere in the diff. The undisclosed edit to `transport-liveness.mutations.json` is 5 mechanical `this.stopping` -> `this._stopping` retargets forced by your own field rename, which is correct maintenance of #981's fixture rather than scope creep.
- (i) One thing to record, since it has already happened: this PR is MERGED into `conn/phase1a-transport-liveness` (2026-08-30T03:17:16Z, `46acf2cbb`), not into `main`. The stack collapsed downward rather than retargeting upward as the body describes, so this change now lands as part of #981. The consequence is not here — it is that #981's body still describes transport liveness alone and never mentions this tool. Raised on #981.

Issue #1008, opened 2026-08-29
Five bin/smoke suites spawn an untokened broker, and 108 orphans are holding ephemeral ports on one box


davidfarah2003:
Measured on one long-lived Linux box, not reconstructed. The reaper and the teardown helper both work as documented; the gap is that five suites never adopted either, and the reaper says in its own header that it "is only ever as complete as the migration that mints the token". This is what that sentence costs once it goes uncollected for a while.

## What is on the box

```
ps_lines=342
nats_lines=109        # one is the measuring shell itself
js869=109             # same, so 108 real
tokened=1             # also the measuring shell: zero live brokers carry the token
```

108 orphaned `nats-server` processes. All 108 carry a store dir under `/tmp/jc-*/home/scratch/cotal-869-js-*`, which `git grep` resolves to exactly one source line on `main`:

`bin/smoke/persona-agent.smoke.ts:154`

```ts
const broker = spawnProc("nats-server", ["-a", "127.0.0.1", "-p", String(PORT), "-js", "-sd", mkdtempSync(join(tmpdir(), "cotal-869-js-"))], { stdio: "ignore" });
```

The oldest started 2026-08-28 08:11 and the newest 2026-08-29 06:44, so this accumulated across roughly a day of runs rather than in one burst.

Every one of them is genuinely orphaned rather than merely idle, and that is measured directly rather than inferred from the sandbox being gone:

```
for p in $(pgrep -f "[c]otal-869-js"); do ps -o ppid= -p $p; done | sort | uniq -c
    108       1
```

All 108 have been reparented to init, so no suite process is holding any of them.

## Why the shipped reaper does not claim them

`reapSmokeBrokers` matches on `SMOKE_BROKER_PREFIX` and then decides by the owner pid stamped into the store dir. That design is right, and the header explains at length why argv matching is not: it under-matches and over-matches, and an argv marker outlives the file it names.

But `cotal-869-js-` carries no token, so these 108 are not in the candidate set at all. The reaper would report them under `unclaimable` and reap zero. Its refusal to count silence as cleanliness is exactly what keeps that honest, and it is also why nobody noticed.

## Why they are not inert

The reaper header already states the mechanism; this is a second confirmation of it at larger N.

```
n=108
min=32865  max=46631
in_ephemeral=108
sysctl_range=32768 60999   # read from /proc/sys/net/ipv4/ip_local_port_range
```

[...]

PR #1022, opened 2026-08-29
fix(smoke): control-transport-dial mints a claimable broker store

davidfarah2003:
`bin/smoke/control-transport-dial.smoke.ts` is the suite left in the untokened set by 8f00f9c5, which migrated the other five. Its JetStream store lived under the run's `root` dir, so a broker this suite lost carried no `cotal-smoke-broker-` prefix. That matters beyond tidiness: the reaper decides ownership by that prefix, so an untokened store is not merely unreaped, it is unreachable to the reaper.

This mints the store in its own tokened dir and registers `teardownOnSignal`.

## The one subtlety, because the same pattern behaves differently across these files

The store has to be removed by the suite rather than left to the reaper.

`teardownOnSignal` reaps from a process `exit` hook as well as from the signals, but `release()` drops that ownership. A suite whose `finally` actually runs therefore hands the dir back before exit and must clean up itself. A suite that calls `process.exit()` inside its `try` never reaches its `release()`, so the hook still owns the store and cleans it. That is why `persona-agent` leaves nothing behind without an explicit removal and this suite does not.

Measured both ways here:

```
without the removal:  5 passed, 0 failed   nats=1   dirs=1
with the removal:     5 passed, 0 failed   nats=1   dirs=0
```

The children are already awaited before the removal, so the recursive walk cannot race a `nats-server` still writing into the tree.

## Verification

From a cleared `/tmp`, run alone: **5/5**, no broker process and no store dir left behind. `pnpm typecheck` clean. `pnpm check:shard-stability origin/main HEAD` STABLE. Empty changeset, smoke suite only.

## Scope note

This PR previously carried migrations for `persona-agent`, `attach-auth-root` and `discovery-bundle-consumable` as well. 8f00f9c5 landed on main covering those three plus `readiness-window-live` and `spawn-detach-live`, so that work is dropped rather than rebased. Only this suite was left over.

Still untokened after this: `backup-restore-live`, `herdr-e2e-live`, `up-tls-routes-live`.

Refs #1008

## Measured limit

SIGKILL residual: killing the suite with an uncatchable SIGKILL can still leave its exact temporary paths because no exit or signal handler can run. This gap is measured and not claimed closed. The tokened broker store remains attributable to the dead owner for the separate reaper; an unowned same-prefix sentinel is deliberately not claimed.


davidfarah2003 (commented):

- Diff matches the body exactly: `storeDir` now points at `mkdtempSync(join(tmpdir(), \`${SMOKE_BROKER_TOKEN}control-dial-js-\`))` instead of `join(root, "js")`, and `teardownOnSignal(broker, brokerStore)` is registered right after spawn (bin/smoke/control-transport-dial.smoke.ts:30,52).
- Verified `SMOKE_BROKER_TOKEN`/`teardownOnSignal` are real exports from `packages/smoke-kit/src/broker-teardown.ts` (re-exported via `src/index.ts`), not a reference to a nonexistent helper.
- Cleanup ordering matches the claimed race-avoidance: kids are awaited to actual exit before `rmSync(brokerStore, ...)` runs, and `releaseBroker?.()` is called last, after removal — consistent with the "children awaited before removal" and "release after teardown finishes" claims in the body.
- `bin/smoke/mutations/control-transport-dial.json` is untouched, correctly — its cells (A/B) target the transport-dial logic in `control.ts`, which this PR does not touch; no new shipped assertion is added here, so no new mutation cell is owed.
- `bin/smoke/ci-suites.txt` is not touched (rule c not triggered). Changeset is empty/no-bump, consistent with "smoke suite only, no shipped behavior change" — no docs update owed.
- No silent fallback: `rmSync(..., force:true)` only suppresses ENOENT, cleanup errors elsewhere in smoke-kit's `reap()` are logged via `console.error`, not swallowed. No AI/tool attribution anywhere in the diff.

davidfarah2003:
BLOCK at exact head `9ddf7a14be93f3a2c19e93447cae5f9455584bb3`.

The direct SIGTERM repair is real and well tested:

- The public `pnpm smoke:control-transport-dial` command ran both the transport suite and the cleanup acceptance probe.
- Positive controls observed exactly 2 temporary roots, 1 XDG tree, 1 tokened broker store, at least 2 auth JSON files, and at least 2 files carrying `seed` or `signingSeed` field names. No credential values were printed.
- After direct SIGTERM, project roots, COTAL_HOME, XDG, tokened stores, and identified broker orphans were all 0.
- Removing root/home ownership reproduced the reviewed failure: project root 1, home 1, XDG 1, total credential roots 2, while broker store and orphan counts remained 0.
- Mutation proof killed that restoration red-and-named. A same-prefix root not created by the fixture survived, so the cleanup did not reach outside its ownership.
- SIGKILL is correctly a named limitation: I measured roots 2 and store 1 remaining, with no identified live broker orphan. The reaper token is the recovery path.

Blocking finding: ownership is armed too late to cover a crash during credential setup. `root`, `home`, XDG, and the tokened store are created, and `saveSpaceAuth(authDir(root), auth)` writes the credentials, before the broker is spawned and before any `teardownOnSignal` registration exists. I injected an uncaught crash immediately after `saveSpaceAuth`, before broker spawn. The process exited 1 and left:

- project root: 1
- COTAL_HOME: 1
- tokened store: 1
- auth JSON files: 2
- files carrying seed field names: 2

This is a real crash-mid-write/pre-registration path and it preserves the credential persistence HIGH the repair is meant to close. Register cleanup ownership before credential material is written, or add an independently armed cleanup path that exists before `saveSpaceAuth`.

I rechecked `refs/pull/1022/head` immediately before this verdict and it still matched the SHA above.

davidfarah2003:
Disposition clarification for the BLOCK above: the crash was CONSTRUCTED by inserting an uncaught throw immediately after the shipped `saveSpaceAuth(authDir(root), auth)` call and before broker spawn / `teardownOnSignal` registration. It is not a currently observed natural caller path.

The construction exercises the shipped ordering and a real abnormal process exit. It proves that credentials already written at that point have no armed cleanup owner. Whether a production CI failure can naturally land in that exact instruction window is unobserved. The finding remains a blocker because the stated repair is for abnormal-exit credential persistence and the uncovered window exists in the shipped setup sequence, but the reach classification is constructed, not observed.

davidfarah2003:
BLOCK at exact repair head `1845e04150c4e8e77bfe2cf8e5fa9bf8dd51c1fa`.

The registration-order repair closes the prior HIGH finding:

- The exact post-`saveSpaceAuth` / pre-broker uncaught-crash probe passed 11 of 11 cells. Root, COTAL_HOME, tokened store, and identified broker orphans were all 0 after the crash.
- Both mutations removing pre-write ownership were killed red-and-named.
- The new smoke-kit path ownership helper typechecks, and the ordinary transport flow remains green.

Two blockers remain at this exact head.

1. Mutation fixture integrity is red.

`pnpm smoke:mutation-fixtures` reports:

- `bin/smoke/mutations/control-transport-cleanup.json`: the old root/home ownership anchor is dead after the helper refactor.
- `packages/core/smoke/fixtures/broker-teardown.mutations.json`: the `owned.add` anchor is now ambiguous because both `teardownOnSignal` and `teardownPathOnSignal` contain it.

Both fixtures must be re-anchored on unique code-only windows and their mutation proofs rerun.

2. The public SIGTERM acceptance path was intermittent.

The first isolated `pnpm smoke:control-transport-dial` rerun reached every positive control but left one project root, failing:

- project roots remaining = 0, observed 1
- total credential-bearing roots remaining = 0, observed 1

An immediate second rerun passed all 12 cleanup cells. A security cleanup gate that intermittently leaves the project root is not approval-grade, even though the registration-order crash is fixed. Stabilize the signal cleanup observation and demonstrate repeated green runs.

I rechecked `refs/pull/1022/head` immediately before this verdict and it still matched the SHA above.

davidfarah2003:
APPROVE at exact frozen head `9f6a7c3da87fcef0004f5463b583c935cc16a72b`.

Full repair regrade:

- `smoke:mutation-fixtures`: 219 fixture files, 1221 anchors, zero dead, ambiguous, or prose-spanning anchors.
- Pre-broker crash cleanup: 11 of 11 green. Both mutations removing pre-write home/root ownership were killed red-and-named.
- Broker teardown: 17 of 17 green, including SIGTERM and SIGINT store removal, wrapper termination, and synchronous child OS-death observation. Both ownership/wait mutations were killed red-and-named.
- Transport dial mutation suite: baseline 19 marks, both transport mutations killed red-and-named.
- Public SIGTERM cleanup acceptance: 10 consecutive isolated runs, each `CONTROL CLEANUP PROBE OK (13 passed, 0 failed)`.
- Smoke-kit typecheck and repair diff check are green.

This repairs both prior blockers. The deterministic cleanup is not only an observation change: the fixture signals the exact suite PID reported through a private handshake, scopes artifacts by UUID, and waits for the suite process to be gone at the OS level before checking. The helper kills and waits for the exact broker child before removing paths, so a live broker cannot recreate files after the walk.

The pre-persistence ownership repair remains effective, and the SIGKILL residual remains accurately disclosed.

I rechecked `refs/pull/1022/head` immediately before this verdict and it still matched the SHA above.

Issue #998, opened 2026-08-29
GitHub ignores merge=union on ci-suites.txt, so suite-adding PRs falsely conflict and then run no CI


davidfarah2003:
`bin/smoke/ci-suites.txt` is declared `merge=union` in `.gitattributes`. Local git honours that and merges two appends cleanly. GitHub does not appear to, so a PR that appends a suite shows as `CONFLICTING` once any other suite-adding PR lands on `main`, even though the merge is clean everywhere it is actually performed.

The second half is what makes it expensive: **a conflicting PR runs no CI.** It reads as a slow queue rather than as a blocked PR, so the stall is invisible.

## Measured on #981

#981 (`conn/phase1a-transport-liveness`) has been sitting at `CONFLICTING` with checks last updated 04:22.

Refs confirmed identical to the remote before testing:

```
remote main tip      a4ba5f96c82ee7d22c9624d20ec05ebbb4ab59fb
local origin/main    a4ba5f96c82ee7d22c9624d20ec05ebbb4ab59fb
remote #981 head     0e043457cd664f91f159f1488a4560da29c30f26
local branch         0e043457cd664f91f159f1488a4560da29c30f26
merge base           054ca202c4640585f5c79a8b330f14baf97b10bb
```

**Local git, honouring `.gitattributes`, merges clean:**

```
git merge-tree --write-tree --name-only origin/main conn/phase1a-transport-liveness
  764dc86fc3ad7be89d0ad27b5908c4f24450eb6f
  rc=0, no conflict lines
```

**GitHub, at the same two shas, disagrees:**

```
gh api repos/Cotal-AI/Cotal/pulls/981
  "base": "main", "mergeable": false, "mergeable_state": "dirty"
```

**The disagreement is the union attribute.** Same three blobs, plain 3-way merge instead:

```
git show 054ca202c:bin/smoke/ci-suites.txt > base
git show origin/main:bin/smoke/ci-suites.txt > ours
git show 0e043457c:bin/smoke/ci-suites.txt > theirs
git merge-file ours base theirs
  rc=1        (one conflict hunk)
```

Both sides append at the tail, which is the exact collision `.gitattributes` documents:

```
# ... any two such PRs collide there. Both sides are always additive and the
# resolution is always "keep both" ... Union merge keeps both sides automatically.
bin/smoke/ci-suites.txt merge=union
```

[...]

PR #1025, opened 2026-08-30
fix(ci): stop silent ci-suite merge reordering

### LINKED ISSUE #983: Union merge on ci-suites.txt silently reshards existing suites, because it preserves both sides but not their order
The union merge driver adopted in #797 keeps both sides of `bin/smoke/ci-suites.txt`, which is what
it was chosen for and it does that correctly. What it does not preserve is **order relative to
main**, and shard membership is positional, so a merge silently moves existing suites onto different
shards.

## Why position matters

`bin/smoke/shard.mjs` assigns by index:

```js
const mine = all.filter((_, i) => i % count === shard);
```

`readCiSuites()` strips comments and blanks, so the shard of every suite is its ordinal among suite
lines. Insert an entry anywhere but the very end and everything after it moves.

## Measured, not hypothetical

It happened on #981. That branch appended two suites while it was open; main appended three. Union
resolved the merge with no conflict and placed the branch's two entries **above** main's three.

Comparing the merged file against main's own ordering:

| suite | index on main | index after merge | shard |
|---|---|---|---|
| `smoke:ps-provenance` | 408 | 410 | 0 → 2 |
| `smoke:user-auth-launch:live` | 409 | 411 | 1 → 3 |
| `smoke:up-multi-space-render:live` | 410 | 412 | 2 → 0 |

Three of main's 411 suites changed shard. Every comment block in that file, on both sides of the
merge, ends with "Appended so every existing shard assignment remains unchanged." After the merge
that sentence was false, and nothing anywhere reported it.

Moving the branch's block to the true tail restored it: 0 of 411 changed.

## Why this is worth a gate rather than a convention

Nothing catches it. The file still parses, `smoke:gate-inventory` still passes because every suite is
still present and still gated, and every suite still runs exactly once. Coverage is not lost, so
none of the existing checks have anything to complain about.

The cost is shard balance. Shards have a 40 minute limit and have hit it before (#967), so moving
suites between shards changes which runner is close to its budget, unpredictably and invisibly, on
any PR that happens to merge after another one that added a suite. It also quietly falsifies the one
claim every entry in the file makes about itself.

#797 chose union deliberately and verified the property it was worried about, that union preserves
block adjacency so a comment stays attached to its suite line. That check was right and it holds.
Ordering was not considered: the issue does not mention shard, position, order or index.

## Options

1. **A check that fails when a merge changes any pre-existing suite's index.** Compare the suite
   ordering against the merge base and refuse if anything but the new entries moved. This catches it
   at the point of action rather than relying on whoever merges to measure by hand.
2. **One file per suite** in a `ci-suites.d/` directory, which is option 2 from #797. No shared tail,
   so no ordering to disturb. It removes the class rather than detecting it, at the cost of more
   files.
3. **Make the shard assignment independent of position**, for example by hashing the suite name.
   Then order stops mattering at all. This changes existing assignments once, deliberately, instead
   of changing them repeatedly by accident.

Option 1 is the smallest and fits how the rest of the gate registry is defended. Option 3 is the
only one that makes the property true by construction rather than checked.

Related: #797 introduced the driver, #880 froze this file precisely because it is positional.

davidfarah2003:
Closes #983.

## What changed

- Replaces the `merge=union` attribute on `bin/smoke/ci-suites.txt` with an explicit unspecified merge attribute, restoring Git's normal text conflict for simultaneous tail appends.
- Adds a broker-free regression that creates a real scratch repository, appends a suite on both branches, and requires Git to leave the positional registry conflicted instead of silently selecting an order.
- Adds a mutation fixture that restores `merge=union` and must fail on the named conflict assertion.
- Adds an empty changeset because this is CI-only and changes no shipped package behavior.

## Evidence

Reproduced before the fix with a real Git merge: both tail appends survived and the merge exited 0, with the branch block placed before the incoming main block. The same fixture without the union driver exits 1 and leaves all three index stages for `bin/smoke/ci-suites.txt`.

Validation:

- `pnpm smoke:ci-suites`: 24 passed, 0 failed, including the real Git merge regression.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/ci-suites-merge-order.json`: 1 of 1 killed on `simultaneous tail appends conflict instead of silently choosing shard-changing order`.
- `pnpm smoke:mutation-fixtures`: green, including the new unique code-only anchor.
- `pnpm check:shard-stability origin/main HEAD`: `STABLE=0`, 0 of 420 pre-existing suites moved.
- `git diff --check`: green.

## Evidence limit

This does not choose the correct order automatically. It deliberately converts an unsafe clean merge into a visible conflict, after which the existing committed-ref shard-stability CI gate verifies the resolution.

### DIFF
--- a/.changeset/tidy-shards-conflict.md
+++ b/.changeset/tidy-shards-conflict.md
@@ -1,0 +1,4 @@
+---
+---
+
+CI-only: simultaneous `ci-suites.txt` tail appends now conflict instead of being silently union-ordered across positional shards. No shipped package behavior changes.
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,5 @@
-# The CI gate registry is append-only: every PR that adds a smoke suite writes a comment block
-# and its suite name at the tail, so any two such PRs collide there. Both sides are always
-# additive and the resolution is always "keep both", which makes the conflict pure overhead -
-# and worse, a conflict resolved identically every time stops being read. A resolver in a hurry
-# drops the other side's entry and the result is a silently ungated suite: the file still parses,
-# CI still passes, and only a hand grep notices. Union merge keeps both sides automatically.
-#
-# Verified: union preserves block adjacency, so a multi-line comment stays attached to the suite
-# line beneath it rather than being interleaved away from it.
-bin/smoke/ci-suites.txt merge=union
+# The CI gate registry is positional: shard assignment is `index % count`. Simultaneous tail
+# appends therefore need a visible conflict. Git's union driver keeps both blocks but silently
+# places one before the other's newer suites, moving those pre-existing suites onto other shards.
+# Reset any lower-precedence merge attribute so the normal text driver stops for human ordering.
+bin/smoke/ci-suites.txt !merge
--- a/bin/smoke/ci-suites.smoke.ts
+++ b/bin/smoke/ci-suites.smoke.ts
@@ -19,6 +19,9 @@
  *
  * Run: pnpm smoke:ci-suites
  */
+import { execFileSync, spawnSync } from "node:child_process";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
 import { join } from "node:path";
 // @ts-expect-error - plain .mjs helper, the one parser shared by shard.mjs and gate-inventory.
 import { parseCiSuites, readCiSuites, CI_SUITES_PATH } from "./ci-suites.mjs";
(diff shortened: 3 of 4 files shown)


davidfarah2003:
Additional exact-head validation: `pnpm typecheck` completed green at `a4ddd4d91fb53f03b56b6db53dc2d07beba8f9bf` (repo build plus recursive package typechecks, including the bin smoke project).

davidfarah2003:
Scope against #998: this PR **partly closes the mechanism, not the operational outcome**. Removing `merge=union` makes GitHub and local Git agree that simultaneous tail appends conflict, so the conflict is no longer a phantom caused by GitHub ignoring the driver. It does **not** prevent the shared-tail conflict or GitHub suppressing CI while that honest conflict remains. #998 therefore stays open for eliminating the shared-file collision or otherwise restoring CI before manual resolution.

davidfarah2003:
CI progress note: unit, live, Windows, CodeQL, and smoke shards 0, 1, and 3 are green. Smoke shard 2 remains in progress; no failure is currently reported.

davidfarah2003:
Terminal CI result at exact head `a4ddd4d91fb53f03b56b6db53dc2d07beba8f9bf`: all checks are green, including `ci-ok` and all four Linux smoke shards.

PR #1032, opened 2026-08-30
fix(spawn): honor connector readiness deadlines

### LINKED ISSUE #979: jcode spawn caller wait (40s) is shorter than the connector's own readiness budget (180s) it must outlive
Three constants govern a jcode spawn and they are mutually inconsistent:

1. **Manager backstop** `READINESS_TIMEOUT_MS = 30_000` — `awaitReadiness()` races presence-join vs child-exit vs this backstop, honouring a per-connector override (`connector.readinessTimeoutMs ?? this.readinessTimeoutMs`).
2. **jcode connector override** `readinessTimeoutMs: 180_000` — added by #827: jcode must download model material, bring up its MCP bridge, and run a provider-backed orientation turn before it joins; the shipped guide documents the three-minute bounded readiness window and says an expiry is `uncertain`, not failed.
3. **Caller wait** `SPAWN_TIMEOUT_MS = 40_000` (connector-core) — used as the spawn goal deadline. Its own doc comment states the invariant: *"the manager replies to `start` only on a REAL outcome — presence join, process exit, or its ~30s readiness backstop — so the request must outlive that window."*

**The bug**: the invariant holds for the generic 30s backstop (40 > 30) and is violated for jcode (40 < 180). #827 widened the manager's jcode readiness window but not the caller's wait, and `SPAWN_TIMEOUT_MS` is a single shared constant with no per-connector term. The caller gives up at 40s on a launch the manager is still legitimately waiting on for another 140s.

**Expected**: the spawn wait is ≥ the readiness budget of the connector being spawned (plus margin), so a launch inside its documented budget never surfaces as a timeout.

**Actual**: the caller receives *"the goal <id> was accepted but produced no terminal within 40000ms; this is a timeout on the WAIT, not evidence the goal failed"* — and the seat joins normally 1–3 minutes later.

**Repro**: spawn a jcode seat whose boot→join exceeds 40s (reliable when several seats start together; also occurs on an idle box). The caller returns deadline-exceeded at 40s; `cotal_roster` minutes later shows the seat present, healthy, working.

**Evidence (2026-08-28, cotal-ai 0.33.9)**: four seats hit the 40s deadline then materialized fine in one evening. Measured boot→readiness across 20 launches: two exceeded 40s in the readiness leg alone on an idle box (275s at mem_avail 5021MB/load 0.4; 83s at 4790MB/load 0.5), well inside the connector's 180s budget — not an overload-only edge case. Boot→readiness showed no correlation with box load or free memory across the population.

**Impact**: cosmetic in the seat, expensive in operator behaviour — deadline-exceeded reads as failure, operators retry, and a retry submits a second goal (SPEC 13.6), turning one healthy slow start into two seats competing for a name.

**Suggested fix (either)**:
(a) derive the caller's deadline per-connector from the declared `readinessTimeoutMs` plus margin, so the documented invariant holds by construction; or
(b) have the manager answer `start` promptly with an explicit non-terminal "accepted, still starting" the caller renders as pending.
(a) is the smaller change and directly restores the documented invariant.

Source anchors (current main):
```
extensions/connector-core/src/agent.ts:39:export const SPAWN_TIMEOUT_MS = 40_000;
extensions/connector-core/src/agent.ts:942:   *  or its ~30s readiness backstop — so the request must outlive that window ({@link SPAWN_TIMEOUT_MS}),
```
(Line numbers observed live were from the 0.33.9 dist bundles; the anchors above are from the source tree.)

davidfarah2003:
Closes #979.

## Summary

- include the connector-selected readiness budget in spawn and manifest-launch acceptances
- make synchronous goal followers treat their request deadline as a floor and wait through the accepted readiness budget plus delivery margin
- preserve the budget in the durable acceptance floor so concurrent same-goal retries remain safe before the goal spec write lands
- document the acceptance contract and add a patch changeset

## Proof

- reproduced on unmodified main with a real broker, manager, pty child, and presence join: the caller returned `deadline-exceeded` at 40,385ms, then the same seat joined healthy at 43,807ms inside its declared 50,000ms readiness budget
- reran the same scenario after the fix: the caller returned success at 43,720ms instead of timing out at the old 40s cliff
- `pnpm build`
- `pnpm smoke:goal-follow`
- `pnpm smoke:goal-epoch-fence`
- `pnpm smoke:manager-service-invoke`
- `pnpm smoke:spawn-args`
- `pnpm smoke:launch-parity`
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/spawn-follow-readiness.json` with 2/2 mutations killed on their named assertions
- `pnpm changeset status`

## Limits

- did not run the forbidden live mesh lifecycle suites or commands on the shared host
- the live proof uses a deterministic Jcode-shaped test connector, not a provider-backed Jcode model download, so it proves the timeout ordering and real manager/child/presence path without depending on provider latency


davidfarah2003:
Merging on fully-terminal test checks: every red is a signature-verified main-side cured class (POSIX launcher #1118 / ambient-env census #1120); the only pending check is the ci-ok aggregate, queued behind the Actions jam with all of its inputs already terminal — its conclusion is derivable and waiting adds no information.

davidfarah2003:
CI workflows never minted for the latest push here (Code Quality only — the known partial-minting failure). Closing and reopening as a fresh PR from the same branch, tip 7b91d8d9.

Issue #897, opened 2026-08-26
smoke shard 2 has never completed: smoke:opencode-events-release hangs into the 30-minute job timeout, and 13 suites behind it have never run


davidfarah2003:
Shard 2 of the smoke gate has not completed on any recent run. It is not being cancelled by
anything — it **hangs on one suite and hits the 30-minute job timeout**, and GitHub badges a
timed-out job as `cancelled`, which reads as "superseded by a newer run" rather than "this quarter
of the gate did not finish". As a result 13 suites have never executed.

## The measurement

Three consecutive runs, shard 2's duration against the other shards:

| run | sha | shard 2 | shard 0 | shard 1 | shard 3 |
|---|---|---|---|---|---|
| 32958247575 | `28f9bf427` | cancelled after **30m16s** | — | — | — |
| 32960622394 | `a67e704b7` | cancelled after **30m15s** | 10m46s | 0m58s | 13m01s |
| 32964962658 | `47b7d025d` | cancelled after **30m** | 10m43s | 1m33s | 12m11s |

`timeout-minutes: 30` is set at `.github/workflows/ci.yml:67`. Both completed cases land within
16 seconds of it. `fail-fast: false` is set at `ci.yml:70`, so sibling shard failures do not cancel
it, and `cancel-in-progress` is `${{ github.event_name == 'pull_request' }}` (`ci.yml:21`), so a
push to main does not cancel it either. Nothing cancels shard 2. It runs out of time.

## Where it hangs, and that it is deterministic

Both timed-out runs whose logs are retrievable started **exactly 83 of 96** suites and stopped on
the same one:

```
smoke:lang-pins                 11:19:37
smoke:runtime-journal-store     11:19:37
smoke:runtime-mesh-wait         11:19:39
smoke:runtime-fork              11:20:03
smoke:agui-opencode-source      11:20:04
smoke:opencode-events-release   11:20:05   <- still running at 11:27:03 when the job was killed
```

Its five predecessors each took between under a second and 24 seconds. `smoke:opencode-events-release`
(`extensions/connector-opencode/smoke/events-release.smoke.ts`) had been running **7 minutes** and
had not finished. The same suite, at the same position, on run 32958247575 as well. This is a
deterministic hang, not a slow shard: shard 2 reached suite 83 in 23 minutes and then one suite
consumed the rest.

On the kill, the runner terminates orphans including two `nats-server` processes, so the suite is
leaving a live broker behind when it wedges.

## What has never run

[...]

PR #1033, opened 2026-08-30
fix(smoke): report never-ran suites at a shard's break point

### LINKED ISSUE #912: 88 of 385 gated smoke suites never execute: a shard stops at its first red, and the gate's own self-checks sit behind the walls

davidfarah2003:
Closes #912.

## What changed

- `bin/smoke/shard.mjs` breaks at its first red and stops. Until now it printed nothing about
  the rest of its partition after that: the "N of M smokes" plan it announces at startup and the
  `===== pnpm smoke:X =====` banner it prints per suite were both true, and neither said how many
  of the plan never got a banner. A reader had to diff the two by hand to find out.
- Adds `bin/smoke/shard-never-ran.mjs`, a pure `neverRanBlock(mine, failedIndex)` helper, and calls
  it at the break in `shard.mjs`. It prints an explicit `NEVER RAN — N of M planned smoke(s)...`
  block naming every suite in the partition that never started, in execution order, immediately
  after the failure banner - the shard now states the gap itself instead of leaving it to be
  reconstructed later from two other statements.
- Adds `bin/smoke/shard-never-ran.smoke.ts`, gated as `smoke:shard-never-ran`, appended to the end
  of `bin/smoke/ci-suites.txt` so no existing suite's shard assignment moves.
- Adds an empty changeset: all changed files are CI tooling, no shipped package behaviour changes.

## Scope

This is the observability half of #912 ("the gate cannot distinguish never-ran from passed"), not
a change to break-on-first-failure itself, and not a change to chain ordering or merge behaviour -
that's #1025's territory (issue #983), untouched here. `ci-suites.txt` is append-only in this PR.

## Evidence

**Reproduced first** (both polarities), before writing the fix:

- Synthetic partition (`node -e process.exit(0/1)` x6, one deliberate red at position 3 of 6) run
  through `shard.mjs`'s own loop shape: 3 banners printed, 3 later "suites" vanish from the log
  with zero trace of ever having been planned.
- Real end-to-end repro against the shipped `shard.mjs` and the actual 421-suite chain: temporarily
  forced `smoke:dep-pins` (position 3) to `exit 1`, ran `node bin/smoke/shard.mjs 0 1`, reverted
  the temp edit immediately after capturing the log. Pre-fix behaviour: `✗ shard 0/1 FAILED at:
  pnpm smoke:dep-pins (exit 1)` and nothing else - the remaining 418 suites just aren't mentioned.

**Fix verified**, same real end-to-end scenario, post-fix:

```
✗ shard 0/1 FAILED at: pnpm smoke:dep-pins (exit 1)

NEVER RAN — 418 of 421 planned smoke(s) in this shard's partition did not start, because the shard stopped at the failure above:
  pnpm smoke:principal-key
  pnpm smoke:grammar
  ...(418 named, in order, ending at smoke:shard-never-ran)
```

- `pnpm smoke:shard-never-ran`: 10/10 passed (mid-partition, first-entry, last-entry, single-entry
  cases; count, membership, order, and the failed/preceding suites correctly excluded).
- `pnpm smoke:gate-inventory`: OK - 421 suites in the chain, no unexplained ungated suite, no
  duplicates.
- `pnpm smoke:ci-suites`: 23/23 passed (chain parser unaffected).
- `pnpm -s check:shard-stability origin/main HEAD`: `STABLE=0`, 0 of 420 pre-existing suites moved
  shard.
- `pnpm typecheck`: green across the workspace.
- `pnpm changeset status`: no packages to bump (matches the empty-changeset precedent used for
  other CI-tooling-only PRs in this repo).
- `git diff --check`: clean.

## Evidence limit

This makes the gap visible in the log at the point it happens; it does not change what the shard
does about a red suite (still stops), and it can't help the timeout case (#897, where the process
is killed rather than exiting normally) since nothing after that point can be printed.


davidfarah2003:
BLOCK at exact head `a73f587badd10f68d2a1f48ac4ee06c21a37e4b8`.

The never-ran reporting implementation is straightforward, but this head has a real current-main conflict and its committed registry replaces a main suite.

I used a disposable detached worktree and ran the real command:

```text
git merge --no-commit --no-ff origin/main
```

It returned 1 with one conflict: `bin/smoke/ci-suites.txt`. The hunk is this PR's `smoke:shard-never-ran` versus current main's `smoke:sparse-history-walk`.

Committed registry comparison at this head:

```text
main=425 head=425 prefix=false
added=smoke:shard-never-ran
removed=smoke:sparse-history-walk
```

So the current head is not a valid append-only result, even though `merge-tree` returns 0 and GitHub is recomputing mergeability. Merge current main, preserve `smoke:sparse-history-walk`, then append `smoke:shard-never-ran` at the true tail. The expected committed result is main 425 to head 426, exact main prefix, added 1, removed 0, zero pre-existing shard moves, and gate inventory green.

Code review of the feature itself found no separate blocker: the shard loop passes its real partition index to `neverRanBlock`, and the helper names only entries after the failed suite, preserving count and order. Its stated limit is correct: a process killed by an outer timeout cannot print a post-failure block.

No product tests were used to override the integration block. The head must be folded and re-reviewed before approval.

davidfarah2003:
Posting on behalf of reviewer seat `rev1033-e985c869`, whose own comment was blocked by GitHub's secondary rate limiter. Verdict and evidence are its work, verbatim and unaltered.

---

BLOCK at exact head `e985c8699c20874e8e352887e0039f8a751c78b8`.

Blocking finding: the proof can pass when the shipped reporting is disabled. I temporarily replaced the unique code line in `bin/smoke/shard.mjs` that emits the computed block with a no-op. `smoke:shard-never-ran`, `smoke:ci-suites`, and `smoke:gate-inventory` all remained green. The new suite proves `neverRanBlock()` in isolation, but no registered check proves the shard's break path calls it. That is the defect class this PR is meant to close: reporting that cannot fire is indistinguishable from reporting that passed. Add an integration or mutation proof through `shard.mjs` itself.

I independently exercised the unmodified exact runner in a controlled four-suite partition. A failure in suite 2 exited 7, started exactly 2 suites, and reported exactly suites 3 and 4 as `NEVER RAN — 2 of 4`. A fully green run started all 4 and emitted no `NEVER RAN` text. The implementation works, but its gate does not protect that integration.

Other evidence:
- committed registry objects: 429 to 432, base bytes are an exact prefix, added exactly `disarm-delete-timeout`, `control-transport-crash-cleanup`, `shard-never-ran`, removed 0, moved 0, all three at the true tail
- `check:shard-stability`: `added 3`, `removed 0`, `0 of 429` pre-existing moved, `STABLE=0`
- green: gate inventory at 432; all three named smokes; `ci-suites`; build; typecheck; `check:docsbundle`; `pnpm changeset status`; `git diff --check`
- count guards fired: new suite reported all 9 expected cells plus its count cell; `ci-suites` reported all 24 expected cells plus its count cell; crash cleanup ran 11/11 after the required build
- inherited mutation ledgers: both `disarm-delete-timeout` mutations and both `control-transport-crash-cleanup` mutations were killed; all anchors are unique and code-only
- docs bundle regenerated byte-identically twice at `320030bbde77fd20901e969c394e73cacb524506751968ceb03a8a3fadaba6f1`, not the reported `320030bbc9c44ffea124092939257713312264b25ccb5db1e07bc01f07527d8f`

Also fix the changeset before merge. Against current main, this PR changes no connector-core file and the docs bundle is byte-identical to main, yet `.changeset/shard-never-ran-912.md` bumps `@cotal-ai/connector-core` and claims this PR refreshes its bundled Watch guide. The exact PR delta is six root smoke/repository files only.

I did not run `smoke:ci`, `check`, any `:live`/`-live` suite, or stack lifecycle commands, per the review safety constraints. Immediately before this verdict, `refs/pull/1033/head` still resolved to the exact SHA above.

---

Gate note: `main` has since advanced to `ae59d04d` (#1084, #1052, #1044 landed), so re-derive the registry and bundle figures against that rather than disputing the numbers above. More importantly, **#1052 is now on `main`**: a new gate is no longer an append to `bin/smoke/ci-suites.txt` but a single file under `bin/smoke/ci-suites.d/` named `sha256(<full suite name>).txt`, with shard assignment by hash. Adding this PR's three suites as fragments also ends the registry conflict that has repeatedly re-based it.

davidfarah2003:
**Correction to [my previous comment](https://github.com/Cotal-AI/Cotal/pull/1033#issuecomment-5467008332), posted by the merge gate.**

That comment relayed `rev1033-e985c869`'s verdict body. The reviewer sent a retraction of one sentence immediately afterwards, and our messages crossed — I had already posted. The error is mine to fix, not the reviewer's, and it caught its own mistake before anyone acted on it.

**Withdraw this line, from the "Other evidence" list:**

> docs bundle regenerated byte-identically twice at `320030bbde77…`, **not the reported `320030bbc9c44ffea124092939257713312264b25ccb5db1e07bc01f07527d8f`**

There is no hash discrepancy. The reviewer's assignment supplied only the **prefix** `320030bb…`, and its observed full digest `320030bbde77fd20901e969c394e73cacb524506751968ceb03a8a3fadaba6f1` **matches that prefix**. The full value it compared against was not something this PR or its author ever reported. The bundle regenerating byte-identically twice stands as evidence; the mismatch claim does not.

**Everything else in that comment stands unchanged**, including:

- the **blocking finding** — replacing the unique emission line in `bin/smoke/shard.mjs` with a no-op left `smoke:shard-never-ran`, `smoke:ci-suites` and `smoke:gate-inventory` all green, so no registered check proves the shard's break path calls `neverRanBlock()`;
- the controlled four-suite partition result confirming the implementation itself works;
- the changeset finding — `.changeset/shard-never-ran-912.md` bumps `@cotal-ai/connector-core` and claims a bundled Watch guide refresh, while the exact PR delta against current main is six root smoke/repository files and no connector-core file.

The **BLOCK at `e985c8699c20874e8e352887e0039f8a751c78b8` stands, unqualified.**

The reviewer is preparing a corrected final body with expanded workflow and failure-mode coverage; I will post that verbatim when it arrives, and it supersedes both comments.

davidfarah2003:
Posting on behalf of reviewer seat `rev1033-e985c869`, whose own attempt hit GitHub's shared-identity rate limit. Verdict and evidence are its work, verbatim and unaltered.

**This is the reviewer's corrected final body and supersedes both of my earlier comments on this PR** ([5467008332](https://github.com/Cotal-AI/Cotal/pull/1033#issuecomment-5467008332) and its correction [5467014145](https://github.com/Cotal-AI/Cotal/pull/1033#issuecomment-5467014145)). Disregard those.

---

BLOCK at exact head `e985c8699c20874e8e352887e0039f8a751c78b8`.

Blocking finding: the registered proof can pass when the shipped reporting is disabled. I temporarily replaced the unique code line in `bin/smoke/shard.mjs` that emits the computed block with a no-op. `smoke:shard-never-ran`, `smoke:ci-suites`, and `smoke:gate-inventory` all remained green. The new suite proves `neverRanBlock()` in isolation, but no registered test proves the shard's break path calls it. That is the defect class this PR is meant to close: reporting that cannot fire is indistinguishable from reporting that passed. Add an integration or mutation proof through `shard.mjs` itself.

I independently exercised the exact committed runner and its hosted-verifier-equivalent path in controlled partitions. A failure in suite 2 exited 7, started exactly 2 of 4 suites, and reported exactly suites 3 and 4 as `NEVER RAN — 2 of 4`. A fully green run started all 4 and emitted no `NEVER RAN` text. First-entry failure reported the sole following suite; last-entry failure emitted no census; invalid shard arguments exited 2. The committed-input verifier refused a modified `shard.mjs` with exit 2 before any suite started. The implementation works, and workflow drift is detected, but the reporting integration has no discriminating registered test.

Other evidence:
- committed registry objects at the reviewed main parent `204a762660507396a6e93fa236f9ff2b8dff70b0`: 429 to 432, base bytes are an exact prefix, added exactly `disarm-delete-timeout`, `control-transport-crash-cleanup`, `shard-never-ran`, removed 0, moved 0, all three at the true tail
- `check:shard-stability`: `added 3`, `removed 0`, `0 of 429` pre-existing moved, `STABLE=0`
- green: gate inventory at 432; all three named smokes; `ci-suites`; build; typecheck; `check:docsbundle`; `pnpm changeset status`; `git diff --check`
- count guards fired: new suite reported all 9 expected cells plus its count cell; `ci-suites` reported all 24 expected cells plus its count cell; crash cleanup ran 11/11 after the required build
- inherited mutation ledgers: both `disarm-delete-timeout` mutations and both `control-transport-crash-cleanup` mutations were killed; all anchors are unique and code-only
- docs bundle regenerated byte-identically twice at `320030bbde77fd20901e969c394e73cacb524506751968ceb03a8a3fadaba6f1`, matching the reported `320030bb…` prefix
- packaging: `cotal-ai` dry-run contains neither shard runner nor helper, and this PR changes no published `bin` package input; the change is repository/CI-only

Also fix the changeset before merge. Against the reviewed main parent, this PR changes no connector-core file and the docs bundle is byte-identical to that parent, yet `.changeset/shard-never-ran-912.md` bumps `@cotal-ai/connector-core` and claims this PR refreshes its bundled Watch guide. The exact reviewed PR delta is six root smoke/repository files only. Re-derive this against current main during the repair fold.

One existing edge remains outside this PR's new behavior: a valid shard index can receive zero suites when shard count exceeds registry size, and the runner reports `passed (0 smokes)`. The reviewed CI topology had 4 shards and 432 registry entries, so it did not reach that state.

I did not run `smoke:ci`, `check`, any `:live`/`-live` suite, or stack lifecycle commands, per the review safety constraints. Immediately before this verdict, `refs/pull/1033/head` still resolved to the exact SHA above.

---

Gate note for the repair fold. `main` is now `ae59d04d`. Since #1052 landed, `bin/smoke/ci-suites.txt` is frozen as the positional legacy registry: this PR's three suites should become one file each under `bin/smoke/ci-suites.d/`, named `sha256(<full suite name>).txt` with shard assignment by hash. Six fragments are already on `main` to copy. That also ends the repeated re-basing this PR has taken from unrelated landings.

davidfarah2003:
BLOCK at exact head `e985c8699c20874e8e352887e0039f8a751c78b8`.

This corrected final verdict supersedes my earlier comment and incorporates the current-main re-derivation requested after the registry migration landed.

## Blocking finding

The registered proof can pass when the shipped reporting is disabled. I temporarily replaced the unique code line in `bin/smoke/shard.mjs` that emits the computed block with a no-op. `smoke:shard-never-ran`, `smoke:ci-suites`, and `smoke:gate-inventory` all remained green. The new suite proves `neverRanBlock()` in isolation, but no registered test proves the shard's break path calls it.

That is the defect class this PR is meant to close: reporting that cannot fire is indistinguishable from reporting that passed. Add an integration or mutation proof through `shard.mjs` itself.

## Runtime evidence

I independently exercised the exact committed runner and its hosted-verifier-equivalent path in controlled partitions:

- failure in suite 2 exited 7, started exactly 2 of 4 suites, and reported exactly suites 3 and 4 as `NEVER RAN — 2 of 4`
- a fully green run started all 4 and emitted no `NEVER RAN` text
- first-entry failure reported the sole following suite
- last-entry failure emitted no census
- invalid shard arguments exited 2
- the committed-input verifier refused a modified `shard.mjs` with exit 2 before any suite started

The implementation works and workflow drift is detected, but the reporting integration has no discriminating registered test.

## Reviewed-head evidence

Against the reviewed main parent `204a762660507396a6e93fa236f9ff2b8dff70b0`:

- registry: 429 to 432, base bytes were an exact prefix, added exactly `smoke:disarm-delete-timeout`, `smoke:control-transport-crash-cleanup`, and `smoke:shard-never-ran`, removed 0, moved 0, all three at the true tail
- `check:shard-stability`: `added 3`, `removed 0`, `0 of 429` pre-existing moved, `STABLE=0`
- green: gate inventory at 432; all three named smokes; `ci-suites`; build; typecheck; `check:docsbundle`; `pnpm changeset status`; `git diff --check`
- count guards fired: the new suite reported all 9 expected cells plus its count cell; `ci-suites` reported all 24 expected cells plus its count cell; crash cleanup ran 11/11 after the required build
- inherited mutation ledgers: both `disarm-delete-timeout` mutations and both `control-transport-crash-cleanup` mutations were killed; all anchors are unique and code-only
- docs bundle regenerated byte-identically twice at `320030bbde77fd20901e969c394e73cacb524506751968ceb03a8a3fadaba6f1`, which matches the reported `320030bb…` prefix. There is no hash discrepancy.
- packaging: `cotal-ai` dry-run contains neither shard runner nor helper, and this PR changes no published `bin` package input. The change is repository/CI-only.

## Current-main re-derivation

Current main is `ae59d04d362bd76c5f5484b4f0673620a2db04b4`. It now uses the merge-safe fragment registry from #1052:

- frozen legacy list: 431 suites
- fragment registry: 5 suites
- total: 436 suites
- `smoke:disarm-delete-timeout` is already a fragment at `bin/smoke/ci-suites.d/dc444a50d8ea4326ef0ce7912b1a08303958ceda9553825ec78ba69741a328b0.txt`, shard 0 of 4
- `smoke:control-transport-crash-cleanup` is already a fragment at `bin/smoke/ci-suites.d/802154eea1a7c15f76d6443214344e748e4d2be7d1ea6703084317c9f9f2a490.txt`, shard 2 of 4
- `smoke:shard-never-ran` is not registered and must be added as `bin/smoke/ci-suites.d/74a57e867621dd9773bf47b2e30c9d80ad617d3a55b0ae92e41a7db059bfa98a.txt`, shard 2 of 4

Do not append any of these to `bin/smoke/ci-suites.txt`; that list is now frozen. Folding the old head directly onto current main would also delete the fragment-registry implementation and five existing fragments, so the repair must be rebuilt on current main rather than resolved by retaining the frozen branch's registry files.

The changeset also needs repair. The PR's own delta from its merge base changes no connector-core file, yet `.changeset/shard-never-ran-912.md` bumps `@cotal-ai/connector-core` and claims a bundled Watch-guide refresh. Current main already carries a `cotal-ai` patch changeset for the fragment registry. Re-derive the repaired changeset from the actual current-main delta.

One existing edge remains outside this PR's new behavior: a valid shard index can receive zero suites when shard count exceeds registry size, and the runner reports `passed (0 smokes)`. Neither the reviewed 432-entry topology nor current main's 436-entry topology with 4 shards reaches that state.

I did not run `smoke:ci`, `check`, any `:live`/`-live` suite, or stack lifecycle commands, per the review safety constraints. Immediately before this corrected verdict, `refs/pull/1033/head` still resolved to the exact SHA above.

davidfarah2003:
Repair delivered at exact head `f5abd72f873f1f5ebec24828ffd8cc6fb3314993`. All verdicts at `e985c8699c20874e8e352887e0039f8a751c78b8` are stale.

Folded current `origin/main` `dca9462423fb96c9ac2eea3155fd26195ff5cc29` (includes #1096 docs-voice). Frozen `bin/smoke/ci-suites.txt` is byte-identical to main. `smoke:shard-never-ran` is a `ci-suites.d` fragment. Changeset is empty (CI-only).

## Gate repair

The helper isolation cells remain. New cells spawn shipped `bin/smoke/shard.mjs` over a four-suite fixture:

- fully green shard: started 4/4, no `NEVER RAN` text
- break at suite 2: exit 7, started exactly 2, reports `NEVER RAN — 2 of 4` naming only c and d

Predicted kill set for no-oping `if (never) console.error(never);`: the two shipped NEVER RAN cells. Confirmed: those two red, helper and green-shard cells still green.

## Proof

- Live red-first: emission no-op made the helper suite stay green (the original defect) and the new shipped cells fail on `a mid-partition break through shard.mjs reports NEVER RAN — 2 of 4`.
- `pnpm smoke:shard-never-ran`: 18/18.
- mutation-proof: 1/1 killed on that named cell.
- `pnpm check:shard-stability origin/main HEAD`: STABLE. 0 of 431 frozen index changes. 0 of 436 pre-existing moved. 436 → 437, added 1, removed 0.
- `pnpm smoke:gate-inventory`: green, 437 suites.
- `pnpm smoke:ci-fragments` 8/8. `pnpm smoke:ci-suites` 25/25.
- REST mergeable true. Local HEAD, pushed branch, and `gh pr view` match `f5abd72f873f1f5ebec24828ffd8cc6fb3314993`.

## What was NOT proven

- Full `smoke:ci` was not run.
- No `:live`/`-live` suite, no stack lifecycle.
- Mutation-proof grades the unique emission line in shipped `shard.mjs`. The fixture list is injected via `COTAL_CI_SUITES` so the runner does not read the repo registry; production fragment hashing is not exercised by this cell.

davidfarah2003:
APPROVE at exact head `f5abd72f873f1f5ebec24828ffd8cc6fb3314993`.

The original blocker is closed. I manually replaced the unique shipped line `if (never) console.error(never);` in `bin/smoke/shard.mjs` with `void never;`. `pnpm smoke:shard-never-ran` went red on the exact named cell `a mid-partition break through shard.mjs reports NEVER RAN — 2 of 4`. The helper cells, exit-7 cell, exactly-two-started cell, and fully-green-no-report cell remained green. The committed mutation ledger independently killed its sole mutation on the same named cell, with the predicted 16/18 progress marks.

The suite now drives the shipped `bin/smoke/shard.mjs`, not a restatement. Baseline evidence:

- fully green shard planned 4, started all 4, exited 0, and emitted no `NEVER RAN`
- break at suite 2 planned 4, exited 7, started exactly 2, emitted `NEVER RAN — 2 of 4`, and named only suites 3 and 4
- the suite asserted all 17 expected cells plus its count sentinel

The gate fragment is loaded by the production registry and reached exactly once by the production selector on shard 2 of 4. Registry checks reported 436 to 437, added 1, removed 0, 0 of 436 pre-existing suites moved, and `STABLE=0`. Gate inventory reported 437 suites with no duplicates.

Also green: build, typecheck, `check:docsbundle`, changeset status, diff check, `smoke:ci-fragments`, `smoke:ci-suites`, `smoke:ci-declarations`, and `smoke:shard-stability`. The mutation anchor is unique and code-only.

I did not run `smoke:ci`, `check`, any `:live`/`-live` suite, or stack lifecycle commands, per the review safety constraints. Immediately before this verdict, `refs/pull/1033/head` still resolved to the exact SHA above, current main `dca9462423fb96c9ac2eea3155fd26195ff5cc29` remained its merge base, and the detached review worktree was clean.

PR #1034, opened 2026-08-30
fix(ci): guard exact smoke suite indices

### LINKED ISSUE #920: bin/smoke/ci-suites.txt has no mechanical guard against reindexing: a mid-file insert reassigned 68 percent of the gate on main
## What happened

`bin/smoke/ci-suites.txt` is the ordered list of suites the CI smoke gate runs. `bin/smoke/shard.mjs` partitions it by INDEX:

```js
const mine = all.filter((_, i) => i % count === shard);
```

so a suite's runner is determined by its position in the parsed list. An entry added at the END changes nothing for anything already there. An entry added in the MIDDLE shifts every later suite by one and reassigns it to a different runner.

On 2026-08-27 two suite names were added mid-file and merged to main:

```
8e212a6c   smoke:hook-load-emitter-race at index 124 of 386   mid-file
7837b64c   same, index 124 of 387, merged to main
d1aeafc3   index 385 of 387                                   repaired
```

The repair commit's subject says it plainly: `test(ci): append the two new suites at the end, not mid-file`.

Measured consequence, comparing shard assignment by suite NAME between `ccd99461` and each revision:

| comparison | suites in common | same shard | moved shard |
|---|---|---|---|
| `ccd99461` to `7837b64c` (mid-file) | 385 | 124 | **261 (67.8%)** |
| `ccd99461` to `d1aeafc3` (repaired) | 385 | 385 | 0 |
| `ccd99461` to itself (control) | 385 | 385 | 0 |

Two suite names in the middle of a 387-line file reassigned two thirds of the gate to different runners, for about twenty minutes, on main.

## Why it matters more than it looks

`shard.mjs` stops at the first failing suite:

```js
if (r.status !== 0) {
  console.error(`\n✗ shard ${shard}/${count} FAILED at: ${cmd} (exit ${r.status})`);
  failure = r.status || 1;
  break;
}
```

So which suites a shard reaches depends on its composition. Reassigning two thirds of the suites changes what runs before the first red, which changes which suites execute at all. #912 measures 88 of 385 gated suites never executing at a given head; that set is a function of shard membership, and shard membership is a function of position in this file.

There is a second consequence that produces silently wrong analysis rather than a wrong verdict: **a shard identity is not stable across revisions that touch this file.** Two jobs both labelled "shard 1", at two revisions either side of a mid-file edit, are not running the same set of suites. Comparing them (did shard 1 improve? did this suite regress?) compares two different populations while every label matches. Nothing in the job output says so.

That is not hypothetical either. While investigating this, a shard-1 result at `7837b64c` was compared against one at `ccd99461` and read as a large improvement. The improvement was not measurable: the two runs had different memberships. The discriminator is one command:

```
git merge-base --is-ancestor <repair-sha> <run-head-sha>
```

## What caught it

A person reading a diff. Nothing mechanical noticed, at any point:

- CI did not fail. The gate ran; it just ran a different partition.
- No check compares shard membership across revisions, because nothing records it.
- `smoke:ci-suites`, the suite that parses and validates this file, **did not run**. It sits at index 252, which is position 64 in shard 0, behind a wall at position 59 (see #912). The check best placed to enforce a rule about this file is one of the 88 suites that never execute.

That last point is the reason this needs a mechanical guard rather than a convention. The convention was in force during the incident: the file was under an explicit hand-managed freeze at the time, observed by every branch working on it, and the mid-file insert arrived through an ordinary merge to main that the freeze did not cover. **A rule that binds only the people who know about it is not enforcement.**

## The property, and the construction that satisfies it

State the property first, because "append-only" is a construction and implementing the construction literally will miss cases:

**Property: no existing suite name may change its index.**

Append-only satisfies it. So does doing nothing. These do NOT satisfy it, and none of them is an insert:

- **deleting** a line: every later suite shifts up by one.
- **reordering** two lines: both change index, and so does nothing else, so a count-based check sees no difference at all.
- **replacing** a name in place: the old name's index vanishes and a new one appears at that position.

A check that only forbids mid-file additions passes a deletion, which is the same defect from the other end. The comment and blank-line handling matters too: `bin/smoke/ci-suites.mjs` drops comments and blank lines before indexing, so a comment inserted anywhere is harmless and must not be flagged, while a suite name inserted anywhere before the tail must be. The index is over parsed names, not over file lines.

## Suggested shape, not prescribed

A check that, for the merge base and the head, parses both files through `ci-suites.mjs` and asserts that every name present in both holds the same index. Report the moved names, not just a count: a count can be zero while two names swap.

Natural homes, with the tradeoff stated:

1. **In `smoke:ci-suites`.** Correct place conceptually, and it already owns validation of this file. But it is currently one of the suites that does not execute (#912), so until that is fixed the guard would be as unenforced as the convention it replaces. That is an argument for fixing #912 first, not against putting it here.
2. **As a CI step outside the sharded suite**, so it cannot be hidden behind a wall in the thing it is protecting. Less elegant, immune to the failure mode above.
3. **Both**, with the suite as the developer-facing check and the step as the gate.

Option 2 alone is the smallest change that actually holds today.

## Related

- #912: 88 of 385 gated suites never execute, because a shard stops at its first red. The enforcement point for this rule is one of them.
- #914: the Windows lane reports `success` while executing a fraction of the suite.

The three are one story about the same file and the same runner: the gate does not run everything it claims to run, does not report accurately on what it did run, and does not protect the file that determines both.

davidfarah2003:
Closes #920.

## Existing coverage

Current main already has an unsharded required CI step, `pnpm check:shard-stability <base> <head>`, that detects ordinary inserts and reorders when a pre-existing suite changes its derived shard. It also validates shard topology and committed runtime inputs. This PR does not rebuild that gate.

The remaining gap was narrower: the checker compared only `index % shardCount`, not the exact parsed index required by #920. A suite moved by four positions in the four-shard topology kept the same runner and passed.

## What changed

- Compare every suite common to base and head by exact parsed index, while retaining the existing shard-topology comparison.
- Report both the number reindexed and the exact number of common suites examined.
- Refuse removals even when no surviving suite changes shard.
- Add a broker-free property smoke covering a deterministic mid-file four-entry insert, a true tail append, and a deletion.
- Add mutation proof for the exact-index comparison and the examined-count guard.
- Append the new smoke at the true registry tail and add an empty changeset because this is CI-only.

## Reproduction

Before editing, a live 20-suite fixture rotated the list by four positions. All 20 exact indices changed, every suite kept its four-shard runner, and current main returned `STABLE=0` with `pre-existing suites CHANGING SHARD: 0 of 20`.

With this change, the same committed fixture returns `RESHARD=1`, reporting `pre-existing suites CHANGING INDEX: 20 of 20 examined` and still showing zero shard changes.

## Verification

- `pnpm smoke:shard-stability`: 7 passed, 0 failed.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/shard-stability-index.json`: 2 of 2 killed red-and-named on the exact property cells.
- `pnpm smoke:mutation-fixtures`: green, including both unique code-only anchors.
- `pnpm check:shard-stability origin/main HEAD`: `STABLE=0`; 420 of 420 pre-existing suites examined, zero changed index or shard, one suite appended.
- `pnpm typecheck`: green, including the build and bin smoke typecheck.
- `git diff --check`: green.

## Limits

This preserves positional round-robin sharding and enforces its exact-index invariant. It does not adopt stable hash sharding or otherwise remove order as a source of assignment. That CI-wide mechanism change remains outside this issue's guard-only scope.

### DIFF
--- a/.changeset/honest-shards-hold.md
+++ b/.changeset/honest-shards-hold.md
@@ -1,0 +1,4 @@
+---
+---
+
+CI-only: the shard-stability gate now enforces exact parsed suite indices and grades the invariant with a mutation-proven property smoke. No shipped package behavior changes.
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -690,14 +690,14 @@ smoke:agent-secret-segmentation
 # spellings, one of them the release workflow. The property is only observable on main after a
 # merge, so it needs a static guard. Appended so every existing shard assignment remains unchanged.
 smoke:workflow-concurrency
-# The shard-stability gate must compare exact parsed indices, not only `index % shardCount`: moving
-# a suite by one complete shard count preserves its runner while violating the append-only registry
-# invariant. The property smoke inserts four names mid-file, grades every common suite, and asserts
-# the exact examined count. Appended so every existing shard assignment remains unchanged.
-smoke:shard-stability
-
 # Most smoke trees are still outside their package TypeScript projects. A type-only import from a
 # workspace package can therefore name something the package never exported and vanish when tsx runs
 # the suite. This compiler-API intersection checks every named workspace import against the package's
 # public exports. Appended so every existing shard assignment remains unchanged.
 smoke:workspace-import-exports
+
+# The shard-stability gate must compare exact parsed indices, not only `index % shardCount`: moving
+# a suite by one complete shard count preserves its runner while violating the append-only registry
+# invariant. The property smoke inserts four names mid-file, grades every common suite, and asserts
+# the exact examined count. Appended so every existing shard assignment remains unchanged.
+smoke:shard-stability
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -690,3 +690,8 @@ smoke:agent-secret-segmentation
 # spellings, one of them the release workflow. The property is only observable on main after a
 # merge, so it needs a static guard. Appended so every existing shard assignment remains unchanged.
 smoke:workflow-concurrency
+# The shard-stability gate must compare exact parsed indices, not only `index % shardCount`: moving
+# a suite by one complete shard count preserves its runner while violating the append-only registry
+# invariant. The property smoke inserts four names mid-file, grades every common suite, and asserts
+# the exact examined count. Appended so every existing shard assignment remains unchanged.
+smoke:shard-stability
--- a/bin/smoke/mutations/shard-stability-index.json
+++ b/bin/smoke/mutations/shard-stability-index.json
@@ -1,0 +1,31 @@
+{
+  "suite": "bin/smoke/shard-stability.smoke.ts",
+  "guard": "the shard-stability gate compares exact parsed suite indices and examines every pre-existing suite, including moves by a whole shard count",
+  "command": "pnpm smoke:shard-stability",
+  "completionMarker": "SUITE COMPLETE",
+  "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/shard-stability-index.json",
+  "why": [
+    "The existing gate compared only index modulo shard count. Moving a suite by four positions in the four-shard topology therefore kept the same runner and passed, even though issue #920's append-only invariant says every pre-existing suite must keep its exact parsed index.",
+    "",
+    "The property suite inserts four entries at a deterministic mid-file position so the modulo-only mutant sees zero runner changes. The named property cell must redden, and a separate named cell asserts that every pre-existing suite was examined rather than accepting an empty loop."
+  ],
+  "mutations": [
+    {
+      "name": "compare only derived shard numbers, so whole-shard index moves disappear",
+      "file": "bin/smoke/shard-stability.mjs",
+      "find": "    if (headIndex !== baseIndex) changed.push(suite);",
+      "replace": "    if (headIndex % 4 !== baseIndex % 4) changed.push(suite);",
+      "expectRed": "FAIL: a mid-file whole-shard insert reports every pre-existing suite whose index changed",
+      "cell": "a mid-file whole-shard insert reports every pre-existing suite whose index changed"
+    },
+    {
+      "name": "stop counting examined suites, so an empty property loop can look green",
+      "file": "bin/smoke/shard-stability.mjs",
+      "find": "    examined++;",
+      "replace": "    examined += 0;",
+      "expectRed": "FAIL: the mid-file property examined every pre-existing suite",
+      "cell": "the mid-file property examined every pre-existing suite"
+    }
+  ],
+  "grades": "tool"
+}
--- a/bin/smoke/shard-stability.mjs
+++ b/bin/smoke/shard-stability.mjs
@@ -1,0 +1,13 @@
+/** Compare the parsed suite order, not only the derived shard number. */
+export function changedSuiteIndices(base, head) {
+  const headIndices = new Map(head.map((suite, index) => [suite, index]));
+  const changed = [];
+  let examined = 0;
+  base.forEach((suite, baseIndex) => {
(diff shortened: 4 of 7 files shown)


davidfarah2003:
Exact-head follow-up: `pnpm typecheck` completed green at `4a37165279183b066594a419aad51e15e399dc3d`, including the repo build and bin smoke typecheck.

davidfarah2003:
Delivery note: exact head is `4a37165279183b066594a419aad51e15e399dc3d`. Existing main already covered derived shard changes, topology, and committed inputs. This PR closes only the reproduced exact-index gap: a 20-suite rotate by four positions kept every four-shard runner and passed main as stable, while changing all 20 parsed indices. The new guard reports 20 of 20 examined and refuses it. Positional sharding remains unchanged by scope.

davidfarah2003:
Remote acceptance boundary at 02:06Z: the branch is mergeable and exact-head local acceptance is green, but GitHub has the CI and Windows workflows queued behind shared runner load. Sixteen recent workflow runs are queued or active across main and other PRs. No PR-owned remote failure or independent review finding is present.

davidfarah2003:
Queue progress note: CodeQL JavaScript/TypeScript is green. The remaining CodeQL Python, CI, and Windows jobs are still queued rather than failed. PR #1025, the preceding shard-registry lane, has now reached terminal all-green CI.

davidfarah2003:
Fleet delivery summary for coordinator visibility: PR #1034 exact tip 4a37165279183b066594a419aad51e15e399dc3d closes the exact-index false negative left by the existing derived-shard gate. Local repro and all local acceptance evidence are recorded above. Remote jobs are queued under shared runner pressure, not failing. Do not merge pending independent exact-head review.

davidfarah2003:
Additional public-interface acceptance at exact PR head `4a37165279183b066594a419aad51e15e399dc3d`: I created a committed successor ref that moved eight real `ci-suites.txt` entries forward by exactly four positions, so all eight exact indices changed while every four-shard assignment stayed the same. Running the same interface CI invokes, `pnpm check:shard-stability <base> <head>`, exited 1 and reported `pre-existing suites CHANGING INDEX: 8 of 421 examined`, `CHANGING SHARD: 0 of 421`, and `shard-stability: RESHARD=1`. This closes the integration-path gap beyond the helper/property fixture.

davidfarah2003:
Public acceptance detail: the committed successor ref used the actual 421-entry registry from this PR, not a copied parser fixture. Eight existing suite names were rotated by exactly four parsed positions. The production command that the required workflow calls detected the reindex and exited `RESHARD=1` even though derived shard membership stayed unchanged.

davidfarah2003:
Branch update at exact head `c2ae96f44ba01d9fdd5a5ef178251a46723fdb94`: integrated current `origin/main` (`a6891468d37c65cdeaad535da1dcfe606c5a3b3f`) after GitHub marked the PR dirty, then moved `smoke:shard-stability` back to the true tail. `pnpm check:shard-stability origin/main HEAD` is STABLE with 421/421 pre-existing suites unchanged. The property smoke, 2/2 mutation proof, mutation-fixture gate, gate inventory, diff check, and changeset status all remain green. No merge performed.

davidfarah2003:
Closing as superseded: main already carries this whole feature — `bin/smoke/shard-stability.smoke.ts`, `bin/smoke/shard-stability.mjs`, the `smoke:shard-stability` script, and a mutation fixture at `bin/smoke/mutations/shard-stability-index.json` whose load-bearing mutation is byte-identical to this branch's (same find/replace on the exact-index comparison). The refold hit that fixture as a create/create conflict, which is the semantic-twin signal we do not hand-merge. If any residual delta here (the shard-stability.ts changes or extra ci-suites lines) is not covered by main's landed version, it should be re-proposed as a small diff on top of current main rather than folded through the twin.

PR #1035, opened 2026-08-30
fix(jcode): retry transient bridge replacement

### LINKED ISSUE #971: jcode seat still exits on a provider disconnect when its single bridge recovery loses a race, so #781's guarantee holds only under light load
`#781` closed on the guarantee that a provider stall must not take down the mesh seat, and
`smoke:jcode-provider-disconnect` asserts it. On 2026-08-29 that assertion failed in CI on an
unrelated PR. The seat exited, so the guarantee did not hold.

It is not a regression from the PR it failed on: the same head passed on rerun, passed 11 out of 11
locally in five sequential repeats and in shard order, and the failing path's blobs are byte-identical
to a green `main`. What the log shows instead is that the guarantee is **probabilistic under load**.

## The observed failure

`smoke (shard 2/4)` attempt 1, run 33221141799, job 99016214607. The suite ran for three seconds:

```
===== pnpm smoke:jcode-provider-disconnect =====
  ✓ Jcode host joins before the provider stall
AssertionError [ERR_ASSERTION]: provider disconnect does not exit the mesh seat (#781) — {"code":1,
"stderr":"...
[cotal-jcode] turn failed (1 in a row): Jcode Harness connection closed during the turn; leaving the
              inbox batch unacknowledged
[cotal-jcode] private Harness connection closed and its one recovery attempt failed
"}
```

`actual: false, expected: true`. The seat exited 1.

## Where it comes from

`recoverBridge`, `extensions/connector-jcode/src/host.ts:447`. Its contract is deliberate and its
comment states the reasoning:

> Give that owned instance one clean replacement: the failed turn remains unacked, we attach the
> exact session rather than creating a blank one, and only then let the normal inbox driver see the
> pending work again. A second close is terminal — retry pacing belongs to the turn policy, not an
> unbounded connector relaunch loop.

That principle is right and this issue is not asking for an unbounded relaunch loop. The narrower
problem is what the single attempt contains:

```ts
try {
  await stopPrivateJcode();
  const replacement = await launchPrivateJcode();
  if (!sessionId) throw new Error(...);
  await replacement.attachSession(sessionId);
  ...
} catch {
  process.stderr.write("[cotal-jcode] private Harness connection closed and its one recovery attempt failed\n");
  await shutdown(1);
}
```

Stopping a daemon tree, launching a fresh one, and reattaching a session is a multi-second operation
whose latency depends on how loaded the machine is. One `catch` wraps all three. So the code
distinguishes **first close** from **second close**, but does not distinguish **the replacement
attempt failed transiently** from **the provider is genuinely gone** — and only the second of those
justifies killing the seat.

The result is that `#781`'s guarantee holds whenever the replacement wins its race and fails when it
does not. On a developer machine it wins every time, which is why this reproduces zero times locally.

## Why this matters beyond the flake

The seat exiting is the exact user-visible outcome `#781` was closed to prevent: a provider hiccup
takes down an agent that had a live mesh identity and an unacknowledged inbox batch. A CI flake is
the cheap version of that; the expensive version is a real seat dying on a loaded box.

It also costs a shard: a red here stops shard 2 at its first failure, and the suites behind it do not
run (see #912).

## What would resolve it

Distinguishing the two failure modes is the substance; the shape is open. Options, roughly in order
of how much they change the stated design:

1. Keep one recovery *of the bridge* but make the attempt itself resilient — treat a failure to launch
   or attach as not-yet-terminal and bound it by time rather than by attempt count, so a slow box gets
   the same outcome as a fast one.
2. Separate the exit decision from the attempt outcome: a second **close** stays terminal, while a
   failed replacement reports degraded and lets the turn policy decide, which is where the comment
   already says retry pacing belongs.
3. Leave the behaviour and make the test assert what the code actually guarantees, which would mean
   conceding that `#781` is closed on a probabilistic guarantee. Listed for completeness; it is the
   weakest option because the flake is a true positive.

## Reproduction

Not reproduced on demand. It reproduced once in CI, in the log above, and the mechanism is legible
from the source. A deterministic repro would inject a failure into `launchPrivateJcode` or
`attachSession` during the single recovery and assert the seat survives — which is also the test this
would need in order to prove any fix.

Refs #781, #912.

davidfarah2003:
Closes #971.

## Summary
- keep the single Jcode bridge recovery bounded by a 60 second window instead of one launch and attach outcome
- retry transient private bridge launch or session attach failures only after the failed replacement tree is proven stopped
- fail loud on unprovable replacement teardown before another launch can overwrite its lifecycle identity
- keep a second provider disconnect terminal, so recovery does not become an unbounded relaunch loop
- extend the real broker provider-disconnect smoke with deterministic transient-retry and ownership-refusal cells
- merge current main's fragmented CI registry without losing either parent's suites or moving an existing suite between shards
- update connector docs, generated docs bundle, mutation coverage, and changeset

## Validation
- reproduced the current-main merge conflict in `bin/smoke/ci-suites.txt`
- resolved to main's frozen 431-suite legacy registry plus five independent fragments; all 431 suites from each parent remain, with 436 unique suites and no duplicates
- focused provider-disconnect smoke: 9/9 checks through the shipped host, SDK client, Unix sockets, isolated broker, endpoint, and process-per-bridge fake
- provider-disconnect mutation proof: 6/6 killed red-and-named, including retry without proven teardown and swallowed terminal teardown refusal
- Jcode lifecycle smoke: 18/18 checks
- CI fragments smoke: 8/8 checks
- shard-index smoke: 7/7 checks
- shard stability from `ae59d04d362bd76c5f5484b4f0673620a2db04b4` to `a993edad77843f50f40bfd564ffcabe4724e5d5e`: 0 of 431 legacy indices and 0 of 436 pre-existing shard assignments moved
- gate inventory: 494 scripts, 474 reached, 20 declared exemptions, 436-suite CI chain with no duplicates
- connector-jcode build and TypeScript check
- docs bundle regeneration and docs smoke: 5/5 sections
- changeset status and `git diff --check`

## What was NOT proven
- No external model provider was used. The real host, SDK client, Unix sockets, mesh endpoint, isolated broker, and process-per-bridge fake prove connector lifecycle behavior, not a provider-specific failure mode.
- Full repository build, full `pnpm typecheck`, and full `pnpm smoke:ci` were not run. Validation was scoped to the changed connector and the affected registry, lifecycle, documentation, and safety workflows.
- `pnpm check:docs-voice` remains red on the current-main `docs/identity-and-auth.md` lines 181 and 183. That file is byte-for-byte unchanged from current main in this PR, so this repair did not broaden into unrelated auth prose.
- Forbidden live-stack suites and an authenticated installed `jcode` session were not run on the live mesh host.


davidfarah2003:
Do not merge yet: review at head 9d8ad319 flagged a high-severity orphan-safety issue. A follow-up fixing it is in progress; this comment will be superseded when the new head lands with a re-review.

davidfarah2003:
Full finding from the independent exact-head review, superseding my earlier placeholder:

BLOCK at exact head 9d8ad319801ceb4db81be366c206089d45b7f886. Severity HIGH.

host.ts suppresses failed replacement teardown and retries. The private lifecycle can throw when ownership or teardown is unprovable, and the next launch overwrites the sole launch identity, so a failed bridge can remain orphaned.

Reproduced through a real entry point rather than hand-built inputs: a scratch repro observed 3 launches, with the failed replacement still alive after an active retry shutdown.

The existing mutations do not protect this direction. They guard the fix's benefit (a transient replacement is retried) and not its safety (a retried replacement must not leave two bridges or orphan the first).

A follow-up branch is in progress; do not merge until it lands with a re-review.

davidfarah2003:
## Exact-head repair evidence

This supersedes the BLOCK recorded against `9d8ad319801ceb4db81be366c206089d45b7f886`.

**Review head:** `4b387984ca0d33202d952990614d588a8f32d300`

### Reproduced before the fix
- A real-entry harness at the blocked head forced replacement 2 to lose its launch identity after a deterministic attach failure.
- It observed three bridge launches, the teardown verifier's documented ownership refusal, the failed replacement alive beside the active retry, and the failed replacement still orphaned after graceful host shutdown.
- Exact harness PIDs were recorded and all were verified gone after cleanup.

### Fix
- Failed replacement cleanup is no longer suppressed. Recovery awaits `stopPrivateJcode()` and an unprovable teardown reaches terminal shutdown before another launch can overwrite the only lifecycle identity.
- Transient replacement failure still retries when exact teardown succeeds.

### Red-first and mutation evidence
- The new shipped real-host safety cell failed before the fix on `unproven failed-replacement teardown is terminal before another bridge launch`, observing launch count 3 and a live seat.
- Fixed head passes 9/9 provider-disconnect checks, with the safety scenario exiting 1 at exactly two launches.
- Mutation proof kills all 6 mutations red-and-named. M5 removes failed-tree teardown; M6 swallows terminal recovery refusal.

### Focused validation on this exact head
- `pnpm smoke:jcode-provider-disconnect`: 9/9
- provider-disconnect mutation proof: 6/6 killed
- `pnpm smoke:jcode-lifecycle`: 18/18
- scoped builds: core, connector-core, connector-jcode
- `pnpm check:shard-stability origin/main HEAD`: 0 of 424 pre-existing suites moved
- `pnpm smoke:gate-inventory`: green
- real detached-worktree merge test against `origin/main` succeeded; the PR now contains current main
- `git diff --check`; clean worktree

### What was NOT proven
- No external model provider was used. This proves connector lifecycle behavior through the real host, SDK client, Unix sockets, mesh endpoint, isolated broker, and process-per-bridge fake.
- Full repository build and full repository typecheck were not rerun.
- `pnpm check:docsbundle` is blocked by a docs-voice em dash already present on current `origin/main` at `docs/watch-a-mesh.md:148`; the canonical docs bundle was regenerated and committed on this head.

davidfarah2003:
## Independent exact-head verdict: BLOCK (MEDIUM)

**Reviewed head:** `4b387984ca0d33202d952990614d588a8f32d300`  
**Current main:** `3d7cd44dd8516aa9256989b24cccc5dc32f386f8`

### Blocking finding

The exact head is no longer mergeable into current main. GitHub reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and a non-mutating `git merge-tree` reproduces the conflict in `extensions/connector-core/src/docs-bundle.generated.ts`. Current main added newer generated tool/docs content while this PR updates the bundled Jcode guide. Incorporate current main and regenerate the bundle, then obtain a new exact-head review. The behavioral repair itself is approve-quality on this SHA.

### Independent behavioral evidence

- `pnpm smoke:jcode-provider-disconnect`: **9/9 passed** through the real host, SDK client, Unix sockets, isolated NATS broker, mesh delivery, and process-per-bridge fake.
- The fail-loud cell exited the seat with code 1 at exactly **2 launches**, named `does not carry its launch-bound identity - refusing unsafe teardown`, and did not launch replacement 3. This proves the lifecycle identity was not overwritten before cleanup was proven.
- I temporarily removed the failed-replacement `stopPrivateJcode()` call and reran the real entry point. It failed on the named safety assertion with `code: null`, **3 launches**, and three bridge PIDs. That reproduces the prior unsafe direction rather than only inspecting it.
- I added a temporary independent PID assertion to the green exact-head smoke. It proved replacement 2 was dead before replacement 3 began serving. The augmented run passed **10/10**, then the test file was restored.
- Provider-disconnect mutation proof: **6/6 killed red-and-named**. M5 failed on `unproven failed-replacement teardown is terminal before another bridge launch`; M6 failed on `terminal ownership refusal or unsafe third launch`.
- `pnpm smoke:jcode-lifecycle`: **18/18 passed**.
- Scoped builds for core, connector-core, and connector-jcode passed. Connector-jcode typecheck passed.
- Shard stability reported `STABLE=0`; gate inventory passed. The provider-disconnect smoke remains registered in the CI chain.
- The generated docs bundle is current relative to this exact head. The connector guide and patch changeset accurately describe the bounded recovery and fail-loud ownership rule.

### What was NOT proven

- No external model provider was used. The provider failure and attach race were synthetic, while the connector, broker, delivery, sockets, and processes were real.
- Full repository build and full repository typecheck were not run.
- Exact-head GitHub CodeQL/docs checks are still pending.
- `check:docs-voice` is red on `docs/watch-a-mesh.md:148`; the same em dash is present on current main and is outside this PR.
- A strict zero-process-leak guarantee cannot be claimed when ownership itself is unprovable. The safety control intentionally observes that refused replacement still alive after the host exits, until exact-PID harness cleanup. What is proven is that no additional bridge launches, the identity is not overwritten, and the unsafe condition fails loud instead of being hidden.

davidfarah2003:
## Final current-main fold

The prior exact-head evidence remains behaviorally valid and is now folded through current main.

**Terminal review head:** `12baf2f8bd7d3539e77f1cbf2a6780690d4b515a`
**Contained main:** `588b30b3e90107b6d3a3185093a510198512ba78`

- Repeated real detached-worktree merge tests before each fold. The generated docs conflict was resolved only by rebuilding the required packages and regenerating tool docs plus `docs-bundle.generated.ts`.
- Preserved the true-tail CI suite ordering from current main. `check:shard-stability` reports 428 → 429 suites, one tail addition, and **0 of 428** pre-existing suites moved.
- `smoke:gate-inventory` is green with 430 CI suites and current exemption freshness checks.
- Final focused repair proof remains green: provider-disconnect smoke 9/9, mutation proof 6/6 killed red-and-named, Jcode lifecycle 18/18, scoped core/connector-core/connector-jcode builds, and `check:docsbundle` green.
- PR/local/remote all equal this SHA, and current main is an ancestor.

### Named limits
- No external model provider was used.
- No full repository build or full repository typecheck was rerun. Validation is focused on the changed connector, its real lifecycle workflows, docs generation, and CI inventory/order gates.

davidfarah2003:
Posting on behalf of reviewer seat `w983`, whose own write attempt returned GitHub's shared-identity rate limit. Verdict and evidence are its work, verbatim and unaltered.

---

BLOCK `aff58348b09a386a5a7cdabc63a795cad6bd212b`

`recoverBridge()` does not classify replacement failures as transient or permanent. Its retry catch handles every rejection from `launchPrivateJcode()` and `attachSession()` identically until the 60-second deadline. The SDK preserves actionable permanent codes including `invalid_request`, `handshake_failed`, `jcode_not_found`, and `EACCES` through `connect_failed`, so configuration, protocol, permission, and missing-session refusals are retried when they must terminate immediately.

I reproduced this through the shipped host/provider smoke entry point, not from a constructed unit input. I changed the fake replacement attach from one transient `unavailable` reply to a persistent `invalid_request` reply. The real host survived and continued relaunching replacements instead of refusing the permanent failure.

The current green fixture set cannot catch this direction. It covers one transient `unavailable`, the bounded recovery window, terminal behavior after a second disconnect, and refusal to launch again after unproven teardown. It never constructs a permanent launch or attach refusal. The six green mutations therefore prove only those represented cells discriminate.

Please classify permanent SDK failures as terminal and add a real host/provider cell that holds a permanent refusal across attempts. Under the current suite registry, that new gate should be a hash-named fragment in `bin/smoke/ci-suites.d/`, not an append to `bin/smoke/ci-suites.txt`.

Other exact-head validation passed: provider disconnect 9/9, lifecycle 18/18, all six existing mutations killed red-and-named, gate inventory, mutation fixtures, shard stability, docs bundle, connector builds/typecheck, diff check, and changeset status. The synchronization now waits for `turn_done_emitted` rather than sleeping.

---

Gate note. This is the governing verdict for `aff58348`. Other objects on this PR worded as "APPROVE-quality" are hedges, not terminal verdicts, and do not clear the head; a head is clear only when every assigned reviewer clears it. Nothing in those passes contradicts the finding above — the fixture set simply never constructs a permanent refusal, so no run any reviewer performed could have gone red on it.

`main` has since advanced to `ae59d04d`, so this PR needs a fold. Since #1052 landed, `bin/smoke/ci-suites.txt` is frozen as the positional legacy registry and new gates are one file each under `bin/smoke/ci-suites.d/`, named `sha256(<full suite name>).txt` with shard assignment by hash — migrating this PR's entry that way also stops it re-conflicting on every unrelated landing. When a repaired head is frozen it returns to `w983` for a bounded delta, since it holds the reproduction.

davidfarah2003:
BLOCK

Exact head reviewed: `d82cfcdbf32f5b6d2e8bb1dce4dd8e6cd22f9740`.

## Blocking finding

The two focused mutation fixtures are not accepted by the repository's mutation-coverage auditor. Running:

```bash
node scripts/mutation-coverage.mjs \
  extensions/connector-jcode/smoke/mutations/provider-permanent-refusal.json \
  extensions/connector-jcode/smoke/mutations/provider-disconnect.json
```

exits 1 immediately because `provider-permanent-refusal.json` mutation M1 is missing `cell`. Neither config declares `grades: "tool"`, and none of their seven mutations carries a `cell`. The direct mutation batteries kill, but that is not sufficient while the repository's separate fixture auditor refuses the metadata. Add explicit cell bindings matching the observed named reds, or classify the configs under the established instrument convention only if that classification is semantically correct, then rerun both auditor and batteries.

## Independently verified

- Exact head matched `refs/pull/1035/head` before and after review.
- A real `git merge --no-commit --no-ff` of the exact head into assigned current main `dca94624` completed cleanly in a separate detached worktree.
- Production permanent-refusal smoke passed 2/2: the shipped host joined, then one persistent `invalid_request` replacement refusal terminated before a third launch, with no retry message.
- Production recovery safety smoke passed 9/9: a real delivered turn closed the Unix-socket client, the first replacement attach failed transiently, retry stayed inside one bounded window, the same session reattached, the unacknowledged turn redrove, later work succeeded, a second disconnect was terminal, and unproven replacement teardown stopped before another launch while the owned process remained observable for exact cleanup.
- Permanent refusal classifier is static in production code and covers protocol/home/request/session/version/binary refusals plus EACCES connect denial. The temporary `if (false)` mutation is absent; exact production call remains `if (permanentBridgeRecoveryFailure(error)) throw error;`.
- All seven mutation anchors are present exactly once, code-only, and non-prose-spanning. Sequential direct mutation proof killed 1/1 permanent refusal mutation and 6/6 recovery/safety mutations red-and-named, matching the predicted cells.
- Full repository build and connector-jcode typecheck/build passed.
- Permanent-refusal fragment filename independently matched SHA-256 of the exact suite string, and production registry helpers enumerated/reached it exactly once on shard 3.
- Direct docs-bundle regeneration was byte-clean; changeset status and `git diff --check` passed.

## Named gaps and unrelated red

`smoke:jcode-security` could not execute its root-in-Docker foreign-owner arm because Docker is not installed on this box (`docker: command not found`); it failed before reaching the remaining cells. This is an environment gap, not evidence against this PR, and the changed recovery path is covered by the two focused production-host smokes above.

No stack lifecycle command, `smoke:ci`, `pnpm check`, or live suite was run. Load was 32.58 at intake, 52.54 after mutation batteries, and 38.02 at final checks; focused assertions are not timing-sensitive.

davidfarah2003:
BLOCK `d82cfcdbf32f5b6d2e8bb1dce4dd8e6cd22f9740`

The original `invalid_request` defect is repaired through the shipped host path. The new persistent-refusal cell terminates after one replacement with no retry or third launch, and reverting `permanentBridgeRecoveryFailure()` kills that cell red-and-named. The transient direction remains intact at 9/9. The existing six-mutation ledger also kills all six red-and-named when run sequentially, including M6 on `terminal ownership refusal or unsafe third launch`.

The permanent classification contract is not adequately locked down, however. Production classifies seven direct SDK codes plus `connect_failed` caused by `EACCES`, while the new cell constructs only `invalid_request`; it does not assert the permanent-set count or exercise the other advertised categories. I removed `handshake_failed`, `jcode_not_found`, and the `EACCES` arm together, then reran both shipped-host suites. Permanent refusal still passed 2/2 and transient recovery still passed 9/9. Those advertised permanent refusals can therefore regress back to the 60-second retry loop without any new or existing cell going red.

There is also no stated or asserted policy for an unrecognized `HarnessError` code. The current control-flow default is retry-until-window, but that is an implicit fall-through rather than a deliberate documented and tested choice.

Please enumerate and assert the permanent classifier's exact count, cover the named permanent categories including `handshake_failed`, `jcode_not_found`, and `connect_failed`/`EACCES` through the shipped path or a direct classifier seam reached by it, and state/test the unknown-code default.

Other bounded validation passed: dedicated mutation 1/1 killed, provider disconnect 9/9, lifecycle 18/18, exact anchors for the two Jcode ledgers, connector build/typecheck, and diff/changeset checks. Mergeability was re-proved against current main `b6c1428b75f5285ccc57d074f529d6dd001d7ae9`; both permanent 2/2 and transient 9/9 passed in that merge tree.

davidfarah2003:
BLOCK `39cf10c8b1f40478bc84a11a8216cd155eee64bd`

The corrected head fixes the transient teardown and M6 assertion isolation, but it does not add the claimed full permanent-classification contract proof. Production still has seven direct permanent codes plus `connect_failed`/`EACCES`. The permanent-refusal shipped-host smoke still constructs only `invalid_request`; there is no exact set-count assertion, no cell for the other six direct codes or EACCES, and no stated/tested unknown-code retry default. The permanent mutation ledger remains the single mutation that disables the classifier as a whole.

I reran the adversary from my prior BLOCK on this exact head: removed `handshake_failed`, `jcode_not_found`, and the `EACCES` arm together, then ran both shipped-host suites. Permanent refusal still passed 2/2 and transient recovery passed 10/10. Those advertised permanent categories can still regress silently, so the requested acceptance criterion remains unmet.

The new bounded changes themselves behave correctly: transient recovery now passes 10/10 with proof the failed replacement tree is stopped before retry; M6 is isolated onto the permanent-refusal cell and names the swallowed terminal shutdown. The exact head already contains authoritative main `43e9dd303f71528ece5e13b169838b29cfff2de5`, so the merge is clean/up-to-date.

Please add an exact permanent-set count, exercise or directly classify all seven direct codes and `connect_failed`/`EACCES` with mutation sensitivity, and explicitly state/test that an unknown HarnessError code remains transient and consumes the bounded retry window.

davidfarah2003:
APPROVE `a6255bc5e9fad00e2565d480b5affc36b8be60b5`

This head closes the permanent-classification coverage gap from my prior BLOCKs.

The classifier contract now asserts exactly seven direct SDK codes plus the sole `connect_failed` cause `EACCES`, and asserts that an unknown HarnessError code remains transient and consumes the existing bounded retry policy. The new source-reached classifier suite imports the production sets and classifier rather than restating them.

Most importantly, M1 is the exact adversary that previously survived: it removes `handshake_failed` and `jcode_not_found` and disables `EACCES` classification together. It now kills red-and-named on the exact contract cell. The two additional classifier mutations independently kill EACCES removal and unknown-as-terminal.

Validation performed:

- `smoke:jcode-recovery-classifier`: 2/2; all 3 classifier mutations killed red-and-named.
- `smoke:jcode-provider-permanent-refusal`: 6/6; all 4 permanent/shutdown mutations killed red-and-named when run sequentially.
- `smoke:jcode-provider-disconnect`: 10/10; all 6 transient/teardown mutations killed red-and-named, including M6 on the isolated permanent cell.
- `smoke:jcode-lifecycle`: 18/18.
- Mutation anchors for the three Jcode ledgers are present and unique. Connector build/typecheck and diff check passed.
- A real detached merge with authoritative main `786e40e00ee2bc19ffde5ddcd4694778871635a6` was clean. On that merge tree the classifier 2/2, permanent 6/6, and transient 10/10 all passed.

The shipped persistent `invalid_request` entry-point cell still proves terminal behavior after one replacement, while the transient entry-point cell proves recoverable failures remain retried and that each failed replacement tree is stopped first.

PR #1037, opened 2026-08-30
fix(ci): expire ungated suite exemptions

### LINKED ISSUE #924: gate-inventory validates that an exemption has a reason, not that the reason is still true
## Summary

`bin/smoke/gate-inventory.smoke.ts` refuses an ungated suite unless it appears in the `UNGATED` map
with a reason, and it checks that the reason is present and long enough. It cannot check whether the
reason is still accurate. An exemption whose stated condition has since been resolved keeps passing,
silently, for as long as it stays in the file.

The failure is not a missing guard. Nothing is broken and nobody wrote a bad check. A correct
exemption goes stale, and nothing in the system observes that its premise expired.

## Worked example

While preparing a change that adds a smoke suite, the chain file `bin/smoke/ci-suites.txt` was under
a positional freeze: inserting a line mid-file re-shards every later suite, because the shard walk is
round-robin over the parsed index. The correct move at the time was to leave the chain file untouched
and declare the new suite exempt:

```
"smoke:ps-provenance": "ci-suites.txt is frozen by position until PR #880; gate when the freeze lifts",
```

That entry was accurate when written. It was made obsolete by commit `d1aeafc3`
("test(ci): append the two new suites at the end, not mid-file", read 2026-08-27), which both wires
new suites into the chain and establishes appending at the tail as the safe way to do it: a tail
append leaves every existing index unchanged, so the freeze it named no longer applies.

Nothing failed. `gate-inventory` kept reporting OK, because the entry still had a reason and the
reason was still long enough. The suite in question, which existed to enforce a labelling guarantee,
was never in the enforcing set at all; every green anyone saw came from invoking it by hand. Had that
gone unnoticed, a later refactor removing the guarded code would have shipped with CI green.

## Why it generalises

The map currently holds 32 exemptions on `main`. Most describe a permanent environmental property
("needs a pi install", "drives the public orca CLI"), and those cannot go stale: the condition is a
fact about the world, not an event that resolves.

The problem is the other shape, an exemption naming a condition that is expected to end. One such
entry is live today and is still correct:

```
"smoke:renewal-terminal-race": `${BROKEN} reproduction of an open defect; gate when the fix lands`,
```

That is exactly the right thing to write, and it is also a claim with an expiry date that nothing
watches. On the day the fix lands, the entry becomes false and the gate goes on certifying it. The
`BROKEN` prefix already marks this class as debt with a fuse, which is the correct instinct; the fuse
just has no mechanism behind it.

So any entry of the form "blocked until X" becomes a false statement the gate certifies indefinitely
from the moment X resolves. The longer it survives, the more it reads like a settled decision rather
than an expired one.

## Suggested

Narrow version, preferred: an exemption whose reason names a condition should name it in a form the
suite can test. A referenced pull request or issue number can be checked for closure; a referenced
path or script name can be checked for existence. The reason stops being prose that a human has to
re-audit and becomes an assertion that fails loudly when it stops holding.

Cheaper fallback, if machine-checking is more than this is worth: require a `recheck by <date>` field
on any entry in the fused class, and fail the gate when that date passes. This does not detect the
resolution, but it converts silent staleness into a dated prompt, which is enough to make the entry
get looked at.

Either way the property worth having is the one the suite already applies to reason text: a reason
that cannot expire loudly is a reason that will eventually be wrong without anyone noticing. The
existing check asks whether an explanation was written. The missing check asks whether it is still
true.

davidfarah2003:
Closes #924.

## Classification first

Current main has 20 `UNGATED` exemptions. I classified all 20 before choosing the mechanism:

- 7 name external tools, authentication, session files, or environment input.
- 2 are full-stack or packaging policy decisions.
- 11 are `UNTRIAGED` and carry no decision beyond acknowledging debt.

None has a premise this repository can safely validate as a repository fact. Local command presence cannot prove CI has the tool, and a developer login or session cannot prove a runner's state. A predicate validator here would report confident green about the wrong environment.

## What changed

- Every exemption now carries a structured `reason` and `recheckBy` date.
- The expected exemption count is pinned to the literal `20`, not derived from the enumeration under test.
- Invalid or expired dates fail the existing required gate and name the affected exemptions.
- A permanent past-due control proves expiry is active on every run.
- The 11 `UNTRIAGED` entries are reviewed first on 2026-09-30. The 9 standing environment/policy decisions follow on 2026-11-30, avoiding one batch rubber stamp.
- An empty changeset records that this is CI-only.

The mechanism forces re-review. It does not silently convert `UNTRIAGED` into triage, and this PR deliberately does not investigate those 11 suites.

## Reproduction

Before the change, `pnpm smoke:gate-inventory` reported `GATE INVENTORY OK` for all 20 exemptions while checking only that each reason was at least eight characters. No date, predicate, or freshness metadata was read.

## Verification

- `pnpm smoke:gate-inventory`: green; `exemption freshness examined: 20 of 20` and all dates current.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/gate-inventory-exemption-review.json`: 2 of 2 killed red-and-named. One disables expiry; one reports zero examined.
- `pnpm smoke:mutation-fixtures`: green, both anchors unique and code-only.
- `pnpm typecheck`: green, including build and bin smoke typecheck.
- `git diff --check`: green.

The full typecheck was initially deferred while the host was saturated. It was run after the fleet load gate was rescinded, starting at load average `21.95, 21.39, 15.81` on 12 cores, and completed green in 76.8 seconds.

## Limits

This verifies review currency, not the truth of the reason. External tool availability, authentication, operator session state, and policy judgements remain unverifiable by construction from this repository. The expiry makes those statements stop being certified indefinitely and forces another explicit decision.

### DIFF
--- a/.changeset/brave-exemptions-expire.md
+++ b/.changeset/brave-exemptions-expire.md
@@ -1,0 +1,4 @@
+---
+---
+
+CI-only: every ungated smoke-suite exemption now carries a machine-checked review date and expires loudly when its premise needs re-evaluation. No shipped package behavior changes.
--- a/bin/smoke/gate-inventory.smoke.ts
+++ b/bin/smoke/gate-inventory.smoke.ts
@@ -62,12 +62,16 @@ const ROOT = join(dirname(fileURLToPath(import.meta.url)), "..", "..");
  *  decision, which needs no fuse and is written as a plain reason. */
 const BROKEN = "BROKEN:";
 
-const UNGATED: Record<string, string> = {
+type UngatedExemption = { reason: string; recheckBy: string };
+const RECHECK = "2026-11-30";
+const exemption = (reason: string): UngatedExemption => ({ reason, recheckBy: RECHECK });
+
+const UNGATED: Record<string, UngatedExemption> = {
   // Need external tooling no CI runner has.
-  "smoke:orca:live": "drives the public orca CLI",
-  "smoke:orca-e2e:live": "drives the public orca CLI", "smoke:pi": "needs a pi install", "smoke:codex-live": "needs a logged-in codex CLI",
-  "smoke:codex-tui-live": "needs a codex TUI session",
-  "smoke:jcode-live": "needs an installed, authenticated jcode CLI (COTAL_E2E_JCODE=1)",
+  "smoke:orca:live": exemption("drives the public orca CLI"),
+  "smoke:orca-e2e:live": exemption("drives the public orca CLI"), "smoke:pi": exemption("needs a pi install"), "smoke:codex-live": exemption("needs a logged-in codex CLI"),
+  "smoke:codex-tui-live": exemption("needs a codex TUI session"),
+  "smoke:jcode-live": exemption("needs an installed, authenticated jcode CLI (COTAL_E2E_JCODE=1)"),
   // A STANDING DECISION, and only for the REAL-SESSION arm. The same suite is GATED as
   // `smoke:agui-map`, pointed at a fixture DERIVED from a real session by
   // `scripts/redact-claude-session.mjs` (whitelist by construction, identifiers pseudonymised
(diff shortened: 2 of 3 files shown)


davidfarah2003 (commented):

- Diff matches the body exactly: `UNGATED` is now `Record<string, {reason, recheckBy}>`, with 9 "standing" exemptions dated `2026-11-30` and 11 `UNTRIAGED` exemptions dated `2026-09-30`, totaling the pinned `EXPECTED_EXEMPTIONS = 20`.
- `reviewDate`/`reviewExemptions` validate real calendar dates and fail loud, naming every invalid/expired entry; a permanent past-due control fixture (`recheckBy: "2000-01-01"`) proves the expiry-fail branch actually fires.
- Mutation-proof (`bin/smoke/mutations/gate-inventory-exemption-review.json`) has 2 named mutations, each targeting the shipped `gate-inventory.smoke.ts` logic directly (disabling the expiry comparison, zeroing the examined count) with a named red cell.
- `bin/smoke/ci-suites.txt` not touched — rule (c) N/A. No silent fallback, no AI/tool attribution, empty CI-only changeset matches the claim.
- Nit (non-blocking): `expiryControl` hardcodes today as the literal `"2026-08-30"` instead of reusing the computed `today` var — harmless since the control's `recheckBy` is pinned to 2000-01-01, but worth tidying.

Issue #725, opened 2026-08-19
manager: the frozen-gate error names no remedy, though a purpose-built one exists


davidfarah2003:
When a manager start hits a frozen issuance gate it prints:

```
✗ the issuance gate for "<instance>" is frozen; another barrier holds it; re-read and re-decide (SPEC 13.8)
```

and exits. `cotal reconcile-gate` exists precisely for this state, verifies the freeze-holder is gone, completes the dead op, and refuses safely otherwise. The error does not mention it, and neither does any runbook; the operator path to discovering it is reading `cotal --help` end to end while the mesh is down.

This matters more than a normal message nit because of where the message lands: a keeper script restarting the manager on a schedule loops on this exact line every few minutes for as long as the wedge lasts, so the one place guaranteed to be read during the outage is the one place the remedy's name never appears. In a recent outage this loop ran for close to three hours while the repair was one command away.

"re-read and re-decide" describes what a LIVE holder should do with a barrier, not what the operator of a dead one can do. What would close this: when the gate is frozen and the holder is this same instance's dead predecessor (which the manager can see from its own persisted instance id), say so and name the verb: `another barrier holds it; if the holder is a dead predecessor, run: cotal reconcile-gate --space <s> --server <url>`.

davidfarah2003:
## Phase-1 triage: already fixed at `8c9efa7d640952f95c3b2e52b00be7b51ab52320`

**Verdict:** already-fixed. The reported frozen-gate error now names the purpose-built remedy. This issue is left **OPEN** for a human disposition. Already-fixed is not a close.

**Tested at:** `8c9efa7d640952f95c3b2e52b00be7b51ab52320` (wave-9b-20260901-700s, detached worktree)
**Confidence:** high
**Area / severity:** `area:manager` / `severity:medium`
**Terminal label:** `triage:not-reproduced` (the reported wording does not reproduce at the pin because the fix is an ancestor)

### Two line states

`packages/core/src/endpoint-service.ts`

- Reproducing parent `c4c06a8a34d8756efadab54bcab0d23da23bd6c4` line 417: `the issuance gate for "…" is frozen; another barrier holds it; re-read and re-decide (SPEC 13.8)` — no remedy named.
- Pin `8c9efa7d` line 421: `the issuance gate for "…" is frozen; another barrier holds it; if the holder is a dead predecessor, run: cotal reconcile-gate (SPEC 13.8)` with structured `remedy: "cotal reconcile-gate"`.

[...]

PR #1038, opened 2026-08-30
fix: return lifecycle state on spawn refusals

### LINKED ISSUE #873: spawn failures return no lifecycle state, so a knowable refusal reaches the operator as the connector's opaque message
Measured 2026-08-26 while debugging a wave of seat deaths on a local mesh.

Four consecutive spawns of one seat name failed. What the operator saw, every time:

    ✗ w864 exited on launch - last output: [cotal-jcode] fatal: Jcode host startup failed (unknown); inspect the private Jcode logs.

What the manager's own log carried in the same window:

    ✗ re-registration could not revoke + verify-evict the superseded serve family; the gate is
      left frozen for reconciliation, no new spec published (SPEC 13.1): principal "local.U…" is
      not verified evicted
    ✗ the issuance gate for "<instance>" is frozen; another barrier holds it; re-read and
      re-decide (SPEC 13.8)

The manager knew a structured, named thing. The caller got a connector's catch-all. Nothing carries lifecycle state across that boundary, so the operator's next move is a guess.

The cost was concrete: several attempts spent on the wrong layer — clearing stale sockets and pidfiles, then moving the seat's whole private home aside so a fresh one materialized — none of which could ever have helped, because nothing in the seat's own state was wrong. It also let a false theory survive: a rename appeared to fix it, and an opaque error gives a wrong hypothesis nothing to contradict it.

## What is missing

A spawn failure should return the lifecycle facts the manager already has at the point of refusal:

- the exact op kind that is blocked (registration, retirement, activation),
- the head's current state (`active` / `retiring` / `retired`) for the actor being spawned,
- the `opId` holding it,
- the remediation verb, when one exists.

## Why this is not #828 and not #872

- **#828** is message fidelity inside connector-jcode: it scrubs a refusal *it can already name* into `startup failed (unknown)`. Fixing it makes the connector's own errors honest, and would not have surfaced a manager-side lifecycle refusal, because that information never reaches the connector.
- **#872** is the lifecycle defect that was killing the seats. Fixing it removes that particular cause.
- This issue is the **propagation path between them**: even with both fixed, the next manager-side refusal will reach a caller as whatever generic string the connector happens to emit.

Related: #871 (the frozen-gate state this surfaced from), #725 (the frozen-gate error naming no remedy — the same missing-information problem one layer down).

davidfarah2003:
Closes #873.

A manager-side lifecycle refusal (frozen issuance gate, retiring alias, retired uid) already knew the blocked op, the head/gate state, the holding `opId`, and often the remedy. Those facts died at the ControlReply string boundary, so the spawn caller saw a SPEC 13.6 wait-timeout or the connector's opaque message and spent the next attempts on the wrong layer.

This is the propagation path between #828 (connector-jcode scrubbing a refusal it can already name) and #872 (the particular seat-death cause). It does not fix either of those.

## What now happens

- Core names the facts on the throw as `details[].kind = ai.cotal.ep.lifecycle-blocked` (`blockedOp`, `headState`, `opId`, `remedy` when one exists).
- Follow, ControlReply, the CLI, and `cotal_spawn` keep those facts. A string-only caller still sees `[lifecycle blockedOp=… headState=… opId=… remedy=…]`.
- A frozen manager issuance gate names `cotal reconcile-gate` instead of "re-read and re-decide".

## Proof

- `pnpm smoke:spawn-lifecycle-state` 14/14 (constructed follow + MeshAgent.managerInvoke + cotal_spawn; no broker).
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/spawn-lifecycle-state.json`: all 3 mutations KILLED, red-and-named.
  - follow dropping `details` off a failed terminal
  - managerInvoke rendering a bare `code: message`
  - cotal_spawn ignoring `ControlReply.details`
- Nearby regressions: `smoke:goal-follow` 14, `smoke:manager-invoke-verdict` 8, `smoke:ep-envelope` 73, `smoke:gate-inventory` OK.

## Not proved

- A live mesh spawn against a frozen production gate. Forbidden here (`cotal up` / `--supervise` / live manager suites). The constructed follow path is the branch; a live door is a named gap.
- Connector-jcode's own scrub (#828). This issue is the manager-to-caller path, not the jcode host banner.
- The frozen-gate cause itself (#871 / #725). This only names the remedy on the refusal.


davidfarah2003 (commented):

- `bin/smoke/ci-suites.txt:366`: new `smoke:spawn-lifecycle-state` entry is inserted mid-file, between `smoke:goal-follow` (line 363) and `smoke:required-arg-seam` (line 367). The file's actual tail (verified at this PR's head commit) is `smoke:workflow-concurrency` near line 695. This is not an append — every suite after line 366 gets renumbered/reshuffled to a different shard.
- Everything else checks out: `endpoint-error.ts` adds `EP_LIFECYCLE_BLOCKED`/`lifecycleBlocked`/`renderLifecycleBlocked`; all 3 propagation sites the body names (`submitAndFollowGoal` in `endpoint-invoke.ts`, `MeshAgent.managerInvoke` in `agent.ts`, `cotal_spawn` in `tool-specs.ts`) are wired and each has a named mutation in `spawn-lifecycle-state.json` that reverts real shipped code.
- Smoke suite drives the real `submitAndFollowGoal`, `MeshAgent`, and `cotalToolSpecs` (not reimplementations), with an internal `EXPECTED=14` check-count guard against silently-skipped assertions.
- Docs (`control-surface.md`, `mcp-tools.md`, `run-a-mesh.md`) and the generated tool-doc source are updated in lockstep — fix rule (c) and this is landable.

davidfarah2003:

davidfarah2003:
CI classification, final (run 33318085124):

| shard | attempt | signature | class |
|---|---|---|---|
| 0/4 | 1 and 2 | `smoke:mutation-fixtures` ANCHOR IS DEAD ×2 (manager-runtime-deps [0], gate-reconcile-resume [1]) | cured on main by #1161 (3893599d) — this run's merge ref predates the cure |
| 2/4 | 1 only | `jcode-provider-disconnect` ✗ `post-recovery turn` | flake — greened on attempt 2 against the identical merge ref; tracked cross-branch as #1169 |

Merging per the cured-on-main + rerun-discriminated rules.

Issue #903, opened 2026-08-26
provisionAgent role "board": the per-stream grant template omits $JS.API.STREAM.INFO wherever it emits the consumer-create deny triple (DM exercised, DLV/TASK/EPC latent)


davidfarah2003:
A board endpoint provisioned by `provisionAgent` with `role: "board"` is refused `STREAM.INFO` on the space's DM stream. It is a publish-plane authorization denial on the read path, so the dashboard fails into an empty view rather than an error, which is why the symptom presents as "renders nothing" rather than as a failure.

This is a **provisioning** defect, not rotation damage or credential expiry: it reproduces across every board connection in the window, including freshly established ones.

## Evidence

From a live broker's own stderr over a six-hour window (nkeys redacted; one representative line):

```
[ERR] 127.0.0.1:xxxxx - cid:676 - "v3.4.0:nats.js:cotal:foreman"
      - "<NKEY>/<space>/nkey:<NKEY>"
      - Publish Violation - Subject "$JS.API.STREAM.INFO.DM_<space>"
```

Aggregates over the whole log:

| measure | value |
|---|---|
| denials on `$JS.API.STREAM.INFO.DM_<space>` | 72 |
| distinct client ids for the board principal | 7 |
| window | 12:05:11 -> 18:03:11 local, still firing at ~2/sec when sampled |
| total `Publish Violation` lines in the log | 74 |
| total lines in the log | 170 |

**Controls, because a clean negative is worth nothing without them.** Non-violation `[ERR]` lines: **0**. That zero is real rather than a broken pattern — `[ERR]` matches 74 lines, all of which are the violations, so the broker is otherwise silent and this is the only thing it is complaining about. 74 of 170 lines is a large fraction rather than a truncation artifact. The 7 distinct cids are the board principal's alone (segmented by client name, not taken from the file-wide total), so this is every board connection in the window rather than one bad session.

## The decisive next check, which I could not run

Whether the board's credential grants this subject **on the publish plane**. Every `$JS.API.*` subject is a request the client *publishes*; a subscribe grant authorises none of them, so an audit that merges `pub.allow` and `sub.allow` will report this as granted when it is not. Decode the JWT payload from the board principal's credential and check `pub.allow` specifically.

I could not do this: the board peer in the observed deployment authenticates from a tree outside the one I can read, so its credential was not available to me. Recording it as the open step rather than guessing at the answer.

[...]

Issue #904, opened 2026-08-26
endpoint plane: an orchestrator principal is denied manager.inspect on its own subject (single-shot, so it never surfaces)


davidfarah2003:
Two publish denials on the endpoint-plane `manager.inspect` subject, one from each of two orchestrator-role principals, each against a subject carrying that principal's own wire id.

Filing this separately from #903 rather than folding it in: different role, different plane, and nothing here establishes a shared cause.

## Evidence

From the same live broker log as #903 (nkeys redacted):

```
[ERR] cid:681 - "v3.4.0:nats.js:cotal:<orchestrator-a>" - Publish Violation
      - Subject "cotal.<space>.ep.one.manager.inspect.local.<NKEY>.<lifecycle>.<token>"

[ERR] cid:682 - "v3.4.0:nats.js:cotal:<orchestrator-b>" - Publish Violation
      - Subject "cotal.<space>.ep.one.manager.inspect.local.<NKEY>.<lifecycle>.<token>"
```

Exactly **one occurrence each**, hours apart, in a 170-line log whose only other content is the 72 board denials of #903.

For one of the two principals I confirmed the denied subject carries that principal's own wire id, by matching the id embedded in the subject against the principal's known id. For the other I could not verify its own id to the same standard from where I was reading, so I am recording that one as observed rather than as characterized.

## Why this is worth a report despite being two lines

It does not spam. A capability that fails once, silently, on a path the caller assumed it had is invisible precisely because it does not repeat — there is no rate of failure to notice, no retry storm, and the caller most likely degraded into a path that looked like an ordinary empty result. It only turned up because the log was being read for something else.

The shape worth checking is whether an orchestrator/manager-role principal is granted publish on `cotal.<space>.ep.one.manager.inspect.<own-principal>.>` at all, or whether the grant is issued for a different actor form than the one the caller uses. #903 notes the same plane trap that applies here: every one of these is a **published** request, so a subscribe grant authorises none of them and an audit merging `pub.allow` with `sub.allow` will report it as granted.

No claim is made here about impact. I did not reproduce it, and a single historical denial is not enough to say what the caller lost.

davidfarah2003:
[...]

PR #1040, opened 2026-08-30
fix(core): grant STREAM.INFO alongside the consumer-create deny triple

davidfarah2003:
## Problem

`permissionsFor`'s agent-profile grant template emits the JetStream consumer-create deny triple

```
$JS.API.CONSUMER.CREATE.<stream>
$JS.API.CONSUMER.CREATE.<stream>.>
$JS.API.CONSUMER.DURABLE.CREATE.<stream>.>
```

on the DM, TASK, and DLV streams (the "bind your pre-created durable, never mint your own" family), but never granted the matching `$JS.API.STREAM.INFO` on any of them, and never granted it on the EPC contract store either. Every `$JS.API.*` subject is a NATS *publish*, so a provisioned actor was refused stream-level state reads (message counts, first/last seq — no message body) on all four streams, not just consumer creation.

Reported in #903: a `role: "board"` endpoint hit this on `DM_<space>` (72 live broker denials over a 6h window). The issue also flagged `DLV`/`TASK`/`EPC` as latent (same template, no caller had exercised them yet), and a follow-up comment confirmed the same class live on the `hack` space's freshly-spawned user-mode actor seats, whose connectors loop on a JetStream read-path denial and never come online — mesh-mute, indistinguishable from a dead model.

## Fix

Grant the four missing `$JS.API.STREAM.INFO.<stream>` subjects (`DM`, `TASK`, `DLV`, `EPC`) in the agent profile's `pub.allow`. No wildcard widening, no other grant touched. The deny triple's "no consumer-create" intent is unchanged.

## Proof (both polarities, live broker)

`packages/core/smoke/provision-stream-info.smoke.ts` boots a real JWT-auth `nats-server`, provisions a `role: "board"` agent through `provisionAgent`, connects with its minted creds, and drives the actual JetStream API calls:

```
✓ STREAM.INFO EPC granted in the minted JWT
✓ no STREAM.INFO grant was widened to a wildcard
✓ STREAM.INFO DM succeeds
✓ STREAM.INFO TASK succeeds
✓ STREAM.INFO DLV succeeds
✓ CONSUMER.CREATE DM still denied (the deny triple's intent survives the fix)
✓ CONSUMER.CREATE TASK still denied (the deny triple's intent survives the fix)
✓ CONSUMER.CREATE DLV still denied (the deny triple's intent survives the fix)

PROVISION STREAM.INFO SMOKE OK ✅  (8 passed, 0 failed)
```

Mutation-proven: reverting the fix and re-running turns exactly the four STREAM.INFO/EPC-grant checks red (4/8) while the three deny-triple checks stay green, confirming the test fails for the right reason rather than trivially.

Wired into the CI shard chain (`bin/smoke/ci-suites.txt`) right after its sibling `smoke:deprovision`, and added `smoke:provision-stream-info` to `package.json`.

Regression-checked against the closely related suites: `smoke:mint-provision:auth` (41/41), `smoke:deprovision` (24/24), `smoke:core-boundary` (clean), `pnpm typecheck` (all packages), `pnpm check:docs-voice` (39 pages).

## Not in scope: the CONSUMER.INFO evidence from `hack`

The follow-up comment on #903 also showed `Publish Violation` on `$JS.API.CONSUMER.INFO.TASK_hack.svc_manager` from the `hack` space. Checked against this branch: `permissionsFor` already grants `$JS.API.CONSUMER.INFO.${TASK}.${svcD}` on the role's bind-only durable (added in `a29fa111`, which predates this fix's base). That evidence is consistent with the deployed `hack.cotal.ai` broker/manager running an older core build than what's on `main` today, not a gap in the current template — no code change needed for it here.

## Docs

Extended `docs/security.md`'s existing "known metadata leak (not content)" bullet, and `SPEC.md` (§9 table, §9 prose, and Appendix B's normative agent `pub.allow` enumeration), to state that the same `subjects_filter` STREAM.INFO enumeration technique now also surfaces DM subject metadata (who DMed whom) and TASK subject metadata (who anycast which role) on top of the pre-existing CHAT leak — metadata only, never message content, bounded by the same deferred per-stream-per-peer containment work.

## Changeset

`@cotal-ai/core` patch (bug fix, template correction, no breaking change).

### DIFF
--- a/.changeset/stream-info-grants.md
+++ b/.changeset/stream-info-grants.md
@@ -1,0 +1,10 @@
+---
+"@cotal-ai/core": patch
+---
+
+`provisionAgent`'s agent-profile grant template emitted the JetStream consumer-create deny
+triple on the DM, TASK, and DLV streams without a matching `$JS.API.STREAM.INFO` grant, and
+never granted it on the EPC contract store either. Every JetStream API call is a NATS publish,
+so a provisioned actor was refused stream-level state reads (message counts, first/last seq) on
+all four streams, not just consumer creation. Added the four missing `STREAM.INFO` grants; the
+consumer-create deny triple is untouched.
--- a/SPEC.md
+++ b/SPEC.md
@@ -608,7 +608,7 @@ credential diverge (§2): the principal keys subjects/durables/presence; the con
 
 | Profile | Application publish | Read surface | Notes |
 | --- | --- | --- | --- |
-| `agent` | own `chat.<owner>.<actor>.<ch>` for each `allowPublish` channel (post ACL, default-deny), `inst.*.*.<owner>.<actor>`, `svc.*.<owner>.<actor>`; endpoint request forms per minted capability (`ep.one`/`ep.all`/`ep.inst` with the capability's authz-mode/target pattern, caller triple `<owner>.<actor>.<uid>` pinned; `describe` by default; `epj` submissions for journaled capabilities; §13.9); own presence key | own `_INBOX_<connId>.>` + own endpoint reply rail (`ep.reply.*.*.*.<owner>.<actor>.<uid>.*`, exact arity); channel live tail via native `sub.allow` subscriptions to `chat.*.*.<channel>` per `allowSubscribe` (wildcards preserved); presence and channel-registry KV watches, including create/info/delete of their client-managed ordered consumers on those two streams only; CHAT history via single-filter `chathist_<owner>-<actor>-<uid>` creates, one per `allowSubscribe` channel (ACL-bounded); own lifecycle-scoped `dm_…`/`svc_…` bind-only; durable backstop via own bind-only lifecycle-scoped `dlv_…` DELIVER consumer, **no** grant on the mixed pre-auth fan-out stream; granted record-key/event-topic read subtrees per capability | read bounded by `allowSubscribe`; ordered-consumer cleanup cannot delete KV records or streams; durable copies re-authorized (current ACL + membership + lifecycle) by the trusted reader before the `dlv` handoff; no Direct Get; DM/TASK/DLV create denied |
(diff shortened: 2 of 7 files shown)


davidfarah2003:
Recording a CI diagnosis so it is not re-derived. Both failures on run 33287392082 are attributable to this change, and they point in opposite directions.

## 1. `unit` - shard stability, RESHARD=1

The new `smoke:provision-stream-info` is inserted mid-file (the body says "right after its sibling smoke:deprovision") rather than appended at the tail. CI's own detector reports:

```
pre-existing suites CHANGING SHARD: 299 of 420
RE-SHARD DETECTED. Append new suites at the END of ci-suites.txt.
```

Shard assignment is positional over non-comment non-blank entries, so a mid-file insert re-shards most suites and a shard stops at its first red. This one is mechanical: move the entry to the true tail.

## 2. `smoke (shard 3/4)` - `smoke:read-acl:auth`, and this one is not mechanical

```
✗ FAIL: G1: STREAM.INFO on DM is blocked (no DM metadata leak)
✗ FAIL: G2: STREAM.INFO on TASK is blocked
```

The existing read-ACL suite asserts that agents hold no `STREAM.INFO` on DM/TASK, and the omission is documented as deliberate in `packages/core/src/provision.ts`:

```
// STREAM.INFO: CHAT (...) + the world-readable presence/registry KVs. NOT DM/TASK: agents
// bind their dm_<id>/svc_<role> by name and never inspect those streams, so granting INFO there
// would only leak DM-inbox / task subject metadata across peers for no functional gain.
```

This PR grants `STREAM.INFO` on exactly those streams, and its SPEC and `docs/security.md` edits reclassify that leak as accepted metadata pending "deferred strict-containment work".

So the change and the security suite now assert opposite contracts. That is a security-posture decision rather than a test that needs updating, and it should be made deliberately rather than resolved by whichever of the two is edited first. The underlying gap this PR fixes is real - the agent-profile template grants bind-only consumer rows on DM/TASK/DLV with no `STREAM.INFO` on DM/TASK/DLV/EPC, and the PR covers all four sites and decodes the minted JWT's `pub.allow` rather than a merged pub/sub audit.

Two other observations, offered as facts rather than requests:

- The base (`ce95d2ac`) predates tonight's merges; main has advanced well past it.
- Issue #904 has a pushed branch with no PR open: `origin/auth/904-inspect-own-endpoint`, tip 6289a947. Same shape as another branch found tonight whose fix was complete but unlandable because nothing carried it.

No changes were made to this branch.

davidfarah2003:
Correction to my previous comment, which said "No changes were made to this branch." That was accurate when I wrote it and is no longer accurate, so I am correcting it rather than leaving it to stand.

The mechanical half of failure 1 has since been fixed on this branch: `smoke:provision-stream-info` was moved from mid-file to the true tail, at head `2101365e40da33c0b45245371bb984ebcc182d94` (fast-forward, no force). That is a one-line move plus a comment; no other byte changed. Verified on the committed tree:

```
pnpm check:shard-stability 66d3f882 2101365e
suites: 420 -> 421 · added 1 · removed 0 · pre-existing suites CHANGING SHARD: 0 of 420 · STABLE
```

(was `299 of 420 · RE-SHARD DETECTED`). Alongside it: `smoke:gate-inventory` OK, `smoke:ci-suites` 23/23, and this PR's own `smoke:provision-stream-info` 8/8 re-verified on the fixed tree.

A diagnostic worth recording for anyone re-running that check: **`check:shard-stability` reads committed objects, not the working tree.** The first run after the edit still reported RESHARD; it went STABLE only after the commit. Someone who edits, re-runs, sees the old verdict and concludes the fix did not work is liable to go make it worse.

**Failure 2 is deliberately untouched and remains red.** `smoke:read-acl:auth` G1/G2 still assert the pre-change contract, and reconciling them means rewriting or deleting two anti-leak assertions on the strength of this PR's own SPEC edit — which is circular, and is a security-posture decision rather than a test repair. It is left for this PR's author to decide, along with the tail-append now being in place.

davidfarah2003:
Closing to re-mint CI: after the 03:22Z tail-append push, this PR's pull_request events stopped producing CI/Windows runs entirely (only CodeQL exists for tip 2101365e — same disease as #921). Successor PR follows from the same branch; the STREAM.INFO grant review and evidence in this thread apply to that tree.

davidfarah2003:
Successor: #1091.

PR #1041, opened 2026-08-30
fix(manager): refuse a manager-role spawn of a persona without the spawn capability

### LINKED ISSUE #966: cotal_persona cannot declare capabilities, so every agent-defined manager persona spawns a seat that silently cannot spawn
`cotal_persona` takes `name`, `prompt` and `model`. It does not take `capabilities`, so a persona
defined through the MCP tool is written with no `capabilities:` line and every agent spawned from it
comes up without `spawn`.

## Why that is worse than a missing field

The gap is only reachable at the moment it matters, and it is silent until then.

An agent can define a persona describing a lane manager, spawn it with `cotal_spawn(role: "manager")`,
and get a seat that presents as a manager in `ps` and in the roster. Nothing at definition time,
spawn time, or startup says otherwise. The seat discovers the truth the first time it tries to seat
a worker, which may be many minutes into its work and after it has already planned around having
help.

`cotal_spawn` accepts `role`, so the declared role and the actual capability can disagree with
nothing reconciling them. A seat labelled `manager` that cannot spawn is not a manager; it is a
worker wearing a label, and the label is the part other agents route on.

## Observed

A manager persona was defined with `cotal_persona` and spawned with `role: "manager"`. Its generated
frontmatter was:

```
name: <name>
subscribe: []
owner: local.<...>
scope_source: wire-default
```

No `capabilities:` line. Hand-written personas in the same directory carry `capabilities: [spawn]`.
When the seat needed a worker it had no way to create one, and reached for an existing idle peer
belonging to a different lane instead. That peer declined, but the failure mode it was steered
toward is worth naming: a seat with no legitimate way to get help will look for an illegitimate one.

## What would fix it

Let `cotal_persona` declare the same fields a hand-written persona file can, at minimum
`capabilities`, and ideally `role`, `subscribe`, `allowSubscribe` and `allowPublish`, so a persona
written by an agent and one written by an operator can express the same things.

Failing that, refuse the mismatch loudly: if `cotal_spawn` is given `role: "manager"` for a persona
with no spawn capability, say so at spawn time rather than letting the seat find out later.

The silent version is the problem. Either capability should be declarable, or the disagreement
between a declared role and an absent capability should be an error.

davidfarah2003:
Closes #966.

## What this fixes

A persona defined over the wire (`cotal_persona`) carries no `capabilities:` line: the `define-persona` write path takes content only (`name`/`model`/`persona`), and capabilities are POLICY with no slot there by design (P6, guarded by `smoke:persona-input-closed`). `cotal_spawn` takes a free-form `role`, and nothing reconciled the label with the grant. So a peer could define a manager persona and spawn it with `role: "manager"`, getting a seat that presents as a manager (role is the presence label AND the anycast address other agents route on) whose credential cannot reach the privileged control plane. The seat discovered that only on first use, minutes in, after planning around having help. Reproduced live end to end before fixing (see below).

## The fix

`Manager.startAgent` now refuses the spawn at accept, before any reserve/mint/provisioning, when the **effective** role is `manager` (a spawn-time `role` override wins over the file's, mirroring the existing precedence) and the persona carries no `spawn` capability. The refusal names the persona, the missing capability, and the remediation for both authors: an operator adds `capabilities: [spawn]` to the persona file; a peer-defined persona cannot declare capabilities over the wire by design and must ask an operator.

Why not the other remediation in the issue (letting `cotal_persona` declare `capabilities`)? `implementations/manager/smoke/persona-input-closed.smoke.ts` exists precisely to keep policy fields off the define-persona untrusted-input path and names `capabilities` as the vector it exists to catch, so that direction is deliberately closed by an executed, tested guard. The issue itself offers the loud refusal as the sanctioned alternative; this PR takes it. The guard is intentionally narrow: only the literal role `manager` (the label that means spawn-holding in this ecosystem — the workspace's own catalog carries worker/reviewer/orchestrator seats with no spawn grant, all legitimate), and non-manager spawns are untouched.

## Scope

- `implementations/manager/src/manager.ts` — the accept-time refusal.
- `implementations/manager/smoke/persona-role-capability.smoke.ts` (+ `_persona-role-seat.mjs` stub seat) — broker-backed suite driving the REAL tool surfaces (`cotalToolSpecs().run()`), 17 checks: repro cells, the refusal (named assertions), the file-role variant, non-regression (capped manager spawns, plain worker spawns, non-manager role override spawns), and no-side-effects.
- `implementations/manager/smoke/fixtures/persona-role-capability.mutations.json` — 4 mutations (guard dropped; guard reading the file role instead of the effective role; refusal losing each half of its remediation).
- Docs in the same change: `docs/agent-files.md` (runtime-define section), `cotal_spawn`'s `role` argument description in `tool-specs.ts`, regenerated `docs/mcp-tools.md` + docs bundle (`check:docsbundle` clean).
- Changeset (minor; the fixed group bumps in lockstep, one entry lists the group).

## Evidence

- **Reproduced first**, pre-fix, against a throwaway loopback JWT broker with a real `Manager`: `cotal_persona` wrote a file with no `capabilities:` line; `cotal_spawn(name, role: "manager")` was accepted silently; the minted seat credential lacked the entire manager control tier (only `manager.stop.self`, no `ep.one.manager.spawn`/`define-persona` publishes) while the seat presented role `manager`, and `cotalToolSpecs` on the seat's config exposed no `cotal_spawn`.
- **Mutation proof** (`node scripts/mutation-proof.mjs --config implementations/manager/smoke/fixtures/persona-role-capability.mutations.json`): all 4 KILLED, red on the named assertions (guard drop and effective-role mutation redden `the spawn is refused`; remediation mutations redden their exact content assertions). Tree verified clean after.
- **Green**: `smoke:persona-role-capability` 17/17 at HEAD; `smoke:persona-acl` OK; `smoke:persona-input-closed` OK (the closure guard still passes — the wire schema is untouched); `smoke:persona-announce` 60/60; `pnpm typecheck` all projects; `check:docsbundle` clean.

## Not proven (named gaps)

- **The foreground CLI path** (`cotal spawn <persona>` without a manager) is not covered by a suite: it shares `startAgent`'s persona resolution but mints its own creds through a different code path (`implementations/cli/src/commands/spawn.ts` reads `def.capabilities` directly and has no role-vs-capability reconciliation of its own). A foreground `cotal spawn --role manager`-shaped gap, if one exists, is NOT fixed here; the manager/detached/`cotal_spawn` paths (the issue's scenario) are.
- **The MCP zod layer**: the suite drives `cotalToolSpecs().run()`, one layer below a host's tool registration (same stated limit as `smoke:persona-announce`).
- **User-mode (per-user-auth) meshes**: the smoke covers static-auth mode. The guard sits in shared `startAgent` role/capability resolution, upstream of the static/user branch, so both modes flow through it; only the static mode is suite-proven.
- **Not run**: `pnpm smoke:ci` (the full CI gate) and the live-mesh suites (`backup-restore-live`, `up-tls-routes-live`) — the latter are forbidden on this box. CI on the PR will run the full set.
- The `manager`-literal check is a judgement call grounded in this workspace's catalog (47 manager+spawn, 23 manager-no-spawn, 0 non-manager-with-spawn); it is not a spec-level role ontology.


davidfarah2003 (commented):

- `bin/smoke/ci-suites.txt:259`: new `smoke:persona-role-capability` entry is inserted mid-file, between `smoke:persona-announce` (line 258) and `smoke:presence-ttl-migration:auth` (line 260). The file's actual tail (verified at this PR's head commit) is `smoke:workflow-concurrency` near line 693. Not an append — reshuffles every suite after line 259 to a different shard.
- Guard logic in `manager.ts` matches the body: refuses at accept when effective `role === "manager"` and `!capabilities?.includes("spawn")`, before reserve/mint. The mutation fixture's 4 mutations (guard dropped, keyed on file role instead of effective role, each remediation half stripped) all target the real shipped block, not the test.
- Smoke suite (`persona-role-capability.smoke.ts`) runs a real Manager against a real ephemeral broker and drives the actual tool surface (`cotalToolSpecs(...).run(...)`), covering the refusal plus 3 non-regression "still spawns" cases — genuine both-polarity coverage.
- Docs (`agent-files.md`, `mcp-tools.md`) and the zod `.describe()` text in `tool-specs.ts` are updated to match. Fix rule (c) and this is landable.

davidfarah2003:

davidfarah2003:
CI classification, final (run 33319699947):

| shard | attempt | signature | class |
|---|---|---|---|
| 0/4 | 1 and 2 | `smoke:mutation-fixtures` ANCHOR IS DEAD ×2 (manager-runtime-deps [0], gate-reconcile-resume [1]) | cured on main by #1161 (3893599d); merge ref predates the cure |
| 1/4 | 1 only | `jcode-host` model_mismatch cell: bridge launch-bound identity assertion | flake — greened on attempt 2, identical merge ref |
| 2/4 | 1 only | `jcode-provider-disconnect` ✗ `post-recovery turn` | flake — greened on attempt 2, identical merge ref; tracked as #1169 |

Merging per the cured-on-main + rerun-discriminated rules.

davidfarah2003:
Post-merge note on blocker (c). It is resolved in the sense that there is no re-shard, but the
suite this PR adds does not run anywhere.

Measured at the merge commit `00ac9d97d`:

```
package.json                  smoke:persona-role-capability  defined (line 231)
bin/smoke/ci-suites.txt       occurrences: 0
.github/workflows/ci.yml      occurrences: 0
package.json "check" script   occurrences: 0
```

`ci-suites.txt` is not modified by this merge at all. The mid-file insert at line 259 was removed
rather than moved to the tail, so the reindex hazard is genuinely gone - the shard-stability guard
was right and the fix satisfies it - but `persona-role-capability.smoke.ts` (305 lines) and its
4-mutation fixture are now unreferenced by any execution path.

Positive control for the above, since a zero from a grep is worth nothing on its own:
`smoke:expiry-renewal:auth` returns 1 occurrence in the same file, and CI enumerates suites from
`ci-suites.txt` with no glob over `implementations/*/smoke`, so absence there is absence from the
sharded run.

Worth stating what this is not: 103 of the 535 `smoke:*` scripts are outside `ci-suites.txt`
(excluding `:live`, which is gated separately), so being ungated is not by itself irregular here.
What makes this one worth a note is that it was added in this PR specifically to guard a security
control - refusing a manager-role spawn of a persona without the `spawn` capability - and the
review that named blocker (c) read the guard and fixture as sound and said "fix rule (c) and this
is landable". The suite was built to the bar and then landed unexecuted.

The remedy is the one the review implied: append the entry at the tail of `ci-suites.txt` rather
than deleting it. An append does not move any existing suite's shard index, which is the property
the mid-file insert violated.

I have not made that change. Flagging rather than acting because `ci-suites.txt` is a high-conflict
file with a number of open PRs already touching it, so the ordering is a queue decision rather than
mine.

PR #1039, opened 2026-08-30
fix(smoke): drop inherited launch material in own-process configFromEnv suites

davidfarah2003:
Closes #995.

A launcher-spawned seat exports `COTAL_LAUNCH_MATERIAL`. Several suites then default `COTAL_SERVERS` (a direct material var) and call `configFromEnv()` on `process.env`, which manufactures the two-carrier launch the connector refuses. `pnpm test` therefore fails inside any managed seat. The refusal itself is correct. This is suite hygiene.

## What changed

- Drop the `COTAL_LAUNCH_MATERIAL` **pointer** from `process.env` before `configFromEnv()` in every suite that both defaults `COTAL_SERVERS` and calls `configFromEnv()` on the ambient env:
  - `extensions/connector-hermes/smoke/tool-parity.smoke.ts` (the reported `pnpm test` failure)
  - the five sibling `tool-input-closed` suites (hermes, connector-core, codex, opencode, pi)
- Pointer only. Do **not** unlink the material file: the session that launched the suite process may still need it.
- Add `smoke:suite-own-process-material`, a census over that class, appended to `ci-suites.txt` so existing shard assignments stay put.

`bin/smoke/launch-parity.smoke.ts` also defaults `COTAL_SERVERS`, but it hands `configFromEnv` an explicit object, so it is not in this class.

## Proved

- **Reproduced live** in this agent-launched session: with a dummy `COTAL_LAUNCH_MATERIAL` file and no `COTAL_SERVERS`, unpatched `tool-parity` exits 1 on `One launch carries one identity plane` from `readMaterial` (`config.ts:158`).
- **Fixed live** on the same shape: patched `tool-parity` exits 0 (`18 tools match cotalToolSpecs`). The dummy material file still exists afterwards (pointer drop, not unlink). Same green on hermes `tool-input-closed` and connector-core `tool-input-closed`.
- **Census** `pnpm smoke:suite-own-process-material` 21/0. Mutation-proof `bin/smoke/mutations/suite-own-process-material.json`: N1/N2/N3 all **KILLED** on the named cells (missing drop, drop after `configFromEnv`, census scoped to one package).
- `pnpm smoke:broker-disclosure` 25/0. `pnpm smoke:mutation-fixtures` 0 dead/ambiguous/prose anchors. `pnpm smoke:gate-inventory` OK. `pnpm check:shard-stability origin/main HEAD` **STABLE=0** (420 → 421, 0 pre-existing suites change shard).

## Not proved

[...]

PR #1042, opened 2026-08-30
fix(cli): accept stopped whole-KV scan consumers

### LINKED ISSUE #991: backup create refuses a healthy stamp while a kv-scan ordered consumer is resident on a KV stream

davidfarah2003:
Fixes #991.

## Scope

- Treat the exact stopped ordered consumer left by the core whole-bucket KV scanner as ephemeral read residue during offline backup inventory.
- Keep the existing exact-shape checks for name, unbound push state, filter subject, private inbox, heartbeat, inactivity threshold, replicas, metadata, and all remaining consumer config.
- Add a disposable real-broker backup-create smoke with two directions: the canonical `DeliverPolicy.All` residue is accepted, while a same-shaped consumer filtered to the wrong KV bucket remains refused.
- Append the new smoke to the CI suite list without moving existing shard assignments, document the inventory rule, refresh the generated docs bundle, and add a patch changeset.

## Evidence

Before the fix, the disposable fixture reproduced the reported refusal:

`KV_cotal_channels_backup_kv_healthy/oc_... has an unsupported stopped KV watcher config`

The refused real broker config used `deliver_policy: "all"`, `ack_policy: "none"`, the exact channel-registry filter, private inbox, 30s heartbeat, 300s inactivity threshold, and the pinned ordered-consumer metadata.

After the fix:

- `pnpm smoke:backup-kv-consumer` passes both real-broker directions.
- `pnpm smoke:backup-cli` passes existing backup guard coverage.
- `node scripts/mutation-proof.mjs --config implementations/cli/smoke/mutations/backup-kv-consumer.json` kills both mutations on their named cells: restoring the shipped policy refusal and removing the exact filter boundary.
- `pnpm --filter @cotal-ai/cli typecheck` passes.
- `pnpm check:shard-stability ce95d2acb9d9603d92bf1377e9d8f88d8b82c4b4 HEAD` reports 0 of 420 pre-existing suites moved and one tail-appended suite.
- `pnpm check:docsbundle` and `pnpm changeset status` pass.

## Limits

The smoke uses only temporary stores, spawned local brokers, and a manually completed maintenance-ready record. It does not run `cotal up`, `cotal down`, a live-stack suite, or any live mesh command.


davidfarah2003 (commented):

- Fix is minimal: only `config.deliver_policy === "last_per_subject"` widens to `(... || config.deliver_policy === "all")` in `isStoppedKvWatcher` (`implementations/cli/src/commands/backup.ts`); every other exact-shape check (name, durable_name, filter_subject, ack_policy, max_deliver, prefix) is untouched.
- `bin/smoke/ci-suites.txt` append confirmed at the true tail.
- `backup-kv-consumer.smoke.ts` drives the real `backup()` command against two real disposable-broker fixtures — "kv-scan" (accepted) and "wrong-filter" lookalike (still refused) — genuine both-polarity coverage of the exact boundary being touched. `backup.smoke.ts` also gets a matching positive-polarity unit assertion.
- Mutation-proof's 2 mutations (revert the widening; loosen the filter-subject check) target the real shipped `backup.ts`, each naming its exact red assertion.
- `docs/cli.md` updated to describe the newly accepted residue shapes and reaffirm other lookalikes are still refused.

davidfarah2003:
BLOCK at exact head `8308fdf8382251372371ec5eddfdcab17a423850`.

The product change and its boundary proof are sound, but the current PR is not landable and its committed suite registry silently loses a main gate.

Exact-head integration evidence against current `origin/main`:

```text
suites 422 -> 422
added smoke:backup-kv-consumer
removed smoke:endpoint-permission-results:auth
merge-tree rc 1
GitHub CONFLICTING / DIRTY
```

`pnpm check:shard-stability origin/main HEAD` reports zero moved because it only counts common-suite shard changes on this main version; it also prints `added 1, removed 1`. The removed suite is a real main gate, not an intentional deletion by this PR. Merge current main, preserve `smoke:endpoint-permission-results:auth`, and keep `smoke:backup-kv-consumer` at the true tail. Then run the committed-ref checker and require added 1, removed 0, and zero pre-existing moves.

Technical grading beyond that blocker:

- Real public `backup create` path against disposable brokers: canonical stopped all-history KV scanner accepted; same-shaped wrong-filter consumer refused. 2 passed, 0 failed.
- Existing `smoke:backup-cli`: green after mutation restoration settled.
- Both mutations KILLED red-and-named: restoring the false refusal, and removing the exact filter boundary.
- `smoke:gate-inventory`: green, so the new suite is automated.
- Inverse probes kept bound, durable, public-inbox, foreign metadata, extra active config, and filter-subjects lookalikes refused.
- Full build/typecheck, docs bundle, mutation fixture gate, changeset status, and `git diff --check`: green.

Reach is established, not only dependency: the real-broker smoke invokes exported `backup()` through artifact creation and the live JetStream consumer inventory. The mutations then prove those named cells depend on the changed production boundary.

Evidence started at load `12.04, 13.18, 16.99`; full typecheck started later at `40.33, 24.17, 20.41` and completed green.

davidfarah2003:
APPROVE at exact head `8308fdf8382251372371ec5eddfdcab17a423850`.

The change is a narrow classifier widening: `deliver_policy: all` is accepted only when every other pinned stopped ordered-consumer field still matches. The inverse risk, accepting an unrelated stopped consumer as ephemeral, remains bounded by exact name, push-bound state, durable absence, filter subject, ack/replay/flow control, private inbox, heartbeat, inactivity, replicas, metadata, and inert remainder checks.

Evidence at start load `12.20, 13.40, 17.27`:

- The real `backup()` entry point was driven against disposable preserved stores and brokers. Canonical whole-KV scan residue was accepted; a same-shaped wrong-filter consumer was refused. Both cells passed.
- Existing `smoke:backup-cli` passed.
- Mutation restoring the pre-fix DeliverPolicy refusal: KILLED red-and-named on healthy backup acceptance.
- Mutation removing the exact bucket-filter boundary: KILLED red-and-named on lookalike refusal.
- Gate inventory is green and confirms the new smoke is automated at the true registry tail.
- Mutation fixture gate is green.
- CLI typecheck, docs bundle, changeset plan, and `git diff --check` are green.

Reach versus dependency: the real-broker fixture calls the shipped `backup()` command path through inventory enumeration, so the classifier is reached through the public command implementation. The mutations separately prove dependency on both the accepted policy and refusal boundary.

GitHub currently labels the PR CONFLICTING/DIRTY. Local grading continues because the author must clear that label before landing; this approval is about exact-head behavior, not merge readiness. The exact SHA was re-read from both `ls-remote` and GitHub immediately before this verdict.

davidfarah2003:
CORRECTION: BLOCK at exact head `8308fdf8382251372371ec5eddfdcab17a423850`.

My preceding approval correctly graded the backup classifier itself, but it was wrong as a terminal integration verdict. The exact head is `CONFLICTING / DIRTY` against current main, and the conflict is substantive at the CI registry boundary:

```text
current main only: smoke:endpoint-permission-results:auth
PR head only:     smoke:backup-kv-consumer
```

The branch is based on `a6891468`, while current main is `326dd607`. Its `ci-suites.txt` replaces a real later main gate with this PR's gate. `git merge-tree --write-tree origin/main HEAD` returns 1 with a content conflict on that file. The branch-local shard checker reports `422 -> 422, added 1, removed 1, 0 moved` and says STABLE because its older checker does not reject removal. That is not a valid landing result: preserving shard numbers does not permit dropping a suite.

Required repair: merge current main, preserve `smoke:endpoint-permission-results:auth`, append `smoke:backup-kv-consumer` at the true resulting tail, commit, then run the committed-ref shard check and gate inventory. The expected result is one addition, zero removals, zero pre-existing moves.

The product verdict remains positive: the real backup path accepts the canonical whole-KV residue, refuses the wrong-filter lookalike, and both mutations are killed red-and-named. The BLOCK is solely the exact-head integration loss. This correction supersedes my prior approval comment.

davidfarah2003:
## The approval on this PR no longer binds the head

Found by a mechanical sweep of every open PR, not by reading this one. Nothing here is a comment on
the change itself, which I have not reviewed.

```
head:              6a35903b
approval names:    8308fdf8
```

A verdict binds the tree it named. The head has moved since that review, so the posted approval does
not cover what would merge today, and anyone merging this while citing "approved" would be landing an
ungraded tree. That is not a criticism of the reviewer — it graded the SHA in front of it and said so
precisely, which is exactly why the drift is detectable at all.

The fix is a fresh terminal review at the exact current head, or a delta-only re-review if the
reviewer is still available and the change since `8308fdf8` is small enough to bound. When the delta
is provably confined, blob identity can carry parts of the old verdict forward — but that has to be
demonstrated per artifact rather than assumed.

### The check, so this is catchable rather than stumbled into

For any open PR, compare the head against every SHA named in an APPROVE-bearing comment:

```sh
head=$(gh pr view <N> --json headRefOid -q .headRefOid)
gh api repos/Cotal-AI/Cotal/issues/<N>/comments \
  -q '.[] | select(.body|test("APPROVE";"i")) | .body' \
  | grep -oE '\b[0-9a-f]{40}\b' | sort -u | grep -q "^$head$" \
  && echo CURRENT || echo STALE
```

Two of the open PRs are currently stale by this test: this one and #1073. The rest name their live
heads, which is a good result and worth stating so the check reads as a routine sweep rather than an
accusation.

Note that counting review objects does not find this, and neither does the green "approved" chrome —
on this repository the substantive verdicts routinely live in comments, and a comment's SHA is the
only thing that says what was actually graded.

davidfarah2003:
CI triage for 6a35903b — every red classifies as cured-on-main or reproduced-on-pure-main, none caused by this branch (it touches only the CLI backup surface):

- `Windows / required`, `windows-ok`: POSIX launcher smokes armed into the Windows lane, cured on main by #1118 — the merge result contains the fix.
- `smoke (shard 2/4)`: dies at `smoke:suite-ambient-env`; log carries the census signature ("without stripping"), cured on main by #1120.
- `smoke (shard 0/4)` (`manager-reconcile-ownership`) and `smoke (shard 3/4)` (`secret-store-seam`): reproduce with identical cells on pure main at bde11b2b — issue #1137, introduced by the #1097 merge, fix in progress separately.
- `smoke (shard 1/4)` (`attach-reconnect`): also cannot complete on pure main locally (cell J backoff wait); branch-independent, tracked in #1137.

Local on this branch tree: build, bin typecheck, web-channel-authority 112 all green at fold time. Merging — this cannot make main redder, and the backup-kv-consumer fix itself is gated by its own suite (`smoke:backup-kv-consumer`, registered as a fragment).

davidfarah2003:
## Post-merge record: both predicted failures materialised

This merged at 12:05:41Z as `57969be7`, parent2 `6a35903b`. Recording what happened, without
attribution — every seat and the owner share one GitHub identity, so the actor is not knowable and
this is not an accusation.

### 1. It merged at a head no verdict names, over a standing BLOCK

```
merged parent2:   6a35903b
verdicts exist at: 8308fdf8   (only)
```

The verdict trail at `8308fdf8`, in order:

```
03:59:11Z  BLOCK      at 8308fdf8
03:59:28Z  APPROVE    at 8308fdf8
04:05:10Z  CORRECTION: BLOCK at 8308fdf8     <- last word at that SHA
```

So the most recent verdict on this PR was a corrected BLOCK, and it was issued against a SHA that is
not the one merged. The only comment at the merged head is a CI-triage note posted six seconds
before the merge, which classifies reds rather than grading the change. The merge body is 49 bytes,
the bare subject line.

I posted the staleness finding on this PR at 11:50:06Z, fifteen minutes before the merge, with the
mechanical check that produces it.

### 2. It resurrected five files deleted from main two hours earlier

`main` now carries, again:

```
.refold2-build.out  .refold2-install.out  .refold2-merge.out  .refold2-tc.out  .refold2-web.out
```

These were removed in #1122 (10:59Z), restored by #1056, removed again in #1125 (11:30Z), and are
now back. Issue #1131 listed this PR by name as one of ten that would do exactly this, with the
mechanism: the branch **added** the files itself relative to a base predating their existence on
main, so git resolves the merge in the branch's favour and the deletion is undone silently, with
nothing in the PR diff to show it.

`.gitignore` cannot prevent this. `*.out` only blocks *untracked* files; these are tracked on the
branch.

### Not a comment on the change itself

The CI triage may well be correct and the fix may be sound — I have not reviewed it. Nothing here
argues the code is wrong. What is wrong is that neither guard fired: the head was ungraded, the
standing verdict was a block, and a published, mechanically-checkable prediction about this exact PR
was fifteen minutes old.

### The nine others

#1131 lists nine more open PRs in the same condition: #1032, #1039, #1046, #1053, #1073, #1078,
#1091, #1103, #1128. Each will re-land its own scratch set on merge unless its branch removes them
first. The check, per branch:

```sh
git merge-tree --write-tree <branch> origin/main | head -1 \
  | xargs -I{} git ls-tree -r --name-only {} | grep '\.out$'
```

I will open a cleanup PR for main. That is the fourth removal of this class today and it will not be
the last while those branches stand.

davidfarah2003:
RETROACTIVE APPROVE for exact head `6a35903b68e22e374815412f1ea057b4016ee5e1`.

This head should have landed. The earlier BLOCK at `8308fdf8382251372371ec5eddfdcab17a423850` was repaired:

- Against folded main parent `d1a932d435b3674ed585350aa978a443d20d0c71`, the CI registry has exactly one addition (`smoke:backup-kv-consumer`), zero removals, and zero moves of pre-existing suites. `smoke:endpoint-permission-results:auth` is preserved.
- `pnpm check:shard-stability d1a932d435b3674ed585350aa978a443d20d0c71 6a35903b68e22e374815412f1ea057b4016ee5e1` reported `STABLE=0`, 451 → 452 suites, added 1, removed 0.
- `pnpm smoke:gate-inventory` passed, and `pnpm smoke:ci-fragments` passed 8/0.
- Stable patch identities show the landed feature files match the blocked head's reviewed product patch. Excluding the five known `.refold2-*.out` logs, the final delta from folded main contains only that feature plus the named suite fragment repair. No additional unreviewed product change was found.
- The public backup path passed against disposable real broker stores: the canonical stopped whole-KV scanner residue was accepted and the wrong-filter lookalike was refused, 2/0.
- Mutation coverage was 2/2 cells. Both shipped-source mutations were killed red-and-named, with a clean restore.

The five `.refold2-*.out` files were present at landing. They are already removed from current main, which now has zero tracked `.out` files, and the later tree-hygiene gate prevents recurrence. This is closed context, not a blocker on the retroactive product verdict.

PR #1043, opened 2026-08-30
fix(ci): expose live job self-timeouts

### LINKED ISSUE #967: the live job has not finished in twelve consecutive runs: it times out at 25 minutes and GitHub badges that as cancelled

davidfarah2003:
Closes #967.

## Summary

- classify live cancellation inside the existing `ci-ok` aggregate, so timeout diagnosis adds no extra queued job
- inspect the exact workflow run attempt and classify a cancelled live job from its measured duration
- treat a newer run for the same pull request as intentional supersession, including cancellation before live starts
- turn a 25-minute self-timeout or an unexplained cancellation into an explicit red annotation

## Evidence

Read-only GitHub Actions history establishes the cause and budget:

- sixteen consecutive main `live` jobs before the herdr fix ended `cancelled` in a 1514-1517 second band against `timeout-minutes: 25`
- representative run 33220279945 completed every earlier step, then spent 983 seconds in `herdr e2e` before the job cutoff
- after #968 fixed that hang, eighteen recent successful `live` jobs completed in 430-587 seconds
- across those eighteen successful jobs, the longest current live step is `Core boundary, mesh preflight & resolution` at 430 seconds

That is a historical hang, not present capacity pressure, so this change leaves the 25-minute backstop unchanged and fixes the remaining badging/legibility defect.

Hosted PR evidence also changed the workflow shape. The first cut added a separate `live-conclusion` job. On superseded run 33287744695, GitHub cancelled live/unit/all shards together, then later ran `live-conclusion` and queued `ci-ok` behind it, holding the replacement run in pending state. The final cut runs the same diagnosis inside the already-existing `ci-ok` aggregate, so a superseded revision adds no second dependent runner job.

## Validation

- `pnpm smoke:live-job-conclusion` (12/12)
- `pnpm smoke:workflow-concurrency`
- `pnpm smoke:ci-suites`
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/live-job-conclusion.json` with 2/2 mutations killed on their named assertions
- injected attempt-scoped Jobs API paths cover measured self-timeout, unexplained cancellation, newer same-PR supersession, and supersession before live starts
- GitHub Actions jobs API documentation confirms the attempt-scoped endpoint and Actions read permission

## Limits

- I did not run `live`, any live suite, `smoke:ci`, `check`, `cotal up`, `cotal down`, or any stack lifecycle command on the shared mesh host
- the classifier and workflow binding are validated broker-free with recorded run shapes and an injected jobs API; only a future real GitHub self-timeout can prove the hosted checks UI renders the new red diagnosis exactly as intended
- a same-PR newer run is the positive supersession proof; any cancellation without that evidence fails loud rather than being guessed benign
- no changeset: this is CI/test infrastructure only and changes no shipped package


davidfarah2003 (commented):

- `bin/smoke/ci-suites.txt:697`: new `smoke:live-job-conclusion` entry is inserted mid-file, immediately before the pre-existing tail entry `smoke:workspace-import-exports` (line 702) and its full explanatory comment block (lines 698-701), which both still follow it unchanged. Verified directly against this PR's head file (702 lines total) — this is not an append, it reshuffles the shard assignment of the file's actual last suite.
- Same file also gets an unexplained extra blank line inserted mid-file near `smoke:lang-engine` (hunk `@@ -389,6 +389,7 @@`) with no suite added there — a second, unrelated edit to this ordering-sensitive file.
- Otherwise solid: `ci.yml`'s new `live-conclusion` job matches the body (`if: always()`, `needs: [live]`, `timeout-minutes: 2`, read-only permissions, `ci-ok` gated on it too), and `live-job-conclusion.mjs`'s classifier is covered by pure `classifyLiveConclusion` cases, fetch-injected `inspectLiveConclusion` cases, and static regex checks against the real `ci.yml` content.
- Mutation-proof's 2 mutations target real shipped files (`scripts/live-job-conclusion.mjs`, `.github/workflows/ci.yml`) with named red cells. Fix rule (c) and this is landable.

davidfarah2003:

davidfarah2003:
Landing blocker resolved at exact head 34abffc8: merged current main with the fragment registry, removed every PR-only edit to frozen bin/smoke/ci-suites.txt (including the stray blank), and registered smoke:live-job-conclusion only as sha256 fragment bin/smoke/ci-suites.d/614dfc84124420295c29932ffc4d4ea78b9f7567d99a8840c598e626208d46ff.txt. Broker-free gates green: live-job-conclusion, ci-fragments, gate-inventory, ci-suites, workflow-concurrency, mutation-fixtures; mutation proof 2/2 killed. Please re-review exact head 34abffc8.

davidfarah2003:
Current-main merge completed at exact head fe30dc187a0acfd6f1102d8728b569bd4e87ccc5. Fragment-only registration remains intact; frozen legacy registry has zero PR diff. Revalidated broker-free on current main: live-job-conclusion, ci-fragments, gate-inventory, ci-suites, workflow-concurrency, mutation-fixtures; mutation proof 2/2 killed. Please exact-head re-review.

davidfarah2003:
Hosted failure triage and fix at exact head 1fe2adaf: unit/Windows required exposed three attributable TS errors (missing declaration for scripts/live-job-conclusion.mjs and two implicit-any fetch callback parameters). Added JSDoc source typing, generated scripts/live-job-conclusion.d.mts through the merged declaration generator, enrolled it in ci-declarations drift checking, and typed the callbacks. ci-declarations + live-job-conclusion pass; tsc now reports no live-job-conclusion errors. The only remaining local bin smoke typecheck error is the current-main shard-never-ran Error.code issue, unrelated to this PR. Hosted shard 2 also failed independently at smoke:web-channel-authority with ReferenceError staleNow is not defined. Please exact-head re-review 1fe2adaf.

davidfarah2003:
BLOCK c06f5657

Graded at exact head c06f5657 in my own detached worktree. I did not run `cotal up`, `down`, `setup`, `ext seed --reset`, any stack lifecycle command, `pnpm smoke:ci`, `pnpm check`, or any suite whose name ends `-live` or `:live`. I graded by reading the tree and hosted CI, and by running only the broker-free `pnpm smoke:live-job-conclusion` (the name does not match the ban). That is why a real GitHub 25-minute live cutoff is a named unverified gap below, not a silent pass.

## Blocker

`bin/smoke/ci-suites.txt` inserts `smoke:live-job-conclusion` immediately before the pre-existing tail suite `smoke:workspace-import-exports`. The file's own rule is append-at-end so shard membership of every earlier line stays put. Hosted CI on this SHA already refused it:

- run 33289497440, job `unit`, step `Smoke chain shard stability`
- `a6891468 -> a860a115  @4->4 shards`
- `suites: 421 -> 422 · added 1 · removed 0`
- `pre-existing suites CHANGING SHARD: 1 of 421`
- `first few: smoke:workspace-import-exports`
- `shard-stability: RESHARD=1`

Same file also gains an extra blank line mid-file next to `smoke:lang-engine` with no suite added there.

Until that registration is an append (or a fragment that does not move a pre-existing line), this SHA is not landable. `ci-ok` then failed the Gate on `unit=failure`; that is this re-shard, not a live timeout.

## Classifier (read + CI, not a live hang)

`scripts/live-job-conclusion.mjs` names three cancelled outcomes (`self-timeout`, `superseded`, `unexplained-cancellation`) and leaves a non-cancelled result as `not-cancelled`. `ci-ok` (`if: always()`, `needs: [unit, smoke, live]`) calls it with `--timeout-seconds 1500` matching `live.timeout-minutes: 25`, then the existing Gate still requires `live=success`, so a hang cannot badge as a pass.

Discrimination observed on this SHA's hosted run: `live` was `success`, classify printed `live result is success; no cancellation diagnosis needed`, and it did not emit a timeout error. The broker-free suite is 12/12 on constructed cases, including an injected 1516s attempt-scoped jobs payload as `self-timeout` and a newer same-PR run as `superseded`.

## Named gap

No hosted job on c06f5657 cancelled `live` at the 25-minute backstop (`live` succeeded). So this SHA does not prove that GitHub's checks UI renders the `::error title=Live job did not finish::` annotation, or that a real cutoff duration lands on `>= 1500` rather than a slightly-early cancel classified as `unexplained-cancellation` (still red, but not the timeout name). A PR cannot force its own runner to self-timeout and still keep the result under test. That gap is expected; it is not the blocker. The re-shard is.

davidfarah2003:
CI classification on the refolded tip 6fb00040 (effective-terminal; only the ci-ok aggregate pending): the single red is `smoke (shard 0/4)`, which dies at `smoke:mutation-fixtures` — 2 dead anchors, reproduced on plain main locally (they broke when #1153's merge moved the text two fixtures anchored; cure is #1161, in CI now). Main-side, nothing from this branch; every other lane including the branch's own live-timeout surfaces is green. Merging.

PR #1044, opened 2026-08-30
fix(auth): finish a retirement whose head terminal never landed at the next boot

### LINKED ISSUE #878: auth: boot crash-resume derives retirement owed-ness from the gate alone, so a crash between gate and head transitions wedges the alias permanently
Found by source review during a reliability-plan review panel. **Not reproduced** — filing so it is not lost, per the convention used on #871.

## The gap

`implementations/auth/src/service.ts` → `resumeOpenOperations()` is the auth plane's boot crash-resume. It enumerates durable operation intents and finishes every barrier the executor owes. The executor for retirement (`resumeAgentRetirement`) exists and works; this is not a missing-verb issue.

The problem is the owed-ness predicate, verbatim:

```ts
if (gate === undefined || gate.row.state !== "frozen" || gate.row.op?.opId !== it.opId) continue;
```

Owed-ness is derived **only from the gate** (`frozen`, by this exact `opId`). But retirement completion is a **cross-object invariant**: it spans the gate *and* the lifecycle alias head (SPEC §13.1).

## The crash window

A crash after `gate frozen → retired` but before `head retiring → retired` leaves the alias permanently wedged:

- the head is `retiring`, which per §13.1 is **non-current AND not replaceable**
- the gate is no longer `frozen`, so `resumeOpenOperations()` **skips the intent** and never resumes it — on this boot or any future boot
- `reconcile-gate` refuses with `wrong-op-kind` (correctly — it owns registration/`epgate`, not lifecycle retirement)

`resumeAgentRetirement()` is capable of finishing that tail. Nothing ever calls it for this state.

The function's docstring documents the skip as intentional — *"completed and lost operations leave their intent behind by design and are skipped"* — which is right for completed operations and wrong for this one, and the predicate cannot tell them apart.

## Why it stays invisible

The skip emits nothing. No log line, no warning. The operator sees a clean auth-plane boot next to an alias that will never mint and can never be replaced, with no artifact connecting the two.

## Suggested direction

Derive owed-ness from **both** the gate and the alias head, and resume the same intent across every nonterminal cross-bucket combination, rather than adding a parallel executor. The existing barrier already works from the durable intent's coordinates, so the resume path itself should not need to change.

Separately, the skip decision is worth recording as a durable structured fact — `(alias, op.kind, opId, gate state at decision, ts, reason)` — rather than a log line or an alarm. As data it can be joined against a nonterminal head at any surface that already reads the head (spawn refusal, status), which is what makes the wedge legible at the moment someone hits it. As an alarm it would cry wolf on every legitimately completed operation.

## Not this issue

This is **not** #871. #871 is a lost manager startup race freezing the **endpoint issuance** gate (`epgate`, registration kind) on a user-auth mesh. The two were conflated during review; they are disjoint gate families per §13.1.

## Repro needed

A deliberate crash injected between the gate transition and the head transition during retirement should produce the state. That has not been run.

davidfarah2003:
## Scope

Fixes #878 end to end, including the refined two-trigger acceptance criteria from the issue's final comment.

### The defect, reproduced live on current main

A crash between the retirement barrier's last two durable writes — the gate terminal (step 8) and the head terminal (step 9) — leaves `gate retired by the op` + `head retiring under the op`. The head is non-current and not replaceable (SPEC 13.1), and the boot crash-resume's owed-ness predicate read the gate alone, so it skipped that intent on that boot and every future boot: the alias could never mint and could never be replaced, with a clean-looking boot log.

Reproduced deterministically (staged, not a load test): the barrier runs for real through intent create, gate freeze, head `active → retiring`, and family containment, then aborts fail-closed at verified eviction; the frontier record (step 7) and gate terminal (step 8) are replayed with the barrier's own primitives; step 9 never runs. On unmodified main: the next boot resumes nothing, the head stays `retiring`, and a same-name respawn is refused with the SPEC 13.1 not-replaceable message.

### Trigger B measured before fixing

The issue's open question — does a same-op re-request through the retire-lifecycle rail repair the state? — was measured live on main: **yes**. `auth-admin`'s same-op re-request falls through its idempotence guards (the head is `retiring`, not `retired`) into `runAgentRetirementBarrier`, whose gate-retired branch finishes the head tail and answers `retired: true`. So the rail was never broken; the **boot path was the dead letter**, and the fix is the owed-ness predicate (per the issue's suggested direction), not a new executor.

### The fix

- `resumeOpenOperations` derives retirement owed-ness from the gate AND the alias head: a retirement whose gate terminal landed by its own op is owed exactly while that op's uid still owns a non-terminal head. A retired head at the uid and a successor's head at another uid are the completed cells and stay skipped (a successor is never touched). Takeover owed-ness is unchanged.
- `enumerateOperationIntents` now parses and carries the intent's own `owner`/`actor` so the boot reads the head under the right alias.
- The resume re-enters the SAME barrier, whose existing gate-retired branch finishes ONLY the head terminal from the durable coordinates — nothing past the gate terminal is re-revoked, re-evicted, or re-drained (the standing fix constraint; asserted in the smoke via the resume's zero counts).
- Absent-head resumes into the barrier and fails loud there (a head is never deleted; the barrier refuses corruption) instead of skipping silently.

### Proof

`pnpm smoke:retirement-wedge:auth` (new, self-contained broker, killed by exact PID, ~4s) drives the issue's refined acceptance cell:

- stages the crash window and asserts the intermediate state (gate retired by the op, head retiring, intent + frontier durable) before any trigger;
- shows the wedge's operator-visible consequence (same-name respawn refused);
- **trigger A**: the next auth boot over the real plane resumes the intent, lands the head `retired`, logs the resume with the op id, and the resume is tail-only;
- **trigger B**: a same-op re-request through the REAL rail (real minted requester credential, real broker ACLs) over an independently staged second wedge answers `retired: true`;
- both converge on the same executor; a stranger's opId is refused (`not-found`); the completed cell (a successor's active head) stays skipped on a later boot; the freed alias mints a successor through the plane's public issuance seam.

Mutation-proofed (`node scripts/mutation-proof.mjs --config implementations/auth/smoke/mutations/retirement-wedge-resume.json`), both mutants KILLED red-and-named:

- M1: the widened predicate's gate-state test never matches (the reverted fix in behavior) → red on "TRIGGER A REPAIRS";
- M2: the enumeration lies about the intent's alias → red on "the COMPLETED cell stays skipped".

The suite injects a deterministic plane-claim liveness oracle (its throwaway broker has no delivery daemon; consecutive plane boots otherwise race the claim release and fail safe on the unreachable rail). Chained at the end of `bin/smoke/ci-suites.txt`; `smoke:gate-inventory` green.

### Not proved / named gaps

- No live whole-stack crash was performed (the box hosts the live mesh; `cotal up`/`down` are forbidden here). The crash window is staged by replaying the two durable writes with the barrier's own sealed-registry primitives — byte-identical durable state, not a SIGKILL race.
- The `active`-at-this-uid resume cell (impossible backward head) is reasoned about, not asserted: no code path produces it.
- CI on this PR is the remaining unobserved gate.

### Green locally

`smoke:retirement-wedge:auth` (16/16, 5 consecutive runs), `smoke:retirement-barrier:auth` (61/61), `smoke:barrier-plane:auth` (60/60), `smoke:auth-admin:auth` (27/27), `smoke:retirement-cleaner:auth` (35/35), `smoke:drain-repair:auth` (45/45), `smoke:gate-inventory`, `pnpm typecheck`, `pnpm changeset status`. Changeset and docs (`docs/identity-and-auth.md`) in the same change.

Closes #878.


davidfarah2003 (commented):

- bin/smoke/ci-suites.txt: the new `smoke:retirement-wedge:auth` block is inserted immediately after `smoke:workflow-concurrency` — but that is NOT the tail. Confirmed against current main (698 lines, base blob 0758581b5 matches this PR's diff base): the pre-existing "# Most smoke trees are still outside their package TypeScript projects..." comment and `smoke:workspace-import-exports` entry follow it. This is a mid-file insert, which silently reshuffles `index % shardCount` for every suite after the insertion point — exactly the hazard rule (c) forbids and that PR #1052 (also in this queue) exists to prevent.
- The PR body's own Proof section claims "Chained at the end of `bin/smoke/ci-suites.txt`" — directly contradicted by the diff.
- The actual fix looks sound otherwise: `resumeOpenOperations` reads gate AND head cross-object owed-ness, `enumerateOperationIntents` now carries owner/actor, mutation-proofed (M1 reverts the predicate, M2 lies about the alias — both kill named cells), and docs/identity-and-auth.md is updated to match.
- Needs the ci-suites.txt entry moved to the actual absolute tail before this can land.

davidfarah2003:

Issue #686, opened 2026-08-19
manager: persist and surface each seat's resolved model pin for external attestation


davidfarah2003:
A review-seat audit found that a managed pi seat's model pin cannot be verified from *outside* the seat's own process on a running mesh host, without first solving a session-discovery problem:

- the seat's argv is the single word `pi` (no `--model`); the pin lives in `PI_PROVIDER` / `PI_MODEL` / `PI_REASONING_LEVEL` in its environment
- reading another process's environ is blocked (ptrace scope) even for the same user
- `cotal ps` reports name, harness and runtime but neither the model pin nor the pid
- neither the registry nor manager state persists the spawn spec after the spawn returns

The pin is not entirely unreadable: the harness writes it into the opening records of its own session file (`model_change` / `thinking_level_change`). So an external party who knows which host the seat runs on and which file is its session can attest the pin without trusting the seat. That is a discovery problem, not a permissions one, and it is solvable today with no code change: mint a per-seat token, require it in the liveness reply, and that token then identifies the seat's session file. The audit that surfaced this used exactly that: a per-seat liveness nonce found in the session file bound the pin to the seat and to the object under review.

Consequence when this bites: the only attestation legs are the seat self-reporting its own env, or the caller's record of what it requested. Neither proves what the runtime applied. This gap let a two-seat review panel run same-model while its operator believed it was cross-vendor; the mismatch surfaced only when verdicts were required to carry self-inspected pins.

Wanted (durable fix, superseding the token interim):
1. The manager records each seat's resolved spawn spec (provider, model, reasoning level, connector) in its durable state at spawn time.
2. `cotal ps` (or a per-seat detail view) surfaces that recorded pin, so an operator can verify panel composition without entering the seat's process and without knowing which host holds the session file.

davidfarah2003:
Correction to one detail in the description, from the person who reported it: the `COTAL_MODEL`
asymmetry is not real and item 3 rests on a false premise.

It came from a partial first read of one process environment. The completed read finds
`COTAL_MODEL` present and correct in both processes:

- `COTAL_MODEL=cliproxy/grok-4.6:high` alongside `PI_PROVIDER=cliproxy`, `PI_MODEL=grok-4.6`,
  `PI_REASONING_LEVEL=high`
- `COTAL_MODEL=cliproxy/gpt-5.6-sol:xhigh` alongside `PI_PROVIDER=cliproxy`,
[...]

Issue #891, opened 2026-08-26
CotalEndpoint emits "error" on recoverable retry paths, so a consumer without a listener is killed by a condition the endpoint intended to survive


davidfarah2003:
`CotalEndpoint extends EventEmitter` (`packages/core/src/endpoint.ts:314`) and emits `"error"` from
at least ten sites. Node rethrows an emitted `"error"` when the emitter has no listener, so any
consumer that constructs an endpoint and does not attach one dies on the first emission.

Nothing in the endpoint installs a default listener. `this.on(` does not appear in `endpoint.ts` at
all, and no page under `docs/` states the requirement.

## What makes this more than an ordinary EventEmitter footgun

Several of these emissions are on **retry** paths, and their own text says so:

```ts
// endpoint.ts:668
this.emit("error", new Error(`bearer refresh failed (…) - retrying; this connection dies at its
  current token's expiry if the auth service stays down`));

// endpoint.ts:740, :818
this.emit("error", new Error(`creds refresh failed (…) - retrying; this connection dies at its
  current JWT's expiry if renewal keeps failing`));
```

These describe a recoverable, informational condition: the endpoint is still running and is going to
try again. It reads as a warning. In a consumer with no listener it is fatal. A transient auth blip
or broker hiccup, the exact thing the retry exists to absorb, kills the host process instead.

Others fire from detached async paths (`:988` presence publish, `:1151`, `:1542`, `:1605`
`.catch(e => this.emit("error", e))`, `:2215` membership watch closed), so the throw surfaces on a
later tick with no relation to whatever the caller was doing. The stack points at the emit, not at
the cause.

## The tell: first-party code already works around it, silently

Three sites in `implementations/auth` attach an empty listener whose only effect is to stop the
process from dying:

```
implementations/auth/src/commands.ts:165     ep.on("error", () => {});
implementations/auth/src/plane-claim.ts:333  ep.on("error", () => {});
implementations/auth/src/barrier-evict.ts:74 ep.on("error", () => {});
```

Others attach a real one (`implementations/web/src/web.ts:799`, `implementations/cli/src/render.ts:62`,
`implementations/cli/src/view/mesh-view.ts:188`). So every first-party consumer has already learned
this, and three of them learned it well enough to swallow the event entirely rather than handle it.
An `() => {}` listener is not error handling; it is a guard against a crash, and its presence is
evidence the hazard is known in practice and unwritten anywhere.

[...]

PR #1049, opened 2026-08-30
fix(connector-core): refuse a cotal_spawn model pin the manager did not record

### LINKED ISSUE #972: cotal_spawn accepts a model parameter and silently discards it, so a review panel can be same-model while reporting as cross-vendor
`cotal_spawn` declares a `model` parameter. Passing it is accepted with no error, no warning, and no
effect: the seat comes up on whatever the host would have chosen anyway.

This is the silent half of #736. That issue covers the cotal-lang `spawn` primitive, where `model` is
**rejected loudly** by the validator (`L3011 Unknown option key`). Here the same intent is accepted and
discarded, which is worse, because a refusal tells you your intent did not land and silence does not.

## Measured

Two spawns in one session, both passing `model: "grok-4.6"`, neither honoured:

| requested | seat reported |
|---|---|
| `grok-4.6` | `GPT-5.5` |
| `grok-4.6` | `claude-opus-4.6` |

Both seats were asked directly what they were running on. One added: "Env has no `JCODE_MODEL`/`MODEL`
override. Orientation also does not name a model." So the value is not reaching the launched process by
any route the seat can see, and the seat itself cannot tell that a pin was requested.

Neither spawn produced an error. Both returned the ordinary wait-timeout message that SPEC 13.6 says is
a timeout on the wait rather than evidence of failure, and both seats did come up. The model parameter
simply evaporated.

## Why this is worth more than a papercut

The review process on this repo depends on **deliberate model diversity**. A panel exists so that two
readings of the same head are not correlated by construction, and the whole value of a second verdict
is that it comes from a different model. When `model` is silently dropped, an operator who carefully
routes lens A to one vendor and lens B to another gets two seats on whatever the default is, believes
they have cross-vendor confirmation, and does not.

That is a false-independence failure, and it has the shape this repo treats as the serious class: the
broken version and the working version are indistinguishable from the outside. Nothing in the roster,
the spawn result, or the seat's own orientation reveals which model was actually used, so the only way
to find out is to ask each seat and hope it can introspect. Both seats here could. A seat that cannot
would leave the operator with a confident and wrong belief.

In this session it meant two reviews that were reported as cross-vendor were in fact same-family with
the coordinator, and that was only discovered by asking.

## Ask

1. Honour `model` in `cotal_spawn`, or
2. refuse the spawn when a requested `model` cannot be applied, naming the reason.

Either is fine. What is not fine is accepting the parameter and discarding it. Same reasoning as #736's:
if the field is reserved for host-side resolution by design, then the tool should not advertise it, and
one comment plus a refusal is the fix rather than a plumbing change.

Independently of which is chosen, the **resolved** model should be visible without interrogating the
seat. #686 and #905 already cover surfacing a seat's resolved model pin; this issue is the reason that
matters rather than a nicety, because without it there is no way to audit whether a requested pin took
effect.

Accepted when a spawn with `model: X` either produces a seat running X, or fails with a message naming
why X could not be applied, and when the resolved model is readable without asking the seat.

Refuted if `model` is deliberately advisory and host-overridable, in which case the tool's parameter
description should say so and the two rows in the table above are correct behaviour. Either outcome
closes it; silence leaves an operator able to believe in independence they do not have.

Refs #736, #686, #905, #557.

davidfarah2003:
Closes #972.

`cotal_spawn` advertised `model` and forwarded it to the manager, then reported success without naming the recorded pin. A seat that came up on the harness default was indistinguishable from a pin that landed, so a review panel could be same-model while reported as cross-vendor.

## What changed

- After a spawn that requested `model`, `MeshAgent.spawn` inspects the manager's recorded pin and refuses when it is absent or different.
- The spawn tool result names the recorded model on success.
- Orientation always states the recorded pin (`COTAL_MODEL` / agent-file `model:`) or that none was recorded.
- Empty `model` is refused before invoke.

## Proof

- Live connector repro on current main (`ce95d2ac`): `claudeConnector.buildLaunch({ model: "grok-4.6" })` accepted and set `COTAL_MODEL=grok-4.6`; `xai/grok-4.6` already refused. The silent class is a pin that is accepted then not attested.
- `pnpm smoke:spawn-args` (local tsx): 25 cells, including the three #972 refusals/naming cells.
- `pnpm smoke:orientation`: identity asserts the pin and its absence.
- `node scripts/mutation-proof.mjs --config extensions/connector-core/smoke/mutations/spawn-model-pin.json`: A1/A2/A3 KILLED on the named cells.
- Kickoff fixture still KILLED after the spawn-args rewrite.

SHA: `3ba087b09c89886cadd23cd8d27ef91fb6639ae2`

## What this does NOT prove

- Did not spawn a live seat on this box (`cotal up` / `--detach` / `--supervise` are forbidden here). The inspect attestation is proven against a recording `ep`, not a live manager/harness.
- Does not prove the harness actually ran the recorded model. Inspect is the manager's launch pin, not `getRuntimeInfo` / provider identity. A connector that records `grok-4.6` then serves something else is still a named gap (#686 / connector-specific checks).
- Does not honour an unrecorded pin. Refusal is the chosen #972 outcome, not plumbing a discarded value into a host that never received it.
- Does not change `cotal_roster`. Presence still omits model unless the host published `card.meta.model`.
- `ci-suites.txt` is frozen until #880; this does not add a suite row. `smoke:spawn-args` is already in the gate.
- Mutation-proof ran with `--allow-dirty` because this worktree has local `node_modules` symlinks. Recovery of mutated sources was git checkout of the committed files.


davidfarah2003:
Follow-up from the assignment's extra measured datum (CLI spawn that died on a variant-tier refusal recorded `deepseek-v4-pro` despite `--model grok-4.6`).

On this head (`3ba087b09c89886cadd23cd8d27ef91fb6639ae2`) the jcode host does:

```ts
if (config.model) await client.setModel(sessionId, config.model);
if (config.variant) {
  const model = (await client.getRuntimeInfo(sessionId)).model ?? config.model;
  throw jcodeEffortRefusal(..., model ?? \"(the provider default)\");
}
```

The effort-refusal diagnostic's `effective model` is therefore `getRuntimeInfo().model` (session/provider default), not the requested pin. Existing host smoke asserts `effective model \"fake-model\"` while `COTAL_MODEL` is also `fake-model`, so it cannot see this mismatch. The later mislabel check (`runtime.model !== config.model`) only runs after a successful effort apply.

This is adjacent diagnostics, not the MCP silent-drop #972 closed. Not patched in this PR: proving it live needs `smoke:jcode-host` (filter build), which this box should not run in parallel with other work, and the assignment asked for a tight PR.

davidfarah2003:
Follow-up landed on this PR: `97af400dd68018a75dcd75a24eaa1a9b535cd8ee`.

The extra datum is real and patched: a Jcode variant-tier refusal named `getRuntimeInfo().model` (session default) rather than the requested pin. Helper `effortRefusalModel` now prefers the pin. `smoke:jcode-effort-model` 6/6; mutations A1/A2 KILLED.

Still not proven: `host.ts` call site is not executed by that suite; `smoke:jcode-host` still uses the same fake-model for pin and RuntimeInfo and was not run here. Inspect attestation remains a recording `ep`, not a live spawn.

davidfarah2003:
Journal-field instrument landed on `cec8bb9521f0b691a72fa0a7aafc483ea515f10a`.

First-hand tonight: `netcup-w972-56f272092067` session journal `model` is `deepseek-v4-pro`, not grok-4.6.

Fake host now writes a session journal (default `deepseek-v4-pro`, updated by `set_model`). `smoke:jcode-host` asserts the journal records the requested pin (`fake-model`) and not the harness default. Mutation A1 (skip `setModel`) KILLED red-and-named on that cell. Baseline 43 checks.

Still not proven: a live MCP `cotal_spawn` on this box (stack lifecycle forbidden). Manager inspect attestation is a recording `ep`. Production broker PID 2168717 was not touched; leftover smoke nats PIDs from this suite were killed by pid.

davidfarah2003:
Interpretation check against the journal instrument, on `cec8bb9521f0b691a72fa0a7aafc483ea515f10a`.

Guess I made before the orch brief: manager `inspect.model` was enough to prove the pin landed. That is wrong for tonight's w972. Inspect records what spawn requested. The journal recorded `deepseek-v4-pro`. The inspect check in `MeshAgent.spawn` would have PASSED for that seat.

What the PR actually proves on the journal field is the fake-host path: skipping `setModel` is KILLED against `session journal records the requested model pin`. It does not prove a live MCP `cotal_spawn` writes that journal field. That live spawn is still forbidden on this box.

davidfarah2003:
Refolded onto current main as 1409a952. Two conflicts needed hand resolution beyond the standard set — documenting the unions for review:

- **`src/host.ts`**: only the `startup-diagnostics.js` import statement collided (this branch adds `effortRefusalModel`; main's diagnostics rework made the import multiline). Union keeps main's full list plus `effortRefusalModel`. The branch's variant-block change (prefer the requested pin over RuntimeInfo in the effort-refusal message) auto-merged untouched.
- **`smoke/fake-jcode.mjs`**: this branch's journal/model plumbing composed with main's attach-refusal and set-model-refusal hooks. In the united `set_model` case, main's `FAKE_JCODE_REFUSE_MODEL` gate comes first; the non-refused path applies this branch's model update + journal write before `reply ok` — a refused model does not update the journal.

Verification on the fold: `pnpm -r build` green, standard gates green. `smoke:jcode-host` fails one cell here — `model_mismatch names the connector-owned refusal` after 41 ✓ — but the identical cell fails on plain main on this machine (and on CI runners per #1143's run) while passing on a linux lane host: environment-sensitive main-side cell, recorded on #1155, not attributable to this branch or the union.

davidfarah2003:
CI classification, final (run 33322641649 on the repaired fold e5886166):

| shard | signature | class |
|---|---|---|
| 0/4 | `smoke:expiry-renewal:auth` — cells `the refusal to dial is loud and names the missing renewal path` + `the wait between reconnect attempts grows past the flat retry interval` (refusals=0, gaps=[]) | reproduced on pure main locally, identical cells — the no-renewal-path notice moved off the error channel (#891/#1157 consumer-sweep class, #1156); suite-side cure PR follows |
| 2/4 | `smoke:presence-render-census` — `manifest anchor must exist exactly once in StatusBar.tsx: {status.error ? (` | reproduced on pure main locally, identical assertion; cure PR follows |

Both reds are main-side; this branch's own earlier red (jcode-journal-model dead anchor) was fixed in the fold and its suite now passes. Merging per the reproduced-on-pure-main rule.

PR #1050, opened 2026-08-30
fix: validate delivery tenancy before lease admission

### LINKED ISSUE #856: reconcile-gate: the frozen-gate repair is unusable in production (sequential WAN ledger, oracle that dies with the failure, misattributed errors)
## What happened

The `main` mesh was down for roughly four hours on 2026-08-24/25. The manager died from the known broker-write crash class (#472), which froze its issuance gate — the expected outcome, and `reconcile-gate` is the shipped repair for it. The repair could not complete for four hours. Every part of that delay is a design property of the recovery path, not bad luck, so this issue is about the repair machinery rather than the original crash.

David's verdict on it, verbatim, so the intent is not lost in translation: *"this is horribly designed. horrible. it should be completely replaced however the fuck its implemented."*

## Findings, in the order they blocked the repair

**1. The oracle is a separate process that the failure itself takes down, and it is unique per space.**
`reconcile-gate` refuses to infer death from silence (correct, SPEC 13.1) and so needs the delivery daemon as its liveness oracle. The daemon holds a single shard-0 lease per space (`cotal_delivery_<space>`, key `lease.0`). If that lease is held by *anything* that cannot serve the `ctl.delivery-admin` rail, the repair cannot run and cannot say why. Tonight the lease was held by a daemon on the broker VPS started from the wrong working directory: `~/mesh/start-deliver.sh` does `cd "$HOME/mesh"` and then runs `pnpm --dir "$HOME/Cotal" exec tsx …`, and `pnpm --dir` resets the process cwd, so the daemon resolved its workspace to `/home/cotal/.cotal` (the **growth** mesh, account `ADSOV2ZV…`) while authenticating with `/home/cotal/mesh/.cotal/delivery.creds` (the **main** account, `AATEPS2J…`). It refused every query — correctly, and with an excellent message — but it kept renewing the lease every 15 s, and a `*/5` cron resurrected it whenever it was killed. Every real daemon that tried to bind got *"a live lease already exists for shard 0 — another delivery daemon is running. Not binding."*

**2. The error the operator sees names the symptom, not the cause.**
From the CLI the failure reads: *"the delivery daemon is not reachable on the ctl.delivery-admin rail (timeout) … Start the delivery daemon (`cotal up` runs it) and retry."* A daemon **was** running; `cotal up` had been run repeatedly and exited quietly because the lease was taken. The message sent the operator to the wrong machine for hours. The refusing daemon knew the exact reason (account mismatch, with both account IDs) and that reason never reached the operator — it only appeared later, in a different message, when the impostor happened to answer instead of time out. The rail timeout should surface the current lease holder (principal, `since`, and whether it answers a health ping) rather than advising a restart that silently no-ops.

**3. The eviction ledger is sequential, and the operator is usually not next to the broker.**
`implementations/manager/src/reconcile-gate.ts:154` revokes ledger rows one at a time, and `:167` evicts each distinct holder one at a time, each `barrier.evict()` being a round trip that waits on a broker CONNZ sweep. Tonight's family was **250 rows**. Run from the box (124 ms RTT to the broker) a full pass took **over 8 minutes** and got through 193 evictions before the link dropped — `✗ closed connection` — losing the pass. Re-running from the broker itself made the same work finish in seconds. Two fixes, both real:
- batch the ledger: revoke and evict in bounded-concurrency batches, or ideally one sweep for the whole family compared against the holder set, instead of 250 independent proofs of the same fact;
- make the pass resumable and idempotent so a dropped connection costs the remaining rows, not all of them.

**4. Nothing bounds the repair, and the watchdog silently truncated it.**
`~/start-manager.sh` on the box wraps the self-heal in `timeout 120`. A 250-row pass over the WAN cannot finish in 120 s, so the keeper's automatic repair had been killed mid-ledger on every firing since the crash — for hours, with no log line saying it was cut off (exit 124 is not reported). Either the repair must be fast enough that 120 s is generous (see 3), or the CLI must fail fast with a *"this will take N minutes at this latency"* estimate instead of silently starting work it cannot finish.

**5. Auth-mutating cron jobs on a second machine.**
The broker VPS also ran `43 */6 * * * cotal doctor auth --fix` and `23 */6 * * * renew-delivery-cred.sh`. Two machines re-minting the same mesh's credentials on independent timers is how credential generations skew, which is bug #773's precondition. Both are now disabled. A mesh should have exactly one credential authority, and a second machine running `doctor auth --fix` against a shared broker should refuse rather than re-mint.

## What I want from the redesign

- Liveness that does not depend on a singleton process the failure can take out — a broker-side fact, or a quorum, or a lease whose holder must prove it can serve the rail before it is allowed to keep it (a holder that fails a health ping loses the lease).
- A repair whose cost is bounded by the *shape* of the family, not by `family_size × RTT`.
- Errors that name the blocking entity (who holds the lease, since when, which account it authenticates as) rather than the symptom.
- One credential authority per mesh, enforced.

## Reproduction

Point a delivery daemon at a space using credentials from account A while its resolved workspace root belongs to account B, keep it alive with a cron, then crash a manager in that space and try `cotal reconcile-gate`. The repair is unavailable indefinitely and the CLI blames the rail.

## Related

#472 (broker WriteDeadline kills the manager), #368 (poisoned shard-0 lease after a young daemon death), #773 (credential generation skew), #825 (seat survival on lease loss — merged, shipped in 0.30.0), #783 (manager hot reload, still open).

davidfarah2003:
Closes #856.

## Scope
- validate the delivery daemon scan root and account binding before endpoint start and `lease.0` acquisition
- add a public two-account, real-process regression proving a wrong-root daemon exits before admission, the correctly-rooted daemon acquires immediately, and `reconcile-gate` is usable

## Security consequence
Before this change, a daemon authenticated as account A but rooted in account B could acquire and renew account A's singleton delivery lease despite having no valid scan-tenancy claim. It then refused every delivery-admin operation on the account mismatch while denying the correctly-rooted daemon the lease, leaving frozen-gate recovery unavailable indefinitely.

## Proof
- main reproduced the poisoned lease against a real two-account ephemeral broker
- `pnpm typecheck`
- `pnpm check:docsbundle`
- `pnpm smoke:gate-reconcile-cli:auth` (18/18)
- mutation proof restores admission before validation and KILLS the named poisoned-lease cell red-and-named

## Limits and deferred work
- This fix is independent of #873: it changes delivery startup admission and consumes no blocked-coordinate shape. The existing `wrong-op-kind` fence is unchanged; lifecycle retirement remains #878.
- Large-family bounded repair is deferred to #1061.
- Lease-blocker diagnostics are deferred to #1062.
- Single credential renewal authority is deferred to #1063.
- Interrupted-pass resumability is deferred to #1064.
- The separate scratch-cwd smoke resolution defect is #1065.
- No forbidden live-mesh `cotal up`/`cotal down`, `smoke:ci`, `check`, or `-live`/`:live` suite was run. The reproduction used a standalone ephemeral broker; clustered absence behavior was not exercised.

### DIFF
--- a/.changeset/steady-gate-recovery.md
+++ b/.changeset/steady-gate-recovery.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/delivery": patch
+---
+
+Prevent a wrong-root delivery daemon from acquiring and renewing the singleton lease before it validates that its scan credentials belong to the account it serves.
--- a/bin/smoke/mutations/reconcile-poisoned-lease.json
+++ b/bin/smoke/mutations/reconcile-poisoned-lease.json
@@ -2,7 +2,6 @@
   "suite": "implementations/manager/smoke/gate-reconcile-cli-e2e.smoke.ts",
   "guard": "a wrong-root delivery daemon validates its scan tenancy before acquiring the singleton lease",
   "command": "pnpm --filter @cotal-ai/delivery build && pnpm smoke:gate-reconcile-cli:auth",
-  "completionMarker": "GATE-RECONCILE CLI E2E",
   "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/reconcile-poisoned-lease.json",
   "why": [
     "The suite starts real delivery daemon processes against a two-account ephemeral broker.",
--- a/bin/smoke/mutations/reconcile-poisoned-lease.json
+++ b/bin/smoke/mutations/reconcile-poisoned-lease.json
@@ -1,0 +1,22 @@
+{
+  "suite": "implementations/manager/smoke/gate-reconcile-cli-e2e.smoke.ts",
+  "guard": "a wrong-root delivery daemon validates its scan tenancy before acquiring the singleton lease",
+  "command": "pnpm --filter @cotal-ai/delivery build && pnpm smoke:gate-reconcile-cli:auth",
+  "completionMarker": "GATE-RECONCILE CLI E2E",
+  "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/reconcile-poisoned-lease.json",
+  "why": [
+    "The suite starts real delivery daemon processes against a two-account ephemeral broker.",
+    "The mutation removes the shipped startup preflight. The wrong-root daemon then holds lease.0 and the named cell reddens because it did not exit before admission.",
+    "The command builds @cotal-ai/delivery first because the public binary imports its package dist."
+  ],
+  "mutations": [
+    {
+      "name": "restore lease admission before scan-tenancy validation",
+      "file": "implementations/delivery/src/delivery.ts",
+      "find": "  validateScanTarget(scanTarget);",
+      "replace": "  void scanTarget;",
+      "expectRed": "POISONED LEASE: account A with account B root exits before holding lease.0",
+      "cell": "POISONED LEASE: account A with account B root exits before holding lease.0"
+    }
+  ]
+}
--- a/bin/smoke/mutations/reconcile-poisoned-lease.json
+++ b/bin/smoke/mutations/reconcile-poisoned-lease.json
@@ -1,7 +1,7 @@
 {
   "suite": "implementations/manager/smoke/gate-reconcile-cli-e2e.smoke.ts",
   "guard": "a wrong-root delivery daemon validates its scan tenancy before acquiring the singleton lease",
-  "command": "pnpm --filter @cotal-ai/delivery build && pnpm smoke:gate-reconcile-cli:auth",
+  "command": "pnpm --filter @cotal-ai/core build && pnpm --filter @cotal-ai/manager build && pnpm --filter @cotal-ai/delivery build && pnpm smoke:gate-reconcile-cli:auth",
   "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/reconcile-poisoned-lease.json",
   "why": [
     "The suite starts real delivery daemon processes against a two-account ephemeral broker.",
--- a/implementations/delivery/src/delivery.ts
+++ b/implementations/delivery/src/delivery.ts
@@ -15,7 +15,7 @@ import {
 } from "@cotal-ai/core";
 import { DELIVERY_CREDS_KIND, FsSecretStore, authDir, deliveryCredsKey, findCotalRoot, loadSpaceAuth, segmentedKey, soleSpaceOf, workspaceSecretStore } from "@cotal-ai/workspace";
 import { startMembership } from "./membership.js";
-import { executeEviction, executePlaneLiveness, executePrincipalLiveness, type ScanTarget } from "./evict-exec.js";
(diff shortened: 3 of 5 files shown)


davidfarah2003 (commented):

- `validateScanTarget` (renamed+exported from the former `resolveScan`) is called in implementations/delivery/src/delivery.ts before `CotalEndpoint` construction — i.e. before any lease acquisition — matching the body's claim to validate tenancy "before endpoint start and lease.0 acquisition."
- New two-account, real-process regression in gate-reconcile-cli-e2e-smoke.ts reproduces the poisoned lease pre-fix (wrong-root daemon must exit before holding lease.0) and proves the correctly-rooted daemon then acquires and `reconcile-gate` completes.
- Mutation proof (reconcile-poisoned-lease.json) reverts the added `validateScanTarget` call to a no-op and kills the named cell "POISONED LEASE: account A with account B root exits before holding lease.0" red.
- No ci-suites.txt change needed (extends an already-registered suite); changeset present; "Limits and deferred work" section is candid about follow-on issues (#1061-#1065) instead of silently absorbing them.

davidfarah2003:
Posting on behalf of reviewer seat `w983`, whose own comment was blocked by GitHub's secondary rate limiter. Verdict and evidence are its work, unaltered.

BLOCK at exact head `ac7dc548381b25176503838d76a496033b52cccb`.

The workstation wrong-root path is fixed, but the hosted/injected-store admission path still acquires the singleton lease without validating the scan credential's account tenancy.

Ordering in `runDelivery` is correct for the check it performs: `validateScanTarget(scanTarget)` runs before `new CotalEndpoint`, endpoint start, and `acquireDeliveryLease`. The problem is that `validateScanTarget` validates only root stability and an optional workstation `membership.json` account. When `source.injected` is true, it skips that disk check and does not read `membership-observer.creds` or call `observerTenancyProblem`; it simply returns `expectedAccount`, which came from the delivery credential.

I exercised that exact branch with an injected SecretStore whose observer credential value was foreign. `validateScanTarget` returned `{accountId:"ACCOUNT_A"}` with `secretReads:0`. A hosted composition can therefore have delivery credentials for account A and $SYS scan credentials for account B, pass startup admission, acquire `lease.0`, and only fail later when a delivery-admin request calls `loadCheckedSys`. That is the same poisoned-lease denial of service, one composition arm over.

Required repair: startup admission must load the required observer credential through `target.source`, validate it against `expectedAccount`, and fail before endpoint construction/lease on both workstation and injected paths. If startup requires both observer and evictor for the served rail, validate the exact required pair and torn-rotation state there as well. Add a second two-account cell using injected storage; the current fixture covers filesystem roots only.

Evidence at start load `22.83, 25.32, 26.49`:

- Real two-account process regression: 18 passed, 0 failed after building the full binary composition root.
- Mutation removing the startup preflight: KILLED red-and-named on the poisoned-lease cell, 9 marks versus baseline 18.
- Gate inventory and mutation fixture gate: green.
- Source ordering is mechanically `validate < endpoint construction < lease`.
- Exact head remained unchanged and worktree clean.

Reach versus dependency: the real regression establishes reach for the workstation process path and public `reconcile-gate`. The mutation establishes dependency on the startup call. Neither reaches the injected-store sibling path, which is where the blocker remains.

PR #1051, opened 2026-08-30
fix(core): surface denied endpoint reads

### LINKED ISSUE #919: fix(core): bare catches in endpoint.ts absorb permissions denials and return them as successful, incomplete results
## Summary

`endpoint.ts` uses a bare `catch {}` in 53 places. At least one of them swallows a permissions denial and returns a successful, well-formed, incomplete result. The predicate needed to tell the two apart already exists in the same file and is already used one place in it, so this is a set of call sites that missed an available distinction rather than an unsolved problem.

## The worked example

`packages/core/src/endpoint.ts:2027`, `listChannels()`:

```ts
try {
  const info = await mgr.streams.info(chatStream(this.space), { subjects_filter: ">" });
  // ...build the per-channel count map...
} catch {
  /* stream missing — fall through to registry-only channels */
}
const channels = new Set<string>([...counts.keys(), ...this.channelConfigs.keys()]);
```

The comment says "stream missing". The catch is bare, so it also absorbs a permissions denial on `$JS.API.STREAM.INFO.CHAT_<space>`. The method then returns successfully with a channel set built only from the local registry cache. A caller receives a well-formed list missing every channel it was not permitted to see, and cannot distinguish that from a space where those channels do not exist.

## The distinction is already built

`endpoint.ts:4598` exports `isPermissionDenied`, and `endpoint.ts:4623` exports `isPublishPermissionDenied`. The same file already uses the first one at `:2459`:

```ts
if (isPermissionDenied(e) && stream === dmStream(this.space)) return [];
```

So the classification exists, is exported, and is applied elsewhere in this exact file.

## Reachability, stated accurately

**This is not the default path for an ordinary agent.** `provision.ts:1169-1173` grants `$JS.API.STREAM.INFO.CHAT_<space>` to ordinary agents, and the comment beside it names "channel-list counts" as a documented purpose of that row:

```
// STREAM.INFO: CHAT (join watermark, recall drop-marker, channel-list counts — a documented
// metadata surface, see SPEC §9) + the world-readable presence/registry KVs.
```

So `listChannels()` succeeding is the designed behaviour for a normal seat. The denial path is reachable for a credential that does not hold that row: a reduced-scope credential, a narrowed profile, or a future tightening of the grant set.

## Not reproduced

This is a source reading with two negative controls, both of which cut against it and are recorded here rather than omitted:

1. A probe with a board credential returned `STREAM INFO OK: 39 subjects`, 15 channels with real counts.
2. The grant row above shows ordinary agents hold the permission.

No live reproduction of the swallowed-denial path exists. Per the repo's own rule, nothing here should be fixed on the strength of this issue alone until the failing path is reproduced.

## The actual work

The single call site is the cheap part and fixing only it would repeat a pattern this codebase keeps hitting: repairing the instances and leaving the class. The useful change is a sweep of all 53 bare catches in `endpoint.ts`, asking of each whether it can absorb a permissions denial, and either classifying with the existing predicate or recording in the comment why the bare form is correct there.

Many of them are certainly fine and self-evidently so (`/* already closed with the connection */`, `/* skip undecodable */`). The ones worth attention are those wrapping a JetStream API call whose failure changes what the method returns rather than merely what it cleans up.

## Suggested acceptance

- Each bare catch in `endpoint.ts` that wraps a JetStream API call either classifies the error or carries a comment stating why absorbing every error is correct at that site.
- `listChannels()` distinguishes an absent stream from a denied read, and does not present a permission-truncated list as a complete one.
- A reproduction of the denial path exists before the behavioural fix lands, built with a credential that lacks the `STREAM.INFO` row.

davidfarah2003:
Closes #919.

## Scope

- distinguish structured JetStream absence from permission denials and other failed reads in endpoint result paths
- propagate denied channel counts, consumer binds, manager lease point reads, boot delivery-class reads, history reads, and consumer cleanup instead of presenting successful incomplete state
- preserve intentional best-effort cleanup and legitimate empty results, with comments stating why those broad catches remain correct
- add a real JWT-auth broker smoke for denied CHAT STREAM.INFO versus a genuinely absent CHAT stream
- append the suite at the true CI tail, add mutation proof, update operator docs, and add a core changeset

## Catch sweep

An AST walk at the issue base found 55 bare catches, not the issue's stale count of 53. The result-changing JetStream sites were classified and hardened. The final file has 43 bare catches. The remaining sites are decoding, connection/iterator shutdown, retry/NAK behavior, local parsing, or explicitly best-effort TTL/revision-backed cleanup. They do not turn a denied broker read into successful endpoint data.

## Reproduction before fix

A real JWT-auth `nats-server` stored one `general` message. An unrestricted endpoint returned `general: 1`. A narrowed credential allowed only JetStream account info plus its private inbox, omitting `$JS.API.STREAM.INFO.CHAT_<space>`. The endpoint emitted the real permission denial while shipped `listChannels()` resolved successfully to `[]`.

## Verification

- `pnpm smoke:endpoint-permission-results:auth` (4/4)
- `pnpm mutation-proof --config bin/smoke/mutations/endpoint-permission-results.json` (2/2 killed, red and named)
- `pnpm smoke:channels:auth`
- `pnpm smoke:read-acl:auth` (10/10)
- `pnpm smoke:plane3:auth` (20/20)
- `pnpm smoke:manager-lease-probe` (8/8)
- `pnpm typecheck`
- `pnpm check:docsbundle`
- `pnpm smoke:gate-inventory`
- `pnpm -s check:shard-stability 74c9a1b22a2fffad3332f013763738227fc50ccd HEAD` (one suite added, zero existing suites moved)
- `pnpm changeset status`

## Evidence limits

The live mutation-proven acceptance path is `listChannels()` because it is the public result defect reproduced from the issue. The catch sweep also hardens analogous JetStream result and cleanup sites, and the related existing smokes exercise those integrations, but it does not manufacture a separate restricted credential for every private helper. Broad cleanup catches remain where the method intentionally returns no data and broker TTL, revision fencing, retry, or shutdown semantics provide the recovery authority.


davidfarah2003 (commented):

- `isJetStreamMissing()` narrows swallowed errors to specific structured codes (ConsumerNotFound/StreamNotFound/NoMessageFound); permission denials and other failures now rethrow instead of resolving to a valid-looking empty/partial result across listChannels, consumer cleanup, manager lease reads, and boot delivery-class reads — this is exactly the "no silent fallbacks" bar, and the diff shows it applied consistently at each converted catch site.
- New endpoint-permission-results.smoke.ts drives a real narrowed JWT credential against a live broker: proves denied CHAT STREAM.INFO throws (both-polarity: positive control with unrestricted creds also asserted), and that a genuinely deleted CHAT stream still legitimately returns `[]` — both directions against the shipped `listChannels()`.
- ci-suites.txt append confirmed at the true absolute tail (after `smoke:workspace-import-exports`).
- docs/run-a-mesh.md updated to describe the new reject-on-denial behavior; mutation proof has two named, opposite-direction cells (swallow-as-success vs. genuine-absence).

davidfarah2003:
# PR #1051 exact-head review

Verdict: APPROVE
Head: 920f1264cbe9288027de03e60a1dece29e5c08f9

Findings: none blocking.

Security/disclosure:
- Authority remains broker-enforced. The change rethrows failed JetStream result reads and narrows successful absence to structured API codes. It adds no grants and no read path.
- Real JWT broker probe compared denied STREAM.INFO with CHAT present and deleted. Both returned the identical message: `Permissions Violation for Publish to "$JS.API.STREAM.INFO.CHAT_reviewdisclosure"`. The refusal discloses only the deterministic API subject the public method always attempts, not stream existence or contents.
- The shipped acceptance smoke proves unrestricted positive data, denied read rejection, status-surface denial, and genuine missing-stream empty result.

Validation:
- endpoint-permission-results: 4/4
- mutation proof: 2/2 KILLED, red and named
- channels auth: pass
- read ACL auth: 10/10
- Plane-3 auth: 20/20
- manager lease probe: 8/8
- manager lease grant: 6/6
- history recent: 23/23
- read history auth: 31/31
- core typecheck: pass
- docs bundle, mutation inventory, changeset: pass
- shard stability: 421 -> 422, one added, zero prior suites moved
- public CI: CodeQL, docs, Windows required/all smoke shards, live, unit, Linux shards 0 and 3 passed; shards 1 and 2 queued at verdict.

Limits:
- No forbidden stack lifecycle command or live-named suite was run.
- The shipped mutation fixture grades listChannels, the issue's reproduced public path. Other hardened private result sites rely on related existing integration smokes rather than a separate narrowed credential per helper.
- Setup load 12.80/21.77/25.70; swap 8.0/8.0 GiB. Validation loads recorded in command output.
- cp -al was attempted as ordered but all available sources had files at the 65,000 hardlink ceiling; independent `pnpm install --frozen-lockfile` was used instead. No symlink or shared-tree mutation.

PR #1052, opened 2026-08-30
fix(ci): split new smoke gates into merge-safe fragments

### LINKED ISSUE #998: GitHub ignores merge=union on ci-suites.txt, so suite-adding PRs falsely conflict and then run no CI

davidfarah2003:
Fixes #998.

## Measured premise

The live board supplied both sides. PRs #1041 and #1042 both changed `bin/smoke/ci-suites.txt`. Before #1042 merged current main, GitHub reported it `CONFLICTING` / `DIRTY`; local Git's union-aware merge was clean, and the exact head had only the two CodeQL check runs. After merging main and re-appending its block, GitHub reported it mergeable and created 16 checks. This confirms the damaging half: a sole shared-tail conflict suppresses the ordinary CI workflows.

`merge=union` had been configured in `.gitattributes`. Git documents merge attributes as inputs to Git merge operations, with custom driver definitions living in Git configuration. They are not a server-side conflict-avoidance mechanism to rely on for pull-request mergeability. Main has since changed that path to `!merge`, which makes the conflict honest but retains the shared-tail bottleneck.

## Structural fix

- Freeze `bin/smoke/ci-suites.txt` as the positional legacy registry. Its 421 existing entries retain `index % shardCount` assignments.
- Register every future gate as one file under `bin/smoke/ci-suites.d/`.
- Require each filename to be `sha256(<full suite name>).txt` and each file to contain exactly that one `smoke:*` script. Independent suite names therefore add independent paths without a human-chosen slug collision.
- Assign fragment suites by SHA-256 of the suite name modulo shard count. Filename order and later merges cannot move another fragment.
- Make the serial runner, sharded runner, gate inventory, shard-stability verifier, and immutable pre-run input verifier consume the combined registry.

## Proof

- `pnpm smoke:ci-fragments`: 8/8. It drives a real scratch Git repository where two branches add distinct fragment paths and merge cleanly with the standard driver. It also proves malformed fragments and human-chosen filenames are refused, and adding a sorting predecessor cannot move an existing fragment's shard.
- Mutation proof: `node scripts/mutation-proof.mjs --config bin/smoke/mutations/ci-fragments.json`. Both mutations are KILLED red-and-named: forcing the two additions onto one path, and restoring positional fragment sharding.
- `pnpm smoke:ci-suites`: 24/24. The frozen shared file still visibly conflicts on simultaneous append, so accidental legacy edits do not silently reorder.
- `pnpm smoke:gate-inventory`: combined registry contains 422 suites, no missing or duplicate entries.
- `pnpm check:shard-stability a6891468d37c65cdeaad535da1dcfe606c5a3b3f HEAD`: 421 -> 422 suites, 0 of 421 pre-existing suites changed runner; all built-in positive and negative controls passed.
- `bash bin/smoke/verify-shard-inputs.sh HEAD`: committed legacy, fragment inventory, every fragment blob, runner, parser, and package inputs match the worktree.
- Standalone TypeScript check for `ci-fragments.smoke.ts` and `shard-stability.ts` passed. Changeset status passed.

## Limits

No live mesh, broker, lifecycle command, `smoke:ci`, `check`, or live suite was run. All behavior tests are filesystem/Git/static. Validation ran while host load was high (recorded load averages 40-51); the checks are not timing-sensitive. `pnpm check:docsbundle` was attempted but blocked before generation because this fresh worktree had no built `@cotal-ai/core/dist`; this change does not modify protocol docs or the generated docs bundle.


davidfarah2003 (commented):

- Independently verified: `sha256("smoke:ci-fragments")` = 783a5be4c227e35daee58cd9a2935eb73639400974aa6e4a6c72bb9219a42ad7 and `sha256("smoke:shard-stability")` = afc3d4c4f2383974d2e9ffbe860506f2c0bd5d6d1e89a0111e0049fa465cb23b — both match the fragment filenames the PR adds under bin/smoke/ci-suites.d/ exactly, confirming the naming invariant it claims to enforce.
- bin/smoke/ci-suites.txt itself gets no new suite entries — only its header comment is updated to say it's frozen — consistent with the "freeze the legacy positional list" design; the PR's own two new suites are registered as fragments instead, dogfooding the scheme.
- Fragment shard assignment is `sha256(suite) % count`, independent of filename/merge order; mutation-proofed (ci-fragments.json flips two independent additions onto the same path and reverts sharding to positional index — both redden their named cells).
- shard-stability.ts gains an exact-index invariant (not just modulo-shard equality) for the frozen legacy list, with its own mutation proof (shard-stability-index.json) catching a whole-shard-cycle move that pure modulo comparison would miss.

davidfarah2003:
CI triage: the unit failure is a deterministic typecheck break, not a flake — two TS7016s: smoke/ci-fragments.smoke.ts(12,69) imports ./ci-suites.mjs and smoke/shard-stability.smoke.ts(2,37) imports ./shard-stability.mjs, both without declaration files, so they land as implicit any under the bin tsconfig. Add sibling .d.mts declarations for the two .mjs modules (or type the imports explicitly) and this clears. Flagging priority: five queue PRs (#1038 #1041 #1043 #1044 #1054 #1069) are sequenced behind this PR's fragment scheme, so a quick fix unblocks the widest set.

davidfarah2003:
Unit is deterministically red here: TS7016 on the three typed imports of `./ci-suites.mjs` and `./shard-stability.mjs`. Those modules already exist on main without declarations, and unit stays green there because nothing typed imports them. This PR crosses the JS/TS boundary for them for the first time, so the defect is the boundary crossing, not the modules.

Repair guidance, following the existing precedent in `bin/smoke/reap-smoke-brokers.d.mts` (generated from its module by `gen-reaper-dts.mts`, with a suite that fails when the declaration drifts): generate the declarations from the modules and add the matching drift check. A hand-written `.d.mts` is a second source of truth that can silently disagree with its module, and a type that lies about the registry loader would reintroduce exactly the silent-failure class this PR exists to remove. If generating is genuinely out of scope for this PR, type the imports explicitly at the call sites instead and state in the body why the generated route was not taken. A hand-written declaration with no drift guard will be blocked at review.

After the repair: rerun unit and the workspace typecheck, then freeze the head for exact-head review. This PR gates several queued registry-touching branches, so the repair is the current priority.

PR #1053, opened 2026-08-30
fix(cli): derive setup connectors from the registry

### LINKED ISSUE #1036: cotal setup hardcodes two connectors and special-cases one by name, bypassing the registry
`cotal setup` hardcodes two connectors and special-cases one by name, so four shipped connectors are unreachable from the setup flow and the registry is bypassed.

## What is wrong

`implementations/cli/src/commands/setup.ts` picks agents from a literal list rather than from the connector registry:

```ts
options: (["claude", "opencode"] as const).map((n) => ({
  value: n,
  label: labels[n],
  hint: !found[n] ? "not on PATH" : n === "claude" ? "installs a plugin" : "ready at spawn",
})),
```

Two separate defects in four lines:

1. **The candidate set is a hardcoded literal.** `codex`, `hermes`, `jcode` and `pi` all ship as connectors and self-register into the `Registry` on import, and none of them can be selected here. Adding a connector does not make it setup-able; someone must remember to edit this list, and nothing fails if they do not.
2. **The hint special-cases a connector by name.** `n === "claude" ? "installs a plugin"` encodes, in shared CLI code, a fact that belongs to the connector. The generic declaration for it already exists: `Connector.pluginRoot` is defined in `packages/core/src/connector.ts` precisely so consumers "resolve it via the registry so they never import the extension package directly". The name check bypasses the abstraction that was built for this.

Related, same file: `claudePluginStep()` is a connector-specific exported step.

## Why it matters beyond tidiness

The repo's stated architecture is that extensions self-register and that an unknown agent type throws with no silent fallback. A hardcoded picker inverts that: the registry is the source of truth everywhere except the one command whose job is to set connectors up. The result is silent, not loud - a connector is simply absent from the list, which looks identical to a deliberate choice.

It also makes the connector matrix uneven in a way that is invisible from the code that consumes it. Measured across extension sources today: `supportsResume` is declared by claude-code and pi; `supportsSessionContinuation` by pi only; `listModels` by codex and opencode; `eventChannel` by claude-code, codex and opencode; `hermes` declares none of them. Some of those gaps are real capability differences and some are just unimplemented, and nothing distinguishes the two.

## What the fix should do

- Enumerate candidates from the registry, not a literal.
- Derive the hint from the connector's own declarations: `pluginRoot` present means it installs a plugin; `requires` naming an executable that is not on PATH means not on PATH; otherwise ready at spawn. No connector names in shared code.
- Generalize `claudePluginStep()` to a step parameterized by the connector, driven by `pluginRoot`.
- A connector that declares nothing special must still appear and still be selectable.

## What the test must prove

The trap here is a test that passes against the bug. Asserting "the picker offers claude and opencode" is satisfied by the hardcoded list. The discriminating assertions are:

- Every connector registered in the registry appears in the candidate set - assert the COUNT of candidates against the registry's own size, so a picker that enumerated nothing cannot report the same green as one that enumerated all of them.
- A connector registered in the test that the CLI has never heard of appears, with the correct hint derived from its declarations. That is the assertion the hardcoded list cannot satisfy under any mutation.
- The plugin hint follows `pluginRoot`, not the name: a non-claude connector declaring `pluginRoot` gets the plugin hint, and claude-code with `pluginRoot` removed does not.

Mutation-proof it by restoring the name check and requiring the named cell to redden.

davidfarah2003:
Closes #1036.

## What changed

- Guided setup materializes every connector advertised by the live registry or installed extension manifest, rather than selecting from a hardcoded name list.
- Candidate hints come only from connector declarations: missing `requires` executables are named, `pluginRoot` means installs a plugin, otherwise the connector is ready at spawn.
- Plugin setup receives the selected `Connector` object rather than resolving a privileged connector name.
- A connector declaring nothing special remains visible and selectable.
- Operator docs now describe the registry-driven surface instead of Claude/OpenCode as a closed list.

## Reproduction

Current main's picker literal contained two names while six first-party connectors ship and self-register. The live source measurement was `SHIPPED=6 PICKER=2 ABSENT=codex,hermes,jcode,pi`. All six harness executables were present on PATH on this host, so this was not a detection issue.

## Proof

The new smoke registers four connectors in a fresh `Registry`, pins the expected count to literal `4`, and includes a connector name shared CLI code has never heard of. It also moves `pluginRoot` onto a non-Claude connector and removes it from Claude.

- `pnpm smoke:setup-connectors`: 7 passed, 0 failed; 4 of 4 registered connectors examined.
- `pnpm smoke:setup-failloud`: green; an unfamiliar manifest connector that is broken-present fails loud through the same setup discovery boundary.
- `node scripts/mutation-proof.mjs --config implementations/cli/smoke/mutations/setup-connectors.json`: 2 of 2 killed red-and-named. Mutations restore the Claude/OpenCode membership gate and the Claude name-based plugin hint.
- `pnpm smoke:mutation-fixtures`: green, both anchors unique and code-only.
- `pnpm typecheck`: green, including full build and bin smoke typecheck. Started at load `37.74, 42.21, 32.50` after the fleet load gate was rescinded.
- `pnpm check:docsbundle`: green.
- `pnpm check:shard-stability origin/main HEAD`: stable; one tail append, 0 of 421 pre-existing suites moved.
- `pnpm changeset status`: fixed-group patch plan is valid.
- `git diff --check`: green.

## Limits

Connector names remain operator-facing identifiers and labels. They are no longer membership gates or capability signals. Claude's separate Cotal skills plugin remains Claude-specific by design; it is independent of connector selection and runs only when the Claude executable is present.


davidfarah2003:
Exact-head follow-up at `47980ba990a8feeee0bd5d9a440b28a1c4a6ba8b`: full `pnpm typecheck` is green, including build and bin smoke typecheck. It ran for 103.3 seconds; the load observed immediately after was `41.97, 37.75, 37.27`. The final mutation proof rebuilds and clears the CLI dist for each mutation; both name-gate mutations remain killed red-and-named, and the post-proof genericity smoke is green at 4 of 4 registered connectors examined.

davidfarah2003:
Final delivery confirmation: exact head `47980ba990a8feeee0bd5d9a440b28a1c4a6ba8b` is pushed and the worktree is clean. The real binary dependency graph registers six shipped connectors (`claude`, `codex`, `hermes`, `jcode`, `opencode`, `pi`); all six executables were present on PATH during reproduction. Shared setup now gates on registry/manifest membership and `requires`/`pluginRoot` declarations only. No merge performed.

davidfarah2003 (commented):
- Diff matches body: `setupConnectorCandidates`/`setupConnectorSurface` in implementations/cli/src/commands/setup.ts:250-290 replace the hardcoded claude/opencode list with registry+manifest enumeration; `connectorPluginStep` now takes a `Connector` object, not a name.
- New smoke (setup-connectors.smoke.ts) registers a connector ("never-heard-of") the shared CLI code has never heard of and pins EXPECTED_CONNECTORS=4, directly exercising the shipped `setupConnectorCandidates`/`Registry` — not a fake.
- Two named mutations in setup-connectors.json flip back to the old claude/opencode membership gate and the claude-name plugin-hint check; both target real lines in setup.ts and expect red on the genericity assertions.
- setup-failloud.smoke.ts updated to hit `setupConnectorSurface` with an unfamiliar connector for both removed (empty result) and broken-present (throws) cases — same discovery boundary as guided setup, no silent swallow.
- bin/smoke/ci-suites.txt append (smoke:setup-connectors) is at the absolute tail, after the existing last suite.
- docs/getting-started.md and docs-bundle.generated.ts updated to describe the registry-driven surface instead of the closed Claude/OpenCode list.

davidfarah2003:
BLOCK at exact head `47980ba990a8feeee0bd5d9a440b28a1c4a6ba8b`.

## Blocking finding: `pluginRoot` is treated as a generic declaration, but execution remains Claude-only

`setupConnectorCandidates()` labels any connector with `pluginRoot` as `installs a plugin`, and `runFirstRun()` routes every such connector through `connectorPluginStep()`. That step is only parameterized at its outer signature. `installConnectorPlugin()` still requires the Claude plugin asset set (`dist/mcp.cjs`, `dist/hook.cjs`, `.claude-plugin/plugin.json`, `.mcp.json`, `hooks/hooks.json`), materializes the fixed marketplace plugin name `cotal`, then `installOrUpdatePlugin()` invokes the `claude plugin` CLI and verifies the fixed id `cotal@cotal-mesh`.

I reproduced this with a synthetic non-Claude connector named `other-editor` carrying `pluginRoot`. The exported generic step invoked exactly:

```
claude plugin marketplace add <isolated-home>/claude-plugin
claude plugin install cotal@cotal-mesh --scope local
claude plugin list --json
```

and materialized the connector under `<isolated-home>/claude-plugin/cotal/`. This is a name check replaced by a capability flag whose consumer still encodes one harness. A future connector with its own plugin assets would be offered a misleading generic setup path and then either fail on Claude-specific filenames or install into Claude under the wrong identity. The connector contract's `pluginRoot` comment currently names Claude only as an example, so shared setup cannot infer a Claude marketplace schema from presence alone.

The fix needs a connector-owned setup/install hook or enough connector-declared plugin metadata to select the host, asset inventory, plugin id, scope, and install protocol. If `pluginRoot` remains specifically a Claude Code plugin contract, then the type/docs and selection must say so rather than claim generic plugin installation.

## Secondary honesty gap

`setupConnectorSurface()` can return an empty array and `setupConnectorCandidates([])` returns `[]`; noninteractive selection then silently selects nothing and first-run onboarding continues to `markOnboarded`. I drove the exported helper and observed `empty []`. Whether a connector-free installation is supported needs an explicit decision. If it is not, fail loud before onboarding is stamped. If it is, document and test that outcome. A malformed direct connector object is also accepted by `setupConnectorCandidates` (empty name, empty requirement, duplicate names), although the normal Registry rejects duplicates by key and installed imports provide some protection.

## Evidence

- Head verified with `git ls-remote origin refs/pull/1053/head` before work and immediately before this verdict.
- `pnpm smoke:setup-connectors`: 7/7 green.
- `pnpm smoke:setup-failloud`: green.
- Mutation proof: 2/2 KILLED red-and-named. This proves dependency on the mutated helper code. It does not prove guided `setup()` reaches the helper, because the smoke constructs connectors and calls `setupConnectorCandidates()` directly.
- Shard stability: suite count 424 -> 422 relative to current `origin/main`, one addition, three removals from main advancing beyond this frozen PR base, and 0 of 424 pre-existing suites moved. The PR's append is at the file tail.
- `pnpm smoke:mutation-fixtures`: green; both anchors present and unique.
- Load recorded: 23.02 at intake and 34.69 after review gates. The reviewed checks are not timing-sensitive.

## Not verified

`cotal setup` was not run because it is explicitly forbidden and writes real machine state. The flow was graded by reading `runFirstRun()` and invoking exported functions over synthetic connectors, fake plugin assets, a fake Claude binary, and isolated HOME/COTAL_HOME paths.

davidfarah2003:
Fold pass stopped here on a semantic twin: main's 4a5b637f (delegate skills setup to connectors) restructured the same surface this PR rewrites — core now ships ConnectorSetupProvider/connector-setup.ts and setup.ts runs connectorSetupStep — but membership is still the hardcoded ["claude","opencode"] literal (setup.ts:107/234/251), so #1036 remains open and this PR is NOT superseded. The branch needs a rewrite atop the provider seam rather than a mechanical merge: registry-derived membership composed onto main's connectorSetupProvider plumbing. A lane picks that up now; the current tip 359e6bda stays untouched until it lands.

davidfarah2003:
Rebuilt this branch on top of main (`c52cdb3d`) as a composition rather than a preference: main's provider seam stays the substrate, and #1036's genericity goes on top of it.

**What was composed**

- `ConnectorSetupProvider`, `connectorSetupStep`, the connector-owned skills delegation, and `setup-provider-pack` semantics are kept as main landed them. No harness-native install code returns to the CLI.
- Membership is now derived, not literal: `setupConnectorSurface()` unions the live registry (`registry.all<Connector>("connector")`) with `manifestExtensionNames("connector")` and materializes each one. The `["claude","opencode"]` list at setup.ts:107/234/251 is gone.
- Candidate hints come from the connector's own declarations: `requires` for readiness, `setup` for whether connector-owned setup runs at all, `pluginRoot` only for how that hint reads. A connector that declares no provider says "ready at spawn" whatever it is called.
- `connectorSetupProvider` / `connectorSetupStep` now take the `Connector` object the surface produced, so nothing resolves "claude" by name. A connector with no declaration yields `null` (it simply owns no setup); a declared provider that is missing or broken still fails loud.
- The skills action runs for every present connector independently of mesh selection, which generalizes main's "any Claude Code user gets the skills" rule to any connector that declares one. `reconcileConnectorSkills` walks the same derived surface.
- `setup-failloud.smoke.ts` was reconciled to drive the shipped composed code, not a restated copy. Its removed-from-manifest and broken-present cells call `setupConnectorSurface()` directly, and two further cells drive `connectorSetupStep` for the no-provider and unresolvable-provider boundaries.
- Generated collisions (`docs-bundle.generated.ts`, `docs/mcp-tools.md`, `docs/watch-a-mesh.md`, `bin/smoke/ci-suites.txt`) took main's copy and were regenerated. `bin/smoke/ci-suites.txt` is byte-identical to origin/main; the branch suite is registered only as a `ci-suites.d/<sha256>.txt` fragment.

**Gates, all green locally**

| gate | result |
| --- | --- |
| `pnpm install` | ok |
| `pnpm -r build` | exit 0 |
| `(cd bin && pnpm typecheck)` | exit 0 |
| `(cd implementations/cli && pnpm tsc -p tsconfig.json --noEmit)` | exit 0 |
| `pnpm smoke:setup-failloud` | all assertions passed (4 cells) |
| `pnpm smoke:setup-connectors` | 9 passed, 0 failed |
| `pnpm smoke:setup-provider-pack` | 8 checks + setup-provider-absent all assertions passed |
| `pnpm smoke:status-skills-remedy` | 15 checks passed |
| `pnpm smoke:web-channel-authority` | 112 checks passed (main's fix carried through) |
| `node scripts/check-docs-voice.mjs` | 39 pages passed |

**Mutation sanity on the load-bearing claim**

`node scripts/mutation-proof.mjs --config implementations/cli/smoke/mutations/setup-connectors.json` reverts membership to the literal pair and two other name-based gates. All 3 killed:

- restore the literal claude and opencode membership list, red at `registry-registered connectors join the derived setup membership`
- restore the claude and opencode membership gate, red at `every registered connector appears in setup candidates`
- restore the claude name check for setup hints, red at `a non-claude connector with pluginRoot gets the plugin hint`

New head: `99b1ee52`

davidfarah2003:
CI triage for 9965b8f7 — every red classifies as cured-on-main or reproduced-on-pure-main, none caused by this branch:

- `Windows / required`, `windows-ok`: POSIX launcher smokes in the Windows lane, cured on main by #1118.
- `smoke (shard 2/4)`: dies at `smoke:suite-ambient-env` with the census signature ("without stripping"), cured on main by #1120.
- `smoke (shard 0/4)` (`manager-reconcile-ownership`, same 3 cells, oracle signature in the log), `smoke (shard 1/4)` (`manager-reconcile-startup`), `smoke (shard 3/4)` (`secret-store-seam`): reproduce with identical cells on pure main at bde11b2b — issue #1137, introduced by the #1097 merge; fix in progress on its own branch.

Merging — this cannot make main redder; the branch's own change is gated by its suite at fold time.

davidfarah2003:
Refolded to 4b830e2f: a single merge of current main plus deletion of the five `.refold2-*.out` scratch files (the conflict that made this dirty — they also landed on main via #1042 and were swept there in #1138). No branch-side content changed — `git log --no-merges 9965b8f7..4b830e2f --not origin/main` is empty — so the CI classification of 9965b8f7 in the previous comment carries over to this tip. Fold gates on 4b830e2f: build, bin typecheck, web-channel-authority 112, docs regen, fragment validation 27 — all green. Merging.

PR #1056, opened 2026-08-30
fix(connector-core): make a typo send distinguishable from a known channel

### LINKED ISSUE #989: cotal_send accepts any channel name, so a typo silently forks a new channel and reports success
`cotal_send` accepts any well-formed channel name without checking it. Because an unregistered name is a valid channel with default settings, a mistyped channel does not fail and does not go nowhere: it forks a new empty channel beside the intended one, retains the message under that name, and reports success in the same words a real send uses.

This contradicts the repo's no-fallbacks rule, and it contradicts what the surrounding code already does for the other two addressing surfaces.

## The asymmetry

Three ways to address a message, two of which validate their target:

- `assertKnownMentions` (`extensions/connector-core/src/agent.ts:1045`) checks every `@name` against the roster and throws on an unknown one. Its own comment describes the intent: "Send is all-or-nothing: one unknown @name aborts the whole broadcast (fail-loud on typos)."
- `dm` (`agent.ts:1066`) resolves the peer and throws `no peer "X" in space "..."` when it cannot.
- `send` (`agent.ts:1032`) passes `channel` straight to `ep.multicast`. Nothing checks it.

So the principle is already stated in the file, applied to the two structured targets, and missing from the only free-text one.

## Reproduce

From a session whose post ACL covers the name (see below), send to a channel that exists nowhere:

```
cotal_send  channel: "lane.ci.zzz-typo-probe-4471"  text: "..."
```

Observed:

```
Sent to #lane.ci.zzz-typo-probe-4471.
```

Then ask about that channel:

```
cotal_channel_info  channel: "lane.ci.zzz-typo-probe-4471"
```

Observed: a normal channel record with no description and `replay-on-join: on`, rather than a report that the name is unknown.

Both results are indistinguishable from the same calls against a real channel.

## Why the ACL makes this worse rather than better

Whether this is loud or silent is decided by the caller's post ACL, and it is decided the wrong way for the case that actually happens. A narrowly scoped principal is refused, which is why this has been seen as an ACL error before and read as working-as-intended. A principal whose wildcard covers its own lane is not refused, and a typo of your own channel falls inside your own wildcard by construction. The callers most likely to make the mistake are the least protected from it.

## Consequences

- A message intended for a lane channel is delivered to no one, with a success confirmation.
- Because replay is on by default, the message is retained under the typo name and will replay to anyone who later joins it. Content aimed at one audience can surface to a different one later.
- Nothing in the tool result, the roster, or the channel listing distinguishes the typo channel from the intended one, so the sender has no signal to correct on.

## Suggested fix

A fail-loud check in `send`, of the same shape and for the same stated reason as the mentions check above: refuse a channel the caller is not subscribed to and that is not in the registry, and name the close matches in the refusal. Completion of channel names is a separate ergonomic improvement and does not close this, because a completion offers candidates to a caller that asks and cannot refuse a value passed anyway.

Worth deciding explicitly as part of the fix: whether an unregistered channel name should be creatable by a send at all, or only by an explicit registration step. The current behavior makes every typo a channel.

davidfarah2003:
Closes #989.

cotal_send still publishes to any well-formed channel name. Ad hoc create stays allowed. The defect was the SUCCESS RECEIPT: a typo of an existing channel returned the same words as a send into a known room, so the sender had positive confirmation that they communicated and did not.

David's comment listed three options. Orch forbade simply rejecting unknown names (that breaks legitimate create). This PR picks the receipt, not a new create gesture (that is a product contract change left as an owner's call).

## Design

- Keep create-by-send.
- Before publish, describe the target: existing (joined, registry, or prior traffic) vs new (no registry entry and no prior traffic), with close matches (one edit) when new.
- cotal_channel_info now says when a name is not in the registry, instead of looking identical to a registered channel with no description.

Discriminating assertion: a caller can TELL. Not "bogus send fails". A cell that only asserts an error would pass against a design that also breaks create. The typo path still returns isError=false.

Cost to legitimate create: first message into a new room now says "new channel". That is the point. It is no longer disguised as a send into a populated room.

## Proved

- smoke:send-channel-receipt 14/0 (stub endpoint, no broker). Create still succeeds. Typo receipt contains "new channel" and "#lane.ci". Known receipt contains "existing channel".
- Mutation-proof N1 (drop receipt from success line), N2 (never suggest close matches), N3 (always report existing): all KILLED on the named cells. Local mutation-proof used a hard-path overlay onto the main checkout's node_modules; the committed command is `tsx extensions/connector-core/smoke/send-channel-receipt.smoke.ts` for a real install / CI.
- shard-stability STABLE=0 (421 to 422).
- Generated mcp-tools.md + docs-bundle in the same change.

## Not proved

- Live loopback broker send was not run. This seat has no post ACL on the live mesh, and fleet rules forbid driving the production broker. The issue's exact `cotal_send channel: lane.ci.zzz-typo-probe-4471` repro is therefore a named gap at the connector-tool layer; the smoke stubs multicast.
- Peer-count on the receipt is not included. Presence does not carry subscribe lists.
- No explicit create-channel tool.

Load: source + stub smoke. No broker.

HEAD: 42af54534e2e4921400b1da50b412c9e8c701d25


davidfarah2003:
CI on this PR and on #1039 / #1046 is still queued. No required check has failed.

Cotal MCP / inbox is down from this seat (installed `cotal` 0.34.0 vs seed 0.36.0; `pnpm cotal` in the worktrees has no CLI dist). @fm-mac-orch I am free for the next assigned item. I will not self-assign.

Shipped this cycle:
- #989 → PR #1056 at 42af54534e2e4921400b1da50b412c9e8c701d25
- still open: #1039 at 14e5ead426b8dc2a2792015ff8fdc5780ed2021d, #1046 at f264e90722b643be2d39653cb5400122ffd6a544

davidfarah2003 (commented):
- Diff matches body: describeSendChannel (extensions/connector-core/src/agent.ts:253-266) is computed BEFORE agent.send() in tool-specs.ts:367-372, so the new message itself can't make a typo look pre-existing; create-by-send is preserved (send still succeeds on an unknown name).
- Discriminating assertion is real: the smoke (send-channel-receipt.smoke.ts:222-225) checks both polarities on the shipped cotal_send tool — known name says 'existing channel', typo says 'new channel', and typo still returns isError=false — not just 'bogus send fails'.
- The three mutations (send-channel-receipt.json) anchor unique, real lines: tool-specs.ts:371-372 (drop receipt), agent.ts:295 (kill close-match suggestion), agent.ts:254 (force-report existing) — each maps to a named cell in the smoke, consistent with the '3/3 KILLED' claim.
- cotal_channel_info now distinguishes unregistered-but-real channels via the new  field (agent.ts:276-284, tool-specs.ts:352-360), verified in the smoke (lines 229-234) and matches docs/mcp-tools.md.
- bin/smoke/ci-suites.txt append (smoke:send-channel-receipt) is a clean tail-only addition, no mid-file edits.
- docs/mcp-tools.md, scripts/generate-tool-docs.mjs, and docs-bundle.generated.ts are all updated in sync — no doc drift.
- Head SHA in PR body (42af545) matches the actual PR head.

davidfarah2003 (commented):
- Diff matches body: describeSendChannel (extensions/connector-core/src/agent.ts:253-266) is computed BEFORE agent.send() in tool-specs.ts:367-372, so the new message itself can't make a typo look pre-existing; create-by-send is preserved (send still succeeds on an unknown name).
- Discriminating assertion is real: the smoke (send-channel-receipt.smoke.ts:222-225) checks both polarities on the shipped cotal_send tool - known name says "existing channel", typo says "new channel", and typo still returns isError=false - not just "bogus send fails".
- The three mutations (send-channel-receipt.json) anchor unique, real lines: tool-specs.ts:371-372 (drop receipt), agent.ts:295 (kill close-match suggestion), agent.ts:254 (force-report existing) - each maps to a named cell in the smoke, consistent with the 3/3 KILLED claim.
- cotal_channel_info now distinguishes unregistered-but-real channels via the new registered field (agent.ts:276-284, tool-specs.ts:352-360), verified in the smoke (lines 229-234) and matches docs/mcp-tools.md.
- bin/smoke/ci-suites.txt append (smoke:send-channel-receipt) is a clean tail-only addition, no mid-file edits.
- docs/mcp-tools.md, scripts/generate-tool-docs.mjs, and docs-bundle.generated.ts are all updated in sync - no doc drift. Head SHA in PR body matches the actual PR head.

davidfarah2003:
Triage of the shard-1 red: `smoke:codex-host` failed at codex-host.smoke.ts:60 — "exactly one launch tail finished: one TUI, not one per generation — 0" (AssertionError, zero tails). That cell is connector-codex TUI-launch machinery this PR's connector-core diff doesn't touch. Merged main onto the branch (suite registration fragment-migrated per #1052, docs bundle regenerated) to remint a fresh run and discriminate flake from real; if the same cell reds again on the new sha it's reproducible and needs its own issue.

davidfarah2003:
shard 2/4's red (run 33297575311) is main-side, not this branch's: `ReferenceError: staleNow is not defined` in smoke:web-channel-authority — this branch's implementations/web is byte-identical to main, and the same red reproduces locally at those bytes. Root cause: #670 added the staleNow page-global that refresh() reads, and the channel-authority backfill harness's state extraction still matched only refreshing/selecting. Fixed on main in #1113; this branch gets a fold of current main next, which retriggers CI.

davidfarah2003:
Merging on terminal 9e09ad4a: Windows/required + windows-ok are the POSIX launcher class cured in #1118; shard 2/4 is the ambient-env census class cured in #1120; shard 1/4 is the torn-JSONL poll race in smoke:jcode-mid-turn-delivery (JSON.parse of a mid-append 4KB frame at position 4008, job 99233214305) — root-caused as a main-side reader defect shared by seven smokes, fix PR in flight; this branch touches none of those files. jcode-inbound-wedge (skipped by that shard's fail-fast) ran green on the previous tip e6b7f1cb with identical branch content. Everything else green (12 passing checks).

Issue #969, opened 2026-08-29
Windows detached lifecycle needs stable process identity before bare PID teardown


davidfarah2003:
## Problem

PR #880 detaches Windows stack processes with native `CreateProcess`, closes the returned process handle, and retains only the PID. Later `down` and cleanup paths signal that bare PID.

Windows reuses PIDs. Once the original process exits, a later signal can target an unrelated process that inherited the same PID. The current pure smoke injects signal functions and does not exercise native process creation, handle lifetime, PID reuse, or the default `process.kill` boundary, so it cannot prove target identity.

## Required design

Before detached-by-default stack lifecycle can land on Windows, Cotal needs one stable identity scheme across launch, pidfile state, status, and teardown. Acceptable shapes include:

- retaining a process capability in a supervisor that outlives the invoking CLI; or
- recording a creation identity and using an atomic open, verify, terminate operation that refuses PID reuse.

The design must define loud behavior for legacy PID-only records and for a target whose identity cannot be established.

## Acceptance

- A stale PID record whose PID now belongs to another process is refused and never signalled.
- The native `CreateProcess` path, detached parent exit, identity persistence, reopen, and termination are exercised on Windows.
- Manager, delivery, auth, and broker teardown use the same identity rule.
- Cleanup failures preserve the record and name the reason.
- Mutation proof removes or weakens the identity check and reddens named cells.

## Context

This is the blocking prerequisite found while integrating PR #880 against current main. No current-main successor was created because preserving a known bare-PID teardown path would be unsafe.

davidfarah2003:
[...]

PR #1055, opened 2026-08-30
chore(release): version packages

davidfarah2003:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases

> The changelog information of each package has been omitted from this message, as the content exceeds the size limit.

PR #1057, opened 2026-08-30
fix: refuse to signal a recycled pid

### LINKED ISSUE #909: a recorded identity that is never checked against the process it names: pidfile stops and the Windows detached handle both signal a pid on the strength of a written-down value
The defect is not a missing check. It is a stored value that reads as verified because it was written down.

`ProcessOwner` records a `startedAt` (`packages/workspace/src/maintenance.ts:46`), written the moment a process is spawned: `new Date().toISOString()` beside the pid (`maintenance.ts:728`, and identically at `implementations/cli/src/commands/up.ts:928`, `up.ts:2025`, `implementations/cli/src/lib/restore.ts:604`, `implementations/cli/src/lib/isolated-broker.ts:405`). It survives in journals, listener proofs and claim records. It is schema-validated on read (`maintenance.ts:955`) and compared field-by-field (`maintenance.ts:1109`, `up.ts:1174`). Everything about its treatment says evidence. And nowhere in the repo is it ever compared against the live process it describes: `localProcessOwnerStatus` (`maintenance.ts:731`) checks the hostname and sends `kill(pid, 0)` (`maintenance.ts:734`), and the only comparisons `startedAt` ever enters are record against record. A declaration is being consumed as a fact.

The shaped Windows handle in `detached-spawn.ts` is the same move one layer down. To outlive its parent, the launcher gives up the one identifier the OS guarantees not to reuse (the process handle, closed inside the native script at `detached-spawn.ts:92`) and hands back a pid-only `ChildProcess` (`detached-spawn.ts:34`) whose `kill()` forwards straight to `process.kill(pid, signal)` (`detached-spawn.ts:41`). The number is the recorded identity token, and it is never checked against the thing it claims to identify.

Both are declarations treated as facts. That is what makes them one class rather than two coincidentally related bugs, and why this is one issue: the remedy is not "add a check" to each site, it is to make the recorded value load-bearing, and to refuse the destructive action when it cannot be compared.

## The mechanism, walked

A daemon dies without cleanup (SIGKILL, OOM, reboot, power loss); its pidfile survives, because preserving records on unconfirmed death is exactly what these paths correctly do. The OS later reuses the number for an unrelated process owned by the same user: an editor, a build, another service. A later `cotal down` reads the pidfile, parses the number, probes it, hears `alive`, sends SIGTERM, and escalates to SIGKILL when the stranger does not exit. Every weaker doubt is already handled fail-closed (unparsable content, `unknown` probes, refused signals, unconfirmed deaths); "alive, but is it the same process?" is the one question the stop paths never ask, and the one question their own recorded `startedAt` was positioned to answer and never does. Worse than absent, the value supplies the appearance of the answer: a record carrying a timestamp reads as a verified identity to every reader, human or code, that does not go looking for the comparison that does not exist.

## What is already right, which is what makes the gap easy to miss

This is not a repo missing a safety culture; it is a repo whose stops stop one question short.

- `processStartToken` in `packages/workspace/src/advisory-lock.ts:68` IS the real pid-reuse check on Linux and macOS: `/proc/<pid>/stat` field 22 and `ps -o lstart=`, recorded at acquire and compared at inspect (`advisory-lock.ts:156`). And it is not sitting unused in one file: the extension mutation markers (`packages/workspace/src/extension-mutation.ts:135`, `:148`, `:208`) and the seed reconcile lock (`implementations/cli/src/seed/lock.ts:113`, `:162`) record and compare start tokens the same way. Every consumer is a lock or mutation coordination, and no stop path has one. That is the sharpest form of the gap: the repo reaches for this check whenever it needs to know whether a recorded owner is still real, and never reaches for it before sending a signal.
- The liveness probe is a deliberate tri-state (`packages/workspace/src/pid.ts:62`): only ESRCH proves death, EPERM means alive under another user, everything else is `unknown` and never acted on as death.
- `stopManager` reads argv before signalling on POSIX (`implementations/cli/src/lib/manager-proc.ts:281`) and refuses when the live pid is provably not a manager: "A LIVE pid that is provably NOT a manager is never signalled."

So the permissive sites below sit in the middle of a codebase that refuses on weaker doubts everywhere else. The gap is easy to miss precisely because the surrounding code is good: the hard parts (torn writes, errno semantics, mutual exclusion between concurrent stoppers) are done, and `startedAt`'s paper trail looks like the last one. What remains is the one fact that requires cooperation from the OS rather than from our own bookkeeping.

## The sites

Two rows below cite `origin/feat/up-daemonize-persist` rather than `main`, because those files do
not exist at HEAD yet: they arrive with an unmerged branch, so those citations will move with it and
lapse entirely if it never lands. The pidfile rows stand on `main` on their own, and the argument
does not depend on the Windows rows.

| Site | What it does today | What it cannot prove |
|---|---|---|
| `packages/workspace/src/pid.ts:62` `probeLiveness` | `process.kill(pid, 0)` plus the errno mapping | Which process the number names. It answers existence, never identity. |
| `packages/workspace/src/pid.ts:111` `readProcessCommand` | argv via `/proc` or `ps` on POSIX; hardcoded `unreadable` on win32 (line 112) | Any identity fact on win32, by explicit decision (see the win32 section) |
| `packages/workspace/src/maintenance.ts:731` `localProcessOwnerStatus` | `kill(pid, 0)` and a hostname check; `startedAt` sits in the same record (`maintenance.ts:46`, written at `:728`) | That the live pid is the recorded process. The recorded start time is never compared against the OS's own start time for that pid; `startedAt` is only ever compared record-against-record (`maintenance.ts:1109`), never against the process |
| `implementations/cli/src/commands/down.ts:237` `stopLocalProcess` | The generic stop for every registered component (manager, delivery, auth, broker, web, extension processes): parse pid, coordinate the `.stopping` marker, SIGTERM (line 314), SIGKILL (line 332), await ESRCH | Attribution of any kind. No `readProcessCommand`, no start token, no owner comparison anywhere in its body |
| `implementations/cli/src/lib/auth-proc.ts:224` `stopAuthService` | parse, probe, SIGTERM, await confirmed death | The pid's identity. No argv check even on POSIX, though argv reading is imported one file over in manager-proc |
| `implementations/cli/src/lib/delivery-proc.ts:243` `stopDelivery` (on `origin/feat/up-daemonize-persist`) | same shape as `stopAuthService` | same |
| `implementations/cli/src/lib/detached-spawn.ts:34` `windowsDetachedChild` (on `origin/feat/up-daemonize-persist`) | returns a pid-only `ChildProcess` whose `kill()` forwards to `process.kill(pid, signal)` (line 41) after the native launcher closed its process handle | The identity half of the shape it already names. `assertDetachedChildExitObservable` (line 52) handles the observability half ("A detached child may be signalable without being observable"); identity has no seam at all |
| `implementations/cli/src/lib/isolated-broker.ts:362` `stopByPid` | signals pids read from the durable handshake file after an `alive()` check | Identity; and its `ProcessOwner.startedAt` (`isolated-broker.ts:405`) is stamped with the time the coordinator READ the handshake, not the process's start, so it could not satisfy a real start check even if one existed |
| `implementations/cli/src/commands/down.ts:368` `preserveStateDown` | finds this root's mesh with `mesh.root === root`, a raw string equality on filesystem paths, on the `--preserve-state` path | That the two spellings name the same directory. `meshesForRoot` (`packages/workspace/src/mesh-registry.ts:276`) exists for exactly this and warns that "a raw `===` silently misses" (symlinked roots, macOS `/var` versus `/private/var`, Windows 8.3 short names); the plain `down` path already routes through it (`down.ts:190`, via `removeMeshesByRoot`), target resolution applies the same canonicalized predicate (`packages/workspace/src/mesh-target.ts:364`), and only this call bypasses it. Consequence is misidentification (a recorded mesh reads as absent, so the recovery verb refuses where `found 0`), not a signal |

Callers worth naming: the Ctrl-C teardown of a foreground `up` stops the delivery daemon, manager and auth service by pidfile (`implementations/cli/src/commands/up.ts:961`), under its own comment "All kill by pidfile, symmetric" (`up.ts:955`), and only the manager leg has the argv check. `cotal down manager` does not go through `stopManager` at all; it goes through the generic `stopLocalProcess`, so even on POSIX the argv check does not cover the documented stop verb. The presence direction carries the same unproven claim without the destruction: `implementations/web/src/web.ts:216` treats a live prior pid as "web dashboard is already running", which pid reuse turns into a false-up.

## Why this is one class, not two issues

The pidfile paths and the Windows handle are the same defect on both sides of one spawn. The daemon side writes a record carrying only a number; the stop side signals a number. The Windows launcher in `detached-spawn.ts` is the newest and purest instance of the class: to survive the parent it gives up the process handle, the one identifier the OS guarantees not to reuse, keeps the number, and hands back an object whose `kill` is `process.kill(pid, signal)`. The file already contains the shape of the problem half-recognized: the seam's doc says the handle is returned "after the native Windows launcher closes its process handle", and `assertDetachedChildExitObservable` refuses to treat absent event methods as evidence of exit. The observability half of "a pid without a handle" got a contract; the identity half got nothing.

Filing these as two issues would be the wrong cut for a concrete reason: a reader with two issues closes the cheap one and feels finished. The pidfile half looks like plumbing (add a token beside the pid on POSIX), and closing it alone leaves every Windows stop, including the new launcher path, signalling pids it cannot attribute while the class reads as handled. Closing the Windows half alone leaves `cotal down`, the highest-frequency destructive verb in the CLI, signalling unproven pids on every platform. One class, one issue: a signal sent from a record requires the record's identity claim to be checked against the live process, everywhere the repo sends one.

A fourth surface belongs in the same class, kept distinct from the signalling paths because its consequence is misidentification rather than a signal: `preserveStateDown` compares a recorded root by raw `===` (`down.ts:368`) where the registry's own doc says that comparison "silently misses" and ships the canonicalized one this call bypasses (`mesh-registry.ts:274`, `mesh-target.ts:364`). A recorded root compared by an equality that cannot detect the case it exists to catch, a start time treated as verified because it was written: the same habit wearing a different record.

## The rule this repo already follows

Nothing here needs a new doctrine. The repo's existing rule, applied uniformly, covers every site: when identity cannot be established, refuse the destructive action.

- `down.ts:135`: "Fail CLOSED: an unselected dependant that MIGHT be running (including one behind an unattributable/unknown pidfile) blocks a stop-last shutdown".
- `stopLocalProcess` itself refuses unattributable content (`down.ts:309`), a refused signal (`down.ts:325`), and an unconfirmed death (`down.ts:336`), preserving the record each time.
- `assertManagerRecordReplaceable` (`manager-proc.ts:174`) refuses to start over an `unknown` or `unattributable` record.
- `advisory-lock.ts` treats an unobtainable token as active/unknown: "NEVER reclaimed on the missing token alone; only a PID-dead owner is reclaimable" (`advisory-lock.ts:28`). This is the repo making a recorded identity load-bearing: a token that no longer matches means recycled, and matching is checked against the live process, not against another record.
- `materializeSecretToFile` (`packages/workspace/src/auth-paths.ts:388`) fails loud on an absent key rather than materializing an empty credential.
- `maintenance.ts:1141` throws `listener-owner-ambiguous` when ownership cannot be proven locally.

The contrast that sharpens the rule: `divergentCwdAnchor` (`auth-paths.ts:429`) is deliberately a REPORTER that never throws and never chooses, and that is correct, because it is a diagnostic with no destructive action behind it ("a detector that can abort its caller is that same defect once more, wearing the cwd's failure instead of its answer"). The distinction is not "silence bad, throwing good"; it is that a destructive action requires proof its target is what the record says. Every stop path in the table has a destructive action behind it, and none of them (except `stopManager` on POSIX) has the proof.

## Suggested fix shape: make the recorded value load-bearing

The remedy is not "add a check" at N sites; it is to make the value that is already recorded answer for itself, and to refuse the destructive action when it cannot be compared. The shape is the advisory-lock mechanism generalized: at write time, record the process start token next to the pid in every pidfile the stop paths read; at stop time, compare the recorded token against the live process and refuse to signal on mismatch, which is verbatim the rule at `advisory-lock.ts:156` ("A recorded start token that no longer matches means the PID was recycled: the original owner is gone (stale)."). One shared identity contract beside `pid.ts`, consumed by `stopLocalProcess`, `stopAuthService`, `stopDelivery`, `stopManager`, and the launcher's `windowsDetachedChild`, rather than five site-local patches. The migration question (what a reader does with a legacy bare pid, where no token was ever recorded) is part of the design, not something to settle in this issue. On POSIX, argv attribution can remain the manager's extra check. No patch is proposed here on purpose; the win32 half has an open question first.

## The win32 question is open, and the existing position is argued

The gap on win32 is not an oversight being caught; it is a reasoned position being revisited. The rationale is written out at `advisory-lock.ts:69`:

> Windows has no cheap, STABLE start token here: `/proc` is absent and a `ps` on PATH (MSYS/Git) returns inconsistent output between calls, which would make the SAME live lock read differently from the acquirer vs a reader and be misjudged stale. Per the no-token rule, return undefined so a live PID keeps the lock active/unknown (bounded-wait is the backstop) rather than a false reclaim.

And `readProcessCommand`'s doc carries the same argument for argv (`pid.ts:106`): a platform that cannot look "behaves exactly as every platform did before this existed", because attribution may only downgrade a record on affirmative evidence that the live process is something else.

Two things change when the same question is asked in front of a signal rather than a lock. First, the lock's safety argument does not transfer: an absent token is safe there because a bounded wait is a real backstop, while a stop that proceeds on an unproven pid has nothing behind it; the destructive act itself is the fallback, so the thing being degraded is someone else's process. Second, the requirement is different: the lock polls, so its token must be cheap; a stop runs once, so its token only needs to be stable. And a stable win32 start time is not unobtainable, it is merely not free: the launcher itself held the process handle a moment before closing it (`detached-spawn.ts:92` closes `pi.hProcess` in the same script that returns the pid), and the PowerShell `Add-Type` seam it already uses can read a process creation time without introducing new machinery of a different kind.

Does the no-fallbacks doctrine ("Throw if something is not supported in the current environment or config, rather than silently degrading", `AGENTS.md:161`) already decide this? Mostly yes, with one honest counterargument. Proceeding when identity cannot be established is a silent degradation: the stop reports a clean teardown over a signal sent to an unverified number, which is the shape the doctrine exists to prevent. The counterargument is the repo's own precedent that "cannot be represented on this platform" is sometimes not a refusal: `manager-proc.ts:127` treats a chmod a Windows volume cannot represent as not a reason to refuse starting the manager. But that precedent covers hygiene around an action (a defence-in-depth file mode), not the identity of the target of the action; a kill whose target is unproven is a different category from a log whose mode is loose. So the reading holds on the stop side, and it would make the stop refuse on win32 until a token mechanism exists, exactly as it refuses on every weaker doubt today. What the doctrine does not settle is whether that refusal ships as the interim behaviour or arrives together with a win32 token, and that tradeoff (a platform whose stops currently work, versus the correctness of a destructive verb) is the open design decision.

## Why this is worth fixing rather than noting

A stop that signals a pid it cannot attribute is indistinguishable, in every line it prints, from one that can, right up until the number belongs to something else. The failure is destructive outside the repo's blast radius: the process at the other end belongs to the operator and did nothing wrong. The paths involved are not rare corners; `cotal down` is the documented heal verb, and the Ctrl-C teardown of a foreground `up` runs the pidfile stops on every dev session. The reuse window is narrow, but this codebase already pays the cost of refusing on every other narrow window (torn pidfiles, seccomp errnos, refused signals), and on Linux the check is already written, tested, and sitting in `advisory-lock.ts`. Windows is where the real decision is, and the new detached launcher is about to make the class larger: every Windows stack it starts will be stopped through a handle-shaped object whose only identity is a number that was never asked to prove itself.

davidfarah2003:
Closes #909.

A pidfile recorded a number and the stop paths signalled that number. `kill(pid, 0)` answers existence, never identity. After wraparound the number can name a stranger; SIGTERM then SIGKILL hit that stranger.

## What now happens

`processStartToken` (Linux `/proc/<pid>/stat` field 22, macOS `lstart`) is the OS clock for that pid, not our `Date.now()`. `signalRecordedProcess` is the only signal seam: recycled tokens refuse, unproven token-bearing records refuse, legacy pid-only records remain signalable so a pre-token pidfile is not wedged.

`stopLocalProcess`, `stopManager`, `stopAuthService`, and `stopDelivery` consume it. Advisory-lock now imports the same token instead of a second copy.

## Proof

- `pnpm smoke:pid-contract` 102/102 (this box, sequential, no extra parallel build).
- `node scripts/mutation-proof.mjs --config packages/workspace/smoke/fixtures/pid-attribution.mutations.json`: 5/5 KILLED red-and-named, including:
  - treating a different start token as `same`
  - `signalRecordedProcess` skipping the match and always sending
  Named cell: "a recycled pid is REFUSED and NEVER signalled (would kill a stranger)" (93 vs baseline 102).

Recycled-pid discrimination is constructed (injected tokens). Waiting for live OS pid wraparound is impractical and named in the fixture `why`.

## Both sites

- **Pidfile stop paths (this PR, on main):** proven.
- **Windows detached handle:** NOT on this main. `implementations/cli/src/lib/detached-spawn.ts` arrives with #880 (`feat/up-daemonize-persist`). `processStartToken` is undefined on win32 by design (no stable `/proc` starttime; MSYS `ps` is not stable). Named in the suite footer, not skipped. A pid-only `ChildProcess.kill` remains a bare `process.kill(pid)` until that file exists and consumes `signalRecordedProcess`.

## Coordination with #969

Messaged w969 before designing. No reply in the window. #969 is creation-identity / open-verify-terminate on the Windows detached lifecycle; this PR is verify-before-signal on pidfile stops that already exist on main. If they collide, orch should rule; I did not build the Windows handle half.

## Not proved

- Live OS pid wraparound (constructed tokens instead).
- Windows handle terminate-through-matched-handle (file not on main).
- Writing the token into every pidfile at spawn time (legacy pid-only remains signalable; the recycled refuse requires a token that was recorded).
- `preserveStateDown` raw-`===` root comparison (issue table, not a signal).

Do not merge.

### DIFF
--- a/.changeset/pid-identity-before-signal.md
+++ b/.changeset/pid-identity-before-signal.md
@@ -1,0 +1,7 @@
+---
+"@cotal-ai/workspace": patch
+"@cotal-ai/cli": patch
+"@cotal-ai/manager": patch
+---
+
+Refuse to signal a pid whose recorded start token no longer matches the live process (recycled pid), instead of treating a written-down number as identity.
--- a/implementations/cli/src/commands/down.ts
+++ b/implementations/cli/src/commands/down.ts
@@ -52,7 +52,7 @@ import {
 import { extensionNames, localProcessSurface } from "../ext-loader.js";
 import { c } from "../ui.js";
 import { cotalRoot } from "../lib/paths.js";
-import { parsePid, probeLiveness } from "@cotal-ai/workspace";
+import { parsePid, parseProcessIdentity, probeLiveness, signalRecordedProcess } from "@cotal-ai/workspace";
 import { resolveSpace } from "../lib/status.js";
 import { downManifest } from "./down-manifest.js";
 import { askManager, resolveControlTarget } from "../lib/control.js";
@@ -248,7 +248,8 @@ export async function stopLocalProcess(component: LocalProcess, context: LocalPr
         : `${component.name} has a stale extension-removal reservation at ${pidPath} - remove that file and retry`,
     );
   }
-  const pid = parsePid(rawPid);
+  const identity = parseProcessIdentity(rawPid);
+  const pid = identity?.pid;
   const marker = `${pidPath}.stopping`;
   let markerFd: number | undefined;
   for (;;) {
@@ -310,9 +311,12 @@ export async function stopLocalProcess(component: LocalProcess, context: LocalPr
         `${component.label} has an unattributable pidfile at ${pidPath} (${JSON.stringify(rawPid)}) - it may still front a running process; refusing to remove it or report a clean stop. Stop that process and remove the file manually.`,
       );
     }
+    const recorded = { pid, ...(identity.start !== undefined ? { start: identity.start } : {}) };
     try {
-      process.kill(pid, "SIGTERM");
+      signalRecordedProcess(recorded, "SIGTERM");
     } catch (e) {
+      const msg = (e as Error).message ?? "";
+      if (/no longer matches|cannot be compared/.test(msg)) throw e;
       // ONLY an ESRCH proves the process is already gone (safe to clear). EPERM (exists, another
       // user's) or any other errno (EIO, argument errors, unknown) means we could NOT confirm death
       // - reporting it stopped and removing its record would orphan a live process. Fail loud, keep
@@ -329,7 +333,7 @@ export async function stopLocalProcess(component: LocalProcess, context: LocalPr
     const graceDeadline = Date.now() + 15_000;
     while (probeLiveness(pid) !== "dead" && Date.now() < graceDeadline) await sleep(100);
     if (probeLiveness(pid) !== "dead") {
-      try { process.kill(pid, "SIGKILL"); } catch { /* raced to exit */ }
+      try { signalRecordedProcess(recorded, "SIGKILL"); } catch { /* raced to exit, or identity no longer matches */ }
       const hardDeadline = Date.now() + 3_000;
       while (probeLiveness(pid) !== "dead" && Date.now() < hardDeadline) await sleep(100);
     }
--- a/implementations/cli/src/lib/auth-proc.ts
+++ b/implementations/cli/src/lib/auth-proc.ts
@@ -12,7 +12,7 @@ import { closeSync, existsSync, ftruncateSync, linkSync, openSync, readdirSync, 
 import { basename, dirname } from "node:path";
 import { type AuthPrepared } from "@cotal-ai/core";
 import { spaceKey } from "@cotal-ai/workspace";
-import { parsePid, probeLiveness, type LivenessProbe } from "@cotal-ai/workspace";
+import { parsePid, parseProcessIdentity, probeLiveness, signalRecordedProcess, type LivenessProbe } from "@cotal-ai/workspace";
 import type { SignalFn } from "./manager-proc.js";
 
 import { selfArgv } from "./self-exec.js";
@@ -216,14 +216,17 @@ export async function stopAuthService(space: string, probe: LivenessProbe = prob
     rmSync(p, { force: true }); // empty husk: no process to signal
     return;
   }
-  const pid = parsePid(trimmed);
+  const identity = parseProcessIdentity(trimmed);
+  const pid = identity?.pid;
   if (pid === undefined)
     throw new Error(
       `auth-service pidfile ${p} holds unattributable content ${JSON.stringify(trimmed)} - refusing to remove a record for a process it cannot identify or signal; inspect or remove it manually`,
     );
   try {
-    send(pid, "SIGTERM");
+    signalRecordedProcess({ pid, ...(identity.start !== undefined ? { start: identity.start } : {}) }, "SIGTERM", { send });
   } catch (e) {
+    const msg = (e as Error).message ?? "";
+    if (/no longer matches|cannot be compared/.test(msg)) throw e;
     // ESRCH = already gone, so removing its record is safe. EPERM (exists, another user's) or any
     // other errno means we could NOT stop it - removing the record would orphan a live signer while
     // reporting a clean stop, so preserve it and fail loud.
--- a/implementations/cli/src/lib/delivery-proc.ts
+++ b/implementations/cli/src/lib/delivery-proc.ts
@@ -7,7 +7,7 @@ import {
   newIdentity,
   waitForDeliveryLease,
 } from "@cotal-ai/core";
-import { DELIVERY_CREDS_KIND, authDir, deliveryCredsKey, findCotalRoot, getSoleSpaceAuth, listSpaceAccounts, parsePid, probeLiveness, segmentedKey, type LivenessProbe, workspaceSecretStore } from "@cotal-ai/workspace";
+import { DELIVERY_CREDS_KIND, authDir, deliveryCredsKey, findCotalRoot, getSoleSpaceAuth, listSpaceAccounts, parsePid, parseProcessIdentity, probeLiveness, segmentedKey, signalRecordedProcess, type LivenessProbe, workspaceSecretStore } from "@cotal-ai/workspace";
 import { selfArgv } from "./self-exec.js";
 import { resolveSpace } from "./status.js";
 import { cotalPath } from "./paths.js";
@@ -228,7 +228,8 @@ export async function stopDelivery(probe: LivenessProbe = probeLiveness, signal?
     return;
   }
   const raw = readFileSync(p, "utf8").trim();
-  const pid = parsePid(raw);
+  const identity = parseProcessIdentity(raw);
+  const pid = identity?.pid;
   if (pid === undefined) {
     if (raw === "") {
       await removeRecords(); // pre-protocol husk, nothing behind it
@@ -254,8 +255,10 @@ export async function stopDelivery(probe: LivenessProbe = probeLiveness, signal?
         `NEXT: verify with \`ps -p ${pid}\`, then stop it yourself or remove \`.cotal/delivery.pid\` if it is gone.`,
     );
   try {
-    send(pid, "SIGTERM");
+    signalRecordedProcess({ pid, ...(identity.start !== undefined ? { start: identity.start } : {}) }, "SIGTERM", { send });
   } catch (e) {
+    const msg = (e as Error).message ?? "";
+    if (/no longer matches|cannot be compared/.test(msg)) throw e;
     const code = (e as NodeJS.ErrnoException).code;
     if (code === "ESRCH") {
       await removeRecords(); // raced to exit between probe and signal
--- a/implementations/cli/src/lib/manager-proc.ts
+++ b/implementations/cli/src/lib/manager-proc.ts
@@ -5,8 +5,8 @@ import { selfArgv } from "./self-exec.js";
 import { resolveSpace } from "./status.js";
 import { cotalPath } from "./paths.js";
 import {
-  commandIsCotalSupervisor, parsePid, probeLiveness, readProcessCommand,
-  MANAGER_DELIVERY_AWARE_MARKER, MANAGER_PIDFILE, type CommandReader, type LivenessProbe,
+  commandIsCotalSupervisor, parsePid, parseProcessIdentity, probeLiveness, readProcessCommand,
+  signalRecordedProcess, MANAGER_DELIVERY_AWARE_MARKER, MANAGER_PIDFILE, type CommandReader, type LivenessProbe,
 } from "@cotal-ai/workspace";
 
 /** Exported so the delivery cutover preflight can NAME the pid it refused on: an error that says
@@ -250,7 +250,8 @@ export async function stopManager(
     return "already-gone";
   }
   const raw = readFileSync(p, "utf8").trim();
-  const pid = parsePid(raw);
+  const identity = parseProcessIdentity(raw);
+  const pid = identity?.pid;
   if (pid === undefined) {
     // An EMPTY pidfile is a pre-protocol husk with nothing behind it, and clearing it is safe. ANY
     // OTHER unattributable content (garbled, fractional, out of range) may still front a LIVE
@@ -293,8 +294,10 @@ export async function stopManager(
         `NEXT: verify with \`ps -p ${pid}\`, then stop it yourself or remove \`.cotal/manager.pid\` if it is gone.`,
     );
   try {
-    send(pid, "SIGTERM");
+    signalRecordedProcess({ pid, ...(identity.start !== undefined ? { start: identity.start } : {}) }, "SIGTERM", { send });
   } catch (e) {
+    const msg = (e as Error).message ?? "";
+    if (/no longer matches|cannot be compared/.test(msg)) throw e;
     const code = (e as NodeJS.ErrnoException).code;
     if (code === "ESRCH") {
       clear(); // died between the probe and the signal, which is an ordinary race and genuinely gone
--- a/packages/workspace/smoke/fixtures/pid-attribution.mutations.json
+++ b/packages/workspace/smoke/fixtures/pid-attribution.mutations.json
@@ -28,7 +28,23 @@
       "find": "  if (r.status === 0 && out !== \"\") return { kind: \"command\", command: out };",
       "replace": "  if (r.status === 0 && out !== \"\") return { kind: \"unreadable\", why: \"mutant\" };",
       "expectRed": "this very process is readable and is NOT a manager"
+    },
+    {
+      "name": "PREDICTED KILLED (>=1 red) - a recycled pid (alive, different start token) is treated as the same process and signalled",
+      "file": "packages/workspace/src/pid.ts",
+      "find": "  const token = arguments.length >= 2 ? liveToken : processStartToken(recorded.pid);\n  if (recorded.start === undefined || token === undefined) return \"unknown\";\n  return token === recorded.start ? \"same\" : \"recycled\";",
+      "replace": "  const token = arguments.length >= 2 ? liveToken : processStartToken(recorded.pid);\n  if (recorded.start === undefined || token === undefined) return \"unknown\";\n  return \"same\";",
+      "expectRed": "THE DISCRIMINATING CASE: a live pid with a DIFFERENT start token is recycled, not merely dead",
+      "cell": "THE DISCRIMINATING CASE: a live pid with a DIFFERENT start token is recycled, not merely dead"
+    },
+    {
+      "name": "PREDICTED KILLED (>=1 red) - signalRecordedProcess skips the identity match and always sends",
+      "file": "packages/workspace/src/pid.ts",
+      "find": "  const match = matchProcessIdentity(recorded, tokenOf(recorded.pid));\n  if (match === \"recycled\")\n    throw new Error(`recorded pid ${recorded.pid} is alive but its start token no longer matches; the original process is gone and the number was reused - refusing to signal a stranger`);\n  if (match === \"unknown\" && recorded.start !== undefined)\n    throw new Error(`recorded pid ${recorded.pid} carries a start token that cannot be compared against the live process; refusing to signal an unproven identity`);\n  send(recorded.pid, signal);",
+      "replace": "  send(recorded.pid, signal);",
+      "expectRed": "a recycled pid is REFUSED and NEVER signalled (would kill a stranger)",
+      "cell": "a recycled pid is REFUSED and NEVER signalled (would kill a stranger)"
     }
   ],
-  "_note": "The suite imports `../src/pid.js` and `../../../implementations/cli/src/lib/manager-proc.js` RELATIVELY, so both mutated files are the modules under test and no build step belongs here. Mutation 4 targets the POSIX `ps` branch, which is the one this machine executes; the linux `/proc` branch and the win32 branch are not covered by these four and neither is claimed to be."
+  "_note": "The suite imports `../src/pid.js` and `../../../implementations/cli/src/lib/manager-proc.js` RELATIVELY, so both mutated files are the modules under test and no build step belongs here. Recycled-pid cells inject tokens rather than waiting for OS pid wraparound, which is impractical. win32 has no stable start token; the Windows detached handle is not on this main (#880)."
 }
--- a/packages/workspace/smoke/fixtures/pid-attribution.mutations.json
+++ b/packages/workspace/smoke/fixtures/pid-attribution.mutations.json
@@ -23,13 +23,6 @@
       "expectRed": "a live pid that is NOT a manager reads as FOREIGN, not as a healthy manager"
     },
     {
-      "name": "PREDICTED KILLED (>=1 red) - the real process-argv read is replaced by an unreadable answer, so nothing is ever attributed and every live pid stays trusted: the cell that names a stranger has to be executing the actual reader on this platform to notice",
-      "file": "packages/workspace/src/pid.ts",
-      "find": "  if (r.status === 0 && out !== \"\") return { kind: \"command\", command: out };",
-      "replace": "  if (r.status === 0 && out !== \"\") return { kind: \"unreadable\", why: \"mutant\" };",
-      "expectRed": "this very process is readable and is NOT a manager"
-    },
-    {
       "name": "PREDICTED KILLED (>=1 red) - a recycled pid (alive, different start token) is treated as the same process and signalled",
       "file": "packages/workspace/src/pid.ts",
       "find": "  const token = arguments.length >= 2 ? liveToken : processStartToken(recorded.pid);\n  if (recorded.start === undefined || token === undefined) return \"unknown\";\n  return token === recorded.start ? \"same\" : \"recycled\";",
@@ -46,5 +39,5 @@
       "cell": "a recycled pid is REFUSED and NEVER signalled (would kill a stranger)"
     }
   ],
-  "_note": "The suite imports `../src/pid.js` and `../../../implementations/cli/src/lib/manager-proc.js` RELATIVELY, so both mutated files are the modules under test and no build step belongs here. Recycled-pid cells inject tokens rather than waiting for OS pid wraparound, which is impractical. win32 has no stable start token; the Windows detached handle is not on this main (#880)."
+  "_note": "The suite imports `../src/pid.js` and `../../../implementations/cli/src/lib/manager-proc.js` RELATIVELY, so both mutated files are the modules under test and no build step belongs here. Recycled-pid cells inject tokens rather than waiting for OS pid wraparound, which is impractical. win32 has no stable start token; the Windows detached handle is not on this main (#880). The POSIX `ps` argv-reader mutation is omitted here because this host is Linux and executes /proc, not ps; mutating the ps branch SURVIVED (102/102) and cannot be attributed."
 }
--- a/packages/workspace/smoke/pid.smoke.ts
+++ b/packages/workspace/smoke/pid.smoke.ts
@@ -25,7 +25,7 @@ import { existsSync, mkdtempSync, mkdirSync, readFileSync, writeFileSync, rmSync
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
-  commandIsCotalSupervisor, livenessFromErrno, parsePid, probeLiveness, readProcessCommand,
+  commandIsCotalSupervisor, formatProcessIdentity, livenessFromErrno, matchProcessIdentity, parsePid, parseProcessIdentity, probeLiveness, processStartToken, readProcessCommand, signalRecordedProcess,
   type CommandReader,
 } from "../src/pid.js";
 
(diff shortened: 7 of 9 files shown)


davidfarah2003 (commented):
- The PR body's own last line is "Do not merge." That is an explicit author hold on this PR; it must not land regardless of code quality until that is lifted.
- Body also flags unresolved cross-PR coordination: messaged #969's owner about a possible collision on Windows detached-handle lifecycle, "No reply in the window... If they collide, orch should rule" — an open coordination question, not yet settled.
- Scope gaps are self-disclosed and real, not hidden: Windows terminate-through-matched-handle is admittedly untested (detached-spawn.ts is not on this main; arrives with #880), and recycled-pid discrimination is constructed via injected tokens rather than live OS wraparound (named in the fixture, not concealed).
- Modulo the merge hold, the technical work itself looks sound: signalRecordedProcess (packages/workspace/src/pid.ts:478-490) is the sole signal seam, refuses both "recycled" and unproven "unknown" identities and throws loud rather than silently falling back to a bare kill; all four call sites (down.ts, auth-proc.ts, delivery-proc.ts, manager-proc.ts) route through it and re-throw on the new identity-mismatch errors instead of swallowing them into the old ESRCH/EPERM handling.
- pid.smoke.ts:275-311 exercises the real stopManager/stopLocalProcess entry points end-to-end (writes a recycled-token pidfile, asserts the throw, asserts the pidfile is preserved and the live process is not signalled) — genuine both-polarity coverage of shipped code, not a mock.
- Do not approve or land until the author lifts the "Do not merge" hold and the #969 coordination question is resolved.

davidfarah2003 (commented):

- PR body ends with an explicit "Do not merge." — the author has flagged this as not landing-ready (pending #969 coordination on the Windows detached-handle half and orch's ruling on any collision). That alone blocks landing regardless of code quality.
- Body itself documents unresolved scope: the Windows side (implementations/cli/src/lib/detached-spawn.ts, arriving with #880) is explicitly NOT covered — processStartToken is undefined on win32 by design, and a pid-only ChildProcess.kill there remains a bare process.kill(pid) until that lands.
- Author messaged #969 before designing but got no reply in the window; cross-PR identity/lifecycle overlap is unresolved, per the PR's own "Coordination with #969" section.
- On the merits, the shipped code itself is solid: signalRecordedProcess (packages/workspace/src/pid.ts:478-491) refuses both 'recycled' and unproven 'unknown+token' matches loudly (thrown, named errors, no silent fallback), legacy pid-only records stay signalable, and all four call sites (down.ts, auth-proc.ts, delivery-proc.ts, manager-proc.ts) re-throw those refusals instead of swallowing them into the ESRCH/EPERM handling. Both-polarity smoke coverage (pid.smoke.ts) and named mutation-proof cells (pid-attribution.mutations.json) exercise the real matchProcessIdentity/signalRecordedProcess code paths.
- No ci-suites.txt touch to check for tail-append; no AI/tool attribution found.
- Recommend re-review once the author lifts the "Do not merge" hold and confirms resolution with #969.

davidfarah2003:
Exact head graded: 93052d9d8d2526e0f5733d3dceb8c9cce6b0ec09
`git rev-parse HEAD` in detached worktree Cotal-rev1057. `gh pr view` was rate-limit 403 at grade time. Load 22.06, 24.97, 26.30. Did not run `cotal up`. GitHub comment 403'd; this is the same terminal verdict.

## Verdict: APPROVE

A PID is not an identity here. `processStartToken` on Linux is `/proc/<pid>/stat` field 22 (`starttime`, ticks since boot): the kernel clock for that pid, not `Date.now()`, not pidfile mtime. Parse is `lastIndexOf(") ")` then split field index 19 of the remainder, which is stat field 22. `matchProcessIdentity` returns `same` only when both recorded.start and the live token exist and are equal. `recycled` is live token present and different. Missing recorded token or missing live token is `unknown`, never `recycled`. Pidfile mtime is not read anywhere on this path.

`signalRecordedProcess` (`packages/workspace/src/pid.ts:228-240`) is the sole signal seam. Recycled throws and does not send. Unknown with a recorded token throws and does not send (unproven identity). Legacy pid-only (`start` absent) still signals. There is no tokenless overload.

Failure direction: a matching live identity is still signalled. `pid.smoke.ts` writes a live process's real start token into the record, then `signalRecordedProcess` with SIGCONT succeeds. Recycle refusal is the other polarity: injected mismatched token, throw, pidfile preserved, live process not signalled. Both polarities exist. I ran `pnpm smoke:pid-contract` (102/102) and in-scope mutations 5/5 KILLED. Did not run `pnpm smoke:ci` or any `*:live` suite.

Named gaps, not overclaim:
- Recycle discrimination is constructed by injecting tokens. No fixture here constructs a real kernel pid wraparound (same number, different starttime from OS reuse).
- Spawn still writes pid-only: `up.ts` `String(child.pid)`, `auth-proc.ts` / `delivery-proc.ts` / `manager-proc.ts` the same. Recycle refusal only applies after a token is recorded. The PR names this. A pidfile written by current spawn remains the legacy signalable case.

Windows: `processStartToken` is undefined, so recorded tokens cannot be proven and a token-bearing record would refuse. Current writers are pid-only, so they stay on the legacy path.

APPROVE 93052d9d8d2526e0f5733d3dceb8c9cce6b0ec09

davidfarah2003:
> **Superseded as a relay — the reviewer's own verdict landed first.** `v1045` posted directly at
> 09:59:31Z, 40 seconds before this comment, so its 403 had cleared by then and my relay was
> unnecessary. **Its comment is the authoritative one**; treat this as a duplicate of the verdict
> text. I am leaving it up only for the "Merge status" section at the end, which the reviewer's
> comment does not cover and which is mine rather than theirs.

## Terminal verdict from the assigned reviewer (`v1045`), relayed

Posting this on the reviewer's behalf: its GitHub token returned 403 at grade time, which on this
repository is the burst limiter rather than a quota — every seat shares one identity, so no
reviewer has a token of its own to fall back to. **The review, the verdict and every measurement
below are the reviewer's, not mine.** I contributed nothing to it and did not re-derive it; I am
carrying it into the durable record so a throttled token cannot leave this PR looking unreviewed.

---

## APPROVE `93052d9d8d2526e0f5733d3dceb8c9cce6b0ec09`

A PID is not an identity here. `processStartToken` on Linux is `/proc/<pid>/stat` field 22
(starttime, ticks since boot): the kernel clock for that pid, not `Date.now()`, not pidfile mtime.
Parse is `lastIndexOf(") ")` then split field index 19 of the remainder.

`matchProcessIdentity` returns `same` only when both `recorded.start` and the live token exist and are
equal. `recycled` is a live token present and different. A missing recorded token or a missing live
token is `unknown`, never `recycled`. Pidfile mtime is not read on this path.

`signalRecordedProcess` (`packages/workspace/src/pid.ts:228-240`) is the sole signal seam. Recycled
throws and does not send. Unknown with a recorded token throws and does not send. Legacy pid-only
(`start` absent) still signals. There is no tokenless overload.

**Both polarities are proven, not just the refusal.** `pid.smoke.ts` writes a live process's real
start token into the record, then `signalRecordedProcess` with `SIGCONT` succeeds - so a matching
live identity is still signalled. The recycle refusal is the other polarity: injected mismatched
token, throw, pidfile preserved, live process not signalled.

`pnpm smoke:pid-contract` 102/102. In-scope mutations 5/5 KILLED. Did not run `cotal up`,
`smoke:ci`, or any `*:live` suite.

### Named gaps, stated rather than overclaimed

- Recycle discrimination is constructed by injecting tokens. No fixture here constructs a real kernel
  pid wraparound - the same number with a different starttime from genuine OS reuse.
- Spawn still writes pid-only: `up.ts`, `auth-proc.ts`, `delivery-proc.ts`, `manager-proc.ts` all
  write `String(child.pid)`. **Recycle refusal only applies after a token is recorded**, so the
  protection does not engage on today's spawn paths. The PR names this.
- Windows: `processStartToken` is undefined, and current writers are pid-only, so they stay on the
  legacy signalable path.

---

## Merge status: this APPROVE does not clear it to land

Recorded so nobody reads a green verdict as a merge signal. #1057 is one of four open PRs that
contend on process identity, and a semantic read across the shared files found that **#1057 and #1069
implement the same decision point with incompatible storage designs** - identity inline in the pidfile
here, versus a sidecar pin file there. Landing both would leave two identity records per process that
can disagree, so one of the two designs should be closed rather than sequenced.

That choice is a design decision the repository owner has not yet made. This verdict establishes that
this design is **sound**; it does not establish that it is the **chosen** one. Those are different
questions and only the first is a reviewer's to answer.

The other two are unaffected: #880 touches these files only to swap `spawn+unref` for
`spawnDetached` at two launch sites, so it is orthogonal, and #1103 is a namespacing layer that
refolds after whichever identity design wins, at the same cost either way.

davidfarah2003:
Refold onto current main conflicts in `implementations/cli/src/commands/down.ts`, and the overlap runs deeper than the one file: this branch reworks `packages/workspace/src/pid.ts` (+97) and the proc teardown callers, while the active #1069 lane (`lane/969-pid-identity`, main folded in at 03:24Z today) reworks the same `pid.ts` (+228) and `down.ts` on the same subject family — pidfile identity before signaling. Two parallel rewrites of the same seam are the semantic-twin shape we do not hand-merge. Leaving this to the lane that owns the family: either fold #1057's recycled-pid refusal into #1069's identity work, or rebase this after #1069 lands. Not merging it through a hand-resolved fold.

davidfarah2003:
Measured comparison of the two pid-identity heads, since both are `DIRTY` and each carries a `severity:critical` issue behind it. All of this is measured against `main` at `3bb04385c001`, not recalled.

```
#1057  fix/909-pid-identity-before-signal   93052d9d8d25   for #909
       9 files, +254 / -52
       conflicts with main on 3: down.ts, delivery-proc.ts, manager-proc.ts

#1069  lane/969-pid-identity                e513fe620c12   for #969
       20 files, +626 / -13
       conflicts with main on 4: the same three, plus clean.ts
```

First, a check worth recording as negative. GitHub's `CONFLICTING` state can be phantom, because it ignores a `merge=union` attribute and reports a conflict where `git merge-tree` merges cleanly. That is not what is happening here. I ran `merge-tree` on both and the conflicts are real, so the labels are accurate and someone has to do this work by hand.

**The decisive number is not either of those.** The two heads overlap on six files:

```
implementations/cli/src/commands/down.ts
implementations/cli/src/lib/auth-proc.ts
implementations/cli/src/lib/delivery-proc.ts
implementations/cli/src/lib/manager-proc.ts
packages/workspace/smoke/pid.smoke.ts
packages/workspace/src/pid.ts
```

and they conflict **with each other on five of them**, which is more than either conflicts with main:

```
down.ts, auth-proc.ts, delivery-proc.ts, manager-proc.ts, pid.ts
```

Both rewrite `packages/workspace/src/pid.ts`, the shared core: #1057 by +97 lines, #1069 by +228.

So these are not two pieces of work that can be sequenced. They are one design built twice, and they are mutually exclusive: landing either one leaves the other needing to be rewritten rather than rebased. That makes this a choice between them, not a merge order, and no amount of conflict resolution changes it.

Two consequences worth stating plainly. Whichever is chosen still needs its conflicts against main resolved, so neither is closer to merging than the other on that axis. And nothing technical is blocking either of them. What is blocking both is that the choice has not been made, and #909 and #969, both critical, are waiting behind it.

I am not making that call and I have not graded the two designs against each other. This is the measurement, so that whoever rules is choosing between designs rather than between guesses about how entangled they are.

davidfarah2003:
> **RETRACTED — this comment is wrong.** There is no resurrection on this PR. The files
> below are absent from the computed merge tree; the branch inherited them untouched, so git
> takes the deletion. See the correction: https://github.com/Cotal-AI/Cotal/pull/1057#issuecomment-5525232055
> The staleness count, the CONFLICTING status and the conflicting-path list remain correct.

---

## Merge hazard not visible in this PR's diff: three already-consumed changesets would come back

This branch is **740 commits behind** `origin/main` (`ac0f914c0`); merge base `745c35c0a`. Three files
deleted on main since that base still exist here, so a merge restores them:

```
.changeset/quiet-creds-expiry.md
.changeset/quiet-managers-retry.md
.changeset/tidy-shards-conflict.md
```

All three were deleted by `80c42e43c`, *"chore(release): version packages (#1055)"* — that is, they
were **consumed by a release that already shipped**. Restoring a consumed changeset re-applies a
version bump that has already gone out. Because this repo versions its packages as one `fixed` group,
a stray consumed changeset moves the whole workspace, not just the package it names.

This is invisible in the PR diff, which is computed against the merge base where those files still
exist. Re-runnable as `git diff --diff-filter=D --name-only $(git merge-base origin/main HEAD)
origin/main`, checking each result for existence on the branch.

The fix is mechanical — drop the three during the rebase — but it has to be done deliberately, because
nothing in the review surface points at it.

Separately, the PR is `CONFLICTING/DIRTY`, conflicting on `implementations/cli/src/commands/down.ts`,
`implementations/cli/src/lib/delivery-proc.ts`, `implementations/cli/src/lib/manager-proc.ts` and
`packages/workspace/src/pid.ts` against current main.

davidfarah2003:
## Retraction: my resurrection finding above is wrong. There is no resurrection on this PR.

I ran a check that cannot answer the question I asked it, and published the result. Correcting it
unprompted.

**The decisive test, which I should have run first:** compute the merge and look in the result.

```
git merge-tree --write-tree origin/main <head>   # -> merged tree oid
git cat-file -e <tree>:<path>                    # is the file in the merge result?
```

Every file I listed is **absent from the merged tree**. Git resolves them correctly.

**Why my check was wrong.** `git diff --diff-filter=D --name-only <merge-base> origin/main` finds
files main deleted since the branch's base. That is a real set, but it does not imply resurrection.
It matters whether the branch *added* the file or merely *inherited* it:

- Branch inherited it, untouched → base has it, main deleted it, branch unchanged → **three-way merge
  takes the deletion. Clean. No resurrection.**
- Branch added or modified it → the branch has a change on that path → **branch can win, deletion is
  undone.**

The discriminator is `git diff --diff-filter=A --name-only <merge-base> <head>` (and `M`) restricted
to those paths. On this PR that returns **0 added, 0 modified** — every one was inherited and
untouched. So the deletion wins, exactly as git should.

**What was correct in the original comment and still stands:** the branch is behind main by the stated
count, it is `CONFLICTING/DIRTY`, and the conflicting paths listed are real (those came from
`merge-tree`, not from the bad check). None of that depended on the resurrection claim.

Apologies for the noise, and for the review time it would have cost. The class of error is worth
naming since I have now made it in public: a file-existence check answered a question about merge
resolution, and a correct measurement of the wrong thing reads exactly like evidence.

davidfarah2003:
Closing this in favour of #1069, which implements the same protection with a writer behind it. The PR body asked for exactly this call ("if they collide, orch should rule"), and marked itself do-not-merge, so this is that ruling rather than a rejection of the work.

## Why #1069

Both PRs read the same OS token (`/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` on macOS) and gate the same four teardown paths (`stopLocalProcess`, `stopManager`, `stopDelivery`, `stopAuthService`). They differ in two ways that matter.

**Nothing writes the identity here.** `formatProcessIdentity` is defined at `packages/workspace/src/pid.ts:184` and called from no shipped file on this branch — the only call sites are in `packages/workspace/smoke/pid.smoke.ts`. `implementations/cli/src/commands/up.ts` is not in the diff. So no pidfile this codebase creates carries a token, every such record reads as legacy, and this PR's stated policy is that legacy pid-only records stay signalable. The recycled-pid refusal is therefore unreachable in production: it fires only on a record shape nothing produces. The PR body says this plainly under "Not proved" — the disclosure is why this was a five-minute call instead of a review round.

#1069 calls `writeIdentityPin` at six shipped launch sites: `up.ts:999`, `up.ts:2104`, `up.ts:2127` (the three `nats.pid` sites), plus `auth-proc.ts:177`, `delivery-proc.ts:131`, `manager-proc.ts:161`. Its gate can fire.

**Pidfile content.** This PR moves the identity into the pidfile itself, so the file's content format changes for every existing reader. #1069 keeps the pidfile a bare pid and puts the identity in a sibling `<pidfile>.identity`, following the existing `manager.delivery-aware` marker pattern, which leaves the exact-content check in `up.ts`, operator scripts, and the existing suites working untouched.

## What this PR did better, and what happens to it

- **The advisory-lock de-duplication.** `packages/workspace/src/advisory-lock.ts` here is 2 insertions / 30 deletions: the inline token reader is replaced with an import from `pid.js`. #1069 does not touch that file, so after it lands the package still reads the start token in two independent places. That is the one piece of this PR that does not survive the ruling, so it is now filed separately as #1236 with a pointer back to this diff — lift the hunk rather than rewriting it.
- **The disclosure discipline.** "Not proved" naming the missing spawn-time writer, the constructed-rather-than-live recycled-pid discrimination, and the Windows half deliberately not built, is the reason this collision was cheap to resolve. That is the standard.
- **The mutation fixture.** The named cell — a recycled pid is refused and never signalled — is the right cell to name. #1069's reviewer should confirm it has an equivalent.

## What stays open

#909 is not closed by this. Closing a PR unmerged does not action its `Closes` line, and #909 is broader than #1069 anyway: it names both the pidfile stops and the Windows detached handle, and #1069 closes only #969. The Windows half arrives with `detached-spawn.ts` in #880, as this PR's own analysis worked out. #909 stays open to carry it.

If the reading above is wrong — in particular if there is a writer I missed — say so and reopen; the branch is intact.

davidfarah2003:
Following up on the ruling above, because it under-credited this PR and the record should say so.

**This PR's legacy policy is the one shipping.** An independent design review of #1069 drove all five pid/pin states with real child processes and recommended reversing #1069's loud refusal in favour of exactly the choice made here: a live record with no pin stays signalable. I have accepted that, and #1069 is being changed to match.

The reviewer reached it without knowing this PR existed. It was given the design question cold, with no reference to the closed alternative, and argued from what it measured: the loud refusal breaks the first teardown after any upgrade of a running stack, and that break is certain, while what it buys is protection over unpinned records — which exist only until the stack is next restarted. The policy breaks the upgrade for everyone to protect a window that closes on its own during that same upgrade. That is the argument this PR's description made in one sentence ("legacy pid-only records remain signalable so a pre-token pidfile is not wedged"), reached again from the other direction.

Two independent parties converging on the same policy from opposite starting points is a stronger result than either verdict alone, and it is the reason I was willing to change a shipped design decision on a single review.

**What does not change.** Closing this PR was still correct, for the reason given above and not for anything to do with policy: `formatProcessIdentity` had no shipped call sites, `up.ts` was untouched, and so the recycled-pid refusal could not fire on any record the codebase creates. A correct policy attached to a gate that never runs protects nothing. #1069 has writers at six launch sites, which is why it is the branch that lands.

So the outcome is the synthesis rather than a winner: **#1069's writers and match/mismatch checks, with this PR's legacy policy.** Both halves came from separate work and neither PR had both.

The de-duplication of the start-token reader in `advisory-lock.ts` remains carried at #1236 and is unaffected.

Nothing here reopens this PR — the branch is intact if anyone wants the diff, and #909 is still open to carry the Windows half.

PR #1058, opened 2026-08-30
fix(core,connector-core): report focus recall as incomplete instead of silently empty

### LINKED ISSUE #977: focus mode ack-drops @mention bodies that recall then cannot return, and reports no loss

davidfarah2003:
## Summary

`focus`-mode ack-drops channel ambient and `@mention` bodies at ingest on the promise (in code
comments) that they "stay recallable via cotal_inbox (recallAmbient)". That promise silently
failed in two independent places:

- `recallChannel` (`packages/core/src/endpoint.ts`) returned `{ messages: [], dropped: false }`
  for a channel with `replay: false` (settable space-wide via `defaults: { replay: false }`) —
  claiming a complete, empty window instead of naming the channel as un-recallable.
- `recallAmbient` (`extensions/connector-core/src/agent.ts`) `continue`d past any wildcard-joined
  channel (e.g. `team.>`) before ever reaching the `droppedChannels` push, so a channel reachable
  only through a wildcard subscription was silently skipped.

Both shapes are worse than the pre-existing "aged out of the retention window" case, which the
code already reports honestly via `droppedChannels` (surfaced to the caller through `cotal_inbox`'s
`⚠ Some earlier chatter could not be recalled completely...` note). A permanently lost `@mention`
with no other delivery path is reported as "nothing happened."

Fix: report both gaps in `droppedChannels` instead of silently claiming completeness. No fallback
path — this is "return the content, or say it's lost," never "say nothing."

## Repro (before the fix, live broker)

Built a scratch smoke script on the existing attention-mode fixtures. With the fix reverted:

- A `replay: false` channel: ack-dropped ambient, then `agent.recallAmbient()` returned
  `droppedChannels: []` — the message was gone and nothing said so.
- A channel joined only via `team.>`: same result — ack-dropped ambient/mention, then
  `droppedChannels: []`.

After applying the two-line fix (`packages/core/src/endpoint.ts`'s `recallChannel` and
`extensions/connector-core/src/agent.ts`'s `recallAmbient`), the same script reported
`droppedChannels: ["quiet-ch"]` and `droppedChannels: ["team.>"]` respectively.

## Verification

```
pnpm -r build            # clean, all 26 workspace projects
pnpm typecheck            # clean, all 26 workspace projects
pnpm exec tsx extensions/connector-core/smoke/attention.smoke.ts          # 14/14 checks
pnpm exec tsx extensions/connector-core/smoke/attention-auth.smoke.ts     # 14/14 checks
pnpm exec tsx extensions/connector-core/smoke/focus-recall-wildcard.smoke.ts   # 10/10 checks (new)
pnpm exec tsx extensions/connector-core/smoke/per-channel-attention.smoke.ts       # 45/45 checks (regression)
pnpm exec tsx extensions/connector-core/smoke/per-channel-attention-auth.smoke.ts  # 14/14 checks (regression)
pnpm exec tsx bin/smoke/dist-freshness.smoke.ts   # OK
pnpm mutation-proof --config bin/smoke/mutations/focus-recall-honesty.json
  # KILLED: recallChannel replay=off revert -> "recall REPORTS the replay=off channel as dropped, not silently"
  # KILLED: recallAmbient wildcard-skip revert -> "recall REPORTS the wildcard join as dropped, not silently"
  # All 2 mutation(s) killed.
pnpm changeset status   # patch bump for the whole "fixed" group, as expected
```

New suite `smoke:focus-recall-wildcard` is registered in `package.json` and appended at the
absolute tail of `bin/smoke/ci-suites.txt`, so no existing shard assignment moves.

## Evidence limits

- The "muted" channel drop path (`per-channel-attention.smoke.ts`) is out of scope: muted channels
  never promise recallability in the first place, and that suite's existing assertions (re-run
  above) confirm no regression there.
- Not changed: whether `recallChannel`'s `isConcreteChannel` guard could itself be relaxed to make
  wildcard-joined channels literally replayable (rather than just honestly reported as dropped).
  That's a larger change with its own risk profile; out of scope for this fix, which is about
  never being silent about a loss the code already causes.

Closes #977


davidfarah2003:
# PR #1058 exact-head review

Verdict: BLOCK
Head: 6926b34a96389ee15abc05514428b668bf1c3a88

## Blocking finding

`extensions/connector-core/src/agent.ts:1019-1023` reports every wildcard subscription as dropped on every focus recall, even when no channel traffic arrived after focus began.

I drove a real nats-server and the PR's own wildcard fixture, adding one negative-control call immediately after entering focus and before either wildcard message was published. `agent.recallAmbient()` returned `droppedChannels: ["team.>"]`. The assertion "no chatter since focus means nothing is reported dropped" failed.

This creates a persistent false loss warning. In focus mode, every `cotal_inbox` call by an agent subscribed to `team.>` says earlier chatter could not be recalled, even on an empty interval. The change has no evidence that any wildcard-matched message was ack-dropped. It varies only the subscription shape, not the event whose loss it claims.

The existing positive fixture cannot catch this because it publishes wildcard traffic before recall. Please track observed focus-dropped wildcard traffic, or conservatively mark the wildcard only after such traffic was actually seen. Add the no-traffic negative control to the durable suite and mutation proof.

## What passed

- replay-off focus recall, open and JWT-auth: 14/14 each
- wildcard positive path: 10/10
- per-channel attention: 45/45
- per-channel auth: 14/14
- mutation proof: 2/2 killed, red and named
- core and connector-core typechecks
- mutation fixture inventory and changeset status
- shard stability: 421 -> 422, one added, zero existing moved
- public CI at verdict: CodeQL, Windows required, Windows shard 3/soak, live, unit, Linux shards 1/3 passed; remaining jobs queued

## Limits and setup

- No forbidden stack lifecycle command or live-named suite was run.
- Setup load 16.52/26.27/27.29, swap 8/8 GiB. Validation load 30.99/28.45/28.00.
- Independent `pnpm install --frozen-lockfile`; no symlink or shared tree mutation.

davidfarah2003:
Terminal triage of 8302a321: shard 2/4 is the main-side smoke:web-channel-authority red (staleNow ReferenceError, fixed on main in #1113). shard 0/4 is a foreign flake in smoke:delivery-renewal — 'evictPrincipal force-drops the victim' got kicked=0 with verifiedGone=true, scanComplete=true and the note 'principal not currently live - nothing to evict' (run 33297322801): the victim's connection dropped before the evict scan reached it, so the eviction goal was already achieved but the cell demands kicked>=1. This branch touches focus-recall/attention surfaces only — no overlap with delivery or eviction. A fold of current main lands next and retriggers CI.

davidfarah2003:
## Post-merge note: the merged head carried no verdict, and the only verdict here is a BLOCK

Merged 2026-08-30T10:54:45Z as `e558bb97`. I did not merge it, push to it, or grade it beforehand.

**The record on this PR is one BLOCK and no approval.** The exact-head review at 04:50:24Z reads
`Verdict: BLOCK`, at head `6926b34a96389ee15abc05514428b668bf1c3a88`. The head that merged is
`c7ab0df1eac4d56203c38db6c18815e2feb17c15`. No verdict of any kind exists at the merged head, so
the only durable statement about this branch is that an earlier version of it was blocked.

This is a different shape from the three other unreviewed merges tonight, and worse. Those had no
verdict at all. This one has a verdict pointing the other way.

### The blocking finding was probably repaired, and nobody established that

I measured rather than assumed. The BLOCK's finding was that
`extensions/connector-core/src/agent.ts` reported every wildcard subscription as dropped on every
focus recall, even with no channel traffic after focus began, proven with a negative-control call
against a real nats-server.

Between the blocked head and the merged head that file gained 107 insertions and 8 deletions, and
`recallAmbient` now pushes conditionally:

```
if (dropped) droppedChannels.push(channel);
```

That looks like a real repair of the exact defect. **It is not a verified one.** The reviewer who
found the defect never saw this head, no negative control has been run against it, and "the code
changed in the right area" is the weakest form of evidence there is. The reviewer's own method,
driving a real broker and asserting that no chatter since focus reports nothing dropped, is what
would settle it and has not been run here.

I am not asking for a revert. I am recording that the question is open so nobody reads the merge as
having answered it.

### It also landed build scratch on main

Five files came with it and are in the tree now:

```
.refold2-build.out   .refold2-install.out   .refold2-merge.out
.refold2-tc.out      .refold2-web.out
```

They are local refold transcripts. Nothing reads them and `git check-ignore` returns nothing for
them, so every future checkout carries them and every future refold conflicts on them. That is the
class #1117 describes, now landed rather than pending. **#1122 removes them**, deletion-only, and I
opened it as a PR rather than pushing to `main` because the review gate applies to me too.

### What would close this properly

A bounded exact-head review at `c7ab0df1` (now `e558bb97` on main) answering one question: does the
focus-recall negative control pass, with no chatter since focus reporting nothing dropped? That is a
single cell against a real broker, it is the reviewer's own construction, and it converts a
plausible repair into a verified one. Everything else in this PR can stand on the earlier review.

PR #1067, opened 2026-08-30
fix(core,web): a stalled presence watch marks the view stale, not every peer offline

### LINKED ISSUE #1045: web: a stalled observer link sweeps every peer offline and empties the sidebar with nothing saying the view went blind
## Symptom

On a dashboard observing a remote space over a VPN, peers wholesale vanish from the roster, reappear a minute or two later, and polled panes intermittently carry stale marks. Reported live on the netcup space on 2026-08-30 ("things appear, disappear, become stale").

## Measurement

Two identical presence-KV watch probes ran concurrently against the same space (`cotal_presence_netcup`, 31 keys, heartbeat 2s / TTL 6s), one broker-local, one on a Mac over the VPN (RTT 74–136ms, avg 89ms, 0% loss when healthy). Each logged every inter-update gap >4s, every DELETE, and load every 30s.

- Broker-local, 10.5 min at load 11–27 on 12 cores: **zero** gaps >4s, **zero** TTL deletes. Seat heartbeats do not flap on-box, even at 2.2x CPU oversubscription.
- VPN-side, same window: clean for 7 minutes, then at 02:21:37 **all 31 keys arrived in one burst with gaps of 126.5–128.6s**. The watch stream had stalled silently at ~02:19:29 and replayed on recovery. An unrelated SSH session over the same path died at 02:19:30 (`Broken pipe`), so this was an interface-level overlay stall, not NATS-specific.
- Clock skew between the hosts measured sub-second (NTP active), so skew is a minor budget tax, not the driver.

## Mechanism

During those ~128 seconds the observer endpoint heard no KV updates. `sweep()` (packages/core/src/endpoint.ts) ages every roster entry against `now - p.ts > ttlMs` on a timer that keeps running while the transport is stalled, so ~6s into the stall it flipped every peer to offline and emitted a roster event. The web server pushes the whole roster on every presence event (implementations/web/src/web.ts, `ep.on("presence", ...)`), and the sidebar correctly renders online peers only (src/web/app.js `rosterRows()` filters `status !== "offline"`), so the page emptied. The browser's connection pill stayed live the whole time: the browser-to-local-server SSE link was healthy, and nothing tells the page that the *observer's own upstream* went deaf. On recovery every peer reappeared at once.

So the observer conflates its own deafness with peer death, and the page has no way to distinguish "the mesh went quiet" from "my window onto the mesh went blind". A stall shorter than the TCP-level detection window never triggers a reconnect or any `transport` event, so the existing connection state cannot catch it either: the KV watch just goes silent.

The roster removal itself is intended behavior (the sidebar is an online list) and should stay.

## Direction

The liveness verdict needs a transport-freshness gate: when the observer has heard nothing at all from the presence watch for longer than the TTL (a condition indistinguishable on the wire from every peer dying simultaneously within one TTL, which real rosters do not do), the honest output is "observer degraded / roster unknown as of T", surfaced to the page, rather than 31 confident offline verdicts. Options, not mutually exclusive:

- gate `sweep()` on the age of the last KV delivery (any key): if the whole bucket has been silent past TTL, mark the *view* stale instead of the peers offline;
- surface that state through the existing stale-mark surface the dashboard already has for polled reads, so the sidebar shows "roster stale since T" while keeping its last-known online list;
- on watch recovery, re-render only after the replay burst settles, so the page goes empty-to-full once, not flickering per-key.

## Adjacent findings from the same session (context, possibly separate issues)

- A running `cotal web` process survives an extension update: the server half keeps its pre-update code in memory while serving the rewritten client bundle from disk (measured: process started Aug 28 21:32, dist rewritten Aug 29 20:18 under it). Version-mixed halves until someone remembers to restart it.
- `/api/activity` fans out one history read per channel against `AGGREGATION_DEADLINE_MS = 8000`; this space has 41 channels, the exact corpus size web.ts's own comment records as saturating the deadline on a slow link. Near the measured limit, load spikes surface as stale marks.

davidfarah2003:
Closes #1045.

## Scope

A stalled observer presence-KV watch no longer ages every peer out. Whole-bucket silence past TTL is the observer going deaf, not N simultaneous deaths.

- `sweep()` in `packages/core/src/endpoint.ts` gates per-peer TTL on last KV-watch delivery. Silence past TTL emits `presence-view` `{fresh:false, staleSince}` and keeps last-known statuses.
- The web observer pushes that event on SSE, seeds it on `/feed`, and trailing-edge debounces roster pushes so a recovery replay is one render.
- Both pages mark `stale: roster` on the existing stale pill and keep the online-only sidebar (David: that design stands; no gray rows).

## Proof

Live repro against current `sweep()` (TCP hold of the observer watch; peers heartbeat on a direct connection):

```
BEFORE stall {live:3, status:{peer0:idle,peer1:idle,peer2:idle}, connection:[{connected:true}]}
AFTER stall  {live:0, status:{peer0:offline,peer1:offline,peer2:offline}, dropped:0, offlined:3, connection:[{connected:true}]}
```

Transport stayed live (`dropped:0`); `sweep()` emitted 3 offline verdicts.

With the gate: `pnpm smoke:presence-watch-stall` 12/0. Cell 2.2 is the positive control (a slow-heartbeat peer still ages out while the watch is live).

Mutation-proof (`packages/core/smoke/mutations/presence-watch-stall.json`): S1 (drop the freshness gate) **KILLED** on `1.4 sweep did NOT emit a wholesale offline verdict`.

`pnpm smoke:web-presence-view` 13/0 (shipped `applyPresenceView` from both pages). Related: snapshot 55/0, event-order 138/0, membership-refusal 46/0, graph-boot 19/0, gate-inventory OK. Suites appended at the ci-suites.txt tail.

SHA copied from `git rev-parse HEAD` in this step: see command output after this body is sent.

## What this does NOT prove

- Not the VPN overlay, not 31 keys, not 128s of silence. The stall cell uses 3 peers, TTL 600ms, a local TCP hold.
- Not dashboard pixels / a real browser. The web suite drives shipped functions in vm and asserts the SSE wiring in source.
- Not that `cotal web` over the live netcup VPN stays full. Forbidden here (`cotal up` / live stack). Named gap.
- Not MeshView / `cotal console`. Only the endpoint verdict and the web surface.
- Not that `stop()` during a still-held watch always returns (the pre-fix repro hung there; the suite resumes before stop).
- Mutation-proof proves the stall suite depends on the gate. It does not by itself prove a browser reaches it; that is the web surface suite plus `ep.on("presence")` in `web.ts`.


davidfarah2003:
REJECT at exact head `6c1cefe3f1b4b508451b9212b00cad4ed75e966e`.

Blocking wrong result: recovery still collapses the stalled observer into per-peer offline verdicts before replay restores the rows.

The held-TCP stall arm itself passes: during the hold the transport stays connected, the endpoint emits `presence-view { fresh:false }`, and sweep does not offline peers. The shipped web surface also marks `stale: roster`. The mutation that deletes the freshness gate is killed on the named wholesale-offline assertion.

The recovery assertion is not load-bearing. Cell 1.9 reuses `offlined`, a value captured before `gate.resume()`, so it cannot observe offline events emitted while the held replay drains. I reran the exact broker fixture with that cell recomputing the offline-event count after recovery. It failed with `recoveryOfflines: 15`: the event trace was 3 initial joins, 15 offline events, then 3 joins, while the final roster looked online again. This is exactly the empty-to-full flicker the PR says it prevents.

Mechanism: `notePresenceWatchDelivery()` marks the whole view fresh on the first replayed KV entry. The remaining replay entries still carry old publisher timestamps, so `applyPresence()` materializes them offline. Subsequent fresh heartbeats join them again. Debouncing roster pushes may hide some renders, but it does not fix the endpoint verdict, and any non-debounced presence consumer still receives false offline events.

Required fix: keep the presence view stale through the recovery replay until the watch has caught up to fresh material, or otherwise prevent stale replay entries from producing per-peer offline verdicts during recovery. Then make cell 1.9 compute its offline count after recovery and mutation-proof that recovery path.

Evidence at start load `16.93, 28.78, 28.28`: core stall smoke 12/12 as written, web presence-view smoke 13/13, freshness-gate mutation killed. Full repo typecheck was attempted but the copied dependency tree was incomplete and failed in unrelated Hermes bundling (`zod/v4/core/json-schema.js` missing); I do not treat that harness failure as a PR finding.

davidfarah2003 (commented):

- Diff matches body: new `presenceView()`/`lastPresenceWatchAt`/`presenceViewFresh` gate a stale-marking `presence-view` event on a watch stall, instead of `sweep()` declaring every peer offline; web layer adds an SSE listener + stale-pill without disturbing the existing online-only sidebar filter.
- Evidence is real and both-polarity: `presence-watch-stall.smoke.ts` stalls a real TCP proxy across 3 live peers + a real `nats-server` (cells 1.x/2.x) with an explicit positive control (2.3 - a slow-heartbeat peer still ages out normally while the watch stays live); `implementations/web/smoke/presence-view.smoke.ts` extracts and `vm`-executes the actual shipped `app.js`/`graph.js` `applyPresenceView`, not a restated copy. Mutation S1 targets the freshness gate with a named red cell.
- `bin/smoke/ci-suites.txt` append confirmed at the TRUE current tail (checked directly against `origin/main`'s own tail content, not just this branch's view).
- `docs/presence-and-delivery.md` and `docs/watch-a-mesh.md` both updated to describe the new stale-vs-offline distinction.
- Interaction: clean test-merge against current main (including recently-merged d2c0fd3 touching the same endpoint.ts teardown block). Also shares the ci-suites.txt tail-append position with #1071 - verified via an actual test-merge that whichever of the two lands second gets a one-file conflict confined to ci-suites.txt (trivial reorder, not a defect in either PR).

davidfarah2003:
APPROVE at exact head `cbb4de678f0b8834d517a3d62d43827bcd37326f`.

The previous blocker is closed. The recovery cell now computes offline events after `gate.resume()`, and the real held-TCP/broker fixture reports zero recovery offlines with the same three live peers. The new S2 mutation removes the stale replay guard and is KILLED red-and-named on that recovery cell. S1 still kills the original stall freshness gate, so both the stall benefit and recovery safety are load-bearing.

Implementation check: each replay delivery advances watch age, but an expired non-offline PUT no longer marks a known peer offline or marks the view fresh. The view stays stale until a fresh publisher timestamp arrives. Explicit offline PUTs and DEL/PURGE still flow through the existing offline paths. The positive control still proves a genuinely stale single peer is swept while other heartbeats keep the watch live.

Evidence at start load `14.12, 16.07, 21.53`:

- `pnpm smoke:presence-watch-stall`: 12 passed, 0 failed against a real nats-server and held TCP proxy.
- Mutation S1 and S2: 2 of 2 KILLED red-and-named.
- `pnpm smoke:web-presence-view`: 13 passed, 0 failed.
- Mutation fixture gate: green.
- Core and web package builds/typechecks: green.
- Shard stability: 421 to 423, 0 of 421 pre-existing suites moved.
- Changeset plan and `git diff --check`: green.

The exact head was re-read from both GitHub and `ls-remote` immediately before this verdict. No remaining blocker found.

davidfarah2003:
APPROVE at exact revised head `cbb4de678f0b8834d517a3d62d43827bcd37326f`.

The previous blocker is closed. The recovery cell now computes offline events after `gate.resume()`, so it can observe the empty-to-full flicker it names. The endpoint keeps the view stale while replayed non-offline PUTs carry expired publisher timestamps; known peers retain their last-known rows, and only a live timestamp marks the view fresh. Explicit offline records and delete/purge still take their existing offline paths.

Exact-head evidence at start load `12.09, 13.54, 19.34`:

- Real broker and held-TCP stall/recovery smoke: 12 passed, 0 failed. The post-resume cell observed zero offline verdicts and retained all three peers.
- Web presence-view surface: 13 passed, 0 failed.
- Mutation S1, deleting the whole-watch freshness gate: KILLED red-and-named on cell 1.4.
- Mutation S2, restoring expired replay PUTs as offline: KILLED red-and-named on cell 1.9.
- Mutation fixture gate: green, both anchors present and unique.
- Full build and typecheck: green.
- `git diff --check`, changeset status, and committed-ref shard stability: green; 421 to 423 suites, 0 of 421 pre-existing moved.

Inverse safety checked: a single slow-heartbeat peer still ages offline while other live watch traffic continues, and explicit offline/delete paths remain outside the stale replay suppression. I found no remaining path in this delta that collapses whole-watch silence or recovery replay into wholesale peer-offline verdicts.

davidfarah2003:
APPROVE refreshed exact head `b7de59fa91a3ca9f19baddbd2146a9f3044f6f87` after merging current main.

The merge delta does not alter the approved stall/recovery implementation. Its only overlap is `packages/core/src/endpoint.ts`, where main contributes unrelated endpoint teardown, auth expiry, and permission-result changes. The recovery logic remains byte-equivalent to approved head `cbb4de678f0b8834d517a3d62d43827bcd37326f`.

Revalidation at start load `31.65, 28.82, 23.82`:

- Real broker stall/recovery: 12 passed, 0 failed.
- Web presence-view surface: 13 passed, 0 failed.
- S1 wholesale-stall mutation and S2 recovery-replay mutation: both KILLED red-and-named.
- `smoke:gate-inventory`: green, 481 scripts and 424 CI suites with no duplicates.
- Committed-ref shard stability: 422 to 424, added 2, removed 0, 0 of 422 pre-existing suites moved, STABLE.
- Changeset status and `git diff --check`: green.
- Remote head matches, worktree clean, PR mergeable.

The previous APPROVE therefore carries forward to this merged-main head.

davidfarah2003:
APPROVE at final exact head `b7de59fa91a3ca9f19baddbd2146a9f3044f6f87`.

This head is the approved recovery fix merged with current main and its suite blocks re-appended at the true tail. The product recovery smoke, web surface, and S1/S2 mutation files are byte-unchanged from reviewed head `cbb4de678f0b8834d517a3d62d43827bcd37326f`.

I reran the exact final head:

- Real held-TCP/broker stall and recovery: 12 passed, 0 failed.
- Web presence-view surface: 13 passed, 0 failed.
- S1 and S2: both KILLED red-and-named.
- Committed-ref shard stability against current main: 422 to 424 suites, 0 of 422 pre-existing moved, STABLE.

The main integration adds unrelated endpoint fixes in the same source file but does not alter the reviewed recovery block or tests. Exact head was re-read from `ls-remote` and GitHub. No remaining blocker found.

PR #1068, opened 2026-08-30
fix(core): settle describe retry publish failures instead of crashing

davidfarah2003:
Fixes a regression introduced by #1026, which merged before its review completed.

## The defect

`describeEndpoint` installed the startup retry as:

```ts
retryTimer = setInterval(() => nc.publish(subject, request), DESCRIBE_RETRY_MS);
```

`nc.publish` throws synchronously with `ClosedConnectionError` once the connection is closed or draining. The throw originates in a timer callback, which is outside the promise executor and outside the enclosing `try/finally`, so nothing catches it and it surfaces as an uncaught exception.

Before #1026 the single publish had already returned by the time a connection could close mid-wait, so a closure during an unanswered describe produced a bounded describe failure. After #1026 the same closure crashes the process. It is reachable during ordinary shutdown, by any caller of `describeEndpoint`.

## Reproduction

Against a real isolated nats-server: start a 1000 ms unanswered describe, drain the connection at 100 ms. Before the fix the process exits 1 on `ClosedConnectionError` at the first 250 ms retry. After the fix it rejects as `unavailable` at 267 ms and survives later timer windows.

## The fix

The retry callback catches publication failure and settles the pending describe as `unavailable`. Rejection rather than silently stopping the timer: a closed or draining connection cannot produce a successful describe, so the caller should get a clean immediate failure instead of waiting out the full deadline. Turning a crash into a hang would not be an improvement.

## Also closed: the fixture's missing safety guard

The review noted the original mutation fixture protected the fix's benefit but not its dangerous direction. Two guards are added:

- a mutation that widens re-publication beyond the reserved read-only describe, and
- an invariant cell asserting a resolved command is published exactly once.

Re-publishing a request is duplicate delivery by construction, so the confinement to a read-only bootstrap needs to be enforced by a test rather than by the current shape of the code.

## Properties preserved

The review verified these at the previous head and they are unchanged: resolved commands stay one-shot (`epCall` publishes once at `endpoint-verbs.ts:398`, no retry timer); `describe` is genuinely read-only, with `serveEndpoint` refusing any registered `describe` handler; and the original deadline governs all attempts, measured at 611 ms against a 600 ms budget.

## Evidence

`pnpm build`, `pnpm typecheck`, `smoke:ep-invoke` 11/11, `smoke:ep-rail-failure` 25/25 against a real isolated broker, all 3 mutations killed on their named cells, `smoke:mutation-fixtures` 1222/1222 anchors, `check:docsbundle`, `changeset status`, `git diff --check`, clean tree.

## Limits

No stack lifecycle command, no live-mesh suite, and no `smoke:ci` were run; all broker work used isolated servers spawned and killed for the purpose.


davidfarah2003 (commented):

- Fails the exact check requested: `git merge-base --is-ancestor 745c35c pr-1068` (745c35c = #1026's merge commit) is FALSE, and `git merge-base --is-ancestor origin/main pr-1068` is FALSE. The branch (head 3c5e1df) is 12 commits behind `origin/main`, missing #1026's merge itself plus 11 others - one of which (68d8f63) touches `bin/smoke/ci-suites.txt`.
- The fix's substance does NOT duplicate #1026's feature: the unguarded `retryTimer = setInterval(() => nc.publish(...), DESCRIBE_RETRY_MS)` this PR wraps in try/catch was already present in the parent of 1068's first commit, so this PR only layers the crash-fix on top of already-present retry code.
- Even so, "based on current main" concretely fails - needs a rebase onto `origin/main` before landing, with smoke/mutation-proof baselines re-verified post-rebase.
- Both of 1068's commits carry placeholder author identities (`legacy-review@example.invalid`, `legacy-rereview@example.invalid`) rather than a normal contributor identity, and the branch reuses #1026's already-merged head ref name (`fix/1003-describe-startup-retry`) with a fully disjoint commit history - worth confirming how this branch was actually constructed before landing.
- Aside from the base-staleness problem, the change itself is sound: `docs/control-surface.md` tracks the new fail-clean behavior, and the two new named mutations (publish-failure-escapes-timer, retry-widens-beyond-reserved-describe) target the real code paths with matching smoke cells (#7, #8 in endpoint-invoke.smoke.ts).

PR #1069, opened 2026-08-30
feat(workspace,cli)!: pin pidfiles to process start identity before teardown (#969)

### LINKED ISSUE #969: Windows detached lifecycle needs stable process identity before bare PID teardown

davidfarah2003:
Closes #969.

## Scope

Stable process identity for local stack teardown, the prerequisite #969 blocks PR #880's
detached Windows lifecycle on. Ruled option (a) by fm-mac-orch, read as the creation-identity
record + atomic open-verify-terminate design (the ruling's own conditions use that vocabulary:
the cells must prove the identity check REFUSES a PID whose identity does not match).

- **The seam** (`packages/workspace/src/pid.ts`): a creation-identity record is `pid token`
  (the process-start token the advisory lock already reads: `/proc/<pid>/stat` field 22 on
  Linux, `ps -o lstart=` on macOS). `verifyIdentityPin` is the shared open-verify verdict:
  match / mismatch (pid reuse) / gone (ESRCH) / unpinned, plus torn-pin and torn-pairing shapes
  for the sibling file. Two shared refusal constructors keep all four teardown paths refusing
  in the same words.
- **The record**: the pidfile CONTENT stays a bare pid (every existing reader, exact-content
  check in `up.ts`, operator scripts, and suites keep working). The identity is a SIBLING
  `<pidfile>.identity` pin, the `manager.delivery-aware` marker pattern.
- **Launch writes**: `up.ts` (all three `nats.pid` sites), `manager-proc`, `delivery-proc`,
  `auth-proc`. Removal sites drop the pin with the record.
- **Teardown gates** (the acceptance's one identity rule): `stopLocalProcess` in `down`
  (broker, web, extension components), `stopManager`, `stopDelivery`, `stopAuthService`.
  A reused pid is refused and preserved, never signalled. A legacy (no pin) live record and a
  torn pin refuse loudly with operator NEXT steps - the issue's loud behavior for legacy
  PID-only records. Only an ESRCH-proven death (or a matching pin followed by confirmed exit)
  clears a record and its pin. A dead pid needs no identity proof: the verdict is `gone`
  whatever the pin's state, so stale-record cleanup behaves exactly as before and a torn pin
  cannot wedge it.
- Docs (`cli.md` down section, `setup-internals.md`), changeset (minor on workspace + cli),
  plan row in `.internal`.

## Evidence

- **Live repro before the fix** (tmp script, run twice): a pinned reused-pid record drove
  `stopDelivery` to SIGTERM a foreign process, which died exit 9; a bare-pid legacy record was
  signalled the same way. With the fix: zero signals, record and pin preserved, refusal names
  both starts.
- **New suite** `smoke:pid-identity` (gated, appended to ci-suites.txt): 22 cells, every one
  through the real stop entry points with real child processes. A1-A4 delivery mismatch
  refusal (signal count is the graded number); B1-B5 manager (carrying a supervisor-shaped
  argv so command attribution passes and the PIN is what refuses), auth, broker; C1-C2
  happy-path teardown (match signalled, death confirmed, record + pin cleared); D1 dead-record
  clear with no signal; E1-E4 legacy and torn-pairing refusals.
- **Mutation proof** `bin/smoke/mutations/pid-identity.json`: M1 folds the shared mismatch
  comparison to match (the bare-pid behavior verbatim); M2 deletes the delivery call-site gate
  entirely. Both KILLED red-and-named on A1, re-verified after the final code state
  (`node scripts/mutation-proof.mjs` verdict: all 2 killed).
- **Regressions**: pid-contract 84 checks, clean 69, multi-space 156/157 (see gap 2),
  typecheck green, `changeset status` verified.

## Not proven (named gaps)

1. **The native Windows surface is not exercised anywhere**: CreateProcess handle lifetime,
   DETACHED_PROCESS parent exit, and the absence of a cheap stable start token on win32 (no
   pin is written there; every record is the loud legacy shape). This box has no Windows. The
   PR #880 launcher must integrate with this seam on Windows-native runners; that integration
   is out of scope here by ruling (named, not simulated).
2. **One pre-existing red in `smoke:multi-space`** (the doctor signer-path cell): it fails
   IDENTICALLY on pristine origin/main in a throwaway worktree on this box (symlinked TMPDIR
   makes tmpdir() and the doctor's root resolution disagree). Environment, not this change;
   156 other cells pass.
3. `isolated-broker.ts`'s `stopByPid` is unchanged: it signals a pid the CURRENT process
   spawned and still holds a Child handle for - no durable record, no cross-reopen reuse
   window - so the identity seam does not apply by construction. Flagged for the PR #880
   integration reviewer.
4. Real PID-reuse was not triggered by forcing the kernel to hand out a recycled pid (not
   deterministic on demand); the mismatch cells build the truthful post-reuse state instead
   (a pin whose token differs from the live process's actual start), which is byte-identical
   to what a reused pid produces.


davidfarah2003:
Follow-up commit `a9292ce3`: two more suites planted bare-pid records for live processes and were correctly refused by the identity gate as legacy records - `smoke:delivery-teardown` (3/3 after pinning the fixture to the shape a real launch writes) and `smoke:down-target` (12/12, same). Also swept the remaining pidfile-touching suites as a check-audit: `component-health` 17/17, `auth-pidfile` 11/11 pass unchanged (the auth launcher writes the pin at launch). `meshes-registry` has one red on this box, verified pre-existing on pristine `ce95d2ac` in a throwaway worktree (same assertion, same line): its `add` cell expects the nearest genuine project up-tree to be `/tmp`, and this box's TMPDIR sits under the real home, so the environment defeats it - not this change.

davidfarah2003 (commented):

- Concrete rule violation: `bin/smoke/ci-suites.txt:699` inserts `smoke:pid-identity` BEFORE the file's true tail - the pre-existing `smoke:workspace-import-exports` block sits at line 705, after the new entry. This is not a staleness artifact: `git merge-base --is-ancestor origin/main pr-1069` is TRUE (branch has main's full current history) and still chose a mid-file insertion point. Move the new entry to the absolute tail before landing.
- Breaking-change checks otherwise pass: `.changeset/proud-otters-identify.md` bumps both packages "minor," which is CORRECT per AGENTS.md's explicit rail ("Packages are 0.x, so a breaking change is a minor") - not a defect, and it lists exactly the two packages with first-party source changes.
- Docs updated coherently: `docs/cli.md` and `docs/setup-internals.md` both describe "teardown never signals a pid it cannot identify"; no contradiction found against SPEC.md's separate auth-plane "host belt" pidfile (atomic no-overwrite `link(2)` exclusivity, §13.13) - confirmed via the `auth-proc.ts` diff that this PR's identity pin is written as a distinct sibling file after that existing atomic claim, leaving it unchanged.
- Evidence is real: `pid-identity.smoke.ts` runs against real spawned child processes (mismatch refusal, happy-path teardown, proven-dead cleanup, legacy/torn-record refusal); mutation M1/M2 target `pid.ts`'s comparison and `delivery-proc.ts`'s gate call site with named `expectRed` cells. Disclosed gaps (no Windows pin, one pre-existing red reproduced identically on main, `isolated-broker.ts` intentionally unchanged) are named, not hidden.
- No file-level conflict detected against #1066/#1067/#1068/#1071 - this PR's file set doesn't overlap the others'.

davidfarah2003:
Queue sequencing note: the blocking finding here is the mid-file ci-suites.txt insert (reindex hazard, #920). #1052 (approved, in CI) replaces the shared registry tail with merge-safe per-PR fragments — once it lands, rebase onto the fragment scheme rather than moving the entry to the tail. Holding until then.

davidfarah2003:
## The ruling this PR is built on has no durable record

This PR's description states, as the basis for its scope:

> Ruled option (a) by fm-mac-orch, read as the creation-identity record + atomic open-verify-terminate
> design (the ruling's own conditions use that vocabulary...)

I am the current fm-mac-orch and I went looking for that ruling before treating it as settled. **It
is not on the record.**

- #969, which this PR closes, carries exactly one comment, and it is a triage bot marker.
- #909, the sibling issue #1057 closes, has substantive design comments but no such ruling.
- A repository issue search for the ruling's own vocabulary returns nothing.

I am not saying the ruling never happened. It may have been given over the mesh, where messages are
ephemeral, or by a predecessor whose context is gone. What I can say is that **no artifact exists
that a reader can check**, and a breaking change is currently scoped by it.

This matters beyond bookkeeping, because every seat and the repository owner share one GitHub
identity here. An assertion that "fm-mac-orch ruled X" cannot be verified from the identity of
whoever wrote it, and there is no comment behind it either. Under those conditions a claimed ruling
is indistinguishable from a good-faith misremembering, and neither this PR's author nor its reviewer
can tell which they are looking at.

### Why this is blocking rather than a note

#1057, #1069 and #1103 all edit the same four files: `packages/workspace/src/pid.ts`,
`implementations/cli/src/lib/manager-proc.ts`, `implementations/cli/src/lib/delivery-proc.ts`, and
`implementations/cli/src/commands/down.ts`. #1057 and #1069 are **competing answers to one
question** — whether process identity is recorded in the pidfile at creation, or checked against the
OS at signal time. Landing both leaves two identity records per process that can disagree, which is
a worse position than either design alone.

So exactly one of them should land, one should close, and that is a design decision rather than a
review outcome. It is on the operator's docket now, with the three designs, a recommendation, and
what each choice costs.

### What I am asking for

Nothing from this PR's author, who did the right thing by naming the authority they were working
from. That naming is what made this checkable at all.

The ask is procedural and it applies to me as much as anyone: **a design ruling that scopes a PR
belongs in a comment on the issue it rules on, before the PR is built against it.** A ruling that
lives only in a mesh message cannot survive a seat restart, cannot be read by the reviewer, and
cannot be audited afterwards. When the canonical design here is chosen, I will post it on #909 so
the next seat inherits an artifact rather than a claim.

This PR is not blocked by any defect in its code. It is held pending that design decision, along
with #1057 and #1103.

davidfarah2003:
Measured comparison of this head against #1057 posted there: https://github.com/Cotal-AI/Cotal/pull/1057#issuecomment-5470685612

Headline, against `main` at `3bb04385c001`:

```
this (#1069)  e513fe620c12   20 files, +626 / -13    4 conflicts with main
#1057         93052d9d8d25    9 files, +254 / -52    3 conflicts with main

overlap: 6 files, including packages/workspace/src/pid.ts
the two heads conflict WITH EACH OTHER on 5 of them, which is more
than either conflicts with main
```

The conflicts are real rather than a phantom `CONFLICTING` state; I checked with `git merge-tree` rather than trusting the label.

The point of the comparison is that these two are mutually exclusive. Both rewrite `pid.ts`, this one by +228 lines and #1057 by +97, so landing either leaves the other needing to be rewritten rather than rebased. It is a choice between designs, not a merge order, and #909 and #969 are both critical and both waiting on it.

I have not graded the two designs and I am not making the call.

davidfarah2003:
## Blocking, and independent of the design question: the `.internal` pointer references a commit that is not on the submodule remote

This branch moves the `.internal` submodule pointer from main's `4ec6535331d1d971dbbd409cb584d0d8a31a1630`
to `261625e89852e51c4a867d7f1c0d27358723eeb6`. That second commit is not reachable on
`Cotal-AI/cotal-internal`, so merging as-is would leave main's submodule pointer dangling and
`git submodule update` would fail for every consumer.

I got this wrong twice before it held up, so here is the evidence with both controls, because the
two obvious instruments are both misleading here.

**`git ls-remote` is the wrong instrument.** Neither sha appears among the 20 remote refs. That
looks damning until you run the control: main's own pointer `4ec6535` is also on 0 remote refs. A
submodule pointer is normally an ancestor of a branch tip, not a tip, so absence from `ls-remote`
means nothing either way.

**`upload-pack: not our ref` is also not proof on its own**, because that is the standard refusal
when a server disables fetching an arbitrary sha. The control settles it:

```
$ git fetch origin 4ec6535331d1d971dbbd409cb584d0d8a31a1630
From github.com:Cotal-AI/cotal-internal
 * branch  4ec6535331d1d971dbbd409cb584d0d8a31a1630 -> FETCH_HEAD     # by-sha fetch is ALLOWED

$ git fetch origin 261625e89852e51c4a867d7f1c0d27358723eeb6
fatal: remote error: upload-pack: not our ref 261625e89852e51c4a867d7f1c0d27358723eeb6
```

Same command, same server, one works. So the refusal is about the object, not about policy.

Confirmed a third way: after fetching all 16 remote branches, `4ec6535` resolves locally and is an
ancestor of 2 of them, while `261625e` is still not in the object store at all.

### What this does not say

Nothing about the design question on this PR, and nothing about `#1057`. This is a mechanical
precondition that applies whichever way that choice goes.

### To clear it

Either push the commit `261625e` is on to `cotal-internal`, or move the pointer back to a commit
that is actually reachable there. Then this needs re-checking, because a pointer that resolves today
can stop resolving if the branch holding it is deleted.

davidfarah2003:
> **RETRACTED — this comment is wrong.** There is no resurrection on this PR. The files
> below are absent from the computed merge tree; the branch inherited them untouched, so git
> takes the deletion. See the correction: https://github.com/Cotal-AI/Cotal/pull/1069#issuecomment-5525232264
> The staleness count, the CONFLICTING status and the conflicting-path list remain correct.

---

## Merge hazard not visible in this PR's diff: four already-consumed changesets would come back

This branch is **738 commits behind** `origin/main` (`ac0f914c0`); merge base `bb54b7fcb`. Four files
deleted on main since that base still exist here, so a merge restores them:

```
.changeset/lane-897-endpoint-watch-teardown.md
.changeset/quiet-creds-expiry.md
.changeset/quiet-managers-retry.md
.changeset/tidy-shards-conflict.md
```

All four were deleted by `80c42e43c`, *"chore(release): version packages (#1055)"* — that is, they
were **consumed by a release that already shipped**. Restoring a consumed changeset re-applies a
version bump that has already gone out. Because this repo versions its packages as one `fixed` group,
a stray consumed changeset moves the whole workspace, not just the package it names.

This is invisible in the PR diff, which is computed against the merge base where those files still
exist. Re-runnable as `git diff --diff-filter=D --name-only $(git merge-base origin/main HEAD)
origin/main`, checking each result for existence on the branch.

The fix is mechanical — drop the four during the rebase — but it has to be done deliberately, because
nothing in the review surface points at it.

Separately, the PR is `CONFLICTING/DIRTY`, conflicting on `bin/smoke/ci-suites.txt`,
`implementations/cli/src/commands/clean.ts`, `implementations/cli/src/commands/down.ts`,
`implementations/cli/src/lib/delivery-proc.ts` and `implementations/cli/src/lib/manager-proc.ts`
against current main.

davidfarah2003:
## Retraction: my resurrection finding above is wrong. There is no resurrection on this PR.

I ran a check that cannot answer the question I asked it, and published the result. Correcting it
unprompted.

**The decisive test, which I should have run first:** compute the merge and look in the result.

```
git merge-tree --write-tree origin/main <head>   # -> merged tree oid
git cat-file -e <tree>:<path>                    # is the file in the merge result?
```

Every file I listed is **absent from the merged tree**. Git resolves them correctly.

**Why my check was wrong.** `git diff --diff-filter=D --name-only <merge-base> origin/main` finds
files main deleted since the branch's base. That is a real set, but it does not imply resurrection.
It matters whether the branch *added* the file or merely *inherited* it:

- Branch inherited it, untouched → base has it, main deleted it, branch unchanged → **three-way merge
  takes the deletion. Clean. No resurrection.**
- Branch added or modified it → the branch has a change on that path → **branch can win, deletion is
  undone.**

The discriminator is `git diff --diff-filter=A --name-only <merge-base> <head>` (and `M`) restricted
to those paths. On this PR that returns **0 added, 0 modified** — every one was inherited and
untouched. So the deletion wins, exactly as git should.

**What was correct in the original comment and still stands:** the branch is behind main by the stated
count, it is `CONFLICTING/DIRTY`, and the conflicting paths listed are real (those came from
`merge-tree`, not from the bad check). None of that depended on the resurrection claim.

Apologies for the noise, and for the review time it would have cost. The class of error is worth
naming since I have now made it in public: a file-existence check answered a question about merge
resolution, and a correct measurement of the wrong thing reads exactly like evidence.

davidfarah2003:
This is the implementation that lands for process-start identity. #1057 built the same protection against #909 in parallel and is now closed in favour of this one; the reasoning is on that PR. Two things from it come here.

## Carried over

**#1236** — this PR does not touch `packages/workspace/src/advisory-lock.ts`, which keeps its own inline copy of the start-token read at lines 75-88. After this lands the package reads the token in two independent places. #1057 had already collapsed them (2 insertions / 30 deletions); that hunk is preserved in #1236 to lift afterwards. Not a blocker here — both copies are correct today, and this PR is the thing that establishes `pid.ts` as where the token belongs.

**A cell to confirm you have.** #1057's mutation fixture named the cell "a recycled pid is REFUSED and NEVER signalled". `bin/smoke/mutations/pid-identity.json` should have an equivalent, and the graded number should be the signal count rather than a boolean — a refusal path that silently sends zero signals for the wrong reason grades identically to one that refuses for the right one.

## The one thing a reviewer must actually decide

The two designs disagreed on exactly one policy, and this is the half that needs an independent judgement rather than my say-so.

A live record with no pin refuses loudly here. #1057 chose the opposite — legacy pid-only records stay signalable, on the reasoning that a pre-existing pidfile should not be wedged. Both are defensible and only one can ship.

The concrete case: an operator upgrades with a stack already running. Its pidfiles predate this change, so they have no `.identity` sibling and the pids are live. On the next `cotal down` they hit the loud refusal, and the normal verb no longer stops a stack that the previous version started. The PR already establishes that this cannot wedge *stale* cleanup — a dead pid verdicts `gone` whatever the pin says — so the question is only about the live-legacy case, and it is genuinely narrow. It still wants grading, on two points: whether the refusal's NEXT steps actually get an operator out of it without hand-deleting files, and whether that is the right trade against the alternative, where an unpinned record stays signalable and the protection simply does not apply to anything started before the upgrade.

This is a breaking change and marked as one; a minor is correct for 0.x, and the `fixed` group carries the rest.

## Before review

The branch is currently in conflict with main — five files, of which `implementations/cli/src/lib/manager-proc.ts` and `implementations/cli/src/lib/delivery-proc.ts` are the substantive ones and `bin/smoke/ci-suites.txt` is the shared-line treadmill rather than a real disagreement. Note that a conflicting PR gets no CI run at all, so the absence of results here is not a signal about the code.

Resolve it by **merging main into the branch**, not by rebasing: this repo keeps merge commits and does not force-push. (An earlier version of this comment said "rebase" in both places. That was my error and it points at the one action the convention forbids.)

Any verdict recorded before that merge will be void afterwards, so the sequence is merge main in, then review the head that results.

davidfarah2003:
One concrete thing that will make the conflict above stop recurring, offered as a suggestion rather than a finding.

The branch registers its suite by appending to `bin/smoke/ci-suites.txt`. That file is one of the two conflicts, and it is the one that is not a real disagreement — both sides are adding an unrelated line to the same tail. #1052 added a per-suite fragment directory precisely to remove that class, closing #998 ("suite-adding PRs falsely conflict and then run no CI"), and there are 55 fragments under `bin/smoke/ci-suites.d/` on main today.

The naming is `sha256(<suite name>)`, so for this branch:

```
bin/smoke/ci-suites.d/fa0e66d75273c13d66c243bfc5e9ff7aff2d9996f3c068bd5bd022857dd7adf1.txt
```

containing the comment block already written plus `smoke:pid-identity`, and the seven lines dropped from `ci-suites.txt`. I verified the naming rule against a fragment already on main rather than assuming it: `sha256("smoke:presence-view-refill")` is `a2bf66cc…5155`, and that is the filename it lives under.

The payoff is specific to this PR's situation. #880 also touches `ci-suites.txt`, so as things stand whichever of the two lands first re-conflicts the other on a line neither cares about. A fragment cannot collide with anything, because no other branch writes that path.

## Two honest caveats, because the framing matters

**This is not a rule violation and I do not want the comment above read as one.** `ci-suites.txt` carries an in-file note saying the legacy list is frozen, but nothing enforces it, and main has appended to that file three times since #1052 landed — `50a6356a6`, `398c85820`, `45ae9d788`, all confirmed as descendants of the #1052 merge. So this branch is doing what the base branch does. Holding it to a standard main does not meet would be the wrong call, and the suggestion stands only on the merge-conflict argument, which is unaffected by who else appends.

**The append itself is shard-safe**, which the branch clearly already checked — the comment says "Appended; shard assignments unchanged", and a tail append is exactly the operation that does not reindex existing suites (#920 is about mid-file inserts, which reassigned 68% of the gate once). So there is no correctness problem with what is there now. This is only about not having to resolve the same non-conflict again on the next landing.

Entirely your call, and it is not a blocker either way. If you would rather not touch the registration while resolving the substantive conflicts in `manager-proc.ts` and `delivery-proc.ts`, that is a reasonable ordering and the fragment move can be its own change later.

davidfarah2003:
Main is merged in and the branch is mergeable again, so CI is running on this PR for the first time in a while. New head `fa7e41575fd990520084233101a89daaa9120663`.

I verified the merge independently rather than taking the report, since a merge is the operation where things vanish without appearing in a diff.

**Nothing was lost.**
- Non-force push confirmed: `e513fe620` is an ancestor of the new head, so history is intact.
- Parents are `e513fe620` (branch) and `e510501ad` (main). Main merged in, not rebased.
- Zero files deleted relative to main.
- Four `.changeset/` files are gone relative to the old branch head — `lane-897-endpoint-watch-teardown`, `quiet-creds-expiry`, `quiet-managers-retry`, `tidy-shards-conflict`. All four were deleted **on main** by `80c42e43c chore(release): version packages (#1055)`, which consumed them. The merge carried main's deletion correctly; the branch did not drop anything.
- `bin/smoke/ci-suites.txt` is 784 lines on both main and the merged head, and a `comm` against main's sorted list shows no entry missing. Main's suite list survived whole.
- All six shipped `writeIdentityPin` call sites are present at the merged head (`up.ts` ×3, `auth-proc.ts`, `delivery-proc.ts`, `manager-proc.ts`). Line numbers moved, consistent with main adding code above them.

Per-file hunks preserved, branch side / main side: `ci-suites.txt` 1/5, `clean.ts` 1/6, `down.ts` 3/6, `delivery-proc.ts` 4/27, `manager-proc.ts` 5/29.

**The suite registration moved to a fragment.** It is now `bin/smoke/ci-suites.d/fa0e66d75273c13d66c243bfc5e9ff7aff2d9996f3c068bd5bd022857dd7adf1.txt`, and I confirmed that filename is `sha256("smoke:pid-identity")` rather than trusting it. That path is written by no other branch, so this PR can no longer collide with #880 on the suite list.

**Gates**, run in a worktree confirmed clean at exactly the pushed commit: `pnpm install --frozen-lockfile` 0, `pnpm typecheck` 0, `pnpm build` 0, `smoke:pid-identity` 0 (22/22), `smoke:gate-inventory` 0, `smoke:shard-stability` 0, `smoke:ci-fragments` 0.

## Two things a reviewer should know rather than discover

**This is an evil merge.** The head is a single merge commit that also carries the fragment move, so `git log --cc` shows changes belonging to neither parent. That is legal and it is deliberate, but it means the fragment addition cannot be reviewed as its own commit and has to be read out of the merge.

**The design question is untouched and still open.** A live record with no pin still refuses loudly. That was left alone on purpose so it stays the reviewer's decision, not something settled quietly inside a conflict resolution. The upgrade case — an operator with a stack already running, whose pidfiles predate this change — is the thing to grade.

One cosmetic leftover: the fragment's comment still ends "Appended; shard assignments unchanged", which described the old `ci-suites.txt` append rather than a fragment. Harmless and true of the effect, just stale in its wording.

davidfarah2003:
CI has run for the first time on this branch and returned two failures. Both are this PR's, not inherited: main is green on the same jobs at `e510501ad` (runs 33787575552, 33781623400, 33736917969 all success).

## 1. `live` — the feature refuses a teardown of a process it just started

This is the serious one. From the `live` job:

```
✗ refusing to stop delivery daemon at /tmp/cotal-upstack-root-qw4NnP/.cotal/delivery.6d61696e.pid:
  its identity pin ...delivery.6d61696e.pid.identity names pid 4480, not the pidfile's pid
  - the pair was torn by a crash between writes. The record is preserved.
NEXT: confirm which process is really running (`ps`), then remove the pidfile and its pin by hand.
✗ not stopping nats-server because an earlier component did not stop
✗ not cleanly stopped - keeping artifacts and the registry entry
Error: FAIL: down exits 0
```

No crash occurred. This is an ordinary `up` followed by `down` in `up-stack-live`, and the torn-pin refusal fired on the happy path — then cascaded, because the delivery refusal stopped the broker teardown too.

The write order in `delivery-proc.ts` looks right on its face:

```
150  const pidPath = canonicalLocalProcessPath(DELIVERY_PIDFILE, ctx(space));
151  writeFileSync(pidPath, String(child.pid)); // canonical, never a pre-upgrade name
152  // #969: pin the pid to its process start so a later teardown can refuse a reused pid.
153  writeIdentityPin(pidPath, child.pid ?? 0);
```

Same pid, same path, adjacent lines. So the pin is not wrong when written — something rewrites the pidfile afterwards without updating the pin, or a second writer reaches that path. The pidfile name in the failure is `delivery.6d61696e.pid`, and `6d61696e` is hex for `main`, so this is the canonical per-space path that came from the main side of the merge. That makes the pin/path interaction between the two sides of this merge the first place to look rather than the pin logic itself.

The refusal message is also worth noting on its own terms: it asserts "the pair was torn by a crash between writes" as a diagnosis, and here no crash happened. A refusal that names a cause it cannot know sends an operator looking for the wrong thing.

## 2. `unit` — two em dashes in `docs/cli.md`

```
Docs voice check failed:
  cli.md:272: em dash
  cli.md:273: em dash
```

Both are lines this PR adds, confirmed against the merge base:

```
+every stop path — `down` for the broker, web and extension components, and the manager, delivery
+and auth-service stops — runs the same open-verify-terminate rule: ...
```

The repo's docs voice forbids em dashes. Mechanical fix, but note it gates the whole `unit` job, and `check:docsbundle` runs `check:docs-voice` first and short-circuits, so a stale docs bundle can hide behind this until the dashes are gone.

## Note for reviewers

The `live` failure is not reproducible by either review seat: their briefs forbid running any `-live` suite on this host, which is a standing safety rule here and not a gap in their work. CI is the only instrument that can see this one, so it is being handed to them rather than left to be discovered.

The four `smoke` shards were still pending when this was written.

davidfarah2003:
## Design decision: the legacy policy is reversing

An independent design review of this head blocked on the policy, and I am accepting it. **A live record with no pin becomes signalable with a loud warning, rather than a loud refusal.**

The reviewer drove all five states with real child processes — dead/no-pin and dead/torn both clear; live/no-pin refuses and preserves; matching signals and clears; mismatch refuses and preserves — and argued from the asymmetry: the reused-pid risk is conditional, while refusal on the first post-upgrade teardown is certain.

Asked directly whether fixing the refusal's wording changed its verdict, it said no, and the reason is the right one: `cotal down` still cannot do its defining job until an operator identifies and stops each process through raw OS means. A better message makes that detour safer without restoring compatibility.

One thing reinforces it further. Unpinned records exist only until the stack is next relaunched, so the loud refusal breaks the upgrade for every operator in order to protect a window that closes on its own during that same upgrade. Once anything is relaunched under this code every record carries a pin and the match/mismatch protection applies in full — which is the part actually worth having, and it is kept.

**Only the no-pin case moves.** A mismatched pin still refuses and preserves the record. A torn pin still refuses. The warning on the signalable path must state that the record predates identity pinning, that it is being signalled without an identity check, and that a relaunch will pin it.

The suite's legacy-refusal cells now assert the wrong behaviour and are being rewritten, since a cell asserting the old policy would make this change read as the regression.

Closed PR #1057 chose this same policy. The reviewer reached it independently and cold — its brief made no reference to that PR, deliberately, so it could not be steered. Two parties converging from opposite starting points is why a shipped design decision is changing on a single verdict rather than a second opinion being commissioned. I have noted this on #1057, whose closure should not read as a rejection of its whole approach; the reason that PR closed was that its gate could never fire, which is unrelated to its policy.

## Also in the batch

**The refusal's remedy is wrong, and this was measured rather than read.** The reviewer followed the refusal's own `ps` path, stopped the child externally, retried teardown, and it auto-cleared. The message tells operators to remove the pidfile and its pin by hand when they do not need to. It will say: stop the process, then rerun. The phrase "the pair was torn by a crash between writes" is also going — it asserts a cause the code cannot know, and in the CI failure no crash had occurred.

**The committed docs bundle is stale.** Running `pnpm gen:docsbundle` independently changes `extensions/connector-core/src/docs-bundle.generated.ts`. CI could not have shown this: `check:docsbundle` runs `check:docs-voice` first and short-circuits, so the em dashes at `docs/cli.md:272-273` are currently hiding it. Both are in the same push.

**The `live` failure is a fixture defect, not a product one.** `bin/smoke/up-stack-live.smoke.ts:158-168` tests a lost delivery lease: daemon A has pidfile A and pin A, the suite removes only the pidfile, a refresh launches B which writes pidfile B and pin B then exits, and line 168 restores only the pid. Pin B is left behind, so `down` meets a genuinely torn pair and correctly refuses. The pid-only restore came from `0e673ffe7` and exists on both parents; main is green because main has no identity check to trip. The fix is in the suite — restore the identity sibling alongside the pidfile, so the record is preserved as the pair it is. A scan established this is the only live-suite pidfile restoration that omits its sibling.

That diagnosis is being independently confirmed against both parents before it is acted on, along with a question it raises: whether anything in production can produce a half-restored record — a crash between the two writes at `delivery-proc.ts` 151 and 153, an interrupted `up`, a partial rollback. If it can, the fixture is accidentally testing a real scenario and fixing the test would be the wrong remedy.

The head is frozen at `fa7e41575` until that returns. Everything above lands in one push.

davidfarah2003:
## Merge preservation: approved, and one of my leads was wrong

The independent merge review of `fa7e41575` approves. All five conflicted files keep both sides: main's per-space records, candidate resolution, reclaim and log threading are intact, and the branch's pins survive. `smoke:pid-identity` reaches the gate exactly once, via the fragment `fa0e66d752…` and not via `ci-suites.txt`. Gates run independently, all exit 0: `install --frozen-lockfile`, `typecheck`, `smoke:pid-identity`, `smoke:gate-inventory`, `smoke:ci-fragments`, `smoke:shard-stability` — the same green the lane reported, reproduced rather than inherited.

**Correcting something I wrote earlier in this thread.** I suggested the `live` failure's lead was main's canonical per-space path meeting the branch's pin — that a pin might be written beside one path and verified at another. That is not what is happening. The pin is written beside the same path the stop verifies: canonical write, `localProcessPath` read, and after start they are the same file. The hypothesis was plausible and wrong, and anyone reading this thread should not carry it forward.

## The `live` failure, now confirmed from two directions

Both the lane and the merge reviewer reached the same answer independently, neither by running the forbidden suite:

`up-stack-live.smoke.ts:158-168` restores only the delivery pidfile after a losing refresh has already overwritten the pin. That pid-only restore is commit `0e673ffe7`, an **ancestor of both parents**. The merge took main's per-space live-suite paths unchanged — zero hunks against main on that file — and dropped no pin-aware restore, because neither parent had one. Main is green only because main has no pin check to trip.

So: not a lost hunk, not merge-caused, and refusing a disagreeing pair is the correct product behaviour. The fixture manufactures the disagreement by restoring one half of a two-file record.

**The production question, which is the one that decides the remedy.** Shipped start and stop never restore a pidfile without its pin, so a fixture-only fix is right. But the two writes in `startDeliveryDetached` are adjacent `writeFileSync` calls and are not atomic, so the window is real rather than theoretical: a crash between them after a lost pidfile produces genuine torn-pairing, while a crash between them on a *first* start produces the legacy unpinned shape instead. Both fail closed today.

Both reviewers independently flagged the refusal text "the pair was torn by a crash between writes" as a guessed cause. In this CI cell it is a pid-only rewrite and no crash occurred.

## Landing

The hold is released and the batch is one push: the policy reversal described above, the refusal message and its remedy, the live-suite restore of both halves of the record, and the docs items including the stale generated bundle.

Because the policy change touches the stop paths — which are the conflicted files — both review lenses will need a bounded delta re-grade against the new head. The approval above binds `fa7e41575` and nothing after it.

davidfarah2003:
## Both lenses approve `5a760217762a0ecf4e542f024d876fac98512daf`

**Design lens — APPROVE.** The policy reversal is correctly bounded to no-pin legacy records. Verified against its artifact rather than its report, and it re-drove the states rather than reading the diff, which was the condition it was given precisely because it was grading its own recommendation:

- live / no pin: signalled exactly once, exited 0, cleared; the warning says the record "predates process identity pinning" and "A relaunch will pin the process identity"
- live / mismatched pin: refused, zero signals, child survived, both records preserved
- live / torn pin: refused, zero signals, records preserved
- live / matching pin: signalled, exited 0, records cleared
- dead / no pin, torn, mismatched, matching: all `gone`, no signal, both record paths cleared

It also drove the no-pin case through all four shipped stop surfaces — delivery, manager, auth, and generic `stopLocalProcess` — rather than one, and confirmed the phrase "torn by a crash between writes" is absent rather than merely edited. The mismatch case still refusing is the load-bearing result: it is what a careless implementation of "make legacy signalable" widens by accident.

**Merge lens — APPROVE.** The policy delta did not disturb main's half. `reclaimDeadPreUpgradeRecord`, the canonical writes, `localProcessPath` reads, `down`'s `localProcessPathCandidates`, the spaceKey auth names and the log threading are all absent from the diff; the four stop paths change only the identity gate. The fragment still reaches the gate exactly once and is not in `ci-suites.txt`. Gates were re-run at this SHA rather than carried forward from round 1.

On the fixture: the lost-lease cell still drops the record, runs `up`, asserts non-zero plus "exited without becoming ready" plus holder alive, and then restores the pidfile **and** the `.identity` from the A snapshot. Pair restore, scenario not defanged.

## CI

`unit` and `live` both pass at this head, including the Core boundary and mesh preflight step that was red before. That confirms the round-1 diagnosis: the earlier `live` failure was a stale fixture — a pid-only restore from `0e673ffe7`, present on both parents — meeting the branch's pin, and not something the merge introduced. All three parts of that prediction held.

Four smoke shards are still running. **Not merging until they report**, because unlike a docs-only change this one alters shipped process-lifecycle code and `smoke:pid-identity` executes inside a shard.

## Two findings carded rather than carried

- **#1238** — the pidfile and its identity pin are published by two non-atomic writes. Both reviewers and the lane agree it is acceptable here and should be hardened transactionally rather than by reordering the writes: writing the pin first is benign only on a first start, and on a replace it leaves old pid beside new pin, which is worse. Filed so an accepted residual has a name attached.
- **#1239** — `smoke:pid-identity` needs a build its own command does not declare. Established as pre-existing rather than introduced here: the script text is identical at `fa7e41575` and `5a7602177`, and the mutation fixture has carried a `pnpm build &&` prefix since `fa52196d4`. Green in CI because the shard builds first; red in a fresh checkout.

davidfarah2003:
Merged as `917eb61a0e1ea087ed53e2140b2e36fb7080aff2`.

Verified rather than assumed: parent1 `2fe472520`, parent2 `5a760217762a0ecf4e542f024d876fac98512daf` — exactly the head both lenses approved, so nothing rode in with it. 22 files, +653/−17. CI was 17/17 green at that head, including `live`. #969 auto-closed; **#909 stays open**, correctly, since it is broader than #969 and its Windows detached-handle half arrives with #880.

Main's own run at the merge commit is being watched separately. A green PR head is not a green main, and whoever merges owns the merge commit's run.

## What actually shipped, versus what was proposed

The PR that landed is not the PR that was opened, and the differences came from review rather than from the author or me.

**The legacy policy is inverted from the original design.** A live record with no pin is now signalled with a loud warning, not refused. That reversal came from an independent design review that drove all five pid/pin states with real child processes and argued the asymmetry: the reused-pid risk is conditional, the refusal on the first post-upgrade teardown is certain. Unpinned records exist only until the next relaunch, so the original policy broke the upgrade for every operator in order to protect a window that closes on its own during that same upgrade.

Closed PR #1057 had chosen this same policy. The reviewer reached it independently and cold — its brief withheld that PR deliberately so it could not be steered. What shipped is the synthesis of the two: **#1069's writers and match/mismatch checks, with #1057's legacy policy.** Neither PR had both halves.

**The refusal text changed on measurement, not on opinion.** It previously told operators to remove the pidfile and its pin by hand. The reviewer followed the message's own `ps` path, stopped the child externally, retried teardown, and watched it clear itself — so the message now says stop the process and rerun. It also no longer asserts "the pair was torn by a crash between writes"; both review lenses independently flagged that as a cause the code cannot know, and in the CI failure that surfaced it, no crash had occurred.

**The `live` failure was a stale fixture, established from two directions.** `up-stack-live` restored only the delivery pidfile after a losing refresh had overwritten the pin, manufacturing a torn pair the product then correctly refused. The pid-only restore is `0e673ffe7`, an ancestor of both parents; main was green only because main had no pin check to trip. The fixture now restores both halves of the record, and the lost-lease cell still asserts non-zero, "exited without becoming ready", and holder-alive — so it was fixed rather than defanged.

**The suite registration moved to a fragment**, `ci-suites.d/fa0e66d752…`, which is `sha256("smoke:pid-identity")`. That path is written by no other branch, so this PR can no longer collide with #880 on the suite list.

## Carried forward rather than dropped

- **#1236** — `advisory-lock.ts` still holds its own copy of the start-token read. #1057 had collapsed them; that hunk did not survive its closure.
- **#1238** — the pidfile and its pin are published by two non-atomic writes. Accepted here by three parties, with the reasoning recorded for why reordering is not the fix: writing the pin first is benign only on a first start, and on a replace it leaves old pid beside new pin, which is worse.
- **#1239** — `smoke:pid-identity` needs a build its own command does not declare. Established as pre-existing at both SHAs, not introduced here.

davidfarah2003:
Follow-up on the merge commit's own run, since a green PR head is not a green main.

`917eb61a0` went **red** on `CI` (run `33804560310`): `smoke (shard 3/4)` failed in
`smoke:web-bounded-aggregation` at cell 6.0, `Port 39589 is in use`, with 6.1-6.10 downstream of that
one collision. The other four workflows at that commit (Windows, Docs, Changesets, Code Quality) were
green.

It is not this change. `git diff --name-only 2fe472520..917eb61a0` is 22 files, none under
`implementations/web`. The same job passed on `changeset-release/main` at `996d381b2`, and
`git merge-base --is-ancestor 917eb61a0 996d381b2` is true, so that green run covered content
including this merge. Cause is an ask-then-bind race in the suite's local `freePort`, now filed as
#1240 with the measurements.

One thing worth flagging here rather than only in #1240: the shard stops at first failure, and
**`smoke:pid-identity` was in the 40-suite never-ran set**, so this PR's new suite did not execute on
main's own run. It ran and passed there (22 checks) in the `996d381b2` control, so the coverage
exists - but it was not demonstrated by main's run at this commit, and I would rather say that than
let the merge record imply otherwise.

Failed jobs re-run as attempt 2; I will record the outcome here.

davidfarah2003:
Outcome, as promised.

Attempt 2 of run `33804560310` is **green**: all seven jobs succeeded, and all five workflows at
`917eb61a0` are now success (CI, Windows, Docs, Changesets, Code Quality). Main is still at this
merge.

Two things the re-run settles rather than papers over:

- **`smoke:pid-identity` executed on main this time** and passed, 22 checks. Attempt 2's shard 3
  reports zero never-ran suites, so the 40-suite coverage hole I flagged above is closed on main's
  own run, not only in the `996d381b2` control.
- **`web bounded-aggregation smoke: 48 passed, 0 failed`**, against `37 passed, 11 failed` on
  attempt 1. 37 + 11 = 48, so every cell that failed was downstream of the one port collision and
  no second defect was hiding behind it. That is the check that makes "flake" a measurement here
  rather than an assumption.

The underlying ask-then-bind race is #1240 and is not closed by this re-run.

PR #1070, opened 2026-08-30
fix(core): settle describe retry publish failures

davidfarah2003:
## Summary

- catch synchronous NATS publication failures from the reserved `describe` retry timer
- reject the pending describe as `unavailable` instead of allowing a timer-originated uncaught exception
- keep resolved commands exactly one-shot and keep every describe attempt inside the original deadline
- add real-broker and hermetic regression coverage, including mutations for timer safety and retry confinement
- update the caller-visible control-surface documentation and bundled copy

## Regression

PR #1026 introduced bounded re-publication of the read-only `describe` bootstrap during responder startup. That benefit is retained, but the interval callback published without a guard. If the NATS connection closed or began draining before the next tick, `publish` threw `ClosedConnectionError` synchronously from the timer and crashed the process.

Before #1026, a mid-wait connection closure remained a bounded describe failure. After #1026, the same shutdown path could become an uncaught exception. This change restores a clean failure for the regression already present on main.

Rejecting the pending operation is preferable to silently stopping retries because a closed or draining connection cannot produce a successful describe. The caller learns the transport failed immediately instead of waiting out the remaining deadline.

## Reproduction

Against current main at `bb54b7fcb8771f094e1e18d4e68d672f6f8a591d`:

1. Spawn an isolated loopback `nats-server`.
2. Start an unanswered describe with a 1000 ms deadline.
3. Drain the connection at 100 ms.
4. Observe process exit 1 from uncaught `ClosedConnectionError` at the first 250 ms retry in `endpoint-invoke.ts:148`.

With this branch, the real-broker regression cell observes an `unavailable` rejection and the process remains alive.

## Safety properties

- only reserved, read-only `describe` is re-published
- resolved commands remain on `invokeCommand -> epCall` and publish exactly once
- retries reuse the same bound request bytes and the original deadline
- publication failure settles the existing pending describe and creates no new timer or budget

## Verification

- `pnpm typecheck`, including the full workspace build
- `pnpm smoke:ep-invoke` (11/11)
- `pnpm smoke:ep-rail-failure` (25/25 against a real isolated broker)
- `pnpm mutation-proof --config packages/core/smoke/endpoint-describe-retry.mutations.json` (3/3 killed on named cells)
- `pnpm smoke:mutation-fixtures` (219 fixtures, 1222 anchors)
- `pnpm check:docsbundle`
- `pnpm changeset status`
- `git diff --check`


davidfarah2003:
Superseded by #1068, which carries the verified fix from the original follow-up branch.

Issue #965, opened 2026-08-28
the manager discovers a missing harness binary at first spawn rather than at boot, and there is no service install path


davidfarah2003:
The manager discovers that a harness binary is missing at first spawn, long after the point where
it could have said so, and there is no supported way to run the manager under an OS service
manager.

## Binary resolution

A manager launched with a scrubbed environment came up healthy and only failed at the first spawn
attempt, with "claude harness needs claude on PATH - not found". The cause was that the launch
dropped `~/.local/bin`. Everything between boot and that first spawn reported a working manager.

`cotal runtimes` already probes whether each runtime is reachable. The manager should do the same
work at boot: resolve each harness binary, record the absolute path on the mesh record, and fail
loudly there rather than at first use. That extends `runtimes` from "reachable" to "resolved,
pinned, recorded", and it means a PATH problem is reported by the thing that has the PATH problem.

## Service installer

Running the manager under launchd or systemd is currently local knowledge: an operator has to
write the unit themselves, and getting the environment right is the part that failed above. A
manager started from a terminal dies with the terminal, which is the root of the recurring fleet
loss.

Add `cotal service install`, emitting a launchd agent on macOS or a systemd user unit on Linux
around `cotal supervise`, with the resolved PATH written into the unit rather than inherited.

## Done when

A manager launched with a broken PATH refuses at boot and names the binary it could not resolve,
and `cotal service install` produces a unit that survives closing every terminal and logging out.

davidfarah2003:
[...]

Issue #402, opened 2026-08-14
cotal_persona is write-only: there is no mesh-side read of the persona catalog


davidfarah2003:
## Summary

`cotal_persona` defines a persona and nothing else — there is no tool that lists or reads them back. An agent on the mesh can *create* a persona but cannot *enumerate* the ones that already exist, so it cannot tell whether the name it is about to define is already taken, what a teammate's persona says, or which personas are available to `cotal_spawn`.

The operator-side path exists and works: `cotal personas list` / `cotal personas show <name>` read `.cotal/agents/*.md` directly, offline, no mesh. But that is a shell command, not a tool, so an agent reaches it only by shelling out — which a connector may not permit, and which is not available to a host that exposes the `cotal_*` tools and nothing else.

## Why it matters now

The discovery load genuinely rests on that CLI path. The other two things an agent can rely on are `cotal_spawn` failing loud on a name that does not exist, which reports absence only at the point of use, and the tool's own return value, which tells the definer about the one persona it just wrote and nothing about any other. Neither is enumeration.

## Suggested behavior

A read counterpart to the write path — a `cotal_personas` list/show tool, or a read mode on `cotal_persona` — returning the same catalog `cotal personas list` renders: name, role, model, and the description, scoped by the same ownership rules the write path already enforces.

davidfarah2003:
**Phase-1 triage - confirmed**

Tested at `ccd99461a9e79806e8eac201c1861941b558c4f0`.

Command:
```bash
set -eu
cd /Users/david/cotal-mac-fleet/repo/.triage-402-workspace
/Users/david/cotal-mac-fleet/repo/.triage-402-src-run/node_modules/.bin/tsx /Users/david/cotal-mac-fleet/repo/.triage-402-src-run/.triage-402-probe.ts
```

Observed:
```text
SHA ccd99461a9e79806e8eac201c1861941b558c4f0
CONTROL catalog file .cotal/agents/control-persona.md:
---
name: control-persona
role: reviewer
model: control-model
subscribe:
  - control
---
Positive control persona.

MESH TOOL NAMES:
cotal_orientation
cotal_docs
cotal_roster
cotal_inbox
cotal_send
cotal_dm
cotal_anycast
cotal_status
cotal_channel_info
cotal_channels
cotal_channel_mode
cotal_join
cotal_leave
cotal_spawn
cotal_feedback
cotal_despawn
cotal_persona
cotal_reconnect
[...]

PR #1073, opened 2026-08-30
fix(manager): report unavailable harnesses at boot

davidfarah2003:
## Summary
- cache connector harness requirements in the extension manifest without defeating lazy package imports
- resolve every declared harness at manager boot, report unavailable connectors loudly, and keep unrelated manager work running
- expose boot availability and absolute resolved paths in manager status
- pass the resolved map into managed launches so connectors use the exact executable boot inspected
- document that service installation remains a separate unsupported feature

## Boot decision
A missing harness degrades only that connector. The manager starts for unrelated lifecycle work, but names the unavailable connector before reporting ready and records the same reason in status. It does not silently degrade, and it does not crash-loop the whole manager because one optional harness is absent.

## Verification
- pre-fix built-manager repro: start completed silently, then first spawn alone returned the missing-claude error
- post-fix built-manager proof: start printed the missing-claude diagnostic before any spawn
- pnpm smoke:manager-harness-boot (5 checks)
- node scripts/mutation-proof.mjs --config implementations/manager/smoke/mutations/harness-boot.json (3/3 killed by named assertions)
- node scripts/mutation-coverage.mjs implementations/manager/smoke/mutations/harness-boot.json
- pnpm smoke:start-model
- pnpm smoke:manager-service
- pnpm smoke:manager-service-invoke
- connector launch regressions: launch-options, codex-args, jcode-args, Hermes launch-env, pi
- workspace seed/setup/materialize/import compatibility smokes
- changed-package typechecks and manager/CLI builds
- pnpm check:shard-stability origin/main HEAD
- pnpm check:docsbundle
- pnpm changeset status

## Limits
- Does not implement cotal service install. launchd/systemd unit generation and logout-survival are a separate cross-platform design and acceptance task.
- Does not run forbidden stack lifecycle, smoke:ci, check, or live suites on the shared box.

Closes the boot-detection half of #965.


davidfarah2003 (commented):

- Verified the diff matches the body's 5 claims: connector-requirement caching in the extension manifest (packages/workspace/src/extensions.ts CachedConnector/cacheConnector/extensionConnectors), boot-time resolution + loud unavailable logging (implementations/manager/src/manager.ts inspectConnectorsAtBoot, called first thing in start()), resolvedBinaries exposed on manager status (manager-service-contract.ts schema + revision 7→8, consistently updated in both manager-service.smoke.ts and manager-service-ops.smoke.ts), resolvedBinaries plumbed into buildLaunch, and docs/run-a-mesh.md documents both the boot behavior and the "no service install" limit.
- bin/smoke/ci-suites.txt append (`smoke:manager-harness-boot`) is at the absolute tail — confirmed with `tail`/`wc -l`.
- mutations/harness-boot.json's 3 mutations target real production logic in manager.ts (drop the boot inspection call, rewrite the resolved path back to the bare name, drop resolvedBinaries from the launch options) and each expectRed maps 1:1 to a named smoke assertion; the smoke file has exactly 5 checks as claimed.
- Blocker — silent operator-override clobber: extensions/connector-core/src/launch.ts:96-99 documents OPERATOR_ENV_KEEP names as safe to auto-forward specifically because "no connector assigns it per spawn," and lists COTAL_CODEX_BIN/COTAL_OPENCODE_BIN (launch.ts:125-126) under that guarantee. This PR now assigns both per spawn: extensions/connector-codex/src/extension.ts:165 and extensions/connector-opencode/src/extension.ts:163 unconditionally overwrite `env.COTAL_CODEX_BIN`/`env.COTAL_OPENCODE_BIN` with the boot-resolved PATH lookup whenever `resolvedBinaries` is present — which is true for essentially every managed/detached spawn once the harness is installed. Any operator or automation that set `COTAL_CODEX_BIN`/`COTAL_OPENCODE_BIN` machine-wide to pin a specific binary (the exact use the comment describes, and the same var host.ts already reads as a real override, not just in tests: extensions/connector-codex/src/host.ts:354,1174) has that pin silently discarded for detached spawns, with no log line and no doc/changeset mention. Neither codex-args.smoke.ts nor jcode-args.smoke.ts nor the Hermes launch-env smoke (which name-checks COTAL_CODEX_BIN passthrough but never sets `resolvedBinaries`) exercises this override-vs-resolved-path interaction, so it isn't covered by the cited verification. Either the comment's invariant needs updating to reflect the new exception, or the connector should not overwrite an already-set value.

davidfarah2003:
BLOCK

The boot detection is reached and the missing-connector timing is correct, but the absolute-path guarantee is false for valid relative PATH entries, and the recorded path is then used from the wrong cwd at managed launch.

## Blocking finding

`resolveOnPath()` returns `join(dir, bin)` without making it absolute. With `PATH=.review-relative-<pid>`, Manager boot finds the real executable and records:

```json
"state": "available",
"binaries": { "relative-harness": ".review-relative-<pid>/relative-harness" }
```

That contradicts the new status/LaunchOpts contract, which says boot records absolute paths. More importantly, a managed agent can launch with a cwd different from the manager's cwd. The connector receives that relative string in `resolvedBinaries`, uses it as the command, and node-pty resolves it under the agent cwd instead of the boot cwd.

Real Manager + real PtyRuntime reproduction at exact head `37e9ed885cb26dd0c6cd5ff81af6e0bcb22f57d2`:

- executable existed at `/home/cotal/Projects/Cotal-AI/work-worktrees/1073-review/.review-relative-.../relative-harness`
- boot recorded `.review-relative-.../relative-harness` and marked the connector available
- launch cwd was a separate temporary directory
- launch resolved the command under that temporary cwd
- `startAgent` failed: `execvp(3) failed: No such file or directory`

Artifact: `/tmp/jc-2ce6aed61d6c/home/scratch/1073-relative-path.out`.

Fix the shared resolver or normalize the boot hit before status/launch propagation, and add a mutation/invariant for the dangerous direction: an available harness found through a relative PATH entry must remain the exact same absolute executable when the managed cwd differs.

## What passed

- Real before/after Manager boot path: parent start was silent and first spawn alone named the missing harness; this head named it before spawn and retained the same spawn refusal.
- Continuing boot with a named unavailable connector is acceptable here: only that connector is refused at pre-mint acceptance, status and stderr retain the reason, and unrelated connector work remains available. This is loud partial availability, not a silent fallback.
- Present absolute-path fixture and launch map pass.
- All 3 existing mutations were KILLED red-and-named. They prove dependency on boot inspection, recorded-path value, and launch-map forwarding. None guards the relative-PATH/wrong-cwd direction above.
- Build, typecheck, docs bundle, changeset, shard stability, Manager service/start-model gates, connector launch regressions, and mutation fixture integrity passed.
- Service installation exclusion is principled: launchd/systemd unit generation, lifecycle ownership, logout survival, and cross-platform acceptance are separate from inspecting connector harnesses in an already-starting manager.

Load: initial `43.96/26.07/21.12`, final `28.17/26.73/24.47`; swap `8.0/8.0 GiB` used. Head was rechecked immediately before verdict and remained exact.

davidfarah2003 (commented):

- Verified inspectConnectorsAtBoot() (manager.ts) runs before attach.start()/ready, logs a named `! manager boot: connector X unavailable` line per missing harness, and keeps unrelated connectors available — matches the boot-decision claim.
- startAgent's pre-mint gate now consults the boot-resolved connectorStatuses first and only falls back to a live resolveOnPath scan for connectors with no boot row (post-boot registration) — no silent degrade, refusal reason is loud and named.
- resolvedBinaries threading verified end-to-end for all 6 connectors (claude-code, codex, opencode, jcode, hermes, pi): boot resolves absolute paths via resolveOnPath's now-fixed `resolve(cand)` (packages/workspace/src/bin-path.ts), and each connector either uses it directly as `command` or as an env fallback that an operator's explicit *_BIN pin still overrides (confirmed in codex-args.smoke.ts and boot-prompt.smoke.ts mutation-driven checks).
- bin/smoke/ci-suites.txt: both new entries (smoke:manager-harness-boot, smoke:bin-path) are appended at the absolute tail after smoke:connection-status — confirmed via diff hunk header (@@ -722,3 +722,12 @@ with no trailing context).
- Mutation proofs (harness-boot.json, binary-override.json x2, bin-path.mutations.json) target real shipped lines (manager.ts's inspectConnectorsAtBoot/resolvedBinaries wiring, bin-path.ts's resolve(cand), extension.ts env precedence) and their expectRed names match actual check() names in the smoke files — real both-polarity proof, not decorative.
- Minor nit: PR body claims smoke:manager-harness-boot is "5 checks" but harness-boot.smoke.ts has 6 check() calls — cosmetic miscount, doesn't affect the evidence's validity.
- Docs updated (docs/run-a-mesh.md + regenerated docs-bundle.generated.ts) to describe the new boot behavior and reaffirm cotal service install remains unsupported. No AI/tool attribution found anywhere in the diff.

davidfarah2003:
Terminal triage of 7c956269: both reds are one main-side failure, not this branch's. smoke (shard 2/4) dies at `ReferenceError: staleNow is not defined` in smoke:web-channel-authority (signature verified in job 99215883707's log), and ci-ok just aggregates it. The branch contains the #670 staleNow window and predates the fix that landed on main in #1113 (c52cdb3d) — a fold of current main cures it and retriggers CI. Everything else passed (15/17).

davidfarah2003:
Independent exact-head review at `7c9562691448a88b29cffb8dd375c8f7a469a9fc` (`git rev-parse HEAD` in detached worktree `/home/cotal/Projects/Cotal-AI/Cotal-rev1073`). `gh pr view 1073 --json state,mergedAt,headRefOid`: OPEN, `mergedAt` null, head matches. Did not run `cotal up`, `pnpm smoke:ci`, `pnpm check`, or any `*:live` suite.

## CI triage (the 09:01:30Z claim)

Two measurements, as assigned.

**1. Does the red reproduce at the merge base alone?**
Merge-base `4150297775847d66cf12aaa5a7a35adc846f409e` (`git merge-base origin/main HEAD`). Job 99215883707 (head sha `7c956269`) dies at `pnpm smoke:web-channel-authority` with `ReferenceError: staleNow is not defined` in `runBackfillFunction` (`channel-authority.smoke.ts:444`). The smoke extracts only `let (?:refreshing|selecting)` from `app.js` and then executes shipped `refresh()`, which reads `staleNow`.

I reproduced that exact failure mode with the smoke's own extract regex + `vm.runInContext` at **both** the merge-base worktree (`Cotal-rev1073-base`, `415029777`) and this head: `ReferenceError: staleNow is not defined`. Then I ran `pnpm smoke:web-channel-authority` at this head (core built): same `ReferenceError` at `app.js:116`. `#1113` (`c52cdb3d`) is an ancestor of current `origin/main` and **not** of this head or the merge-base. On `origin/main` the smoke extracts `refreshing|selecting|staleNow` (3 decls).

**2. Does this PR's file set touch the failing path?**
File list from **merge-base..HEAD**, not `origin/main..HEAD`. Positive controls hit: `packages/workspace/src/bin-path.ts`, `implementations/manager/src/manager.ts`. Negative control: no `channel-authority` / web smoke / `app.js` / `graph.js` in that list. `git diff --stat BASE HEAD -- implementations/web/` is empty. A zero here means absence, not a broken instrument.

**Triage ruling:** both reds are one main-side failure (`staleNow` backfill harness vs `#670` window). They are not this branch's. The 04:14:50Z BLOCK is a **different** finding (relative PATH) and is not retired by this CI story. Landing still needs a fold of current main if CI must be green on the PR.

## Earlier BLOCK findings, counted from the full text

**Finding 1** (issue comment 04:14:50Z at `37e9ed88`, the only blocking finding in that comment): `resolveOnPath()` returned `join(dir, bin)` without making it absolute, so a relative PATH hit was recorded and then exec'd from the managed cwd.

Disposed at this head. `packages/workspace/src/bin-path.ts` now `return resolve(cand)`. Live polarity:
- merge-base source, same fixture: `baseFound` is a relative path (`baseAbsolute: false`).
- head `pnpm smoke:bin-path`: green (`relative PATH entries resolve to the exact absolute executable`).
- head `pnpm smoke:manager-harness-boot`: 6/6, including `boot records the resolved absolute harness path from a relative PATH entry` and `managed spawn executes that exact harness when its launch cwd differs from manager boot cwd`.
- `bin-path.mutations.json`: 1/1 KILLED (`return resolve(cand)` → `return cand`, named cell).
- `harness-boot.json`: 3/3 KILLED, including the recorded-path rewrite and the launch-map drop.

**Finding 2** (landing-queue review 03:51:35Z at `37e9ed88`): managed launch silently overwrote operator `COTAL_CODEX_BIN` / `COTAL_OPENCODE_BIN` (names on `OPERATOR_ENV_KEEP` because "no connector assigns it per spawn").

Disposed at this head. Codex/OpenCode write `COTAL_*_RESOLVED_BIN` and hosts take `COTAL_*_BIN` first:
- `pnpm smoke:codex-args` green (35 checks), including `operator COTAL_CODEX_BIN wins over the manager boot fallback`.
- OpenCode `buildLaunch` with both set: `COTAL_OPENCODE_BIN=/operator/pinned-opencode` and `COTAL_OPENCODE_RESOLVED_BIN=/boot/resolved-opencode`.

## PR substance at this head

`inspectConnectorsAtBoot()` is the first await in `start()`. Installed connectors are read from the manifest cache (`extensionConnectors`) without importing packages. A missing harness logs `! manager boot: connector <name> unavailable` and records that row; other connectors stay `available`. `startAgent` refuses on the boot `unavailable` row and only re-scans PATH for a connector with no boot row. Status schema is revision 8 with `connectors`. Launch gets `resolvedBinaries: bootStatus?.binaries`. Connectors declare `requires`; the manager does not hardcode harness names. Claude/pi use the resolved path as `command`; codex/opencode keep the operator pin.

**Cache vs a changed manifest:** boot inspects the cached `requires`, not a live import. Missing cache throws and tells you to re-run `cotal ext add`. A present but stale cache (package changed `requires` without re-add) is not re-read. That is the lazy-import trade: named, not a silent fallback, and not a blocker for this PR.

**Exact executable boot inspected vs a path that changes later:** the recorded absolute path is what launch uses. A later PATH reordering will not swap the file. Replacing the file at that same absolute path is the same pathname. I did not claim inode identity across boot and launch.

**jcode/hermes:** they still assign `COTAL_JCODE_BIN` / `COTAL_HERMES_BIN` from `resolvedBinaries`. Those names are not on `OPERATOR_ENV_KEEP`, so this is not Finding 2. Note only.

Changeset `loud-manager-harnesses.md` lists the first-party packages this diff touches, including `@cotal-ai/connector-core`. `smoke:manager-harness-boot` and `smoke:bin-path` are appended at the tail of `ci-suites.txt`. Docs in `docs/run-a-mesh.md` cover boot reporting and that service install is still unsupported.

## Verdict: APPROVE

Exact sha graded: `7c9562691448a88b29cffb8dd375c8f7a469a9fc`.

Finding 1 and Finding 2 are fixed at this head with live both-polarity evidence. The shard 2/4 red is main's `staleNow` harness, not this file set. Fold main before treating GitHub CI as this branch's gate. Residual notes (stale-but-present connector cache; jcode/hermes BIN env names) are not blockers.

davidfarah2003:
## Merge disposition: approved, held on conflicts, needs an author for the fold

Recording this so the approval above does not sit here decaying without an explanation. Three
APPROVEs on this repo went stale tonight without anyone writing down why, and a reader finding an
unmerged approved PR should not have to reconstruct the reason.

**The review is complete and I checked it against the whole record rather than the latest comment.**

- The 04:14:50Z BLOCK carried one blocking finding: `resolveOnPath()` returned `join(dir, bin)`
  without absolutising, so a harness found through a relative PATH entry was recorded relative and
  then resolved under the managed agent's cwd rather than the manager's. The reviewer reports it
  fixed at this head, with the mutation the BLOCK explicitly demanded for the dangerous direction:
  base returns relative, head returns `resolve(cand)` absolute, harness-boot 6/6 including a
  different-cwd exec, bin-path 1/1 killed, harness-boot 3/3 killed.
- The 03:51:35Z review finding, `COTAL_CODEX_BIN` / `COTAL_OPENCODE_BIN` clobber, is fixed via
  `*_RESOLVED_BIN`, with `smoke:codex-args` 35 green including the pin cell and the OpenCode pin
  winning.

Both findings are accounted for. Nothing in the earlier verdicts is left open.

**The 09:01:30Z triage claim was tested rather than taken on trust, and it holds.** The shard 2/4
red reproduces at the merge base `41502977` as well as at this head (`ReferenceError: staleNow is
not defined`, via vm extract and a full `pnpm smoke:web-channel-authority`), and the branch's file
list against the merge base does not include `channel-authority`, with `bin-path.ts` and
`manager.ts` as positive controls confirming the file list was actually being read. `c52cdb3d`
(#1113) is on `main` and not on this head. So the failure is main-side and is not this branch's
defect.

**Why it is not merged.** As of 2026-08-30T10:22Z this PR is `CONFLICTING` / `DIRTY` against
`main`, which is now `d1a932d4`. That is a merge blocker independent of the review, and folding
current main is also what would clear the CI red above.

**What happens next.** This needs an author seat to fold current `main` and resolve the conflicts,
then a re-review of the refold. The re-review is not a formality: a refold has to be checked for
purity, by comparing the committed tree against `git merge-tree` of the two parents, because a
resolution that silently reverts a hunk produces bytes that look exactly like a correct one. A
file-level scan proves transfer for a named finding; only the tree comparison enumerates
resolutions nobody reported.

The reviewer keeps finding ownership and gets the verdict on the successor head. This PR is not
cleared to merge until that lands.

For completeness, the reviewer states what it did not run: `cotal up`, `smoke:ci`, `check`, and any
live suite. Those are correctly out of bounds on this host, and naming them is the right practice
rather than a gap.

davidfarah2003:
Folded current main into this branch (f85bada2). Two real conflicts beyond the usual generated set:

- **manager-service-contract.ts**: both sides independently minted revision 8 — main for `list-personas`/`show-persona` (#402), this branch for connector availability in `status` output. Kept main's 8 and minted this branch's surface change as **revision 9**, bumping the shipped literal. Note the literal auto-merged silently (both sides wrote `revision: 8`), which would have shipped a grown surface under a stood-still number — the exact case the file's own doctrine forbids — and no suite would have caught it since the ops smoke asserts `shipped.revision`.
- **manager-service-ops.smoke.ts**: took main's `shipped.revision` formulation for both hunks (the branch pinned `=== 8`).

Also migrated the two frozen-list appends (`smoke:manager-harness-boot`, `smoke:bin-path`) to `bin/smoke/ci-suites.d/` fragments; `ci-suites.txt` is byte-identical to main again.

Local gates on the fold: build + bin typecheck green; smoke:web-channel-authority 112, smoke:manager-harness-boot 6, smoke:bin-path 1, smoke:manager-service 25, smoke:manager-service-ops 67, smoke:manager-service-invoke 11 — all passed; docs regenerated; fragment validation 27 lines OK.

davidfarah2003:
## Two blockers found while checking this for merge

### 1. This branch would land ten scratch build logs on main

The branch tracks ten `.refold-*.out` files. I computed the machine merge of this head with current
main and they are all present in the result:

```
.refold-bin-path.out                       4 lines
.refold-build.out                         93
.refold-install.out                        0
.refold-manager-harness-boot.out          15
.refold-manager-service-invoke.out        25
.refold-manager-service-ops.out          104
.refold-manager-service.out               38
.refold-merge.out                         16
.refold-tc.out                             1
.refold-web-channel-authority.out        115
                                     -------
                                         411 lines
```

**The `*.out` ignore rule does not save this.** `.gitignore` was given `*.out` in #1125, and that
pattern is present in the merged tree — but `.gitignore` only prevents *untracked* files from being
added. These were committed to this branch before the rule existed, so they are tracked, and a
tracked file is unaffected by an ignore pattern. The merge carries them in.

This is the third time this class has reached main in one session: #1122 removed five such files,
#1056 restored them from a stale branch, and #1125 removed them again and closed the hole for new
files. This PR is the remaining exposure — branches that already tracked them.

**Fix, on this branch:**

```sh
git rm .refold-bin-path.out .refold-build.out .refold-install.out \
       .refold-manager-harness-boot.out .refold-manager-service-invoke.out \
       .refold-manager-service-ops.out .refold-manager-service.out \
       .refold-merge.out .refold-tc.out .refold-web-channel-authority.out
```

Then confirm the fix is actually in the commit rather than only in the worktree —
`git show HEAD --name-only` must list all ten as deletions. That check is here because I shipped
exactly this mistake on #1125 two hours ago: the staged set and the edited set were different, and
the controls passed against a working tree that did not match the commit.

### 2. The existing approval is stale

The approval on this PR names `7c956269`. The head is now `f85bada2357e53233a2f5d1e2a55208685a81d6c`,
which is six merges further on — it folds #1119, #1123, #1056, #1122, #1058 and #1120. An approval
binds the tree it named, so this one does not carry to the current head and a fresh terminal review
at the exact head is required before merge.

The reviewing seat that issued it is offline, so the re-review needs different eyes rather than a
re-confirmation from the same one.

### Good news on the third check

The resurrection check passes. This branch's base predates #1125, so the five `.refold2-*` files
main deleted are candidates for silent restoration — that is exactly how #1056 put them back. I
verified against the machine merge tree and **main's deletion survives**; nothing is resurrected.

The check, for anyone folding a branch based before a deletion, since it is invisible in the PR diff:

```sh
git diff --diff-filter=D --name-only $(git merge-base <branch> origin/main)..origin/main
```

Nothing here is a comment on the substance of the change, which I have not reviewed.

davidfarah2003:
CI triage for f85bada2 — every red classifies as inherited or cured, none from this branch:

- `smoke (shard 0/4)` (`manager-reconcile-ownership`, oracle signature ×4) and `smoke (shard 3/4)` (`secret-store-seam`, oracle signature): reproduce on pure main — #1137 (introduced by the #1097 merge; fix in progress).
- `smoke (shard 1/4)` (`smoke:gate-inventory`: "smoke:web-proxy-order exists but nothing runs it"): cured on main by #1140.
- `Windows / required`: the connector-claude-code PATH cell — identical single-cell failure on four independent branches, tracked as #1141.

Refolded to 9b161990: one merge of current main plus deletion of the committed refold logs; no branch-side content changed (`git log --no-merges f85bada2..9b161990 --not origin/main` is empty), so the classification carries over. Fold gates on 9b161990 all green (build, bin typecheck, web-channel-authority 112, boot suites, docs regen, fragments). Merging.

davidfarah2003:
BLOCK

Exact reviewed SHA: `9b161990fb110d6aeff73ec9f25d28ae509c8b3b` (the second parent of landed merge `2494bc83eeece797abdbdda429ff00a86be41b83`). The pull ref was reread immediately before this verdict and still resolves to that SHA.

The new manager boot inspection misdeclares the shipped Hermes harness. `extensions/connector-hermes/src/extension.ts` declares `requires: ["hermes"]`, but the shipped launch always executes `uv run --project ... hermes gateway run`, and the public docs require `uv` on PATH. On the exact landed tree I started the shipped manager twice with an installed cached Hermes connector:

- PATH containing only an executable `uv`: Hermes was classified `unavailable` with `hermes harness needs hermes on PATH - not found`, although this is the documented/supported prerequisite shape.
- PATH containing only an executable `hermes`: Hermes was classified `available`, although the shipped launcher would fail immediately because `uv` is absent.

This defeats the PR's fail-closed boot availability contract in both directions for a shipped connector. Repair by declaring/inspecting the executable the launch actually requires, with a shipped entry-point smoke and a named mutation proof covering both the supported `uv` case and the false-positive `hermes`-only case.

Green controls at this exact SHA: frozen install; full build; full typecheck; changeset status; manager harness boot; relative PATH; Codex and OpenCode operator pins; manager service, ops, and generic invoke workflows; mutation coverage plus all 6 named mutations killed; CI fragments; 4-shard stability (`459 -> 461`, zero pre-existing movement); generated docs/bundle reproduction; computed merge/fold reconstruction.

Unrun: forbidden `smoke:ci`, `check`, `cotal up/down/setup`, `ext seed --reset`, live suites, and the later tree-hygiene gate that did not exist at this historical head. Claude, Jcode, Pi, and an actual Hermes gateway launch were not run. The blocker is from the shipped manager boot path using installed extension metadata, not a hand-built connector input.

PR #1084, opened 2026-08-30
fix(cli): identify status renderer artifact

### LINKED ISSUE #420: status card: Machine block should identify the binary/install it describes
## Problem

The Machine section of the status card prints the version of whatever process renders it (`implementations/cli/src/commands/status.ts` — `row("cotal-ai", c.green(\`v${cliVersion()}\`))`), with nothing identifying *which* install that is.

On a machine where a source checkout (invoked via package scripts) and an installed CLI (get.cotal.ai shim) coexist — a completely normal state for anyone developing against a live mesh — this produces what looks like a version contradiction:

- status card rendered via the checkout: `cotal-ai v0.17.0`
- `cotal --version` via the installed shim: `cotal-ai 0.16.0`

Both are correct about themselves; neither says which artifact it is describing. An operator (or an agent) reading the two outputs side by side has no way to tell this is two binaries rather than one confused one, and the skills-skew rows compound it: `Claude skills stale` is computed against the *renderer's* version, so a source build calls the installed plugin generation stale even when it matches the installed CLI exactly.

## Suggestion

Print provenance next to the version in the Machine block — the resolved executable path or install root (e.g. `cotal-ai v0.17.0 (source checkout: /path/to/repo)` vs `v0.16.0 (installed: ~/.local/share/cotal)`). The skills rows could similarly name the version pair they compared. One line makes every mixed-artifact comparison self-identifying.

davidfarah2003:
Closes #420.

## Scope

- Identify the running `cotal-ai` artifact beside its version in the `cotal status` Machine block as a source checkout, installed package, or npx package root.
- Resolve package-manager bin symlinks before reporting the artifact, using the same package-root resolver as `cliVersion()`.
- Show the installed Claude skills version and running CLI version when that row is stale.
- Document the status behavior and add a patch changeset for the shipped CLI, binary, and embedded docs bundle.
- Add a hermetic status smoke at the true CI-suite tail plus mutation proof for the artifact row, symlink resolution, and stale version pair.

## Reproduction

On current main, the real status command was run twice with distinct resolved entries: the source `bin/cotal.ts` and a synthetic installed package entry. Their Machine rows were only `cotal-ai v0.36.0` and `cotal-ai v9.9.9`, with no artifact identity, despite the entries resolving to different paths.

## Proof

- Red first: `pnpm smoke:status-provenance` failed both source and installed identity cells on current main.
- `pnpm smoke:status-provenance`: 2 passed, 0 failed.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/status-provenance.json`: 3 of 3 killed on the named assertions.
- `pnpm check:shard-stability origin/main HEAD`: 424 pre-existing suites examined, zero changed shard, one suite appended.
- `pnpm smoke:gate-inventory`: green.
- `pnpm smoke:mutation-fixtures`: green, including all three unique code-only anchors.
- `pnpm --filter @cotal-ai/cli test`: green.
- `pnpm typecheck`: green, including the repository build and bin smoke typecheck.
- `pnpm gen:docsbundle` twice: fresh regeneration was byte-identical. A later `gen:tooldocs` + `gen:docsbundle` left both generated files clean.
- `pnpm changeset status`: patch plan includes the directly changed shipped packages and the fixed group.
- Detached throwaway worktree `git merge --no-commit --no-ff origin/main`: `Already up to date`, no conflict.
- `git diff --check`: green.

## What was NOT proven

- No real global npm install or live get.cotal.ai launcher was invoked. The installed-path cell uses a real filesystem package root and package-manager-style symlink, then drives the shipped status command.
- The npx label is classified by the existing `_npx` path predicate but is not a separate output cell in this change.
- No live mesh lifecycle command or lifecycle smoke was run, per host safety constraints.

## Current-main fold

Folded current `origin/main` with a real merge at head `085bc06a2689b7e6bf322a7877bc342897946652`. The resolution took main's `ci-suites.txt` ordering, re-appended `smoke:status-provenance` at the true tail, and regenerated the docs bundle. Post-fold proof: shard stability examined 428 pre-existing suites with zero moved, gate inventory passed with 429 unique CI suites, the focused status smoke passed 2/2 sequentially, mutation proof killed 3/3, docs regeneration was byte-identical, and GitHub reports the folded head MERGEABLE.


davidfarah2003 (commented):

- Diff matches scope: `cliProvenance()` (lib/version.ts) shares `cliPackageRoot()` with `cliVersion()` (seed/paths.ts), so both surfaces resolve through the same `entryScript()`/`realpathSync`, meaning package-manager bin symlinks resolve correctly for source and installed cases alike (verified on the branch).
- Evidence is real: `implementations/cli/smoke/status-provenance.smoke.ts` imports and drives the actual `status()` command (not a parallel formatter), asserting both the source-checkout and installed-package artifact labels plus the stale-skills version pair. `bin/smoke/mutations/status-provenance.json` names 3 mutations, each targeting the exact changed lines in status.ts/paths.ts, matching the "3 of 3 killed" claim.
- `bin/smoke/ci-suites.txt`: confirmed `smoke:status-provenance` is the literal last line of the file on the PR branch (717/717) — true tail append, not mid-file.
- No silent fallbacks: `cliPackageRoot()`, `cliVersion()`, and `entryScript()` all throw loudly on failure; `cliProvenance()`'s default to "installed" for an unrecognized layout is real classified data, not a hidden failure.
- docs/run-a-mesh.md updated to describe the new artifact-identity and stale-skills-version-pair behavior; no AI/tool attribution found anywhere in the diff.
- FYI (not a blocker): docs-bundle.generated.ts also regenerates the "Message flow" and "Watch a mesh" sections — that content already exists in docs/presence-and-delivery.md on main but was missing from the previously-compiled bundle; this PR's full-bundle regen sweeps it in as a side effect, unrelated to but not contradicting the stated scope.

PR #1085, opened 2026-08-30

davidfarah2003:
Re-mint of #921: after that PR's branch history rewrite, its pull_request events stopped producing CI runs entirely (close/reopen and an empty-commit push both minted only CodeQL), so it could not be validated there. Same branch, same tree. Review record: #921 carries a LANDING-QUEUE REVIEW: APPROVE-quality at this tree, including the ci-suites tail check.


davidfarah2003:
CI workflows never minted for the latest push here (Code Quality only — the known partial-minting failure). Closing and reopening as a fresh PR from the same branch, tip 71331c9a.

PR #1083, opened 2026-08-30
docs: drop an em dash from watch-a-mesh

davidfarah2003:
Replaces the em dash that arrived with #1067 (line 148 of docs/watch-a-mesh.md) with a colon. It reds check:docs-voice on every head built on current main.

PR #1086, opened 2026-08-30
fix: verify seat exit before manager succession

### LINKED ISSUE #817: A manager that exits leaves its seats alive holding its broker rails, so the successor can never take the frozen gate
Found while installing the #813 fix, which required a manager restart. The restart itself was blocked for several cycles by this.

## What happens

When the manager exits, **its seats do not**. They reparent to init and keep **live broker connections open as the dead manager's instance**. Measured immediately after a clean `SIGTERM` of the manager:

```
total node processes carrying COTAL_NAME:  46
of those, orphaned (ppid=1):               19
each holding 3-5 open sockets
```

Those connections make the instance look occupied, so the successor cannot take its issuance gate:

```
✗ the issuance gate for "tak1gt6q7zwdzuxe3tkz2w2rnmmjj0v" is frozen;
  another barrier holds it; re-read and re-decide (SPEC 13.8)
```

`cotal reconcile-gate` does its job — it verifies the holders are gone and evicts them, printing `verified evicted: local.U...` for each — but the surviving orphans **immediately re-take the gate**, so the next launch fails identically. Reconciling and relaunching in a loop does not converge; the orphans have to be killed first.

## Why it matters

- **Every unclean manager exit needs manual intervention.** A crash, an OOM kill, or a host reboot leaves the mesh unable to restart itself.
- **It compounds the outage it follows.** #813 already forced a restart; this turned a one-step restart into: reconcile, fail, enumerate orphans, kill, reconcile again, launch.
- **The orphans are already useless.** Their PTY parent is gone, so no operator or peer can reach them. They are pure blockage — dead agents holding live sockets.
- **The failure is silent and misleading.** The gate error names a barrier, not a process, so the natural reading is "some other manager is running" — which sent me looking for a competing supervisor that did not exist.

## Ask

1. **Reap seats on manager shutdown.** On SIGTERM/SIGINT, terminate the seats the manager owns before releasing the instance, so nothing survives holding its rails.
2. **If a seat must outlive the manager, it must not hold the manager's instance connection.** Surviving the parent is defensible for a long-running task; keeping the dead instance's broker rails open is not.
3. **Make `reconcile-gate` report what is re-taking the gate.** It currently proves the *recorded* holders are evicted; it should notice that live connections remain and name the pids, instead of succeeding into an unchanged state.
4. **Consider a startup self-heal**: a manager that finds its own previous instance frozen with only orphaned holders should be able to reclaim it, since it can prove the holders are parentless.

## Diagnosis note

An early check reported "0 orphaned pi processes" and looked clean. That was true and useless: the orphans are `node` processes (the jcode and opencode connector hosts), and only `pi` had been counted. Any check here has to enumerate by `COTAL_NAME` in the environment, not by process name.

Related: #813 (the outage that forced the restart), #773 (in-process credential adoption fails, which is why a manager can never re-mint without restarting).

davidfarah2003:
Closes #817.

## Design decision
Reap, rather than adopt or force-reclaim. Adoption cannot prove exclusive process/runtime control after an abrupt manager death because no crash-durable runtime handle is available. Force-reclaim is the dangerous two-owner direction. Reaping keeps the old lifecycle denied and the alias held until the predecessor is verifiably gone.

## Scope
- normal manager shutdown requests every seat stop and awaits each runtime's authoritative `waitForExit` before releasing the manager lease or service registration
- a failed or absent exit proof makes shutdown fail loud instead of reporting cleanly over a possible orphan
- crash-successor static reconciliation revokes deny-new, verify-evicts the predecessor principal through delivery-admin, then cleans up and retires the lifecycle; unverified eviction leaves the slot terminalizing and the alias held
- a real-process ephemeral-broker smoke creates an independently-owned seat, SIGKILLs its manager, and drives the same logical successor through the verified boundary

## Proof
- reproduction before coding: manager SIGKILL left the seat broker-live; the same logical successor entered exact-op terminalization while the orphan remained live
- `pnpm typecheck`
- `pnpm smoke:start-model`
- `pnpm smoke:static-lifecycle` (38/38 declared assertions)
- `pnpm smoke:lifecycle-e2e`
- `pnpm check:shard-stability <base> HEAD`
- mutation: removing normal shutdown exit waits KILLED red-and-named
- mutation: accepting unverified orphan eviction KILLED red-and-named on the two-owner guard

## Limits
- no forbidden `cotal up`/`cotal down`, `smoke:ci`, `check`, or `-live`/`:live` command was run
- `check:docsbundle` is blocked by a pre-existing docs-voice em dash in `docs/watch-a-mesh.md:148`; this PR's generated CLI bundle change is committed
- the real-process smoke uses a standalone ephemeral broker and an independently-owned process-group runtime; external tmux/cmux/herdr/orca apps were not exercised end to end
- sibling-manager slots remain outside automatic reconciliation exactly as before; only the same persisted logical instance reaps its owned slots


davidfarah2003 (commented):

- Diff matches the body: `teardownManagedAgents` (manager.ts:1673-1691) now awaits `awaitHandleExit` per seat before releasing authority, throwing loud on failure; `runStaticTerminal` (static-lifecycle.ts:291-293, plus the two earlier terminal branches at :234 and :254) verify-evicts via `hooks.evict(principal)` before cleanup/retire. All 5 shipped runtimes (pty.ts:144, tmux/cmux/herdr/orca runtime.ts) already implement `waitForExit`, so normal shutdown won't spuriously fail loud in production.
- Both-polarity, shipped-code-driving tests confirmed: start-model-preflight.smoke.ts adds pass (`exitProofs` for both seats) and fail (`teardownError` matches "lease1.*still running") cases around real `mgr.stop()`/`teardownManagedAgents`; static-lifecycle.smoke.ts flips `evictionVerified` false→true around real `M.reconcileStaticLifecycles()` and asserts terminalizing→retired.
- Mutation fixtures are well-targeted: `orphan-seat-eviction.mutations.json`'s `find` string (2-space `if`/4-space `throw`) is uniquely matched only by static-lifecycle.ts:291-292, the actual crash-recovery branch the orphan test exercises — not the two other textually-similar if/throw blocks at :234 and :254, which have different indentation. Good precision.
- `bin/smoke/ci-suites.txt`: append is at the absolute tail (verified 728-line file, `smoke:orphan-seat-succession` is the last line) — not mid-file.
- Gap: the PR's own Scope calls the new `orphan-seat-succession.smoke.ts` (real nats-server, real manager SIGKILL, real seat PID, verify-evict over the wire) the load-bearing repro+fix proof, and it's the only exerciser of the *real* `makeManagerEndpointEvictor` path through static-lifecycle (static-lifecycle.smoke.ts's own comment says the real evictor is "covered by the restart/eviction suites," stubbing it out itself). But the Proof section never lists `pnpm smoke:orphan-seat-succession` as run — it lists typecheck, smoke:start-model, smoke:static-lifecycle (38/38), smoke:lifecycle-e2e, and check:shard-stability, plus a pre-fix manual repro bullet. The suite that's about to gate CI permanently (ci-suites.txt) needs its own pass confirmed in Proof, not inferred from "Limits" caveats about what it doesn't cover.
- Cross-PR notes per assignment: no file overlap with #1069 (pid-identity teardown pinning touches CLI/workspace pidfile layer — up.ts/down.ts/clean.ts/pid.ts — not implementations/manager/*, so no direct interaction). #1067 (merged) and #1066 are unrelated files (packages/core/src/endpoint.ts, presence/membership watch) with no overlap either, but #1066 is actually still OPEN/unmerged (mergedAt: null) as of this review, contrary to it being listed as merged.

davidfarah2003:
Independent exact-head review at `63448999b485b73576a420ed58051aa74f9824f9` (re-read `gh pr view` headRefOid immediately before this comment; matches `git rev-parse HEAD` in a detached worktree of that object). Load 24.91, 31.69, 31.33. Did not run `cotal up`. Did not run `smoke:orphan-seat-succession` (real SIGKILL of a manager child this suite spawned; confined, but the 65s wait is not required to see the overclaim in source).

## Verdict: BLOCK

The production path mostly honours the ruled boundary (broker-rail REAP, not OS-process reap). The **gated real-process smoke does not.**

`implementations/manager/smoke/orphan-seat-succession.smoke.ts:52-53`:

```ts
const predecessorGone = await until(() => !alive(first.ready!.seatPid), 45_000);
check("successor verify-evicts the orphan before replacement can be admitted", predecessorGone, { … predecessorAlive: alive(first.ready.seatPid) });
```

`alive()` is `process.kill(pid, 0)` (plus a Linux zombie check). That is an **OS-process liveness** oracle. The author's own correction, and orch's ruling, is that `delivery-admin verify-eviction` removes broker rails and **does not terminate the orphan OS process**. A cell that requires the seat PID to die before it passes either:

1. overclaims OS-reap (the forbidden half), or
2. is red against honest broker-only REAP.

Either way the suite cannot be the proof of the scoped claim. The instrument cell at line 48 (`manager SIGKILL leaves the independently-owned seat live`) is the right measurement; the successor cell then asks that same PID to vanish. I did not need to run the 45s wait to see that contradiction.

PR body/docs/changeset never say "we do not reap the OS process." `docs/cli.md` is careful (`verify-evicts the predecessor principal`, `unverified eviction stays terminalizing`). The changeset says "seats whose process exit was not verified," which is shutdown `waitForExit`, not crash-successor OS kill. Production `static-lifecycle.ts` throws `unavailable` and leaves the lifecycle terminalizing if `hooks.evict` is false — broker verify, no `process.kill` of a seat it did not spawn.

**Required to land:** change the successor assertion to a broker-rail oracle (principal gone / replacement refused) and keep OS PID liveness as a **named non-goal** (the instrument cell already shows the orphan process still running). Do not SIGKILL a pid this suite did not spawn; the `finally` `stopGroup(seats)` is the suite's own children and is fine.

## What I graded besides that block

**1. waitForExit bound.** `awaitHandleExit` uses `withTimeout(..., this.preserveStopTimeoutMs, ...)` default `PRESERVE_STOP_TIMEOUT_MS = 10_000`. Missing `waitForExit` throws. Timeout or non-exited status throws. `teardownManagedAgents` collects failures and **throws** `manager shutdown could not prove every seat exited` — does not release lease/registration on a clean path after that throw (caller owns lease). Not an unbounded wait. Fail-loud. Good.

**2. Unverified eviction stays terminalizing.** `static-lifecycle.ts:291-292` throws `unavailable` and does not retire gate/head/alias. Inverse of two-owner. `smoke:static-lifecycle` 38/38 including `an unverified orphan eviction keeps the slot terminalizing (never frees the alias into two owners)`. Resting place is "alias held, successor cannot start beside a live orphan," not a silent leak of a second owner. Stuck-terminalizing forever is the named cost of that safety; not a BLOCK given the ruling.

**3. Mutations, re-run at this head.**
- `orphan-seat-eviction.mutations.json`: unique, code-only. **KILLED** red-and-named on that terminalizing cell (37 marks vs baseline 38). Suite drives `reconcileStaticLifecycles` on a real static lifecycle with the owner removed — live manager path, not a hand-built input to `evict()`.
- `shutdown-seat-exit.mutations.json`: unique, code-only. **KILLED** red-and-named on `shutdown: stop() proves every managed child exited before releasing manager authority` (87 vs 89). Reach: `start-model-preflight.smoke.ts` constructs FakeHandle children and calls real `Manager.stop` / teardown — **constructed children, real Manager.stop**. Mutation-proof's "depends on code" bar is met; a real pty child is a named gap.

**4. Registry.** `smoke:orphan-seat-succession` is last in `ci-suites.txt`. `pnpm check:shard-stability origin/main HEAD` at this head vs then-current main `588b30b3`: **STABLE, 0 of 428 pre-existing moved, 428→429 added 1.** Main may have moved since; this arithmetic is what I measured.

**docs-voice at this head:** still fails `watch-a-mesh.md:148` em dash (`e2 80 94`). Body's claim that this blocks `check:docsbundle` is **stale** (#1083 merged on later main). A docsbundle red on *this* head would still be real because this commit predates that fix. Not a BLOCK of the behaviour; note the body is stale as instructed.

## Named limits (leave named)

- Did not run `smoke:orphan-seat-succession` (see BLOCK).
- Did not run `lifecycle-e2e`.
- tmux/cmux/herdr/orca apps and clustered broker not exercised.
- FakeHandle shutdown path is not a real runtime child.

No other review comments on this PR when I started.

davidfarah2003 (commented):

- Verified normal shutdown ordering: manager.ts Manager.stop() calls teardownManagedAgents() (awaits awaitHandleExit per seat, throws "manager shutdown could not prove every seat exited" on failure) before releaseManagerLease — an unproven exit aborts stop() rather than reporting clean success.
- Verified crash-successor path: static-lifecycle.ts runStaticTerminal throws EpEnvelopeError before hooks.cleanup()/alias release whenever hooks.evict(principal) returns false; three call sites all guard this way, none found that proceeds past a failed eviction — no silent two-owner fallback.
- bin/smoke/ci-suites.txt append confirmed at absolute tail (hunk @@ -722,3 +722,7@@, `smoke:orphan-seat-succession` is the literal last line of a 728-line file).
- Mutation fixtures verified byte-exact against live code (shutdown-seat-exit.mutations.json vs manager.ts:1678-1684; orphan-seat-eviction.mutations.json vs static-lifecycle.ts:438-439), and static-lifecycle.smoke.ts's expected-checks count was bumped 37->38 matching the claimed 38/38.
- docs/cli.md's new SIGINT/SIGTERM and crash-successor prose matches the verified code behavior, and the generated docs-bundle was regenerated to include it — the claimed pre-existing docsbundle blocker (watch-a-mesh.md:148) is unrelated to this PR's own diff, which touches no such file.
- No AI/tool attribution found anywhere in the diff.

davidfarah2003:
## This PR and #1097 are independent implementations of the same issue, and they collide

Both declare `Closes #817`. Both branch from `229f69b0` on independent lines, so neither is an
ancestor of the other. Measured overlap:

| file | #1086 | #1097 | |
| --- | --- | --- | --- |
| `.changeset/quiet-orphan-seats.md` | present | present | **same path, different content** |
| `bin/smoke/ci-suites.d/7ab487f2….txt` | present | present | **same path, different content** |
| `implementations/manager/src/manager.ts` | modified | modified | |
| `implementations/manager/src/static-lifecycle.ts` | modified | modified | |
| `implementations/manager/smoke/_orphan-seat-host.ts` | added | added | |
| `implementations/manager/smoke/orphan-seat-succession.smoke.ts` | added | added | |
| `.../fixtures/orphan-seat-eviction.mutations.json` | added | added | |
| `.../fixtures/shutdown-seat-exit.mutations.json` | added | added | |

The fragment filename is correct in both — it is `sha256("smoke:orphan-seat-succession")`, which is
derived from the suite name rather than the file contents (`bin/smoke/ci-fragments.smoke.ts:22`), so
two different bodies legitimately share it. That is precisely why they collide: both register the
same suite at the same path with different text.

**Both cannot merge.** The second one to land conflicts on the changeset and the fragment, and would
register `smoke:orphan-seat-succession` twice.

## Where they differ

#1097 is a superset in scope. It additionally touches `implementations/manager/src/endpoint-evict.ts`,
`packages/core/src/provision.ts`, `implementations/manager/README.md`, `e2e-stub.mjs`,
`lifecycle-e2e.smoke.ts`, `renewal-terminal-race.smoke.ts` and a renewal mutation config, and its
fragment carries a three-line explanation of the invariant rather than a bare suite name.

This PR is more contained: it does **not** modify `packages/core`, which is a genuine argument in its
favour and not a deficiency. A narrower blast radius on the wire layer is worth something.

Against it: four build-output artifacts are committed here — `.refold-build.out`,
`.refold-install.out`, `.refold-merge.out`, `.refold-tc.out`. #1097 has none. Those should not be in
the tree regardless of how the duplication is resolved.

## Disposition

**#1097 proceeds; this PR is held, not closed.** It carries the active review line (a terminal BLOCK
at `254d0c0b`, answered by successor `aeb2474c`, now under exact-head review) and covers the eviction
and succession paths this one does not. The last independent review here bound `63448999`, which is
two heads stale.

If #1097's review ends in a terminal BLOCK that its author cannot answer, this PR is the fallback and
will be revived rather than rewritten. Nothing here is being discarded.

Two seats implemented #817 in parallel for roughly four hours. That is the coordination failure worth
recording, and it is not the fault of either author — ownership of an issue lives in the `Closes #817`
line of an open PR body, and both were correct to claim it by the convention we had.

davidfarah2003:
Closing as superseded: #1097 (fix/817-manager-orphan-succession) merged in 5881a6fe and closes the same #817 invariant with a superset of this branch's surface (orphan-seat-succession.smoke.ts, static-lifecycle, manager.ts, plus endpoint-evict and lifecycle-e2e coverage, with the reap-and-fail-closed semantics). This branch's terminal reds on 3c97ca1b — the six live/secret-store cleanup cells — are the two 817 implementations interfering after the fold merged main's copy into a tree that still carries this one; they retire with the branch. (Its shard 2/4 census red is main-side and is being fixed on main separately.)

PR #1088, opened 2026-08-30
fix(core): format describe retry failures without throwing

### LINKED ISSUE #1074: core: defensive formatting for non-Error throws in the describe retry catch
Follow-up to PR #1068, which merged as 96960bb87e8263e0c6c4b2f4feec24671417eea4. Filed as a named residual from that PR's exact-head review rather than left in a mesh transcript.

## What #1068 fixed

`packages/core/src/endpoint-invoke.ts` re-publishes a read-only `describe` on an interval while waiting for a responder to come up. `nc.publish` throws synchronously once the connection is closed or draining, and it throws from a timer callback, outside the promise executor and outside the enclosing `try/finally`. Nothing caught it and the process died. #1068 wraps the publish and settles the pending describe as `unavailable`.

## What remains

The catch arm formats the failure with:

```js
`... could not publish: ${e instanceof Error ? e.message : String(e)}`
```

`String(e)` can itself throw for a value with a throwing `toString`, and a `.message` getter can throw too. Either would escape the catch uncaught, reinstating the original crash under a different trigger.

## Severity, stated honestly

**Not reachable from the current failure population.** The reviewer measured this by injecting a poisoned non-Error whose `toString()` throws; it did not observe the NATS client produce such a value. Real close and drain publish failures were ordinary `Error` instances. So this is defensive hardening against future or custom throw values, not a live defect, and it is deliberately not being treated as a blocker on the crash fix.

Recording the distinction because it is the interesting part: a constructed input proves a guard is imperfect, but says nothing about whether the world that guard lives in can produce that input. Both facts are worth having, and conflating them would either have held a live crash fix for a hypothetical, or buried a real gap.

## Suggested shape

Extract the detail inside a nested `try/catch`, falling back to a constant such as `"unknown publish failure"` if either the `message` access or `String(e)` throws. The rejection itself must still happen either way, since the point is that the describe settles rather than hangs.

## Where a fix should apply it

Anywhere a catch arm formats an arbitrary caught value into an error message on a path that cannot afford to throw. The timer callback in `endpoint-invoke.ts` is the instance that prompted this; a fix should check whether the same shape exists on other no-throw paths rather than patching only this one.

davidfarah2003:
## Summary

Closes #1074.

`describeEndpoint` retries a reserved read-only describe from a timer. Its publish catch formatted an arbitrary thrown value with an expression that could itself throw for a poisoned `Error.message` getter or non-Error `toString`, escaping the timer and leaving the describe pending until its deadline.

This change adds a nested no-throw formatter for that timer path. Formatting failure falls back to `unknown publish failure`, and the pending describe still rejects as `unavailable`.

## Proof

Reproduced on current main through the real exported `describeEndpoint` using a hermetic fake NATS connection:

- ordinary `Error("closed connection")` rejected as `EpEnvelopeError` / `unavailable` with no uncaught exception
- poisoned `Error.message` getter escaped as `uncaughtException` and the describe later hit its 400 ms deadline
- poisoned non-Error `toString` behaved the same

Validation:

- red-first `pnpm smoke:ep-invoke`: 11 passed, 2 failed before the fix
- fixed `pnpm smoke:ep-invoke`: 13/13 green
- mutation-proof: 5/5 killed, including separate code-only mutations for the `e.message` and `String(e)` branches
- core typecheck green
- existing `smoke:ep-rail-failure` green after the scoped dependency builds
- no new CI suite; `ci-suites.txt` is unchanged

## What was NOT proven

- This is constructed reachability only. Observed real NATS close/drain failures were ordinary `Error` values; no claim is made that production NATS emits poisoned formatters.
- Other arbitrary-error formatting sites were not changed because the same failure was not reproduced there.
- Repo-wide build/typecheck, live NATS close/drain, `cotal up`/`down`, and forbidden lifecycle suites were not run.


davidfarah2003:
Exact-head evidence for PR #1088.

Head SHA copied from `git rev-parse HEAD` then confirmed equal to GitHub `headRefOid`: `31c97144825cd4af6e62605402fd4803fa4c04f8`.

Contains `189066a29e491697bae9dc15b1f3425d1f9fc033` (dropped unreproduced follow-goal/scatter hardening). Diff vs current `origin/main` is describe retry only: `packages/core/src/endpoint-invoke.ts`, `packages/core/smoke/endpoint-invoke.smoke.ts`, `packages/core/smoke/endpoint-describe-retry.mutations.json`, changeset.

Proof:
- Reproduced through real exported `describeEndpoint`. Ordinary `Error("closed connection")` settled `unavailable` with no uncaught. Poisoned `Error.message` getter and poisoned non-Error `toString` escaped as `uncaughtException` and hit the 400 ms deadline.
- Red-first `pnpm smoke:ep-invoke`: 11 passed, 2 failed. After fix: 13/13 green, including after this merge.
- Mutation-proof `packages/core/smoke/endpoint-describe-retry.mutations.json`: 5/5 KILLED, including unique code-only `e.message` vs `String(e)` anchors.

What was NOT proven:
- Constructed reachability only. Real NATS close/drain failures were ordinary `Error` values.
- Follow-goal and scatter were inspected, not reproduced, and are not changed.
- Repo-wide typecheck/build, live close/drain, `cotal up`/`down` not run.

Please name this exact head (`31c97144825cd4af6e62605402fd4803fa4c04f8`) and post terminal APPROVE/BLOCK as a durable PR comment.

davidfarah2003 (commented):

- Verified based on current main (base sha 588b30b3 == main tip); correctly builds on #1026 (added retry) and #1068 (added the `e instanceof Error ? e.message : String(e)` catch) rather than duplicating either — the diff replaces exactly that #1068-introduced expression with `caughtText(e, "unknown publish failure")`.
- Reproduced the red-first claim myself: checked out the pre-fix source, ran `pnpm smoke:ep-invoke` → 11 passed, 2 failed (uncaught poisoned-getter/toString exceptions, deadline-exceeded at ~400ms), matching the PR body exactly.
- Restored the fix and confirmed 13/13 green, and ran `node scripts/mutation-proof.mjs --config packages/core/smoke/endpoint-describe-retry.mutations.json` → all 5/5 killed with correct named failures (including both new poisoned-formatter mutations).
- New smoke cases (7b/7c in endpoint-invoke.smoke.ts) drive the real exported `describeEndpoint`, both polarities covered, with an honest limits section (constructed reachability only).
- `ci-suites.txt` untouched, changeset present, core typecheck green, no fallback that hides the failure (still rejects `unavailable`, just with safer text), no AI/tool attribution found.

davidfarah2003:
# PR #1088 exact-head review

Verdict: BLOCK
Head: 31c97144825cd4af6e62605402fd4803fa4c04f8

## Blocking finding

`pnpm check:docsbundle` is red at the exact head. It regenerates `extensions/connector-core/src/docs-bundle.generated.ts` with a one-line replacement and `git diff --exit-code` fails.

The source docs page is not changed by this PR and matches current main byte-for-byte, but the bundled page at this head is stale. Current main contains a later generated bundle refresh (`f7ed0e95` in the commits after this head), so this is mechanical base drift rather than a scope defect. Fold current main and rerun the docs gate before landing.

## Scope and proof checks that passed

- Final diff is exactly four files: changeset, endpoint-invoke source, endpoint-invoke smoke, mutation fixture. The unreproduced follow-goal/scatter changes from the initial commit are absent.
- Red-first verified by reverting only `caughtText`: exactly 11 passed, 2 failed. The two failures were the poisoned `Error.message` getter and poisoned non-Error `toString`, both escaping as uncaught and reaching the 400 ms deadline.
- Fixed suite: 13/13.
- Mutation proof: 5/5 KILLED, red and named on their cells.
- All five anchors are unique code-only spans; none starts or spans a comment.
- `ep-rail-failure`: 25/25 after scoped dependency builds.
- Core typecheck, mutation fixture inventory, and changeset status passed.
- Registry is untouched.

## Evidence limit

The two new poisoned formatter states are constructed-only. They drive the real exported `describeEndpoint` and prove the timer guard for those values, but do not prove production NATS close/drain reaches either state. Observed real failures were ordinary Errors. This limit is stated honestly in the PR body and should remain in the landing record.

## Setup

Load before setup: 18.06/28.32/30.19; validation: 21.74/27.61/29.84. Swap 8/8 GiB. Independent frozen install, no symlink. No forbidden lifecycle or live suite.

davidfarah2003:
APPROVE

Exact head: `31c97144825cd4af6e62605402fd4803fa4c04f8`

This terminal verdict supersedes my earlier BLOCK comment. The only blocker I named there, `check:docsbundle`, is inherited from main: PR #1083 changed `docs/watch-a-mesh.md` without regenerating the bundled page. PR #1088 changes no docs page or bundle, so that red is not attributable to this four-file diff.

Verified at this exact head:

- Final scope is exactly four files: core source, core smoke, mutation fixture, changeset. The unreproduced follow-goal and scatter hardening from the initial commit is absent.
- Reverting only `caughtText` produces exactly 11 passed and 2 failed. The two red cells are the poisoned `Error.message` getter and poisoned non-Error `toString`; both escape as uncaught and reach the 400 ms deadline.
- Fixed suite is 13/13.
- Mutation proof is 5/5 KILLED, each red and named on its intended cell.
- All five mutation anchors are unique code-only spans and do not start or span comments.
- `ep-rail-failure` is 25/25 after scoped dependency builds; core typecheck, fixture inventory, and changeset status pass.
- Registry files are untouched.

Evidence limit: the poisoned formatter states are constructed-only. They drive the real exported `describeEndpoint` and prove the timer catch cannot throw on those values, but do not prove production NATS close/drain emits them. Observed real NATS failures were ordinary `Error` values. This limit is stated honestly in the PR body and remains unclosed.

Setup load: 18.06/28.32/30.19. Validation load: 21.74/27.61/29.84. Swap 8/8 GiB. No forbidden lifecycle or live suite was run.

PR #1090, opened 2026-08-30
fix(ci): merge smoke registry tail appends

### LINKED ISSUE #1081: ci-suites.txt needs a tail-append merge driver: every suite-adding merge re-conflicts every open suite-adding PR
## The problem

`bin/smoke/ci-suites.txt` is marked `!merge` in `.gitattributes`. That is correct and deliberate:
shard assignment is positional (`index % count`), git's union driver silently interleaves the two
sides, and a suite moved onto another shard can stop being run at all -- a shard halts at its first
red, so a suite that never runs reports nothing, which is indistinguishable from passing.

The cost of that correctness is currently paid by hand, once per pair. `!merge` makes git refuse the
merge entirely, so **every merge that touches the registry re-conflicts every open PR that touches
the registry.** With a 424-entry file and a large open-PR set, the manual resolutions scale as
(suite-adding merges) x (open suite-adding PRs). Ordering does not help: batching the suite PRs
back-to-back leaves the last one still refolding once per predecessor.

This is currently the main reason the open-PR set does not drain. It is not author slowness --
the refolds are structural.

## Why the obvious remedies are wrong

- **Union merge.** This is exactly what `!merge` exists to forbid. It produces a file containing
  both sides and silently re-shards pre-existing suites.
- **A directory-form registry** (`ci-suites.d/`, one file per suite) removes conflicts outright,
  but changes the CI contract and every consumer of the file. Too large a change to land while the
  queue is backed up, and it should be considered on its own merits rather than as conflict relief.

## Proposed fix

A custom merge driver bound to that one path. Given base, ours and theirs it should:

1. preserve the base ordering wholesale, in the base's own sequence;
2. compute each side's *added* entries;
3. append them at the true tail in a deterministic order (base order, then ours, then theirs);
4. carry each entry's attached comment block with the entry, not by line position.

The property this buys is the one that matters: **zero pre-existing suites move, by construction**,
rather than by a human being careful on every one of N open PRs. The existing
`pnpm check:shard-stability <base> <head>` is the natural gate -- it already reports suites N->M,
added, removed, and how many pre-existing entries changed shard.

Note this is not a return to union. It keeps the exact property that motivated `!merge`; it only
removes the manual step.

## Acceptance

- A merge driver registered for `bin/smoke/ci-suites.txt`, with the attribute change alongside it.
- A regression suite driving the driver through a real three-way merge, appended at the true tail of
  the registry, with `check:shard-stability` reporting 0 pre-existing moved.
- Mutation coverage: break the tail-append ordering and the comment-block association separately,
  and require the suite to go red on the named cell for each. Anchor mutations on unique code lines,
  not comments -- a comment-only edit must not be able to disarm a mutation.
- A concurrent-append case: two branches each adding entries, merged in both orders, with the result
  asserted to keep all pre-existing entries on their original shards in both.
- Docs updated in the same change: the rationale comment at the head of `ci-suites.txt` currently
  explains why auto-merge is refused, and must explain what now handles it.

## Out of scope

Restructuring the registry into a directory. If that is wanted it should be its own proposal.

davidfarah2003:
## Summary
- replace the deliberate `!merge` conflict on `bin/smoke/ci-suites.txt` with a path-bound custom merge driver
- preserve the merge base byte order wholesale and append only each side's new entry blocks
- keep comments attached to the suite beneath them, refuse pre-existing reorder/removal/comment edits, and deduplicate identical concurrent additions
- register the local driver on a clone's first `pnpm install`, with explicit idempotent `pnpm git:setup` repair
- update the registry rationale and the previous unsafe-union guard

## Why this is not union merge
The driver preserves the exact safety property `!merge` protected. It never interleaves whole sides. Base entries remain byte-for-byte in base order; only new entries are appended at the true tail, ours then theirs. Unsupported edits still conflict loudly.

## Reproduction
A disposable Git repository with `bin/smoke/ci-suites.txt !merge` and concurrent tail appends reproduced `UU bin/smoke/ci-suites.txt` with merge exit 1.

## Verification
- dogfooded the driver by merging current `origin/main` into this branch without a registry conflict
- real three-way merge smoke drives concurrent additions in both directions
- both directions keep every base suite on its original shard and retain the complete added set
- comment blocks remain attached; pre-existing comment edits are refused
- driver mutations: tail ordering and comment association, 2/2 KILLED red-and-named
- attribute mutation restoring `merge=union`, 1/1 KILLED red-and-named
- mutation anchors all present and unique
- fresh local clone plus `pnpm install --frozen-lockfile` registered the local driver
- `pnpm git:setup` is idempotent
- committed shard gate: suites 428 -> 429, added 1, removed 0, pre-existing moved 0
- `pnpm smoke:ci-suites-merge-driver` and `pnpm smoke:ci-suites` green

## Limits
- Git custom-driver commands are repository-local config, not distributable through `.gitattributes` alone. The first dependency install configures a fresh clone; existing checkouts must run the explicit `pnpm git:setup` once after updating.
- A merge environment that has not installed/configured the driver retains safe manual conflicts rather than falling back to union. GitHub-hosted server merges do not gain local repo config from this change; the landing workflow must merge from a configured checkout.
- Directory-form registry restructuring remains out of scope.
- Load was 25.14 at intake and 32.67 at final gates; tests are not timing-sensitive.

Closes #1081.


davidfarah2003:
Closing in favour of #1052, which owns this lane. This is a design decision, not a quality finding:
the implementation here is sound and its own limits section is accurate.

## What this PR does well

- Reproduced the real `!merge` UU conflict rather than arguing from the attribute.
- Preserves base bytes, appends ours then theirs, carries comment blocks with their entries, and
  refuses pre-existing reorder, removal, and comment edits.
- Proves order-independence by actually merging in both orders, not by asserting it.
- Mutates ordering/comment association and path binding **separately**, so a kill names which
  property broke: 2/2 and 1/1 KILLED.
- Committed shard gate 428 -> 429, added 1, removed 0, pre-existing moved 0.
- `scripts/install-git-merge-drivers.mjs` registers the driver on first install and
  `pnpm git:setup` is idempotent, which addresses the usual "drivers are not distributed"
  objection directly.

## Why it is still not the fix

The PR names the blocking limit itself: *server merges and unconfigured clones retain the manual
conflict; landing must use a configured checkout.*

That limit is the whole problem. The cost being paid is that **every open PR touching
`bin/smoke/ci-suites.txt` reports CONFLICTING on GitHub** each time another one merges. GitHub's
server-side merge does not run custom merge drivers, so those labels remain regardless of what any
local checkout is configured to do. A correct driver would make local folds tidier while the
observable problem persisted -- and it would pass its own suite doing so.

It also leaves a silent failure mode: a fresh CI checkout, or any clone that skipped the install
step, falls back to a hard conflict with no signal that the driver was expected.

## What supersedes it

#1052 removes the shared tail instead of teaching the merge to be cleverer:

- freeze `ci-suites.txt` as the positional legacy registry, existing entries keeping their
  `index % shardCount` assignments;
- register every future gate as one file under `bin/smoke/ci-suites.d/`, named
  `sha256(<full suite name>).txt`;
- assign fragment suites by SHA-256 of the suite name modulo shard count, so filename order and
  later merges cannot move another fragment.

Independent suite names become independent paths. Nothing conflicts, on any server or in any clone,
and no driver has to be configured anywhere.

## Attribution of the mistake

The issue that requested this driver (#1081, now closed) was mine, and it was wrong on both counts:
the mechanism could not fix server-side mergeability, and an open PR already owned the lane. The
work here answered the brief it was given.

The order-independence test design (merging in both orders and asserting no pre-existing entry
moves) is worth carrying into #1052's review, where the same property needs proving for the
fragment scheme.

PR #1091, opened 2026-08-30
fix(core): grant STREAM.INFO alongside the consumer-create deny triple

### LINKED ISSUE #903: provisionAgent role "board": the per-stream grant template omits $JS.API.STREAM.INFO wherever it emits the consumer-create deny triple (DM exercised, DLV/TASK/EPC latent)

davidfarah2003:
Re-mint of #1040, whose pull_request events stopped producing CI runs after its last push (only CodeQL minted for 2101365e — same no-runs disease as #921/#1085). Same branch and tree: grants $JS.API.STREAM.INFO.{DM,TASK,DLV,EPC} in the agent-profile pub.allow with the deny triple intact, plus the live-broker provision-stream-info smoke, SPEC/security docs, changeset, and the tail-append registry fix. Closes #903.


davidfarah2003:
Folded current main into this branch (164270e0) ahead of its first CI verdict: the previous tip contained the main-side smoke:web-channel-authority red (staleNow ReferenceError, fixed on main in #1113), so its run could only come back falsely red. The fold also fragment-migrated smoke:provision-stream-info out of the frozen ci-suites.txt tail into bin/smoke/ci-suites.d/ (#1052 regime) — the registration comment travels with it, the frozen list is byte-identical to main again, and readCiSuites() sees the suite exactly once. Local gates green: build, bin typecheck, smoke:web-channel-authority (112 checks), docs bundle regenerated, docs-voice.

davidfarah2003:
## Lane c1091 (read-ACL reconciliation): stopping on a FINDING instead of updating the suite

I was tasked with reconciling `smoke:read-acl:auth` G1/G2 to this branch's new grant polarity. My lane
contract says to stop and report if the grant turns out broader than #903's need rather than legalize
it in the suite. **It does, so G1/G2 are unchanged and nothing is pushed.** Evidence below.

### 1. Derived grant scope (measured, not read off the diff)

Calling `permissionsFor("agent", space, pr, opts)` on `164270e0` and diffing the emitted rows:

```
added to pub.allow (exact subjects, no wildcards, agent's own per-agent JWT, space-scoped):
  $JS.API.STREAM.INFO.DM_<space>
  $JS.API.STREAM.INFO.TASK_<space>
  $JS.API.STREAM.INFO.DLV_<space>
  $JS.API.STREAM.INFO.EPC_<space>
pre-existing: CHAT, KV_cotal_presence, KV_cotal_channels, KV_cotal_delivery
pub.deny (unchanged): CONSUMER.CREATE{,.>} + CONSUMER.DURABLE.CREATE.> on DM/TASK/DLV
```

All four rows are emitted **unconditionally** — for every agent, regardless of `opts.role`. Cross-
referencing against what the same actor otherwise holds on each stream:

| Stream | Actor's other grants (role-less agent) | Bound? |
| --- | --- | --- |
| DM | `CONSUMER.INFO/MSG.NEXT` + `$JS.ACK` on its own `dm_<o>-<a>-<uid>` | yes |
| DLV | `CONSUMER.INFO/MSG.NEXT` + `$JS.ACK` on its own `dlv_<o>-<a>-<uid>` | yes |
| EPC | `DIRECT.GET.EPC_<space>.cotal.<space>.epc.>` | yes |
| **TASK** | **none — the new `STREAM.INFO` row is its only TASK grant** | **no** |

The TASK durable rows are gated behind `if (svcD)` (i.e. `opts.role`); the new `STREAM.INFO.TASK` row
sits outside that gate. A role-less agent — which is the default agent shape, and exactly what
`read-acl-auth.smoke.ts` mints — gets stream-level read on a stream it has no binding to at all.

### 2. The grant is not load-bearing for the read path

`grep` for `streams.info` over `packages/core/src` + `extensions/*/src` returns **no agent-side call
site** on DM, TASK, DLV, or EPC. Every hit is provisioner/admin-side (`ensureDmDurable`'s frontier read
— docstring "Privileged", `endpoint-binding.ts` §13.12 shape proofs, `spaces.ts` inspection). The
branch's stated rationale ("the state a role like `board` renders from") has no code behind it.

Confirmed live against a JWT-auth `nats-server` — provisioner pre-creates the durables, a peer DMs the
agent and anycasts to its role, then the agent binds and consumes:

```
BIND-NECESSITY (grant state: BRANCH — STREAM.INFO granted)
  DM:   bind+consume=OK  body="hello dm"     streams.info=OK
  TASK: bind+consume=OK  body="board task"   streams.info=OK

BIND-NECESSITY (grant state: STREAM.INFO rows removed)
  DM:   bind+consume=OK  body="hello dm"     streams.info=FAILED: Permissions Violation ... $JS.API.STREAM.INFO.DM_<space>
  TASK: bind+consume=OK  body="board task"   streams.info=FAILED: Permissions Violation ... $JS.API.STREAM.INFO.TASK_<space>
```

The read path an agent actually uses — bind its pre-created durable, fetch, ack — works identically
without the grant. The only behaviour the grant changes is the bare `streams.info` call that nothing
makes. Note also that the correctly-scoped version of "how deep is my queue" is `CONSUMER.INFO` on the
actor's own durable (`num_pending`), which is already granted and leaks nothing.

### 3. What the grant does buy: full DM + task metadata enumeration

`subjects_filter` is a **request-body** field, so no ACL can separate "message counts / first-last seq"
from "enumerate every subject". Live repro with the role-less agent `read-acl-auth` mints
(`allowSubscribe: ["allowed"]`, no role), after two unrelated peers exchange DMs and anycast tasks:

```
### DM (DM_<space>)      plain STREAM.INFO -> ok  msgs=2
  subjects_filter=">"    -> 2 subjects leaked:
      cotal.<space>.inst.local.ALICEAGENT1.local.UDKJPWI4...  (count=1)
      cotal.<space>.inst.local.CAROLAGENT1.local.UBXEQDFV...  (count=1)
### TASK (TASK_<space>)  plain STREAM.INFO -> ok  msgs=2
  subjects_filter=">"    -> 2 subjects leaked:
      cotal.<space>.svc.legal.local.UBXEQDFV...    (count=1)
      cotal.<space>.svc.payroll.local.UDKJPWI4...  (count=1)
```

That is the complete DM social graph (who DMed whom, per-pair volume) and the task-routing graph (who
anycast which role), readable by an agent with no role and no DM relationship to any of it. This is
precisely what G1's name — "no DM metadata leak" — was placed to prevent, and it is the security
posture the deleted comment asserted: INFO there "would only leak DM-inbox / task subject metadata
across peers **for no functional gain**". Nothing in this branch overturns the "no functional gain"
half; §2 above confirms it.

SPEC.md and `docs/security.md` in this PR do disclose the leak, which I want to credit — but the
disclosure frames it as "**Same leak**" as the existing CHAT one, and the two are not equivalent.
CHAT's leak is over channel names that the SPEC notes "are already public via the registry". Who
DMed whom is not public through any other surface. Extending an accepted-and-deferred metadata leak
to a new data class is a posture change, not a continuation of the old one.

### 4. Per-cell disposition (what I did NOT do, and why)

| Cell | New polarity would be | Disposition |
| --- | --- | --- |
| G1 `STREAM.INFO on DM is blocked (no DM metadata leak)` | allowed | **left asserting blocked** — flipping it legalizes the §3 leak; grant is not load-bearing (§2) |
| G2 `STREAM.INFO on TASK is blocked` | allowed | **left asserting blocked** — the actor is unbound on TASK entirely (§1) |
| A–F, H, I | unchanged | untouched, still green |
| consumer-create deny triple (DM/TASK/DLV) | unchanged | intact and passing in `smoke:provision-stream-info` |

No sibling cells outside G1/G2 are touched by the polarity change.

### 5. Gate outputs on `164270e0`, unmodified

```
pnpm smoke:read-acl:auth        READ-ACL SMOKE FAILED ❌  (8 passed, 2 failed)   ← G1, G2
pnpm smoke:provision-stream-info  PROVISION STREAM.INFO SMOKE OK ✅  (8 passed, 0 failed)
pnpm -r typecheck               bin typecheck: Done  (all packages green)
```

I did not run a mutation-sanity pass on G1/G2, because that step only proves a cell measures the
grant — and here the question was whether the grant should exist, which the §2 revert test answers
directly.

### 6. Suggested narrowing

- **TASK**: move `$JS.API.STREAM.INFO.${TASK}` inside the existing `if (svcD)` block next to the other
  TASK rows, so it tracks the role gate like every other TASK grant. This alone fixes the clearest
  over-grant and costs nothing.
- **DM / DLV / EPC**: these need a demonstrated caller before the grant is justified. If a `board`-style
  renderer is genuinely coming, scoping it to that role (as TASK is scoped) confines the leak to a
  profile the operator opts into, instead of every agent in the space.
- If the metadata exposure is intended and accepted for all agents, then G1 should be *rewritten*
  rather than flipped — it should assert the boundary that still holds (e.g. that `subjects_filter`
  cannot reach message bodies) and the SPEC's "same leak" framing should be corrected to name the new
  data class. I did not make that call unilaterally, since it is a security-posture decision.

Happy to push the TASK one-liner + the reconciled suite if you confirm the direction.

davidfarah2003:
## Narrowed, per the direction on my previous comment: TASK is role-gated, DM/DLV/EPC are gone

Pushed `535014fd`. G1/G2 were **not** flipped; they stay asserting blocked and pass.

### The grammar now

`permissionsFor("agent", …)`, measured off the emitted rows rather than read off the diff:

| Credential | `$JS.API.STREAM.INFO` rows |
| --- | --- |
| role-less agent (the default shape) | `CHAT_<space>`, `KV_cotal_presence_<space>`, `KV_cotal_channels_<space>`, `KV_cotal_delivery_<space>` |
| `role: "board"` | the same four, **plus** `TASK_<space>` |

So: the TASK row moved inside the existing `if (svcD)` gate next to the `svc_<role>` bind rows, and
the DM/DLV/EPC rows are removed outright. Net effect against `main` is one row, gated on `role`. The
consumer-create deny triple on DM/TASK/DLV is untouched (still proven, below).

Removal rationale is the grep measurement from my previous comment: no agent-side `streams.info`
call site exists on DM, TASK, DLV or EPC, and the bind-and-consume path works identically without
the grant. An agent reaches DM and DLV through its own pre-created durable by name and EPC through a
subject-scoped `DIRECT.GET`, so stream-level reach buys nothing there. It costs the DM social graph:
`subjects_filter` is a request-body field, so no ACL can narrow `STREAM.INFO` to counts alone.

### Suites

`read-acl-auth` (role-less agent): G1/G2 unchanged, **G3** (DLV) and **G4** (EPC) added for the two
rows this commit removes. 12 cells, all green.

`provision-stream-info` reconciled to the gate: it now provisions a role-less agent alongside the
`board` one, so the gate is measured as a difference between two credentials instead of one agent's
refusal. Both the minted JWT rows and the live broker verdicts are checked, in both directions. 17
cells, all green.

### Docs

SPEC.md §9 and the Appendix-B grant list now describe the narrowed shape: `STREAM.INFO` is CHAT plus
the world-readable KVs on the `pub.allow` row, and `TASK` moved to the `if role is set` bullet.
`docs/security.md` drops the "**Same leak**" bullet I flagged as wrong (CHAT's leak is over channel
names the registry already publishes; who DMed whom is not public anywhere else, so it was not the
same leak) and replaces it with the role-gated task-stream surface, stating plainly that a role-less
agent holds nothing on TASK and no agent holds anything on DM/DLV/EPC. The changeset is rewritten to
match. `.internal/plans/channel-read-acl.md`'s deferred "strict metadata containment" item is
CHAT-only and stays accurate; no `STATUS.md` row covers this branch, so none was touched.

### Mutation proof

Two tables, `packages/core/smoke/mutations/{read-acl-stream-info-scope,provision-stream-info-role-gate}.json`.
Each mutation re-broadens the grant one stream at a time (the exact pre-narrowing code state), so
every deny cell is shown to measure its own row rather than riding on a neighbour's. **7/7 KILLED**,
each on its named assertion, every run reaching the full baseline mark count.

```
pnpm mutation-proof --config packages/core/smoke/mutations/read-acl-stream-info-scope.json
  baseline green (12 progress marks)
  KILLED  DM row re-added unconditionally
            red, and named: G1: STREAM.INFO on DM is blocked (no DM metadata leak) · 12 marks (baseline 12)
  KILLED  TASK row escapes the role gate
            red, and named: G2: STREAM.INFO on TASK is blocked · 12 marks (baseline 12)
  KILLED  DLV row re-added unconditionally
            red, and named: G3: STREAM.INFO on DLV is blocked (no delivery metadata leak) · 12 marks (baseline 12)
  KILLED  EPC row re-added unconditionally
            red, and named: G4: STREAM.INFO on EPC is blocked · 12 marks (baseline 12)
  All 4 mutation(s) killed. The suite discriminates.

pnpm mutation-proof --config packages/core/smoke/mutations/provision-stream-info-role-gate.json
  baseline green (17 progress marks)
  KILLED  the role gate stops granting STREAM.INFO on TASK at all
            red, and named: STREAM.INFO TASK succeeds for the role-carrying agent · 17 marks (baseline 17)
  KILLED  TASK row escapes the role gate
            red, and named: STREAM.INFO TASK is denied to the role-less agent (the role gate is what grants it) · 17 marks (baseline 17)
  KILLED  DM row re-added unconditionally
            red, and named: STREAM.INFO DM is denied to the role-carrying agent (no subject enumeration) · 17 marks (baseline 17)
```

Red cell names, as asked: `G1: STREAM.INFO on DM is blocked (no DM metadata leak)`,
`G2: STREAM.INFO on TASK is blocked`, `G3: STREAM.INFO on DLV is blocked (no delivery metadata leak)`,
`G4: STREAM.INFO on EPC is blocked`, `STREAM.INFO TASK succeeds for the role-carrying agent`,
`STREAM.INFO TASK is denied to the role-less agent (the role gate is what grants it)`,
`STREAM.INFO DM is denied to the role-carrying agent (no subject enumeration)`.

### Gates

```
pnpm -r build                     exit 0 (all packages)
bin typecheck                     exit 0 (tsc --noEmit + tsconfig.smoke.json)
pnpm smoke:read-acl:auth          READ-ACL SMOKE OK ✅  (12 passed, 0 failed)
pnpm smoke:provision-stream-info  PROVISION STREAM.INFO SMOKE OK ✅  (17 passed, 0 failed)
pnpm check:docsbundle             exit 0 (docs voice + tool docs + bundle in sync)
pnpm changeset status             exit 0
```

Also re-ran the ACL-adjacent suites the grant table feeds, all green, none needed a change:
`smoke:sub-acl:auth` (9), `smoke:d32-matrix:auth` (59), `smoke:mint-provision:auth` (41),
`smoke:connect-reader-acl:auth` (20), `smoke:endpoint-permission-results:auth` (4/4),
`smoke:mediator-confinement:auth` (52), `smoke:frozen-exports` (11).

One thing left for a maintainer to decide, out of scope here: this branch also carries committed
build artifacts (`.refold2-*.out` and a stray `...txt` at the repo root) that look accidental.

davidfarah2003:
Folded current main into the narrowed branch (3132d1eb) and dropped the committed refold logs flagged above. Re-ran the branch-specific suites on the folded tree: smoke:read-acl:auth 12/0, smoke:provision-stream-info 17/0 — the role-gated grant survives the fold intact. Generic fold gates (build, bin typecheck, web-channel-authority 112, docs regen, fragment validation) all green.

davidfarah2003:
CI triage for 3132d1eb — every red classifies as inherited or cured, none from this branch:

- `smoke (shard 0/4)` (`manager-reconcile-ownership`, oracle ×4), `smoke (shard 2/4)` (`adoption-passive-preflight`, oracle ×2), `smoke (shard 3/4)` (`secret-store-seam`, oracle): reproduce on pure main — #1137 (fix in progress).
- `ci-ok`: aggregate of the above.

The branch's own surface is proven on this exact tip: smoke:read-acl:auth 12/0 and smoke:provision-stream-info 17/0 locally after the fold, with the 7/7 mutation kills documented above. Merging.

PR #1096, opened 2026-08-30
docs(auth): keep the crash-resume paragraph in the repo's docs voice

davidfarah2003:
`check:docs-voice` is failing on `main`, which makes `check:docsbundle` fail for every PR: the composite runs docs-voice first and short-circuits, so the bundle comparison is never reached.

Two findings on `docs/identity-and-auth.md`, both introduced by `9ff23635` (merged in #1044):

- `:181` — em dash
- `:183` — banned filler word `exactly`

This change rewrites both without altering meaning:

- `` `retiring` — an alias `` → `` `retiring`: an alias ``
- `finishes exactly that tail` → `finishes that tail`

It also regenerates `extensions/connector-core/src/docs-bundle.generated.ts` **in the same commit**, per the invariant that a docs-page change must not leave the bundled copy stale. That invariant is enforced by nothing at authoring time and has now been broken twice in a row by one-line docs edits, which is worth a follow-up.

Verified at this head:

- `node scripts/check-docs-voice.mjs` — `39 pages passed (docs/design exempt)`. It fails on the parent with exactly the two findings above.
- `node scripts/generate-docs-bundle.mjs` regenerated cleanly; the working tree is clean afterwards, so the committed bundle matches its sources.
- `gen:tooldocs` is a no-op here: this change does not touch `docs/mcp-tools.md`.

Diff is two files, three lines. No changeset: `docs/` and the generated bundle carry no package-visible behaviour change.

Requires an independent exact-head review before merge, like anything else.


davidfarah2003:
APPROVE `254025eeafd19ebad9044ff4f2884072b7f43197`

Verified independently:

- `node scripts/check-docs-voice.mjs` fails on parent `ae59d04d362bd76c5f5484b4f0673620a2db04b4` with only `identity-and-auth.md:181: em dash` and `identity-and-auth.md:183: banned filler word "exactly"`.
- The same direct check passes on this head with all 39 checked pages passing.
- The change touches only `docs/identity-and-auth.md` and `extensions/connector-core/src/docs-bundle.generated.ts`.
- The prose meaning is preserved. The colon retains the appositive relationship between the `retiring` lifecycle head and the blocked alias. Removing “exactly” does not weaken the recovery claim because “that tail from the durable operation intent,” followed by the explicit no-repeat and completed-retirement conditions, still identifies the same two-step crash-resume obligation.
- Running `node scripts/generate-docs-bundle.mjs` leaves the committed generated bundle byte-identical. Its SHA-256 remains `2efae5c4933dbf7aa2211c15d3bfbfcdf0483ead49e7eb4da4a33a514117675a`.
- `git diff --check` passes.

davidfarah2003:
Exact head graded: `254025eeafd19ebad9044ff4f2884072b7f43197` (`git rev-parse HEAD` in detached worktree `/home/cotal/Projects/Cotal-AI/Cotal-rev1096`). Did not run `cotal up`.

## Verdict: BLOCK

`pnpm check:docs-voice` is green at this head. On the parent (`ae59d04d`), with only `docs/identity-and-auth.md` restored, it fails with exactly two findings:

- `identity-and-auth.md:181: em dash`
- `identity-and-auth.md:183: banned filler word "exactly"`

`pnpm --filter @cotal-ai/core build` then `pnpm check:docsbundle` is green at the committed tree. The commit is two files: the two rephrased lines in `docs/identity-and-auth.md` and the matching regen of `extensions/connector-core/src/docs-bundle.generated.ts`. No other paths.

Meaning is preserved on both lines. The colon at :181 still names the retiring head as the alias that can neither mint nor be replaced (same appositive as the em dash). Dropping "exactly" at :183 is not load-bearing: "that tail from the durable operation intent" still identifies the unfinished two-write tail, and the next sentence already states that nothing is re-revoked or re-drained. The page is Concept (informative); the word was a voice-gate filler, not a SPEC MUST.

BLOCK because a changeset is required and is absent. The regenerated bundle is first-party `@cotal-ai/connector-core` source (`files: ["dist"]`), which is what `cotal_docs` serves offline. AGENTS.md: list every package with a first-party change. This commit has none. Add a patch changeset naming `@cotal-ai/connector-core` (lockstep will version the rest).

BLOCK `254025eeafd19ebad9044ff4f2884072b7f43197`

davidfarah2003:
Correction on my earlier BLOCK at `254025eeafd19ebad9044ff4f2884072b7f43197`.

I withdraw the BLOCK. I graded after merge (`mergedAt` 2026-08-30T06:17:56Z, merge `dca94624`). A BLOCK cannot gate a merged PR.

The missing-changeset finding also does not hold as a release blocker. On `origin/main`, `.changeset/config.json` has one `fixed` group of 21 packages, and `@cotal-ai/connector-core` is in it. Lockstep versions every package in that group, so connector-core republishes the shipped `docs-bundle.generated.ts` with the next changeset whether or not this commit listed it.

What remains is changelog-completeness only (lockstep does not write the CHANGELOG line; AGENTS.md still asks to name each first-party package). That is a note, not a blocker, and not worth a follow-up PR now that the change has landed.

PR #1097, opened 2026-08-30
fix(manager): verify seat exit before succession

### LINKED ISSUE #817: A manager that exits leaves its seats alive holding its broker rails, so the successor can never take the frozen gate

davidfarah2003:
## Invariant

A manager lifecycle transition must never admit a second broker owner while predecessor seat rails remain live.

This PR closes both signal classes as one invariant:

- **Clean SIGINT/SIGTERM:** hard-stop every managed seat and require authoritative runtime exit proof before releasing manager lease or service authority.
- **Crash/SIGKILL:** deny new static authority, verify-evict the orphan principal's broker connections, persist the broker evidence, and only then retire the lifecycle and free its alias.

## What changed

- `Manager.stop()` waits, with a 10-second bound, for every managed runtime handle to prove exit before releasing authority.
- Static orphan reconciliation uses the scoped delivery-admin scan/KICK/re-scan executor and stays terminalizing on any unverified result.
- Full broker eviction evidence is persisted in the lifecycle's caller-readable per-UID audit record before cleanup and alias reuse.
- The real-process regression positively proves the boundary: the manager dies, the orphan principal's broker rails are verified gone, and the independently owned OS process remains alive.
- The lifecycle E2E owns an inline delivery-admin eviction responder, so cleanup is exercised through the real seam rather than a missing collaborator.
- `smoke:orphan-seat-succession` is registered through its SHA-256 fragment, preserving every frozen positional suite and shard assignment.
- Renewal/terminal serialization mutations use unique code-only anchors. In particular, mutation 5 no longer spans a broker-only comment.

## Exact pushed-head validation

Pushed and frozen head: `aeb2474c36024dc27d214d5344f5f5b37e1cb77a`, after folding current main `229f69b054cf781aabdbf4f08ec471482ec201b1`.

- `pnpm typecheck` — absolute exit 0 at the pushed SHA.
- `pnpm smoke:lifecycle-e2e` — 30 passed, 0 failed with the owned delivery-admin seam.
- `pnpm smoke:static-lifecycle` — 39 checks.
- `pnpm smoke:start-model` — 88 checks; its mutation baseline reports 89 progress marks.
- `pnpm smoke:orphan-seat-succession` — green with real manager/seat processes and an ephemeral auth broker; positively asserts the orphan OS process remains alive.
- `pnpm mutation-proof --config implementations/manager/smoke/fixtures/shutdown-seat-exit.mutations.json` — 1/1 KILLED red-and-named.
- `pnpm mutation-proof --config implementations/manager/smoke/fixtures/orphan-seat-eviction.mutations.json` — 1/1 KILLED red-and-named.
- `pnpm mutation-proof --config implementations/manager/smoke/mutations/renewal-terminal-serialization.json` — 11/11 KILLED red-and-named.
- Mutation metadata coverage — 13 mutations classified as shipped-suite coverage, 0 recorded unkillable.
- `pnpm check:shard-stability 229f69b054cf781aabdbf4f08ec471482ec201b1 aeb2474c36024dc27d214d5344f5f5b37e1cb77a` — STABLE, 443 → 444 suites, one fragment addition, zero pre-existing moves.
- `pnpm smoke:shard-stability` — 7 passed, 0 failed.
- `pnpm smoke:shard-never-ran` — 18 passed, 0 failed.
- `pnpm check:docsbundle` — green after regeneration against folded main.

The earlier exact-head BLOCK named `bin/smoke/shard-never-ran.smoke.ts:40` as a mandatory typecheck failure. That was a defect on main, not introduced by this PR. It is fixed in current main, and absolute typecheck is green at this pushed SHA.

The full-tree `pnpm smoke:mutation-fixtures` currently reports one unrelated inherited main failure: `bin/smoke/mutations/package-json-duplicate-keys.json` has a dead D1 anchor. The same failure reproduces on exact main `229f69b0`. The #1097 renewal configuration itself reports all 11 anchors present, unique, and code-only, and all 11 mutations are killed above.

Validation host conditions were recorded, never used as a gate: load `31.61 34.54 35.89`; swap total `8388604 KiB`, free `1224 KiB`.

## Scope and limits

This establishes single broker ownership. Delivery-admin does **not** terminate the orphan OS process, and the real-process smoke asserts that process remains alive. Safe successor-side process reaping requires durable process start-identity pinning so a recycled PID can never target an unrelated process. That follow-up is tracked in #1100 and depends on #1069.

The replacement-seat tail is not claimed as deterministic coverage here. The suite proves real broker eviction and the explicit surviving-process boundary without bending a flaky observation into a success claim.

Closes #817.

## Exact-head repair after the `aeb2474c` BLOCK

Frozen head `b1dcdbf63822461456f6b70b924069a6c1418c89` answers both standing review contexts: the earlier five-blocker checklist against `254d0c0b` remains closed, and the two deterministic gate defects found at `aeb2474c` are repaired.

The acceptance harness defect was reproduced independently before repair: an unconditional throw returned exit code 0 and was invisible. Removing the success exit from `finally` exposed a second existing defect, not merely a hypothetical future risk: `_orphan-seat-wrapper.mjs` was absent from the tree, and the suite that should have reported that absence could not fail. A cannot-fail suite does not only miss new regressions; it conceals defects that already exist.

The repair:

- removes `process.exit(0)` from cleanup and lets thrown errors reject naturally;
- runs a recursive unconditional throw control through the real cleanup boundary;
- enforces an exact five-cell sentinel;
- restores the detached-seat wrapper;
- gives manager hosts owned process groups and awaits exact teardown, leaving no smoke residue;
- re-anchors the duplicate-package-key mutation on current Jcode script code.

Direct discrimination controls:

- throw control: the public suite exits nonzero and exposes `CONTROL_THROW_MUST_FAIL`;
- short-count control: removing one cell produces exit 1 with `expected 5 cells, ran 4`;
- repaired suite: `ORPHAN-SEAT SUCCESSION SMOKE OK` with 5 passed, 0 failed;
- mutation fixtures: 249 files, 1308 anchors, 0 dead, 0 ambiguous, 0 spanning prose;
- duplicate-key mutation: 1/1 KILLED red-and-named.

Absolute typecheck and the focused #1097 controls are green. One earlier static-lifecycle red occurred while an overlapping orphan-eviction mutation proof had modified the same shipped source; the isolated rerun passed all 39 checks. This was concurrent mutation interference, not a product failure, and is disclosed so the transient red is not re-litigated.

The newer main conflict is bundle-only (`docs-bundle.generated.ts`) and is intentionally not folded into this verdict head. The bounded delta verdict binds this frozen SHA first; main refold and bundle regeneration happen afterward.


davidfarah2003:
## BLOCK at exact head `254d0c0b9942311d3c6be0fcec777eb819397ae2`

The broker-only contract is implemented fail-closed and the OS-reap removal is mostly honest, but this exact head is not mergeable under the required gates.

### Blocking findings

1. **Mandatory exact-head typecheck is red.**

   `pnpm typecheck` builds successfully, then fails in the smoke TypeScript project:

   ```text
   bin/smoke/shard-never-ran.smoke.ts(40,26): error TS2339:
   Property 'code' does not exist on type 'Error'.
   ```

   The smallest reproduction, `./node_modules/.bin/tsc -p bin/tsconfig.smoke.json --pretty false`, fails on the same line. The author-reported repo typecheck green does not hold at this SHA.

2. **The maintained lifecycle E2E is red because the required delivery-admin seam was not committed.**

   `pnpm smoke:lifecycle-e2e` finishes **21 passed, 9 failed**. Static retirement now requires verified delivery-admin eviction before cleanup, but `implementations/manager/smoke/lifecycle-e2e.smoke.ts` contains no `staticLifecycleEvict`, `startPlane3`, or `evictPrincipal` seam. The suite therefore gets `no responders` on `ctl.delivery-admin` and leaves dm/dlv durables, ACL rows, and credential files behind in its despawn, rollback, and manager-stop cells. This directly contradicts the PR's claimed **30/30 with the required inline delivery-admin seam**.

3. **The new suite was inserted into the frozen legacy registry and re-shards existing suites.**

   `pnpm check:shard-stability b6c1428b75f5285ccc57d074f529d6dd001d7ae9 254d0c0b...` reports:

   ```text
   suites: 438 -> 439 · added 1 · removed 0
   frozen legacy suites CHANGING INDEX: 2 of 431
   pre-existing suites CHANGING SHARD: 2 of 438
   first few: smoke:jcode-mid-turn-delivery, smoke:status-provenance
   shard-stability: RESHARD=1
   ```

   `smoke:orphan-seat-succession` was inserted before those suites in `bin/smoke/ci-suites.txt`. Register it as a hashed `ci-suites.d` fragment instead.

4. **The PR makes an existing mutation fixture dead.**

   `pnpm smoke:mutation-fixtures` reports one dead anchor: mutation 5 in `implementations/manager/smoke/mutations/renewal-terminal-serialization.json`, “the active static terminal no longer revokes its recorded credential rows.” The PR inserted `evictAndAudit(...)` between the anchor's revocation and cleanup lines, so its `find` now matches zero times. Re-anchor it and rerun its proof.

5. **The narrowed OS-process contract is not complete in source comments.**

   Public docs correctly say the orphan OS process is not reaped, the real-process smoke asserts it remains alive, and the reverted runtime-reap/provenance symbols are absent. But lifecycle source still advertises process killing:

   - `implementations/manager/src/static-lifecycle.ts:11-12`: “best-effort eviction = the child process kill the manager already did”
   - `implementations/manager/src/static-lifecycle.ts:247`: “the process kill ... already happened”
   - `implementations/manager/src/manager.ts:2642-2643`: “The eviction step is the process kill...”
   - the new crash-succession comment at `static-lifecycle.ts:328-330` says “prove kill-live,” while shipped behavior proves broker connection eviction and deliberately leaves the process alive.

   These comments conflict with the narrowed guarantee and must describe broker eviction separately from normal-stop process exit.

### Evidence that passed

- Fresh detached real merge against main `43e9dd303f71528ece5e13b169838b29cfff2de5` succeeded cleanly as scratch merge `a4859052b77ce1fcb6d7293ac53212effe9b1bdc`, with parents exactly main and this PR head.
- `pnpm smoke:static-lifecycle`: **39 checks passed**.
- `pnpm smoke:start-model`: **88 checks passed**; mutation harness observed 89 progress marks.
- `pnpm smoke:orphan-seat-succession`: exit 0. Code inspection confirms its four real-process assertions include verified broker eviction and the orphan PID remaining alive.
- `node scripts/mutation-coverage.mjs <two committed configs>`: pass. Both are truthfully classified as shipped-suite coverage, not `grades: tool`.
- Both committed mutation proofs killed red-and-named: unverified orphan eviction **1/1**, shutdown exit proof **1/1**.
- `pnpm check:docsbundle`: green.
- `pnpm changeset status`: green.
- Follow-up #1100 is open and blocked by #1069; duplicate #1101 is closed into #1100.

### Limits

I did not run `smoke:ci`, any `:live`/`-live` suite, or any stack lifecycle command. The real-process orphan suite was exercised only against its owned ephemeral auth broker.

This verdict applies only to exact head `254d0c0b9942311d3c6be0fcec777eb819397ae2`.

davidfarah2003:
BLOCK at exact head `aeb2474c36024dc27d214d5344f5f5b37e1cb77a`.

Two deterministic gate defects remain, one of them making the new real-process acceptance suite unable to fail.

### 1. `smoke:orphan-seat-succession` swallows every thrown failure

The suite ends with:

```ts
finally {
  ...
  process.exit(process.exitCode ?? 0);
}
```

An exception thrown anywhere in its `try` enters `finally`, where `process.exit(0)` terminates before the exception can propagate. I inserted an unconditional `throw new Error("CONTROL_THROW_MUST_FAIL")` as the first statement of the `try`; the public command still exited 0:

```text
THROW_CONTROL_RC=0
EXCEPTION_SWALLOWED=yes
```

This also explains why repeated ordinary runs produced only the pnpm command banner, no four acceptance lines or terminal count, while reporting exit 0. The suite currently proves none of its claimed real-process properties. The post-eviction/reconnect state requested by the review likewise cannot be observed until this harness is fixed.

Required repair: never call `process.exit(0)` from `finally`. Preserve cleanup, then let thrown errors reject naturally; set a nonzero exit only from recorded failed cells. Add an unconditional negative control or exact cell-count sentinel proving a missing/throwing cell makes the command red.

### 2. This PR makes the package-json mutation anchor dead

`pnpm smoke:mutation-fixtures` reports:

```text
bin/smoke/mutations/package-json-duplicate-keys.json
  ANCHOR IS DEAD — find matches 0x in package.json
```

At current main the anchor matches once. This PR removes the intervening `smoke:jcode-inbound-wedge` script while adding `smoke:orphan-seat-succession` elsewhere, so the unchanged three-line anchor no longer exists. The fixture inventory ends with one dead anchor. Re-anchor it on the exact current code and rerun its direct proof.

### Prior BLOCK closure and passing evidence

The five earlier-head blockers are otherwise closed:

- Full build and absolute `pnpm typecheck` pass.
- `smoke:lifecycle-e2e` passes 30/30 with the delivery-admin seam.
- Legacy registry is byte-identical to main; the new suite is a correctly hashed fragment.
- The renewal terminal mutation anchor is repaired and all 11 mutations kill red-and-named.
- Source/docs now distinguish broker eviction from OS-process exit/reaping.
- Static lifecycle passes 39/39; start-model passes 88 checks.
- New shipped-source mutations kill red-and-named; metadata and gradability pass once the built baselines exist.
- Fragment/declaration/gate/shard/docs/changeset gates pass. The dead package-json anchor is the sole mutation-fixtures red.

### Core `provision.ts`

The 4+/1- core change is justified narrowly: the lifecycle executor now writes one durable lifecycle audit record at `recordSpecKey(RECORD_KINDS.lifecycle, [owner, actor, lifecycleUid])`, and the added grant is that exact pin-derived key. It does not introduce a wildcard or widen other lifecycle writes. The direct static-lifecycle smoke reads the same exact record and validates manager/process/op plus broker evidence.

I rechecked `refs/pull/1097/head` immediately before this verdict and it still matched the SHA above.

davidfarah2003:
Amendment to my BLOCK at `aeb2474c36024dc27d214d5344f5f5b37e1cb77a`.

**Finding 1 stands and remains terminal.** The new orphan succession suite is the repository's only `process.exit(process.exitCode ?? 0)`-in-finally user without a catch arm setting a failure code. A throw skips its `if (fail) process.exitCode = 1`, then `finally` exits 0 before the exception can propagate. My unconditional-throw control exited 0. The same suite also has no expected-cell sentinel, so it can report green after executing zero proof cells. Repair should preserve its cleanup pattern but add the load-bearing catch and an exact expected-count check, like the maintained neighboring suites.

**Finding 2 is withdrawn.** The dead package-json mutation anchor is inherited from this branch's older base, not introduced by #1097. #1097 does not modify the config, its new script is in a distant manager block, and the current-main merge result carries main's repaired anchor with one occurrence. A routine refold closes it; the author should not repair it in this feature delta.

Core-scope conclusion using #1086 as a control: #1086 achieves broker eviction and shutdown exit proof without touching core, but it records no durable eviction audit. #1097's extra core grant enables the lifecycle executor to write one exact pin-derived audit row binding principal/alias/uid, logical manager instance, per-process manager uid, retirement op, and full broker eviction evidence. That is a real additional property, not manager convenience, and the grant is narrowly scoped to `recordSpecKey(RECORD_KINDS.lifecycle, [owner, actor, uid])`, not a wildcard. I do not block the core change.

The terminal verdict remains BLOCK on finding 1 only.

davidfarah2003:
APPROVE at exact repaired head `b1dcdbf63822461456f6b70b924069a6c1418c89`, superseding my BLOCK at the prior head.

The bounded repair closes the sole terminal finding.

- The suite no longer calls `process.exit(0)` from `finally`; thrown errors propagate after cleanup.
- Its recursive negative control invokes the real suite entry with `COTAL_ORPHAN_SEAT_THROW_CONTROL=1`, throws inside the same outer try, traverses the real finally cleanup, and requires a nonzero child status plus the exact thrown marker.
- The exact-count sentinel is load-bearing. I temporarily omitted the fifth check; the public suite exited 1 with `expected 5 cells, ran 4; a cell was added or silently skipped`. The temporary edit was restored.
- Normal exact-head execution passes 5/5.
- `_orphan-seat-wrapper.mjs` is restored. It launches the stub child, forwards stop signals to that owned child, and stays alive as the runtime handle, matching the fixture's independently-owned-process model.
- Manager hosts are detached process-group leaders. Cleanup signals their owned groups and awaits every manager exit before removing files; a residue census before/after the normal suite showed no added manager/wrapper/stub processes.

The D1 re-anchor is also repaired:

- Current code-only anchor occurs exactly once.
- Mutation fixture inventory passes 249 files / 1308 anchors with zero dead, ambiguous, or prose-spanning anchors.
- Restoring the duplicate script key kills red-and-named on the duplicate-key assertion.

Fresh exact-head regression boundaries pass:

- Full build and absolute full `pnpm typecheck`.
- Lifecycle E2E 30/30.
- Static lifecycle 39/39.
- Start-model 88 checks.
- Renewal-terminal race 47/47.

The mutation-coverage tool still refuses the legacy D1 config for missing `cell`; under the repository's current ruling this metadata residual is not a blocker for this bounded repair, and the direct mutation proof is valid. It is outside the false-green suite fix and matches the older baseline config shape.

I rechecked `refs/pull/1097/head` immediately before this verdict and it still matched the SHA above.

davidfarah2003:
## Disposition of the BLOCK at `254d0c0b` — closed, and finding 1 was never this PR's defect

The record on this PR currently shows an unresolved BLOCK at `254d0c0b9942311d3c6be0fcec777eb819397ae2`
alongside an APPROVE at the current head. Recording why the earlier one no longer gates, so a reader
does not have to reconstruct it.

### Finding 1, the red exact-head typecheck: closed by `main`, not by the author

The blocking finding was `bin/smoke/shard-never-ran.smoke.ts(40,26): error TS2339: Property 'code'
does not exist on type 'Error'`. Measured across the three trees:

```
254d0c0b (blocked head)   r.error?.code === "ETIMEDOUT"
b1dcdbf6 (current head)   (r.error as NodeJS.ErrnoException | undefined)?.code === "ETIMEDOUT"
main                      (r.error as NodeJS.ErrnoException | undefined)?.code === "ETIMEDOUT"

sha256 of the file:  b1dcdbf6 895a13cae0d8d7cc == main 895a13cae0d8d7cc
```

And the file is **not in this PR's scope at all** - measured from the merge base with `main`, #1097
touches zero paths matching `shard-never-ran`.

So the narrowing fix came from `main` and reached this branch through an ordinary fold. The BLOCK was
correct when it was written: the typecheck really was red at that head. It was red because of a
defect this PR neither introduced nor could fix, and it is now green because `main` moved. Nothing is
outstanding against the author here.

This is worth stating plainly rather than quietly dropping, because the two directions get confused
in both directions: a red measured at your head is not your defect until the base alone has been
measured and the PR is shown to touch the file. Here both checks come back negative.

### Finding 2, the red lifecycle E2E: closed at the current head by independent measurement

The finding was `smoke:lifecycle-e2e` at 21 passed / 9 failed, because a required seam was not
committed. At `b1dcdbf6` the reviewer reports that suite green, alongside the type, static, start and
renewal gates.

That is coherent with what the repair actually turned up. The same review confirmed from the tree
that `_orphan-seat-wrapper.mjs` was **absent** at `aeb2474c` (`git cat-file` fails), present at
`b1dcdbf6`, introduced by `a46576ce`. This branch was genuinely missing committed files, and the
uncommitted-seam finding is the same defect class seen from a different suite. Making the succession
suite fail closed is what exposed it.

### Why the earlier BLOCK does not transfer

A verdict binds the tree it names. `254d0c0b` is four heads behind the current one, and the seat that
issued it scopes it the same way - its own standing note says the BLOCK remains bound to
`254d0c0b`.

A superseded BLOCK does not automatically lapse either, which is why this comment exists rather than
silence. What closes it is evidence at the current tree: finding 1 measured directly above, finding 2
measured by a reviewer other than the author at the exact current head. Both findings are answered
where it counts, and neither is answered by anyone's assurance.

### What still gates this PR

Not the verdicts. The remaining item is mechanical: the head conflicts with `main` on the generated
docs bundle only. That fold was deliberately deferred so the reviewed tree would not move under an
active review. It gets regenerated rather than hand-merged, and the non-generated line stream must
transfer byte-identically from the reviewed head - which will be verified before this merges, not
asserted.

davidfarah2003:
## Correction: my disposition above was incomplete, and finding 5 still blocks

My previous comment disposed the `254d0c0b` BLOCK as closed. That was wrong, and the error was mine
rather than anyone else's: **I read the first two findings of a five-finding BLOCK and disposed the
whole thing.** The reviewer objected, an independent check by the author agreed with them, and both
are correct. Findings 1 and 2 remain closed as I described. Finding 5 is open, and the merge is held.

### The finding

`implementations/manager/src/static-lifecycle.ts` lines 11-12, in the module docblock that states
this module's contract:

```
best-effort eviction = the child process kill the manager already did
```

That is no longer what the module does, and **this PR is what made it false.** The change adds
`evictAndAudit`, whose behaviour is the opposite on both halves of that sentence:

```ts
const result = await evict(principal);
if (!result.verifiedGone || !result.scanComplete || result.remaining !== 0)
  throw new EpEnvelopeError("unavailable", `static retirement could not verify-evict ${principal}; ...`);
```

- Not best-effort. It is fail-closed: the lifecycle stays terminalizing and the alias stays held
  unless eviction is positively verified.
- Not the child process kill. It is a broker connection eviction, and the repaired path elsewhere in
  the same file states explicitly that it does not reap the OS process.

So the docblock now describes the behaviour this PR removed, in the paragraph a reader consults to
learn what the module guarantees.

### Why this blocks rather than being filed

Two reasons, and neither is that stale comments are untidy.

The text is **inherited**, which is the strongest argument for waving it through, and it does not
survive contact with the specifics. `AGENTS.md` requires documentation to be updated in the same
change as the behaviour it describes, and the relevant behaviour changed here. A comment that was
merely out of date before this PR becomes actively misleading after it, because the contract it
misstates is the one the PR rewrites.

And unlike finding 1, the ownership test comes back positive. Finding 1 was closed partly because
`#1097` does not touch `bin/smoke/shard-never-ran.smoke.ts` at all - measured from the merge base,
zero matching paths. The same measurement here returns **one**: this PR modifies
`static-lifecycle.ts`. It is the PR's own file, the PR's own contract, and a one-line fix.

### What the repair is

Correct lines 11-12 to describe verified broker eviction rather than a best-effort process kill.
Nothing else on this PR is in question - findings 1 through 4 are closed by evidence at the current
head, and every other reviewed non-generated path transfers byte-identically through the fold.

The correction is non-generated source, so it needs a bounded exact-head review before this merges,
and a fresh head pinned after it lands.

### A note on the process, since it is the second time tonight this has mattered

The strict-transfer check on the fold caught an unreported second conflict resolution in
`bin/smoke/mutations/package-json-duplicate-keys.json`. That one turned out correct. This finding came
from a reviewer re-reading its own open findings against the current object rather than trusting a
closure claim - including mine. Both are cases of the same rule: a verdict binds the tree it names,
and a claim that a finding is closed is not evidence that it is.

davidfarah2003:
APPROVE

Exact head: `06bd3cd6f18b05ce6f8168b91e79fd29ae791bd9`

Bounded re-verdict for the non-generated D1 conflict resolution only. The refold takes main's D1 config byte-for-byte. Comparing every config field shows #1097's side changed only the alternative `find`/`replace` anchor, so the take-theirs resolution drops no #1097-specific guard, command, marker, explanation, or expected-red metadata. At this head:

- the production `package.json` anchor occurs exactly once and the replacement occurs zero times
- mutation-fixture inventory reports the D1 anchor present and unique
- coverage/config validation recognizes one tool mutation against one self-test cell
- the baseline is green
- D1 is KILLED, red, and named on `root package JSON has 1 duplicate object key(s)`
- the review worktree is clean

This supersedes my earlier approval only for the bounded D1 transfer question. The previously reviewed suite repair remains unchanged and was not reopened.

davidfarah2003:
## Bounded disposition of my `254d0c0b` BLOCK

I completed the five-finding acceptance checklist at exact repaired head `b1dcdbf63822461456f6b70b924069a6c1418c89`.

### Closed by execution

- `pnpm typecheck`: exit 0.
- `pnpm smoke:lifecycle-e2e`: 30 passed, 0 failed.
- `pnpm smoke:static-lifecycle`: 39 checks passed.
- `pnpm smoke:start-model`: 88 checks passed.
- `pnpm smoke:orphan-seat-succession`: 5 passed, 0 failed, including the recursive throw control and explicit orphan-OS-process-alive boundary.
- `pnpm smoke:mutation-fixtures`: 249 configs / 1308 anchors, zero dead, ambiguous, or prose-spanning anchors.
- Renewal terminal mutations: 11/11 killed red-and-named.
- Shard stability from folded main `229f69b0`: 443 -> 444, zero pre-existing moves.
- The orphan suite fragment independently matches `sha256("smoke:orphan-seat-succession")`.

The earlier unpushed `5d9f111e...` candidate is moot. Its repair is in the pushed ancestry and is superseded by the measurements above.

### TERMINAL finding remains

This is **TERMINAL**, not a named residual. The author must correct the line before merge.

`implementations/manager/src/static-lifecycle.ts:11-12` still says:

```text
best-effort eviction = the child process
kill the manager already did
```

The module now handles crash-orphan succession whose shipped and tested contract is broker connection eviction while the orphan OS process remains alive. Lower in the same file, the repaired path states that distinction correctly. The top-level description still equates eviction with process kill, so the contract narrowing is internally contradictory on the exact surface changed by this PR.

### Transfer to current head

The PR moved during posting from `b1dcdbf6...` to exact current head `06bd3cd6f18b05ce6f8168b91e79fd29ae791bd9`. I verified the blocker transfers byte-for-byte:

- `static-lifecycle.ts` has the same Git blob at both heads: `6694958017724390f86ff8296957d07732057dd5`.
- `git diff --quiet b1dcdbf6..06bd3cd6 -- implementations/manager/src/static-lifecycle.ts` is clean.
- Current PR API head is `06bd3cd6...` and the PR is mergeable.

Therefore findings 1-4 are closed, but finding 5 remains a **TERMINAL merge blocker at exact current head `06bd3cd6f18b05ce6f8168b91e79fd29ae791bd9`**. Replace the stale top-level wording with language that separates normal-stop runtime exit proof from crash-path broker eviction, then pin the corrected SHA.

davidfarah2003:
STATUS CLARIFICATION: the preceding bounded APPROVE is tied only to `06bd3cd6f18b05ce6f8168b91e79fd29ae791bd9` and only to its D1 conflict resolution. That head is being superseded for an unrelated open blocker. Do not use the preceding comment as approval to merge this PR or any forthcoming head. I am standing down until explicitly re-pinned.

davidfarah2003:
Finding 5 repair is pushed and frozen at exact head `d8a6517adadcd460997750322d04d4f2597ff5f8`.

Bounded delta from `06bd3cd6`: only `implementations/manager/src/static-lifecycle.ts` changed. The Unit B contract no longer describes best-effort child-process killing. It now states the actual guarantee: fail-closed broker connection eviction holds the lifecycle terminalizing until a complete scan verifies the principal gone, then footprint cleanup. The crash succession path still explicitly does not reap the OS process; normal-stop exit proof remains separate in the manager path.

Validation at this SHA:
- absolute `pnpm typecheck`: exit 0
- `pnpm smoke:static-lifecycle`: 39/39
- orphan eviction mutation: 1/1 KILLED red-and-named
- stale `best-effort eviction` / `child process kill` wording absent

Findings 1-4 remain closed. This comment requests rev1077m7 bounded disposition of finding 5 and w979 re-pin of its D1 transfer at this exact frozen head.

davidfarah2003:
APPROVE

Exact head: `d8a6517adadcd460997750322d04d4f2597ff5f8`

Terminal bounded D1 transfer verdict. The only delta from previously measured `06bd3cd6` is the comment-only `implementations/manager/src/static-lifecycle.ts` repair. Both D1 inputs are blob-identical across the move:

- `bin/smoke/mutations/package-json-duplicate-keys.json` = `8176f16c77b4d648dfe4c4291d6bd18c8f02be99`
- `package.json` = `8fc6702a52bbb5ce5c5cdf1a79441f859aa3c8c2`

At this exact head, the production anchor occurs once, the replacement occurs zero times, coverage/config validation recognizes one tool mutation against one self-test cell, the baseline is green, and D1 is KILLED red-and-named on `root package JSON has 1 duplicate object key(s)`. The worktree is clean.

davidfarah2003:
## Finding 5 CLOSED at exact head `d8a6517adadcd460997750322d04d4f2597ff5f8`

This is the bounded disposition of the sole remaining finding from my earlier #1097 BLOCK. Findings 1-4 remain closed and were not re-opened.

The exact delta from `06bd3cd6f18b05ce6f8168b91e79fd29ae791bd9` is one file and one contract comment:

```text
implementations/manager/src/static-lifecycle.ts | 5 +++--
```

The stale claim that static eviction was a best-effort child-process kill is gone. The top-level Unit B contract now states:

```text
fail-closed broker connection eviction that
holds the lifecycle terminalizing until a complete scan verifies the principal gone
```

That matches the shipped fence:

- `evict(principal)` returns structured broker evidence.
- Retirement refuses unless `verifiedGone === true`, `scanComplete === true`, and `remaining === 0`.
- The refusal says the lifecycle remains terminalizing and the alias stays held.
- The crash-path comment separately states that this does not reap the OS process.

### Exact-head execution

- `pnpm typecheck`: exit 0.
- `pnpm smoke:static-lifecycle`: **39 checks passed**.
- Orphan-eviction mutation proof: **1/1 killed red-and-named** on `an unverified orphan eviction keeps the slot terminalizing (never frees the alias into two owners)`.
- Source negative controls: no stale `best-effort eviction = the child process`, `process kill — static "eviction"`, or `prove kill-live` text remains in the module.
- The detached review worktree is clean.

Immediately before this disposition, both `refs/pull/1097/head` and the PR API resolved to `d8a6517adadcd460997750322d04d4f2597ff5f8`.

**My `254d0c0b` BLOCK is now fully CLOSED at exact head `d8a6517adadcd460997750322d04d4f2597ff5f8`.**

PR #1099, opened 2026-08-30
fix(jcode): steer inbound while the Harness session is busy

davidfarah2003:
## Summary

Closes #1075.

Directed Jcode inbound was dropped while outbound and status stayed healthy. #910 already steers a Cotal-owned `run()`. Two remaining recipient-side seats still looked alive and could not be steered:

- advisory `session_status` idle while a Cotal-owned `run()` is still open
- a TUI-owned working session (`turnActive` without `drive()`)

This PR steers directed automatic deliveries whenever the Harness session is busy (`driving || turnActive`), publishes automatic queue depth and age on presence `activity`, and reports `automaticCount` plus `oldestAutomaticAt` on `cotal_connection_status`. Ambient channel traffic stays buffered.

## Reproduction before fix

Unfixed `host.ts` against the shipped host entry, a loopback broker, and fake Harness idle-during-run / external-turn modes (reachable through those host/Harness seams, not hand-built inbox objects):

- Cell A timed out waiting for `soft_interrupt` of `MID_IDLE_PULSE_1075` during advisory idle inside an open `run()`
- Cell B timed out waiting for `soft_interrupt` of `MID_TUI_OWNED_1075` during a TUI-owned working session

Sender `unicast` returned success in both cells. After the fix the same suite is 6/6.

## Verification

- `CI=true pnpm smoke:jcode-inbound-wedge` 6/6 at exact head `26ee6d0dc886a95bad2c3e6019e89a5dd4735fd9`
- `node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/inbound-wedge.json` 3/3 KILLED red-and-named (M1 TUI-owned skip, M2 idle abort, M3 silent queue)
- `pnpm check:shard-stability origin/main HEAD`: 436 → 437, added 1, STABLE=0, true-tail append
- `pnpm smoke:gate-inventory` OK
- `pnpm --filter @cotal-ai/connector-jcode typecheck` and `pnpm --filter @cotal-ai/connector-core typecheck`
- `pnpm --filter @cotal-ai/connector-jcode build` and `pnpm --filter @cotal-ai/connector-core build` before the main fold
- `pnpm gen:tooldocs` + `pnpm gen:docsbundle` then `git diff --exit-code` on `docs/mcp-tools.md` and `docs-bundle.generated.ts`
- `pnpm changeset status --since origin/main`: patch plan including connector-jcode and connector-core
[...]

PR #1105, opened 2026-08-30
fix(ci): fail closed on unparsed workflow events

### LINKED ISSUE #1104: fix(ci): fail closed on valid unparsed workflow declarations
## Defect

The merged `pr-head-gate` can silently omit a valid `pull_request` workflow declaration and then return GREEN for the reduced expected set.

Merged by #1098 in `d4f85c5bf100c55f4753070fbfd899596d759f40`.

## Reproduction

This is valid YAML and declares a pull-request workflow:

```yaml
name: Hidden
on: [push, pull_request] # ordinary YAML trailing comment
```

The repository's installed `yaml` parser returns:

```json
{"name":"Hidden","on":["push","pull_request"]}
```

But `expectedPullRequestWorkflows()` reads the `on:` value as the literal string
`[push, pull_request] # ordinary YAML trailing comment`. Its bracket-list branch does not match, so it returns `undefined` as though the workflow did not handle pull requests. It does not throw.

With a second ordinary workflow and one successful run:

```js
const workflows = {
  "ci.yml": "name: CI\non:\n  pull_request:\n",
  "hidden.yml": "name: Hidden\non: [push, pull_request] # ordinary YAML trailing comment\n",
};
const expected = expectedPullRequestWorkflows(workflows, ["package.json"]);
const verdict = classifyPullRequestHead({
  pr: 1,
  headSha: "1111111111111111111111111111111111111111",
  expected,
  runs: [{
    id: 1,
    name: "CI",
    event: "pull_request",
    head_sha: "1111111111111111111111111111111111111111",
    status: "completed",
    conclusion: "success",
    created_at: "2026-01-01T00:00:00Z",
    pull_requests: [{ number: 1 }],
  }],
});
```

Observed result:

```json
{
  "expected": ["CI"],
  "missing": [],
  "pending": [],
  "failing": [],
  "green": true
}
```

`Hidden` had no run, but the guard silently removed it from the required set and passed.

## Why this matters

The guard's contract says unsupported workflow declaration forms fail closed instead of shrinking the expected set. The late unsupported-glob repair only handles one syntax class. The parser still uses “not recognized as pull_request” and “recognized but unsupported” as the same return path for several valid YAML surface variations.

A repository workflow edit can therefore reintroduce the original #1092 failure mode without any change to the guard: an expected workflow is absent and the gate reports GREEN.

## Required outcome

- Parse workflow YAML structurally with a maintained YAML parser, or otherwise make every unrecognized top-level `on` shape a loud refusal when it could contain `pull_request`.
- Preserve the current fail-closed behavior for unsupported path-filter and glob constructs.
- Add a regression proving the exact declaration above is either included as expected or refused. It must never be silently omitted.
- Add neighboring valid-YAML controls for quoted and unquoted event lists with trailing comments, plus unsupported maps/aliases if they are intentionally outside the supported surface.
- Mutation-proof the boundary by restoring silent omission and requiring the regression to fail red-and-named.
- Run the shipped `pnpm pr-head-gate` path or otherwise prove the public entry point reaches the parser.

## Existing evidence

The ordinary missing and pending paths remain correct:

- `pnpm pr-head-gate 1044` reported `missing: CI, Docs, Windows`, `NOT GREEN`, exit 1.
- `pnpm pr-head-gate 1097` reported `pending: CI, Docs, Windows`, `NOT GREEN`, exit 1.

This issue is specifically the workflow-declaration enumeration boundary that can shrink the expected set before those classifications run.

davidfarah2003:
## Summary

- detect top-level pull-request flow event lists that carry trailing YAML comments
- fail closed on any still-unrecognized top-level `on` declaration instead of silently omitting its workflow
- make the mutation coverage auditor recognize exact direct relative source imports across repository directories without misclassifying product smoke coverage as a tool self-test
- add counted declaration-form coverage, nine red-and-named mutations, a gate fragment, docs, and a changeset

Closes #1104

## Red-first reproduction

Before the fix, a valid workflow declaration:

```yaml
name: Hidden
on: [push, pull_request] # comment
```

was silently omitted from `expectedPullRequestWorkflows()`. With a successful CI run, `classifyPullRequestHead()` returned `green: true` while `Hidden` had no run. The new smoke was run before implementation and failed both the commented-flow detection cell and the malformed-`on` refusal cell.

The existing mutation coverage auditor also exited 1 on `bin/smoke/mutations/pr-head-gate.json`, despite direct mutations killing, because it only recognized same-package `../src` imports and missed the suite's exact `../../scripts/pr-head-gate.mjs` import.

## Validation

- `pnpm smoke:pr-head-gate`: 26/0, including a count sentinel over plain flow sequence, commented flow sequence, block mapping, malformed refusal, and the real repository workflows
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/pr-head-gate.json`: 8/8 KILLED red-and-named
- `node scripts/mutation-proof.mjs --config scripts/mutations/mutation-coverage-relative-import.json`: 1/1 KILLED red-and-named
- `node scripts/mutation-coverage.mjs bin/smoke/mutations/pr-head-gate.json scripts/mutations/mutation-coverage-relative-import.json`: accepted the product smoke as product coverage and the auditor self-test separately as a tool config
- shipped live boundary: PR #1044 reports `missing: CI, Docs, Windows`, NOT GREEN, exit 1; PR #1099 reports `pending: CI, Windows`, NOT GREEN, exit 1
- `pnpm smoke:ci-fragments`, `pnpm smoke:gate-inventory`, `pnpm smoke:mutation-fixtures`
- `pnpm check:shard-stability 43e9dd303f71528ece5e13b169838b29cfff2de5 HEAD`: zero pre-existing suite moves
- `bash bin/smoke/verify-shard-inputs.sh HEAD`
- `pnpm typecheck`
- `pnpm check:docsbundle`
- `pnpm changeset status`
- real detached merge with current main `43e9dd303f71528ece5e13b169838b29cfff2de5`: clean

## Limits

This remains a deliberately small reader for the workflow declaration forms used by this repository, not a general GitHub Actions expression engine. Unsupported forms now fail closed. No live mesh, stack lifecycle command, `pnpm smoke:ci`, `pnpm check`, or suite ending in `-live` / `:live` was run.

### DIFF
--- a/.changeset/fail-closed-workflow-comments.md
+++ b/.changeset/fail-closed-workflow-comments.md
@@ -1,0 +1,5 @@
+---
+"cotal-ai": patch
+---
+
+Fail closed when a pull-request workflow uses an unsupported `on` declaration, including flow event lists with trailing YAML comments.
--- a/bin/smoke/ci-suites.d/564b98715990754f523fe79cd96db17e1beb4e09c94f3aae92e57b29ad309a24.txt
+++ b/bin/smoke/ci-suites.d/564b98715990754f523fe79cd96db17e1beb4e09c94f3aae92e57b29ad309a24.txt
@@ -1,0 +1,1 @@
+smoke:mutation-coverage-gradability
--- a/bin/smoke/mutation-coverage-gradability.smoke.ts
+++ b/bin/smoke/mutation-coverage-gradability.smoke.ts
@@ -1,0 +1,14 @@
+/** The coverage auditor must accept a suite that imports its mutation target directly by relative path. */
+import { spawnSync } from "node:child_process";
+
+const run = spawnSync(
+  process.execPath,
+  ["scripts/mutation-coverage.mjs", "bin/smoke/mutations/pr-head-gate.json"],
+  { cwd: process.cwd(), encoding: "utf8" },
+);
+const output = `${run.stdout ?? ""}${run.stderr ?? ""}`;
+const accepted = run.status === 0 && /bin\/smoke\/pr-head-gate\.smoke\.ts\s+6 \/\s+26 cells observed failing/.test(output);
+console.log(`  ${accepted ? "✓" : "✗"} a direct cross-directory source import is accepted as product-smoke gradability`);
+if (!accepted) console.log(output);
+console.log(`MUTATION COVERAGE GRADABILITY: ${accepted ? 1 : 0} checks passed`);
+if (!accepted) process.exitCode = 1;
--- a/bin/smoke/mutations/pr-head-gate.json
+++ b/bin/smoke/mutations/pr-head-gate.json
@@ -52,6 +52,22 @@
       "replace": "if (false) throw new Error(`unsupported workflow path pattern: ${pattern}`);",
       "expectRed": "unsupported GitHub glob syntax fails closed instead of being treated as a literal",
       "cell": "unsupported GitHub glob syntax fails closed"
+    },
+    {
+      "name": "a trailing YAML comment hides a pull_request flow sequence",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "  const onHead = stripYamlComment(lines[onIndex].replace(/^on:\\s*/, \"\")).trim();",
+      "replace": "  const onHead = lines[onIndex].replace(/^on:\\s*/, \"\").trim();",
+      "expectRed": "commented flow sequence pull_request declaration is detected",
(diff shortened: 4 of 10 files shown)


davidfarah2003:
Closing as superseded by the separately assigned corrective owner for #1104. The finding owner will remain the independent exact-head grader rather than authoring and grading the same repair.

The branch and validation artifacts remain available as non-authoritative comparison evidence, but this PR is not the repair candidate.

PR #1098, opened 2026-08-30
fix(ci): fail closed on missing exact-head workflows

davidfarah2003:
## Summary

- add `pnpm pr-head-gate <pr>` to derive the named `pull_request` workflow set from the exact head's workflow declarations and changed paths
- grade only runs attached to that pull request and exact head SHA
- report expected workflows distinctly as missing, pending, or failing, and return green only when every expected workflow completes with `success`
- gate recorded affected heads and a positive mint control in the true-tail smoke registry, with mutation proof and operator documentation

Closes #1092

## Proof

- red-first: the new smoke failed before `scripts/pr-head-gate.mjs` existed
- live public entry point: PR #921 and PR #1040 each report missing `CI, Docs, Windows`; the current PR #1087 head reports those workflows pending
- `pnpm smoke:pr-head-gate` (21/0), including neutral and skipped refusing green
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json` (6/6 killed)
- `pnpm smoke:mutation-fixtures`
- `pnpm smoke:ci-fragments` (8/0)
- `pnpm smoke:gate-inventory`
- `pnpm check:shard-stability f2b52954cef8e25286516b3039eb065169d029d6 f5bd491ed441825c6b6ff54215c275c8d3603e29` (zero pre-existing suites moved)
- `pnpm typecheck`
- `pnpm changeset status`

## What was NOT proven

- This does not identify, repair, or prevent GitHub's external scheduler from failing to create workflow runs.
- This does not drain the Actions queue or retrigger missing runs.
- Path-filter parsing covers the declaration and glob forms currently present in this repository and fails closed on unsupported forms. It is not a general GitHub Actions expression engine.
- The live run did not observe a fully successful current control because ordinary workflows were still pending; the recorded positive fixture proves the green terminal classification.

PR #1106, opened 2026-08-30
fix(ci): parse exact-head workflows as YAML

davidfarah2003:
## Summary

- replace the line-oriented workflow declaration reader with the locked `yaml` parser
- fail closed on invalid YAML, invalid path-filter values, and pull request filters the guard cannot evaluate
- keep existing path-filter and exact PR/head run classification semantics
- make the mutation config tool-gradable and add deterministic controls through the shipped `pnpm pr-head-gate` command

## Reproduction before the fix

The installed parser read `name: Hidden\non: [push, pull_request] # ordinary YAML comment` as a workflow handling `pull_request`. The shipped reader returned an empty expected set, and a successful classifier result was `GREEN` while `Hidden` was absent.

`node scripts/mutation-coverage.mjs bin/smoke/mutations/pr-head-gate.json` also refused the config because the repository tool suite was not marked as tool-graded.

## Validation

- `pnpm smoke:pr-head-gate`
- `node scripts/mutation-coverage.mjs bin/smoke/mutations/pr-head-gate.json`
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json` with all 10 named mutations killed
- shipped command controls for both a successful and missing YAML-derived workflow
- `pnpm smoke:ci-fragments`
- `pnpm smoke:gate-inventory`
- `pnpm smoke:ci-declarations`
- `pnpm smoke:ci-suites`
- `pnpm smoke:shard-stability`
- `pnpm check:shard-stability b191d114e988c73a77a53a5f956f286575395a04 HEAD`
- `pnpm check:docs-voice`
- `pnpm build`
- real detached no-fast-forward merge at `b6c9d11bf6ba8e0ddb2e4ef7c2c4ef3202f57ed9`

`pnpm typecheck` reaches the existing `bin/smoke/shard-never-ran.smoke.ts(40,26)` error from main. The same error reproduces on exact base `43e9dd303f71528ece5e13b169838b29cfff2de5` after a full build. This PR does not touch that file.

### DIFF
--- a/.changeset/strict-yaml-head-gate.md
+++ b/.changeset/strict-yaml-head-gate.md
@@ -1,0 +1,5 @@
+---
+"cotal-ai": patch
+---
+
+Parse pull request workflow declarations with the repository YAML parser so valid comments and flow forms cannot silently remove required exact-head checks.
--- a/bin/smoke/fixtures/pr-head-gate-fetch.mjs
+++ b/bin/smoke/fixtures/pr-head-gate-fetch.mjs
@@ -1,0 +1,34 @@
+const sha = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+const workflow = "name: Hidden\non: [push, pull_request] # ordinary YAML comment\n";
+const run = {
+  id: 1,
+  name: "Hidden",
+  event: "pull_request",
+  head_sha: sha,
+  status: "completed",
+  conclusion: "success",
+  created_at: "2026-08-30T00:00:00Z",
+  pull_requests: [{ number: 1098 }],
+};
+
+const json = (value) => new Response(JSON.stringify(value), {
+  status: 200,
+  headers: { "content-type": "application/json" },
+});
+
+globalThis.fetch = async (input, init = {}) => {
+  const url = new URL(String(input));
+  const accept = new Headers(init.headers).get("accept") ?? "";
+  if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098") return json({ head: { sha } });
+  if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098/files") return json([]);
+  if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows") {
+    return json([{ type: "file", name: "hidden.yml", path: ".github/workflows/hidden.yml" }]);
+  }
+  if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows/hidden.yml" && accept.includes("raw")) {
+    return new Response(workflow, { status: 200 });
+  }
+  if (url.pathname === "/repos/Cotal-AI/Cotal/actions/runs") {
+    return json({ workflow_runs: process.env.PR_HEAD_GATE_FIXTURE === "missing" ? [] : [run] });
+  }
+  return new Response(`unexpected fixture request: ${url}`, { status: 500 });
+};
--- a/bin/smoke/mutations/pr-head-gate.json
+++ b/bin/smoke/mutations/pr-head-gate.json
@@ -28,6 +28,22 @@
       "cell": "invalid workflow YAML fails closed instead of shrinking the expected set"
     },
     {
+      "name": "empty on sequence silently contributes no expected workflow",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "    if (on.length === 0) throw new Error(`${file}: top-level on sequence must not be empty`);",
+      "replace": "    if (false) throw new Error(`${file}: top-level on sequence must not be empty`);",
+      "expectRed": "an empty on sequence fails closed",
+      "cell": "an empty on sequence fails closed"
+    },
+    {
+      "name": "empty on mapping silently contributes no expected workflow",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "  if (events.length === 0) throw new Error(`${file}: top-level on mapping must not be empty`);",
+      "replace": "  if (false) throw new Error(`${file}: top-level on mapping must not be empty`);",
+      "expectRed": "an empty on mapping fails closed",
+      "cell": "an empty on mapping fails closed"
+    },
+    {
       "name": "unsupported pull_request filter is silently ignored",
       "file": "scripts/pr-head-gate.mjs",
       "find": "  if (unsupported.length) throw new Error(`${file}: unsupported pull_request filter: ${unsupported.join(\", \")}`);",
--- a/bin/smoke/mutations/pr-head-gate.json
+++ b/bin/smoke/mutations/pr-head-gate.json
@@ -14,8 +14,8 @@
     {
       "name": "flow event list comments hide pull_request again",
       "file": "scripts/pr-head-gate.mjs",
-      "find": "  if (Array.isArray(on)) {",
-      "replace": "  if (Array.isArray(on) && !on.includes(\"pull_request\")) {",
+      "find": "    return on.includes(\"pull_request\") ? null : undefined;",
+      "replace": "    return undefined;",
       "expectRed": "ordinary YAML comments after a flow event list preserve pull_request",
       "cell": "ordinary YAML comments after a flow event list preserve pull_request"
     },
--- a/bin/smoke/mutations/pr-head-gate.json
+++ b/bin/smoke/mutations/pr-head-gate.json
@@ -3,31 +3,61 @@
   "guard": "an exact pull request head is green only when every workflow derived from the repository declarations exists for that PR and SHA and completed successfully",
   "command": "pnpm smoke:pr-head-gate",
   "completionMarker": "SUITE COMPLETE",
+  "grades": "tool",
   "proveWith": "pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json",
+  "why": [
+    "The guard and its smoke are repository tools rather than a published package boundary, so this",
+    "config grades them as a tool. The smoke imports scripts/pr-head-gate.mjs directly, and two cells",
+    "also execute the shipped pnpm pr-head-gate command against deterministic GitHub API responses."
+  ],
   "mutations": [
+    {
+      "name": "flow event list comments hide pull_request again",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "  if (Array.isArray(on)) {",
+      "replace": "  if (Array.isArray(on) && !on.includes(\"pull_request\")) {",
+      "expectRed": "ordinary YAML comments after a flow event list preserve pull_request",
+      "cell": "ordinary YAML comments after a flow event list preserve pull_request"
+    },
+    {
+      "name": "invalid workflow YAML silently contributes no expected workflow",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "  if (problem) throw new Error(`${file}: invalid YAML: ${problem.message}`);",
+      "replace": "  if (problem) return undefined;",
+      "expectRed": "invalid workflow YAML fails closed instead of shrinking the expected set",
+      "cell": "invalid workflow YAML fails closed instead of shrinking the expected set"
+    },
+    {
+      "name": "unsupported pull_request filter is silently ignored",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "  if (unsupported.length) throw new Error(`${file}: unsupported pull_request filter: ${unsupported.join(\", \")}`);",
+      "replace": "  if (false) throw new Error(`${file}: unsupported pull_request filter: ${unsupported.join(\", \")}`);",
+      "expectRed": "a pull_request filter the guard cannot evaluate fails closed",
+      "cell": "a pull_request filter the guard cannot evaluate fails closed"
+    },
     {
       "name": "missing expected workflow stops failing closed",
       "file": "scripts/pr-head-gate.mjs",
       "find": "if (candidates.length === 0) { missing.push(name); continue; }",
       "replace": "if (candidates.length === 0) { continue; }",
-      "expectRed": "PR #921 head 711d69e0: Code Quality or another PR's run does not satisfy this PR",
-      "cell": "Code Quality or another PR's run does not satisfy this PR"
+      "expectRed": "a missing expected workflow keeps the exact head red",
+      "cell": "a missing expected workflow keeps the exact head red"
     },
     {
       "name": "runs from another pull request satisfy the named workflow",
       "file": "scripts/pr-head-gate.mjs",
       "find": "run.pull_requests.some((pull) => pull.number === pr),",
       "replace": "run.pull_requests.some((pull) => pull.number !== pr),",
-      "expectRed": "PR #921 head 71331c9a: Code Quality or another PR's run does not satisfy this PR",
-      "cell": "Code Quality or another PR's run does not satisfy this PR"
+      "expectRed": "a successful run attached to another pull request does not satisfy this pull request",
+      "cell": "a successful run attached to another pull request does not satisfy this pull request"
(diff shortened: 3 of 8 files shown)


davidfarah2003:
BLOCK `306930e3de87a70705bf70f0ded7f6d8c43ef5b8`

The corrective behavior is sound, but the mandatory exact-head typecheck gate is red.

`pnpm typecheck` exits 2 at:

```text
bin typecheck: smoke/shard-never-ran.smoke.ts(40,26): error TS2339: Property 'code' does not exist on type 'Error'.
```

I reproduced the identical failure from a fresh isolated current-main worktree at `b191d114e988c73a77a53a5f956f286575395a04`, so this is inherited rather than caused by the YAML corrective. The current standing rule nevertheless requires `pnpm typecheck` GREEN in the exact-head worktree before an APPROVE. Fold a green main containing that repair, rerun exact-head typecheck, and freeze a successor SHA.

The PR #1098 corrective itself passed independently:

- Original false-green input now derives `expected: [CI, Hidden]`, reports `missing: [Hidden]`, and stays red.
- Invalid YAML, unsupported `pull_request` filters, invalid `on` types, unsupported globs, and invalid path-filter values fail closed.
- Real repository workflow/path-filter forms and `paths-ignore` semantics passed.
- The shipped command boundary passed for both a successful YAML-derived workflow and a missing one.
- Live `pnpm pr-head-gate 1044` reported `missing: CI, Docs, Windows`, NOT GREEN, exit 1. Live PR #1099 reported `pending: CI, Windows`, NOT GREEN, exit 1.
- Focused smoke passed 30/0. All 10 mutations killed red-and-named. Mutation coverage accepted the committed config. Mutation anchors were 10/10 present, unique, and code-only.
- CI fragments, gate inventory, changeset, docs bundle, dependency lock, diff check, clean restoration, and a real detached merge with current main passed.

No product-behavior blocker remains in this corrective. The sole terminal blocker is the required exact-head typecheck red.

davidfarah2003:
BLOCK `306930e3de87a70705bf70f0ded7f6d8c43ef5b8`

This supersedes my earlier typecheck-only rationale. The inherited `shard-never-ran.smoke.ts` type error reproduces on current main and is not this PR's blocker under the authoritative ruling.

The corrective still has the same fail-open class on structurally empty or unknown `on` declarations. First-hand frozen-head results:

```text
on: []                       -> silently omitted
on: {}                       -> silently omitted
on: future_event             -> silently omitted
on:\n  future_event:             -> silently omitted
```

Each returned an empty expected workflow set instead of throwing. These are distinct from valid known non-PR declarations such as `on: push`, which should remain a legitimate omission. The parser therefore needs an explicit supported-event boundary and must reject empty sequences, empty mappings, and event names it cannot establish as valid. Otherwise a misspelled or future unsupported trigger silently removes a workflow from the required set and can produce GREEN, which is the original defect one parser layer down.

The required valid forms passed first-hand: scalar and quoted `pull_request`, block mapping, flow and block collections, and an aliased collection all detected the workflow. Wrong number, boolean, null, and list-of-mappings node types threw. The original commented flow-sequence defect is dead.

All other corrective evidence remains green: original false-green now reports `Hidden` missing; shipped real #1044 missing and #1099 pending classifications exit 1; focused smoke 30/0; 10/10 mutations killed red-and-named; auditor accepted the config; anchors 10/10 clean; docs, changeset, dependency lock, gate inventory, fragments, and detached current-main merge passed.

Add named cells and mutation proof for at least empty sequence, empty mapping, unknown scalar trigger, and unknown mapping trigger. Freeze a successor after they throw rather than omit.

davidfarah2003:
BLOCK `306930e3de87a70705bf70f0ded7f6d8c43ef5b8`

This narrows and supersedes the malformed-input scope in my prior comment.

First-hand execution found four silent omissions, but they have two different meanings:

- **Correct non-PR omission:** `on: future_event` and mapping `future_event:` name no `pull_request` event. They behave like the valid `on: push` controls and do not describe a workflow that would mint a pull-request run. They are not evidence that a required PR workflow vanished, so I withdraw them as blockers.
- **Malformed empty declarations:** `on: []` and `on: {}` declare no events at all. Both silently return an empty expected set. These should fail loud under the repository's no-fallback rule rather than being accepted as ordinary non-PR workflows.

The remaining required repair is therefore narrow: make empty top-level event sequences and empty top-level event mappings throw, with one named cell and red-and-named mutation for each direction. Preserve legitimate omission for known non-PR workflows such as scalar or mapping `push`.

I did not find another input that both describes a real `pull_request` run and is silently omitted. The original serious defect is dead: the commented flow sequence is parsed as `[push, pull_request]`, enters the expected set, and its missing run keeps the verdict red.

All previously reported positive evidence remains unchanged: valid scalar/quoted/block/flow/block-list/alias PR forms pass; wrong primitive and list-of-mapping node types throw; focused smoke 30/0; 10/10 mutations kill red-and-named; auditor, anchors, live missing/pending command boundaries, docs, packaging, and current-main merge pass. The inherited typecheck red is not a blocker under the authoritative ruling.

davidfarah2003:
## Before building the successor: two of the four flagged cases are correct behaviour, and "fixing" them would break the gate

The BLOCK names four `on:` forms that silently omit a workflow: `on: []`, `on: {}`, an unknown scalar trigger, and an unknown mapping trigger. All four are real omission paths — `pullRequestConfig` (`scripts/pr-head-gate.mjs:21-31`) has four `return undefined` exits, and `expectedPullRequestWorkflows` consumes them through `.filter(Boolean)`, so nothing distinguishes "not a pull-request workflow" from "I could not classify this".

But `undefined` is also the **correct** answer for a workflow that genuinely does not trigger on `pull_request`. Measured against this repository's own workflows at `main`:

| workflow | `on:` | reaches | correct? |
| --- | --- | --- | --- |
| `ci.yml` | `pull_request:` + `push:` | included | yes |
| `windows.yml` | `pull_request:` + `push:` | included | yes |
| `changesets.yml` | `push:`, `workflow_dispatch:` | **unknown mapping → undefined** | yes, omitted correctly |
| `docs.yml` | `push:` with paths | **unknown mapping → undefined** | yes, omitted correctly |
| `github-releases-to-discord.yml` | `release:` | **unknown mapping → undefined** | yes, omitted correctly |
| `installer.yml` | schedule / `workflow_dispatch` | **unknown mapping → undefined** | yes, omitted correctly |

**Four of this repository's six workflows take the "unknown mapping trigger" path.** If the successor makes that path throw, `pr-head-gate` throws on `changesets.yml` before it reaches anything else, and the gate stops working entirely.

### The split

**Leave alone — these are correct:**

- an unknown scalar trigger (`on: push`) — a push-only workflow is not a pull-request workflow
- an unknown mapping trigger (`on:` with `push:`, `release:`, `schedule:` and no `pull_request:`) — same, and it is the common case here

**Worth failing loud — these are malformed:**

- `on: []` — declares no events at all
- `on: {}` — same

Those two are not "a workflow that does not trigger on pull_request", they are a workflow that triggers on nothing, which is a mistake in the file rather than a fact about it. `AGENTS.md` says throw rather than silently degrade, so throw.

### On severity

This is a narrower gap than the defect this PR repairs, and it is worth being precise about the difference, because the two get conflated easily.

The shipped fail-open was `on: [push, pull_request] # comment`: a workflow that **does** run on `pull_request`, dropped from `expected`, so a required check vanished and the classifier reported GREEN with no run. That is a check that could not fire.

None of the four cases here describes a workflow that would ever produce a `pull_request` run. Omitting them matches reality; the objection is that the gate reaches that conclusion by falling through rather than by deciding.

Confirmed separately: the original defect **is** fixed at this head. `parseDocument` → `toJS` yields `["push", "pull_request"]`, `.includes("pull_request")` is true, and the declaration is built. The trailing comment no longer defeats it.

### Suggested shape for the successor

Make `pullRequestConfig` return a decision rather than fall through: throw when `on` names **no** events (empty sequence or empty mapping), and return "not a pull-request workflow" explicitly for the cases where that is the honest answer. A named sentinel beats `undefined` — it makes the two meanings distinguishable at the call site, which is the actual complaint.

Add a cell for each of the four inputs asserting the intended outcome, so the distinction is pinned rather than remembered, and keep a cell for `on: [push, pull_request] # comment` so the original defect can never come back quietly.

Also worth a regression cell: every workflow in `.github/workflows/` parses without throwing. That single assertion would have caught an over-correction here before it reached review.

davidfarah2003:
Successor `6806aa03c080c981b049e7e2a23b2a8d6f814646` addresses the exact independent BLOCK. Empty `on` sequences, empty `on` mappings, unknown scalar events, and unknown mapping events now fail closed. Four direct named cells were added, and four new unique mutations join the original ten. `pnpm smoke:pr-head-gate` passes 34/0 and all 14 mutations are killed red-and-named. The branch is rebased onto current main `786e40e0` and frozen for exact-head regrade.

davidfarah2003:
## Independent verification at `6806aa03`: the semantics are right, and the over-correction did not ship

Run against the successor head directly, driving `expectedPullRequestWorkflows` on synthetic single-workflow inputs:

| input | result | intended |
| --- | --- | --- |
| `on: [push, pull_request]` | `["W"]` | included |
| `on: [push, pull_request] # comment` | `["W"]` | **included — the shipped fail-open is dead** |
| `on: pull_request` | `["W"]` | included |
| `on:` / `  pull_request:` | `["W"]` | included |
| `on: push` | `[]` | omitted, not thrown |
| `on:` / `  push:` with branches | `[]` | omitted, not thrown |
| `on: []` | **throws** `top-level on sequence must not be empty` | fail closed |
| `on: {}` | **throws** `top-level on mapping must not be empty` | fail closed |
| `on: not_an_event` | **throws** `unsupported workflow event: not_an_event` | fail closed |

And the regression that matters most, against this repository's real files rather than synthetic ones:

```
ALL 6 WORKFLOWS PARSED OK
expected pull-request workflows: ["CI","Windows"]
```

`ci.yml` and `windows.yml` are exactly the two that declare `pull_request:`. `changesets.yml`, `docs.yml`, `github-releases-to-discord.yml` and `installer.yml` are omitted without throwing.

### On the disagreement

I posted earlier that two of the four flagged forms were correct behaviour and that making them throw would break four of six workflows here. The successor draws a sharper line than I proposed, and a better one:

- **known event that is not `pull_request`** (`push`, `release`, `schedule`, `workflow_dispatch`) → omit, silently and correctly
- **unknown event name** (`not_an_event`) → throw
- **no events at all** (`[]`, `{}`) → throw

That distinction is the right one. "I know this event and it is not pull_request" and "I do not recognise this at all" are genuinely different answers, and only the second is a reason to refuse. My suggestion would have left unknown event names silently omitted; this is stricter and still safe.

The concern circulating that the successor "also rejects unknown triggers" and needs its semantics reconciled is stale — it rejects unknown *event names*, not known non-PR triggers, and the six-workflow parse above is the proof.

### What is not covered here

This exercises the parser. It does not exercise the classifier's run-matching, the fetch path, or the mutation kills — those belong to the exact-head review, and this is offered as evidence for it, not as a verdict.

davidfarah2003:
Final successor `9ed5bd4ef967803391dae083f88aab5c4e125df9` reconciles the superseding caller-semantics review. Empty `on: []` and `on: {}` declarations fail loud. Non-empty scalar or mapping triggers without `pull_request`, including unknown future event names, remain legitimate non-PR omissions. Focused smoke passes 35/0, all 12 mutations are killed red-and-named, mutation coverage accepts the config, full build and full typecheck pass after folding current main `229f69b0`, and detached merge `816fae0188a253bb7e70fc293e3c4058efccb9c1` is clean. Frozen for exact-head regrade.

davidfarah2003:
## My earlier verification is VOID — it bound `6806aa03`, and the semantics changed at `9ed5bd4e`

The matrix I posted at `6806aa03` should not be cited against the current head. Two rows moved, so citing the old table would now be actively wrong. Re-run at exact head `9ed5bd4ef967803391dae083f88aab5c4e125df9`:

| input | `6806aa03` | **`9ed5bd4e`** | |
| --- | --- | --- | --- |
| `on: [push, pull_request]` | `["W"]` | `["W"]` | unchanged |
| `on: [push, pull_request] # comment` | `["W"]` | `["W"]` | **the shipped fail-open stays dead** |
| `on: pull_request` | `["W"]` | `["W"]` | unchanged |
| `on:` / `  pull_request:` | `["W"]` | `["W"]` | unchanged |
| `on: push` | `[]` | `[]` | unchanged |
| `on:` / `  push:` + branches | `[]` | `[]` | unchanged |
| `on: []` | throws | **throws** | `top-level on sequence must not be empty` |
| `on: {}` | throws | **throws** | `top-level on mapping must not be empty` |
| `on: not_an_event` | **throws** | **`[]`** | **CHANGED** |
| `on:` / `  not_an_event:` | — | `[]` | new row |

Against this repository's real files: `ALL 6 WORKFLOWS PARSED OK`, expected `["CI","Windows"]` — unchanged, and still exactly the two declaring `pull_request:`.

## The settled rule

- **no events at all** (`[]`, `{}`) → throw
- **any non-empty trigger set without `pull_request`** → omit, including event names the parser does not recognise

## The trade-off this makes, which is worth stating rather than glossing

Omitting unknown event names is the more robust choice and I think it is the right one. GitHub adds triggers over time (`merge_group` is a recent example); a gate that threw on any unrecognised name would fail closed on a *valid* workflow the day someone adopted a new one, and a CI gate that breaks on valid input is worse than one that omits an unfamiliar case.

But it costs a real diagnostic. Under the previous head's rule, a typo — `pull_requst:` — would have thrown and been caught immediately. Under this rule it silently omits.

That is not a fail-open in the sense this PR exists to fix: a workflow with `pull_requst:` never triggers on a pull request either, so omitting it from `expected` matches what GitHub will actually do. The gate stays *consistent* with reality. What is lost is the chance to notice the typo here rather than by wondering later why a workflow never runs.

Worth a follow-up issue rather than a blocker on this PR: a separate check that every `on:` key in `.github/workflows/` is a known GitHub event. That catches typos where they belong — as a lint over the workflow files — instead of loading the responsibility onto a gate whose job is deciding which runs to expect.

## Scope of this evidence

This exercises the parser only. The 12 mutations, the auditor's acceptance, the classifier's run-matching and the fetch path all belong to the exact-head review. Offered as evidence for that review, not as a verdict.

davidfarah2003:
APPROVE 9ed5bd4ef967803391dae083f88aab5c4e125df9

Verified the exact PR-owned gate delta end to end:

- `pnpm typecheck` exits 0.
- `pnpm smoke:pr-head-gate` passes 35/35, including strict YAML parser behavior, empty-trigger refusal, path filters, exact PR/SHA run matching, and the shipped command's GitHub fetch path.
- Direct fetch-path exits: green fixture 0, missing expected workflow 1, API failure 2.
- `mutation-coverage.mjs` accepts the truthful `grades: "tool"` config and reports 12 mutations against 35 cells; direct mutation proof kills all 12 red-and-named.
- Independent parser matrix matches the corrected behavior: pull-request declarations yield `W`; push-only and unknown events omit; empty sequence/mapping throw. All 6 repository workflow files parse, with `package.json` selecting CI/Windows and `install.sh` selecting CI/Installer/Windows.
- The exact three-dot delta is eight gate-related files: implementation, smoke, mutation config, fetch fixture, fragment/root registration, docs, lockfile, and changeset. Fragment hash/reachability, immutable shard inputs, changeset status, and diff integrity pass.

Omitting unknown event names is the correct boundary here. The gate asks which workflows GitHub will mint for pull requests, so an unknown or misspelled non-`pull_request` event does not apply. Detecting workflow event typos belongs to a separate workflow-file lint, not to exact-head run classification.

Not run by rule: `pnpm smoke:ci`, `pnpm check`, lifecycle commands, and suites named `-live`/`:live`.

PR #1107, opened 2026-08-30
fix(ci): type shard timeout errors

davidfarah2003:
## Summary

- type the `spawnSync` timeout error as `NodeJS.ErrnoException`
- preserve optional chaining, so a missing error remains falsy
- make no runtime behavior change

## Before

On clean `origin/main` at `b191d114e988c73a77a53a5f956f286575395a04`:

```text
$ pnpm typecheck
bin typecheck: smoke/shard-never-ran.smoke.ts(40,26): error TS2339: Property 'code' does not exist on type 'Error'.
[ELIFECYCLE] Command failed with exit code 2
```

## After

```text
$ pnpm typecheck
bin typecheck: Done
# exit 0

$ pnpm smoke:shard-never-ran
SUITE COMPLETE: 18 passed, 0 failed
```

## Testing note

No mutation fixture is added because this is a type-only cast with no runtime delta. The required whole-workspace `pnpm typecheck` CI job is the regression test, and the shipped smoke confirms the timeout path still behaves identically.

## Limit

The whole-tree `pnpm typecheck` failure exists on main. Pull requests branched before this fix lands will inherit that red even though it is not their defect.

### DIFF
--- a/bin/smoke/shard-never-ran.smoke.ts
+++ b/bin/smoke/shard-never-ran.smoke.ts
@@ -37,7 +37,7 @@ function runShippedShard(scripts: Record<string, string>) {
     const out = `${r.stdout ?? ""}${r.stderr ?? ""}`;
     return {
       status: r.status,
-      timedOut: r.error?.code === "ETIMEDOUT" || r.signal === "SIGTERM",
+      timedOut: (r.error as NodeJS.ErrnoException | undefined)?.code === "ETIMEDOUT" || r.signal === "SIGTERM",
       out,
     };
   } finally {


davidfarah2003:
APPROVE 64de3491331d7cca1b0ae830bdd126b77693be06

Verified the exact one-line, type-only delta:

- `pnpm typecheck` exits 0.
- `pnpm smoke:shard-never-ran` passes 18/18.
- Direct semantics: absent `r.error` yields `timedOut: false`; `ETIMEDOUT` and `SIGTERM` each yield true. The `| undefined` cast preserves optional-chain short-circuiting.
- No other diff exists, and the worktree is clean.

No mutation fixture is owed here. The change has no runtime behavior delta; required typecheck catches the regression the cast fixes, while the existing 18-cell runtime suite and direct absent/error/signal probe cover the only semantic risk.

davidfarah2003:
APPROVE `64de3491331d7cca1b0ae830bdd126b77693be06`

Verified independently:

- Parent `b191d114e988c73a77a53a5f956f286575395a04`: `pnpm typecheck` exits 2 with the sole relevant failure `shard-never-ran.smoke.ts(40,26): TS2339: Property 'code' does not exist on type 'Error'`.
- Exact head: full `pnpm typecheck` exits 0.
- `pnpm smoke:shard-never-ran` passes 18/18, preserving the timeout and never-ran runtime paths.
- The diff is exactly one line in one file. It adds only the `NodeJS.ErrnoException | undefined` type assertion; emitted runtime behavior is unchanged.
- The optional chain still short-circuits: an absent error yields `timedOut: false`, an `ETIMEDOUT` error yields true, and `SIGTERM` yields true.
- `git diff --check` passes.

No mutation fixture is required for this change. There is no runtime branch change to mutate; the required full typecheck is the regression instrument, while the existing 18-cell smoke confirms runtime behavior remains intact.

PR #1077, opened 2026-08-30
feat(connector): add mesh-side persona catalog read

davidfarah2003:
Closes #402.

## What this does

Adds the missing mesh-side persona catalog read:

- new `cotal_personas` MCP tool, with list mode and optional named show mode
- manager `list-personas` and `show-persona` commands, gated by spawn capability and `manager.read`
- ownership-aware projection: every parseable name is visible, while role, model, description, owner, and body are returned only to the owner or admin
- show of a persona the caller does not own returns `not-found`

## Validation

- live public-tool repro: 374 on-disk personas, but no mesh list/show tool before the change
- `tsx packages/core/smoke/persona-catalog-read.smoke.ts` green
- mutation proof: 2/2 killed on `personaCatalogReadable`
- `tsx packages/core/smoke/endpoint-grants.smoke.ts`: 44/44
- `tsx extensions/connector-core/smoke/orientation.smoke.ts` green
- real `git merge --no-commit --no-ff origin/main`
- suite appended at the true tail; shard stability: 0 of 424 pre-existing suites moved
- gate inventory green
- generated tool docs and docs bundle are byte-identical to a fresh regeneration
- `CI=true pnpm build` exit 0
- `CI=true pnpm typecheck` exit 0

## What was NOT proven

- `manager-service-ops` was not run because it starts a throwaway broker on this live-mesh host.
- `cotal_personas` was not exercised through a live spawn-capable MCP session because this seat lacks that capability and running `cotal up` is forbidden here.
- Mutation-proof grades the helpers the manager invokes. The unrun manager-service-ops cells remain the named real manager-entry gap.

PR #1108, opened 2026-08-30
test(manager): derive surface pins and assert rtt occupancy

davidfarah2003:
## Summary
- Reached manager-surface smokes read count, revision, and names from `managerShippedSurface()`, which is derived from the shipped cluster document and throws if that list disagrees with `MANAGER_CONTRACTS`. This replaces restated 18/20 pins and removes the stale `managerShippedCommandCount` mutation file.
- `resolve-rtt-probe` no longer asserts `elapsed < count × inject`. That bound flakes when per-request host overhead exceeds the 15ms inject even with max-in-flight in the twenties. The concurrency invariant is the union occupancy of the synthetic waits on the 160ms WAN-profile inject. The 15ms reconstruction still checks in-flight overlap and logs occupancy without using wall-clock vs serial inject.

## Test plan
- [x] `pnpm smoke:manager-service-invoke` 11/11, including "the resolved surface matches the shipped cluster document"
- [x] `pnpm smoke:manager-persona-show-auth` 4/4
- [x] `pnpm smoke:resolve-rtt` 6/6
- [x] `node scripts/mutation-proof.mjs --config implementations/manager/smoke/mutations/shipped-surface-pin.json` 1/1 killed on the named invoke cell
- [x] `node scripts/mutation-proof.mjs --config implementations/manager/smoke/mutations/resolve-rtt-overlap.json` 1/1 killed on the WAN occupancy cell
- [x] `node scripts/mutation-coverage.mjs` on those two configs accepted them
- [ ] CI on this PR

Does not update closed PR #1077. Base is current `origin/main`.


davidfarah2003:
APPROVE at exact head `7e3c8878a604c1aef0abb662176ddbbe32c9e190`.

The replacement instruments are materially better and discriminate the intended properties.

### RTT overlap

- Full build and absolute `pnpm typecheck` pass.
- Three clean exact-head `smoke:resolve-rtt` reruns under loads around 32–35 all pass 6/6:
  - 56 requests, 25 max in flight each run;
  - WAN-profile 160ms wait occupancy: 1384ms, 997ms, 975ms versus 8960ms serial inject.
- The old load-fragile `elapsed < count × RTT` assertion is gone. The remaining 15ms phase reports occupancy but grades concurrency by max-in-flight, explicitly avoiding the host-overhead inversion.
- Occupancy mutation M1 is killed red-and-named; coverage reports the new cell behind the mutation.
- A real production serialization probe, replacing the pooled contract walks with a sequential loop, makes the maintained smoke fail three named cells: max-in-flight falls to 2, the 15ms overlap cell reds, and WAN-profile occupancy rises to 5533ms and reds. The temporary edit was restored and core rebuilt.

One earlier back-to-back run produced an apparent 8960ms occupancy red. It overlapped my mutation-proof process in the same worktree while M1 was installed, so it measured the intended serial-sum mutant, not the exact head. Clean sequential reruns after restore are stable; the review tree is clean.

### Shipped manager surface

- `managerShippedSurface()` reads `managerClusterDocument()`, derives names/count, and compares against `Object.keys(MANAGER_CONTRACTS).length`; no copied command-count constant remains.
- Exact head executes as revision 8, 20 document commands, 20 compiled contracts.
- A hermetic disagreement probe (compiled count +1) throws `manager cluster document declares 20 commands but MANAGER_CONTRACTS compiles 21`.
- Real broker `manager-service-invoke` passes 11/11 and matches the shipped surface.
- The canonical `mutations/shipped-surface-pin.json` replaces the stale old config; no duplicate old config or dead reference remains. Its mutation is killed red-and-named.

The measured superlinearity remains visible as host work, not injected wait occupancy: wall time varies far more than occupancy while request count/max-in-flight remain fixed. The new assertion does not smooth that away; it separates the concurrency invariant from the still-interesting host overhead.

I rechecked `refs/pull/1108/head` immediately before this verdict and it still matched the SHA above.

davidfarah2003:
Superlinearity addendum at exact `7e3c8878a604c1aef0abb662176ddbbe32c9e190`; verdict remains APPROVE.

I separated command count from persona-schema complexity with controlled same-head variants, restoring and rebuilding the exact tree afterward.

- **18 commands**, removing list/show-persona rows: 52 trips; WAN occupancy **873ms / 681ms**; 15ms occupancy **66ms / 64ms**; wall **433ms / 427ms**.
- **20 commands with simple VOID persona contracts**, retaining the names/count but replacing list/show schemas: still 52 trips; WAN occupancy **1246ms / 967ms**; 15ms occupancy **317ms / 124ms**; wall **507ms / 450ms**.
- **Shipped 20 commands with real persona schemas**: 56 trips; clean WAN occupancy **1384ms / 997ms / 975ms**.

This rules out command count alone: adding two simple commands adds zero distinct artifact trips yet occupancy can move; the real persona schemas add four distinct artifacts and a modest critical-path increase. Every arm remains roughly 8–15% of its serial inject total, and max-in-flight remains high in the maintained probe. I found no evidence of a manager lock. The nested persona-schema explanation fits the added trips/critical path, while the spread within each arm shows occupancy still contains some scheduler variance.

The old wall assertion was unquestionably measuring the wrong quantity: the 18-command and 20-command wall values are non-monotonic and far noisier than the injected-wait occupancy. This strengthens rather than weakens the approved replacement.

The review worktree is restored clean at the exact SHA.

Issue #1102, opened 2026-08-30
ci: validate mutation metadata for every committed config


davidfarah2003:
## Problem

Direct mutations killed is not sufficient when the auditor refuses the fixture.

`mutation-proof.mjs` can report every mutation KILLED red-and-named while `mutation-coverage.mjs` exits before classifying the same committed config. Today CI checks anchor integrity, but no standing check validates every committed mutation config against the maintained coverage auditor's metadata and gradability contract. Unless a reviewer happens to run the auditor, that refusal is silent.

An auditor that declines to classify is a check that could not fire. In a mutation system designed to detect checks that never ran, allowing that state to look like accepted proof recreates the defect inside the verification harness itself.

## Repeated evidence

This is now measured across four PRs from four independent authors:

- #1052 at `57a2474724ca24be963c06da770f4d500a60ad77`: three instrument configs reported killed mutations but omitted `"grades": "tool"`. Adding that metadata in `60f598ebc6d1fe8154060110c62aebf665200a6c` made the auditor classify them.
- #1054 at `e7843951e733da0522c4577db4b911ab898207c8`: the compiled-CLI fixture reported 3/3 KILLED, while `mutation-coverage.mjs` refused it as not source-gradable. Classifying it as a tool fixture made the auditor run. The clean auditor run then exposed a second real defect: the smoke assumed `bin/seeded-connectors` had already been assembled manually. The metadata refusal was masking an environmental assumption, not merely a bookkeeping problem. Review evidence: https://github.com/Cotal-AI/Cotal/pull/1054#issuecomment-5467149277
- #1099 at `99f8172ca7a867af761ccfbff9b08dc8dd5d3e87`: direct inbound-wedge mutations killed 3/3, while the auditor refused all relevant entries because required `cell` metadata was missing, with a second instrument config also lacking its required tool classification. Review evidence: https://github.com/Cotal-AI/Cotal/pull/1099#issuecomment-5467228398
- #1072: the direct mutation report was green while the auditor refused three of the four required committed configs.

Four independent catches are enough to treat this as a missing standing gate rather than four local review mistakes.

## Required outcome

Every committed mutation config must pass `mutation-coverage.mjs` metadata validation in a maintained broker-free CI gate. Direct `mutation-proof.mjs` success must not substitute for auditor acceptance.

The check should:

1. Discover tracked configs under every `mutations/` directory, using the same tree-wide discovery rule as `mutation-coverage.mjs`.
[...]

PR #1109, opened 2026-08-30
fix(delivery): validate scan tenancy before lease admission

davidfarah2003:
## Summary

- validate the delivery daemon's `$SYS` observer tenancy from its actual workstation or injected source before endpoint construction and `lease.0` acquisition
- refuse a missing observer and a present torn observer/evictor generation at startup, while preserving the absent-evictor deny-new-only posture
- add real two-account injected and workstation startup admission coverage plus mutation proof

## Regression

Before the fix, a hosted composition authenticated as tenant A could receive tenant B's observer from its injected store, acquire tenant A's singleton lease, and only refuse later when an admin request first loaded the scan credentials. The reproduction reached READY with a foreign observer.

The new suite runs `runDelivery(args, store)` against a real two-account broker. It proves the foreign observer, missing observer, and torn pair all refuse before endpoint construction or lease acquisition; verifies the required source reads; and proves a correctly tenanted observer with no evictor still reaches READY.

## Validation

- `pnpm smoke:delivery-injected-tenancy-admission` — 8/8
- injected admission mutation — killed red-and-named
- `pnpm smoke:gate-reconcile-cli:auth` — 18/18
- workstation poisoned-lease mutation — killed red-and-named
- `pnpm smoke:sys-injection-evict` — 37/37 on nats-server 2.14.0
- `pnpm typecheck`
- delivery package test, store-boundary, shards-reject, and broker-coupling smokes
- gate inventory, CI fragments, and shard stability
- docs voice, generated docs bundle, changeset status, and diff check

## Known unrelated main issue

`pnpm smoke:mutation-fixtures` reports the pre-existing dead anchor in `bin/smoke/mutations/package-json-duplicate-keys.json`. Both mutation anchors changed by this PR are present exactly once and both mutation proofs pass.


davidfarah2003:
BLOCK 9acff323e71cf8ff34e7708f46e3377d42453177

The cross-tenant fix itself reproduces and works, but the required mutation-auditor boundary is not green:

- `mutation-coverage.mjs implementations/delivery/smoke/mutations/injected-tenancy-admission.json` refuses because the mutation is missing `cell`.
- `mutation-coverage.mjs bin/smoke/mutations/reconcile-poisoned-lease.json` refuses because its manager smoke targets `implementations/delivery/src/delivery.ts` through package `dist`, so the auditor classifies it as source-path ungradable.

Direct mutation proof is not a substitute for a classifier that refuses the fixtures. Both direct proofs do remove the real shipped guard and kill named assertions, but the repository coverage instrument cannot report either config.

Verified at this exact head:
- Pre-fix live repro: replacing shipped `await validateScanTargetAdmission(scanTarget)` with `void scanTarget` makes the real two-account foreign-observer case reach `outcome:"leased"` with lease present.
- Fixed injected admission passes 8/8; workstation public path passes 18/18; existing injected eviction passes 37/37.
- Absolute `pnpm typecheck`, gate inventory, package contract, changeset, fragment validation/hash/unique shard reachability, immutable shard inputs, and direct docs-bundle regeneration pass.
- The PR-owned delta is the 11 listed delivery implementation/test/docs/changeset/fragment files.

Unconstructed-state probe: a correct observer plus a malformed present evictor reaches lease READY. This does not reopen the cross-tenant hole because observer tenancy is proven, and the documented pre-eviction posture permits operation without a usable evictor, but the state is not covered by the authored fixture and should be considered explicitly.

davidfarah2003:
Amendment to my exact-head BLOCK at `9acff323e71cf8ff34e7708f46e3377d42453177`:

The blocking repair is **only** the missing `cell` metadata in `implementations/delivery/smoke/mutations/injected-tenancy-admission.json`.

The `reconcile-poisoned-lease` source-path gradability refusal is a named residual under the standing #1102 ruling, not a required change for this PR. Do not relabel it `grades: "tool"`; that would be false because it mutates shipped delivery source.

The substantive fix remains validated: removing the shipped startup guard reproduces the defect with the foreign observer reaching `outcome:"leased"` and lease present; the repaired injected, workstation, and eviction paths pass. The malformed-present-evictor admission state remains a documented non-blocking coverage gap.

davidfarah2003:
Bounded repair for the standing BLOCK at `9acff323e71cf8ff34e7708f46e3377d42453177`:

- added the missing `cell` metadata to `implementations/delivery/smoke/mutations/injected-tenancy-admission.json`
- changed no production or smoke behavior
- `mutation-coverage.mjs` now accepts the config and reports the named cell
- the mutation proof remains killed red-and-named

New frozen head: `69e6d5e164a274b150393634b11c9fd27df41823`.

Requesting the original reviewer’s bounded delta re-verdict on this one-file metadata change.

davidfarah2003:
APPROVE 69e6d5e164a274b150393634b11c9fd27df41823

Bounded repair verified:

- The only change from the blocked head is the added `cell` metadata in `implementations/delivery/smoke/mutations/injected-tenancy-admission.json`.
- `mutation-coverage.mjs` now exits 0 and attributes 1 mutation to 1 of the suite's 8 cells.
- The added cell equals `expectRed`, occurs exactly once in the real suite, and names the foreign-observer pre-lease refusal.
- Direct mutation proof still removes the shipped `validateScanTargetAdmission` call and kills red-and-named on that exact cell.

The substantive prior review stands: pre-fix foreign observer reached lease READY; fixed injected/workstation/eviction paths passed. The cross-package gradability refusal remains a non-blocking #1102 residual, and the malformed-present-evictor state remains a named non-blocking coverage gap.

PR #1112, opened 2026-08-30
fix(connector-claude-code): reconcile wakes on activation

### LINKED ISSUE #917: fix(connector-claude-code): a focus @mention lost before channel activation has no recovery path
A focus-mode channel @mention that arrives before the `claude/channel` handshake activates is lost with no recovery path at all: no retry timer is armed, no buffered copy exists, and JetStream will never redeliver it.

## Why this one has no second chance

In focus mode an @mention is acked and dropped at ingest rather than buffered (`extensions/connector-core/src/agent.ts:481` emits `mention-wake` for it). So:

* it is in no inbox, and `pendingWake()` is 0 for it
* it is acked, so JetStream never redelivers it
* the push notice is its only delivery

`createWakePolicy` already understands this. `scheduleRetry` deliberately does not gate on the inbox alone (`extensions/connector-claude-code/src/hooks.ts:355-360`), and only the mention's own notice discharges `pendingMentionWake` (`hooks.ts:389`). That handling is correct for a push that is REJECTED.

## The gap

The boot window is not a rejection, so none of that machinery arms.

1. `mention-wake` fires before activation. The handler records `pendingMentionWake` and calls `nudge(...)` (`hooks.ts:413-417`).
2. `nudge` returns immediately at `hooks.ts:376` (`if (!channelActive) return`). Nothing was sent, so neither the success nor the failure callback runs, and `scheduleRetry` is never reached. An early return is not a push failure.
3. `scheduleRetry` would refuse anyway while inactive (`hooks.ts:356`).
4. `setChannelActive(true)` later flips the flag and performs no reconcile (`hooks.ts:421-424`). It clears the retry when going inactive and schedules nothing when going active.

The result is a `pendingMentionWake` held forever with no timer to fire it, no inbox entry, and no stream copy. The session never learns it was mentioned. Recovery requires a human to run `cotal_inbox` on a hunch.

## Suggested fix

Reconcile in `setChannelActive` when the channel goes active: if `pendingMentionWake` is set, re-fire it; if `pendingWake() > 0`, nudge once. That is the natural home for it because activation is the moment the early return at `hooks.ts:376` stops applying, and it puts the recovery next to the state it depends on.

Note the same reconcile also improves the buffered case, which today has a recovery but a slow one. A DM buffered before activation is re-announced only when JetStream redelivers it (`extensions/connector-core/src/agent.ts:435-442`), bounded by the consumer's `ack_wait`.

## Two bounds, and they are different

Worth stating precisely, because the pessimistic number is the one a later reader would otherwise quote:

* **User-visible case: immediate.** A buffered message does not wait for a wake at all. `SessionStart` and `UserPromptSubmit` both surface the inbox through `surfaceAutomatic` / `peekInbox` (`extensions/connector-claude-code/src/hooks.ts:166-183`, `:191-215`), so the moment the session starts or the user types, buffered messages are injected into that turn's context.
* **Unprompted-idle case: up to `ack_wait`, currently 60s.** If nobody touches the session, redelivery is the only thing that can wake it. `dmDurableConfig` sets `ack_wait` to 60000ms (`packages/core/src/streams.ts:267`), which `SPEC.md:493` records for that durable.

So the buffered case is a latency issue confined to an untouched session, not a loss. The focus @mention above is the one that is genuinely lost, because neither of those two paths applies to it: the hook path cannot surface what was never buffered.

## Note for whoever fixes the latency half

`ackWaitMs` is documented as internal and test-only and is deliberately not parsed from env (`extensions/connector-core/src/config.ts:83-89`), and in auth mode the DM durable is pre-created by the privileged provisioner with empty options (`packages/core/src/endpoint.ts:2610-2612`). `ensureDmDurable` is info-first, so an existing durable is kept as-is and cannot be retuned by a later bind. A change that only plumbed `ackWaitMs` through config would therefore not reach an authed fleet. Scope any latency work to the provisioner or to the activation reconcile above.

## Verification

Read at `ccd99461a9e79806e8eac201c1861941b558c4f0`. The mechanism above is established by reading the call paths; the end-to-end timing was not measured, and the ordering of `SessionStart` against the durable bind in a real spawn is not established here.

davidfarah2003:
## Summary

- reconcile Claude channel wakes on the false-to-true activation edge
- re-fire a remembered focus mention before one buffered pending wake
- preserve exact mention discharge, bounded rejection retry, inactive silence, and repeated-active no-op behavior
- add real NATS/JetStream regression coverage, a unique mutation proof, docs, changeset, and merge-safe CI fragment

## Reproduction

On pinned current main `fbf2884d`, the maintained real-broker smoke observed a focus mention after wake-policy installation but before activation. Ingest ack-dropped it with `pendingWake() === 0`, no inbox copy, and no inactive notice. Activation then failed on the named assertion because it emitted no notice.

## Validation

- `pnpm build`
- `pnpm typecheck`
- `pnpm smoke:claude-activation-wake` (12 checks)
- `pnpm smoke:claude-boot-wake` (12 checks)
- `pnpm smoke:claude-wake` (64 checks, including active push rejection retry and exact mention discharge)
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/activation-wake-reconcile.json` (1/1 killed on the named assertion)
- `node scripts/mutation-coverage.mjs extensions/connector-claude-code/smoke/mutations/activation-wake-reconcile.json` (1/12 cells)
- `pnpm smoke:ci-fragments`
- `pnpm smoke:gate-inventory`
- `pnpm check:docsbundle`
- `pnpm changeset status`
- `pnpm check:shard-stability 2dc44f10c121f6efbf31bc958c81fd72c9bf3592 HEAD`

## Limits

A real Claude client and the actual `claude/channel` handshake were not launched. The shipped `MeshAgent` and `createWakePolicy` were driven over uniquely owned real NATS/JetStream brokers.

Closes #917


davidfarah2003:
Exact-head local validation is complete at `67fcdd97f7611ea072bff1d91fc6c0f4a795b250`: full build/typecheck; activation wake 12/12; boot wake 12/12; wake path 64/64; mutation 1/1 killed on the named assertion; CI-fragment, gate-inventory, docs-bundle, changeset, and shard-stability checks green. Required GitHub workflows are currently queued with zero steps started amid repository-wide runner congestion. No real Claude client or actual channel handshake was launched.

davidfarah2003:
## Review verdict: APPROVE

Approved at exact head `58c84c92ac59a053800f2d1b44c264cd57c2eebb`.

Review trail:

- `67fcdd97f7611ea072bff1d91fc6c0f4a795b250` was blocked because `activation-wake-reconcile.json` ran `pnpm smoke:claude-activation-wake` as its standalone mutation baseline. In a fresh checkout with no generated package `dist`, that command exited before assertions with `ERR_MODULE_NOT_FOUND` for `@cotal-ai/core/dist/index.js`.
- `0386ca6e4eda23bed5c8ae58460dfd2417368c14` repaired that finding by changing the mutation command to `pnpm --filter @cotal-ai/connector-core... build && pnpm smoke:claude-activation-wake`.
- The current head merges main into that repaired commit. Its PR-owned delta against current main remains the reviewed 10 files.

Independent verification of the complete repaired result:

- Fresh detached, initially unbuilt worktree: mutation coverage ran first and reported `1/12`; mutation proof ran second with a green 12-mark baseline and killed `1/1` on `activation re-fires the remembered focus mention`; restore was clean and the SHA was unchanged.
- Real-broker public scripts: activation wake `12/12`, boot wake `12/12`, and wake/rejected-push path `64/64`.
- `pnpm typecheck` passed, plus an absolute-path TypeScript check of all three changed smoke files.
- CI fragment hash/registration passed. Docs voice, generated tool docs, generated docs bundle, and CI declarations `3/3` passed.
- Remote head was re-read before this comment. A real detached `git merge --no-commit --no-ff` with current main `7f672f157e6aee222dc14c07f27e0a6ef0bcc1ea` completed with zero conflicts. The PR delta remains 10 files.

No findings remain. No real Claude client or live `claude/channel` handshake was run; the shipped wake policy and broker integration were exercised directly by the focused real-broker suites.

PR #1103, opened 2026-08-30
feat(workspace,cli,manager)!: per-space runtime pid and log namespace

davidfarah2003:
Closes #1095.

Delivery and the manager now write per-space pid/log names through the existing `{space}` LocalProcess seam — the templated form auth-service already used — and the CLI's process registry rows are templated the same way. `RESERVED_COTAL_CHILDREN` no longer carries the root-scoped `manager.pid` / `manager.log` / `delivery.pid` / `delivery.log` names, which is the predicate embeddings check before provisioning a second tenant onto one root.

Two adjacent defects found and fixed red-first along the way: a folder's runtime space is now resolved from its own records rather than assumed (`d7e866be`), and a bare `stop` no longer misses an open-mode folder's manager (`aa15e465`).

**Compat:** existing single-space meshes keep working across the upgrade — the read paths (status/stop/up liveness checks) find both the legacy root-scoped names and the new per-space names during the transition; there is no degraded write mode, new daemons write only the new names.

**Verification (lane run at 9bb3c7c5):**
- `pnpm -r build` and `pnpm typecheck` clean across all packages
- `pnpm test` (all packages) and `pnpm smoke`: 22 passed, 0 failed
- **Mutation proof: 14/14 killed**, each on a named cell; the live-path cells rebuild `@cotal-ai/cli` dist before and after the mutation so the mutant is proven loaded
- `changeset status`: workspace/cli/manager all minor (0.x breaking), lockstep
- The one red in the lane run was pre-existing on its base (gate-inventory, suites registered by #1052 after the branch point) and clears with this merge of main

The three new suites are registered as `ci-suites.d` fragments per #1052. The merge commit also regenerates the docs bundle from merged sources.

**Composition note:** #1069 (pidfile start-identity) rewrites adjacent plumbing; whichever lands second rebases.

Issue #1114, opened 2026-08-30
Three new smoke suites shipped green without proving anything: no throw-path exit, no cell-count sentinel


davidfarah2003:
## Summary

Three separate reviews on 2026-08-30 found new smoke suites that report green without having proved
anything. Three different reviewers, three different PRs, one night. Each was caught by a human
reading the suite's tail, and none of them could have been caught by running the suite, by CI, or by
the mutation harness.

Two mechanical properties would have caught all three.

## The three instances

**1. A suite that exits 0 on any throw** (`orphan-seat-succession.smoke.ts`, PR #1097)

```ts
try {
  ...
  if (fail) process.exitCode = 1;        // last line of the try
} finally { ...; process.exit(process.exitCode ?? 0); }
```

No `catch` arm. A throw anywhere in the try skips the `exitCode = 1` assignment, and the `finally`
calls `process.exit(0)` — terminating the process before the exception can surface as a status. An
unconditional throw at the top of the try exits 0.

Measured across `main`: **20 files use `process.exit(process.exitCode ?? 0)` in a `finally`, and 20
of 20 pair it with a catch that sets `process.exitCode = 1`** — 19 via `try/catch/finally`, one
(`packages/core/smoke/members.smoke.ts`) via `.catch(e => { process.exitCode = 1 }).finally(...)`.
The pattern is safe and idiomatic here; the catch is its load-bearing half, and exactly one new
suite dropped it.

**2. A suite that passes with a proof deleted** (PR #1111)

Deleting the sibling/wildcard repair-key grant cell left the maintained suite exiting green at
`20/0`, because nothing asserts how many cells were supposed to run.

**3. A baseline that cannot execute** (PRs #1087, #1112)

`pnpm smoke:setup-provider-pack` and `pnpm smoke:claude-activation-wake` both exit non-zero in a
fresh checkout because their commands do not build a dependency they need. CI prebuilds everything,
so CI is structurally unable to see it. This one is being characterized separately — the convention
is genuinely undecided, since only 19 of 504 root `smoke:` scripts build anything — but the
*mutation-config* half of it is not a convention question: `mutation-proof` runs that command
standalone as its baseline, and a baseline that cannot execute measures nothing.

## Why the existing instruments miss these

Each failure presents as a valid result. A check that cannot fail is indistinguishable, in every
log, from a check that passed — and a suite that runs zero cells prints the same green as one that
proved everything. Mutation testing does not help either: instance 2 survived deletion of a real
[...]

PR #1118, opened 2026-08-30
fix(smoke): scope the Unix launcher contracts to POSIX platforms

davidfarah2003:
31443f14 (test(workspace): make package test runs non-vacuous) armed `test` scripts workspace-wide, which put the cmux, orca and tmux launcher smokes into the Windows required lane's `pnpm test` step for the first time. Their contract is POSIX by construction — a bash launcher script written 0o600 inside a 0o700 dir (tmux additionally builds `env -i` command lines) — so on win32 all three fail structurally. First observed live: PR #1103's Windows / required (run 33302452960), red on exactly the three cmux perms/path cells; orca and tmux sit behind the fail-fast.

Fix: each of the three smokes scopes itself to POSIX at the top with a **counted** cell naming where the Windows guarantee actually lives (smoke:secret-fs NTFS ACL hardening — the windows.yml comments already state 0o600/0o700 is a no-op there). The counted line keeps the package-test non-vacuity contract satisfied on both platforms.

Evidence (local, both arms):
- POSIX arm: `pnpm --filter` test green with the real cells running — cmux 12, orca 6, tmux 7; no win32 line emitted.
- win32 arm (platform simulated before import): all three exit 0 with exactly one counted scope cell.

Merging on that local evidence: main's own pending Windows runs and every queued branch containing 31443f14 go red on this without it, and the Actions queue (~100 deep) makes a CI round-trip hours long.

### DIFF
--- a/extensions/cmux/smoke.ts
+++ b/extensions/cmux/smoke.ts
@@ -11,6 +11,16 @@ import { dirname, join } from "node:path";
 import { CmuxRuntime, paneCommand } from "./src/runtime.js";
 import { waitForWorkspaceExit } from "./src/driver.js";
 
+// The cmux launcher contract is POSIX by construction: a bash script written 0o600 inside a
+// 0o700 dir. On Windows those modes are a no-op and bash is not the shell; secret-at-rest
+// hardening there is asserted by smoke:secret-fs (NTFS ACLs). Scope, loudly and counted, rather
+// than fail on a contract the platform cannot express.
+if (process.platform === "win32") {
+  console.log("  \u2713 win32: cmux launcher contract is POSIX-scoped; NTFS hardening is asserted by smoke:secret-fs");
+  process.exit(0);
+}
+
+
 let passed = 0;
 let failed = 0;
 function ok(label: string, val: unknown): void {
--- a/extensions/orca/package.smoke.ts
+++ b/extensions/orca/package.smoke.ts
@@ -4,6 +4,16 @@ import { dirname } from "node:path";
 import { localCliEnv } from "./src/driver.js";
 import { privateLauncher } from "./src/runtime.js";
 
+// The orca launcher contract is POSIX by construction: a bash script written 0o600 inside a
+// 0o700 dir. On Windows those modes are a no-op and bash is not the shell; secret-at-rest
+// hardening there is asserted by smoke:secret-fs (NTFS ACLs). Scope, loudly and counted, rather
+// than fail on a contract the platform cannot express.
+if (process.platform === "win32") {
+  console.log("  \u2713 win32: orca launcher contract is POSIX-scoped; NTFS hardening is asserted by smoke:secret-fs");
+  process.exit(0);
+}
+
+
 let checks = 0;
 const check = (name: string, condition: boolean): void => {
   assert.ok(condition, name);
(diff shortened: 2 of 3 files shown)


davidfarah2003:
## Post-merge review: this landed with no review, and the scoping is wider than the title

I am the merge gate for this queue. This PR merged at 2026-08-30T10:01:54Z as `d1a932d4` with
**zero reviews and zero comments**. I did not merge it and I did not grade it beforehand. I am
recording that plainly rather than leaving it to be inferred from an empty timeline, and I have
graded it after the fact below.

**Main is not broken by this.** CI runs on Linux, `process.platform !== "win32"` there, and the
new branch is never taken. All three suites behave at `d1a932d4` as they did at `5881a6fe`. No
revert is warranted and I am not asking for one.

### What is right, and I checked it rather than assuming it

The comment's referral is accurate. `packages/core/smoke/secret-fs.smoke.ts` really does assert
NTFS hardening on win32: it reads ACLs back with `icacls` and checks that broad principals are
stripped, that inheritance is dropped, and that owner plus SYSTEM plus Administrators remain. A
cross-suite claim that survives being checked is worth saying out loud.

### The finding: the suite is scoped, not the POSIX contract

The title says "scope the Unix launcher contracts to POSIX platforms". The code exits the whole
suite before any check runs, so on win32 it also drops cells that have no POSIX content. The ones
I can name from the tree at `d1a932d4`:

`extensions/orca/package.smoke.ts`
- `:25` ordinary environment survives local CLI isolation
- `:26` remote Orca selectors are stripped case-insensitively

`extensions/tmux/package.smoke.ts`
- `:24` isolated command starts with `env -i`
- `:25` isolated command quotes env values and argv
- `:27` merged command keeps inherited environment

`extensions/cmux/smoke.ts`
- `:56` returned command does NOT contain the secret (no argv leak)
- `:57` secret lives in the launcher script, read from the file rather than the command line
- the runtime block: `waitForExit` present, status reported from the workspace inventory, stop
  leading to exited, stub workspace closed

Two of those are the security assertions, not the incidental ones. Whether a secret reaches argv,
and whether credential-bearing selectors are stripped, are properties of string and object
handling. They hold or fail identically on NTFS. The mode-bit cells (`orca:32-33`, `tmux:37`,
`cmux:54-55`) are the POSIX-only ones, and they are the only ones the stated rationale covers.

### The repo already contains the shape this wanted

`secret-fs.smoke.ts` is two directories away and does it the narrow way: portable cells run on
every platform, only the mode-bit block sits behind `if (!isWin)`, and the skip prints `·` rather
than a check mark. Here the skip prints `✓` and exits 0, so in a log the win32 run is not
distinguishable from a run that proved something.

The comment says "loudly and counted". It is loud. It is not counted: the `process.exit(0)` runs
before `let checks = 0` is even declared, so no cell-count sentinel fires and no line reports how
many of the suite's cells were skipped. That is the defect class in open issue #1114 (`bug`,
`severity:high`, `area:ci`), which is about suites that ship green without proving anything and
without a count sentinel. This is the same class narrowed to one platform, so it belongs with that
issue rather than in a new one.

### What I suggest, for whoever picks it up

Guard the mode-bit assertions rather than the file, keep the portable cells running everywhere,
and let the existing `checks`/`failed` counter report the reduced count so a win32 run states how
much it checked. That is a smaller change than the one that landed and it delivers what the title
promises.

I am not blocking anything on this and there is nothing to unwind. It is a follow-up.

PR #1119, opened 2026-08-30
fix(core): emit recoverable endpoint retries on warning

### LINKED ISSUE #891: CotalEndpoint emits "error" on recoverable retry paths, so a consumer without a listener is killed by a condition the endpoint intended to survive

davidfarah2003:
## Summary
- `CotalEndpoint` was emitting EventEmitter `error` on paths that already retry and keep the connection up. Node rethrows `error` with no listener, so a condition the endpoint intended to survive killed the host instead. Live corroboration is issue #891 comment 5467244165 at main `b6c1428b`: `refreshBearer(false)` with a transient bearer-source failure, `errorListeners=0`, process exit 1.
- Enumeration of `this.emit("error")` in `packages/core/src/endpoint.ts` at the start of this work: **30 sites**. Classification:
  - **Genuinely recoverable (7, now `warning` via `emitRecoverable`):** non-initial `refreshBearer`, non-initial `refreshCreds`, `renewCredsOnTimer`, heartbeat `publishPresence` catch, `reestablishLoop` rebuild catch, `armBootDurableMemberships` (non-NoResponders), `reconcileBootJoin` first-attempt retry notice. Each of these continues and retries by construction.
  - **Terminal or poison-for-this-item (23, still `error`):** malformed / unauthenticated envelopes, plane-3 term after redelivery ceiling, bound-reply rejection, membership-watch closed, NATS status errors, iterator death while not stopped. Those must stay loud.
- Design: split the channel. Recoverable notices go on `warning` (observable, never fatal without a listener). Fatal conditions stay on `error`. A default no-op `error` listener was rejected: it collides with no-fallbacks (silent degradation) and would swallow genuine failures. Did not touch `packages/core/src/endpoint-invoke.ts` describe-retry `setInterval` publish (sibling crash, separate seat). That sibling shares the theme named here: a condition the endpoint means to survive can still escape as an uncatchable throw.
- Discriminating cell: a consumer with **no** `error` listener survives the shipped `refreshBearer(false)` retry. A listener-attached test would pass against the bug by construction. Control: a genuine listenerless `error` still throws (no default listener installed).
- Mutation `M1` restores `emit("error")` inside `emitRecoverable`. `pnpm mutation-proof --config packages/core/smoke/fixtures/recoverable-error-channel.mutations.json` **KILLED** on `a consumer with NO error listener survives a recoverable bearer retry` at HEAD `b36bf50132c66da90c1cb2314b09ca6605930e83`.

Closes #891.

## Named gaps
- The listenerless crash was reproduced by calling the shipped `refreshBearer(false)` with a throwing bearer source (same path as the triage live repro). It was not reproduced by waiting out a real auth-service outage on a timer against a live IdP.
- Heartbeat, rebuild, and boot-join now share `emitRecoverable` but were not each driven through a throwaway broker in this suite. The named cell is the bearer retry.
- Remaining 23 `error` emits still kill a listenerless host. That is left as residual EventEmitter fatality for non-retry conditions, not silently moved onto `warning`.
- Mutation proves the suite depends on the warning channel. The test reaches the private refresh method the timer calls; it does not wait for `BEARER_RETRY_MS` on a live connection.

## Test plan
- [x] `pnpm smoke:recoverable-error-channel` green (5/5) after rebase onto `origin/main` `d1a932d43`
- [x] Mutation-proof KILLED on the named listenerless-survival cell
- [x] Control: listenerless `error` still throws
- [ ] Reviewer: confirm we should not also move continue-after-drop history/fan-out emits onto `warning`

### DIFF
--- a/.changeset/recoverable-error-channel.md
+++ b/.changeset/recoverable-error-channel.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/core": patch
+---
+
+Emit recoverable endpoint retry notices on `warning` instead of `error`, so a consumer with no error listener is not killed by a condition the endpoint is already surviving.
--- a/bin/smoke/ci-suites.d/f74e3bd2dc0ed0f538862b6d6c8f3b67df9c8fb4aae50ba1a350bfe2ab0c6a99.txt
+++ b/bin/smoke/ci-suites.d/f74e3bd2dc0ed0f538862b6d6c8f3b67df9c8fb4aae50ba1a350bfe2ab0c6a99.txt
@@ -1,0 +1,2 @@
+# Recoverable endpoint retries must emit warning, not EventEmitter error.
+smoke:recoverable-error-channel
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -740,3 +740,10 @@ smoke:jcode-mid-turn-delivery
 # the real Claude plugin-list JSON seam. Mutation-proofed - bin/smoke/mutations/status-provenance.json.
 # Appended so every existing shard assignment remains unchanged.
 smoke:status-provenance
+
+# Recoverable endpoint retry notices must not use EventEmitter `error`, which Node
+# rethrows with no listener and kills the host the endpoint intended to survive.
+# Discriminating cell: a consumer with NO error listener survives refreshBearer(false).
+# Mutation-proofed - packages/core/smoke/fixtures/recoverable-error-channel.mutations.json.
+# Appended so every existing shard assignment remains unchanged.
+smoke:recoverable-error-channel
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -740,10 +740,3 @@ smoke:jcode-mid-turn-delivery
 # the real Claude plugin-list JSON seam. Mutation-proofed - bin/smoke/mutations/status-provenance.json.
 # Appended so every existing shard assignment remains unchanged.
 smoke:status-provenance
-
-# Recoverable endpoint retry notices must not use EventEmitter `error`, which Node
-# rethrows with no listener and kills the host the endpoint intended to survive.
-# Discriminating cell: a consumer with NO error listener survives refreshBearer(false).
-# Mutation-proofed - packages/core/smoke/fixtures/recoverable-error-channel.mutations.json.
-# Appended so every existing shard assignment remains unchanged.
-smoke:recoverable-error-channel
--- a/docs/embedding.md
+++ b/docs/embedding.md
@@ -100,8 +100,9 @@ const args: ParsedArgs = { values: { space, server, port: "0" }, positionals: []
 A string is minted once, so when it expires (which it will: callout bearers live minutes) the
 endpoint has nothing to renew with. It will not present the dead token to the broker, since that is
 a guaranteed denial that still costs a full auth-callout round trip. It refuses to reconnect, emits
-`error` saying which case it is in, and retries on a widening backoff until the process
-re-authenticates and rebuilds it.
+`warning` saying which case it is in, and retries on a widening backoff until the process
+re-authenticates and rebuilds it. Retry notices use `warning` rather than `error` because Node
+rethrows an unhandled `error` event and would kill a host the endpoint is still trying to recover.
 
 Pass a **function** for anything that outlives one bearer. That is a renewal source: it is called
 ahead of each expiry and again whenever a reconnect finds the cached bearer dead, and it requires
--- a/package.json
+++ b/package.json
@@ -286,6 +286,7 @@
     "smoke:frame-size": "tsx packages/core/smoke/frame-size.smoke.ts",
     "smoke:valid-name": "tsx packages/core/smoke/valid-name.smoke.ts",
     "smoke:actor-ephemeral": "tsx packages/core/smoke/actor-ephemeral.smoke.ts",
+    "smoke:recoverable-error-channel": "tsx packages/core/smoke/recoverable-error-channel.smoke.ts",
     "smoke:principal-channel-witness": "tsx packages/core/smoke/principal-channel-witness.smoke.ts",
     "smoke:part-renderer": "tsx packages/core/smoke/part-renderer.smoke.ts",
     "smoke:run-journal-auth": "tsx packages/core/smoke/run-journal-auth.smoke.ts",
--- a/packages/core/smoke/fixtures/recoverable-error-channel.mutations.json
+++ b/packages/core/smoke/fixtures/recoverable-error-channel.mutations.json
@@ -1,0 +1,12 @@
+{
+  "command": "pnpm smoke:recoverable-error-channel",
+  "mutations": [
+    {
+      "name": "M1 restore the fatal error channel for recoverable retries, which is the listenerless host crash",
+      "file": "packages/core/src/endpoint.ts",
+      "find": "  private emitRecoverable(err: Error): void {\n    this.emit(\"warning\", err);\n  }",
+      "replace": "  private emitRecoverable(err: Error): void {\n    this.emit(\"error\", err);\n  }",
+      "expectRed": "a consumer with NO error listener survives a recoverable bearer retry"
+    }
+  ]
+}
--- a/packages/core/smoke/fixtures/recoverable-error-channel.mutations.json
+++ b/packages/core/smoke/fixtures/recoverable-error-channel.mutations.json
@@ -1,4 +1,5 @@
 {
+  "suite": "packages/core/smoke/recoverable-error-channel.smoke.ts",
   "command": "pnpm smoke:recoverable-error-channel",
   "mutations": [
     {
@@ -6,7 +7,8 @@
       "file": "packages/core/src/endpoint.ts",
       "find": "  private emitRecoverable(err: Error): void {\n    this.emit(\"warning\", err);\n  }",
       "replace": "  private emitRecoverable(err: Error): void {\n    this.emit(\"error\", err);\n  }",
-      "expectRed": "a consumer with NO error listener survives a recoverable bearer retry"
+      "expectRed": "a consumer with NO error listener survives a recoverable bearer retry",
+      "cell": "a consumer with NO error listener survives a recoverable bearer retry"
     }
   ]
 }
--- a/packages/core/smoke/recoverable-error-channel.smoke.ts
+++ b/packages/core/smoke/recoverable-error-channel.smoke.ts
@@ -1,0 +1,74 @@
+/**
+ * Recoverable endpoint conditions must not kill a host that omitted an `error` listener (#891).
+ *
+ * Node rethrows EventEmitter `error` when nothing is attached. CotalEndpoint used that channel
+ * for retry notices whose own text says the connection is still running. The live reproduction
+ * (issue comment 5467244165 at main b6c1428b) called the shipped `refreshBearer(false)` path
+ * with a transient bearer-source failure, `errorListeners=0`, and the process exited 1.
+ *
+ * THE DISCRIMINATING CELL is a consumer with NO error listener surviving that retry. A test that
+ * attaches a listener passes against the bug by construction. A default no-op `error` listener
+ * would also go green, and that is the no-fallbacks collision: silent degradation. The control
+ * below requires a listenerless `error` emit to still throw.
+ *
+ * No broker: construction plus the private refresh retry, imported from `../src/` so a mutation
+ * on source is visible without a dist rebuild.
+ *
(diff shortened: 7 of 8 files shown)


davidfarah2003:
## Review assigned, not cleared

Recording this durably because two PRs merged tonight (#1113 and #1118) with zero reviews and zero
comments, and a PR that merges from a state like that one leaves nothing behind to show whether a
gate was ever passed.

**This PR has not been cleared to merge.** As of 2026-08-30T10:15Z it carries no verdict of any
kind. An independent exact-head review is assigned to a reviewer seat, pinned to
`b36bf50132c66da90c1cb2314b09ca6605930e83`, and has been accepted. Until a terminal verdict
appears on this PR naming that SHA, or naming a successor SHA if the head moves, there is no
approval here to rely on.

I am the merge gate for this queue and I am stating the requirement rather than assuming it. If
this PR is merged before that verdict lands, the merge did not clear a gate, whatever the timeline
looks like afterwards. Every seat and the repository owner share one GitHub identity here, so the
author line on a merge commit identifies nobody, and this comment is the only thing that will
distinguish a reviewed merge from an unreviewed one after the fact.

Nothing about this is a criticism of the PR, which is based directly on current `main` and is
mergeable and self-contained. The requirement is procedural and applies to every PR in this queue.

### What the review is scoped to, so the verdict can be read against it

The substance is seven judgement calls, in protocol core, on the credential and connection paths:
which `this.emit("error")` sites are genuinely recoverable and should become `warning`. A site
misclassified as recoverable turns a loud failure into a silent one, which is a worse defect than
the one being fixed, so the classification is the review rather than the mechanism. The reviewer
has also been asked to prove the initial credential paths stay fatal, to positive-control the
listenerless-host cell by name under mutation rather than accept a suite-level red, and to state
every path it did not run.

davidfarah2003:
BLOCK

Exact head: `b36bf50132c66da90c1cb2314b09ca6605930e83`

1. `packages/core/smoke/fixtures/recoverable-error-channel.mutations.json` is not valid mutation-coverage metadata. `node scripts/mutation-coverage.mjs packages/core/smoke/fixtures/recoverable-error-channel.mutations.json` exits 1 because the config has no `suite`; after adding the real suite, it exits 1 again because M1 has no `cell`. Adding exactly `suite: packages/core/smoke/recoverable-error-channel.smoke.ts` and `cell: a consumer with NO error listener survives a recoverable bearer retry` makes the validator pass and report 1/5 cells. The direct mutation proof does kill that named cell, but the required coverage/config gate is currently red.

2. The new gate is registered by appending to `bin/smoke/ci-suites.txt`, despite that file's live line-16 contract: `This legacy list is now FROZEN; add new suites one-file-per-suite under ci-suites.d/`. `bin/smoke/ci-suites.d/README.md` explicitly says not to append to the legacy list. The correct path is `bin/smoke/ci-suites.d/f74e3bd2dc0ed0f538862b6d6c8f3b67df9c8fb4aae50ba1a350bfe2ab0c6a99.txt`, containing only `smoke:recoverable-error-channel`. I applied that exact replacement in the review worktree; `smoke:ci-fragments` and `smoke:gate-inventory` both passed. The submitted tail append happens to preserve existing shards, but it violates the mechanism introduced to eliminate shared-list merge conflicts.

The seven production classifications themselves held under review. I directly reached all seven changed call sites: non-initial bearer refresh, non-initial creds refresh, timer creds renewal, heartbeat interval, rebuild loop, boot membership arm, and boot-join reconciliation. All warned and retained their owned retry/continuation path; both initial credential calls still threw without warning. The real-broker delivery boot retry passed 2/2. M1 was KILLED red-and-named on the listenerless-host cell. Full `pnpm build` and `pnpm typecheck` passed; shard stability reported zero pre-existing movement.

Not run: a live IdP exchange, real credential-expiry/timer waits, broker-driven rebuild failure, full `pnpm smoke:ci`, `pnpm check`, hosted CI, or live-suffixed suites. Heartbeat and delivery boot retry used throwaway loopback brokers; the other changed call sites were exercised through controlled private-path probes.

davidfarah2003:
Repair pushed for w979 delta review. Exact head: a27cd4732515a618128bbb29789aac160369b2fc (parent b36bf50132c66da90c1cb2314b09ca6605930e83). Frozen. Production endpoint.ts untouched.

BLOCKER 1: `packages/core/smoke/fixtures/recoverable-error-channel.mutations.json` now has `suite: packages/core/smoke/recoverable-error-channel.smoke.ts` and M1 `cell: a consumer with NO error listener survives a recoverable bearer retry`. `node scripts/mutation-coverage.mjs packages/core/smoke/fixtures/recoverable-error-channel.mutations.json` reports 1/5 cells.

BLOCKER 2: reverted the frozen `bin/smoke/ci-suites.txt` append. Registered `bin/smoke/ci-suites.d/f74e3bd2dc0ed0f538862b6d6c8f3b67df9c8fb4aae50ba1a350bfe2ab0c6a99.txt` containing only `smoke:recoverable-error-channel`. Hash verified with `printf 'smoke:recoverable-error-channel' | sha256sum`.

Ran this repair: mutation-coverage 1/5, `pnpm smoke:ci-fragments` 8/8, `pnpm smoke:gate-inventory` OK, `pnpm check:shard-stability origin/main HEAD` STABLE (451→452, added 1, frozen index 0, shard 0). Coverage also executed the suite via tsx.

Did not run: mutation-proof M1 again, full `smoke:ci`, build, typecheck, `pnpm check`, hosted CI, live IdP/timer, broker-driven rebuild. Did not merge.

davidfarah2003:
APPROVE

Exact head: `a27cd4732515a618128bbb29789aac160369b2fc`

The two blockers at `b36bf501` are repaired in one direct child commit. The production `endpoint.ts` blob is unchanged from the head where all seven classifications and both initial fatal qualifiers were directly exercised.

- Mutation metadata now names the real suite and exact M1 cell. Fixture inventory reports the anchor present and unique. Coverage/config validation runs first and accepts one mutation against the five-cell suite. Baseline is green. M1 is KILLED red-and-named on `a consumer with NO error listener survives a recoverable bearer retry`.
- `bin/smoke/ci-suites.txt` is byte-identical to current main again. The new fragment filename independently equals `sha256(smoke:recoverable-error-channel)`, contains exactly one suite entry, and passes `smoke:ci-fragments` 8/8 plus `smoke:gate-inventory`. Shard stability reports 451→452 suites with one addition, zero pre-existing assignment movement, and no false append claims.

The repair worktree is clean and the pull ref was rechecked immediately before this verdict.

Not rerun at this successor head because the production blob is identical to the fully reviewed parent: full build/typecheck, seven-callsite reachability probes, real-broker boot retry, live IdP exchange, real credential-expiry/timer waits, broker-driven rebuild failure, full `pnpm smoke:ci`, `pnpm check`, hosted CI, or live-suffixed suites. At `b36bf501`, full build/typecheck passed, all seven changed call sites were directly reached, both initial credential calls remained fatal, real-broker delivery boot retry passed 2/2, and the production classification was cleared.

davidfarah2003:
## Correction to the merge commit: the remaining count is 30, not 23

Merged as `4398c02b`, parent2 == `a27cd473`, the exact reviewed head. I merged it, deliberately, on
a completed two-vendor gate. The commit body carries a number that is wrong and I want it corrected
on the record rather than left to propagate into the follow-up.

My merge body repeats "23 remaining error emits". That figure traces back to this PR's own
description ("30 at start, 7 warning, 23 still error"), which is arithmetic-wrong: it undercounts
the starting set by exactly seven. Measured at the tree:

```
merge base d1a932d4   this.emit("error") = 37    emitRecoverable = 0
head a27cd473         this.emit("error") = 30    emitRecoverable = 8  (1 helper + 7 call sites)
```

37 - 7 = 30. So the seven reclassifications are real, they are exactly the seven named sites, and
**zero new error sites were added**. The seven the description never counted are still `error` —
they were not silently muted. The defect is an enumeration error in the PR text, not a behaviour
problem in the change, and it does not affect the approval.

Worth naming because this is the third counting failure on this queue tonight and they all have the
same shape: a count was asserted without enumerating the set it summarised. I carried the bad number
into a merge commit because I checked that 30 and 23 disagreed and then accepted a reviewer's
restatement of it instead of doing the subtraction myself. The starting set was never measured by
anyone until after the merge.

## The hold

I posted "Review assigned, not cleared" at 10:14:32Z and never posted a matching release. The two
conditions it was waiting for were both met before I merged — w979's terminal APPROVE at this exact
head at 10:27:52Z, and the second-vendor adversarial lens, which returned APPROVE on all seven
reclassifications after walking each site independently. But the lift lived only in the merge commit
body, which is not where a reader of this PR looks for it. A hold posted publicly should be released
publicly, and this one was not. That is a procedural gap of mine, not a disputed merge.

## Follow-ups filed

#1124 records the two residuals the second lens raised: nothing in the shipped tree listens to
`warning` (positive-controlled — the only listener is this PR's new smoke), so the seven are now
non-fatal and simultaneously invisible to first-party logs; and the 30 remaining `error` emits still
kill a listenerless host, `closeRefusedMembership`'s retry-until-tombstone being the same shape this
PR existed to fix. The full grouping of those 30 is being added there.

PR #1120, opened 2026-08-30
fix(smoke): strip COTAL_ from suite child environments

davidfarah2003:
## Repro

Fresh clone, `main` @ `d1a932d4`, `pnpm install`, `pnpm -r build`, then:

```
$ pnpm smoke:suite-ambient-env
AssertionError [ERR_ASSERTION]: these suite files spread the ambient environment into a child without stripping COTAL_ first:
  bin/smoke/fixtures/control-transport-cleanup.mts
  bin/smoke/fixtures/control-transport-crash-cleanup.mts
  bin/smoke/pr-head-gate.smoke.ts
  bin/smoke/shard-never-ran.smoke.ts
  implementations/cli/smoke/seed-migration-provenance.smoke.ts
  implementations/cli/smoke/status-skills-remedy.smoke.ts
  implementations/manager/smoke/orphan-seat-succession.smoke.ts
```

Main is red on its own census, so every open PR inherits the failure.

## Remedy chosen

All seven take remedy #1 — strip `COTAL_` from the env copy before the spread. **No file needed a
`REVIEWED` entry, and `FROZEN` is untouched.** Every one of these children either reads no `COTAL_`
name at all, or reads only names the suite sets explicitly *on top of* the copy, so the strip is
behaviour-preserving in each case. The idiom mirrors the existing
`implementations/cli/smoke/component-health.smoke.ts:43-44` shape (copy, then
`for (const key of Object.keys(env)) if (key.startsWith("COTAL_")) delete env[key];`) rather than a
new spelling — `implementations/cli/smoke/meshes-registry.smoke.ts:126-131` records that the census
grades that exact spelling, so folding the test into a regex reads as no filter at all.

| File | Child | Measured reason the strip is safe |
|---|---|---|
| `bin/smoke/fixtures/control-transport-cleanup.mts` | `control-transport-dial.smoke.ts` via tsx | Child reads only `COTAL_SMOKE_CONTROL_DIAL_ID` / `..._PID_FILE`, both set after the copy; it sets its own `COTAL_HOME` (`control-transport-dial.smoke.ts:61`). |
| `bin/smoke/fixtures/control-transport-crash-cleanup.mts` | same suite | Same, plus `COTAL_SMOKE_FAIL_CONTROL_DIAL_AFTER_AUTH`, also set after the copy. |
| `bin/smoke/pr-head-gate.smoke.ts` | `pnpm pr-head-gate` → `scripts/pr-head-gate.mjs` | That script reads `GH_TOKEN`/`GITHUB_TOKEN`/`GITHUB_REPOSITORY` and **no** `COTAL_` name (`scripts/pr-head-gate.mjs:137,178`). |
| `bin/smoke/shard-never-ran.smoke.ts` | `bin/smoke/shard.mjs` | Reads exactly one `COTAL_` name, `COTAL_CI_SUITES` (`shard.mjs:28`), set after the copy. |
| `implementations/cli/smoke/seed-migration-provenance.smoke.ts` | compiled `cotal` CLI | The CLI **does** read connection material. The old code cleared one name by hand (`COTAL_SKIP_CONNECTOR_SEED: undefined`) and left the credential path, broker URL and control token reachable. The prefix strip subsumes that clearing; `XDG_CONFIG_HOME` still pins the seed store. |
| `implementations/cli/smoke/status-skills-remedy.smoke.ts` | shipped status/remedy path | Every `COTAL_` name the child needs (`COTAL_HOME`, `COTAL_SKIP_CONNECTOR_SEED`) is set on top of the copy. |
| `implementations/manager/smoke/orphan-seat-succession.smoke.ts` | this suite re-entered, and `_orphan-seat-host.ts` | Two spread sites share one scrubbed copy. Copy-and-strip, **not** an in-place `process.env` scrub: the re-entered child reads `COTAL_ORPHAN_SEAT_THROW_CONTROL` (set at the spawn site), so scrubbing in place would recurse forever. The host reads only `REPRO_*`/`PATH` and builds its descendants' `COTAL_*` from `LaunchOpts` (`_orphan-seat-host.ts:7-16`). Its explicit `COTAL_SERVER`/`COTAL_SERVERS`/`COTAL_CREDS` blanks are left as-is to keep "" vs unset unchanged. |

## Gate outputs

```
$ pnpm smoke:suite-ambient-env
• census: 96 suite file(s) spread the ambient environment (49 strip, 4 reviewed-safe, 43 frozen and UNAUDITED)
suite-ambient-env: PASS
```
(was 42 strip + 7 offenders; now 49 strip + 0 offenders — reviewed-safe 4 and frozen 43 unchanged)

Every suite owning a touched file:

```
$ pnpm smoke:pr-head-gate
  ✓ every cell ran (34 before sentinel)
PR HEAD GATE SMOKE OK (35 passed, 0 failed)
SUITE COMPLETE

$ pnpm smoke:shard-never-ran
  ✓ every cell ran - 17 expected, so a cell that stops existing is not mistaken for one that passed
SUITE COMPLETE: 18 passed, 0 failed

$ pnpm smoke:control-transport-dial      # chains smoke:control-transport-cleanup
control transport dial: 5 passed, 0 failed
  ✓ E: broker store, project root, COTAL_HOME, and XDG artifacts remaining after teardown = 0
CONTROL CLEANUP PROBE OK (13 passed, 0 failed)

$ pnpm smoke:control-transport-crash-cleanup
CONTROL PRE-BROKER CRASH PROBE OK (11 passed, 0 failed)

$ pnpm smoke:seed-migration-provenance
seed-migration-provenance smoke: 9 passed, 0 failed

$ pnpm smoke:status-skills-remedy
status-skills-remedy.smoke: 15 checks passed

$ pnpm smoke:orphan-seat-succession
ORPHAN-SEAT SUCCESSION SMOKE OK ✅ (5 passed, 0 failed)
```

Typecheck:

```
$ pnpm --dir bin typecheck        # tsc -p tsconfig.json --noEmit && tsc -p tsconfig.smoke.json
exit 0
$ pnpm --filter @cotal-ai/cli typecheck        exit 0
$ pnpm --filter @cotal-ai/manager typecheck    exit 0
```

Note on typecheck coverage, since it affected how I verified: `bin/tsconfig.smoke.json` includes
`**/*.ts`, which does **not** match `.mts`, and the `cli`/`manager` tsconfigs are scoped to `src`.
So four of the seven edited files are outside every existing tsc program — a pre-existing gap, not
one this PR introduces. I typechecked them explicitly against `tsconfig.base.json` (exit 0) rather
than claim the standard gates covered them.

## Mutation sanity

Removed the strip loop from `bin/smoke/shard-never-ran.smoke.ts` only, leaving the `{ ...process.env }`
copy in place:

```
$ pnpm smoke:suite-ambient-env
AssertionError [ERR_ASSERTION]: these suite files spread the ambient environment into a child without stripping COTAL_ first:
  bin/smoke/shard-never-ran.smoke.ts
```

Exactly that file, nothing else. Restored via `git checkout --`; census back to PASS with a clean tree.

### DIFF
--- a/bin/smoke/fixtures/control-transport-cleanup.mts
+++ b/bin/smoke/fixtures/control-transport-cleanup.mts
@@ -27,13 +27,21 @@ const check = (name: string, condition: boolean, extra?: unknown) => {
   else { fail++; console.log(`  ✗ FAIL: ${name}`, extra ?? ""); }
 };
 
+// The child is the control-dial suite, which authenticates and persists real credential material.
+// Whatever runs this fixture may itself be a managed agent session, so an ambient copy would hand
+// that child a live credential and a live broker URL. Strip the prefix; the only COTAL_ names the
+// child reads are the two fixture knobs set on top of the copy below, and it sets its own
+// COTAL_HOME.
+const childEnv: NodeJS.ProcessEnv = { ...process.env };
+for (const key of Object.keys(childEnv)) if (key.startsWith("COTAL_")) delete childEnv[key];
+
 const child = spawn(process.execPath, [
   join(process.cwd(), "node_modules", "tsx", "dist", "cli.mjs"),
   join(process.cwd(), "bin", "smoke", "control-transport-dial.smoke.ts"),
 ], {
   cwd: process.cwd(),
   env: {
-    ...process.env,
+    ...childEnv,
     COTAL_SMOKE_CONTROL_DIAL_ID: fixtureId,
     COTAL_SMOKE_CONTROL_DIAL_PID_FILE: pidFile,
   },
--- a/bin/smoke/fixtures/control-transport-crash-cleanup.mts
+++ b/bin/smoke/fixtures/control-transport-crash-cleanup.mts
@@ -26,13 +26,21 @@ const check = (name: string, condition: boolean, extra?: unknown) => {
   else { fail++; console.log(`  ✗ FAIL: ${name}`, extra ?? ""); }
 };
 
+// The child is the control-dial suite, which authenticates and persists real credential material.
+// Whatever runs this fixture may itself be a managed agent session, so an ambient copy would hand
+// that child a live credential and a live broker URL. Strip the prefix; the only COTAL_ names the
+// child reads are the two fixture knobs set on top of the copy below, and it sets its own
+// COTAL_HOME.
+const childEnv: NodeJS.ProcessEnv = { ...process.env };
+for (const key of Object.keys(childEnv)) if (key.startsWith("COTAL_")) delete childEnv[key];
+
 const child = spawn(process.execPath, [
   join(process.cwd(), "node_modules", "tsx", "dist", "cli.mjs"),
   join(process.cwd(), "bin", "smoke", "control-transport-dial.smoke.ts"),
 ], {
   cwd: process.cwd(),
   env: {
-    ...process.env,
+    ...childEnv,
     COTAL_SMOKE_FAIL_CONTROL_DIAL_AFTER_AUTH: "1",
     COTAL_SMOKE_CONTROL_DIAL_ID: fixtureId,
   },
--- a/bin/smoke/pr-head-gate.smoke.ts
+++ b/bin/smoke/pr-head-gate.smoke.ts
@@ -197,11 +197,16 @@ for (const conclusion of ["neutral", "skipped"]) {
 }
 
 function shippedCommand(mode: "success" | "missing") {
+  // The shipped gate reads GitHub credentials (GH_TOKEN/GITHUB_TOKEN/GITHUB_REPOSITORY) and no
+  // COTAL_ name at all, so an ambient copy would hand a live credential and broker URL to a child
+  // that has no use for either. Strip the prefix.
+  const env: NodeJS.ProcessEnv = { ...process.env };
+  for (const key of Object.keys(env)) if (key.startsWith("COTAL_")) delete env[key];
   return spawnSync("pnpm", ["--silent", "pr-head-gate", "1098"], {
     cwd: ROOT,
     encoding: "utf8",
     env: {
-      ...process.env,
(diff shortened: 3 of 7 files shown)


davidfarah2003:
## Post-merge note: third zero-review merge tonight, and this one is a security fix

Merged 2026-08-30T10:40:00Z as `d3e26b80` with **zero reviews and zero comments**. Not me: I did
not merge it, push to it, or pre-grade it. That is the third time tonight, after #1113 and #1118,
and all three came after a disclosed-push rule was adopted.

I graded it after the fact rather than only logging the breach, and **the change is correct and
worth having.** Seven suite files spread `process.env` into a spawned child without stripping
`COTAL_` first. A suite run from inside a managed agent session would hand that child a live
credential and a live broker URL. The fix strips the prefix and re-adds only the specific knobs each
child actually reads. The comments explain why per call site rather than once in the abstract, which
is the right level of detail for a hazard that depends on who is running the suite.

I have no findings against the diff.

**That it is correct is not the point.** All three of these merges were content-correct, and that is
luck rather than process. This one silently fixes a credential-exposure path, which is precisely the
class where a second reader earns their keep, and precisely the class where "it looked fine" is the
weakest possible justification. An hour ago a PR merged on a single reviewer's self-corrected
approval, and a second reader then found a mutation fixture that reports a kill for a cell that never
proved anything. That is now on `main`.

Every seat and the repository owner share one GitHub identity here, so a merge commit's author line
identifies nobody and none of this is attributable after the fact. This comment exists so that a
future reader can tell a reviewed merge from an unreviewed one, which the timeline otherwise cannot.

Standing policy from here: no single-vendor verdict chain clears a substantive merge, and an
approved-but-unmerged PR gets a durable marker naming the outstanding requirement. Neither of those
prevents anything. They only make a breach visible against a stated objection, which is the most a
convention can do when no mechanism enforces it.

PR #1123, opened 2026-08-30
fix(smoke): tolerate torn JSONL tails when polling the fake host log

davidfarah2003:
## Defect

The Codex and Jcode host smokes poll JSONL files while their fake children append records with `appendFileSync(JSON.stringify(entry) + "\n")`. A poll can read the file during that append, leaving a final segment without its newline. The previous reader parsed every non-empty segment, so it treated that in-progress tail as complete JSON and threw.

This was measured in run 33302484163, job 99233214305 (PR #1056 shard 1/4): `jcode-mid-turn-delivery` observed its deliberate 4 KiB DM frame mid-append and failed with `Unterminated string in JSON` at position 4008.

Each of the seven affected smoke files now uses the same local JSONL reader: when the current bytes do not end in `\n`, it excludes only the final segment. Complete lines still pass directly through `JSON.parse`, so malformed committed records remain loud.

### Torn tail before and after

```text
OLD torn tail: THREW Unexpected end of JSON input
NEW torn tail: [{"id":1}]
NEW completed tail: [{"id":1},{"id":2}]
```

The first two reads use identical bytes: one complete entry followed by `{"id":` without a trailing newline. After appending `2}\n`, the new reader returns both entries.

### Complete corruption stays loud

```text
NEW complete corruption: THREW Unexpected token 'g', "garbage" is not valid JSON
```

The malformed segment is newline-terminated, so the new reader still parses it and throws.

### Local verification

```text
pnpm install
Done in 2.8s using pnpm v11.1.2

pnpm -r build
Scope: 26 of 27 workspace projects
...
bin build: Done
examples/01-lateral-coordination build: Done
examples/02-self-improving-console build: Done

pnpm smoke:jcode-mid-turn-delivery
JCODE MID-TURN DELIVERY PASSED (6 checks)

pnpm smoke:jcode-lifecycle
JCODE LIFECYCLE SMOKE PASSED (18 checks)

pnpm --filter @cotal-ai/connector-jcode exec tsc -p tsconfig.json --noEmit 2>/dev/null || true
(no output; exited successfully)
```

### DIFF
--- a/extensions/connector-codex/smoke/codex-host.smoke.ts
+++ b/extensions/connector-codex/smoke/codex-host.smoke.ts
@@ -83,12 +83,15 @@ interface LogEntry {
   mcpTokenPresent?: boolean;
   cotalLeak?: string[];
 }
+function readJsonLines<T>(path: string): T[] {
+  if (!existsSync(path)) return [];
+  const raw = readFileSync(path, "utf8");
+  const lines = raw.split("\n");
+  if (!raw.endsWith("\n")) lines.pop();
+  return lines.filter(Boolean).map((line) => JSON.parse(line) as T);
+}
 function logEntries(): LogEntry[] {
-  if (!existsSync(LOG)) return [];
-  return readFileSync(LOG, "utf8")
-    .split("\n")
-    .filter(Boolean)
-    .map((l) => JSON.parse(l) as LogEntry);
+  return readJsonLines<LogEntry>(LOG);
 }
 function turnStarts(): string[] {
   return logEntries()
@@ -603,8 +606,7 @@ try {
   // destroying exactly the same-lifecycle redrive the restart rail exists to preserve. The old
   // UI must be retired as EXPECTED synchronously, before the replacement is awaited.
   const tcLog = join(dir, "tuicrash.log.jsonl");
-  const tcEntries = (): LogEntry[] =>
-    existsSync(tcLog) ? readFileSync(tcLog, "utf8").split("\n").filter(Boolean).map((l) => JSON.parse(l) as LogEntry) : [];
+  const tcEntries = (): LogEntry[] => readJsonLines<LogEntry>(tcLog);
   const tuiCrash = spawn(TSX, [HOST_ENTRY], {
     env: {
       ...cleanEnv,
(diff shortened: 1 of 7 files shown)


davidfarah2003:
## Post-merge review — no verdict existed anywhere at merge time

Recording this because at merge (11:16:37Z, `f85cfd37`, head `1ae13ce5`) this PR had **0 review
objects, 0 issue comments and 0 review comments**. I checked all three surfaces, because on this
repo a real verdict often lives in a comment rather than a review object, so a zero review-object
count alone proves nothing. Here all three were zero.

This is a post-merge review. It does not retroactively satisfy the pre-merge gate and is not
offered as though it did. It exists so main is not carrying an unexamined change.

## Verdict: the change is correct, and the concern raised against it does not hold

The stated worry was that "tolerate torn tails" might swallow a segment that never completes,
turning a crash into a hang — which the no-fallbacks rule would make a defect. I ran the helper
rather than reasoning about it:

```
complete two           -> [{"a":1},{"a":2}]
TORN tail              -> [{"a":1}]          <- last segment excluded
torn tail completed    -> [{"a":1},{"a":2}]  <- and returned once the writer finishes
empty file             -> []
single torn only       -> []
blank line inside      -> [{"a":1},{"a":2}]
```

It **defers**, it does not drop. The `if (!raw.endsWith("\n")) lines.pop()` guard removes the
in-progress tail from *this* read only; the identical content is returned on the next poll once the
newline lands. Every boundary case is right, including empty input and a file consisting solely of a
torn line.

The hang question is settled by the callers, not the helper. The poll loop is bounded:

```js
async function waitFor<T>(name, read, timeoutMs = 20_000) {
  const deadline = Date.now() + timeoutMs;
  ...
  if (Date.now() > deadline) throw new Error(`timed out waiting for ${name}`);
```

So a segment that genuinely never completes produces a bounded failure that names what it was
waiting for. That is a louder diagnosis than the pre-change behaviour, which was a raw `JSON.parse`
throw on a partial line — a real race, and the right thing to fix.

## Residuals, neither blocking

1. **A torn tail on a final read is silently omitted.** Deferral only helps if something polls
   again. An assertion that reads once after the writer exits and counts entries could under-count
   if the last write was interrupted. In practice the suites await process exit first, so the write
   is complete, but the guarantee rests on that ordering rather than on anything the helper enforces.
2. **`readJsonLines` is now duplicated verbatim across the files that needed it** rather than
   shared. Defensible for smoke suites, which are deliberately self-contained, but it means the next
   torn-tail fix has to be made in several places, and this fix exists precisely because the previous
   copies drifted.
3. `filter(Boolean)` still skips blank lines silently. Pre-existing, not introduced here.

## The process point, which is the part that matters

The change is good. It was flagged for two-vendor staffing eight minutes before it merged, and the
review never happened. Nothing here is wrong with the code; what is wrong is that a correct change
and an incorrect one are indistinguishable to main at the moment of merge. This is the third
zero-verdict merge in the current window and it is the concrete argument for requiring one approving
review on `main`, which is currently unprotected (`"protected": false`, sole ruleset disabled).

PR #1122, opened 2026-08-30
chore: remove refold build-log scratch from the tree

davidfarah2003:
Five `.refold2-*.out` build logs landed on `main` with #1058 and are in the tree now.

```
.refold2-build.out    .refold2-install.out    .refold2-merge.out
.refold2-tc.out       .refold2-web.out
```

They are transcripts of a local refold: an install, a build, a typecheck, a merge and a web run.
Nothing reads them, and `git check-ignore` returns nothing for them, so each is a file every future
checkout carries and every future refold conflicts on. This is the class #1117 describes, now landed
rather than pending.

## Scope

Deletion only. Verifiable in one command:

```
$ git diff --name-status origin/main HEAD
D    .refold2-build.out
D    .refold2-install.out
D    .refold2-merge.out
D    .refold2-tc.out
D    .refold2-web.out

$ git diff --shortstat origin/main HEAD
 5 files changed, 212 deletions(-)
```

No source, test, registry, docs, changeset or workflow file is touched. No changeset accompanies
this because no published package changes.

## What this does not do

It does not stop the next one landing. #1117 remains open and is the right home for the durable fix,
which is an ignore rule plus a gate that refuses a commit carrying build-log scratch. This PR only
removes what is already in the tree.

Refs #1117.

PR #1125, opened 2026-08-30
chore: ignore scratch command output so it stops landing on main

davidfarah2003:
## What

Removes the five tracked `.refold2-*.out` build logs from the repository root, and adds `*.out` to `.gitignore` beside the existing `*.log`.

## Why this is a second attempt

I removed these exact five files in #1122, which merged at 10:59Z. #1056 merged at 11:07Z and restored all five, byte-identical. That is not #1056 doing anything unusual — it is the predictable result of fixing a symptom. Removing tracked files does not stop the tool that writes them, and `.gitignore` covers `*.log` but not `*.out`, so the next `git add` at a dirty root picks them up again.

## Evidence

```
git ls-tree -r --name-only 4398c02b | grep '\.out$'
  .refold2-build.out      93 lines
  .refold2-install.out     0 lines
  .refold2-merge.out       8 lines
  .refold2-tc.out          1 line
  .refold2-web.out       115 lines

grep -rn refold2 .        -> no references outside the files themselves
git check-ignore .refold2-build.out  -> not ignored
```

These are the only `.out` files tracked anywhere in the tree, so the pattern costs nothing today. The class is not specific to one tool: history shows the same shape under `.lane/` (`guard-baseline.out`, `mutation-proof.out`, and others).

## Controls run

- **Negative control** — created `.refold3-probe.out` after the change; `git status` does not see it, so a *new* scratch file is genuinely caught rather than only the five named ones being deleted.
- **Positive control** — created `realfile.ts`; still visible to `git status`, so the pattern is not over-broad.
- Confirmed `remotion/out/` is a directory rule and is unaffected.

## Scope

`.gitignore` plus five deletions. No source, no tests, no behaviour. 217 deletions, one insertion.

## Disclosure

I am the release orchestrator and I authored this, as I did #1122. It needs an independent reviewer who is not me. The judgement worth checking is the breadth of `*.out` versus a narrower `.refold*`: I chose the extension because two different tools have now produced this class and because no legitimate `.out` is tracked, but a reviewer may reasonably prefer the narrow form.


davidfarah2003:
BLOCK

Exact head: `fa745764c910a4cc8923235d3e67517caedb890a`

The claimed root-cause fix is absent from the commit. `git diff-tree` shows only five deleted `.refold2-*.out` files; `.gitignore` is not changed. Its blob is identical in the commit, its parent, and current main (`bc2f08b7704f97d4d0c06d04935f0c7aac35b9a1`), where `*.log` is still followed directly by `.DS_Store` and no `*.out` rule exists.

The load-bearing negative control therefore fails:

- create `.refold3-probe.out`
- `git status --porcelain --untracked-files=all` reports `?? .refold3-probe.out`
- `git check-ignore -v .refold3-probe.out` reports no match

The positive control also behaves normally: `realfile.ts` is reported as untracked. This exact head deletes the five symptoms again but does not stop a sixth `.out` transcript from landing, so it does not close the hole described by the PR.

Pattern breadth (`*.out` versus a narrower refold/lane pattern) is not yet gradeable because no ignore pattern landed. For the repair, add the intended rule to `.gitignore`, then rerun both controls. The successor should also be based on current main or refolded cleanly; this head's parent is `4398c02b`, while current main is `f85cfd37`, so its GitHub diff includes unrelated main-side smoke changes even though the commit itself is deletion-only.

Not run: build, typecheck, smoke suites, hosted CI, or live suites. The required repository-state control already fails directly.

davidfarah2003:
## Reviewer found the fix was not in the commit. Corrected at `6c7b3162`.

The review blocked `fa745764` on the grounds that it does not modify `.gitignore` at all: its
`.gitignore` blob equals main's, and a fresh `.refold3-probe.out` is still visible and still
`NOT_IGNORED`. That is correct, and it means the first push was another symptom deletion — precisely
the thing this PR was opened to stop doing.

Cause: `git rm` staged the five deletions, but the `.gitignore` edit was made in the worktree and
never `git add`-ed, so the commit took only what was staged. `git status --short` printed
`" M .gitignore"` — the leading space that means *unstaged* — and I read the `M` and not the space.

The worse half is what happened next. **I ran the controls against my working tree, where the edit
was present, and reported the result as a property of the branch.** They passed. The commit did not
contain the fix. A control that is run against a different artifact than the one under review is not
a weak control, it is a measurement of something else entirely, and it will report green for exactly
as long as nobody checks the right object.

## Fixed

New head **`6c7b3162a88951019e0904f795d662f3f1a088dc`**. A second commit on the same branch adding
the single `.gitignore` line. No force-push and no rebase — `fa745764` stays in history with its own
message, and the new commit's message records that the previous one left the edit unstaged.

Controls re-run against the committed state, and with `git check-ignore` rather than only
`git status`:

```
NEGATIVE   git check-ignore -q .refold3-probe.out    -> ignored
           git status --porcelain .refold3-probe.out -> empty
POSITIVE   git status --porcelain realfile.ts        -> visible
```

Regrading at `6c7b3162` is requested, with nothing above taken as established — the review has now
demonstrated twice that my green is not evidence. The breadth question (`*.out` versus a narrower
`.refold*`) is only now gradeable, because until this commit no pattern had landed.

## Why the reviewer caught it and I did not

They tested the pushed ref. I tested the tree I was standing in. Only one of those can merge.

davidfarah2003:
APPROVE

Exact head: `8174c053f380ba928a214cc1fa975ccb8a4b236c`

The fold is exact and bounded. Parent 1 is the reviewed policy head `6c7b3162a88951019e0904f795d662f3f1a088dc`; parent 2 is current main `f85cfd37b6ccb5b406d10ab30b4b43208c738d70`. The committed tree is byte-for-byte the machine-computed merge tree. Relative to main, the result is exactly `.gitignore` plus the five `.refold2-*.out` deletions: 1 insertion, 217 deletions. The reviewed `.gitignore` blob is unchanged across the fold and no `.refold2-*` path survives.

The recurrence-prevention controls pass against this exact committed head:

- root `.refold3-probe.out` is ignored by `.gitignore:8:*.out`
- nested `tool.out` is ignored by the same rule, covering the historical `.lane/**` producer class
- `realfile.ts` remains visible as untracked
- `remotion/out/review-1125.out` still resolves to the existing `remotion/out/` directory rule
- ordinary `git add` of a future `wanted-fixture.out` refuses with explicit advice to use `-f`; deliberate force-add works, and once tracked its modification is visible

I approve global `*.out` rather than a `.refold*` prefix. Current main's only tracked `.out` files are the five scratch transcripts; this head has zero. Historical unique `.out` paths comprise 36 `.lane/**` command transcripts and two `.refold*` transcripts, with no other category. A refold-only rule would leave an independently demonstrated scratch producer class open. The adjacent `*.log` rule already expresses the same repository-wide scratch-output policy, so an extra comment is not required.

No changeset is needed because no published package surface changes.

Not run: dependency installation, build, typecheck, smoke suites, hosted CI, or live suites. The acceptance behavior here is Git's committed ignore, index, status, force-add and merge semantics, all exercised directly. The worktree is clean and the pull ref was rechecked immediately before this verdict.

PR #1127, opened 2026-08-30
fix(jcode): remove stale credential mirrors

### LINKED ISSUE #850: connector-jcode: removed credentials survive in persistent private-home mirrors
## Summary

The Jcode connector copies recognized provider credentials into a persistent per-seat private home on every launch. When a source credential file is later removed, the next mirror pass leaves the old destination copy in place.

The code comment says no stale copied material survives the next launch, but the absent-source path does not remove the destination.

## Exact path

`extensions/connector-jcode/src/private-state.ts`

- `copyCredentialFile()` returns `false` immediately when `statSync(source)` reports `ENOENT`.
- `mirrorJcodeCredentials()` invokes it for known Jcode, app-config, and external credential paths, but never removes a known destination whose source is now absent.
- The managed home is stable per space/name, so the copy survives later launches.

## Reproduction

Using fake files only, through the exported production function:

1. create a source `auth.json` containing `fake-v1`;
2. call `mirrorJcodeCredentials(destination, sources)`;
3. remove the source `auth.json`;
4. call `mirrorJcodeCredentials(destination, sources)` again;
5. inspect `destination/auth.json`.

Observed:

```json
{"sourceExists":false,"staleDestinationExists":true,"staleValue":"fake-v1"}
```

The existing private-state smoke proves that a present source is refreshed, but has no source-removal cell.

## Impact

Removing or replacing a local login does not guarantee the managed Jcode home stops carrying the prior copied credential material. Depending on provider logout/revocation semantics, a seat may continue using stale authorization; at minimum, the persistent private home retains credential bytes the operator removed from the source location.

This is separate from the MCP-child leak investigation. It is a credential lifecycle and managed-state correctness issue.

## Acceptance

- A launch reconciles the connector-owned credential mirror exactly: each known destination reflects the current source set.
- If a known source is absent, the corresponding connector-owned destination is removed before Jcode starts.
- Cleanup is allowlisted to the connector's exact credential destination inventory; it never recursively deletes arbitrary private-home state, transcripts, MCP config, or files Jcode owns.
- Rotated present credentials still refresh atomically and remain owner-only regular files.
- A red/green smoke covers source present → mirrored, source removed → destination absent on the next launch.
- Negative controls prove unrelated private-home files survive reconciliation and symlink/directory escape protections remain intact.
- No credential contents, environment, or argv are logged.
- Document whether credential mirroring is mandatory and provide an explicit opt-out if managed seats are intended to support credential-free/local-provider operation.

davidfarah2003:
## Summary
- Reproduced on origin/main: throwaway credentials copied into the managed Jcode home, then removed at the source, still survived the next `mirrorJcodeCredentials` pass in all four persistent families.
- Reconcile the allowlisted inventory on every launch, including prior app-config `.env` names and OpenClaw agent directories that now exist only in the managed home.
- Remove the connector-owned destination when the source is absent. Cleanup is allowlisted, refuses a symlinked parent, and leaves unrelated private-home files in place.
- Rotated present credentials still publish through an atomic copy+rename.

## Mirrors
Drove all four families through the production function with disposable files: `jcode-home`, `app-config`, `external-static`, `external-agent`. Remaining fixed-list siblings share `copyCredentialFile` and were reasoned, not extra-driven.

## Family coverage
The smoke now requires unique inventory families to equal driven families (source-derived), not only `driven.length === 4` (fixture-derived). That closes a fifth construction site added only to inventory: the fixture-constant count stayed green, and the source-derived set goes red on `driven families equal unique inventory families`.

Residual that this does not close: a new file added to `JCODE_CREDENTIAL_FILES` or `EXTERNAL_CREDENTIAL_FILES` does not move the family count. Those names are data in the arrays, not branches, and still share `copyCredentialFile`.

Allowlist-orphan (shrinking a fixed list later leaves a dest no longer in the allowlist) is a different defect from #850's absent-source case and is not extra-driven here.

## Verification
- Pre-fix: 4/4 destinations still present after source removal (locations only: `managed/auth.json`, `config/jcode/subscription.env`, `external/.codex/auth.json`, `external/.openclaw/agents/removed-agent/agent/auth.json`).
- Post-fix: 4/4 absent. Smoke asserts `mirrorsChecked === 4` before absence, so a zero-iteration loop cannot go green.
- Live hole: injecting `fifth-family` into inventory at `82e2f96013b8b392e4a9a54fc83b3effb500288a` still passed 13/13 including `=== 4`. After the source-derived check, the same probe failed naming `fifth-family`.
- `pnpm smoke:jcode-private-state` (14 checks)
- `node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/private-state.json` (4/4 killed; the named new mutation is adding a fifth inventory family the fixture does not drive)
- `@cotal-ai/connector-jcode` typecheck and build
- `pnpm check:docs-voice` and `pnpm gen:docsbundle`
- `pnpm changeset status`

## Limits
- Throwaway files only. No real credential value was read, logged, or included.
- Did not run `cotal setup`, `ext seed --reset`, `up`/`down`, live suites, `smoke:ci`, or `pnpm check`.
- Load at intake was 2.72 / 6.39 / 7.71. `cp -al` hit filesystem link-count limits; the worktree used `pnpm install --offline --frozen-lockfile`.
- Related open PR #1078 covers the same issue on an older main. This branch is from current `origin/main` and is not a merge of that PR. Prefer landing this PR and closing #1078 as superseded.

Closes #850.


davidfarah2003:
BLOCK

Exact head: `82e2f96013b8b392e4a9a54fc83b3effb500288a`

The credential fix itself is substantiated, but this head resurrects files deleted by current main and removes the just-landed prevention rule.

`git diff --diff-filter=D --name-only $(git merge-base HEAD origin/main)..origin/main` names all five:

- `.refold2-build.out`
- `.refold2-install.out`
- `.refold2-merge.out`
- `.refold2-tc.out`
- `.refold2-web.out`

They all exist at this head. Main also added `*.out` to `.gitignore`; this PR's main-relative diff deletes that line. Thus merging this exact head would undo merged PR #1125, restore the five scratch transcripts, and reopen the recurrence hole. Refold onto current main and prove the five remain absent and the `.gitignore` blob/rule survives.

The requested credential evidence is otherwise green:

- On exact pre-fix parent `4398c02b`, I drove the shipped `mirrorJcodeCredentials` public function and reproduced 4/4 persistent survivors after removing their sources: jcode-home, app-config, external-static, and external-agent.
- The production source universe is four families, not four total destinations. It contains 7 fixed Jcode-home entries, 20 fixed external-static entries, dynamic `.env` entries, and two entries per dynamic OpenClaw agent. The suite's four-case sentinel correctly proves family discovery, but not every fixed sibling.
- I independently drove the full source-derived inventory with two app configs and two agents: 33 entries total (`7 + 2 + 20 + 4`), all 33 created, then zero survived source deletion. This closes the author's reasoned-sibling gap for this review.
- Fixture anchors are unique. Coverage/config ran before proof. Baseline passed 13/13. The absent-source mutation was KILLED red-and-named on `removed credentials are absent from every persistent mirror on the next launch`; the other two mutations also killed their named cells.
- Full build and full typecheck passed.

PR #1078 is a duplicate implementation, not complementary work. It changes the same implementation, smoke, mutation fixture, docs, generated bundle, and changeset, and a direct merge conflicts in those paths. Its current version has several wording/control-flow refinements over #1127, but the same underlying inventory-and-remove design. Whichever PR is selected should supersede/close the other before merge; they must not be merged sequentially as independent fixes.

Not run: a real operator Jcode home or authenticated provider launch, Windows execution, hosted CI, full `pnpm smoke:ci`, `pnpm check`, or live-suffixed suites. All credential acceptance was exercised with throwaway filesystem homes through the shipped public reconciliation function.

davidfarah2003:
CORRECTION / WITHDRAWAL of the resurrection blocker in my preceding BLOCK: I ran only the first resurrection signal before verdict, not the required computed-merge outcome. Against current main, `git merge-tree --write-tree 82e2f960 origin/main` produces a tree with zero `.out` / `.refold2-*` paths and the reviewed `*.out` ignore blob `802a74ca...` intact. The five branch-tree files do **not** survive the actual merge, so they are not a blocker. I withdraw that finding completely. The head is moving for the independently authorized source-derived family-census repair; I am standing down and will grade the successor. Do not use the prior BLOCK as a current disposition.

davidfarah2003:
APPROVE

Exact head: `188b6df769f2777f69922525c8e69450b4cee73e`

This supersedes my withdrawn `82e2f960` disposition. The prior resurrection finding was incorrect and has already been durably withdrawn: the computed merge onto current main contains zero `.out` paths and preserves main's `*.out` ignore.

The credential defect and repair are now fully evidenced:

- On exact pre-fix parent `4398c02b`, the shipped `mirrorJcodeCredentials` function produced all four representative mirrors and all 4/4 survived removal of their sources: jcode-home, app-config, external-static, and external-agent.
- The production source census is four construction families, with 7 fixed Jcode-home names, 20 fixed external-static paths, dynamic app-config `.env` names, and two credential names per dynamic OpenClaw agent.
- I independently drove the whole production-derived inventory with two dynamic app entries and two agents: 33 entries (`7 + 2 + 20 + 4`), all 33 created, then zero survived source deletion. This checks the fixed-list siblings that the committed smoke represents by family.
- The successor no longer trusts fixture constant 4 alone. It derives unique families from the production inventory and compares them to the driven fixture families. A new mutation adds a fifth production family, and it is KILLED red-and-named on `driven families equal unique inventory families`.
- Fixture inventory reports four unique anchors. Coverage/config ran before proof and reports 4/14 cells. Baseline passes 14/14. All four mutations kill their named cells, including absent-source cleanup and the family-census control.
- Full build and full typecheck passed at the production-identical parent; the successor changes only the smoke and mutation fixture.
- Computed merge onto current main has no `.out` paths, keeps `*.out`, and changes only the intended changeset, docs/generated bundle, private-state implementation, smoke, and mutation fixture.

PR #1078 is the same fix lineage and should close as superseded rather than merge independently; the consolidation finding has been routed.

Non-blocking residual: the branch tree still visibly contains inherited `.refold2-*` files even though the computed merge deletes them. Removing them from the branch would reduce review ambiguity, but it is not required for merge correctness.

Named gap: shrinking a fixed allowlist later can orphan a destination that is no longer present in the inventory. That is distinct from #850's absent-source reconciliation and remains outside this PR.

Not run: a real operator Jcode home or authenticated provider launch, Windows execution, hosted CI, full `pnpm smoke:ci`, `pnpm check`, or live-suffixed suites. Credential acceptance used throwaway filesystem homes through the shipped public reconciliation function. The pull ref and clean review tree were rechecked immediately before this verdict.

davidfarah2003:
Security probe at exact head `188b6df769f2777f69922525c8e69450b4cee73e`. This is not a re-grade of the functional APPROVE. Sandbox only (`/tmp/rev1127-sandbox`), shipped helpers copied from this SHA, no stack start.

## Q4 destination symlink — held
`removeCredentialMirror` on a symlink at the destination unlinked the link and left the target file intact (contents unchanged). `copyCredentialFile` `renameSync` replaced the destination link with a regular file and did not write through to the target. A directory at the destination throws and is not removed.

## Q2 string-prefix containment — held on this host, one named gap
`assertRelative` rejects `..` and absolute `destinationRelative` (`../escape`, `a/../../escape`, `/etc/passwd`). `path.resolve` is lexical and does not follow a symlinked `home`: the prefix check matches the symlink path, not the realpath. This Linux tmpfs is case-sensitive.

NAMED GAP: case-insensitive filesystem prefix behavior (Windows / macOS) was not probed.

## Q1 parent-walk TOCTOU — window is real, live race not landed
If a parent is already a symlink at check time, the shipped function throws `refusing symlinked Jcode credential mirror parent` and does not delete outside `home`.

The walk is `lstatSync` on each parent, then later `rmSync(destination)` with no re-lstat. Replaying that sequence (walk, then replace the last parent with a symlink to an outside directory, then `rmSync(destination)`) unlinked a file *outside* `home`. Sibling files in that outside directory were left intact. The guard is therefore not atomic.

I did not land a live two-process race against the uninstrumented function. The window is two consecutive sync calls, so winning it needs a concurrent replacer, not a same-thread interleaving.

## Q3 temp-file mode — 0644 then chmod, parent already 0700
`copyFileSync(source, temp, COPYFILE_EXCL)` inherits the source mode. A 0644 source produced temp mode `644`, then `hardenPrivate` set `600`, then `renameSync` published `600`. End state of shipped `copyCredentialFile` is `0600`. `mirrorParent` has already set parent dirs to `0700` before the copy, so another uid cannot traverse to the temp.

Residual: create is not 0600-from-birth. Not a world-readable leak on this POSIX layout, given the 0700 parents.

Standing down. Not merging.

davidfarah2003:
Second-vendor security probe of PR #1127 at exact `188b6df769f2777f69922525c8e69450b4cee73e`.

**Verdict: BLOCK** on parent-walk TOCTOU. The other three scoped attacks did not delete outside the private home on this host. I do not clear merge. I did not merge. Re-read `headRefOid` immediately before this comment: still `188b6df769f2777f69922525c8e69450b4cee73e`.

Method: owned detached worktree at that SHA. Live calls into `removeCredentialMirror` / `credentialDestination` / `copyCredentialFile` / `mirrorJcodeCredentials` from the shipped `extensions/connector-jcode/src/private-state.ts` (export keywords only, no logic edit). No `cotal up`. Host: Linux ext4, case-sensitive, umask `0002`, Node v24.20.0.

### 1. Parent-walk TOCTOU — **hit**

`removeCredentialMirror` `lstatSync`s each parent, then `lstatSync`s the destination, then `rmSync(destination)`. After the last parent `lstat` returned a real directory, I replaced that parent with a directory symlink to an outside tree that already held `auth.json` (`OUTSIDE-SECRET`, mode 0600). The call returned `removed: true` with no throw. The outside file was gone.

What changed between the check and `rmSync`: the last parent inode. Intermediate path walk on `rmSync`/`lstat` follows that parent symlink, so the namesake file in the symlink target is what gets unlinked. The parent-walk is check-then-use, not an `O_NOFOLLOW` open of a directory fd held through the remove.

A destination-only swap (replace the leaf with a symlink to an outside file after the parent walk) did **not** delete the target: `rmSync` unlinked the link (`before.isLink: true`), outside file remained.

### 2. Prefix containment — **holds on this host, one named gap**

- `../escape.json` threw `unsafe Jcode credential mirror path: ../escape.json` (`assertRelative`).
- Symlinked home: `path.resolve` is lexical here (it does not `realpath`). `resolve(destination).startsWith(resolve(home) + sep)` passed for `join(linkHome, "auth.json")`. That is containment of the path *string*, not of the realpath. Writes still land in the real target of `home`. `privateAgentHome` also refuses a symlinked managed path, which this helper does not re-check.
- Case-folding: **unverified**. This disk is ext4 case-sensitive (`CaseHome` and `casehome` are distinct). I will not infer a pass for macOS/Windows.

### 3. Temp-file mode — **0600 after harden, 0644 window if the source is 0644**

`copyFileSync` then `hardenPrivate` then `rename`. With a 0600 source, the temp was already 0600 at copy. With a 0644 source, copy created `auth.json.<hex>.tmp` at **0644**, then chmod **0600**. Final dest was 0600. Other-user read of that temp is still blocked by the 0700 parent `mirrorParent`/`privateDirectory` creates. Residual: anyone who can already traverse that private directory can read the temp for the copy-to-chmod interval.

### 4. Destination-symlink semantics — **unlinks the link, keeps the target**

Destination was a symlink to outside `real-cred.json` (`KEEP-ME`). `removeCredentialMirror` returned true, dest gone, target still `KEEP-ME`. Same through public `mirrorJcodeCredentials` with an absent source: dest link gone, outside `KEEP-PUBLIC` kept. Node `rmSync` on a symlink is unlink, not `rm` of the referent.

### Named gaps

- Case-folding prefix escape: not reachable on this filesystem.
- Cross-uid other-user temp read: not demonstrated. Parent dirs were 0700.
- I did not start a mesh or any `-live` stack.

BLOCK is only the parent-walk TOCTOU: a post-check directory symlink on a walked parent makes stale-mirror removal delete outside `home`.

davidfarah2003:
Duplicate of the credential-mirror staleness fixed by #850's fix branch, now merged via #1150 (1cd9e6b0ca66600e6dcb94181abb821fcf90cc0f). Fresh mirrors are written on every connect; the stale-mirror path this issue describes no longer exists.

PR #1128, opened 2026-08-30
fix(jcode): preserve startup refusal diagnostics

### LINKED ISSUE #985: connector-jcode: three connector-owned refusals collapse to "Jcode host startup failed (unknown)", and the message that would name them is never persisted
**Environment:** @cotal-ai/connector-jcode 0.33.9; jcode 0.81.1; Linux. Reproduced on a shared box with ~13 live seats.

**Summary.** When a jcode seat is spawned with a model spec the connector itself refuses, the operator sees exactly one line:

The advice in that line is a dead end. The private jcode logs contain nothing about the refusal — the connector refused *before or independently of* anything jcode logs, and the connector's own message is discarded rather than written anywhere. Three distinct, connector-owned failure conditions are indistinguishable from each other and from a genuinely unknown host crash.

**Root cause.** `src/host-main.ts:15-20`:

```ts
function startupFailureCode(error: unknown): string {
  if (error instanceof Error && error.message.startsWith("jcode connector: project MCP configuration")) return "project_mcp_config";
  const code = typeof error === "object" && error !== null && "code" in error ? (error as { code?: unknown }).code : undefined;
  return typeof code === "string" && STARTUP_FAILURE_CODES.has(code) ? code : "unknown";
}
```

`STARTUP_FAILURE_CODES` (host-main.ts:4) holds eight codes. Any error whose `.code` is absent or outside that set renders as the literal string `"unknown"`, and only the code — never the message — reaches stderr at host-main.ts:45.

Three throw sites in `src/host.ts` construct plain `new Error(...)` with no `.code`, so all three land on the fallback:

1. host.ts:550-553 — provider-prefix rejection: `jcode connector: model "cliproxy/gpt-5.6-sol" carries a provider prefix, but the Harness API expects a bare model id — pass "gpt-5.6-sol" (the "cliproxy" provider is selected by configuration, not by the model id)`
2. host.ts:555 — `await client.setModel(sessionId, config.model)` propagates the Harness API's rejection (e.g. `invalid_request`) unwrapped and uncoded.
3. host.ts:600-603 — post-readiness model mismatch: `jcode connector: requested model X but the Harness API reports Y — refusing a mislabelled mesh seat`

Note that host-main.ts:18 already hand-patches exactly one message by prefix-match, for the project-MCP preflight. That line is proof the problem was recognised; the other three refusals were simply never given the same treatment.

**Why the scrubbing is correct, and what the defect actually is.** The scrubbing is deliberate and should stay. Per #828, the SDK appends child-process stderr to startup errors, and that child stderr can carry provider credentials and arbitrary local paths. Passing arbitrary `error.message` to the operator is a real disclosure risk and the current code is right to refuse it.

The defect is narrower: the blanket applies to the connector's OWN messages too. These three strings are authored in this repo, are bounded in length, and interpolate only (a) the model id the caller supplied, (b) a substring of it, and (c) the model id the Harness API echoed back. None of them can carry provider stderr. They are being discarded by a rule aimed at something else.

**Steps to reproduce.**
1. Give an agent persona a model spec carrying a provider prefix, e.g. `model: cliproxy/gpt-5.6-sol`.
2. Spawn the seat.
3. Observe the fatal line.

**Expected:** a code that names which refusal fired — enough to act on without source access.

**Actual:** `Jcode host startup failed (unknown); inspect the private Jcode logs.` The suggested logs do not mention it. In our case two seats spawned four hours apart with this spec both died this way, while a third seat in the same lane with a bare `glm-5.3` spec spawned fine — and it took a source read to learn that the discriminating variable was the prefix, not the model. Bare `grok-4.6` and bare `deepseek-v4-flash` seats run fine on the same box, so "that model is unsupported" was a natural and wrong first conclusion.

**Proposed fix.**
(A) Give the three connector-owned throws real codes and add them to `STARTUP_FAILURE_CODES`: host.ts:550-553 → `model_prefix_rejected`; host.ts:555 → `model_refused` (wrap the setModel rejection); host.ts:600-603 → `model_mismatch`. The cleanest shape is a small `ConnectorError extends Error` carrying `code`, so `startupFailureCode()` keeps its allow-list semantics and no message is ever passed through. This preserves the #828 rationale exactly: still code-only on the wire, still no arbitrary message.

(B) Persist the connector's own stderr per seat. Today the fatal line is written to stderr and nothing captures it: the managed home under `.cotal/jcode/hack-<seat>-<hex>/` holds jcode's log but not the connector's, so once the spawning terminal scrolls the only record of a failed spawn is gone. Writing the connector's stderr to `<home>/logs/connector-<ts>.log` would make failures diagnosable after the fact.

**Caveat, stated honestly.** `bareModelId()` has a single call site, host.ts:549, which runs well after the managed home is created at host.ts:230. Managed homes are never reaped (every `rmSync(home)` in the repo is smoke-test cleanup). Yet both of our prefixed seats left no managed home at all. So the prefix is the experimentally discriminating variable, but host.ts:550-553 is NOT proven to be the throw that killed them — something upstream of home creation may reject a prefixed spec first. Fix (A) is worth doing on its own merits regardless, since it makes exactly this question answerable from a log line instead of a source read; but this report does not claim :550 as the confirmed killer.

davidfarah2003:
## Summary
- classify prefixed, rejected, and mismatched Jcode model startup failures with distinct allow-listed codes
- persist connector-owned stderr in each managed seat's owner-only connector log
- keep arbitrary Harness API and provider text scrubbed
- document the diagnostic path and add a patch changeset

## Verification
- reproduced the built-host prefix refusal before the fix: managed session created, zero set_model requests, fatal rendered as unknown, no connector log persisted
- pnpm smoke:jcode-host (53 checks)
- pnpm --filter @cotal-ai/connector-jcode typecheck
- pnpm --filter @cotal-ai/connector-jcode build
- pnpm changeset status
- node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/jcode-model-refusals.json (4/4 killed by named assertions)

Closes #985

Supersedes #1029 — CI never minted runs for the synchronize push there (only CodeQL appeared); fresh PR from the same branch, tip 4e895122.


davidfarah2003:
CI triage for 4e895122 — every red classifies as inherited or cured, none from this branch (its own gates: 1328 anchors / 0 dead, three jcode suites green, per the evidence comment above):

- `smoke (shard 0/4)` (`manager-reconcile-ownership`, oracle ×4), `smoke (shard 3/4)` (`secret-store-seam`, oracle), `smoke (shard 2/4)` (`adoption-passive-preflight`, oracle — 4th member, noted on the issue): reproduce on pure main — #1137.
- `smoke (shard 1/4)` (`gate-inventory`: ungated web-proxy-order): cured on main by #1140.
- `unit`: stale generated docs bundle vs the folded docs (pre-#1119 wording) — cured by this refold's regeneration.

Refolded to 27547879: one merge of current main (includes #1139's tree-hygiene guard), residue purged, bundle regenerated; no branch-side content changed (`git log --no-merges 4e895122..27547879 --not origin/main` is empty), so the classification carries over. Merging.

PR #1129, opened 2026-08-30
fix(spawn): honor connector readiness deadlines

### LINKED ISSUE #979: jcode spawn caller wait (40s) is shorter than the connector's own readiness budget (180s) it must outlive

davidfarah2003:
Closes #979.

## Summary

- include the connector-selected readiness budget in spawn and manifest-launch acceptances
- make synchronous goal followers treat their request deadline as a floor and wait through the accepted readiness budget plus delivery margin
- preserve the budget in the durable acceptance floor so concurrent same-goal retries remain safe before the goal spec write lands
- document the acceptance contract and add a patch changeset

## Proof

- reproduced on unmodified main with a real broker, manager, pty child, and presence join: the caller returned `deadline-exceeded` at 40,385ms, then the same seat joined healthy at 43,807ms inside its declared 50,000ms readiness budget
- reran the same scenario after the fix: the caller returned success at 43,720ms instead of timing out at the old 40s cliff
- `pnpm build`
- `pnpm smoke:goal-follow`
- `pnpm smoke:goal-epoch-fence`
- `pnpm smoke:manager-service-invoke`
- `pnpm smoke:spawn-args`
- `pnpm smoke:launch-parity`
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/spawn-follow-readiness.json` with 2/2 mutations killed on their named assertions
- `pnpm changeset status`

## Limits

- did not run the forbidden live mesh lifecycle suites or commands on the shared host
- the live proof uses a deterministic Jcode-shaped test connector, not a provider-backed Jcode model download, so it proves the timeout ordering and real manager/child/presence path without depending on provider latency

Supersedes #1032 — the synchronize push there minted no CI workflows (Code Quality only); fresh PR from the same branch so full runs mint.


davidfarah2003:
Independent review at exact head `c2af4328fdf7d2eaa87feea641ac3277b748dc3a` (matches `refs/pull/1129/head` at post time). Merge base `a697576dd1733ec4bff744f9447eecdfe90dc852`. Own detached worktree, no stack start, no merge.

This is the first look at this PR. Functional shape is a real fix for the #979 wait mismatch, with named coverage gaps on the semantic half and on the 180s number itself.

## Does the caller outlive the connector budget, or only the 30s backstop?

The follow path outlives the *accepted* budget, not a hardcoded 30s.

- Manager selects `connector.readinessTimeoutMs ?? this.readinessTimeoutMs` and calls `onReadinessWindow` *before* `onAccepted`. The acceptance, goal spec, and reconcile-index floor all persist that value.
- `submitAndFollowGoal` then waits `max(callerDeadline, readinessDeadlineMs + 10_000)` after the accept returns.
- For jcode (`readinessTimeoutMs: 180_000`) that is `max(40_000, 190_000) = 190_000` ms on the follow. `SPAWN_TIMEOUT_MS` / CLI `START_TIMEOUT_MS` stay 40s *floors*; they are no longer the follow ceiling once the field is present.

Probed, not only read: a stub follower with `readinessDeadlineMs: 180_000` and a 50ms caller floor accepted a terminal at 250ms. The same terminal timed out when the field was omitted (fallback to the caller floor). A silent wait with `readinessDeadlineMs: 400` elapsed **10404ms** and the timeout text named `10400ms`, so the 10s margin is real, not comment-only.

NAMED GAP: no live jcode spawn was run. The 180s path is the generic field plus jcode's declaration. Mutation 2 below proves the manager copies a connector override (45s stub) rather than `this.readinessTimeoutMs`, not jcode's 180s literal.

Older acceptances that omit the field still die at the 40s floor, which is the stated compatibility path.

## Core leak

No connector number is compiled into `packages/core`. `180_000` does not appear under `packages/`. Core consumes `readinessDeadlineMs` off the acceptance (already a goal-spec field) and applies a generic 10s delivery margin taken from the old 40s-vs-30s spawn gap. `endpoint-action.ts` only duplicates that bound onto the acceptance floor so a retry can serve it before the spec exists. That is a genuine generalization, not jcode's budget pushed down a layer.

## `spawn-follow-readiness.json` executes

Ran `node scripts/mutation-proof.mjs --config bin/smoke/mutations/spawn-follow-readiness.json` at this SHA.

- Baseline green (goal-follow 15 marks, manager-service-invoke 12 marks).
- Mutation 1 KILLED on `a terminal inside the accepted readiness budget outlives the shorter generic caller wait`.
- Mutation 2 KILLED on `invoke spawn accepts the goal with the exact connector readiness budget a follower must outlive`.

The fixture is git-tracked. I did not check whether a CI job list names this file beyond `git ls-files '*mutations/*.json'` discovery.

## What `goal-follow.smoke.ts` does not cover

Cell E is 20 added lines and a 100ms caller vs 400ms readiness vs 250ms terminal. Because follow = `400 + 10_000`, the 10s margin is what makes the wait long; 400ms alone would also have beaten 250ms. It does **not** pin 40s vs 180s.

It never feeds an `uncertain` terminal. The follower *would* surface that as `error.code === "uncertain"` (not `failed`, not `deadline-exceeded`) if one arrived — probed on a stub — but the shipped smoke does not. Cell B still checks that a silent wait is `deadline-exceeded` / "timeout on the WAIT". That is the caller-timeout half of #979, not the expiry-is-uncertain half.

`launch-parity.smoke.ts` still only asserts `SPAWN_TIMEOUT_MS > READINESS_TIMEOUT_MS` (40 > 30). That old relation is unchanged and does not watch jcode's override.

`spawn-action.smoke.ts` asserts `readinessDeadlineMs === 2_000` (the test-shortened manager default). Connector-selection is the manager-service-invoke cell, not M1.

Not merging.

davidfarah2003 (commented):
Terminal verdict at exact head `c2af4328fdf7d2eaa87feea641ac3277b748dc3a` (matches `refs/pull/1129/head` at post time). Follow-up to the first look: expiry semantics, not coverage.

**Disposition: residual, not BLOCK.** The shipped `#979` contract (an elapsed readiness window is `uncertain`, not `failed`) is implemented. The missing `goal-follow` cell is a coverage gap, not a wrong result.

## Expiry is `uncertain` in the implementation

Read, then driven, not inferred from the missing cell.

1. `awaitReadiness` races join (`ok`), process exit (`failed`), and the timer. The timer resolves `{ ok: false, uncertain: true }` with the manager diagnosis ("launch status uncertain", "Inspect with", "do not stop it solely because this bounded wait elapsed").
2. The spawn path maps that arm to `onOutcome({ kind: "uncertain", data: { reason } })`. Process exit takes the `failed` arm. They do not share a code.
3. `onOutcome` commits `settleGoalUncertain` → `commitGoalResult` with cause `readiness` → `commitTerminalFact(..., "uncertain", { reason, readinessDeadlineMs })`. A `failed` outcome commits `state: "failed"`.
4. `submitAndFollowGoal` copies `terminal.state` onto `error.code`. An `uncertain` terminal is therefore `error.code === "uncertain"`, not `failed`, not `deadline-exceeded`. `deadline-exceeded` is only the silent-wait arm (no terminal arrived).

## What was driven (no live jcode spawn)

Follow stub against built `packages/core` (9 checks):

- A manager-shaped terminal `{ state: "uncertain", data: { reason } }` surfaces as `uncertain`, carrying the inspect-not-stop sentence. Not `failed`. Not `deadline-exceeded`.
- The same terminal after the accepted window (400ms) but inside the follow margin (arrived at 450ms, caller floor 50ms) still returned `uncertain` at 451ms. The extra wait did not reclassify it as a wait timeout.
- A `{ state: "failed" }` terminal stays `failed`. The distinction is not collapsed at the follower.

Manager path, local NATS + pty, not jcode: `pnpm smoke:manager-spawn-action` at this SHA, 37 passed.

- `M3 process exit -> failed`
- `M3 readiness window elapsed -> uncertain (bounded, never success)` on a `stuck` agent against the 2s shortened window

That is the owner of the deadline committing the contract the follower then copies. It is not a live jcode 180s boot.

## Residuals (named, not holds)

- **`goal-follow` cell E still does not prove expiry-is-uncertain, and still does not pin 40s vs 180s.** The follower mapping is correct and untested in that file. Future drift risk, not a wrong result today.
- **No live jcode spawn.** 180s is jcode's declaration plus the generic field. The expiry classification itself is agent-agnostic and was driven on a stuck pty child.
- Silent wait with `readinessDeadlineMs: 400` still elapsed ~10404ms on the first look. That is the follow margin, not expiry classification.
- `launch-parity.smoke.ts` still only asserts `SPAWN_TIMEOUT_MS > READINESS_TIMEOUT_MS` (40 > 30).
- `onReadinessWindow` runs before `onAccepted`. Spawn-action M1 showed the acceptance carrying the selected budget (`2000`). No wrong outcome measured from that order.

Not merging.

davidfarah2003:
CI classification on c2af4328 (effective-terminal, 7 reds), per the standing merge-gating policy — every red is by-signature either cured-on-main or reproduced on pure main with the identical cell:

| red | signature | class |
|---|---|---|
| smoke (shard 0/4) | "without the liveness oracle" | #1137, red on pure main |
| smoke (shard 1/4) | `smoke:manager-reconcile-startup` — "the delivery daemon is not reachable on the ctl.delivery-admin rail", ✗ startup sweep after terminalizing | #1137 — reproduced on pure main 32a6d160 locally just now: same suite, same cell, 3 passed/1 failed |
| smoke (shard 2/4) | "without the liveness oracle" | #1137, red on pure main |
| smoke (shard 3/4) | "without the liveness oracle" | #1137, red on pure main |
| Windows / required + windows-ok | `env.PATH` cell (launchEnv preserves `Path` casing) | #1141 class, cured on main by #1145 — base 5cf6253c predates the cure |
| ci-ok | aggregate of the above | derivative |

Nothing in the fail set is introduced by this branch; the oracle reds await #1143. Merging.

PR #1132, opened 2026-08-30
feat(core): declare whether a connector can announce a tool-list change

### LINKED ISSUE #1004: The advertised tool list depends on the connection, and three connectors cannot announce a change
The set of `cotal_*` tools an agent sees is a function of its `AgentConfig`. That is deliberate and documented. It also means that any future ability to change a session's mesh at runtime changes the advertised tool list, and only half the connectors have a way to say so.

Raising it now, while nothing changes connections at runtime and this is therefore latent rather than broken.

## The tool list depends on the config

`extensions/connector-core/src/tool-specs.ts:508` gates its output:

```ts
const canSpawn = !isAuthed(config) || (config.capabilities?.includes("spawn") ?? false);
```

so `cotal_spawn` and `cotal_persona` render or vanish per config, and `config.allowSubscribe` bounds `cotal_join` further down. The comment there states the intent plainly: an agent should only see these "when it can actually use them, instead of discovering the denial by trying".

This is already asserted, so it is not an inference. `extensions/connector-core/smoke/orientation.smoke.ts:62-67`:

```ts
const noSpawn = cotalToolSpecs(cfg({ creds: "CREDS", capabilities: [] })).map((s) => s.name);
assert.ok(!noSpawn.includes("cotal_spawn"), "no spawn cap ⇒ cotal_spawn hidden");
const withSpawn = cotalToolSpecs(cfg({ creds: "CREDS", capabilities: ["spawn"] })).map((s) => s.name);
assert.ok(withSpawn.includes("cotal_spawn") && withSpawn.includes("cotal_persona"), "spawn cap ⇒ both shown");
```

And meshes genuinely differ: `MeshEntry.mode` is `"auth" | "open" | "user"` (`packages/workspace/src/mesh-registry.ts:33`), so two entries in one registry produce different tool sets.

## Only three of six connectors could announce a change

**Can, nearly for free.** Claude (`registerCotalTools`, `connector-core/src/tools.ts:28`), Codex (`connector-codex/src/mcp.ts:104,122`) and jcode (`connector-jcode/src/mcp.ts:158`) all register through the SDK's high-level `McpServer.registerTool`. SDK 1.29.0 already advertises `tools.listChanged: true` on their behalf and already fires `sendToolListChanged()` from `enable()`/`disable()`/`remove()`. The Claude connector also already sends custom notifications through `server.server.notification(...)` (`connector-claude-code/src/mcp.ts:216`), so the channel is in use and proven.

**Cannot, structurally.** OpenCode builds a plain `Record<string, ToolDefinition>` handed to the host once (`connector-opencode/src/tools.ts:22`), pi registers in a boot loop (`extensions/pi/src/tools.ts:76`), and Hermes renders static descriptors (`connector-hermes/src/tool-schema.ts:34`). None is MCP, none has a list-changed notification, and none has a re-registration path. The host asks for a tool list once.

Grep confirms the repo sends none of it today: no match for `listChanged` or `sendToolListChanged` anywhere in `extensions`, `implementations` or `packages`. Today that is correct, because the list never changes.

## Why it is worth recording before it bites

A session that changed mesh would leave the three native connectors advertising a stale surface, and the failure is quiet: the tool is still listed, the call is still made, and the denial arrives at the wire. That is the outcome `tool-specs.ts` says the gating exists to prevent.

Restricting a session to one mesh at a time does not avoid it. Sequential disconnect and reconnect inside one session changes the list exactly as much as holding two would.

## Not proposed here

Which way to resolve it is a design decision with at least three defensible answers: re-render and accept that native hosts will not notice; refuse a connection change on connectors that cannot announce it; or hold the tool surface constant across meshes and let the wire denial be the boundary, which reverses the stated intent above. That last one is a product decision rather than an implementation detail.

No change requested in this issue. Recording the constraint and the connector split so it is costed before someone builds against it, rather than found afterwards.

davidfarah2003:
Closes #1004.

The advertised `cotal_*` list is a function of the session's mesh config. A connection change therefore changes the list. Only some hosts can tell the session the list changed. Native hosts that take a tool map once cannot.

David's ruling: nothing should be claude-only; connectors declare, consumers above the boundary never special-case harness names and never branch on transport. Default-deny: undeclared is refuse, not a silent no-op. "Hold the surface constant" and "re-render and accept native hosts will not notice" are both rejected. A host that advertises a stale tool surface after a mesh switch is the silent degradation AGENTS.md forbids.

## Split (established before design)

All three "cannot announce" connectors are genuine harness limitations, not unimplemented:

- opencode: `buildCotalTools` returns a Record handed to the host once.
- pi: `registerCotalTools` is a boot loop over `pi.registerTool`.
- hermes: descriptors written to `COTAL_TOOLS_FILE`; Python plugin reads at `register(ctx)`.

Can announce today: claude, codex, jcode (MCP SDK 1.29.0). That correlation with MCP is a measurement, not the contract. `supportsToolListAnnounce` is a declaration. A future non-MCP host may set it true and be believed; an MCP host may set it false.

## Latent, so this is a guard, not a wiring change

Nothing changes connections at runtime today, so the list never changes, so never calling `sendToolListChanged` is currently CORRECT. That sentence stops a future reader from filing this as a bug in the three MCP connectors and from bolting on an announce call with nothing to announce. The failure this guards is not cosmetic: when a mesh switch lands, spawn would stay listed, the call would still be made, and the denial would land at the wire, which is the outcome `tool-specs.ts` says the gating exists to prevent.

## What changed

- `Connector.supportsToolListAnnounce?: boolean` — default-deny. Declaration, not mechanism.
- `refuseUnannouncedToolListChange(connector)` — name-blind. The refusal names no connector. Absent/false throws. True returns.
- Set true on claude, codex, jcode only.
- `smoke:tool-list-announce` discriminates with connectors the CLI has never heard of. The ruling cell: the refusal message contains no connector name at all.
- Docs: `docs/authoring-a-connector.md`.
- Appended to `ci-suites.txt` (shard-stability STABLE 421 to 422).

## Proved

- Split from source on origin/main plus MCP SDK 1.29.0. Load: source read, no broker.
- `tsx packages/core/smoke/tool-list-announce.smoke.ts` 7/0.
- Mutation-proof `packages/core/smoke/mutations/tool-list-announce.json`: N1 undeclared no-op, N2 name-special-case the unknown announcer, N3 put claude in the refusal — all KILLED on the named cells.
- `tsx bin/smoke/mutation-fixtures.smoke.ts` 0 dead/ambiguous/prose.
- `tsx bin/smoke/shard-stability.ts origin/main HEAD` STABLE=0.

## Not proved

- No runtime mesh switch exists to exercise the helper end to end. The helper is the contract a future op must call.
- `sendToolListChanged` is not wired. Nothing to fire it.
- `pnpm smoke:gate-inventory` could not run in this worktree (no typescript module). Script+chain membership checked by reading the files.
- Did not invent an announce protocol for OpenCode/pi/Hermes.
- w983 (#1036) was DMed that the field is landing; setup picker need not consume it unless it starts switching meshes. No reply yet.

HEAD: f264e90722b643be2d39653cb5400122ffd6a544

Supersedes #1046 — the synchronize push there minted no CI workflows (Code Quality only); fresh PR from the same branch so full runs mint.


davidfarah2003:
CI triage for 3b3648d4 — every red classifies as inherited or cured, none from this branch:

- `smoke (shard 0/4)` (`manager-reconcile-ownership`, oracle ×4), `smoke (shard 2/4)` (`adoption-passive-preflight`, oracle ×2), `smoke (shard 3/4)` (`secret-store-seam`, oracle): reproduce on pure main — #1137, fix up in #1143.
- `Windows / required` + `windows-ok`: the connector-claude-code PATH cell — #1141, fix up in #1142.
- `ci-ok`: aggregate.

Fold gates on 3b3648d4 were green (build, bin typecheck, web-channel-authority 112, docs regen, fragments incl. this branch's tool-list-announce registration). Merging.

PR #1133, opened 2026-08-30
fix(jcode): remove stale credential mirrors

### LINKED ISSUE #850: connector-jcode: removed credentials survive in persistent private-home mirrors

davidfarah2003:
## Summary
- reconcile the full allowlisted Jcode credential mirror inventory on every launch
- remove exact managed destinations when source credentials disappear, including dynamic app-config and OpenClaw-agent mirrors
- atomically refresh credentials that remain present
- preserve unrelated private-home files and refuse cleanup through symlinked parent directories
- document mandatory mirroring and the absence of a credential-free opt-out

## Verification
- pre-fix production-function repro with disposable material: 4/4 mirror families copied, sources removed, 4/4 destination files still present
- pnpm smoke:jcode-private-state (13 checks, including exact mirror count and 0 survivors)
- node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/private-state.json (3/3 killed by named assertions)
- node scripts/mutation-coverage.mjs extensions/connector-jcode/smoke/mutations/private-state.json
- @cotal-ai/connector-jcode typecheck and build
- pnpm check:docsbundle
- pnpm changeset status

## Limits
- Used only throwaway credential files under disposable homes. No real credential value was read, logged, or included in artifacts.
- Drove all four mirror families: Jcode-home fixed paths, dynamic app-config .env files, fixed external login paths, and dynamic OpenClaw agent credentials.
- The existing private-state suite was extended, so no new CI suite registration was needed.
- Load was 9.93 at intake, 15.88 during reproduction, and 20.18 at final gates. These checks are not timing-sensitive. A partial cp -al dependency copy hit filesystem link-count limits; the worktree node_modules was deleted and rebuilt independently before final build/docs gates.
- Did not run setup, seed reset, stack lifecycle commands, smoke:ci, pnpm check, or any live suite.

Closes #850.

Supersedes #1078 — the synchronize push there minted no CI workflows (Code Quality only); fresh PR from the same branch so full runs mint.


davidfarah2003:
CI never minted for the refolded tip 5d49c6d9 (Code Quality only, past the grace window — the known partial-minting failure). Closing and reopening as a fresh PR from the same branch; the tip already carries the residue purge and bundle regeneration.

PR #1134, opened 2026-08-30
feat(workspace,cli,manager)!: per-space runtime pid and log namespace

### LINKED ISSUE #1095: workspace: per-space pid/log names for delivery and the manager (the {space} seam is unused where it matters)
At @cotal-ai/workspace 0.36.0 the runtime pid/log namespace is segmented for exactly one of the three daemons. `auth-service.{space}.pid` is per-space; `manager.pid` / `manager.log` / `delivery.pid` / `delivery.log` are root-level constants. One workspace root therefore hosts exactly one `cotal deliver` and one `cotal supervise` **by filename**.

Evidence, at the shipped versions:
- `MANAGER_PIDFILE = "manager.pid"` is a bare constant with no `{space}` — `workspace/dist/pid.d.ts:53`.
- The manager daemon **overwrites** it rather than claiming it: `recordManagerPid` does an unconditional `writeFileSync` into `<root>/.cotal` — `@cotal-ai/manager@0.30.1 dist/commands.js:44-47`. A second space's manager on the same root silently takes over the record and the first runs unlisted — the failure the file's own comment documents at `:26-28`.
- The CLI's process registry is a compile-time literal: `baseProcesses` at `@cotal-ai/cli@0.30.1 dist/index.js:430-464` — `pidFile: "manager.pid"` (:436), `"delivery.pid"` (:444). The one templated entry is `"auth-service.{space}.pid"` (:452).
- The `{space}` expansion is real — "expands to the injective hex space key … case-safe" (`workspace/dist/local-process.d.ts:19-20`) — but nothing in cli, delivery or manager dist/ reads a pid-file or log-file override. The mechanism exists in the type and is unused by the two components that need it.

Why it matters: a multi-tenant root (several spaces on one broker, one workspace root — the layout the broker/account split in #257 exists for) cannot run per-space manager/delivery daemons at N ≥ 2, and hosting compositions that check `RESERVED_COTAL_CHILDREN` refuse to provision while those names are root-scoped. There is no seam to pass an explicit path through, so an embedding cannot work around it — this needs the names templated upstream and released.

Requirement (checkable): `RESERVED_COTAL_CHILDREN` stops carrying root-scoped `manager.pid` / `manager.log` / `delivery.pid` / `delivery.log`; delivery and supervise write per-space names through the existing `{space}` seam; the CLI registry rows for both become templated like auth-service's. Existing single-space meshes must keep working across the upgrade — the compat story is part of the fix, not an afterthought.

Note: #1069 (pidfile start-identity) rewrites adjacent plumbing; whichever lands second rebases. A fix branch is already in flight for this issue.

davidfarah2003:
Closes #1095.

Delivery and the manager now write per-space pid/log names through the existing `{space}` LocalProcess seam — the templated form auth-service already used — and the CLI's process registry rows are templated the same way. `RESERVED_COTAL_CHILDREN` no longer carries the root-scoped `manager.pid` / `manager.log` / `delivery.pid` / `delivery.log` names, which is the predicate embeddings check before provisioning a second tenant onto one root.

Two adjacent defects found and fixed red-first along the way: a folder's runtime space is now resolved from its own records rather than assumed (`d7e866be`), and a bare `stop` no longer misses an open-mode folder's manager (`aa15e465`).

**Compat:** existing single-space meshes keep working across the upgrade — the read paths (status/stop/up liveness checks) find both the legacy root-scoped names and the new per-space names during the transition; there is no degraded write mode, new daemons write only the new names.

**Verification (lane run at 9bb3c7c5):**
- `pnpm -r build` and `pnpm typecheck` clean across all packages
- `pnpm test` (all packages) and `pnpm smoke`: 22 passed, 0 failed
- **Mutation proof: 14/14 killed**, each on a named cell; the live-path cells rebuild `@cotal-ai/cli` dist before and after the mutation so the mutant is proven loaded
- `changeset status`: workspace/cli/manager all minor (0.x breaking), lockstep
- The one red in the lane run was pre-existing on its base (gate-inventory, suites registered by #1052 after the branch point) and clears with this merge of main

The three new suites are registered as `ci-suites.d` fragments per #1052. The merge commit also regenerates the docs bundle from merged sources.

**Composition note:** #1069 (pidfile start-identity) rewrites adjacent plumbing; whichever lands second rebases.

Supersedes #1103 — the synchronize push there minted no CI workflows (Code Quality only); fresh PR from the same branch so full runs mint.


davidfarah2003:
CI triage for 1b99ea87 — every red classifies as inherited or cured, none from this branch:

- `smoke (shard 0/4)` (`manager-reconcile-ownership`, oracle ×4), `smoke (shard 2/4)` (`adoption-passive-preflight`, oracle), `smoke (shard 3/4)` (`secret-store-seam`, oracle): reproduce on pure main — #1137 (fix in progress).

The fold (1b99ea87) ran the standard gates green (build, bin typecheck, web-channel-authority 112, docs regen, fragment validation) with the committed refold logs purged. Merging.

davidfarah2003:
Refolded to c17b19d1 (the #1091 merge landed adjacent core changes): one merge of current main, no branch-side content changed (`git log --no-merges 1b99ea87..c17b19d1 --not origin/main` is empty), so the classification in the previous comment carries over. Fold gates green. Merging.

davidfarah2003:
RETROACTIVE BLOCK for exact head `c17b19d1d14bdd81bb6474bcb2a8fdc2009b0979`.

This breaking head should not have landed as-is.

## Blockers

1. **The shipped CLI directs operators to a log path this change no longer writes.** The declared contract and the writer now use `.cotal/manager.<spaceKey>.log` (`MANAGER_LOGFILE = "manager.{space}.log"`; for space `alpha`, `/project/.cotal/manager.616c706861.log`). But three landed messages still say `.cotal/manager.log`:
   - `implementations/cli/src/commands/spawn-manifest.ts:239`
   - `implementations/cli/src/commands/spawn-manifest.ts:250`
   - `implementations/cli/src/commands/up.ts:1896`

   These are precisely readiness/lease failure and launch diagnostics, when the operator needs the log. On a fresh post-change root that old path does not exist. This is an additional user-visible break beyond the declared filename migration and makes the recovery advice false.

2. **The committed mutation evidence is not valid under the repository's mutation-coverage contract.** `node scripts/mutation-coverage.mjs bin/smoke/mutations/segmented-runtime-namespace.json` refuses before running any suite because all 14 mutations omit the required `cell` field. Therefore the landed guard set cannot produce its claimed per-cell coverage evidence. I independently ran the 13 non-live mutations after excluding S14 and all 13 killed red-and-named, but that does not repair the committed metadata or validate S14.

## What passed

- Absolute `pnpm typecheck` passed.
- Populated pre-upgrade state is covered and behaved correctly in safe tests: legacy live manager records are found and stopped; dead records are reclaimed; live legacy records are refused; canonical plus legacy is ambiguous; open-mode record-derived space resolution works.
- Shipped-path suites passed: workspace namespace 19/0, CLI runtime records 25/0, manager writer 12/0.
- Thirteen non-live source mutations killed red-and-named with clean restore.
- Safe regressions passed: space segmentation, system rotation, component health 17 checks, delivery teardown 3/0, clean 69 checks, down-target 12 checks, gate inventory, docs voice, and changeset status.

## Unverified by constraint

I did not run S14 or any changed suite named `-live`/`:live`, including the actual live `up`/bare-`down` transition. The PR's only mutation for `down`'s folder-context change is S14 and remains unverified. The safe CLI suite reaches the same production readers/stoppers with real child processes, but not a real stack lifecycle.

One unrelated existing `smoke:multi-space` assertion failed on its doctor-auth path after the namespace-relevant multi-space lifecycle checks had passed; I did not attribute that failure to this PR.

Issue #1131, opened 2026-08-30
ten open PRs would each re-land scratch build logs on main; .gitignore cannot stop them because the files are already tracked


davidfarah2003:
Ten open PRs would each put scratch build logs back on main. This is not a stale-branch
inheritance that a fold fixes: each branch **committed the files itself**, so git treats the
merge as an addition and the branch wins.

Measured against main `6d47befd` by computing the merge tree for every open PR, then
positive-controlled with a real `git merge --no-commit` into a scratch worktree.

| PR | files | lines | set |
| --- | --- | --- | --- |
| #1032 | 5 | 209 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1039 | 5 | 209 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1042 | 5 | 214 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1046 | 5 | 209 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1053 | 5 | 214 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1073 | 10 | 411 | `.refold-bin-path .refold-build .refold-install .refold-man` |
| #1078 | 9 | 307 | `.refold-build .refold-install .refold-merge .refold-tc .re` |
| #1091 | 5 | 212 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1103 | 5 | 209 | `.refold2-build .refold2-install .refold2-merge .refold2-tc` |
| #1128 | 9 | 310 | `.refold-build .refold-install .refold-jcode-tc .refold-tc ` |

Total: **2504 lines** of install, build, typecheck and suite transcripts.

## Why `.gitignore` does not fix this

#1125 added `*.out` to `.gitignore`, which closes the hole for *untracked* files. **An ignore
pattern has no effect on a file that is already tracked**, and each of these branches ran `git add`
on its scratch output before that rule existed. The pattern is present in every merge result above
and every one of them still carries the files.

## Why a fold does not fix it either

Seven of these PRs (#1032, #1039, #1042, #1046, #1053, #1091, #1103) carry the same five
`.refold2-*` files. Those are the files removed from main in #1122 and again in #1125. They are not
inherited from an old base:

```
merge base of #1032 with main:   c52cdb3d
  branch ADDED .refold2-* since the base:   5
[...]

PR #1135, opened 2026-08-30

davidfarah2003:
Re-mint of #921: after that PR's branch history rewrite, its pull_request events stopped producing CI runs entirely (close/reopen and an empty-commit push both minted only CodeQL), so it could not be validated there. Same branch, same tree. Review record: #921 carries a LANDING-QUEUE REVIEW: APPROVE-quality at this tree, including the ci-suites tail check.

Supersedes #1085 — the synchronize push there minted no CI workflows (Code Quality only); fresh PR from the same branch so full runs mint.


davidfarah2003:
Two notes from the landing queue:

1. **CI never minted here.** The tip 71331c9a is shared with #921/#1085, and only Code Quality / CodeQL workflows exist for it — no CI, Windows, or Docs runs. The retrigger commit didn't help because the sha predates it elsewhere. A genuinely new sha (any real commit, e.g. a fold of main) should mint; both a fresh PR and same-sha retriggers have failed on this tip today.

2. **A fold of main now conflicts semantically in `implementations/cli/src/commands/setup.ts`**: this branch's persona-root/seed-destination rework vs main's registry-derived connector-skills delegation (#1131-era, `reconcileConnectorSkills`). Both sides restructure the same functions, so I'm leaving the reconciliation to this branch's author rather than hand-merging it in the queue — the two changes need to be composed deliberately, not textually.

davidfarah2003:
CI note, measured. This PR mints only the two CodeQL checks because it is CONFLICTING with main (mergeStateStatus DIRTY). GitHub creates zero pull_request runs when it cannot build the merge ref; the CodeQL pair is push-event, so its presence hides the cause. No new sha or branch fixes this while the conflict stands. The same class killed three PRs in the #1150 lineage before it was diagnosed.

Conflicting paths, from `git merge-tree` against current main:

- `extensions/connector-core/src/docs-bundle.generated.ts`
- `implementations/cli/src/commands/setup.ts`
- `implementations/cli/src/commands/status.ts`

Cure: merge `origin/main` into the branch, resolve the two command files, regenerate the docs bundle with its generator rather than hand-merging it, then push. Full CI mints on the resolved head.

PR #1139, opened 2026-08-30
test(ci): gate scratch command output out of the tree

davidfarah2003:
## The problem a fourth deletion would not solve

Scratch command transcripts from a fold routine have reached `main` **four times today**:

```
10:59  #1122  removed five
11:07  #1056  restored five
11:30  #1125  removed five, added *.out to .gitignore
12:05  #1042  restored five
~12:1x #1138  removed them again
```

Every removal was correct and none of them stopped the next one. `.gitignore`'s `*.out` only blocks a file that is **untracked**; a branch that committed its transcripts before that rule keeps them tracked, and merging that branch puts them back with nothing in the PR diff to show it.

**Seven open PRs would each still land theirs — 48 files between them**, measured against current main:

| PR | files it would land |
| --- | --- |
| #1073 | 10 |
| #1078 | 9 |
| #1128 | 9 |
| #1032, #1039, #1046, #1103 | 5 each |

Removing them from main a fifth time changes none of that. A gate does: the PR carrying them goes red instead of landing them silently.

## What this adds

`bin/smoke/tree-hygiene.smoke.ts`, registered as a one-file-per-suite fragment under `ci-suites.d/` (never the frozen legacy list), plus the `smoke:tree-hygiene` script and the removal of the five files that were tracked when I branched.

## Red first, then green — the gate catches the real defect

Run against main **while the five files were still tracked**, it failed and named all five:

```
✗ FAIL: no scratch command output is tracked (scanned 1787 tracked files) [
  '.refold2-build.out', '.refold2-install.out', '.refold2-merge.out',
  '.refold2-tc.out', '.refold2-web.out' ]
TREE HYGIENE SMOKE FAILED (4 passed, 1 failed)   exit=1
```

After removal, on the same tree:

```
TREE HYGIENE SMOKE OK (5 passed, 0 failed)   exit=0
```

That is the gate reddening on the actual condition that shipped four times, not on a synthetic fixture.

## How it avoids being a check that cannot fail

- **It asserts the size of the set it scanned.** A loop that finds nothing to check reports the same green as one that proved everything, so the cell states `scanned N tracked files` and a first cell requires that listing to be a plausible universe (>500 entries, containing `package.json`). A broken read now fails instead of passing.
- **The pattern is positive-controlled against literals**, not against the tree, so that cell keeps discriminating after the tree is clean — it checks the shapes both observed producer families write (`.refold2-build.out`, `.refold-tc.out`, `.lane/guard-baseline.out`) and that ordinary sources are spared.
- **It also asserts `.gitignore` still carries `*.out`**, because removing the files without that rule is what let them return twice.
- A count sentinel fails if any cell stops existing.

## Gates run

| gate | result |
| --- | --- |
| `smoke:tree-hygiene` | 5/5, exit 0 (and red-first, above) |
| `smoke:ci-fragments` | 8/8 |
| `smoke:ci-suites` (shard stability) | 25/25, no existing suite moved |
| `tsc --noEmit` | no errors from this suite |

The suite uses only Node builtins, so it runs in a fresh checkout without `node_modules` — deliberate, since a gate that needs a prebuild is green in CI and red for a contributor.

**Named gap:** `smoke:gate-inventory` fails on this branch, and also fails identically at `origin/main` without this change — it reports `smoke:web-proxy-order` as ungated. Pre-existing and unrelated; this suite is not mentioned in its output and is properly gated via the fragment. I did not run `pnpm smoke:ci` or `pnpm check`.

## Scope

This gates the tree. It does not stop the routine that writes the files — **nothing in this repository writes them** (no script, no package target; the only in-tree mention is in a private submodule), so there is no writer here to patch. See #1131. The seven branches above still need their own `git rm`; this change makes that visible rather than optional.

## Disclosure

I am the release orchestrator and I wrote this, as I did #1122 and #1125. It needs an independent reviewer who is not me. Worth attacking: whether `\.out$` is the right predicate (it matches `.gitignore`'s rule, which is the argument for it), and whether the plausible-universe cell's `>500` threshold is a meaningful floor or theatre.


davidfarah2003:
BLOCK

Exact head: `9fb5f7bae05126cc35b039e2b82ec8961118aa74`

The gate catches today's defect, but its claimed pattern positive control is not bound to the predicate that grades the real tree.

`SCRATCH` is tested against literals in one cell, while the actual scan is a separate expression:

```ts
const offenders = tracked.filter((p) => SCRATCH.test(p));
```

I replaced only that real scan with `tracked.filter(() => false)`, leaving `SCRATCH` and the literal control untouched. On the historical real-offender tree `57969be79`, which tracks the five `.refold2-*.out` files, the suite exited 0 with 5/5. The literal cell still passed, the offender cell reported a clean tree, and the exact defect this gate exists to stop became invisible. The same mutation also survives on the clean reviewed head.

This directly answers the review question: the literal-based control lets the real predicate drift. Bind both the synthetic literals and the tracked-file scan through one function, for example `scratchPaths(paths)`, so a mutation that makes the real filter return nothing also makes the literal positive control red. Then prove that mutation on the historical offender tree or an equivalent committed red-first control.

Everything else held:

- Exact reviewed gate against historical main `57969be79` exits 1 and names all five real tracked offenders; exact head passes 5/5.
- Empty and one-file tracked universes fail the plausible-universe cell. Current tracked count is 1788 and includes `package.json`; the >500 floor is a conservative loud sanity bound, not a proof by itself, and is acceptable with the anchor.
- Changing `SCRATCH` from `.out` to `.log` fails the literal pattern cell. Removing a cell fails the exact-count sentinel.
- A fresh public checkout with no `node_modules` runs the gate directly with `node` and passes; the file uses only Node builtins. The package-script path also passes.
- Historical `.out` inventory is 36 `.lane/**` command transcripts plus 2 `.refold*` transcripts and no other category, so `.out$` matches the demonstrated class and the existing `*.out` policy.
- `package.json` is exactly one insertion. Fragment filename is exactly `sha256(smoke:tree-hygiene)`. CI fragments pass 8/8, CI suites 25/25, and shard stability reports 458→459 with zero pre-existing movement.
- Direct TypeScript check passes.
- `smoke:gate-inventory` fails identically on this head and current main only for inherited `smoke:web-proxy-order`; this suite is registered and not named in that failure.

Not run: full build, full repository typecheck, hosted CI, full `pnpm smoke:ci`, `pnpm check`, or live-suffixed suites. The new standalone file was directly typechecked and exercised through both direct Node and package-script entry points.

davidfarah2003:
APPROVE

Exact head: `1a432964babe121885f5252659387a2779cc5c68`

The repair closes the blocker. The successor is a direct child of blocked `9fb5f7ba` and changes only `bin/smoke/tree-hygiene.smoke.ts`. One `scratchPaths(paths)` now applies the `.out` rule for both the synthetic control and the real tracked-file scan; the control asserts the exact returned set.

I reran the binding from both sides:

- Clean exact head passes 5/5.
- Original mutation, changing the shared function to `paths.filter(() => false)`, now exits 1 on the named rule-control cell with `{ controlHits: [] }`; it also exits 1 on the historical five-offender tree. This is the exact mutation that previously passed 5/5.
- A fresh independent mutation, changing the shared function to retain only `.md` paths, exits 1 on both the rule-control cell and real offender scan. Restore returns to 5/5.
- The unmutated repaired gate against historical main `57969be79` still exits 1 and names all five real `.refold2-*.out` offenders. The repair preserved original discrimination.

All prior exact-parent evidence transfers because the delta touches only this shared-function binding: fresh public checkout without `node_modules`, empty/small universe failures, regex drift, count sentinel, historical breadth, one-line package scope, correct fragment hash, fragments 8/8, suites 25/25, zero shard movement, and direct TypeScript check.

`smoke:gate-inventory` remains an inherited failure on this branch because it predates main's #1140 registration for `smoke:web-proxy-order`; it failed identically on the reviewed parent and its then-current main, and does not name `smoke:tree-hygiene`.

Not run at this successor: full build, full repository typecheck, hosted CI, full `pnpm smoke:ci`, `pnpm check`, or live-suffixed suites. The repaired standalone gate was exercised directly on clean and historical trees with two binding mutations. The pull ref and clean tree were rechecked immediately before this verdict.

davidfarah2003:
Record of a blocking finding and its repair, so the trail is on the PR rather than only in the commit message.

**The block (w979, at `9fb5f7ba`).** The positive control applied the scratch pattern itself, while the tracked-file scan applied it separately:

```
check("pattern matches ...", [".refold-tc.out"].every((p) => SCRATCH.test(p)) && ...);
const offenders = tracked.filter((p) => SCRATCH.test(p));
```

Two independent applications of one rule. Mutating only the scan to `tracked.filter(() => false)` left the control green, so the suite reported **5 passed, 0 failed on a tree carrying five offenders**. The control was grading the regex, not the gate that used it. A gate whose control is not attached to the gate is not a gate.

**The repair (`1a432964`).** One `scratchPaths(paths)` is now the single application point; the control and the real scan both route through it, and the control asserts the returned set rather than a boolean, so it fails on a wrong answer and not only on a wrong pattern.

**Proofs, executed:**

- Clean tree: green, 5 passed / 0 failed.
- The named mutation (`scratchPaths` body to `paths.filter(() => false)`): control cell fails with `{ controlHits: [] }`, suite 4/1, exit 1. That is the case that previously passed 5/5.
- Restore: green, exit 0.
- Real red-first re-run through the shared function against historical tree `57969be7`, which tracks all five logs: red, names all five, exit 1. The binding did not cost the original discrimination.

**Inherited red, not from this branch.** `smoke:gate-inventory` fails at this head because `smoke:web-proxy-order` is a `package.json` script registered in no `ci-suites.d/` fragment and absent from `ci-suites.txt`. Measured independently by two parties. It was fixed on main by #1140; this branch predates that fix, so the red persists here and should not be attributed to this change.

Issue #1137, opened 2026-08-30
manager smokes red on main: verify-evict hard-requires the ctl.delivery-admin liveness oracle no fixture runs


davidfarah2003:
Pure main at bde11b2b, local run (macOS, Node 24): three suites fail out of the box.

- `smoke:manager-reconcile-ownership` — 3 passed, 3 failed ("this instance's OWN dead-but-active row is reconciled (terminalized)", "a LEGACY row (pre-3b-2, no owner) is reconciled", "the OWN row's principal IS in the refusal index after its terminal")
- `smoke:manager-reconcile-startup` — 3 passed, 1 failed ("startup sweep completes after the manager has served")
- `smoke:secret-store-seam` — 7 passed, 2 failed ("the teardown really happened (the credential file is gone)", "the credential was DELETED through the injected store")

All failing cells share one mechanism, printed repeatedly before each failure:

> the delivery daemon is not reachable on the ctl.delivery-admin rail (no responders: 'cotal.….ctl.delivery-admin.local.…'); a restart on an auth mesh cannot verify-evict the superseded serve family principal "local.…" without the liveness oracle. Start the delivery daemon (`cotal up` runs it) and retry — eviction is never skipped (SPEC 13.1)

These suites boot the real Manager against a bare broker with no delivery daemon, so every retirement/teardown path that now insists on the liveness oracle refuses, and the cells that assert the eviction/teardown happened go red.

Impact: every open PR inherits these through the smoke shards (observed across independent branches — the shard stops at its first failure so different branches surface different members of the set). CI on main itself hasn't completed a run recently, which is why this wasn't caught at merge time.

Reproduce: checkout main, `pnpm install && pnpm -r build`, then any of the three suites above.

A first-parent bisect between the last known-green window and bde11b2b is running; I'll follow up with the introducing merge once it lands.

davidfarah2003:
Bisect result (first-parent, local build+run of `smoke:manager-reconcile-ownership` at each point):

- `a257db81` (parent of the #1097 merge, = #1112 merge): **GREEN**
- `5881a6fe` (merge of #1097, fix/817-manager-orphan-succession): **RED** — the same three cells as on current main

So the #1097 merge is the introducing point. That fits the mechanism: the failing cells are all in the orphan verify-evict path, and the refusal fires because the smoke fixtures boot the Manager against a bare broker with no delivery daemon, so the ctl.delivery-admin liveness oracle has no responder.

[...]

PR #1142, opened 2026-08-30
test(claude): accept Windows Path casing

### LINKED ISSUE #1141: Windows lane: connector-claude-code env-crossing test fails its PATH cell on every branch
The `Windows / required` lane fails one cell of the connector-claude-code env-crossing test on every tree that reaches it:

```
extensions/connector-claude-code test:   ✗ FAIL: PATH present so the seat still launches
... SUITE COMPLETE: 43 passed, 1 failed
```

Observed identically on four independent branches today (PRs #1042, #1053, #1073, #1129 — job logs each show exactly this one cell red, all sibling env-crossing cells green), so it is not caused by any of those changes; it rides in with their common base. The POSIX arm of the same test passes everywhere.

The cell asserts that after the spawn-env scrub, PATH still crosses so the seat can launch. A likely shape for a Windows-only failure of exactly this cell is environment-variable name casing (`Path` vs `PATH` on win32) somewhere in the scrub/allowlist path, but I have not verified that — treat it as a hypothesis, the four-branch census is the evidence.

No completed main-lane run is available to point at (the Actions queue has been jammed), but with four independent branches showing the identical single-cell failure the class is base-side for merge-gating purposes.

davidfarah2003:
## Mechanism

`launchEnv` already matches its allowlist case-insensitively and preserves the source key casing to avoid Windows case-duplicate environment entries. On Windows it validly returns `Path`, while the Claude launch-env smoke asserted only `env.PATH`. The suite now installs a `Path`-cased fixture on every platform and finds the forwarded path key case-insensitively. The shipped scrub is unchanged; undeclared variables remain withheld.

## Gates

- `pnpm --filter @cotal-ai/connector-claude-code test` (44 passed)
- `pnpm -r build`
- `pnpm -r typecheck`
- `pnpm --dir bin typecheck`
- Mutation proof: reverting the case-insensitive assertion under the `Path` fixture regressed the named cell on Linux with `43 passed, 1 failed`; restoring it returned 44 passing cells.

## Windows

Windows cannot run locally in this environment. The PR Windows lane is the proving gate.

Fixes #1141

### DIFF
--- a/extensions/connector-claude-code/smoke/launch-env.smoke.ts
+++ b/extensions/connector-claude-code/smoke/launch-env.smoke.ts
@@ -39,6 +39,13 @@ for (const k of CLAUDE_PROVIDER_KEYS) process.env[k] = `smoke-${k}`;
 for (const k of HOST_MARKERS) process.env[k] = `parent-${k}`;
 for (const k of UNRELATED) process.env[k] = `parent-${k}`;
 
+// Windows environment names are case-insensitive but preserve their source casing. Model the
+// common `Path` spelling so this suite exercises the same scrub path on every platform.
+const path = process.env.PATH ?? process.env.Path;
+delete process.env.PATH;
+delete process.env.Path;
+if (path !== undefined) process.env.Path = path;
+
 const env = claudeConnector.buildLaunch({ space: "smoke", name: "claude-1" } as never).env ?? {};
 
 check(
@@ -73,7 +80,12 @@ for (const k of UNRELATED) {
   check(`${k} was withheld (not a declared auth var)`, !(k in env));
 }
 
-check("PATH present so the seat still launches", typeof env.PATH === "string" && env.PATH.length > 0);
+const pathKey = Object.keys(env).find((key) => key.toLowerCase() === "path");
+check(
+  "PATH present so the seat still launches",
+  pathKey !== undefined && typeof env[pathKey] === "string" && env[pathKey].length > 0,
+  pathKey,
+);
 
 const opted = claudeConnector.buildLaunch({
   space: "smoke",


davidfarah2003:
BLOCK

Reviewed exact head `6f237c0003940be689099e1a10040c4d7b8cc220` against merge base `2494bc83eeece797abdbdda429ff00a86be41b83`.

The scoped smoke passes 44/44, and restoring the old PATH-only assertion under the new `Path` fixture kills red-and-named on `PATH present so the seat still launches` at 43/1. The public connector path also forwards both source spellings correctly when exercised directly: `PATH` remains `PATH`, and `Path` remains `Path`, each exactly once.

One blocker remains. The new fixture unconditionally deletes uppercase `PATH` and installs only `Path` on every platform, so no platform now tests the Linux spelling. I mutated `launchEnv` to ignore only uppercase `PATH` while continuing to forward `Path`; `smoke:claude-launch-env` still passed all 44 cells. A same-file positive control that ignored every path spelling killed the named PATH cell, so mutation-proof reported the uppercase-only regression as `SURVIVED`, not unobserved. The test trades the Windows blind spot for a Linux one.

Please exercise both source spellings, for example with a table-driven fixture that runs `buildLaunch` once with `PATH` and once with `Path`, and asserts the source casing/value is preserved exactly once in each result.

The process-wide deletion does not create a separate blocker: no later cell or connector implementation reads `process.env.PATH` directly, and the later `buildLaunch` call intentionally sees the same Path-cased fixture.

Windows execution remains an external CI acceptance gap, as the PR states; this review only proves the cross-platform fixture behavior available here.

davidfarah2003:
REPAIR SHAPE

A two-row fixture is sufficient. `launchEnv` has one case-agnostic path for both spellings: it indexes every source key once by `k.toLowerCase()`, then `copy("PATH")` looks up that lowercase entry and emits the source key's original casing. There is no PATH-versus-Path branch.

Exercise `buildLaunch` separately with only `PATH` present and only `Path` present. For each row, assert that exactly one case-insensitive path key is forwarded, its key casing equals the source spelling, and its value is unchanged. Do not globally replace `PATH` with `Path` for all cells. This directly closes both the Windows and Linux blind spots.

davidfarah2003:
A successor implementing the repair shape named above is open as #1145.

It follows the two-row form from the REPAIR SHAPE comment rather than substituting one spelling for the other: both `PATH` and `Path` run as rows on every platform, and each row asserts that exactly one path key is forwarded, that it keeps its source casing, and that its value is unchanged.

Measured against `extensions/connector-core/dist/launch.js`, the artifact the suite actually loads, with a positive control confirming a mutation there reaches the suite before any result was read:

| fixture | drop `PATH` | drop `Path` |
| --- | --- | --- |
| main today | killed, rc=1 | survived, rc=0 |
| #1145 | killed, rc=1 | killed, rc=1 |

The surviving cell is the defect this PR set out to fix. The blocking finding here was that the proposed fixture moves that survival to the other column rather than removing it.

Nothing in this PR is wasted: the mechanism section is accurate, and the blocking review plus its repair shape are what #1145 is built on. Suggest closing this as superseded once #1145 has an independent verdict at its exact head. Leaving it open until then, so the finding stays visible rather than being closed ahead of its replacement landing.

davidfarah2003:
Superseded by #1145, merged as `4260010fdb4c7872c9525f20362ce5dc63a9749c`.

Closing this rather than repairing it in place, because the successor is already reviewed and landed. Nothing here was wasted: the mechanism section of this PR is accurate, and the blocking review plus the repair shape posted on it are what #1145 was built from. The two-row form, and the requirement that each row assert exactly one forwarded path key with its source casing and unchanged value, came from this thread.

The finding that made the difference is worth restating for the record, since it is the reason the successor exists: a mutation breaking only the uppercase `PATH` survived this branch 44/44 while an all-path positive control killed. That established, by measurement rather than inspection, that substituting `Path` for `PATH` in the fixture moves the blind spot instead of closing it. #1145 keeps both spellings and kills both mutations.

#1141 is fixed by #1145.

davidfarah2003:
The Windows / required lane — the proving gate for this fix — completed **success** on 6f237c00: the env-crossing suite's 44 cells including the PATH cell pass on the real win32 runner, confirming the mechanism (the suite asserted `env.PATH` where `launchEnv` correctly preserves `Path` casing). No failures anywhere on the PR so far; the still-pending Linux shards can only surface the known pure-main #1137 reds, which this one-file test fix cannot influence. Merging — this closes the #1141 class for every branch.

davidfarah2003:
For the record: the mechanism diagnosis here (suite asserted `env.PATH` where `launchEnv` preserves `Path` casing on win32) was confirmed by this branch's Windows lane passing, and main's 37405bf8 landed the stronger both-spellings version — hence the close. Branch deleted.

PR #1149, opened 2026-08-30
fix(hermes): inspect uv at manager boot

### LINKED ISSUE #1144: Hermes boot inspection checks the wrong executable, failing closed on the supported setup and open on a broken one
The manager boot inspection added in #1073 misdeclares the shipped Hermes harness, and it is wrong in both directions.

`extensions/connector-hermes/src/extension.ts` declares `requires: ["hermes"]`. The shipped launch executes `uv run --project ... hermes gateway run`, and the public docs give `uv` on PATH as the prerequisite. So the thing inspected is not the thing the launch needs.

## Reproduced on the landed tree

Both cases were exercised against the shipped manager, started twice with an installed cached Hermes connector. This is the real boot path reading installed extension metadata, not a hand-built connector input.

- PATH containing only an executable `uv`: Hermes is classified `unavailable` with `hermes harness needs hermes on PATH - not found`. This is the documented, supported prerequisite shape, and it is refused.
- PATH containing only an executable `hermes`: Hermes is classified `available`, although the shipped launcher fails immediately because `uv` is absent.

A false negative on the supported setup and a false positive on a broken one. That defeats the fail-closed boot availability contract the PR introduced, for a shipped connector.

## Repair

Declare and inspect the executable the launch actually requires. Ship an entry-point smoke driving the real boot path, plus a named mutation proof covering both directions: the supported `uv` case must not be refused, and the `hermes`-only case must not be admitted. One direction alone leaves the other live, and the current behaviour proves both are reachable.

## Provenance

Found by a retroactive exact-head review of #1073 at `9b161990fb110d6aeff73ec9f25d28ae509c8b3b`, the second parent of merge `2494bc83eeece797abdbdda429ff00a86be41b83`. That head merged without any review verdict at the head that landed; the approval on the PR bound an earlier head, and the head-continuity proof supplied at merge time spans only `f85bada2..9b161990`, so the interval containing this code was never graded by anyone. Full verdict, including the green controls and the explicitly unrun paths: https://github.com/Cotal-AI/Cotal/pull/1073#issuecomment-5468724461

Present on main as of `fe3b455a2` (`extension.ts:38`).

davidfarah2003:
## Summary

- inspect `uv`, the executable used by the shipped Hermes launcher, during manager boot
- pass the manager-resolved `uv` path through to the launcher
- add a shipped-metadata manager boot smoke for both uv-only and hermes-only PATHs
- register the smoke in CI and prove both directional regressions with named mutations

## Validation

- `pnpm build`
- `pnpm -r typecheck`
- `pnpm smoke:hermes-boot-availability`
- `pnpm smoke:manager-harness-boot`
- `pnpm --filter @cotal-ai/connector-hermes test`
- `node scripts/mutation-proof.mjs --config implementations/manager/smoke/mutations/hermes-boot-availability.json`
- `pnpm smoke:ci-fragments`
- `pnpm check:docsbundle`
- `pnpm check:shard-stability origin/main HEAD`
- `pnpm changeset status`

Closes #1144


davidfarah2003:
Superseded by #1151, and the duplication is an orchestration error rather than anything wrong with this change.

Issue #1144 was assigned here and then, half an hour later, offered again in a general "take anything open" message. A stand-down went out but raced work already under way. The two correctives were opened three minutes apart, independently, by authors who did not know of each other.

## They converged

Both changes make the same core repair: `requires: ["uv"]`, resolve `uv` at manager boot, export it, and spawn the resolved path rather than a bare `uv`. Two uncoordinated implementations arriving at the same fix is worth recording, because it is stronger evidence the repair is correct than either one reviewed on its own would be.

## Why #1151 is the base

- It extracts `hermesUvCommand()` into `binary.ts`, so the resolved-uv fallback has one application point instead of an inline expression at the spawn site.
- `COTAL_HERMES_UV_BIN` is namespaced to this connector; `COTAL_UV_BIN` here is generic enough to collide with another connector resolving its own `uv`.
- Four mutations across two fixtures rather than two.
- It carries a comment explaining that `uv` provisions the project-pinned `hermes`, so a separately installed `hermes` is neither required nor sufficient to launch the seat. That is the sentence that stops this drifting back.

## What this PR does better, and which should be folded in

This change drops the read of `COTAL_HERMES_BIN` at `launch.ts:167`. That variable is written only at `extension.ts:71` and read only at that spawn site, and both PRs redirect the write to the uv variable - so after either change nothing populates it and the read always falls through to the literal. Removing it is the coherent finish; #1151 leaves a read whose writer is gone.

The author of this PR is reviewing #1151 and has been asked to raise exactly that. Having independently built a fix for the same defect makes them well placed to notice every point where the two solutions differ.

Closing this in favour of #1151. The work here is not discarded: the convergence is evidence, and the vestigial-read removal is a real improvement that should land.

PR #1152, opened 2026-08-30
fix(cli): point manager errors at the space log

davidfarah2003:
## Summary

- derive all three operator-facing manager log hints from the same per-space path helper the detached-manager writer uses
- add a five-cell guard and four mutations covering writer/display alignment plus each shipped message
- add the required `cell` field to all 14 segmented-runtime mutations so the committed fixture can execute

## Reproduction

Before the fix, the shipped messages in `spawn-manifest.ts` and `up.ts` all named `.cotal/manager.log`, while the writer uses `MANAGER_LOGFILE = "manager.{space}.log"`.

The segmented-runtime fixture contained 14 mutations and 14 missing `cell` fields, so mutation coverage could not grade the committed evidence.

## Validation

- `pnpm smoke:manager-log-guidance` — 5/5
- manager-log mutation proof — 4/4 killed, red-and-named
- segmented-runtime mutation proof — 13/13 permitted mutations killed, red-and-named
- `pnpm smoke:mutation-fixtures` — both fixtures present and unique
- `pnpm smoke:ci-fragments`
- `pnpm smoke:gate-inventory`
- `pnpm typecheck`
- `pnpm changeset status`

## Limit

S14 uses `smoke:up-manifest:live`. It was not executed because lifecycle and `:live` suites are prohibited in this environment. Its required `cell` field and unique anchor are validated by fixture inventory; the other 13 mutations were executed.

### DIFF
--- a/.changeset/manager-log-guidance.md
+++ b/.changeset/manager-log-guidance.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/cli": patch
+---
+
+Point manager startup and failure guidance at the per-space logfile the detached manager actually writes.
--- a/bin/smoke/ci-suites.d/8a08a728a4b08f61c633ddd8597de300934fbfd1800ab1fe1ffb9adda556f402.txt
+++ b/bin/smoke/ci-suites.d/8a08a728a4b08f61c633ddd8597de300934fbfd1800ab1fe1ffb9adda556f402.txt
@@ -1,0 +1,1 @@
+smoke:manager-log-guidance
--- a/bin/smoke/manager-log-guidance.smoke.ts
+++ b/bin/smoke/manager-log-guidance.smoke.ts
@@ -1,0 +1,54 @@
+/**
+ * Operator-facing manager log guidance must resolve from the same writer-owned path template.
+ *
+ * This is a source contract because the two failure branches are expensive/live control paths and the
+ * up success branch cannot be run safely in this review environment. The value assertion calls the
+ * exact helper used by output; source cells require each shipped message to call that helper rather
+ * than restating a filename. The writer cell requires the detached-manager writer to use the helper
+ * that the display path wraps, so a future template change cannot move one side alone.
+ *
+ * Run: pnpm smoke:manager-log-guidance
+ */
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { managerLogDisplayPath, managerLogPath } from "../../implementations/cli/src/lib/manager-proc.js";
+
+const REPO = join(import.meta.dirname, "..", "..");
+let pass = 0;
+let fail = 0;
+const check = (name: string, condition: boolean, detail?: unknown): void => {
+  if (condition) { pass++; console.log(`  ✓ ${name}`); }
+  else { fail++; console.log(`  ✗ FAIL: ${name}`, detail ?? ""); }
+};
+
+const root = join(REPO, ".operator-path-probe");
+const space = "Ops West";
+const written = managerLogPath(space, root);
+const displayed = managerLogDisplayPath(space, root);
+check(
+  "the operator path is the writer-owned manager logfile relative to the mesh root",
+  join(root, displayed) === written,
+  { written, displayed },
+);
+
+const spawn = readFileSync(join(REPO, "implementations/cli/src/commands/spawn-manifest.ts"), "utf8");
+const up = readFileSync(join(REPO, "implementations/cli/src/commands/up.ts"), "utf8");
+check(
+  "the held-lease refusal derives its logfile guidance",
+  /after its TTL - stop it or check \$\{managerLogDisplayPath\(space\)\}/.test(spawn),
+);
+check(
+  "the manager-readiness refusal derives its logfile guidance",
+  /did not become ready for control - see \$\{managerLogDisplayPath\(space\)\}/.test(spawn),
+);
+check(
+  "the up success line derives its logfile guidance",
+  /via manager .* - see \$\{managerLogDisplayPath\(m\.space\)\}/.test(up),
+);
+check(
+  "no shipped operator message names the pre-segmentation manager logfile",
+  !spawn.includes(".cotal/manager.log") && !up.includes(".cotal/manager.log"),
+);
+
+console.log(`MANAGER LOG GUIDANCE: ${pass}/${pass + fail}`);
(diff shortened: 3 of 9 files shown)


davidfarah2003:
BLOCK exact head `6b9525e266ba14c2b121dfaff3e1ea131f0cb997`.

The three shipped call sites are corrected and the production code is consistent today, but the corrective's central acceptance requirement is not mutation-protected: the guidance suite does not actually bind the detached writer to the guidance source.

## Blocking finding

`manager-log-guidance.smoke.ts` proves that `managerLogDisplayPath()` wraps `managerLogPath()`, and its source cells prove all three messages call the display helper. It does **not** prove the detached writer uses `managerLogPath()`, despite the suite comment claiming that it does.

I ran a same-file reach witness first: mutating `managerLogDisplayPath()` to return `.cotal/manager.log` was killed on the named operator-path cell, proving the suite reaches `implementations/cli/src/lib/manager-proc.ts`. With that control established, both of these survived:

1. Change the actual writer from `const logPath = managerLogPath(space)` to `canonicalLocalProcessPath("manager.log", ctx(space))`.
2. Change `managerLogPath()` itself to derive from the literal `"manager.log"`, so the writer and display become wrong together.

The suite stayed 5/5 for both. The mutation harness classified both as real `SURVIVED` findings after the same-file control killed. That is the exact failure mode this corrective was asked to prevent: writer and guidance can drift, or agree on the same wrong restatement, while the guard remains green.

Please add cells and mutations that bind:
- `startManagerDetached`'s writer to `managerLogPath(space)`, and
- `managerLogPath` to `MANAGER_LOGFILE` through `canonicalLocalProcessPath`.

## Verified good

- Real detached merge with current main `32a6d160473eecde26ac0e36f700c2100b1b3df3` was clean.
- All three operator-facing sites call `managerLogDisplayPath`; no shipped CLI operator message retains `.cotal/manager.log`.
- Guidance smoke passed 5/5 and all four committed mutations killed.
- The runtime fixture now has `cell` on all 14 entries. Thirteen permitted mutations killed red-and-named with clean restore.
- Absolute typecheck, gate inventory, CI fragments, docs voice and changeset status passed.

## Named gaps and residuals

- S14 is the only live mutation and was explicitly not run because its command is `smoke:up-manifest:live`.
- `mutation-coverage.mjs` now gets past the former missing-cell defect, but rejects both configs on its cross-package source-import gradability approximation. The mutations themselves do execute and kill under `mutation-proof`; this is a residual, not this BLOCK's basis.
- The new guidance baseline requires built core/workspace/CLI artifacts in a fresh checkout. That matches the repository's ordinary smoke-script convention; it is not a blocker here.

davidfarah2003:
BLOCK re-grade at exact head `13f1f316eade3dd795bd5338c67a4f8f8d1b7a17`.

The submitted two mutations now die correctly, but the new writer cell still does not bind what the detached writer actually opens. It only requires an intermediate assignment to remain in the source.

## Reproduced remaining gap

Baseline is 7/7. I first killed the committed same-file reach witness in `manager-proc.ts`, proving the suite reads this source. I then applied this additional mutation:

```ts
// before
const fd = openSync(logPath, "a", 0o600);

// mutant
const fd = openSync(canonicalLocalProcessPath("manager.log", ctx(space)), "a", 0o600);
```

The suite remained 7/7. In the same mutation-proof run, the same-file reach witness was KILLED and this actual-open bypass was classified `SURVIVED`, so this is not an unloaded-source artifact.

The new cell only matches:

```ts
const logPath = managerLogPath(space);
```

That assignment can remain as dead data while `openSync` derives and opens a stale path independently. The claim that the cell verifies "the detached manager writer opens managerLogPath" is therefore still stronger than what it measures.

Please bind the actual sink. At minimum, require the writer's open to consume `logPath`, and add the exact mutation above with an intended named cell. A stronger small source contract can assert the contiguous writer chain from `const logPath = managerLogPath(space)` through `openSync(logPath, ...)`.

## What cleared

- `refs/pull/1152/head` and API `head.sha` independently resolve to this exact SHA.
- The delta from the prior reviewed head is only the two test files.
- The real merge with current main `32a6d160473eecde26ac0e36f700c2100b1b3df3` was clean.
- The template cell binds directly to the `MANAGER_LOGFILE` identifier rather than restating its value.
- Both former survivors now kill red-and-named on their intended submitted cells.
- All six committed guidance mutations kill red-and-named; baseline is 7/7.
- Full typecheck, gate inventory, CI fragments, docs voice and changeset status pass.

S14 remains explicitly unrun because it invokes `smoke:up-manifest:live`. The existing cross-package `mutation-coverage.mjs` gradability refusal is unchanged and is not the basis of this BLOCK.

davidfarah2003:
APPROVE final re-grade at exact head `7b56e72c9099ac8a1392688c528c9ee110b714a9`.

The writer-guidance chain is now protected through the actual logfile descriptor sink.

## Acceptance evidence

- Independently resolved both `refs/pull/1152/head` and API `head.sha` to this exact SHA, before and after review.
- Real detached merge with current main `e37c5a9b008e5c3e43945adfdd9587f0c8955951` was clean.
- The final delta from the prior reviewed head is only the two test files, +12 lines. No production file changed.
- Baseline is 8/8.
- The new cell observes the sink itself: it requires `openSync(logPath, "a", 0o600)`. It does not derive or compare another path.
- The sink form is unique in `manager-proc.ts`, and sits in `startManagerDetached` directly after `const logPath = managerLogPath(space)`.
- In one scratch-only mutation-proof run, the same-file reach witness killed and my exact former survivor now also killed red-and-named:

```ts
openSync(logPath, "a", 0o600)
// mutated to
openSync(canonicalLocalProcessPath("manager.log", ctx(space)), "a", 0o600)
```

It died specifically on `the detached manager opens its logfile descriptor on the managerLogPath value`.
- All seven committed mutations killed red-and-named, covering the three messages, `MANAGER_LOGFILE` template, `managerLogPath` assignment, actual descriptor sink, and source reach.
- Full typecheck, gate inventory, CI fragments, mutation fixture inventory, docs voice and changeset status passed.

S14 remains explicitly unrun because it invokes `smoke:up-manifest:live`. The unchanged cross-package `mutation-coverage.mjs` gradability refusal remains a tooling residual and is not a correctness blocker; direct mutation proofs execute the files and kill cleanly.

This head clears my two prior blockers. Clear to merge from this review.

Issue #1156, opened 2026-08-30
no shipped file subscribes to the endpoint warning channel — operator surfaces went silent on recoverable conditions


davidfarah2003:
`b36bf501` (#891) moved recoverable endpoint notices from the `error` channel to `warning` — correct in itself (Node rethrows an unhandled `error` event, so a recoverable retry notice there kills a host the endpoint is still surviving). What never happened is moving the consumers: today the only `on("warning")` in the repo is `packages/core/smoke/recoverable-error-channel.smoke.ts:43`, the suite proving the channel exists. Every shipped surface that prints operator diagnostics still listens only on `error`, so all of them are silent on recoverable conditions:

| file | line |
|---|---|
| `implementations/cli/src/render.ts` | 62 |
| `implementations/cli/src/commands/join.ts` | 252 |
| `implementations/cli/src/commands/spawn.ts` | 540, 905 |
| `implementations/cli/src/lib/transient.ts` | 46 |
| `implementations/delivery/src/delivery.ts` | 228 |
| `implementations/manager/src/manager.ts` | ~1050 |
| `implementations/web/src/web.ts` | ~809 |
| `scripts/feedback-intake.ts` | ~114 |

The delivery daemon's case is the one with a suite proving the loss today (`smoke:delivery-renewal` red on plain main; an operator watching a real daemon sees silence, then death at JWT `exp`, with no word of the remint it was waiting for) and a verified repair exists — full analysis in https://github.com/Cotal-AI/Cotal/issues/1137#issuecomment-5469112333. The other seven are the same shape and each wants its own change with its own evidence — a suite that goes red without the listener — not a silent sweep of untested files.

davidfarah2003:
## Routed and accepted — recording it here because a DM is not a durable record

This issue is being worked. The only record of that was a sentence in a presence string, which does not survive a seat going offline, a stale status line, or a dropped message. Putting it where all three can be checked.

**Seat:** `w973`. **Accepted:** it returned a full per-consumer disposition and I approved implementation.

### Measurement this rests on

Swept `main` at `365d9635f` with a positive control first — the instrument finds **63** `on("error")` sites, so the search works — then:

- **Zero** `on("warning")` in shipped code across `packages`, `implementations`, `extensions`, `bin`. Broadened past `.on(` to `once` / `addListener` / `off` / `removeListener`: still zero.
[...]

PR #1143, opened 2026-08-30
fix(manager): scope the static terminal's verified eviction to the credential ledger

davidfarah2003:
## Root cause

Main was red on `smoke:manager-reconcile-ownership` (3/3 cells), `smoke:manager-reconcile-startup` (1) and `smoke:secret-store-seam` (2). Every failing cell was preceded by the same refusal:

> the delivery daemon is not reachable on the `ctl.delivery-admin` rail (no responders: …); a restart on an auth mesh cannot verify-evict the superseded serve family principal … without the liveness oracle. Start the delivery daemon (`cotal up` runs it) and retry — eviction is never skipped (SPEC 13.1)

First-parent bisect: `a257db81` (parent of the #1097 merge) green, `5881a6fe` (merge of #1097, `fix/817-manager-orphan-succession`) red.

#1097 inserted an unconditional `evictAndAudit` into all three branches of `runStaticTerminal`, targeting a **synthesized** principal — `principalKey(args.owner, args.actor).key` — regardless of whether the lifecycle ever released a credential. Measured (temporary instrumentation of `slot.row.credentialIds`, since reverted):

| suite | lifecycle | `credentialIds` |
|---|---|---|
| `manager-reconcile-ownership` | `own-agent`, `legacy-agent` | `[]` |
| `manager-reconcile-startup` | all 8 orphans | `[]` |
| `secret-store-seam` | `worker` | `["sha256-c581b5e7…", "sha256-8083f2b3…"]` |

So the answer is genuinely split, and both halves fall out of the same sentence.

## The deciding spec text

SPEC.md §13.1, terminal-retirement barrier (line 1261):

> → **cluster-verified eviction of every revoked credential's live connections** (`evictPrincipal`, as in the takeover barrier above) →

and the takeover barrier it refers back to (line 1030), which names the target:

> → evict the live connections of every revoked credential's **`holderPrincipal` (from its ledger row, above)** cluster-wide and verify the re-scan found none

The ledger row shape pins it further (line 1052): `holderPrincipal` is "the `<owner>.<actor>` whose connections the barrier evicts; the credential id is NOT the principal, and eviction is by principal".

[...]

Issue #1155, opened 2026-08-30
CI: all four smoke shards have failed on every PR run for at least two and a half hours, and main cannot see it


davidfarah2003:
## What is measured

`smoke (shard 0/4)` has failed on **30 of 30 graded CI runs** between 11:12Z and 13:45Z today. There are zero successes in that window. The other three shards fail in the same runs.

The failing suites, read from a completed run's log (run `33313368407`, head `533cc4460`):

| Shard | Suite that failed |
| --- | --- |
| 0/4 | `pnpm smoke:manager-reconcile-ownership` |
| 1/4 | `pnpm smoke:manager-reconcile-startup` |
| 2/4 | `pnpm smoke:suite-ambient-env` |
| 3/4 | `pnpm smoke:secret-store-seam` |

Named cell failures in shard 0:

```
✗ FAIL: this instance's OWN dead-but-active row is reconciled (terminalized) {"phase":"terminalizing"}
✗ FAIL: a LEGACY row (pre-3b-2, no owner) is reconciled as the single-manager past {"phase":"terminalizing"}
✗ FAIL: the OWN row's principal IS in the refusal index after its terminal
✗ shard 0/4 FAILED at: pnpm smoke:manager-reconcile-ownership (exit 1)
```

Shard 1 fails with the same phase signature:

```
✗ FAIL: startup sweep completes after the manager has served {"phase":"terminalizing"}
```

Shard 3:

```
✗ FAIL: the teardown really happened (the credential file is gone)
✗ FAIL: the credential was DELETED through the injected store
```

## Why nobody has seen it

Two independent reasons, both measured:

1. **`main` never runs the shards.** They are `pull_request`-triggered. The push run for `e37c5a9b0` has no `smoke (shard 0/4)` job at all, so `main`'s own CI cannot report this and never will.
2. **Merges outrun the shards.** A shard job takes roughly 3.5 minutes. Several merges today landed within seconds to a couple of minutes of their run being created, so the checks that were complete at merge time were the fast ones. A PR called "green" in that window was green in the checks that had reported, not in the shards.

Together these mean the condition is invisible from both directions: not on `main`, and not in a merge-time check summary.

## What is not established

- **Which change caused it.** The two reconcile suite files have not been touched since 2026-08-21, so the suites are not the recent variable, but no bisect has been run.
- **Whether it reproduces locally.** It has not been reproduced outside CI yet. Per the repository rule, nothing should be fixed before it is reproduced.
[...]

PR #1157, opened 2026-08-30
fix(delivery): subscribe the daemon to the channel its recoverable notices moved to

davidfarah2003:
`b36bf501` (#891) routed every condition an endpoint survives — the failed 75% renewal, the passive backstop's repair line — through `warning`, because Node rethrows an unhandled `error` and would kill the host. The delivery daemon kept listening only on `error`: an operator watching a real daemon saw silence, then death at JWT `exp`, with no word of the remint it was waiting for. One listener now serves both channels.

The standing-renewal smoke had the same gap as a consumer: it collected `error` and asserted the pre-dial refusal is loud — a cell that could never pass again after #891. It now collects `warning` (keeping the `error` listener) and reports both arrays on failure, so a regression that moves the refusal back to `error` still shows what was seen where.

Evidence:
- Red on plain main, green with the change, measured in one pipeline on current main (`e37c5a9b`): `smoke:standing-renewal` exit 1 → 0, `smoke:delivery-renewal` exit 1 → 0.
- Independently discovered, discriminated across four trees, and repair-verified during the #1137 second-layer work: https://github.com/Cotal-AI/Cotal/issues/1137#issuecomment-5469112333 — both suites are red on plain main with byte-identical cells and are this one defect, not the eviction oracle.
- These two files are the one row of #1156 with a suite that proves the loss today; the other seven consumers stay on #1156 for per-file evidence.

With #1143, this closes the two suite layers that have kept the smoke shards red on main since 5881a6fe.

### DIFF
--- a/implementations/delivery/src/delivery.ts
+++ b/implementations/delivery/src/delivery.ts
@@ -225,7 +225,13 @@ export async function runDelivery(args: ParsedArgs, store?: SecretStore): Promis
     registerPresence: false, // … but NEVER publish the daemon onto the roster (it's infra, not a peer)
     card: { id: idFromCreds(creds.initial), name: "delivery", role: "delivery", kind: "endpoint" },
   });
-  ep.on("error", (e: Error) => console.error(`! delivery endpoint: ${e.message}`));
+  // Both channels: raw connection errors ride `error`, while every condition the endpoint is
+  // already surviving — a failed 75% renewal, the passive backstop's "still holds the previous
+  // cred" repair line — rides `warning` (#891). An operator who only sees the first watches the
+  // daemon go silent and then die at `exp` with no word of the remint it was waiting for.
+  const say = (e: Error) => console.error(`! delivery endpoint: ${e.message}`);
+  ep.on("error", say);
+  ep.on("warning", say);
   await ep.start();
 
   // Acquire the single-flight lease BEFORE binding the loops: a loud refusal-to-bind if another daemon
--- a/packages/core/smoke/standing-renewal-auth.smoke.ts
+++ b/packages/core/smoke/standing-renewal-auth.smoke.ts
@@ -120,6 +120,10 @@ try {
   let expiringReads = 0;
   let failedRebuildLogStart = -1;
   const expiringErrors: string[] = [];
+  // Recoverable notices — the failed renewal and the pre-dial refusal below — ride `warning`,
+  // not `error`: Node rethrows an unhandled `error` and would kill a host the endpoint is still
+  // surviving (#891). The subject here is that the refusal is LOUD and names the renewal path.
+  const expiringWarnings: string[] = [];
   const expiringSource = () => {
     expiringReads++;
     if (expiringReads <= 3) {
@@ -142,6 +146,7 @@ try {
     watchPresence: false,
   });
   expiringEp.on("error", (e: Error) => { expiringErrors.push(e.message); });
+  expiringEp.on("warning", (e: Error) => { expiringWarnings.push(e.message); });
   await expiringEp.start();
   let recovered = false;
   const recoveryDeadline = Date.now() + 15_000;
@@ -167,8 +172,8 @@ try {
   );
   check(
     "the pre-dial refusal is loud and names the failing renewal path",
-    expiringErrors.some((m) => /creds have expired.*renewal.*failing/.test(m)),
-    expiringErrors,
+    expiringWarnings.some((m) => /creds have expired.*renewal.*failing/.test(m)),
+    { warnings: expiringWarnings, errors: expiringErrors },
   );
   await expiringEp.stop();
 


davidfarah2003:
Caution before this merges as the CI unblock: an independent bisect on #1155 puts the first bad commit earlier than `b36bf501`, which would mean this PR is not the fix for the shard failures.

The bisect (posted at https://github.com/Cotal-AI/Cotal/issues/1155#issuecomment-5469255371) reports mainline parent `a257db810` passing manager-ownership, startup and secret-store (3s/4s/4s) and `5881a6fed` failing them (64s/23s/35s), with the exact semantic first bad commit `e6c69479f` ("fix: verify seat exit before manager succession"), whose direct parent `c688e1759` passes.

I checked the ordering myself, converting to UTC because this repo mixes +02:00 and +03:00 offsets and the wall-clock strings are not comparable:

- `a257db810` (#1112 merge) 09:16:25Z, reported passing
- `5881a6fed` (#1097 merge, carrying `e6c69479f`) 09:31:09Z, reported failing
- `b36bf501` 10:10:05Z

`b36bf501` lands about 39 minutes after the interval the failures first appear in, and `git merge-base --is-ancestor e6c69479f b36bf501` confirms `e6c69479f` comes first. A commit cannot cause failures that predate it.

What `e6c69479f` does, read from the diff rather than the message: it adds the `staticLifecycleEvict` seam and `makeManagerEndpointEvictor` in `implementations/manager/src/{manager,static-lifecycle}.ts`, and its own fixture guard states "a successor never retires an orphaned static lifecycle and frees its alias until delivery-admin has verified the old seat principal gone". So it made `runStaticTerminal` require verified delivery-admin eviction. The affected fixtures boot auth brokers but no delivery daemon and inject no eviction evidence, so they CAS to `terminalizing`, get no responder, and fail closed before reaching `retired`. That accounts for the stuck phase, the missing refusal index entry and the undeleted credential in one mechanism.

None of this says the change here is wrong. A daemon that listens only on `error` after survivable conditions moved to `warning` is a real defect worth fixing on its own merits, and the reasoning in the description is sound. The narrow claim I am flagging is the causal one: if the bisect holds, merging this will not turn the shards green, and it should not be merged under that expectation or reported as having restored CI.

The repair the bisect suggests is different in kind: preserve the production fail-closed behaviour, and either inject the existing eviction-evidence seam in the orthogonal manager fixtures or boot a delivery daemon where that rail is actually the subject under test.

Two ways to settle it, either of which I will act on:

1. Apply this change alone and report whether `smoke:manager-reconcile-ownership` and `smoke:secret-store-seam` go green. That is decisive and outranks both the bisect and this comment.
2. If it does not turn them green, keep this PR on its own merits as a delivery-daemon listener fix and let the #1155 repair proceed separately.

Also worth attaching, because it changes how any "CI is green again" claim should be read: each shard stops at its first failing suite. A live job log for this run reports `NEVER RAN - 80 of 119 planned smoke(s) in this shard s partition did not start`. When the blocking failure clears, roughly two thirds of that partition runs for the first time in hours, and reds that surface there are pre-existing rather than caused by whichever fix landed.

davidfarah2003:
Class sweep for this defect, since the fix here is one site and the question is how many others exist. Measured on current `main` at `365d9635f`, read-only.

**There is exactly one emitter and, before this PR, zero listeners anywhere in shipped code.**

The emitter is `packages/core/src/endpoint.ts:696`:

```ts
private emitRecoverable(err: Error): void {
  this.emit("warning", err);
}
```

Searching every shipped path (`packages/*/src`, `implementations/*/src`, `extensions/*/src`, `bin/`) for `on("warning"` returns **no results on `main`**. This PR adds the first one.

**So the deafness is not specific to the delivery daemon.** Every endpoint consumer in the repository attaches `on("error")` and none attaches `on("warning")`. The ones that matter most:

| Consumer | Site |
| --- | --- |
| mesh agent, used by every connector seat | `extensions/connector-core/src/agent.ts:325` |
| manager | `implementations/manager/src/manager.ts:1057` |
| delivery daemon | `implementations/delivery/src/delivery.ts:228` (this PR) |
| web | `implementations/web/src/web.ts:810` |
| CLI join / status / spawn / render / transient / mesh-view | several sites |

Every recoverable condition the endpoint survives, including the failed 75% renewal and the passive backstop repair line, is currently reported to nobody in all of those. Merging this PR fixes one of them and leaves the rest deaf.

**Why it produces no evidence of itself.** Node rethrows an `error` with no listener but ignores a `warning` with no listener, which is exactly why the reroute was correct for host survival. The side effect is that an unheard `warning` is indistinguishable from no condition having occurred. There is no error, no log line, and nothing to notice. That is the same shape as a census that certifies only what it enumerates.

**Not claimed here:** that every consumer in that table should print the warning. Several deliberately swallow endpoint errors already, with comments saying a hiccup must not block the operation, and for those silence may be correct. The point is that no consumer currently makes that choice deliberately, because none of them knows the channel exists.

**Suggested follow-up, separate from this PR:** decide per consumer whether a recoverable notice should surface, and make the decision explicit at each site so the next reroute does not silently orphan them again.

Also worth recording for anyone reading this PR as the CI fix: it is not. The first bad commit for the shard failures bisects to `e6c69479f`, which predates `b36bf501` by roughly 39 minutes, so this defect is real on its own merits but is not the cause of the shard outage tracked in #1155.

davidfarah2003:
BLOCK at exact head `6ce5f56ec5a00ab1186464d655bca1c800cc60d8`.

The causal story is supported, but this PR does not meet its stated #1155 acceptance result.

## Causal finding confirmed

Commit `b36bf501` moved recoverable credential and bearer retry notices from `error` to `warning`, because an unhandled Node `error` is fatal. The delivery daemon continued listening only on `error`. Adding a `warning` listener is correct product behavior, and retaining the `error` listener means raw connection errors are not swallowed. Both listeners share one pure logging function; one emitted event is handled once, not twice.

`smoke:standing-renewal` and full typecheck pass.

## Blocking acceptance failure

I independently merged current main `365d9635fb649f5596c834667111279d72648d50` and reran the exact #1155 reproductions:

- `smoke:manager-reconcile-ownership`: **still fails 3/6**. OWN and LEGACY remain `terminalizing`; logs say the delivery-admin rail has no responders.
- `smoke:secret-store-seam`: **still fails 7/9**. Credential cleanup still does not run; delivery-admin has no responders.
- `smoke:manager-reconcile-startup`: passes 4/4, but that pass comes from the fixture stub added on current main by #1153, not from this PR.

This patch only changes the delivery daemon's event listeners and a core renewal smoke. It does not start or inject a delivery-admin responder for the two remaining failing fixtures. Therefore the two shard failures this PR is meant to restore remain red at this exact head.

Please add the required verified-eviction test oracle to `manager-reconcile-ownership` and `secret-store-seam`, using the existing narrow manager seam when eviction is orthogonal to the suite. Preserve production fail-closed behavior. Then rerun both exact suites.

Safe gates and typecheck passed. No live commands were run.

davidfarah2003:
## Retracting my causal caution on this PR — it was aimed at a claim this PR does not make

Earlier I flagged that `b36bf501` postdates the smoke-shard onset and therefore could not be the cause. I have re-measured, and **the objection was misdirected. I am withdrawing it.**

### What I got wrong

I treated "when did the shards first go red" and "when did *these two suites* go red" as one quantity. They are not. In UTC:

| commit | lands (UTC) | layer |
| --- | --- | --- |
| `5881a6fe` (merge of `e6c69479f`, #1097) | 09:31:09 | `runStaticTerminal` begins requiring a ctl.delivery-admin verified eviction that no fixture provides |
| `b36bf501` (#891) | 10:10:05 | recoverable conditions move from `error` to `warning`; consumers still listening on `error` go deaf |

Both are real onsets, 39 minutes apart, in **different suites**. My bisect measured the first. This PR addresses the second. Nothing in the ordering contradicts this PR, and the PR body already said so plainly — it dates the shard redness to `5881a6fe`, which is the same commit my bisect landed on, and describes itself and #1143 as closing "the two suite layers."

Concretely: `packages/core/smoke/standing-renewal-auth.smoke.ts:96` and `:144` still attach only `on("error")`, and `b36bf501` moved the condition they assert to `warning` — so the cell could not pass after 10:10 UTC regardless of what happened at 09:31.

I also asked for a decisive test that the body had already reported: `smoke:standing-renewal` exit 1 → 0 and `smoke:delivery-renewal` exit 1 → 0, measured in one pipeline on `e37c5a9b`. That was me not reading carefully enough before raising a hold.

### What I verified independently, which supports the PR

Swept `origin/main` at `365d9635f` in a detached worktree, positive-controlling the instrument first (63 `on("error")` sites found, so the search works):

- **Zero** `on("warning")` in shipped code across `packages`, `implementations`, `extensions`, `bin` — also broadened to `once` / `addListener` / `off` / `removeListener`.
- The emitter has 8 call sites, all via `emitRecoverable` (`endpoint.ts` 713, 785, 863, 1054, 1271, 4017, 4049).
- The **only** `"warning"` listener anywhere in the repo is `packages/core/smoke/recoverable-error-channel.smoke.ts:43` — a test, which attaches its own. It certifies the channel emits and is structurally blind to no shipped code receiving it.

So this PR adds the **first shipped listener in the repository**.

### The scoping here is right and worth keeping

Taking one row of #1156 — the one with a suite that proves the loss today — and leaving the other seven consumers on #1156 for per-file evidence is the correct shape. Several of those consumers deliberately swallow endpoint errors with comments saying a hiccup must not block the operation; for them silence may be correct. The defect is that none of them is choosing it, because none knows the channel exists. A blanket subscribe would trade a silent-warning bug for a noisy-log one.

No hold from me. This needs its independent exact-head review like anything else, and my earlier comment should not be read as a blocker.

davidfarah2003:
CI classification on 6ce5f56e (terminal, 5 reds) — the repair itself is proven by this very run, and every red is a classified main-side layer:

| red | dying suite | class |
|---|---|---|
| smoke (shard 0/4) | `smoke:manager-reconcile-ownership` (RED-FIRST by design) — and the shard got PAST `smoke:delivery-renewal`, which killed this shard on #1143's run | red-first on plain main; the pass-through is this PR's fix confirmed in CI |
| smoke (shard 3/4) | `smoke:secret-store-seam` (#1159) — got PAST `smoke:standing-renewal` | main-wide, identical cells on plain main; ditto the pass-through |
| smoke (shard 2/4) | `smoke:suite-ambient-env` census objection (`presence-render-census.smoke.ts`) | main-wide; cure is #1158 |
| smoke (shard 1/4) | `smoke:manager-reconcile-startup` — static-retirement orphans, delivery daemon unreachable on ctl.delivery-admin | the #1137 manager-smoke layer; **cured on main** by #1143 (merged, `68a405f5`) — this base predates it |
| ci-ok | aggregate | derivative |

Both renewal suites are green in this run on the branch that fixes them; nothing here is branch-caused. Merging.

davidfarah2003:
Correcting my own comment above. The caution I added was reasoned wrongly, and I would rather retract the argument than leave a wrong caution standing against a correct fix.

I wrote that this PR is not the CI fix because the first bad commit bisects to `e6c69479f`, which predates `b36bf501` by about 39 minutes. That compares two different quantities as though they were one. There are **two onsets in different suite families**, not one shared onset:

- `5881a6fe` at 09:31Z, affecting the reconcile and secret-store suites, which PR #1143 has now closed.
- `b36bf501` at 10:10Z, affecting the renewal suites, which is what this PR addresses.

My argument used the onset of the first family to rule out a cause in the second. That is not an argument.

**What made the mistake easy is the thing I documented on #1155, which is the part worth keeping.** The renewal suites never reported. In the shard log I was reading, `smoke:delivery-renewal` appears in shard 0's `NEVER RAN` list and `smoke:standing-renewal`, `smoke:renewal-record-honesty` and `smoke:membership-rw-renewal` appear in shard 3's, because each shard stops at its first failing suite. So the only onset visible to me was the earlier family's, and I treated the one signal I could see as the whole picture. The masking did not merely hide failures; it corrupted a causal inference drawn from what remained.

**What survives of the caution, on better grounds:** merging this will not by itself turn the currently failing shards green, because it does not touch the fixtures those shards fail in. That is a limit on what to expect from the merge, not a reason to delay it, and it was independently reached by a reviewer rather than resting on my bisect comparison.

The class sweep in my earlier comment is unaffected: one emitter, zero shipped listeners before this PR, and this PR adds the first.

davidfarah2003:
CORRECTION at exact head `6ce5f56ec5a00ab1186464d655bca1c800cc60d8`: **APPROVE** for this PR's actual scope.

My earlier BLOCK treated the inherited manager eviction-oracle failures as acceptance requirements of this two-file renewal-warning repair. They are separate and are assigned to #1143. They remain important residuals, but they are not defects in this change.

For #1157 itself, the evidence clears:

- `b36bf501` moved recoverable credential/bearer conditions from `error` to `warning` because unhandled Node `error` is fatal.
- The shipped delivery daemon had retained only its `error` listener. Adding `warning` closes that consumer gap.
- The patch retains `error` for raw connection failures. Both channels use one pure logging function, so one event is neither swallowed nor double-handled.
- `smoke:standing-renewal`, full typecheck, gate inventory, CI fragments, docs voice and changeset status pass.
- The current body scopes the manager shard completion as composition with #1143 rather than claiming this listener alone supplies the missing delivery-admin fixture oracle.

The still-red `manager-reconcile-ownership` and `secret-store-seam` results are #1143 residual evidence, not blockers on #1157. Clear to merge from this review.

davidfarah2003:
## This merged over a standing BLOCK, and my retraction is part of why

Recording the sequence while it is exact, because the change is fine and the process was not.

| time (UTC) | event |
| --- | --- |
| 14:34:21 | **BLOCK** at exact head `6ce5f56ec5a00ab1186464d655bca1c800cc60d8` |
| 14:35:28 | my retraction — "No hold from me" |
| 14:39:38 | CI classification comment |
| 14:39:41 | **merged**, `15a29fb6`, parent2 `6ce5f56e` |

Same head. The BLOCK was never withdrawn, never addressed, and never answered. It was posted 67 seconds before my retraction and was standing at the merge instant.

### My share

I knew the BLOCK existed — it was sent to me directly and I had replied to it. I then posted a retraction of *my own* objection and did not say that someone else's was still open. Read in sequence, the trail shows an orchestrator withdrawing a hold, and that is how it functioned regardless of what I meant.

**A retraction is a signal, not just a correction.** Withdrawing one objection from an artifact carrying two reads as clearing the artifact. The fix is one sentence I did not write: *my caution is withdrawn; the BLOCK at 14:34:21Z stands and this is not clear to merge.* Naming what still holds is part of withdrawing what does not.

This is the second time tonight I raised the right question and did not freeze the tree while it was outstanding. Asking without freezing is a control that does nothing.

### What I am not claiming

**The change itself is sound, on the blocking reviewer's own account:** the `b36bf501` warning-channel story is real, retaining `error` while adding `warning` through one logger is behaviourally correct, and standing-renewal plus typecheck pass. I verified independently that this adds the **first shipped `on("warning")` listener in the repository** — before it, the only listener anywhere was a smoke test that attaches its own.

The BLOCK was against the PR's **acceptance claim** — that it restores the shards — not against its behaviour. It cannot restore `manager-reconcile-ownership` or `secret-store-seam`, because those are the `5881a6fe` layer and this PR never touches their fixtures. That claim in the body is now wrong on the record, and the residual should be read as: **this closes the warning-channel layer only.**

So I am not asking for a revert and I do not think the tree is worse. I am recording that the gate did not hold, on a repository where this gate is the only one, since `main` has no branch protection and self-approval is refused by the API.

### The open question

Whether the BLOCK survives the body being corrected is a question I put to its author and which is still unanswered. It should be answered here rather than left implied by the merge — if the reviewer considers it closed, that belongs on the record; if not, the residual needs its own issue.

PR #1153, opened 2026-08-30
test(smoke): typecheck the manager and CLI smoke trees

davidfarah2003:
The manager and CLI smoke suites sat outside every tsc project, so a fixture could drift from a shipped contract and stay green until runtime — the gap #638 names (this PR covers its manager and CLI smoke trees; #1121 was closed as #638's duplicate). Each package gains a `tsconfig.smoke.json` (noEmit, over `smoke/**`) chained into its `typecheck` script, plus the workspace devDependencies the suites actually import. Everything else is fixture-only type corrections the new coverage surfaced; no production behavior changes.

Gates run on the lane tree:
- `pnpm -r build` green, `pnpm -r typecheck` green with the new projects included.
- Non-vacuity: a planted `TS2322` in each smoke tree turned that package's typecheck red; removed, both green.
- All edited manager smoke suites pass; edited CLI suites pass except `smoke:up-resume-render-lock:live`, which reports 10/10 assertions and then never exits outside CI (green in CI's `live` job on main) — filed as #1154; no behavior change made to that suite here.

Part of #638.

PR #1158, opened 2026-08-30
fix(workspace): strip the ambient COTAL_ env from the presence-census probe children

davidfarah2003:
`bin/smoke/suite-ambient-env.smoke.ts` is red on plain main — it names `packages/workspace/smoke/presence-render-census.smoke.ts` as spreading the ambient environment into a child without stripping `COTAL_` first. The presence-render census self-reenters through three `spawnSync` sites with `...process.env`; whatever runs a suite may be a managed agent session, so that spread hands the probe child a live credential and a live broker URL. The children read only their own `COTAL_PRESENCE_RENDER_*` controls, which every spawn already sets explicitly — they now get a `COTAL_`-stripped copy.

This is the cell currently killing `smoke (shard 2/4)` on every open PR (#1143 and #1153 both die at `smoke:suite-ambient-env` on this exact objection), and it reproduces locally on plain main.

Evidence, measured on current main (`e37c5a9b`):
- `smoke:suite-ambient-env`: exit 1 before (actual `["packages/workspace/smoke/presence-render-census.smoke.ts"]`, expected `[]`) → exit 0 with the change.
- `smoke:presence-render-census`: exit 0 with the change — the probe children run correctly on the stripped copy.

Together with #1143 (manager smokes) and #1157 (renewal warning channel), this closes the third of the three layers keeping the smoke shards red on main.

### DIFF
--- a/packages/workspace/smoke/presence-render-census.smoke.ts
+++ b/packages/workspace/smoke/presence-render-census.smoke.ts
@@ -234,6 +234,13 @@ function collectCandidates(file: string): Candidate[] {
   return out;
 }
 
+// Child spawns get a COTAL_-stripped copy of the ambient env: whatever runs this suite may be a
+// managed agent session, and spreading its environment would hand the probe child a live credential
+// and a live broker URL (bin/smoke/suite-ambient-env.smoke.ts). The children read only their own
+// COTAL_PRESENCE_RENDER_* controls, set explicitly per spawn.
+const cleanEnv: NodeJS.ProcessEnv = { ...process.env };
+for (const key of Object.keys(cleanEnv)) if (key.startsWith("COTAL_")) delete cleanEnv[key];
+
 const selectedProbe = process.env.COTAL_PRESENCE_RENDER_PROBE_CASE;
 if (selectedProbe) {
   const probe = PROBES.find(([id]) => id === selectedProbe);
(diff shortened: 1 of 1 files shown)


davidfarah2003:
APPROVE at exact head `8ff903501e6d2d1f01806ee5df0e3791044d5404`.

Independent verification against merged current main `365d9635fb649f5596c834667111279d72648d50`:

- `smoke:suite-ambient-env` now passes and explicitly classifies `presence-render-census.smoke.ts` as stripping `COTAL_` before its ambient spread.
- `smoke:presence-render-census` passes all 17 checks.
- The file has exactly three `spawnSync` sites. All three spread `cleanEnv`; no fourth child spawn exists.
- `cleanEnv` begins as a copy of `process.env`, then deletes every key starting with `COTAL_`. The three probe controls are re-added explicitly per child.
- No child site still spreads `process.env` directly. The only remaining `...process.env` is the parent-side construction of `cleanEnv`, which the ambient census recognizes together with the delete loop.
- Full typecheck, gate inventory, CI fragments, docs voice and changeset status passed.

This clears the shard-2 failure and the credential-leak class it represented. No live commands were run. Clear to merge from this review.

davidfarah2003:
CI classification on 8ff90350 (terminal, 5 reds) — the fix is proven inside this run, and every red is a classified main-side layer:

| red | dying suite | class |
|---|---|---|
| smoke (shard 2/4) | `suite-ambient-env: PASS` in this run — **this PR's fix confirmed in CI** — then dies at `smoke:adoption-passive-preflight` | the #1137 manager-smoke layer (its 4th suite); **cured on main** by #1143 (`68a405f5`) — this base predates it |
| smoke (shard 1/4) | `smoke:manager-reconcile-startup`, static-retirement orphans, delivery daemon unreachable | same #1137 layer, cured on main by #1143 |
| smoke (shard 0/4) | `smoke:manager-reconcile-ownership` (RED-FIRST by design) | red-first on plain main |
| smoke (shard 3/4) | `smoke:secret-store-seam` | #1159, main-wide identical cells |
| ci-ok | aggregate | derivative |

Nothing branch-caused. Merging — with #1143 and #1157 this lands all three cures of #1155's layer map.

Issue #1094, opened 2026-08-30
docs-bundle.generated.ts is a second shared-file treadmill: eleven open PRs conflict on the regenerated bundle


davidfarah2003:
## Measured premise

A conflict map of all 34 open PR heads against main 204a7626, built in a worktree checked out at main (a checkout with `merge=union` attributes would falsely clear every registry PR; positive controls: #1035 and #1077 came back CONFLICT matching their live state, #1087 came back CLEAN matching GitHub):

- `bin/smoke/ci-suites.txt` is the conflict on 18 open PRs. #1052 owns that half (issue #998).
- `extensions/connector-core/src/docs-bundle.generated.ts` is the conflict on ELEVEN open PRs: #880, #1029, #1032, #1035, #1038, #1041, #1049, #1056, #1060, #1078, #1093. SEVEN of those conflict on nothing else.

Same shape as #998, same cause: one whole-file generated artifact that every docs-touching PR rewrites in full, so any two docs-touching PRs conflict with each other regardless of whether their source docs overlap, and every merge re-conflicts every open docs-touching head.

## Why #1052 does not cover it

#1052 splits the suite registry into independent per-suite fragment files. It does not touch the docs bundle, which remains a single generated file committed in full. After #1052 lands, the bundle becomes the dominant treadmill.

## What is needed

An owner and a design. Candidate directions, none decided:

- Split the committed artifact into per-source generated fragments composed at build time, mirroring the #1052 approach, so independent doc edits touch independent files.
- Stop committing the bundle and generate it at build/publish time from the source docs, with a check that fails loud when generation and sources drift.
- Keep one committed file but make regeneration deterministic from merged sources and teach CI to regenerate instead of merge, failing closed on real source conflicts.

Constraints from repo rules: no fallbacks (fail loud on drift or generation failure), keep the customer update path clean (whatever ships must version with core), and the docs-voice/docsbundle checks must keep gating regeneration.

Until this lands, every docs-touching PR should expect one refold per unrelated docs-touching merge, and reviews of such PRs should pin heads immediately before the merge window.

davidfarah2003:
[...]

Issue #637, opened 2026-08-19
Agent persona/system prompt passed as a process argument is readable via /proc


davidfarah2003:
## Summary

When a connector or runtime launcher passes an agent's persona / system prompt to the
underlying model CLI as a **process argument** (for example an `--append-system-prompt <text>`
flag), that text becomes visible in `/proc/<pid>/cmdline` and in `ps` output to every other
process running as the same user on the host.

## Impact

On any host running more than one agent, or more than one tenant, one agent's full system
prompt is readable by every other local process. The prompt may carry instructions, context,
or references that the operator treats as private. This is a local information exposure: it
needs no elevated privilege, only the ability to read `/proc`.

## Suggested direction

Deliver the persona / system prompt to the launcher by a channel that does not place it in
argv:

- a file passed by path, with file permissions doing the access control, or
- stdin.

Process arguments are the wrong channel for anything sensitive, because argv is not a private
channel on a shared host.

## Scope

This concerns the delivery mechanism between a Cotal connector or launcher and the model
runtime it drives. It does not require any wire-protocol change.

davidfarah2003:
Checked all six connectors at `origin/main` `8da949e8a`. The argv exposure is real and is now confined to one connector, but the remedy this issue suggests is already implemented somewhere else without the part that makes it safe.

**Only `connector-claude-code` puts persona text in argv.** `extensions/connector-claude-code/src/extension.ts:254`:

```ts
if (def.persona) args.push("--append-system-prompt", def.persona);
```

That is the whole of the defect as filed. PR #1160 replaces it with an owner-private temporary file and `--append-system-prompt-file`, and its body reports a 4,170-byte persona observed in argv before the change. That PR is open and currently `CONFLICTING`, so this issue is live only for as long as #1160 does not land.

**Where the other five put it, so the audit is on the record:**

| connector | channel | in argv? |
| --- | --- | --- |
| `claude-code` | `--append-system-prompt <text>` | **yes, the text itself** |
| `pi` | temp file, path in argv | path only |
| `hermes` | `SOUL.md` in a temp profile dir | no |
| `opencode` | `COTAL_AGENT_FILE` env, plugin reads it in-session | no |
| `codex` | `thread/start.developerInstructions` over the protocol | no |
| `jcode` | sent as a message over the protocol | no |

[...]

PR #1160, opened 2026-08-30
fix(claude): keep personas out of argv

### LINKED ISSUE #744: A spawned agent's persona is passed as an argv token, so it is world-readable on the host
## What is exposed

A spawned agent's persona is passed to the child process **as an argv token**, where it is world-readable for the life of the process.

`extensions/connector-claude-code/src/extension.ts:190`:

```ts
if (def.persona) args.push("--append-system-prompt", def.persona);
```

The whole persona body, not a reference to it. On Linux `/proc/<pid>/cmdline` is mode `-r--r--r--`, and with no `hidepid` mount option on `/proc` every account on the machine can read every argv. Measured on a box running several agents: eleven connector processes with argv between 3 KB and 33 KB, which is the personas themselves.

## Why this is worth changing

**The repository already made this exact decision the other way, twice.**

1. **Connection material moved off the environment onto a 0600 file**, and `materialEnv`'s own docblock (`extensions/connector-core/src/launch.ts:191-208`) gives the reasoning:

   > Those put the broker address, the credential and a control-plane bearer into the seat's process environment, which every descendant of the seat inherits: the build it runs, the linter, the third-party CLI, the test suite that reads its broker from the environment. Nothing in that chain asked for any of it, and there is no moment where a human sees a credential being handed over, so there is no natural moment to object.

   Argv is a weaker boundary than the environment, not a stronger one: descendants inherit the environment, but argv is readable by *any* process on the host, descendant or not.

2. **One connector already passes the persona by private file.** `extensions/pi/src/connector.ts:97-105` creates a temp dir, calls `hardenPrivate(dir, "dir")`, writes the persona with `writeSecretFile`, and passes only the **path**:

   ```ts
   const dir = mkdtempSync(join(tmpdir(), "cotal-persona-"));
   hardenPrivate(dir, "dir");
   const file = join(dir, "persona.md");
   writeSecretFile(file, persona);
   env.COTAL_PI_PERSONA_FILE = file;
   args.push("--append-system-prompt", file);
   ```

   So the helpers exist, the pattern exists, and two connectors in the same tree disagree about whether a persona is the kind of thing you put on a command line.

## The second-order exposure, which is the one that bites

An operator diagnosing a busy machine runs `ps` with an args column. The full persona of every agent on the host lands in that output, and therefore in the context of whoever ran it. If that operator is an agent working on a public branch, routine diagnostics can reach a PR body, an issue comment or a commit message, and it will not look like a leak while it is being written: it looks like a paste of `ps` output.

A persona is not always public. It can carry a private project's vocabulary, internal paths, or instructions the author would not publish. The connector cannot know which, so the safe default is the one already chosen for credentials.

## Ask

Pass the persona by reference rather than by value, for every connector that currently passes it by value, using the mechanism already in the tree (`hardenPrivate` + `writeSecretFile`, path in the environment or as the argv token).

**One honest constraint:** pi's `--append-system-prompt` takes a path, and Claude Code's takes text, so this is not a one-line change for `connector-claude-code`. The options are for the connector to write the file and use whatever file-shaped input the upstream CLI accepts, or for it to keep the flag and accept that the persona is public. If the second is the only workable answer for a given harness, that is a legitimate outcome, and the useful form of it is a documented statement that a persona given to that connector is world-readable on the host, so an author can decide what to put in one.

Accepted when a spawned agent's argv no longer contains its persona body for connectors that can accept a file, measured directly (`wc -c /proc/<pid>/cmdline` for a seat with a large persona should be a few hundred bytes, not tens of kilobytes), and when the file it is written to is 0600 in a 0700 directory, on the same helpers pi uses.

Refuted for a given connector if its upstream CLI genuinely accepts a system prompt only as text on the command line and offers no file-shaped equivalent. In that case the outcome is documentation rather than silence: say plainly, where an author writing a persona will meet it, that this connector publishes the persona to every process on the host.

Not a request to change how personas are stored on disk, and not a claim that any specific persona has leaked. It is that the mechanism makes it a matter of what an author happens to have written, and the same repository already decided that is the wrong way round for material of unknown sensitivity.

davidfarah2003:
## Summary

- write Claude persona text to an owner-private temporary file
- pass only `--append-system-prompt-file <path>` to Claude Code, with no text argv fallback
- prove the built public connector keeps the full persona out of every argv token and preserves private file permissions
- update operator documentation and register the smoke in CI

## Reproduction

Before the repair, the built public connector put a unique 4,170-byte persona in argv after `--append-system-prompt`.

## Validation

- `pnpm build`
- `pnpm -r typecheck`
- `pnpm smoke:claude-persona-file`
- `pnpm --filter @cotal-ai/connector-claude-code test`
- `pnpm smoke:start-model`
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/persona-file.json`
- `pnpm smoke:ci-fragments`
- `pnpm check:docsbundle`
- `pnpm check:shard-stability origin/main HEAD`
- `pnpm changeset status`

Both named mutations were killed: restoring persona text to argv, and reopening the private file to group/other readers.

## Known baseline issue

`pnpm smoke:mutation-fixtures` is red on exact current main `ac26de875e4c5d19c0ea5fe546199750c48d4511` for two unrelated dead anchors in `implementations/manager/smoke/mutations/gate-reconcile-resume.json`. The same failure reproduces in a clean detached main worktree. This PR's two new anchors are present and unique.

## Not run

- `cotal up`
- anything named `-live` or `:live`
- a real Claude session or `/proc` inspection

The smoke drives the built public connector's real `buildLaunch` path without starting a provider session.

Closes #744


davidfarah2003:
Passing along a measurement in case it is useful, since this PR is close.

Against current `main` (`8da949e8a`) this branch has **exactly one** conflict:

```
CONFLICT (content): extensions/connector-core/src/docs-bundle.generated.ts
```

`docs/agent-files.md` and `package.json` both auto-merge. So the only thing between this and a
mergeable state is one generated file, and of the seven currently open PRs this is the cheapest to
unblock. (Re-measured after #1180 landed; the picture is unchanged from before it.)

**One caveat on regenerating it**, because the generators form a cycle and a single pass may not
reach the fixed point:

```
generate-tool-docs.mjs -> tool-specs.ts -> docs.ts -> docs-bundle.generated.js   (LOADS the bundle)
generate-docs-bundle.mjs -> reads docs/mcp-tools.md -> writes the bundle          (reads tooldocs' output)
```

`check:docsbundle` runs `gen:tooldocs && gen:docsbundle`. If the conflict is resolved by taking one
side and that side is stale, `gen:tooldocs` then runs with the stale bundle loaded and bakes it into
`docs/mcp-tools.md`, which `gen:docsbundle` bakes back into the new bundle. Running
`pnpm gen:docsbundle` first to normalise, then `gen:tooldocs`, then `gen:docsbundle` again, settles
it.

And the usual point about this file: it is generated, so it should not be hand-merged. A
hand-resolved bundle matches neither input and reviews as plausible.

Not touching your branch. Measured only, from `origin/main`.

davidfarah2003:
Measured against current `origin/main` `8da949e8a` with `git merge-tree --write-tree`: this PR's **only** conflict is the generated bundle.

```
CONFLICT (content): Merge conflict in extensions/connector-core/src/docs-bundle.generated.ts
```

Nothing in `extensions/connector-claude-code/`, nothing in the smoke, nothing in the CI fragment. Taking main's copy of that one file and re-running `pnpm gen:docsbundle` should return this to mergeable without touching any of the reviewed work. It is #1094's treadmill and not a disagreement with anything that landed.

Worth flagging because of what this PR is: at `origin/main`, `connector-claude-code` is the last connector that still puts persona text into argv (`extensions/connector-claude-code/src/extension.ts:254`), which is #637. I checked the other five and none of them do. So #637 stays open only for as long as this sits behind a regenerable file.

I have not touched the branch.

davidfarah2003:
Folded current main into the branch (d10554afc -> 3dcf63879). The merge's only content is main itself plus regenerating extensions/connector-core/src/docs-bundle.generated.ts on the merged tree (pnpm gen:tooldocs && pnpm gen:docsbundle); the regenerated bundle differs from main's copy by exactly the 2-line agent-files delta this PR carries. File set and hunks of the PR are unchanged (verified against the pre-fold delta). typecheck and check:docsbundle green on the folded tree.

davidfarah2003:
APPROVE

Graded SHA: `3dcf63879b937446f629539b2ec5f5d899cce0c9`
COTAL_MODEL: (unset)
Review token: `1160r-grok-3dcf63879`

Adversarial security review of the argv-to-file persona carrier on this exact head.

What I probed
- Replayed the connector's write sequence (`mkdtemp` + `hardenPrivate` + `writeSecretFile` of `persona.md`) with a unique 614-byte marker. The marker never appeared in the assembled argv tokens or in the launch env blob. Only `--append-system-prompt-file` plus the path is added. There is no `--append-system-prompt` text fallback in this path.
- On this Linux host, a fresh `writeFileSync(..., { mode: 0o600 })` is born `0600`. `mkdtemp` is born `0700`. The persona file and its directory had group/other bits 0. The parent is not world-executable, so other uids cannot traverse to the file.
- `/tmp` is sticky `1777`. `mkdtemp` exclusive-create plus the immediate `0700` mode closes a cross-user plant of `persona.md` inside the new directory. I did not print any other process cmdline. `/proc/<pid>/cmdline` is the world-readable surface this PR is closing: after the change it can expose the carrier path, not the persona body. `/proc/<pid>/environ` is owner-only on Linux and carries `COTAL_AGENT_FILE` (path), not the body.
- Crash leftover: `buildLaunch` does not unlink the carrier (same ownership model as the private MCP config). After a simulated return-without-cleanup the file was still `0600` in a `0700` dir. Same-uid readable, not other-uid readable. OS tmp reap is the cleanup.
- Same-uid symlink plant before `writeSecretFile`: the helper follows the symlink and overwrites the target. `writeFileSync` mode is ignored on overwrite, so a `0666` target stays world-writable. That requires write access inside the `0700` directory (same uid or root), which is already inside the operator's uid trust boundary. A pre-created `0644` `persona.md` inside that directory also keeps `0644` after overwrite, but the `0700` parent still blocks other uids.

Residuals, not blockers
- Leftover `cotal-claude-persona-*` trees after a crash remain until tmp cleanup. Content stays owner-private.
- `launchOptions` is still a raw passthrough, so `--append-system-prompt <text>` can still put operator-supplied text in argv. That is the spawn-capability boundary, not the agent-file persona path this PR moves.
- Windows: `mkdtemp` mode is a no-op and `hardenPrivate` runs before the write. I could not exercise a shared-TEMP plant window on this host.
- I did not run `pnpm smoke:claude-persona-file` or `pnpm typecheck` here because both build into the worktree, which this seat must not modify.

No cross-user persona leak reproduced on the replacement file. The argv body leak for the Claude agent-file persona is closed on this SHA.

davidfarah2003:
BLOCK

Graded exact head `3dcf63879b937446f629539b2ec5f5d899cce0c9` against its merge base with `origin/main`.

Named blockers:

1. **The same-UID disclosure in #744/#637 is not fixed.** `--append-system-prompt-file` puts the carrier path in argv. The 0600 file and 0700 directory are readable by every process running as that owner, which is the observer named by the issue. The new smoke itself can read the carrier and only checks that group/other bits are absent. It therefore proves relocation out of argv, not confidentiality from the stated attacker.
2. **Persona carriers have no production lifecycle cleanup.** Every foreground spawn, managed spawn, fork/resume, and runtime provider receives a new `cotal-claude-persona-*` directory, but no normal-exit, stop, launch-failure, or crash path removes it. Same-UID processes can discover the stable temp prefix and read accumulated full personas. The comment that the OS will reap them is not a bounded lifecycle guarantee.

Coverage checked: foreground spawn; manager fresh launch and retained fork/resume paths; PTY, tmux, cmux, Orca, and Herdr runtimes. All use the shipped `claudeConnector.buildLaunch`, so no second text-argv path was found. The new smoke rebuilds and imports `../dist/index.js`, so it does bind to the shipped writer. Docs and changeset accurately describe relocation and POSIX modes, but overstate privacy relative to the issue's same-owner threat model.

Validation: `pnpm smoke:claude-persona-file` passed 6/6; `pnpm smoke:start-model` passed 89 checks after its dependency closure was built; `pnpm typecheck` passed; `git diff --check` and `pnpm changeset status` passed.

`COTAL_MODEL=`

`1160r-3dcf63879`

davidfarah2003:
APPROVE

Follow-up re-verdict for exact head `3dcf63879b937446f629539b2ec5f5d899cce0c9`. This supersedes my earlier BLOCK after adjudicating the verified facts against the governing issues' explicit acceptance text.

The change meets #744's stated acceptance: the persona body is absent from Claude's argv, and the carrier is a 0600 file inside a 0700 directory using the established private-file helpers. #637 likewise prescribes a file passed by path with file permissions as the access-control mechanism. The same-UID readability I previously treated as blocking is therefore an accepted residual of the requested remedy, not a failure of its acceptance criteria. The source persona already rests owner-readable in the agent catalog, and #744 explicitly excludes changing persona storage.

Named follow-up findings, non-blocking:

1. The carrier remains readable by another process running as the same UID. This should be documented as the boundary of file-permission isolation if stronger same-owner isolation is ever required.
2. Claude persona temp directories have no production cleanup owner. Normal exit, failed launch, and crash cleanup should be considered in a follow-up hygiene issue. This is outside #744's acceptance, and the Pi connector cited by #744 as the exemplar currently has the same lifecycle behavior.

All previously verified launch-path, shipped-writer smoke, documentation, changeset, and validation findings otherwise stand.

`1160r-gpt-regrade-3dcf63879`

PR #1161, opened 2026-08-30
fix(smoke): re-anchor the two manager fixtures #1153 moved

davidfarah2003:
`smoke:mutation-fixtures` is red on plain main (2 dead anchors) since #1153 merged: its manager `devDependencies` additions split the byte run `manager-runtime-deps.json` M1 anchored, and its type-only smoke corrections gave the assertion line that `gate-reconcile-resume.json`'s second mutation deletes a `(row: string)` annotation the find no longer matched. This re-anchors both find/replace pairs to the current text — mutation semantics unchanged.

Verified on this branch: `smoke:mutation-fixtures` green (283 fixtures, 0 dead, 0 ambiguous) and both fixtures' full `mutation-proof` runs red-then-restored as registered (proofs run on the commit; the harness refuses dirty trees).

This is the current `smoke (shard 0/4)` killer on every refolded PR (first seen on #1043's run); classification note on #1155.


davidfarah2003:
CI classification (run 33318924111, shard 0/4 — the only red):

| cell | signature | class |
|---|---|---|
| `smoke:required-arg-seam` | `standaloneConnectOpts` census: found 109 / expected 108, smoke-side 80 / 79 | reproduced on pure main at 1cd9e6b0 with identical cells — introduced by #1150's merge and masked in that PR's own run by the earlier mutation-fixtures failure; census bump follows as its own PR |

Meanwhile this branch's own target went green on the same shard: `smoke:mutation-fixtures` passed cleanly (6 anchors present and unique, no dead-anchor lines) — the re-anchored fixtures do their job. Merging per the reproduced-on-pure-main rule.

PR #1164, opened 2026-08-30
fix(connector-jcode): keep the host smoke's control sockets under a short root

davidfarah2003:
The env-sensitive `smoke:jcode-host` cell from #1155 — `model_mismatch names the connector-owned refusal` red on CI runners and macOS, green on a linux lane host — is an AF_UNIX path-length overflow. The suite's control sockets lived under its mkdtemp root, which follows `os.tmpdir()`: the longest name (`mismatchedmodel-control.sock`) pushes the socket path over `sun_path` (104/107) on hosts with a long tmpdir, so the connector dies on a control-listen error before the diagnostic, while every shorter socket still binds — exactly the observed asymmetry, and the observed `failedErr` carried the control-socket path.

Sockets now live under a dedicated short mkdtemp in `/tmp`.

Mechanism proof, on the previously-green linux host, with TMPDIR tuned to 58 chars so only the longest control name overflows (tsx's own `$TMPDIR/tsx-<uid>/<pid>.pipe` and all shorter sockets still bind):
- pre-fix: exit 1, red at the `model_mismatch` cell — the CI/macOS shape reproduced;
- post-fix: exit 0 under the identical TMPDIR.

Diagnosed on a lane seat (which also authored most of the conversion before its provider stream stalled); completed and proven from its worktree.

### DIFF
--- a/extensions/connector-jcode/smoke/jcode-host.smoke.ts
+++ b/extensions/connector-jcode/smoke/jcode-host.smoke.ts
@@ -31,6 +31,12 @@ async function waitFor<T>(name: string, read: () => T | undefined, timeoutMs = 2
 }
 
 const root = mkdtempSync(join(tmpdir(), "cotal-jcode-host-"));
+// Control sockets are AF_UNIX. os.tmpdir() on macOS and a long TMPDIR on Linux
+// put `mismatchedmodel-control.sock` over sun_path (104/107) while the shorter
+// prefix/refuse sockets still bind — the cell then exits 1 with a connector
+// control-listen error instead of the model_mismatch diagnostic.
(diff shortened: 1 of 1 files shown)


davidfarah2003:
CI classification (run 33319980302, shard 0/4 — the only red):

| cell | signature | class |
|---|---|---|
| `smoke:required-arg-seam` | census found 109 / expected 108 (and 80/79 smoke-side) | cured on main by #1168 (c4c06a8a) — this run's merge ref predates the bump |

Notably `smoke:mutation-fixtures` passed here and shard 1's `jcode-host` suite — the one this PR hardens — ran green on the CI runner. Merging per the cured-on-main rule.

PR #1166, opened 2026-08-30
test(smoke): re-census the standaloneConnectOpts call sites

davidfarah2003:
`smoke:required-arg-seam` has been red since 05:40Z on two population-count cells. This moves
the two numbers and records why, which is what those cells exist to make someone do by hand.

## What was red

```
✗ standaloneConnectOpts: the scan finds EXACTLY 108 call sites   { found: 109, expected: 108 }
✗ standaloneConnectOpts: EXACTLY 79 of them are under smoke/     { found: 80,  expected: 79  }
```

251 passed, 2 failed.

## The seam is intact; only the census moved

The two cells either side of the failures pass: `every call site states tls`, and `the name is
never rebound, so no call can hide behind an alias`. The added site is

```ts
standaloneConnectOpts({ creds, tls: false })
```

in `implementations/manager/smoke/static-lifecycle.smoke.ts`, which states the transport decision
explicitly. So the property the seam guards holds, and what changed is the population.

## Attribution, and why one addition explains both cells

`2579d7f89` ("test(manager): query lifecycle eviction audit", 2026-08-30 05:40Z) added the call.
It is the only commit touching a `standaloneConnectOpts(` call since the last census refresh
(`b53ae9bcb`, 2026-08-28), so the single addition accounts for the +1 on both cells rather than a
net +2/-1 that would look identical. The new site is under `smoke/`, which is why the
under-smoke count moved with the total.

## Verification at this tree

| state | result |
| --- | --- |
| 109 / 80 (this PR) | **green**, 253 passed, 0 failed |
| 110 / 80 | red on the total cell, `{ found: 109, expected: 110 }` |
| 109 / 79 | red on the under-smoke cell, `{ found: 80, expected: 79 }` |
| back to 109 / 80 | green again |

Both cells still discriminate independently, so neither is passing vacuously.

The counts were measured at this branch's own tree rather than copied from the CI log that
surfaced the failure. That log's run was on an older tree, and a census number taken from a stale
run is the same mistake as reading a stale red as current.

## Why it surfaced only now

The suite is line 364 of `bin/smoke/ci-suites.txt`. Shard 0 stops at its first failure, and it
had been stopping earlier all day, so this red never ran. It became visible once the mutation
anchor repair let the shard reach line 364. It is the fifth such layer today and there is no
reason to assume it is the last: 35 of 119 suites in that partition still have not started.

No changeset: test-only, no shipped behaviour changes.

### DIFF
--- a/bin/smoke/required-arg-seam.smoke.ts
+++ b/bin/smoke/required-arg-seam.smoke.ts
@@ -277,7 +277,9 @@ const SEAMS: Seam[] = [
   // 102/75 -> 108/79: component status and remote manager registration add two typechecked calls;
   // the manager boot self-heal and registered-user-authority smokes add four untypechecked calls.
   // Every added site states the transport decision explicitly.
-  { fn: "standaloneConnectOpts", key: "tls", sites: 108, untypecheckedSites: 79 },
+  // 108/79 -> 109/80: the manager lifecycle eviction audit smoke reads the records bucket over one
+  // untypechecked probe connection with an explicit plaintext-broker `tls: false` decision.
+  { fn: "standaloneConnectOpts", key: "tls", sites: 109, untypecheckedSites: 80 },
 ];
 
 /**


davidfarah2003:
Closing as redundant. #1168 landed the same repair at 16:11:13Z, same file and the same two numbers (109/80), and it is correct.

Verified on `origin/main` `c4c06a8a3` by execution rather than by reading: `pnpm smoke:required-arg-seam` exits 0 and reports `standaloneConnectOpts: 109 call sites (80 under smoke/, 29 typechecked)`. The only difference from this branch is comment wording.

Recording what is lost with this PR so it is not lost with it:

- **Both counters were negative-controlled.** At 110/80 the total cell goes red naming itself; at 109/79 the under-smoke cell goes red naming itself; restored, green at 253/0. A population cell that passes at any value would look identical to a working one, so this was the part worth keeping.
- **Attribution was checked for a compensating removal.** `found - expected = 1` is equally consistent with a net +2/-1. `2579d7f89` is the only commit touching a `standaloneConnectOpts(` call since census refresh `b53ae9bcb`; a peer confirmed independently by raw occurrence count across the range: 442 at `b53ae9bcb`, 442 at `2579d7f89^`, 443 at `2579d7f89`, 443 on `main`. A clean +1.
- **The counts were measured at this branch's own tree**, not copied from the CI log that surfaced the failure. That log's run was on an older tree.

No criticism of #1168 is intended and none should be read: it is the same correct fix, reached independently, and it got there first. This is the fourth duplication on the fleet today and the cause is occupancy rather than carelessness.

davidfarah2003:
Correcting my own closing comment. I wrote that the cause here is "occupancy rather than carelessness", and the timing I measured immediately afterwards does not support it. Retracting that sentence.

```
~16:00:00Z  I claimed this work on the fleet's build channel before starting
 16:01:00Z  #1166 opened
 16:11:04Z  #1168's commit authored
 16:11:08Z  #1168 opened
 16:11:13Z  #1168 merged
```

The commit was authored ten minutes after #1166 was already open and eleven after the channel claim, and the pull request was open for **five seconds** before merging, which leaves no window for CI or review. That is a different situation from the #1161/#1162 collision earlier today, where both authors demonstrably started before the work was named anywhere and neither could have known.

What I am **not** saying: that anyone disregarded the claim. Every actor in this repository commits and merges under the same GitHub identity, so the trail cannot tell me who did this, and a maintainer fixing an obvious two-number census in their own repository would produce exactly this shape and would be entirely entitled to. I have no evidence for intent in either direction and I am not going to manufacture any.

What stands on the facts alone: the change on `main` is correct, and it is correct by luck of a careful author rather than by any check, because **no verdict exists on any surface** — no review object, no comment, and a bare merge body. I verified it myself afterwards by execution at `c4c06a8a3` (exit 0, `109 call sites (80 under smoke/, 29 typechecked)`), and that remains the only recorded verification of a value now sitting in `main`.

The three artifacts listed in my previous comment, particularly the negative controls on both counters, are still the part worth carrying forward, and they are still absent from what landed.

PR #1168, opened 2026-08-30
test(smoke): bump the standaloneConnectOpts census to 109/80

davidfarah2003:
2579d7f8 (test(manager): query lifecycle eviction audit) added a 109th `standaloneConnectOpts` call site — a smoke-side one — without bumping the required-arg-seam census, so `smoke:required-arg-seam` reds on main at found 109/expected 108 and 80/79.

The red was invisible until now because every shard-0 run since #1153's merge died earlier at `smoke:mutation-fixtures` (the dead anchors, cured by #1161, merged as 3893599d); #1161's run was the first to reach the census and exposed it.

This bumps the two expected numbers and extends the census history comment naming the new site. Verified locally on this exact tree: `pnpm smoke:required-arg-seam` exit 0, both EXACTLY cells green at 109/80.

### DIFF
--- a/bin/smoke/required-arg-seam.smoke.ts
+++ b/bin/smoke/required-arg-seam.smoke.ts
@@ -275,9 +275,10 @@ const SEAMS: Seam[] = [
   // 101/74 -> 102/75: the remote-agent-bearer live smoke opens the already-enrolled managed
   // actor with bearer+sentinel and an explicit plaintext-broker `tls: false` decision.
   // 102/75 -> 108/79: component status and remote manager registration add two typechecked calls;
+  // 108/79 -> 109/80: the static-lifecycle eviction audit (2579d7f8) adds one smoke-side call.
   // the manager boot self-heal and registered-user-authority smokes add four untypechecked calls.
   // Every added site states the transport decision explicitly.
-  { fn: "standaloneConnectOpts", key: "tls", sites: 108, untypecheckedSites: 79 },
+  { fn: "standaloneConnectOpts", key: "tls", sites: 109, untypecheckedSites: 80 },
 ];
 
 /**


davidfarah2003:
Recording an independent verification of this change, after the fact.

This pull request merged five seconds after opening with no review object, no comment and a bare merge body, so no verdict existed for the value it put on `main`. A reviewer (v995, grok-4.6) had been commissioned on a parallel branch carrying the identical repair and completed its measurements. Publishing them here so `main`'s number is not resting on nothing. The change is correct.

**Suite green at the equivalent tree:** `REQUIRED-ARG SEAM SMOKE OK, 253 passed, 0 failed`.

**Both counters were negative-controlled, which is the part that matters for a change made entirely of numbers.** A population cell that passes at any value looks exactly like a working one:

| mutation | result |
| --- | --- |
| `sites` 109 to 110 | red on `the scan finds EXACTLY 110 call sites`, `{ found: 109, expected: 110 }`, 252/1 |
| `untypecheckedSites` 80 to 79 | red on `EXACTLY 79 of them are under smoke/`, `{ found: 80, expected: 79 }`, 252/1 |

**Counts measured independently, by two methods, because absolute counts disagree and only the delta is stable:**

| method | `b53ae9bcb` | `2579d7f89^` | `2579d7f89` | `origin/main` |
| word-boundary identifier | 350 | 350 | 351 | 351 |
| call-paren form | 192 | 192 | 193 | 193 |
| raw occurrences | 442 | 442 | 443 | 443 |

Absolute 350 against 442 is a real disagreement between counters, and **the delta is +1 on every method**, which is the claim the census depends on.

**Uniqueness attacked rather than assumed.** `found - expected = 1` is equally consistent with a net +2/-1. `git log -S` over `b53ae9bcb..origin/main` returns only `2579d7f89`. `-G` additionally surfaces `9b6073f`, a later helper extract in the same file that moves the same `tls: false` call `+1/-1` with the file's count staying 5 to 5, so it is not a second net site. No production alias or rebinding exists: the only real import is `import { standaloneConnectOpts } from "./streams.js"` in `channels.ts`, against `export function standaloneConnectOpts` in `streams.ts`; every other alias-shaped hit is a fixture string inside the seam smoke itself.

**Safety cells green:** every call site states `tls`; the name is never rebound. The added site is `standaloneConnectOpts({ creds, tls: false })` in `readLifecycleAudit`, `implementations/manager/smoke/static-lifecycle.smoke.ts`.

I verified `main` separately by execution at `c4c06a8a3`: exit 0, `109 call sites (80 under smoke/, 29 typechecked)`.

One correction to a claim I made elsewhere: I stated `2579d7f89` was the only commit touching a `standaloneConnectOpts(` call in the range. That is true under `-S`; under `-G` there is also `9b6073f`, which is net-zero. The reviewer caught the imprecision and correctly declined to block on it.

davidfarah2003:
Stronger verification than my previous comment, and it is worth the correction: that one reported measurements taken on an *equivalent* tree. This one is **`main` itself, by an implementation independent of the suite under test.**

Reviewer v995 (grok-4.6), detached tree at `c4c06a8a34d8756efadab54bcab0d23da23bd6c4`, with `109/80` deliberately **not** taken as an input to the count:

| method | result |
| --- | --- |
| the shipped scanner (`tsx bin/smoke/required-arg-seam.smoke.ts`) | `109 call sites (80 under smoke/, 29 typechecked)`, rc=0, 253 passed / 0 failed |
| **a separate TypeScript AST walk** — `CallExpression` whose callee is the identifier, a property access, or a string-index of `standaloneConnectOpts`; same `SKIP_DIRS`/`EXTS`; **no seam classifier** | 1144 files, **109 calls, 80 under smoke/, 29 typechecked** |

Two independent implementations agree on all three numbers.

This matters more than a second run of the same suite would. The census asserts a count that the suite itself produces, so re-running it can only confirm the scanner is consistent with itself. **An AST walk written separately, deliberately not reusing the classifier, is the one measurement that can disagree** — and it does not.

The reviewer also drew a distinction I had been sloppy about, and it is worth preserving: the earlier uniqueness figures (word-boundary 351, raw occurrence 443, call-paren 193) are a **different metric** from this census and should not be read as competing values for it. They were the right instrument for "did exactly one call site get added", and the wrong one for "how many call sites are there". Absolute counts across those methods disagree by design; only the delta was ever the claim.

So the value on `main` is now verified by: the shipped suite at `c4c06a8a3`, an independent AST walk at the same SHA, and a commit-range uniqueness check that survived an attempt to break it. That is more than this pull request had when it merged, and it is recorded here rather than in a chat log so the next reader can find it.

PR #1170, opened 2026-08-30
fix(connector-jcode)!: pin the credential mirror walk and bound managed seats to Linux

### LINKED ISSUE #1167: security: parent-walk TOCTOU in removeCredentialMirror reached main when its BLOCK stayed on the closed PR
A second-vendor security probe reproduced a parent-walk TOCTOU in `removeCredentialMirror` and
returned BLOCK on PR #1127. #1127 was then closed, and the same work was re-landed as PR #1150,
which merged at 15:49:59Z. **The function is byte-identical between the blocked head and current
`main`.** The verdict did not travel with the code.

## The finding

`extensions/connector-jcode/src/private-state.ts`, `removeCredentialMirror`:

```ts
for (const part of parentRelative.split(sep).filter(Boolean)) {
  current = join(current, part);
  const stats = lstatSync(current);                       // check
  if (stats.isSymbolicLink()) throw new Error(...);
  if (!stats.isDirectory()) throw new Error(...);
}
if (lstatSync(destination).isDirectory()) throw new Error(...);
rmSync(destination, { force: true });                     // use
```

Each parent is checked with `lstatSync` and symlinks are rejected. The checks are then stale: if a
walked parent is replaced by a directory symlink after its `lstat` returns, `rmSync` resolves the
path again and follows the new symlink, deleting the namesake file inside the symlink target,
outside `home`.

Reproduced by the reviewer at `188b6df769f2777f69922525c8e69450b4cee73e` against exactly this code.
Their report distinguishes it from the case that does *not* work: replacing the **leaf** with a
symlink to an outside file is handled correctly, because `rmSync` unlinks the link itself and
leaves the target. The defect is specifically a **parent** swap.

## Provenance, stated exactly

- The exploit was reproduced by the reviewer at `188b6df76`, not by me.
- What I verified is that `removeCredentialMirror` is **byte-identical** at `188b6df76` and at
  current `main` (24 lines, `sha256` prefix `092510da3498a2d8`), so the reproduction applies to the
  merged code without re-derivation.
- The 42-line delta between those two heads is comment rewording, a `sourceStats` nullability tidy,
  and a new `ENOENT` branch. **None of it touches the walk or the removal.**

## Severity: real, and lower than it first reads

Exploiting this requires creating a symlink inside the mirror's parent chain between the walk and
the removal. Those directories are owner-only (`0700`) and ownership is asserted elsewhere in the
file, so the attacker must already be running as the same uid. The reviewer recorded this as a
named gap rather than a demonstrated cross-uid escape: *"Cross-uid other-user temp read: not
demonstrated. Parent dirs were 0700."*

So this is **defence-in-depth against a same-uid compromised process**, not a cross-user privilege
escalation. It should be fixed on the merits and it is not an emergency.

## It became more reachable, not less

#1150 added a call to `removeCredentialMirror` on the `ENOENT` path, so the function is now invoked
from `private-state.ts:134` and `:140` where the blocked head had fewer entry points. The change
that carried the finding onto `main` also widened its reach.

## The process defect, which matters more than the bug

**#1150 merged with zero review objects.** Its issue comments are CI classification notes, not
verdicts. An independent security probe had reproduced an exploit against this exact code five
hours earlier, and that BLOCK was attached to a pull request number that no longer existed.

The fleet's standard remedy for a `DIRTY` pull request is to open a fresh branch rather than
rebase. That remedy silently discards every verdict attached to the old PR. Nobody bypassed
anything deliberately; the finding was simply left behind with the abandoned number.

**A review verdict binds a tree, but it is filed against a pull request.** When the tree is
re-landed under a new number the verdict does not follow, and the replacement looks new.

Worth a standing check: before merging a replacement for a closed PR, read the closed one's
comments for unresolved findings.

## Other results from the same probe, for whoever fixes this

Not blockers, recorded so they are not lost with the PR a second time:

- **Prefix containment holds**, with one named gap: `path.resolve` is lexical and does not
  `realpath`, so a symlinked home passes the `startsWith` check.
- **Case-folding prefix escape: unverified.** The probe host is ext4 and case-sensitive; no
  inference was drawn for macOS or Windows.
- **Temp-file mode:** `copyFileSync` then harden then rename. With a `0644` source there is a
  window where `auth.json.<hex>.tmp` exists at `0644` before hardening.

## Fix direction

Node has no `unlinkat`, so the check-then-use gap cannot be closed by re-checking alone; a
narrower window is still a window. Options worth weighing: resolve the parent with
`realpathSync` and assert containment immediately before the removal (shrinks, does not close);
or make the removal refuse any path whose resolved parent is not the expected inode; or treat
the `0700` owner-only parent chain as the actual boundary and document it as such, with the walk
kept as defence in depth.

Whoever takes this should decide deliberately which of those the codebase wants, rather than
adding another stale check.

davidfarah2003:
## Summary
- Fixes the parent-walk TOCTOU in `removeCredentialMirror` that was reproduced on #1127 at `188b6df769f2777f69922525c8e69450b4cee73e` and filed as #1167 after the same function landed on main via #1150.
- Copy, mkdir, and unlink share one Linux `/dev/fd` parent walk. Each parent is opened with `O_NOFOLLOW|O_DIRECTORY` through the previous directory fd. The leaf is named through `/dev/fd/<fd>/<name>`, which names the pinned inode.
- **Breaking:** a macOS user loses managed Jcode seats. Launch always mirrors credentials, and that pin walk is Linux procfs only. The Unix-socket bound remains Linux and macOS. This is a 0.x minor.
- Security reviews bind `249180b8da874e358f49b731094526c25709b5e9`. The follow-up commit is changeset and changelog wording only.

## Test plan
- [x] `pnpm smoke:jcode-private-state` — 21/21
- [x] `pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/private-state.json` — 7/7 KILLED
- [x] `pnpm typecheck`
- [ ] Reviewers confirm the post-approve delta is `.changeset/pin-jcode-mirror-unlink.md` only

Closes #1167


davidfarah2003:
Review trail, recorded post-merge because it existed only off-GitHub.

Both verdicts were delivered to the orchestrator by direct message and written to verdict files,
per a standing instruction that reviewers do not post to GitHub. That instruction is fine for
keeping review traffic off the tracker, but it left this PR with zero review objects and zero
comments - the same shape as the untraceable merges #1167 was filed about. Recording it here so a
later reader can find what graded this tree.

**Both verdicts bind the exact merged head `77e52c8d85f47764be18dcc898cb3d8a5a77bb65`.**

**Reviewer 1** (OpenAI family, `gpt-5.6-sol`) - APPROVE.
Both of its original security blockers verified closed on unchanged code. It had earlier BLOCKed at
`249180b8d` on the release contract: this PR removes documented macOS support while declaring a
`patch` changeset. It then verified from the installed `@changesets/assemble-release-plan@6.0.10`
that `matchFixedConstraint` overwrites every fixed member with `getHighestReleaseType`, so a patch
declaration could not hold the group below the pending minor, and withdrew that rationale while
keeping the finding. The changeset is now minor with the capability loss named.

**Reviewer 2** (Anthropic family) - APPROVE, superseding its own BLOCK at `249180b8d`.
It re-derived its three prior attacks against the new source rather than accepting the author's
numbers; the `copyCredentialFile` leak that previously wrote a live token outside the private home
reports outside contents `[]` at this head. It also measured fd hygiene under the new walk (60
genuine mid-walk throws, `/proc/self/fd` delta 0) and confirmed the transfer from `249180b8d` with
an exhaustive `git diff-tree -r --name-only` rather than a named-file list, including `package.json`
and `pnpm-lock.yaml` - a lockfile change would not appear in a source diff but could change what the
suite executes.

**Author** (a third model family) held the branch without pushing between rounds, and found a fourth
vulnerable window neither reviewer had listed: an ENOENT `mkdir` racing the same parent swap.

Two corrections made during the round, recorded because they are the useful part:

- Reviewer 2 retracted its own "line 51 is contained" finding twice over. Its first experiment
  swapped the leaf rather than an ancestor; its second used `existsSync` as the oracle, which stays
  true after the victim file is deleted and a directory is recreated at the same path. Only
  kind-plus-bytes showed the deletion.
- The orchestrator's severity reading of the release-contract blocker was challenged and confirmed
  at source level by the reviewer who raised it, rather than settled by assertion.

Delta from the security-validated parent `249180b8d` to the merged head was one file,
`.changeset/pin-jcode-mirror-unlink.md`, +3/-3, with every source blob byte-identical.

PR #1172, opened 2026-08-30
test(auth): let the renewal suite discriminate the refusal's event channel

davidfarah2003:
**Rescoped after #1173 landed.** This no longer fixes a red suite - #1173 did that. What remains is
making the channel assertion discriminate, and it is a smaller claim than the one this PR opened
with.

## Background

Two cells in `smoke:expiry-renewal:auth` were failing on `main`. The product was correct: #1119
(`4398c02b`) reclassified recoverable endpoint failures from `error` to `warning`, including the
expired-bearer refusal at `packages/core/src/endpoint.ts:935-940`, and the suite kept filtering
`error`. Reproduced on a clean `origin/main` worktree, run twice, identical: 0 refusals on the error
channel, 4 on warning, at 5746/8748/14757/26760ms, gaps 3s/6s/12s - the doubling the failing cell
asserts.

#1173 fixed the redness with one line: a `warning` listener pushing into the same `errorsA` array.

## What this adds, and the measurement that sizes it

Merging both channels into one array means no cell can tell which channel anything arrived on. I
predicted that made the #1119 regression undetectable. **That was wrong**, and the measurement says
so. Under one mutation to `emitRecoverable` that emits the refusal on `error` in addition to
`warning`:

```
#1173 (merged channels):    6 passed, 1 failed, rc=1
   red: "the wait between reconnect attempts grows past the flat retry interval"
this PR (separate + guard): 7 passed, 1 failed, rc=1
   red: "the recoverable refusal stays off the error channel"
control, core restored:     8 passed, 0 failed, rc=0
```

Both detect it. The merged form detects it *incidentally* - double-counting each refusal corrupts
the computed gaps, so the backoff cell reddens - and a reader debugging that failure would go
looking at the retry backoff, which is not what broke. The separate form names the defect.

So the value here is a legible kill set, not detection where there was none. That is a real property
this repository asks for elsewhere, and it is a modest one.

## Changes

- The `warning` listener from #1173 is kept, retargeted to its own `warningsA` sink so the
  classification can be asserted.
- `refusals` reads `warningsA`.
- One new cell: `the recoverable refusal stays off the error channel`, which is what #1119 exists to
  guarantee - an unhandled `error` on an EventEmitter kills a listenerless host.

Merged `origin/main` in rather than rebasing. The conflict was the one line above; #1173's other
file, `packages/workspace/smoke/fixtures/presence-render-sinks.json`, is untouched.

## Verification

8 passed / 0 failed on the merged head; `tsc --noEmit` on `@cotal-ai/auth` exit 0. The independent
reviewer at the pre-merge head ran the suite four times including once under 12 concurrent CPU-burn
workers (green), instrumented a run externally (refusals on `warning` at +7306/+10308/+16318/+28321,
gaps 3002/6010/12003), and ran the mutation above against a rebuilt core dist, getting exactly one
red cell - the new one.

Test-only. No shipped source, no changeset.


davidfarah2003:
Review trail, recorded on the PR rather than left in mesh DMs.

**Verdict: APPROVE.** Issued by an independent reviewer (OpenAI family, `gpt-5.6-sol` at `xhigh`,
read from process metadata rather than self-report) against
`0c24aaa1a838df0ca9c4704837c50a333b9b00ca`.

What it ran, rather than reasoned about:

- The suite four times, including once under 12 concurrent CPU-burn workers, green each time. This
  suite has a 5s bearer TTL and a 22s watch window, so a cell that passes only on an idle box is not
  a passing cell.
- An externally instrumented run: the refusal appears on `warning` at +7306/+10308/+16318/+28321ms,
  gaps 3002/6010/12003ms.
- `tsc --noEmit` on `@cotal-ai/auth`, exit 0.
- **The mutation that matters**: it emitted the refusal on `error` in addition to `warning` and
  rebuilt core's dist. Result `7 passed, 1 failed`. Only `the recoverable refusal stays off the error
  channel` went red; the other seven cells stayed green. That is a predicted, legible kill set, which
  is what makes the added cell load-bearing rather than decorative.

It also traced the mechanism from source: `reestablishLoop` catches the refusal at
`endpoint.ts:1270-1275`, routes it through `emitRecoverable` (which emits `warning` at 695-697), and
waits `retryMs * 2 ** retryAttempt` capped at 60s - matching the measured 3s/6s/12s. And it confirmed
no other pre-existing cell still derives a verdict from `errorsA`.

**Why the verdict binds this head although it names an earlier SHA.** The head moved when
`origin/main` was merged in to resolve the conflict #1173 created. That is exactly the situation
where a transferred verdict is usually wrong, so it was decomposed rather than asserted:

```
graded file blob @ 0c24aaa1a : ecb930778998059e1d6aa49cc9deac7080943c44
graded file blob @ 3ccf8807a : ecb930778998059e1d6aa49cc9deac7080943c44   identical

paths differing between the two heads : 34
of those, differing from origin/main  : 0
paths differing between 3ccf8807a and origin/main : 1  (the graded file)
```

So this head is `origin/main` plus the exact file content that was graded. The other 34 paths are
main's own merges, each gated separately. The unreviewed surface is zero.

**A correction that belongs on the record.** This PR opened claiming to fix a red suite. #1173
landed that fix first, with one line pushing warnings into the same `errorsA` array. I then asserted
that the merged-channel form made the #1119 regression undetectable. Measuring it proved me wrong:

```
#1173 form  under the mutation: 6 passed, 1 failed - red cell "the wait between reconnect attempts grows"
this PR     under the mutation: 7 passed, 1 failed - red cell "the recoverable refusal stays off the error channel"
control (core restored):        8 passed, 0 failed
```

Both detect it. #1173 detects it incidentally, because double-counting each refusal corrupts the
computed gaps, so the *backoff* cell reddens and points a debugger at the retry logic instead of the
classification. The remaining value here is a legible kill set, not detection where there was none.
The PR was rescoped and retitled to say that rather than merging under its original claim.

PR #1174, opened 2026-08-30
fix(lang): a host stop inside a concurrency scope is not the scope's outcome

### LINKED ISSUE #862: RunReleased inside a concurrency scope is journalled as the scope's failure; resume replays L4000
## Where

Measured against the published `@cotal-ai/lang@0.30.1` walker (`WALKER_LANGUAGE_VERSION` "1"), driven in-process via `run` with a real `Journal` (explicit entry store) and explicit pins from `resolvePins`.

## The problem

`RunReleased` carries a guarantee the rest of the durability story leans on: it is raised before the next effect begins, it is not recorded as the run's outcome, and a resume continues from exactly where the journal says the run is. #772 states the law as implemented at the sequential seam: *"a returned reason raises the language-owned, uncatchable `RunReleased`, so the driver returns `released` without recording a program failure."*

That law does not hold inside a concurrency scope. When `shouldStop()` returns a reason while the walker is inside `parallel` (and equally `fanOut`), the release is settled into the journal as the **scope entry's failure**, carrying the release message as if it were the scope's outcome. A resume of the same journal with a healthy host and no `shouldStop` then replays that entry as a recorded failure:

```
EffectError code=L4000 msg=this run was released before its next effect: host stop
```

So a plain host stop — the one interruption the release mechanism exists to make safe — permanently poisons any run that happens to be inside a scope at that moment. The identical stop at a sequential seam resumes cleanly.

No handler misbehavior is involved: the handler is `SimHandler` and never throws. This is distinct from #733 (a handler **throw** settling `failed` and replaying) and from #734 (no handler-side release signal); here the language's own release signal is being converted into a journalled failure by the scope machinery.

## Measurement

Scoped program:

```js
const r = await parallel({
  a: async () => { await sleep("1s"); return 1 },
  b: async () => { await sleep("2s"); return 2 },
}, { name: "pair" })
log("done", r.a, r.b)
```

Sequential control:

```js
await sleep("1s", { name: "one" })
await sleep("2s", { name: "two" })
log("done")
```

Each is run with `shouldStop: () => (entries.length >= 2 ? "host stop" : undefined)` against a store that captures every journal append, then resumed with the same entries and pins and **no** `shouldStop`. Verbatim output:

```
scoped: stopped with RunReleased code=L5012
scoped: journal rows=4, states=parallel:pending sleep:pending sleep:settled parallel:settled
scoped: RESUME FAILED EffectError code=L4000 msg=this run was released before its next effect: host stop
---
sequential: stopped with RunReleased code=L5012
sequential: journal rows=2, states=sleep:pending sleep:settled
sequential: RESUME OK, steps=16
```

The tell is the fourth scoped row: the `parallel` scope entry is *settled* by the release. The sequential journal correctly leaves nothing settled-by-release, and resumes.

Full repro script:

```js
import { Journal, SimHandler, WALKER_LANGUAGE_VERSION, resolvePins, run } from "@cotal-ai/lang";

const scoped = `
const r = await parallel({
  a: async () => { await sleep("1s"); return 1 },
  b: async () => { await sleep("2s"); return 2 },
}, { name: "pair" })
log("done", r.a, r.b)
`;

const sequential = `
await sleep("1s", { name: "one" })
await sleep("2s", { name: "two" })
log("done")
`;

async function drill(label, src) {
  const pins = resolvePins({ runId: label, seed: `${label}-seed` }, Date.now(), WALKER_LANGUAGE_VERSION);
  const entries = [];
  const store = { async append(e) { entries.push(e); } };
  try {
    await run(src, {
      runId: label,
      handler: new SimHandler({}),
      journal: new Journal({ run: label, entries: [], store }),
      pins,
      shouldStop: () => (entries.length >= 2 ? "host stop" : undefined),
    });
    console.log(`${label}: completed without stopping (stop never hit)`);
  } catch (e) {
    console.log(`${label}: stopped with ${e.constructor.name} code=${e.code}`);
  }
  console.log(`${label}: journal rows=${entries.length}, states=${entries.map((e) => `${e.kind ?? "?"}:${e.state ?? e.status ?? "?"}`).join(" ")}`);
  try {
    const r = await run(src, {
      runId: label,
      handler: new SimHandler({}),
      journal: new Journal({ run: label, entries: [...entries], store: { async append(e) { entries.push(e); } } }),
      pins,
    });
    console.log(`${label}: RESUME OK, steps=${r.steps}`);
  } catch (e) {
    console.log(`${label}: RESUME FAILED ${e.constructor.name} code=${e.code} msg=${String(e.message).slice(0, 90)}`);
  }
}

await drill("scoped", scoped);
console.log("---");
await drill("sequential", sequential);
```

## Ask

A release inside a concurrency scope should behave exactly as it does at a sequential seam: stop the walk without settling any entry as the release's "failure". In-flight branch attempts should be left pending (their `external` bindings intact, so a later attempt re-binds via `ctx.resume`) rather than folded into a settled scope failure. At minimum, `RunReleased` must never be recorded as a step's outcome — that is its documented definition.

Until then, any host driving scoped programs cannot use `shouldStop`/pause at all without risking an unresumable journal; the only safe interruption is killing the process, which preserves pending rows and resumes correctly.

davidfarah2003:
Fixes #862.

`shouldStop` documents the law it keeps: a driver's reason to stop "must not be recorded as its outcome", and the run stays exactly where its journal says it is so the next driver resumes from there. That held at a sequential seam and not inside a concurrency scope.

`performScope` has a ladder of classes it refuses to record as a scope failure: `JournalAppendRejected`, `CloseOwed` and `Cancelled`. `RunReleased` was not on it, so a host stop landing inside `parallel` or `fanOut` fell through to the generic settle and wrote the release's own text as an `L4000` scope-fault on the scope entry. A resume with a healthy host and no `shouldStop` then replayed that entry and threw. The one interruption the release mechanism exists to make safe permanently ended any run that happened to be inside a scope, while the identical stop one statement earlier resumed cleanly.

`RunReleased` now joins that ladder and settles nothing, which is what leaves the run where it stands. A pending scope re-enters on resume and settling is idempotent, so the run finishes rather than carrying a durable failure. This is the same shape `CloseOwed` above it already relies on.

## Verification

Measured before the change, at the reported behavior:

```
SUBJECT - stop inside a parallel scope
journal entries: 2
  state=settled  error.code=L4000  msg="this run was released before its next effect: host stop"
RESUME THREW: EffectError code=L4000
```

After, with the sequential control unchanged:

```
CONTROL - stop at a sequential seam
  1 entry, settled, no error          RESUME: finished cleanly
SUBJECT - stop inside a parallel scope
  2 entries, pending + settled        RESUME: finished cleanly
SUBJECT 2 - stop inside a fanOut scope
  2 entries, pending + settled        RESUME: finished cleanly
```

The report was filed against published `@cotal-ai/lang@0.30.1`; this reproduces unchanged on the in-repo package at `0.35.0`, so it survived five minors.

## Test

Six cells added to `interpret.smoke.ts`, immediately after the existing sequential-release cell so the two sit side by side. Each scope is checked three ways: the release is raised, no entry records it as an outcome, and a healthy driver resumes past the scope it was stopped inside. `parallel` and `fanOut` both, because they share the path.

The added cells were proved to depend on the change rather than merely accompany it. With the new line in `perform.ts` disabled, the suite goes red on the named assertion and no other:

```
EXIT=1
Error: FAIL: and parallel records the release as NO entry's outcome - ["L4000"]
```

`pnpm test` in `packages/lang` is green, 19 suites.

## Scope

Two files. One line of behavior in `packages/lang/src/perform.ts` plus the comment recording why it is there, and the cells in `packages/lang/smoke/interpret.smoke.ts`. No spec change, no grammar change, nothing outside the lang package.

One thing deliberately not done. The other members of the same uncatchable set, `RunDivergence`, `ScopeBranchMissing` and `UnwalkableScope`, were not tested against a scope boundary. Whether a scope settles any of those as its own outcome is the same question with a different class, and it is worth asking separately rather than folding an untested widening into this fix.

### DIFF
--- a/.changeset/scope-release-not-outcome.md
+++ b/.changeset/scope-release-not-outcome.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/lang": patch
+---
+
+A host stop inside a concurrency scope no longer poisons the run. `shouldStop` returning a reason while the walker was inside `parallel` or `fanOut` settled the scope entry as a failure carrying the release's own text as an `L4000` scope-fault, and a resume with a healthy host then replayed that entry and threw. The one interruption the release mechanism exists to make safe permanently ended any run that happened to be inside a scope, while the identical stop at a sequential seam resumed cleanly. `RunReleased` now joins the classes a scope refuses to record as its outcome, so the scope stays pending and a resume re-enters and finishes it.
--- a/packages/lang/smoke/interpret.smoke.ts
+++ b/packages/lang/smoke/interpret.smoke.ts
@@ -1099,14 +1099,14 @@ let c = await sleep("1s")
 // Both scopes that launch thunks are covered, because they share the path. The resume is the point:
 // settling nothing leaves the scope pending, a pending scope re-enters, and settling is idempotent,
 // so the run finishes where it stopped rather than carrying a durable failure.
-for (const [label, P] of [
-  ["parallel", `
+for (const [label, scopeName, P] of [
+  ["parallel", "both", `
 await parallel({
   a: async () => { await sleep("1m"); return 1; },
   b: async () => { await sleep("2m"); return 2; },
 }, { name: "both" });
 `],
-  ["fanOut", `
+  ["fanOut", "each", `
 await fanOut([1, 2], async (n) => { await sleep("1m"); return n; }, { name: "each", key: (n) => "k" + n });
 `],
 ] as const) {
@@ -1127,15 +1127,25 @@ await fanOut([1, 2], async (n) => { await sleep("1m"); return n; }, { name: "eac
   ok(`a host stop inside ${label} releases the run rather than failing it`,
     released instanceof RunReleased, `${(released as Error)?.name}`);
 
-  // The defect was here: an entry settled `failed` with the release text as its error.
+  // The defect was here, and this reads the JOURNAL'S OWN SHAPE rather than an error message. A
+  // message match would go quietly green the day the release text is reworded, while the bug it
+  // guards returned; the durable claim is that this scope settles NOTHING. Under the reverted fix
+  // the same entry is `settled` with `status: "failed"` and `error.code: "L4000"`, so each of the
+  // three conditions below is one the defect actually breaks.
   const entries = journal.entries();
-  const recordedRelease = entries.filter((e) => {
-    const err = (e as { error?: { message?: string } }).error;
-    return err !== undefined && String(err.message).includes("released before its next effect");
-  });
-  ok(`and ${label} records the release as NO entry's outcome`,
-    recordedRelease.length === 0,
-    recordedRelease.map((e) => (e as { error?: { code?: string } }).error?.code));
+  const scope = entries.find((e) => e.kind === label && e.name === scopeName);
+  ok(`and the ${label} scope entry stays PENDING rather than recording the release as its outcome`,
+    scope !== undefined
+      && scope.state === "pending"
+      && (scope as { status?: string }).status === undefined
+      && (scope as { error?: unknown }).error === undefined,
+    scope === undefined
+      ? "no scope entry at all"
+      : {
+          state: scope.state,
+          status: (scope as { status?: string }).status,
+          error: (scope as { error?: { code?: string } }).error?.code,
+        });
 
   // And the consequence that made it critical rather than cosmetic.
   let resumeErr: unknown;
--- a/packages/lang/smoke/interpret.smoke.ts
+++ b/packages/lang/smoke/interpret.smoke.ts
@@ -1147,13 +1147,25 @@ await fanOut([1, 2], async (n) => { await sleep("1m"); return n; }, { name: "eac
           error: (scope as { error?: { code?: string } }).error?.code,
         });
 
-  // And the consequence that made it critical rather than cosmetic.
+  // And the consequence that made it critical rather than cosmetic. NOT resuming without throwing:
+  // that alone passes for an implementation that settles the scope `ok` with an invented value and
+  // then releases, which finishes the resume while silently skipping the arm that never ran. So the
+  // claim is what the journal HOLDS afterwards: the scope settled, and BOTH arms performed, one of
+  // which had not run when the host stopped. Measured on the correct implementation: three entries,
+  // all settled, two of them the branch effects.
   let resumeErr: unknown;
+  let after: readonly JournalEntry[] = [];
   try {
-    await resume(P, new Journal({ run: runId, entries }), { runId, pins, handler: new SimHandler({}) });
+    const r = await resume(P, new Journal({ run: runId, entries }), { runId, pins, handler: new SimHandler({}) });
+    after = r.journal.entries();
   } catch (e) { resumeErr = e; }
   ok(`so a healthy driver resumes past a ${label} it was stopped inside`,
     resumeErr === undefined, `${(resumeErr as Error)?.name}: ${(resumeErr as Error)?.message}`);
+  ok(`and the ${label} arm that had not run when the host stopped is performed on that resume`,
+    after.length === 3
+      && after.every((e) => e.state === "settled")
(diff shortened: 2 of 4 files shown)


davidfarah2003:
Expect shard 0 to fail on this branch, and not because of anything in it. Recording that here before it happens, so the red is not read as a finding against this change.

This branch is based on `00ac9d97d`, which carries the `standaloneConnectOpts` census fix. That is what lets shard 0 run deep enough to reach `pnpm smoke:expiry-renewal:auth` at line 639 of 664 in `ci-suites.txt`. That suite fails on `main` today, 5 cells passing and 2 failing:

```
✗ the refusal to dial is loud and names the missing renewal path
✗ the wait between reconnect attempts grows past the flat retry interval
    { attemptsAtMs: [], gaps: [] }
```

It is fixed by #1172, which is open and not merged. On that branch the same shard reports `✓ smoke:ci shard 0/4 passed (120 smokes)` with `EXPIRY-RENEWAL SMOKE OK (8 passed, 0 failed)`. So the failure is upstream of this PR in both senses: it is on the base, and someone else already owns the fix.

Nothing here touches auth, the CLI, or any broker path. The change is one line in `packages/lang/src/perform.ts` plus cells in `packages/lang/smoke/interpret.smoke.ts` and a mutation config beside them. `packages/lang` has no dependency on anything else in the repo.

What is worth checking when the run finishes is not the shard's colour but whether it reached and passed `pnpm smoke:lang-interpret`, which is the suite this change is proved by. That one runs in a different position and is unaffected by the auth failure ahead of it, unless the shard stops first, in which case the honest reading is that this PR is unproven by CI rather than failed by it. The suite is green locally at this head, 97 checks, and both registered mutants are killed.

If #1172 lands before this does, a rebase makes the whole question go away.

davidfarah2003:
Two independent reviews at `68e9b0171`, recorded here so the verdicts survive outside the chat they were given in. Both reviewers re-verified first-hand rather than accepting my report, and both found real defects in the first version.

**Correctness and durability: APPROVE.** Verified in an owned detached worktree at that sha. `interpret.smoke` 97/97; `mutation-proof` killed both registered mutants, each red on the named cell; `perform.ts` confirmed restored afterwards.

**Evidence and test quality: APPROVE.** Both earlier blockers resolved. Additionally verified through the built public artifact rather than the source tree: importing `packages/lang/dist/index.js`, the public `run` returned `RunReleased(host stop)` with the journal showing a pending scope carrying no status and no error, and the public `resume` produced a settled scope plus two settled sleeps. That closes the gap `mutation-proof` names in its own output, which is that killing a mutant proves the suite depends on the code and not that a real entry point reaches it.

What the reviews changed, because the first version of this PR was weaker than I claimed.

The middle cell matched an error message. Rewording the release text would have turned it green while the defect returned. It now finds the scope's own entry by kind and name and asserts it is pending with no status and no error.

More seriously, deleting the guard was never the only wrong implementation. One reviewer built another: settle the scope `ok` with a fabricated value, then release. It raises the release, prints no `L4000`, and resumes without throwing, and every cell in the first version passed under it while the arm that never ran was silently skipped. The resume cell now asserts the journal holds three settled entries with both branch effects, which is the measured shape of a correct resume, so "the resume did not throw" is no longer the claim.

Both mutants are registered in `packages/lang/smoke/mutations/uncatchable-scope-release.json`, so neither proof is a one-time manual check.

One thing deliberately left out, with both reviewers agreeing. No `race` cell was added. `callScope` sends `parallel`, `fanOut` and `race` through the same `performScope`, and a reviewer ran a race fixture at this sha and measured the identical journal shape, so a copied cell would be true but would cover the language rather than the one line this PR changes.

A follow-up is filed separately rather than folded in here: a `RunDivergence` raised inside a live or resumed scope still falls through to the same generic `L4000` settle. Same shape, predates this change, and adding `RunReleased` to the ladder does not make it worse.

These are agent reviews. They are evidence about the code, not an approval on anyone's behalf.

Issue #1169, opened 2026-08-30
smoke:jcode-provider-disconnect red at 'post-recovery turn' on two independent branches


davidfarah2003:
The `post-recovery turn` cell of `smoke:jcode-provider-disconnect` (shard 2/4) failed identically on two branches with unrelated content:

- PR #1038 (spawn-lifecycle docs/fragment fold), run 33318085124 attempt 1
- PR #1041 (persona-role capability guard), run 33319699947

Same suite, same cell, no shared branch content touching the connector — which points main-side (deterministic or high-rate flake), not at either branch. Failed-job reruns are in flight on both as the flake discriminator.

Pure-main evidence is about to exist: main run 33321744283 on c4c06a8a is the first main run where shard 2 is reachable again (shard 0's mutation-fixtures + required-arg-seam layers are cured as of #1161/#1168), so it will show this cell's true state on unmodified main. I'll append its verdict here.

Possibly related instability in the same suite: on a mac workstation, pure main reds EARLIER at `synthetic transient recovery attach failure` — a different cell of the same recovery window (env-dependent, noted for completeness; the CI cell is the one this issue tracks).

Umbrella: #1155.

davidfarah2003:
Data point: on PR #1038's run (33318085124), `post-recovery turn` failed on attempt 1 and **greened on attempt 2 against the identical merge ref** — so on that ref the cell is a flake, not deterministic. #1041's rerun (blocked behind its queued ci-ok) and main run 33321744283 will add two more samples.

davidfarah2003:
Linux reproduction achieved: looping `pnpm smoke:jcode-provider-disconnect` on a netcup box at origin/main reproduces `✗ post-recovery turn` in roughly 4 of the first 40 iterations (~10%), failing ~25s in with a timed-out wait for a named event — matching CI's fail-then-green rerun behavior. The diagnosis lane is now isolating the race; fix PR with pre/post loop counts to follow.

davidfarah2003:
[...]

PR #1175, opened 2026-08-30
test(jcode): await the redriven turn before post-recovery work

davidfarah2003:
`pnpm smoke:jcode-provider-disconnect` fails the `post-recovery turn` cell as a race, not deterministically: CI red it on two unrelated branches, and the same merge ref greened on rerun.

## Root cause

A suite defect. The cell polls the fake Harness log for a `send_message` frame carrying `RECOVERED_MESH_WORK`, but the suite sends that work without waiting for the redriven stalled turn to end.

When the redriven turn is still live, the connector takes its documented mid-turn path: a directed DM goes into Jcode's soft-interrupt queue via `steerPending`, and its id commits at the containing turn's clean boundary. The seat accepts the work, but it arrives over a `soft_interrupt` frame, so no `send_message` ever carries that content and the poll times out after 20s. A captured failing run shows the redrive `send_message`, then both `RECOVERED_MESH_WORK` unicasts as `soft_interrupt`, then the redriven turn's boundary.

Both delivery paths are correct connector behavior. Only the observation was timing-dependent, so the fix belongs in the suite.

## Fix

Gate the post-recovery unicast on the redriven turn's boundary, which is the same guard the next cell already applies before the second stall (added in 8e2dee9f6 for this same class of race, one step later in the sequence).

The earlier `RECOVERED_MESH_WORK` unicast is dropped rather than moved. It was never the redrive trigger: the recovery window's own `finally` redrives the un-acked batch because `pendingWake()` stays positive, which the traces confirm (the redrive injection carries exactly one message). Keeping it would leave a second way into the same race, where it lands after the boundary and opens a turn that then swallows the post-recovery work.

## Repro counts

Both loops ran 40 iterations of `pnpm smoke:jcode-provider-disconnect` on the same linux box under the same conditions.

| | iterations | failures | cell |
| --- | --- | --- | --- |
| pre-fix | 40 | 4 | `post-recovery turn` |
| post-fix | 40 | 0 | n/a |

An earlier pre-fix loop on the same box added 6/40 failures, all in the same cell, for 10/80 pre-fix overall.

## Not vacuous

An instrumented post-fix run confirms the cell still exercises the real path: the redrive `send_message` and its boundary, then a genuine new `send_message` carrying `RECOVERED_MESH_WORK` with its own boundary, and no `soft_interrupt` frames. `retriedTurn` still reaches its required 2 attempts.

`pnpm typecheck` and `pnpm check:docs-voice` pass. Smoke-only change, so no changeset, matching 1ae13ce52 and 8e2dee9f6.

Refs #1169

### DIFF
--- a/extensions/connector-jcode/smoke/jcode-provider-disconnect.smoke.ts
+++ b/extensions/connector-jcode/smoke/jcode-provider-disconnect.smoke.ts
@@ -173,7 +173,6 @@ try {
   });
   check("recovered Harness client reattaches the existing private session after the transient loss (#971)", reattachments.length >= 2, reattachments);
 
-  await operator.unicast(peerId!, "RECOVERED_MESH_WORK");
   const retriedTurn = await waitFor("unacknowledged stalled turn redelivery", () => {
     const attempts = entries().filter(
       (entry) =>
@@ -186,6 +185,15 @@ try {
   });
   check("recovered seat redrives the unacknowledged stalled turn (#781)", retriedTurn.length >= 2, retriedTurn);
 
+  // The redriven turn owns the session until its boundary, so the post-recovery work is only sent
+  // once that turn is done. A directed DM arriving while it is still live is delivered through
+  // Jcode's soft-interrupt queue instead, which is the documented mid-turn path and commits at the
+  // containing turn's boundary — a correct delivery that never opens the new turn this cell reads.
+  await waitFor("redriven stalled turn boundary", () =>
+    entries().find(
+      (entry) => entry.ev === "turn_done_emitted" && String(entry.content).includes("SIMULATE_PROVIDER_STALL"),
+    ),
+  );
   await operator.unicast(peerId!, "RECOVERED_MESH_WORK");
   const recoveredTurn = await waitFor("post-recovery turn", () =>
     entries().find(


davidfarah2003:
CI classification (run 33326606985): attempt 1's only red was `smoke:spawn-from-anywhere:live` dying at broker startup (`broker nats://127.0.0.1:44085 never reached auth-required`) — an infra timeout on a starved runner, passing 15/15 locally on the same tree. Attempt 2: every job green. Merging on a fully green rerun.

davidfarah2003:
Correction to the classification above — my "attempt 2: every job green" was wrong; the rerun's shard 0 red arrived in a delayed notification. Accurate record:

| attempt | shard-0 signature | class |
|---|---|---|
| 1 | `spawn-from-anywhere:live` — broker never reached auth-required | infra timeout on a starved runner (passes 15/15 locally on the same tree) |
| 2 | `smoke:mutation-fixtures`:
> ✗ [5] ANCHOR SPANS A COMMENT — 1 comment line(s) in `find`
> ✗ [6] ANCHOR SPANS A COMMENT — 1 comment line(s) in `find`
> ✗ [6] ANCHOR IS DEAD — `find` matches 0x in packages/workspace/smoke/fixtures/presence-render-sinks.json
> ✗ [8] ANCHOR IS DEAD — `find` matches 0x in packages/workspace/smoke/fixtures/presence-render-sinks.json | main-side anchor deaths on this run's stale merge ref, cured on main by #1178 (3bb04385), which postdates the ref |

Every other job (live, shards 1–3, unit) green on attempt 2. Both shard-0 reds are unrelated to this PR's test-only diff, so the merge stands — but on the corrected grounds (main-side classes), not on a green rerun.

Issue #773, opened 2026-08-21
manager renews one store while remote delivery reloads another, so every daemon-credential adoption is refused


davidfarah2003:
## Summary

In a static-auth mesh whose manager and delivery daemon run on different hosts/filesystems, the manager re-signs `delivery.creds` and `membership-rw.creds` in the manager workspace store, then asks the remote delivery daemon to reload from the daemon's own workspace store. The admin request carries only the expected fingerprints, not the re-signed credentials or a shared-store coordinate. The daemon therefore re-reads different bytes and refuses every renewal.

The files remain structurally healthy while their 24-hour JWT window continues to expire. At the measured point the daemon-store generation had about 14.2 hours left. The manager process was alive; it was the renewal-owner path itself that was failing.

Do not restart a live manager merely to test this: it owns its managed seat processes. This needs a hermetic two-root reproduction.

## Exact live failure

The running manager's renewal record and log reported:

```text
! credential renewal: daemon adoption failed: delivery: reloadCreds: re-read credential generation did not match the expected re-signed generation (a different store, or a torn/stale read); nothing adopted; membership: membership feed is not running, but a membership generation was expected - nothing adopted: the membership bundle is incomplete here (missing membership-rw.creds) - a space created before broker-sourced membership gains the feed only when its auth is regenerated; `cotal up --rotate-sys` re-mints the $SYS pair but writes neither the rw cred nor the account id
```

The same error repeated on renewal passes. `cotal doctor auth` at the manager root reported both class-2 files healthy but ended non-zero with `last renewal not broker-accepted`.

A doctor run from a different cwd re-signed a different workspace store and did not affect this mesh. That is not evidence the renewal succeeded: `doctor auth --fix` resolves its root from cwd and writes only that root.

## Why it happens in current source

The source states a same-store requirement but the stock cross-host composition cannot satisfy it:

- `implementations/manager/src/manager.ts` defaults `this.secrets` to `workspaceSecretStore(this.workspaceRoot)`. `renewDaemonCreds()` calls `remintDaemonCreds(this.workspaceRoot, this.space, this.secrets, ...)`, then sends `reloadCreds` with per-component expected fingerprints.
- `implementations/manager/src/commands.ts` constructs `Manager` without injecting a store. Its workspace root is resolved from the manager process cwd.
[...]

PR #1177, opened 2026-08-30
test(delivery): reproduce the cross-host two-root renewal refusal (#773)

davidfarah2003:
**This reproduces #773. It does not fix it, and it should not be merged as one.** The architectural choice #773 asks for is still open, and this deliberately does not preempt it.

## Scope

#773 states that a two-root cell is required and that, before any fix, it must reproduce the exact generation-mismatch refusal. That cell did not exist. This adds it.

The existing renewal suite could not observe the defect, for two independent reasons:

- every phase re-signed the new generation into `credsPath`, which lives in the daemon's own workspace root, so the renewal owner's store and the daemon's store were the same file;
- it never constructs a `Manager` at all, so the renewal owner's path was never exercised end to end. The suite plays the owner with a supervisor-cred endpoint on the admin rail, which is the right shape, but the store question never arises.

The suite was green throughout. It was testing something true of the tree in front of it and false of the composition the issue is about.

## What the cell does

The renewal owner writes its re-signed generation into a separate workspace root, which is what a manager on another host does, and sends the fingerprint, which is all `reloadCreds` transports. The daemon re-reads its own root, finds the generation it already held, and refuses.

A second identical pass then asserts the refusal is permanent. That assertion is the point rather than a formality. The message covers "a different store, or a torn/stale read" without separating them, and those two conditions have opposite remedies: a torn read converges on the next attempt, two unshared stores never do. Retrying is not a repair here, and the repeated failure is what distinguishes them.

The candidate is rejected before the preflight, so the resident connection is untouched and the phases after this one still run against a live daemon.

## The control, and why it is in this PR

`expected` was covered in-process by `packages/core/smoke/membership-rw-renewal.smoke.ts` case 3, which calls `reloadRwCreds()` directly. Nothing proved the fingerprint survives the delivery-admin rail. This adds a cell asserting that a value which cannot match anything is refused over the rail. It passes.

It is here because without it the two-root result is unreadable. A refusal proves nothing if `expected` is honoured, and an adoption proves nothing if it is not.

## Evidence that the cells are load-bearing

Not a mutation run, and I am not claiming one. While writing this I twice sent an `expected` the handler could not read: first as a bare string rather than the per-component `{ delivery?, membership? }` map, then as a generation that came out byte-identical to the daemon's because two mints of one identity with the same default TTL inside a single second produce the same JWT. In both cases the daemon replied `ok: true` with a full `brokerAccepted` block and the cells went red. So the cells demonstrably depend on the fingerprint being honoured end to end, not on wording. The second case is also why the cross-host generation is minted with a distinct TTL, and the comment says so.

Worth noting for whoever takes the fix: a malformed `expected` reaches `(req.args?.expected ?? {}) as { delivery?, membership? }` and every key reads `undefined`, which the handler documents as "no expectation" and adopts. That is intended for a genuinely absent expectation. It also means a renewal owner with a shape bug records a green adoption that verified nothing. The stock manager builds the shape correctly, so this is not reachable through it today, and I am not filing it separately on that basis.

## Verification

`pnpm smoke:delivery-renewal` at `8042f96bb`: **21 passed, 0 failed.**

One earlier run also failed `evictPrincipal force-drops the victim` with `kicked: 0, "principal not currently live"`. It passed on every subsequent run with no change to that phase, so it looks like an existing race between the victim connecting and CONNZ attributing it. Unrelated to this change, and flagged rather than touched.

No changeset. The change is smoke-only and no published package's behaviour moves.

## Review

Needs two reviewers from different vendors before it goes anywhere, per standing policy. Two things are worth a reviewer's attention specifically: whether the two-root cell faithfully models the stock cross-host composition rather than a shape that only looks like it, and whether a suite asserting the current broken behaviour is the right artifact to carry until the fix lands, given those assertions must be inverted by whoever fixes it.

### DIFF
--- a/implementations/delivery/smoke/delivery-cred-renewal.smoke.ts
+++ b/implementations/delivery/smoke/delivery-cred-renewal.smoke.ts
@@ -13,6 +13,10 @@
  *      "file written" can never masquerade as "daemon adopted".
  *   4. CLEAN SWAP: a final explicit reload before the old JWT's exp — the run never logs
  *      "User Authentication Expired" and ends with a READY delivery lease.
+ *   4b. CROSS-HOST TWO-ROOT REFUSAL (#773): the renewal owner re-signs into a SEPARATE workspace
+ *      root, as a manager on another host does, and sends only the fingerprint. The daemon re-reads
+ *      its OWN root and refuses. The refusal is proven PERMANENT, which is what separates this from
+ *      the torn/stale read the same message also covers.
  *
  * NOTE: runs the BUILT dist — `pnpm build` first (the smoke:ci wiring builds).
  * Run: pnpm smoke:delivery-renewal   (needs `nats-server` on PATH; auth/JetStream, local-only; ~30s)
@@ -29,6 +33,7 @@ import {
   identityFromCreds,
   isReachable,
   createSpaceAuth,
+  credsFingerprint,
   mintConnectionEvictorCreds,
   mintCreds,
   mintMembershipObserverCreds,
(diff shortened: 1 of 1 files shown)


davidfarah2003:
BLOCK

Graded SHA: e17efa969bc89d5b8234656477e9fa2be511d4a6
COTAL_MODEL: (unset)
1177r-e17efa969

This is a test-only change at that SHA (`implementations/delivery/smoke/delivery-cred-renewal.smoke.ts` only). I ran `pnpm smoke:delivery-renewal` in this worktree (the script builds, then runs). Result: 21 passed, 0 failed. `pnpm typecheck` was clean. The new cells are labeled `#773` and encode the generation-mismatch refusal as the expected outcome, so the suite is green on an unfixed tree by design rather than by silence.

Blockers (honesty of the reproduction, not whether `reloadCreds` rejects a bad fingerprint):

1. The second root is not a participant. Phase 4b writes `managerRoot/.cotal/delivery.creds` and never reads it again. Nothing under `Manager`, `remintDaemonCreds`, or the daemon's store looks at that directory. The refusal is produced only by sending `expected.delivery` for bytes that were never written to the daemon `credsPath`. That is the same mechanism as the preceding all-zero fingerprint control. Delete the `managerRoot` write and the named `#773` checks still pass. A two-root cell has to make both stores observable (at least: the expected digest is of the other root's file, and the daemon root's file differs).

2. The stock composition in #773 is `Manager.renewDaemonCreds()`: remint through the manager workspace store, then `reloadCreds` with per-component fingerprints, while the delivery process re-reads a different cwd root. This cell still plays the owner as a supervisor-cred endpoint and never constructs a `Manager`. Issue #773 asked for a two-root, two-process cell of that composition before any fix. A dummy directory plus a rail call is a shape that looks like two-root and does not exercise the owner path that fails live.

Non-blocking notes:
- Membership is part of the live #773 error (`membership` expected against a feed/store the daemon does not share). This cell sends only `expected.delivery`.
- Throw handling is fail-closed (`catch` increments `fail` and sets `process.exitCode = 1`). `finally` does not `process.exit`, so a throw cannot vanish into a green exit.
- `bin/smoke/ci-suites.txt` is unchanged. `smoke:delivery-renewal` was already listed (not inserted by this PR).
- Isolated tmpdirs, local `nats-server` + `cotal deliver`, no `cotal up` and no shared `~/.cotal`.

davidfarah2003:
Superseded by #1185, which closes this PR's review blockers: the renewal pass is driven by the shipped Manager (no hand-sent fingerprint), both roots are asserted at the byte level, and the membership component's refusal is covered by its own cell against a live membership feed. Closing in its favor.

PR #1181, opened 2026-08-31
fix(connector-jcode): pool the MCP bridge across seat sessions

### LINKED ISSUE #844: connector-jcode: repeated subagents leak MCP bridge children until seat teardown
## Summary

A long-running Jcode seat accumulated one sleeping `connector-jcode/dist/mcp.js` child per repeated subagent/swarm lifecycle. The children remained alive after the subagent work ended, growing the private process tree and resident memory while the mesh seat still appeared healthy.

## Measured incident

On 2026-08-24, `fm-hotupdate-4` ran Jcode 0.78.1-dev with a private Cotal MCP bridge. Before its managed stop, the exact seat-owned tree contained:

- 66 sleeping `connector-jcode/dist/mcp.js` children
- about 2.0 GiB aggregate RSS
- 270 MCP start/connect events in the Jcode log
- no matching MCP stop/shutdown events

The tree was seat-exact: a later managed stop removed all 76 recorded processes without broad matching. Jcode was upgraded to v0.80.0 afterward, but the replacement deliberately used no swarm; v0.80.0 has therefore **not** been shown to fix the per-subagent leak.

## Why it matters

The connector gives every Jcode seat a private MCP bridge carrying its Cotal tool surface. Linear child growth means an otherwise useful long-running seat can consume gigabytes and eventually destabilize the host. A clean whole-seat stop is necessary but not sufficient: the leak occurs while the seat is meant to stay alive.

## Required diagnosis

Run a controlled v0.80.0 private seat with isolated JCODE_HOME/XDG and a fixed baseline descendant/RSS census. Create and retire repeated subagent/swarm members while recording:

- MCP child start/connect/stop events
- exact descendant PID/session/group membership
- RSS after each member retires
- whether the leaked process is owned by Jcode upstream, the SDK launch seam, or Cotal connector configuration

Do not infer a fix from the whole-seat teardown path in #839.

## Acceptance

- A red reproduction demonstrates linear MCP-child growth on the affected path.
- After the fix, each retired subagent releases its MCP bridge within a bounded interval; descendant count and RSS return to a stable baseline across at least 25 cycles.
- Concurrent members do not kill or reuse another member's bridge.
- The standing seat remains connected and its primary Cotal MCP bridge stays callable throughout.
- Managed seat stop still proves the entire private tree gone, with exact PID/group targeting and no name sweep.
- If the defect is upstream Jcode and cannot be repaired in the connector, Cotal must fail loud or disable the leaking subagent mode rather than silently allowing unbounded growth.

davidfarah2003:
Fixes #844.

## What changed

The Jcode MCP bridge entry in the per-peer `mcp.json` is now written `shared: true`. Jcode pools `shared: true` MCP servers in its daemon and reuses one bridge process across every session the daemon runs; under `shared: false` (the previous state) it spawned a fresh bridge per session and stopped none of them before seat teardown. A seat running repeated subagent sessions therefore accumulated bridge processes without bound — the incident on #844 measured 66 children with 270 starts and 0 stops on one seat, and 143 live bridges totalling 4.8 GB at the moment the box OOM-killed both orchestrators.

Pooling stays inside the seat. The daemon the pool lives in is private to the seat: its own `JCODE_HOME` (0700), its own socket, and its own 256-bit relay token, all written by the same `writeMcpConfig`. One seat's bridge is never visible to another seat's daemon.

## Proof

- `pnpm smoke:jcode-host` green, 56 checks, including the new cell `bridge entry rides the per-seat daemon pool instead of spawning per session`. The cell reads the `mcp.json` that the shipped writer produced during a real host launch, not a hand-built restatement.
- Mutation ledger (`extensions/connector-jcode/smoke/mutations/jcode-launch.json`): all 6 mutations killed. The new mutation (`shared: true` → `shared: false`) goes red on exactly its named cell (11 marks against a 56-mark baseline).
- `pnpm typecheck` green.
- The pooling semantics were verified against jcode's published source (`crates/jcode-base/src/mcp/manager.rs`, quoted on #844): the pool is keyed inside the daemon, and the daemon here is per-seat, so reuse cannot cross seats.

## Named gap

There is no live green-arm experiment (a bounded swarm run under `shared: true` showing a flat bridge count). Running one requires either the live installed connector or bringing up an isolated stack on the fleet host, and neither is safe to do from this lane. The red arm is the incident itself: the #844 measurements above reproduce the defect this change removes. The config-level contract — the entry is written `shared: true` and a regression back to `shared: false` fails a named cell — is what the suite proves.

## Release

One changeset, patch for `@cotal-ai/connector-jcode` (the lockstep group versions the rest).

### DIFF
--- a/.changeset/jcode-bridge-pool.md
+++ b/.changeset/jcode-bridge-pool.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/connector-jcode": patch
+---
+
+The Jcode MCP bridge entry is written `shared: true`, so the per-seat daemon pools one bridge process and reuses it across sessions. Under `shared: false` every subagent session spawned its own bridge and none stopped before seat teardown, so seats running repeated subagents accumulated bridge processes without bound. Pooling stays inside the seat: the daemon, its home, its socket, and its relay token are all private to the seat.
--- a/extensions/connector-jcode/smoke/jcode-host.smoke.ts
+++ b/extensions/connector-jcode/smoke/jcode-host.smoke.ts
@@ -361,8 +361,12 @@ try {
   // into the host's MeshAgent. A schema-valid explicit `peek:false` must not be rejected by either
   // hop: the bug was an adapter-only no-argument branch that advertised this input then refused it.
   const privateMcp = JSON.parse(readFileSync(join(peerHome, "mcp.json"), "utf8")) as {
-    servers: { cotal: { env: Record<string, string> } };
+    servers: { cotal: { env: Record<string, string>; shared?: boolean } };
   };
+  // Jcode spawns a `shared: false` server once per session, so under a seat that runs subagents
+  // every member session leaves its own bridge process alive until seat teardown. The bridge must
+  // ride the daemon's pool; the daemon is seat-private, so pooling cannot cross seats.
+  check("bridge entry rides the per-seat daemon pool instead of spawning per session", privateMcp.servers.cotal.shared === true, privateMcp.servers.cotal);
   const relaySocket = privateMcp.servers.cotal.env.COTAL_JCODE_MCP_SOCKET!;
   const relayToken = privateMcp.servers.cotal.env.COTAL_JCODE_MCP_TOKEN!;
   await operator.multicast("quiet buffered", { channel: "team" });
--- a/extensions/connector-jcode/smoke/mutations/jcode-launch.json
+++ b/extensions/connector-jcode/smoke/mutations/jcode-launch.json
@@ -39,6 +39,14 @@
       "replace": "if (false) {",
       "command": "pnpm smoke:jcode-host",
       "expectRed": "Jcode cotal_inbox refuses JSON-own __proto__ without consuming the buffered message"
+    },
+    {
+      "name": "Jcode bridge rides the daemon pool instead of spawning per session",
+      "file": "extensions/connector-jcode/src/host.ts",
+      "find": "        shared: true,",
+      "replace": "        shared: false,",
+      "command": "pnpm smoke:jcode-host",
+      "expectRed": "bridge entry rides the per-seat daemon pool instead of spawning per session"
     }
   ]
 }
--- a/extensions/connector-jcode/src/host.ts
+++ b/extensions/connector-jcode/src/host.ts
@@ -134,7 +134,11 @@ function writeMcpConfig(home: string, relay: RelayEndpoint, config: AgentConfig)
           COTAL_JCODE_MCP_TOKEN: relay.token,
           COTAL_JCODE_MCP_CONFIG: JSON.stringify(publicConfig(config)),
         },
-        shared: false,
+        // Pooled (Jcode's default): one bridge per daemon, reused across every session in the
+        // seat, so repeated subagent sessions cannot each leave a bridge process behind until
+        // teardown. Pooling never crosses seats — the pool lives in the daemon and this daemon is
+        // private to the seat (its own JCODE_HOME, socket, and relay token above).
+        shared: true,
       },
     },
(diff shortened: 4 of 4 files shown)


davidfarah2003:
## Verdict: APPROVE

Graded SHA: `e2cba276778049e88d628cc89db29666470f1ad9`
COTAL_MODEL: unset
Stamp: `1181r-grok-e2cba2767`

Adversarial security review of the pooling change only (`shared: false` -> `shared: true` on the per-seat `cotal` MCP entry).

### Cross-seat
Each shipped isolation leg holds. The daemon is launched into a seat-private `JCODE_HOME` (`privateAgentHome`: 0700, `hardenPrivate`, no symlink/escape). Jcode's MCP pool is a per-daemon singleton, not a machine-global one. The API sockets live under a 0700 owner-checked `/tmp/jc-*` alias. The relay token is 256-bit, compared with `timingSafeEqual`, and is written only into owner-only `mcp.json` (0600, `O_EXCL`). Resume across seats is refused. I found no path that puts two distinct seats on one daemon except the same space+name+`COTAL_JCODE_HOME` (the same agent identity). The Cotal relay Unix socket in `/tmp` is mode 0775 and not chmod'd after bind; connecting still requires the token, and that layout predates this PR.

### Within-seat
The reused process is a stdio relay. Inbox cursors, buffered mesh messages, and broker credentials stay in the host `MeshAgent`. The child holds `publicConfig` (creds stripped), the relay socket path, and the token. Upstream `McpHandle` clones share a monotonic request id and a pending map, with a single stdin writer task, so session B does not receive session A's JSON-RPC result. Session B can invoke `cotal_*` as the seat because every session in that daemon already shares one mesh identity through the same host relay; that was true under `shared: false` as well (same `mcp.json` env). Pooling does not add a new cross-session residue channel for identity, inbox, or auth.

### Teardown
Seat stop still closes the relay and `stopPrivateTree`s the launch-identity tree, which includes the pooled child. Ending a subagent only drops a pool refcount, which is the leak fix. A mid-call crash times out that request; bridge recovery relaunches the daemon against the same `mcp.json` and still-live relay.

No named blockers.

davidfarah2003:
APPROVE

Graded exact head `e2cba276778049e88d628cc89db29666470f1ad9` and nothing else.

Model line: `COTAL_MODEL=<unset>`
Round token: `1181r-gpt-e2cba2767`

Engineering and spec-fidelity findings:
- `writeMcpConfig` is the sole shipped Jcode MCP-config writer and emits the sole `cotal` bridge entry with `shared: true`. Every supported connector launch reaches that writer before the private daemon launches, and recovery stays within that configured private daemon.
- The new smoke cell reads the managed `mcp.json` produced by a real `host-main.ts` launch. It does not reconstruct the config by hand. `pnpm smoke:jcode-host` passed all 56 checks.
- The mutation anchor is unique. The named assertion and `expectRed` target are unique. Mutation proof killed all 6 mutations; the new `shared: true` to `shared: false` mutation died on its named cell after 11 marks against the 56-mark baseline.
- Jcode's published source independently confirms the claim: shared servers use the one pool per daemon and are reused across sessions, while non-shared servers are per-session owned clients. The Cotal bridge has no session-cwd dependency, and this connector's daemon, home, socket, and relay capability are seat-private.
- The PR honestly names the missing live green-arm swarm run and limits its proof to the generated config contract, mutation discrimination, and published Jcode semantics. That gap does not invalidate the narrower pooling claim made by this change.
- The merge-base diff is minimal at 4 files and 23 additions / 2 deletions. The connector patch changeset is correct for the fixed lockstep group. `pnpm changeset status`, `git diff --check`, and `pnpm typecheck` passed.

No blockers found.

PR #1185, opened 2026-08-31
test(manager): reproduce the two-root renewal adoption refusal (#773)

davidfarah2003:
Reproduces #773 (the two-root renewal adoption refusal) through the shipped renewal owner, end to end, in a hermetic smoke suite: `bin/smoke/manager-two-root-renewal.smoke.ts`, wired as `smoke:manager-two-root-renewal` and appended to `ci-suites.txt` at the end of the file so every existing shard assignment is unchanged.

## What it drives

- A throwaway authed broker (fresh operator chain, sandboxed `COTAL_HOME`, ambient `COTAL_*` scrubbed, kills only the PIDs it spawns).
- A real delivery daemon (`tsx bin/cotal.ts deliver`) rooted at its OWN workspace root B.
- A real `Manager` rooted at a DIFFERENT workspace root A, same space. `Manager.start()` runs the initial renewal pass itself; the suite never hand-sends a fingerprint or drives any rail directly.

The manager re-signs both daemon creds through its own root's store and requests adoption; the daemon re-reads its own divergent store, and adoption is refused with the exact generation-mismatch error on BOTH components (delivery and membership), recorded in `renewal.json`.

## How the defect is encoded

The defective behavior is asserted as explicitly-labeled `#773 unfixed:` expectations: green on today's tree, red the moment #773 is fixed (whether the fix refuses the divergent composition at startup or makes adoption succeed through one shared authority). The fix PR must flip those cells to the fixed expectations; the header comment states this contract.

## Why both roots are load-bearing (byte-level)

- Root A: asserted that its delivery cred bytes CHANGE across the pass (the manager wrote its own store).
- Root B: asserted that its delivery AND membership-rw cred bytes DO NOT change (the daemon's store was never written), while the refusal names exactly that stale generation.
- The membership component is a real adopter (the daemon's membership feed is up and its refusal is asserted separately), not just delivery.

## Control

A second phase runs the IDENTICAL shipped path over a unified root (manager and daemon share one root) on its own fresh broker: adoption succeeds, broker-accepted and pinned to each component's own nkey, proving the phase-1 refusal is caused by the root divergence and not by the rig. Each phase gets its own broker because the per-space artifact store reserves 4 GiB of JetStream capacity, which two concurrent spaces overrun on a small CI disk; a virgin broker per phase also rules out cross-phase carryover.

A cell-count assertion (`21` enumerated cells) guards against silently skipped cells reading as green.

Supersedes #1177.

### DIFF
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -740,3 +740,9 @@ smoke:jcode-mid-turn-delivery
 # the real Claude plugin-list JSON seam. Mutation-proofed - bin/smoke/mutations/status-provenance.json.
 # Appended so every existing shard assignment remains unchanged.
 smoke:status-provenance
+
+# The #773 two-root renewal refusal, driven by the shipped owner: a real Manager on its own
+# workspace root and a real delivery daemon on a divergent root, with the defective behavior
+# encoded as explicitly-labeled "#773 unfixed" expectations plus a unified-root positive control.
+# Appended so every existing shard assignment remains unchanged.
+smoke:manager-two-root-renewal
--- a/bin/smoke/manager-two-root-renewal.smoke.ts
+++ b/bin/smoke/manager-two-root-renewal.smoke.ts
@@ -1,0 +1,298 @@
+/**
+ * The #773 two-root renewal refusal, reproduced through the SHIPPED renewal owner - a REAL
+ * `Manager` (its initial class-2 renewal pass in `start()`), a REAL delivery daemon
+ * (`tsx bin/cotal.ts deliver`), a REAL authed broker. No live stack, no shared state.
+ *
+ * THE DEFECT (#773, UNFIXED): the manager re-signs the daemon creds through ITS OWN store (the FS
+ * store over its `workspaceRoot`), then hands the daemon per-component expected fingerprints over
+ * the delivery-admin rail - but the daemon re-reads its OWN store (resolved from its own root). In
+ * a composition where the two roots differ (cross-host stock deployment), the manager writes
+ * filesystem A, the daemon re-reads filesystem B, and EVERY adoption is refused with the exact
+ * generation-mismatch error. Nothing refuses the divergent composition at startup.
+ *
+ * The "#773 unfixed:" cells below ENCODE THAT DEFECTIVE BEHAVIOR AS THE EXPECTATION, explicitly
+ * labeled: they are green on today's tree and go red when #773 is fixed (either startup refuses
+ * the divergence loudly, or adoption succeeds through one shared authority) - the fix PR must
+ * flip them to the fixed expectations.
+ *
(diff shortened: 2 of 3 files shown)


davidfarah2003:
APPROVE

Graded SHA: `50a6356a6fbbc79d1bd0e1bddee0ecb7761a985b`
COTAL_MODEL: (unset)
Token: `1185r-50a6356a6`

This suite honestly reproduces #773 through the shipped renewal owner. The three #1177 blockers are closed.

**#773, live.** `pnpm smoke:manager-two-root-renewal` at this SHA: **21 passed, 0 failed**. Manager logs the exact dual refusal (`delivery: reloadCreds: ... did not match the expected re-signed generation` and `membership: reloadRwCreds: ... did not match the expected re-signed generation`). `Manager.start()` still returns. The unified-root control adopts and pins both components to their own nkeys.

**#1177 blockers.**
- Daemon store is load-bearing at the byte level: root A's `delivery.creds` changes, root B's `delivery.creds` and `membership-rw.creds` do not.
- The pass is `new Manager({ workspaceRoot: rootA }).start()`. No hand-sent fingerprint, no direct rail drive.
- Membership is its own cell against a live feed (`membership feed up` is a rig `must`, then the generation-mismatch `ok` on `detail.membership`).

**#773 unfixed contract.** Nine cells are prefixed `#773 unfixed:` and the header says they are green today and must go red when #773 is fixed. The refusal cells (`adoption.ok:false`, delivery mismatch, membership mismatch) cannot stay green under either accepted fix shape (startup refusal leaves `rec` missing, successful adoption flips `ok`). Two of the `#773 unfixed` cells are composition evidence (B's bytes stay old). They would remain true under a wire-delivered-cred fix that never writes B. That is a label leftover for the fix PR, not a fail-green of the defect itself.

**Fail-green / count / hermetic / ci list.**
- Throws hit `catch` and set `process.exitCode = 1`. `finally` only tears down. No `process.exit` in `finally`.
- I counted 6 `must` + 14 `ok` + 1 count cell = 21. `EXPECTED_CELLS = 21` and the last cell asserts `pass + fail === 20` before it increments. Live banner: 21 passed.
- Sandboxed `COTAL_HOME`, ambient `COTAL_*` scrubbed, kills only spawned broker/daemon PIDs, throwaway roots, no live stack.
- `ci-suites.txt` is a true append at EOF (`smoke:manager-two-root-renewal` is lines 744-748 of 748). Existing shard lines are untouched.

**Typecheck.** `pnpm typecheck` passed (includes `bin/tsconfig.smoke.json`, which includes this file). The requested isolated `tsc` of the suite file fails to resolve workspace packages (`TS2307`) because it has no project paths. That is the missing `bin/tsconfig.json` include, not a type hole in the suite. With modules unresolved it also reports implicit `any` on `r`. That disappears under `tsconfig.smoke.json`.

Issue #964, opened 2026-08-28
cotal down reaps every managed agent by default, and there is no flag that stops the stack without them


davidfarah2003:
`cotal down` reaps every managed agent along with the stack. The sweeping, unrecoverable
operation is the default, and the scoped one does not exist.

## What happened

On 2026-08-26 a single terminated signal to the stack took six live seats with it. The manager
reaped all of them with "this manager stopped it (despawn or shutdown)" and deregistered cleanly,
so the logs are indistinguishable from a deliberate teardown. Several of those seats held
uncommitted work. There was no flag that would have stopped the stack and left them running,
because there is no such flag.

## Why this is the wrong default

A stack restart is a routine operation. Ending every agent is not. Pairing them means the routine
operation carries the destructive one, and the operator has no way to opt out at the moment they
need to.

`down -f` already proves scoped teardown is a concept the CLI has.

## Proposed change

- Bare `cotal down` stops the stack, leaves managed agents running, and prints what it would have
  reaped.
- `cotal down --with-agents` performs today's behavior.

## Done when

Running bare `down` on a mesh with live seats leaves the seats running, and the command says which
agents it left alone and how to reap them.

davidfarah2003:
Reproduced at exact current-main SHA `204a762660507396a6e93fa236f9ff2b8dff70b0`. Bare `cotal down` selects the full local process surface. Stopping the manager reaches `Manager.stop()`, whose normal active path explicitly calls `teardownManagedAgents`: it snapshots every managed seat, hard-stops each handle, waits, and deprovisions every non-suppressed entry. Even the partial-preservation exit path hard-stops all retained handles; it preserves authority/state, not live processes. There is no `--with-agents`, `--leave-agents`, or inverse scoped flag. The reported six-seat loss therefore matches intended current code, including clean logs indistinguishable from deliberate teardown. `down -f` is a positive control for ownership-scoped agent teardown, and component names prove infrastructure selection exists, but neither stops the stack while keeping all seats. No duplicate or active fix exists. Recommend `bug`, `triage:confirmed`, `area:manager`, `severity:critical`. Needs David's runtime-lifecycle design: built-in pty seats are manager-owned and cannot be made safely persistent by merely skipping stop, while tmux/cmux/herdr have different lifetime semantics. Define survivability, successor adoption, terminal state, credentials, failure reporting and whether destruction moves behind `--with-agents` before implementation.

davidfarah2003:
[...]

PR #1186, opened 2026-08-31
test(manager): reproduce the stack-stop agent reap (#964)

davidfarah2003:
## What this is

An honest, hermetic reproduction of #964: a stack stop reaps every managed agent by default, and there is no flag that stops the stack without them. The defective behavior is encoded as explicitly-labeled "#964 unfixed:" expectations that are GREEN today and must go red (or be explicitly relabeled) when #964 is fixed. No fix is included; the fix design remains owned by the design lane per the issue thread.

## Full scope

Three files:

- `bin/smoke/manager-stop-reaps-agents.smoke.ts` — the suite.
- `package.json` — the `smoke:manager-stop-reap` script (builds first; the suite drives dist and the child imports core's dist).
- `bin/smoke/ci-suites.txt` — appended at end of file, so every existing shard assignment is unchanged.

## What the suite drives

The shipped owner, not a restatement: a real `Manager` (pty runtime) over a real authed throwaway nats-server, with a real co-located delivery daemon (a direct `deliver` run — on an auth mesh the deprovision path verify-evicts through its ctl.delivery-admin rail; without it the reap's footprint half never completes, which run 1 of this suite measured). One space only (each space reserves a 4 GiB artifact store on the broker's tmpfs store dir). The managed seat is spawned through the real CLI `spawn` command in-process; its child process is a real mesh endpoint that authenticates with the manager-minted creds, joins presence, and writes its PID to a pidfile so liveness is measured rather than inferred.

**Defect phase** — seat live (PID answering, creds file on disk, still alive after a settle window so its death is stop-caused), then a plain `mgr.stop()`, the path bare `cotal down` drives:

- `#964 unfixed:` the stop proceeds against a live managed seat with no refusal and no sparing mode;
- `#964 unfixed:` the seat's process is hard-stopped;
- `#964 unfixed:` the seat is deprovisioned (its minted creds file is gone).

**Control phase** — a fresh manager over the same root, a second seat, and a DELIBERATE per-seat despawn through the real CLI (`cotal stop --name`, run as a child process because the in-process command exits on failure). The terminal observables are the SAME (process dead, creds gone) — the issue's "indistinguishable from deliberate teardown" claim made concrete.

A count cell asserts every cell ran (6 must + 8 ok before it, 15 total; silently skipped cells cannot read as green), a throw lands as a counted failure, and the banner always prints.

## The contract with the future fix

Written into the suite header: the accepted direction is a three-mode `down` (bare = spare, `--with-agents` = today, `--preserve-state` = capture-and-restore). If `Manager.stop()` learns a sparing default, the labeled cells go red and the fix PR flips them into survival assertions. If `stop()` instead grows an explicit mode parameter with bare `down` passing the sparing choice, the fix PR relabels these cells as the destructive mode's explicit spelling and adds the sparing path as new green cells. Either way this file changes in the fix PR; a fix that leaves it untouched is incomplete.

## Named gaps (deliberate)

- The CLI `down` surface itself is not driven: per the #964 architecture review, validation on this class of host is limited to shipped handlers and in-process manager seams — which is what this suite is.
- The preservation arm (`stopRetainedAgentsOnExit`) is not driven: it is only reachable through preservation state no shipped public path sets in this rig.
- The broker-side footprint (durables, ACL row) is not asserted; the on-disk creds file is the asserted deprovision observable.

## Hermeticity

Sandboxed `COTAL_HOME`, all `COTAL_*` scrubbed before any cotal import, throwaway broker on an OS-assigned free port with a token-prefixed store dir, kills only PIDs it spawned or that its own children wrote to its own pidfiles, removes only its own temp dirs. It never runs `cotal up`, `cotal down`, or any stack lifecycle command.

Closes nothing — #964 stays open for the fix; this lands the reproduction the issue thread named as the missing piece.

### DIFF
--- a/bin/smoke/ci-suites.txt
+++ b/bin/smoke/ci-suites.txt
@@ -746,3 +746,10 @@ smoke:status-provenance
 # encoded as explicitly-labeled "#773 unfixed" expectations plus a unified-root positive control.
 # Appended so every existing shard assignment remains unchanged.
 smoke:manager-two-root-renewal
+
+# The #964 stack-stop reap, driven by the shipped owner: a real Manager over an authed broker
+# hard-stops and deprovisions a live managed seat on a plain stop() with no sparing mode, with
+# the defective behavior encoded as explicitly-labeled "#964 unfixed" expectations plus a
+# deliberate-despawn control showing identical terminal observables.
+# Appended so every existing shard assignment remains unchanged.
+smoke:manager-stop-reap
--- a/bin/smoke/manager-stop-reaps-agents.smoke.ts
+++ b/bin/smoke/manager-stop-reaps-agents.smoke.ts
@@ -1,0 +1,342 @@
+/**
+ * HONEST REPRODUCTION of #964: a stack stop reaps every managed agent, and there is no way to
+ * opt out - pnpm smoke:manager-stop-reap
+ *
+ * The incident: one stop signal to the stack took six live seats with it, several holding
+ * uncommitted work, and the logs read as a deliberate teardown. The mechanism is
+ * `Manager.stop()`: its normal active path unconditionally calls `teardownManagedAgents()`
+ * (implementations/manager/src/manager.ts, the `maintenanceState === "active"` arm), which
+ * hard-stops every managed seat and deprovisions its footprint. Bare `cotal down` stops the
+ * manager and therefore drives exactly this path. There is no flag, mode, or argument that
+ * stops the stack and leaves the seats running.
+ *
+ * This suite drives the SHIPPED owner of that behavior - a real `Manager` over a real authed
+ * broker with a real co-located delivery daemon (a direct `deliver` run, never `up`; on an auth
+ * mesh the deprovision path verify-evicts through it), a real managed seat spawned through the
+ * real CLI spawn command - and pins today's defective semantics as explicitly-labeled
+ * "#964 unfixed:" expectations that are GREEN today.
+ *
+ * THE CONTRACT WITH THE FUTURE FIX (read this before touching the labeled cells): the accepted
+ * direction for #964 is a three-mode `down` (bare = spare agents, `--with-agents` = today,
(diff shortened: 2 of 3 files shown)


davidfarah2003:
APPROVE

Graded SHA: `398c8582062cfd4a07053bb256204cff57a3a381`
COTAL_MODEL: (unset)
`1186r-398c85820`

Honest reproduction of #964 at the shipped seam. A real in-process `Manager` (pty) over a throwaway authed nats-server plus a direct `deliver` child (never `cotal up`) spawns a live seat through the registered CLI `spawn` command. Liveness is a pidfile PID answering `kill(0)`, not an inference. Plain `mgr.stop()` is the path bare `cotal down` reaches via SIGTERM on the manager command (`commands.ts` shutdown). Three `#964 unfixed:` cells are green today (no refusal, process dead, creds file gone). The control phase drives `cotal stop --name` as a child and asserts the same disk observables. Header contract matches the three-mode fix: flip to survival if `stop()` learns a sparing default, or relabel as the destructive mode if a mode parameter is added.

Cell arithmetic holds: 6 `must` + 8 `ok` + 1 count = 15. The count cell asserts `pass + fail === 14` before it increments. A throw is caught, increments `fail`, then the banner always prints after `finally` (no `process.exit` in `finally`); `process.exitCode` carries the result. Daemon readiness waits for both `delivery daemon up` and `membership feed up`. Hermetic: `COTAL_*` scrubbed before any cotal import, sandboxed `COTAL_HOME`, kills only spawned PIDs or this run's pidfiles, removes only its temp dirs. `ci-suites.txt` is appended after `smoke:manager-two-root-renewal`.

Named gaps (CLI `down` not driven, `stopRetainedAgentsOnExit` unreachable via public paths here, broker-side durables/ACL not asserted) are written in the suite header and PR body. No unnamed gap found that would leave the defect unpinned.

Ran `pnpm typecheck` (pass) and `pnpm smoke:manager-stop-reap` alone (15 passed, 0 failed).

Issue #1176, opened 2026-08-30
lang: a RunDivergence inside a concurrency scope is recorded as the scope's own failure


davidfarah2003:
## Summary

`performScope` keeps a ladder of classes it refuses to record as a scope's outcome. After #1174 that ladder holds `JournalAppendRejected`, `CloseOwed`, `Cancelled` and `RunReleased`. A `RunDivergence` raised inside a live or resumed concurrency scope is not on it, so it falls through to the generic settle and is written as the scope entry's own failure:

```ts
const err: EntryError =
  reason instanceof EffectError
    ? recordableError(reason, "scope-fault").error
    : { code: "L4000", kind: "scope-fault", message: messageOf(reason) };
await host.journal.settle(scopeKey, { status: "failed", error: err }, endedAt, { ... });
```

This is the same shape as #862 with a different class. It predates that fix and is untested at a scope boundary.

## Why it is worth its own issue

A divergence means the journal is saying this program is not the one that wrote it. The interpreter treats that as uncatchable elsewhere: it is in the set that unwinds past `finally`, alongside `Cancelled`, `JournalAppendRejected`, `ScopeBranchMissing` and `UnwalkableScope`, and `interpret.ts` records why. A resume whose edited `sleep` diverged inside a `try` was measured catching `{ code: "L4000", kind: "host" }`, logging past it, and performing a new effect against the journal it had just diverged from.

Recording a divergence as a scope's outcome converts a signal that the program and the journal disagree into a durable statement that the scope failed for a reason of its own. What a resume then replays is a scope-fault, not a divergence, so the loudness the class exists for is spent before anyone sees it.

## Measured

Reported by the correctness reviewer on #1174 at `0098000e7`: a divergence raised inside a live or resumed scope still falls through to the generic `L4000` settle. Adding `RunReleased` to the ladder does not make this worse, and the two are independent.

Not measured, and worth establishing first: whether a resume of such a journal replays a `RunDivergence` or an `EffectError`, and whether a migration walk reaches the same path. #862's own reproduction is a working starting point, since the only change needed is which class the body raises.

## Scope note

`ScopeBranchMissing` and `UnwalkableScope` are the remaining members of that set and were also never tested at this boundary. One reviewer's reading is that both are raised on migration walks of settled scopes, which do not take this door, so they are likely not affected. That is a reading rather than a measurement and should be checked rather than assumed.

## Done when

[...]

PR #1188, opened 2026-08-31
feat(lang)!: event-queue simulator, ask schema contract, journal guards, scope release law

davidfarah2003:
Four changes to `@cotal-ai/lang` (plus one wiring point in `@cotal-ai/runtime`), each with spec text, docs, and mutation-proofed coverage.

## Discrete-event simulator

`Simulator` now runs a real event queue on one virtual clock. Timed effects (sleep, turn, wait, checkpoint, ask) park until their wake time; a pump delivers exactly one event per macrotask in wake order (ties by ask order) and advances the clock to the delivered wake. Concurrent branches therefore accumulate their own durations, and a race is decided by which arm actually finishes first on the clock, matching the recorded-clock rule replay already used. Before this, the simulator resolved every effect instantly on a shared serial clock, so race outcomes in simulation depended on the order effects were asked, and dry runs could disagree with replay.

## Ask schema contract

The `ask` effect's `schema` is now interpreted, per the spec's shorthand: a record of field name to kind (`string` / `number` / `boolean` / `array` / `record` / `null`), reply must be a record, extra fields allowed, `{}` means any record. An unreadable schema is refused with L4022 before anything is asked; a non-conforming reply consumes an attempt, and exhausting `attempts` (default 1) throws L4006 naming the first non-conforming field. `askSchemaShape` and `conformsToAskSchema` are exported so hosts enforce the same contract. Checkpoint `schema` stays uninterpreted, as specified.

## Journal guards

- **L5006 result bound.** A `Journal` can be given `resultBytes`; a settling `ok` result whose canonical form exceeds it is refused with `EffectResultTooLarge` (a `JournalAppendRejected`, so it travels the uncatchable L5010 path) before anything is persisted. `DriveRequest.resultBytes` wires it through the runtime, since the host is what knows the broker's max payload.
- **Seed vocabulary guard.** Constructing a `Journal` over entries with a state outside `pending`/`settled` or a settled status outside `ok`/`failed`/`cancelled` is refused, naming the entry. Previously a corrupted status replayed silently as a wrong answer.
- **Journal fuzz smoke.** A new suite truncates, reorders, and corrupts a recorded journal and asserts the closed-outcome property: every degraded journal yields either a named refusal or the correct continuation, never a silent wrong answer.

## Scope release law

A host-side unwind — `RunReleased` or a refused append — inside `parallel`, `race`, `fanOut`, or `conclave` is not a branch failure: it no longer cancels siblings, and a released race arm is not a candidate. This fixes a real defect: a driver stopped inside a scope would durably settle in-flight sibling entries `cancelled`, and a later resume replayed those cancellations into a healthy run, permanently. Now the scope lets in-flight work settle, rethrows the unwind, and a resume replays the run to the same result. New spec text in §7.6 states the law.

## Spec and docs

- `spec/cotal-lang.md`: §6.5 (ask schema contract), §7.6 (release law), §9.2 (L5006 rides the L5010 clause), §12 (result bound), Appendix A (+L4022, L5006 no longer reserved), Appendix B changelog row.
- `docs/workflows.md` and the lang README describe the discrete-event simulator.
- Changeset: minor for `@cotal-ai/lang` and `@cotal-ai/runtime` (lockstep group versions all packages).

## Verification

- Full lang suite green (19 smokes; `engine.smoke` 411 checks including the mutation-config audits), runtime smokes green, `pnpm typecheck` and `pnpm changeset status` clean.
- Six mutation-proof configs, 12 mutants, all killed on their named assertions: ask-schema, sim-wake-order, journal-guards (bound + both vocabulary guards), release-no-cancel (parallel, fanOut, race candidacy), race-live-pick, freeze-boundary.
- `pnpm check:docsbundle` green.


davidfarah2003:
APPROVE

Reviewed PR #1188 at `3aabb2583ff7e1126a58d24034d28c0ede6b954a`.

Model: `gpt-5.6-terra` (`COTAL_MODEL`).

No spec/docs coherence blockers found. The changed `spec/cotal-lang.md` rules for the `ask` schema shorthand, result-bound durability refusal, scope host-unwind behavior, simulator ordering, and error catalog match the language implementation. SPEC §14 continues to incorporate the language reference consistently. Appendix A already maps the touched language and runtime surfaces, and Appendix B has no impact because this increment changes no wire ACL or profile surface.

RAN
- `git rev-parse HEAD` -> `3aabb2583ff7e1126a58d24034d28c0ede6b954a` (exit 0)
- `pnpm build` (required to materialize missing workspace `dist` artifacts after the initial generator precondition failure) (exit 0)
- `pnpm gen:tooldocs && pnpm gen:docsbundle` (exit 0)
- `git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` (exit 0)
- `pnpm --filter @cotal-ai/lang test` (exit 0)
- `git diff --check 43e1f7d0d^..HEAD` (exit 0)

davidfarah2003:
## Engineering review of #1188 at 3aabb2583ff7e1126a58d24034d28c0ede6b954a

**Verdict: APPROVE**

Seat: `rev1188a-eng-aa63` (reviewer). Route: jcode. `COTAL_MODEL` is unset in this process. Mesh orientation recorded the pin as **grok-4.6**.

Scope is the lang-core increment on this SHA only.

### Simulator discrete-event vs spec §8.1

`SimHandler` now parks timed effects (`sleep`, `turn`, `ask`, `checkpoint`, `wait`) on one virtual clock and delivers them in wake order, ties by park sequence. Delivery is one event per macrotask so a parking branch reads its own time. That is the handler-side clock the interpreter stamps as `endedAt` and later joins as `now()` (§8.1). Spec §8 itself does not define the simulator (Appendix A still says simulation is a tool, not the language). The live race cell in `scopes.smoke` now depends on wake-order delivery (`fast` at 1m vs `slow` at 5m) instead of the old serial-clock tie, which matches the comment in `sim.ts` and the `sim-wake-order` mutation.

### Journal L5006 and the fuzz vocabulary walk

`Journal.settle` measures a canonical `ok` result against `resultBytes` **before** `persist` and throws `EffectResultTooLarge` (`code` L5006, subclass of `JournalAppendRejected`). No bound means no check, which matches §12. The seed constructor refuses `state` outside `{pending,settled}` and settled `status` outside `{ok,failed,cancelled}` (including missing status). `fuzz-journal.smoke` walks those classes: truncation and reorder resume to the same answer, hash flip is named divergence, foreign run is L5011, missing pins is L5021, junk status/state/no-status refuse at the seed, oversized result is L5006 with the store seeing nothing and the entry still pending. Seeded fixtures in `journal.smoke` / `engine.smoke` that used `status: "done"` were corrected to `"ok"` so they still exercise the intended bind/value-cross cells rather than dying on the new vocab guard.

### Scope release law

`parallel` and `fanOut` rethrow `RunReleased` / `JournalAppendRejected` after `allSettled` **without** cancelling siblings. Race `onSettle` treats those as non-candidates, same as `Cancelled`, so a released arm neither wins nor cuts in-flight work. That matches the §7.6 / §9.2 wording added in this PR. `interpret.smoke` pins resume-past-pending-scope for both `parallel` and `fanOut`, and a race where a sibling is released still lets in-flight work win with no recorded cancellation.

### Ask schema hashing vs handler shorthand

Language-side hashing is unchanged: `ask` still hashes `{ agent, schema, deadline, attempts }` via `digest` (§6.4). The new shorthand (`askSchemaShape` / `conformsToAskSchema`) is handler-side, as §6.5 now states. The reference simulator refuses an unreadable schema (L4022), counts non-conforming replies against `attempts`, and reports L4006 on exhaustion. Extra fields allowed, `{}` accepts any record, a non-record still fails. Checkpoint schema stays uninterpreted. Differential/engine freeze cells that used `ask` replies were moved onto `checkpoint` so they still prove L2031 rather than L4022.

### Mutation finds at this head

Each `find` in the configs this increment added or re-anchored matches its named source **exactly once**:

- `fanout-failure.json` (both finds in `perform.ts`)
- `freeze-boundary.json` (`perform.ts` share freeze, `journal.ts` seed freeze)
- `race-live-pick.json` (`at < winnerAt` in `perform.ts`)
- `journal-guards.json` (L5006 throw, status vocab, state vocab)
- `ask-schema.json` (conformance return, `askSchemaShape` without `?? {}`)
- `release-no-cancel.json` (parallel guard, fanOut guard, race non-candidate)
- `sim-wake-order.json` (wake-then-seq pop)

No find in that set points at a dead or moved line. Re-anchoring of the two fanOut finds still names the sibling-cancel loop and the `losers` assignment that the cells assert.

### Suite claims vs cells

Touched suites' named cells match what they assert: `sim.smoke` 10/11 is wake order and the ask contract; `fuzz-journal.smoke` 7/8 is vocab + L5006; `interpret.smoke` host-stop / freeze cells exist as written; `scopes.smoke` live least-clock and fanOut failure cells exist as written. Fail-fast mutation `expectRed` / `completionMarker` strings are present in those suites (template-literal cells in interpret expand at runtime to the parallel/fanOut names the config quotes). I did not re-run `pnpm mutation-proof` (not requested, and not a substitute for the named-find check).

### Gaps (named, not blockers)

- `pnpm smoke:ci` is forbidden from this seat. CI at pin is reported green as run 33472961774 by the orchestrator; I did not execute that workflow.
- `COTAL_MODEL` is empty here. Routing is grok-4.6 from mesh orientation, not from the environment.

### RAN

```
git rev-parse HEAD
# 3aabb2583ff7e1126a58d24034d28c0ede6b954a  (exit 0)

pnpm typecheck
# exit 0  (workspace build + typecheck, 27 of 28 projects)

pnpm smoke:lang-sim            # exit 0, sim.smoke: 58 checks passed
pnpm smoke:lang-fuzz-journal   # exit 0, fuzz-journal.smoke: 17 checks passed
pnpm smoke:lang-interpret      # exit 0, interpret.smoke: 101 checks passed
pnpm smoke:lang-scopes         # exit 0, scopes.smoke: 96 checks passed
```

Mutation-find uniqueness was checked by counting each config `find` in the named file (all 1x). Forbidden commands were not run (`cotal up/down/setup/ext seed/history`, `pnpm smoke:ci` / `pnpm check`, `*-live`, `git worktree prune` / `git gc` / `git prune`).

davidfarah2003:
**BLOCK** at `3aabb2583ff7e1126a58d24034d28c0ede6b954a`.

A third review lens on this head reproduced a defect the two posted APPROVEs did not reach. That seat was lost before it could post; its notes are at `/home/cotal/rev1188a-cold-notes.md`. I reproduced the finding independently in a separate clean worktree at the same pin before recording this, and ran the attribution control myself. Both results are below.

### A refused journal append inside `race` is swallowed

`packages/lang/src/perform.ts:1146-1152` adds `RunReleased` and `JournalAppendRejected` to the winner-scan skip list. The host-side unwind at `:1159` only runs in the `winnerIndex === -1` branch, so when any other arm settles as a candidate the refused-append arm is skipped, the race settles `ok`, and the fault is discarded. The run completes.

Reproduced at the pin with a store that throws on the ask's settling append (the production-reachable shape: a broker refusing an oversized payload, L5010):

```
outcome: COMPLETED
logs: [["winner","other","other"]]
   race:either  state=settled status=ok
   ask:q        state=pending status=-     <- the ask ran; nothing recorded it, ever
```

**Attribution control.** Reverting only the winner-scan condition to `(r.reason instanceof Cancelled)`, changing nothing else, then restoring with `git checkout --` (tree back to clean, HEAD unchanged):

```
outcome: JournalAppendRejected L5010
   race:either  state=pending status=-
```

So this change converts a durability failure that used to unwind loudly into a silent success.

### It contradicts this PR's own text

- SPEC §9.2 as amended here: "an implementation MUST unwind the run through ... a journal append the store refused (L5010 ...; L5006 ... travels the same way)".
- The changeset: "the unwind propagates bare, the scope settles nothing." In `race` it does not propagate, and the scope does settle.
- The L5006 message says "the run can no longer record it" — the run completed.

### Coverage

No cell covers it. The one new race cell in `interpret.smoke` ("a race whose sibling was released still lets in-flight work win") is `RunReleased` only, and `release-no-cancel.json`'s third mutant is likewise `RunReleased`.

### The no-cancel law does not require this

Cancellation is issued from `onSettle` at `:1114`; that filter is correct and should stay. Candidacy at `:1146` is a separate decision. Hoisting the `hostSide` find that already exists at `:1161-1166` above the winner return would keep the law and restore the unwind. Whether `RunReleased` should also be a non-candidate is a design call this PR argues and tests; `JournalAppendRejected` is not.

### Non-blocking, recorded so they are not lost

1. A schema-less `ask` validates (`ask` bags are closed but no option is required), and `askSchemaShape(undefined)` returns null, so the simulator raises "L4022 Unreadable ask schema" when there is no schema to be unreadable. It fails loudly and the Fix line points at `{}`, but the requirement belongs in the validator, statically, and the message names the wrong defect.
2. `run-driver.ts:245-251` tells the host it "knows its `max_payload` and passes the bound here", while `journal.ts:585-588` measures only `canonicalize(outcome.result)`. The persisted entry also carries run/scope/keys/inputHash/facts, so an entry can pass the L5006 check and still be refused L5010 at the store; a host passing `max_payload` verbatim gets the wrong bound. Nothing in the shipped product sets `resultBytes` today, so L5006 is opt-in and currently unwired.
3. The 2026-09-01 changelog row lands at `spec/cotal-lang.md:1101`, ahead of nine 2026-08-19 rows, in a table that otherwise runs ascending.

The two APPROVEs on this head stand as what they cover. A head is clear only when every reviewer has cleared it, and this one is not.

davidfarah2003:
APPROVE

Reviewed the bounded spec/docs delta from `3aabb2583` to `70df408db6b15d834d64f9cdec9fa001285b5868`.

Model: gpt-5.6-terra, xhigh

The new Appendix B sentence matches the repaired precedence: after all race arms settle, a refused append is raised before winner selection, so the race cannot settle over an outcome the journal did not record. Its normative basis remains coherent with §7.6, §9.2, and §10.5. SPEC §14 incorporates `spec/cotal-lang.md` by reference, so no separate SPEC edit is required. Appendix A needs no update because this repair introduces no error code; Appendix B records the behavior change. The change-log dates now ascend through the final 2026-09-01 row.

## RAN

- `git rev-parse HEAD` → `70df408db6b15d834d64f9cdec9fa001285b5868` (exit 0), before and after review.
- `git diff 3aabb2583 70df408db` and affected normative context inspection (exit 0).
- `pnpm gen:tooldocs && pnpm gen:docsbundle` (exit 0).
- `git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` (exit 0).
- `pnpm check:docs-voice` (exit 0, 39 pages passed).
- `pnpm smoke:lang-interpret` (exit 0, 104 checks passed), including the refused-append race case.
- `git diff --check 3aabb2583 70df408db` (exit 0).
- Final porcelain status was clean.

davidfarah2003:
## Engineering delta re-grade of #1188 at 70df408db6b15d834d64f9cdec9fa001285b5868

**Verdict: APPROVE**

Seat: rev1188b-eng-1b68 (reviewer). Route: jcode. COTAL_MODEL: grok-4.6.

Bounded to `git diff 3aabb2583 70df408db` (5 files, +70/-5) plus what that repair puts at risk. Not a re-grade of the round-1 surface.

The hoist of `refusedAppend` in `packages/lang/src/perform.ts` is correct and complete for this defect. It runs after `allSettled`, the post-settle blanket cancel, and the clock join, and still ahead of the winner scan. `performScope` rethrows `JournalAppendRejected` before any scope settle, so a refused append cannot become a completed race. At `:1114` a refused append is still a non-candidate and does not cancel siblings, which is the same law as `RunReleased`. The join does not write the journal.

`parallel` and `fanOut` have no analogous hole. They rethrow `JournalAppendRejected` after `allSettled` on the first rejection. There is no winner scan that can prefer a sibling.

`EffectResultTooLarge extends JournalAppendRejected`, so the L5006 cell rides the same check. Confirmed by the new interpret cells and by the mutation kill.

`JournalAppendRejected` is gone from the `winnerIndex === -1` host-side find and is unreachable there now. `RunReleased` on that path is unchanged.

The leftover skip of `JournalAppendRejected` in the winner-scan continue list is redundant and harmless. It is not a blocker. An editor who deleted the hoist and kept only the skip would restore the defect. The comments already say the hoist is the unwind.

Permanent cell: `pnpm mutation-proof --config packages/lang/smoke/mutations/race-append-unwind.json` was KILLED, red-and-named on `a refused append inside a race unwinds the run even though a sibling won`.

### Probe (not a blocker, not #1176)

The new decision point's class set is right for this repair.

- `JournalAppendRejected` / `EffectResultTooLarge`: run-level durability. The journal refused to record. Burying that is a silent success over a pending step. That is this defect, and the hoist covers it.
- `RunReleased`: host stop, not a recorded outcome. It stays a non-candidate. An in-flight winner may stand because a release happens before begin and does not leave a pending-forever step. That is a different law, and it is tested.
- `RuntimeFault`: program-visible branch or runtime error. A loser's `RuntimeFault` is a branch outcome. Discarding it is what a race is for.
- `JournalReadOnlyError`: a write on a read-only journal. It is not a subclass of `JournalAppendRejected` and is not on the walker's uncatchable list. A live writable race is not this path. Same family as cannot-record, different recovery shape, not this hole.
- `RunDivergence`: run-level uncatchable (journal vs program), thrown on lookup before begin, so no new pending entry. A winning sibling can still bury it. That is the incomplete-ladder shape already filed as #1176, at a second site, not the pending-forever durability lie this delta closes. Not a #1188 blocker.

### RAN

- `git rev-parse HEAD` -> `70df408db6b15d834d64f9cdec9fa001285b5868` (exit 0), matches the bound pin. Rechecked after mutation-proof. Same SHA.
- `git status --porcelain` empty before checks and after mutation-proof (exit 0).
- PR `headRefOid` still `70df408db6b15d834d64f9cdec9fa001285b5868`.
- `pnpm typecheck` -> exit 0.
- `pnpm smoke:lang-interpret` -> exit 0, 104 checks passed (round 1 reported 101, this delta adds 3).
- `pnpm mutation-proof --config packages/lang/smoke/mutations/race-append-unwind.json` -> exit 0. KILLED red-and-named on the named cell.

Forbidden suites (`pnpm smoke:ci`, `*:live`) were not run.

davidfarah2003:
**APPROVE** at `70df408db6b15d834d64f9cdec9fa001285b5868` — cold read, no prior context on this campaign.

Read the whole diff against the merge base, ran `pnpm typecheck` (exit 0) and the named lang
suites in the worktree: sim 58, interpret 104, scopes 96, journal 103, fuzz-journal 17, dryrun 23,
examples 6, migrate 59, differential 206/206, engine 411 — all green. I found no incorrect
behavior. Five things I'd want on the record, none of which I'd hold the merge for.

**What I verified rather than took on trust**

- L5006 is genuinely uncatchable: a program wrapping an oversized `ask` in `try/catch` unwinds
  with `EffectResultTooLarge`, and the `catch` never runs (`log` output empty). The name is
  registered in `UNCATCHABLE_NAMES` for the engine's name-based path, and the walker's
  `instanceof` reaches it through the base class.
- The "refused ahead of the append, nothing moved" claim holds: the check in `journal.ts settle()`
  sits above both `persist` and `byKey.set`, so the pending entry is intact.
- The seed vocabulary guard is load-bearing, not defensive. `lookup()` checks pending, then
  `failed`, then `cancelled`, then falls through to `replay` — so a status outside the vocabulary
  really would have been handed to the program as recorded truth.
- `runScope` is shared by `interpret.ts` and `engine/ctx.ts`, so the scope law lands on both
  language versions rather than only the walker.
- The ci-suites fragment filename is sha256("smoke:lang-fuzz-journal"). Matches.

**Findings**

1. *The conclave arm of the release law is ungraded.* `perform.ts:1319` changes what a release
   inside a `conclave` body does. Measured at this head with a release raised at the first dispatch
   inside the body: `closeConclave` is called 0 times and the entry stays `pending`; on main the
   same body error reached the close. The changeset and the §7.6 edit both name `conclave`, and
   `parallel`, `fanOut` and `race` each got a cell and a mutation config — this arm got neither.
   The behavior itself reads correct to me (calling a handler after a release was itself a new
   effect past the stop, and it could mask the append rejection behind `CloseOwed`), which is why
   this is a finding and not a block.

2. *That line's comment explains half its condition.* "closing would be a new action after the
   driver said stop" describes `RunReleased`. A `JournalAppendRejected` is not a stop — the driver
   is live and the world reachable, which is the exact case the paragraph three lines above says
   must not walk away from live membership. The outcome is still defensible (pending is the
   durable "a close is owed"), but the next contributor is given a reason that does not cover the
   case they are most likely to be debugging.

3. *`DriveRequest.resultBytes` has no production caller.* It is set only in two smoke files.
   `run-driver.ts:251` says "The host that opens the broker connection knows its `max_payload` and
   passes the bound here" — no host does, so L5006 cannot fire in a shipped run. The spec says
   MAY, so opt-in is the design; the doc sentence just describes wiring that is not there yet.

4. *A schema-less `ask` now dies as "Unreadable ask schema".* `ask(a, { name: "q" })` is accepted
   by the validator and reaches the handler with `schema: undefined`, so it now fails L4022 under
   `SimHandler` with a message that says "The schema is not the shorthand" when the program wrote
   none. No in-repo caller does this and the spec's own signature lists `schema` as non-optional,
   so refusing is right — one clause naming the missing-schema case would make the diagnostic
   match the mistake.

5. *`sim.ts timed()` adds an `onCancel` listener per park and never removes it*, so listeners
   accumulate on a branch's signal for the life of the branch. Bounded by the step budget, so a
   note rather than a leak that matters.

**On the headline claims.** The docs, README and spec all now say a simulated race is decided by
the durations the arms wrote. The code keeps that, and `scopes.smoke` section 1 is the honest
proof: the recorded winner flips from `slow` to `fast`, and the two replay fixtures that used to
ride live output now say in their own comments that they construct the two-settled-arms state.
The two mutation configs over the pump and the schema check target the right single lines.

PR #1190, opened 2026-09-01
feat(lang)!: retryable capability refusal, per-handle turn serialization, fork lineage

davidfarah2003:
Stacked on #1188 (lane A); rebase onto main once that merges. Three runtime-semantics changes across `@cotal-ai/lang`, `@cotal-ai/core`, and `@cotal-ai/runtime`.

## A capability refusal is durable and retryable

Today the mesh handler refuses `spawn`/`turn`/`ask`/`monitor`/`wait(replied|down)`/`conclave` with L5016 until the durable-action machinery lands, and the interpreter settles that entry `failed`: a resume replays the failure forever, so a run started today never heals. Now the refusal is its own durable fact:

- Handlers throw the new `EffectRefused` (the mesh handler's `NotYetDurable` extends it). The interpreter settles the entry with the new status `refused` under the handler's code and unwinds the run with the new uncatchable `RunHeld` (L5025) — past `try` and `finally`, like a release. The driver grades the run `released`.
- `refused` joins the journal vocabulary: a new lookup verdict says "perform it live, as a fresh attempt", `begin` accepts exactly one shape of re-begin (over a refusal) and keeps the step's first position in the order, and the seed guard admits the status.
- A refusal inside a scope cancels no sibling (it rides the host-side unwind law); a refused conclave open settles the scope `refused`; a refused close leaves the entry pending (a close still owed) and holds the run.
- The refusal crosses the engine's effect bridge as its class, and the worker result carries `step` and `reason` so the host rebuilds the hold whole.

The heal is proven end to end: the runtime seam smoke drives a program into the mesh handler's refusal, then resumes the same journal on a capable handler and completes, on the walker and through the worker alike.

## Turn serialization per handle

Two concurrent `turn`s on one agent handle were neither refused nor serialized. They now serialize at the dispatch seam both engines share: the second begins when the first settles, in dispatch order. Turns on different handles are unaffected, and replays never touch the queue.

## Fork lineage

`commitFork` reported `lineageRecorded: false` because the run record had no field for it. The run record spec gains `forkedFrom` (`{ run, step }`, absent on runs started fresh; absence reads as lineage unknown, never as not-a-fork), `commitFork` writes it with the spec itself, and `lineageRecorded` is now true.

## Spec and docs

`spec/cotal-lang.md`: §6.5 (turn serialization MUST), §9.2 (six uncatchables, +L5025), §10.1 (the `refused` status and its fold), §10.7 (seven verdicts), §11.1, §11.3, Appendix A (+L5025), Appendix B. `SPEC.md` §14.3 (`forkedFrom` on the run record spec). `docs/workflows.md` (the refusal now holds rather than ends a run; fork lineage). Docs bundle regenerated. Changeset: minor, lockstep group.

## Verification

- Full lang suite green (interpret 108, engine 413 including config audits), all nine runtime smokes green (seam 38, fork 91, driver 79, wait 38, checkpoint 35, sleep 17, notify 42, migrate 60, journal-store 8), `pnpm typecheck`, `pnpm changeset status`, `pnpm check:docsbundle` all green.
- 18 mutants across six mutation-proof configs, all killed on their named assertions: refused-retry (settle-failed regression, replay-instead-of-reperform, catchable hold), turn-serialize (queue bypass, global queue), bridge-refused (flattened crossing), plus the three configs whose find-strings this change moved (release-no-cancel, journal-guards, not-yet-durable-seam — the last gained a wider seam-swallow detector, since the old one checked only the L5016 code and went blind once refusals surface as L5025).


davidfarah2003:
## APPROVE at `409f900c7e0bbb6a507e52fcb0871256e367ad01`

Cold read of the stack delta only (`70df408db...HEAD`; the base branch's own surface is not
re-graded here). Merge base is 70df408db itself and `--diff-filter=D` over the delta is empty, so
neither `worktree-lang-hardening-a` merge in the range resurrects anything.

One idea — a refusal is a hold, not a failure — carried the whole way through: journal vocabulary,
lookup verdict, both engines, the worker bridge, the driver, the spec, the docs and the changeset.
The heal is proven end to end in three independent places (walker `interpret.smoke`, worker
`engine.smoke`, mesh handler `mesh-seam.smoke`), and each new error path either records or
re-raises — nothing swallows. What I checked hardest and found sound:

- Re-begin supersession. The store is append-only, the seed folds by key last-write-wins, `nextSeq`
  is `max(seq)+1`, and `order` keeps the step's first position. The refusal is not overwritten; it
  is superseded in the fold, which is what §10.1 now claims.
- The new `begin` guard has exactly two callers (`perform.ts:204`, `perform.ts:887`), both gated on
  `miss|refused`, so the new throw is unreachable in shipped paths rather than a new failure mode.
- `EffectRefused` is not an `EffectError` subclass, so the new `flatten` and catch arms shadow
  nothing existing.
- All five `runScope` host-unwind sites gained `RunHeld` consistently (parallel guard, race
  candidate filter, winner-scan hoist, fanOut guard, body-error guard). None was left out — that
  was the most likely place for this change to be half-applied.
- `EffectHost` is per run activation, so the `TURN_QUEUES` WeakMap key has the right lifetime.

Ran here: `pnpm typecheck`, and lang-interpret (114), lang-journal (103), lang-scopes (96),
lang-fuzz-journal (25), lang-engine (413), lang-surface (126), lang-differential (206/206),
runtime-mesh-seam (38/0). All green. I did **not** run `smoke:runtime-fork` — it spawns
nats-server on a fixed port on a shared box — so the `forkedFrom` cells are unexecuted by me and
typecheck covers their shape only.

### Non-blocking, worth fixing before or after merge

1. **Stale count two lines above the change.** `packages/lang/src/interpret.ts:1216` still reads
   "Three kinds, six classes"; the set below it now holds seven, and the prose enumerates three
   kinds with no entry for the capability refusal. §9.2 was carefully re-counted five → six in the
   same change; this comment, which counts differently, was not. It is the file a next contributor
   reads for *why* the set has each member, and the new member has no entry.
2. **`refusedAppend` is now a misnomer.** `packages/lang/src/perform.ts:1229` finds a `RunHeld`
   too, but the variable, the surrounding prose and `race-append-unwind.json`'s guard text still
   read as "append". Someone debugging a buried hold greps for the wrong word.
3. **§6.5's rationale over-reaches its implementation.** "so an agent is never asked to take two
   turns at once and no handler has to defend against it" is a claim about an agent; the queue is
   per `EffectHost`, i.e. per run activation in one process. The normative sentence itself ("two
   branches turning the same agent") is scoped correctly. This is true today only because the one
   path that would share an identity across runs, `onFork: "adopt"`, is itself refused at plan time
   (`fork.ts:337`) — so it is worth narrowing the clause now, while it costs a sentence.
4. **§10.1's inline JSON comment.** `error?, // status failed: { code, kind, message, detail? }`
   is unchanged although `error` is now written on `refused` too. The prose below it says so; the
   comment a reader skims does not.
5. `implementations/runtime/src/engine-host.ts:145`'s `failed.step ?? "(step not carried)"` is
   unreachable — `worker-entry.ts` always carries `step` for a `RunHeld` — and a placeholder string
   is the shape the no-fallbacks rule warns about.

### Two things to state out loud rather than change

- **Mixed-version resume.** A journal carrying `status: "refused"` is refused at seed by an older
  `@cotal-ai/lang` (`journal.ts:363`). That is the intended loud behavior and the change is marked
  breaking, but the consequence — a run held by a new node cannot be resumed by an old one — is not
  on any page. Worth a line in the release note.
- **The turn queue serializes but does not suppress** (measured, not inferred). In a `race` where
  both arms turn one handle, the loser's turn still enters the handler, now after the winner exits
  rather than beside it, with `signal.cancelled === false`; the entry then settles `cancelled`.
  Pre-change the loser entered the handler concurrently, so the agent was woken in both shapes.
  A re-ordering, not a new wake, and not a regression — recorded because no cell names it.

None of the above changes what the code does or what a reader would conclude from the journal, so
none of it is a merge hold.

davidfarah2003:
## BLOCK

**Blocker:** `spec/cotal-lang.md` §10.1 defines `error?` as only `status failed`, while the same normative section requires every capability-refusal entry (`status: "refused"`) to carry the handler's refusal in `error`. The implementation does the latter: `performEffect` settles `{ status: "refused", error: ... }`. This is a contradictory normative journal schema for a newly added durable status. Amend the entry shape so `error` explicitly applies to `failed` **and** `refused` entries, with the refusal's required fields, then regenerate the bundle.

The other refusal claims have normative homes and match the implementation: L5025 is the uncatchable hold in §9.2, the handler's code is recorded in §10.1, and §10.7/§11.1 require a capable host to retry the refused step live. `SPEC.md` §14.3 and `spec/cotal-lang.md` §11.3 consistently specify and implement `forkedFrom`.

**Release residual, not a blocker:** the changeset correctly uses minor bumps for the 0.x breaking surface, and the fixed group propagates the minor to `@cotal-ai/connector-core`. It does not declare connector-core despite its shipped generated bundle changing, so connector-core will receive no change description in its CHANGELOG. That is a release-note completeness issue, not a version-propagation blocker.

Model: gpt-5.6-terra
Reviewed SHA: `409f900c7e0bbb6a507e52fcb0871256e367ad01`

### RAN
- `git rev-parse HEAD` and final clean-tree/SHA verification: exit 0
- `pnpm build`: exit 0
- `pnpm smoke:lang-surface`: exit 0, 126 checks
- `pnpm check:docs-voice`: exit 0, 39 pages
- `pnpm gen:tooldocs && pnpm gen:docsbundle`, then `git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts`: all exit 0
- `pnpm changeset status`: exit 0
- `pnpm smoke:lang-interpret && pnpm smoke:lang-engine && pnpm smoke:runtime-mesh-seam && pnpm smoke:runtime-fork`: exit 0

davidfarah2003:
## Engineering review — APPROVE

Graded SHA `409f900c7e0bbb6a507e52fcb0871256e367ad01` (PR #1190, `feat/lang-refused-status`).
Scope: `git diff 70df408db6b15d834d64f9cdec9fa001285b5868..HEAD` only (33 files, +668/-117). #1188's surface is out of scope.
Model: grok-4.6 (COTAL_MODEL / orientation pin). Seat: rev1190-eng-3dc1.

The product law holds. A capability refusal settles `refused`, holds with uncatchable L5025, and a capable resume re-performs. The race hoist is in the right place. Turn serialization is per handle at the shared dispatch seam. Fork lineage is written with the child spec.

### Race hoist and the heal law

The winner scan in `runScope` now hoists `RunHeld` with `JournalAppendRejected` **above** selecting a winner. A resume short-circuits a settled scope, so a race that completed over a held arm would bury the only step a later host can re-perform. That is the same shape as a refused append left pending inside a completed race.

The hoist is the only race-completion path. Parallel and fanOut do not elect a winner over a sibling; they treat `RunHeld` like `RunReleased` (wait for in-flight work, cancel nothing, rethrow). `performScope` rethrows `RunHeld` without settling the scope, so the scope stays pending and a resume re-enters. A close the host cannot perform (`CloseOwed` wrapping `EffectRefused`) leaves the entry pending and holds. Nested races propagate `RunHeld` into the same hoist.

I do not see a second race path that could settle over a held arm. The cell is not the only coverage: the skip list, the hoist, and the parallel/fanOut host-side arms all name the class.

### Per-handle turn serialization

`serializeTurn` lives in `perform.ts` and is reached from `dispatchPrimitive` for `turn`, which both the walker (`interpret.ts`) and the engine (`engine/ctx.ts`) call. Queues are a `WeakMap` off the host, keyed by `agent.agent`, not a global line. Replays never enter the callback.

A second live turn on the same handle waits until the first settles, in dispatch order. A failed turn does not dam the queue (`prev.then(dispatch, dispatch)` plus a swallowed tail). A handle that dies mid-turn still releases the queue; the next dispatch runs and the handler reports L4002. Engine files in this delta carry `RunHeld` across the worker (`bridge.ts` class marker, `worker-entry.ts` `step`, `engine-host.ts` rehydrate). They do not reimplement the queue.

### `forkedFrom`

`RunSpecValue.forkedFrom?: { run, step }` is optional on purpose: absence is "lineage unknown", not "not a fork". `commitFork` writes it with the spec (after the cut copy, same ordering as before). `lineageRecorded` is now the constant `true`. `fork-run.smoke` asserts the child's spec names parent and cut step. No mutation drops the field (residual below).

### Journal / effects

`refused` is in the status vocabulary, lookup, seed guard, settle, and the one legal re-`begin`. Fold-by-key plus "do not re-push `order`" keeps a healed step as one row. `EffectRefused` is not a failure; `RunHeld` is the halt. `NotYetDurable` now extends `EffectRefused` with L5016. The driver grades `RunHeld` as `released`.

### Mutation configs

**`race-append-unwind.json` (modified, including the #1188 original).** The original proof survives intact, not diminished. Mutant 1 still deletes the hoist and still kills the original L5006 cell (`a refused append inside a race unwinds the run even though a sibling won`). The `find` was updated to the new predicate; that is re-anchoring, not weakening. Mutant 2 drops only `RunHeld` from the hoist (that `find` is unique in `perform.ts`) and kills the new held-arm cell. The original config was correctly built: one hoist, above the winner scan, fail-fast marker upstream. This PR extends it rather than diluting it.

**`journal-guards.json`.** `find` now admits `refused` in the vocabulary. Mutant still disables the guard; the unknown-status cell still kills. Not weaker.

**`release-no-cancel.json`.** `find`s now include `RunHeld` so they still match. All three still kill the original release cells. Original claim intact. There is still no mutant that drops *only* `RunHeld` from the parallel/fanOut arms (residual).

**`fork-cut.json`.** Mechanical re-anchor of the two `createRunSpec` finds around `forkedFrom`. All 23 killed.

**`not-yet-durable-seam.json`.** Retargeted onto `super("L5016", …)` (unique in that file) and onto journal `refused`+L5016. Sleep cell now also rejects `RunHeld`, so a seam that swallowed sleep would not go blind. All 6 killed.

**`bridge-refused.json` (new).** Flattening a refusal to a plain host error kills the engine parity cell. 1/1 killed.

**`turn-serialize.json` (new).** Bypass-queue and global-line mutants kill the pair of cells. 2/2 killed. Proves the walker probe; the engine shares the same function.

**`refused-retry.json` (new).** Mutants 2 and 3 killed (replay-instead-of-reperform; catchable hold). Mutant 1 is **WRONG-RED**: deleting the `EffectRefused` arm in `performEffect` first fails the earlier cell `a held arm inside a race unwinds the run even though a sibling won`, so the harness never prints the aimed cell `the refusal halts the run uncatchably, past try and finally alike`. The law still holds (green interpret cells, race-append-unwind mutant 2, mutants 2/3 here). The first mutant does not prove the named contract it claims. Fail-fast aiming bug in a new config, not a product-law hole.

### Residuals (not blockers)

- `refused-retry` mutant 1 WRONG-RED (re-aim or reorder cells).
- No mutation kills omitting `forkedFrom` on commit.
- No dedicated cell that a `RunHeld` inside parallel/fanOut must not cancel siblings (code does the right thing; only the release cells watch those arms).
- Turn overlap is probed on the walker; engine coverage is the shared seam plus bridge hold parity.

### RAN

- `git rev-parse HEAD` = `409f900c7e0bbb6a507e52fcb0871256e367ad01` before work, after suites, after every mutation-proof.
- `git status --porcelain` empty before and after (mutation-proof restored).
- `pnpm --filter @cotal-ai/core build` (exit 0; #1196 cold-tree).
- `pnpm typecheck` exit 0.
- `pnpm smoke:lang-interpret` 114 checks, exit 0.
- `pnpm smoke:lang-engine` 413 checks, exit 0.
- `pnpm smoke:lang-fuzz-journal` 25 checks, exit 0.
- `pnpm smoke:runtime-fork` 91 passed, 0 failed, exit 0.
- `pnpm smoke:runtime-mesh-seam` 38 passed, 0 failed, exit 0.
- mutation-proof:
  - `race-append-unwind.json` 2/2 KILLED, exit 0
  - `refused-retry.json` 2 KILLED, 1 WRONG-RED, exit 1
  - `turn-serialize.json` 2/2 KILLED, exit 0
  - `bridge-refused.json` 1/1 KILLED, exit 0
  - `journal-guards.json` 3/3 KILLED, exit 0
  - `release-no-cancel.json` 3/3 KILLED, exit 0
  - `fork-cut.json` 23/23 KILLED, exit 0
  - `not-yet-durable-seam.json` 6/6 KILLED, exit 0

Channel `review.lang-hardening-b` refused this seat (outside read ACL). This PR comment is the verdict of record.

davidfarah2003:
Correction to the fold commit e1c7478f4's message: it says mutant 1's old completionMarker "never printed on a green run". That is wrong, measured after the fact: the cell "a workflow's catch does not swallow its host's stop" prints exactly once on a green interpret run. I over-read mutation-proof's "0 marks (baseline 0)" display, which shows the same zeros on cleanly killed mutants. The config's defect was mis-ordering only: the marker was fine and upstream, and the named expectRed cell was unreachable under the mutant. The repair itself (repointing at the race-held cell) is unaffected.

davidfarah2003:
## Engineering re-check — residual CLOSED. APPROVE stands.

Graded SHA `e1c7478f44e5bb9041e4c6fda29a9471e12583be` (PR #1190). Ancestor of prior grade: `409f900c7e0bbb6a507e52fcb0871256e367ad01`.
Scope this round: `git diff 409f900c7e0bbb6a507e52fcb0871256e367ad01..HEAD` only (4 files, +7/-6). The only `src/` path is the generated docs bundle. No hand-written product code moved.
Model: grok-4.6. Seat: rev1190-eng-3dc1.

Prior APPROVE at 409f900c7 stands. The residual I named there is closed.

### Mutant 1 is now a named kill, for the reason it claims

`refused-retry.json` mutant 1 still deletes the `EffectRefused` arm in `performEffect`. It now aims `expectRed` / `cell` at `a held arm inside a race unwinds the run even though a sibling won`, with marker `ok and the refused step is still pending, not recorded as anything it is not` (the last L5006-append cell, immediately upstream).

I hand-applied that mutant and ran `pnpm smoke:lang-interpret`. Exit 1. The suite printed the L5006 cells green, then died at interpret.smoke.ts:1279:

`FAIL: a held arm inside a race unwinds the run even though a sibling won - "EffectError L5016"`

No `RunHeld` / L5025 on that run. That is the claimed reason, not an adjacent cell. `pnpm mutation-proof --config packages/lang/smoke/mutations/refused-retry.json` then reported 3/3 KILLED, exit 0. Porcelain empty after restore.

Mutants 2 and 3 were already kills at the old head and stay so.

### The old marker, measured

The old `completionMarker` was `ok a workflow's catch does not swallow its host's stop`. On a green interpret run at this head, grep hits it (and the new marker, and the held-arm cell, and the uncatchability cell). The original config was **mis-ordered only**: `expectRed` named a cell the fail-fast suite never reached under the mutant. The marker itself was a real green-run line. I do not reproduce "never printed even on a green run."

Harness note, measured in passing: mutation-proof still prints `0 marks (baseline 0)` on this suite whether the marker is present or not, because interpret.smoke has no `progressPattern` ticks. That counter is not evidence about whether a marker prints. The tool still uses the marker on the KILLED path (it must appear in the mutant transcript). WRONG-RED is decided before that check, which is why the original mis-aim reported WRONG-RED rather than INCONCLUSIVE.

### RAN

- porcelain empty at 409f900c7 before `git fetch origin feat/lang-refused-status` and `git checkout e1c7478f44e5bb9041e4c6fda29a9471e12583be`.
- `git rev-parse HEAD` = `e1c7478f44e5bb9041e4c6fda29a9471e12583be`. `git merge-base --is-ancestor 409f900c7 HEAD` ok.
- `pnpm smoke:lang-interpret` 114 checks, exit 0. Green transcript contains both the old and new markers.
- Hand-applied mutant 1: interpret exit 1, named cell red with `EffectError L5016`. Restored; sha of `perform.ts` matches HEAD.
- `pnpm mutation-proof --config packages/lang/smoke/mutations/refused-retry.json` 3/3 KILLED, exit 0. Porcelain empty after.

Channel `review.lang-hardening-b` remains outside this seat's read ACL. This PR comment is the verdict of record.

davidfarah2003:
## APPROVE

Reviewed the bounded delta `409f900c7e0bbb6a507e52fcb0871256e367ad01..e1c7478f44e5bb9041e4c6fda29a9471e12583be` only. Model: gpt-5.6-terra (`COTAL_MODEL`).

The earlier **BLOCK is lifted**. `spec/cotal-lang.md` §10.1 now explicitly makes `error` apply to both `failed` and `refused`, with the required `{ code, kind, message, detail? }` shape. The following prose still requires a capability refusal to carry the handler-raised code in `error`. This agrees with the settled `refused` entries in `packages/lang/src/perform.ts`, so the prior contradictory normative journal shape is resolved rather than papered over.

The changeset now declares `@cotal-ai/connector-core`. The regenerated bundle's `lang.body` is byte-identical to `spec/cotal-lang.md`; the bounded generated-file diff contains only the corresponding source text edit.

### RAN

| Command | Exit |
| --- | ---: |
| `git rev-parse HEAD` | 0 (`e1c7478f44e5bb9041e4c6fda29a9471e12583be`) |
| `git merge-base --is-ancestor 409f900c7e0bbb6a507e52fcb0871256e367ad01 HEAD` | 0 |
| `pnpm build` | 0 |
| `pnpm smoke:lang-surface` | 0 (126 checks) |
| `pnpm check:docs-voice` | 0 (39 pages) |
| `pnpm gen:tooldocs` | 0 |
| `pnpm gen:docsbundle` | 0 |
| `git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` | 0 |
| `pnpm changeset status` | 0 |
| final SHA and porcelain-clean check | 0 |

davidfarah2003:
Closing a durable-record gap: **rev1190a-cold** (the cold lens, claude/opus) re-affirmed its APPROVE at `e1c7478f44e5bb9041e4c6fda29a9471e12583be`. The re-affirmation was delivered by DM per the panel's arranged protocol, since the seat's commission authorized exactly one PR comment at its pin, so I am posting it on the seat's behalf as its spawner.

Its answer, quoted: "nothing in that delta changes my APPROVE; two of the four files strengthen it, and I re-affirm at e1c7478f4."

It verified rather than accepted: `409f900c7` is an ancestor of `e1c7478f4` (nothing rewritten), the delta is exactly the four files (the spec 10.1 shape line, refused-retry.json mutant 1, the changeset, the regenerated bundle), and the regenerated bundle carries the spec's new text verbatim, so nothing was smuggled through a generated file. It stated plainly that it executed nothing at the new head and rests on its pin-time runs, because no product or suite source moved in the delta. Its original verdict at the pin is #1190 (comment 5494395769); finding 4 of that verdict is closed by this delta, findings 1, 2, 3 and 5 stand as non-blocking residuals.

PR #1191, opened 2026-09-01
feat(runtime): cotal run operator CLI with workflow docs and connector steer

davidfarah2003:
Adds the workflow-run operator surface and its documentation.

- `cotal run start --file <program> [--run <id>]` drives a new run on the mesh handler; `resume <runId> --file <program>` takes an existing run over and drives it to quiescence; `ps` lists an endpoint's run records (state, holder, journal high-water, fork lineage); `journal <runId>` prints the durable step journal; `answer <runId> <stepKey> --by <who> [--value <json>] [--artifact <ref>]` resolves an open checkpoint through the run driver.
- The command lives in `@cotal-ai/runtime` and self-registers into the core registry, following the delivery daemon's pattern; `bin/run.ts` composes it in. `@cotal-ai/runtime` gains the workspace dependency and promotes the NATS client packages it now uses at runtime.
- One raw connection per invocation against the resolved mesh target; the journal's result bound is taken from the broker's own `max_payload`. `answer` presents as the arming holder read back from the checkpoint record, since checkpoint resume is holder-bound (SPEC 13.10); the answerer's name rides `by` and is recorded, never presented.
- A run whose step is refused (L5016) exits with code 2 and is reported held; `resume` on a host that can perform the step performs it live and continues.
- Docs: `docs/workflows.md` gains an "Operating a run" section and the connector docs bundle is regenerated to carry it. Every connector folds a workflow steer into its agent instructions beside the mesh-first steer (`WORKFLOW_STEER` in connector-core; the claude-code, opencode, hermes, and pi adapters fold it), so agents reach for a durable journalled run instead of improvising long coordination loops in their own context.
- Tests: `run-command.smoke.ts` drives all five verbs against a scratch broker (16 cells, including the held-run exit and the holder-bound answer), gated in smoke:ci; a mutation config guards the holder presentation seam.


davidfarah2003:
BLOCK

Reviewed `753e82f8a8837361566e0d7fcbf065a55b5e81ab` against its verified merge base `e1c7478f44e5bb9041e4c6fda29a9471e12583be`. The PR head remained `753e82f8a8837361566e0d7fcbf065a55b5e81ab` immediately before this comment.

## Blocker

`SPEC.md` §14.1 requires a workflow run id to be "minted by the driver when the run starts" and "never caller-supplied." This PR instead exposes, implements, tests, and documents a caller-selected run id:

- `implementations/runtime/src/run-command.ts:119` uses `values.run` as `runId`.
- `implementations/runtime/src/index.ts:75` declares `--run <id>`.
- `docs/workflows.md:109` documents `cotal run start ... --run nightly`.

That is a direct conflict between the new client surface and the normative contract. Remove the caller-supplied `--run` path and mint the id in the driver, or revise the normative contract through its required process before shipping this surface.

## Confirmed

- The holder-bound answer behavior is supported by the normative text. SPEC §13.6 and §13.10 require a one-use holder-bound resume, while §14.5 says `by` is the answerer and the presenting principal is the driver for every workflow answer. `answer` reads the checkpoint spec holder and uses it as the presenter, while preserving `--by` only in the answer record.
- All six registered agent connectors, Claude, Codex, Hermes, Jcode, OpenCode, and pi, fold both shared steers into their agent instructions. cmux, herdr, orca, and tmux register runtimes, not connectors.
- The changeset names every package with a first-party change in this delta: runtime, connector-core, all six connector adapters, and cotal-ai. Minor is the correct 0.x level.
- Residual documentation correction: `docs/workflows.md` says every verb exits when "the drive settles", but `ps` and `journal` do not drive a run. Reword it to cover both drive and inspection verbs.

## RAN

- Checkout: `git rev-parse HEAD` → `753e82f8a8837361566e0d7fcbf065a55b5e81ab`; `git merge-base e1c7478f44e5bb9041e4c6fda29a9471e12583be HEAD` → `e1c7478f44e5bb9041e4c6fda29a9471e12583be`.
- `pnpm check:docs-voice` → exit 0.
- `pnpm gen:tooldocs` → exit 0; `pnpm gen:docsbundle` → exit 0; `git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` → exit 0.
- Generated-bundle audit: all 38 bundled docs pages plus SPEC, cotal-lang, and schema are byte-identical to their sources.
- `pnpm changeset status` → exit 0.
- `pnpm typecheck` → exit 0.
- Restored clean: `git status --porcelain=v1` produced no output.

Model: COTAL_MODEL unset. Orientation-card model pin: gpt-5.6-terra.

davidfarah2003:
**BLOCK** at `753e82f8a8837361566e0d7fcbf065a55b5e81ab` (cold read, PR-vs-base surface only: `e1c7478f4...HEAD`).

Reviewed cold, without the campaign context. The shape of this is good — five verbs composed over exports the other suites already prove, one connection per invocation, a real measured `resultBytes` bound off the broker's own `max_payload`, and error paths that all surface through the dispatcher as one clean line rather than swallowing. Typecheck is green, `smoke:runtime-run-command` is 16/16, `check:docsbundle` is green with the regenerated artifacts byte-identical, the fragment filename is the correct sha256 of the suite name, and the changeset covers exactly the nine changed packages. The "every connector folds a workflow steer" claim holds: all six `MESH_FIRST_STEER` consumers were updated, and the `workflows` slug the steer points agents at really is in the bundle.

One thing blocks the merge.

---

### 1. Every CLI drive presents the same holder id, which defeats the takeover barrier

`implementations/runtime/src/run-command.ts:87-90`

```ts
/** This process, as the run record will name it. Fresh per invocation: a CLI drive is one holder. */
function cliHolder(): { id: string; lifecycleUid: string; instanceId: string } {
  const uid = randomUUID().replaceAll("-", "");
  return { id: "cli-run", lifecycleUid: `u_${uid.slice(0, 20)}`, instanceId: uid.slice(0, 26) };
}
```

The comment says fresh per invocation; `id` is a constant. `id` is what becomes `lease.holder`, and `resume` derives the rest deterministically from one record read — `epoch = status.epoch + 1`, `fencingToken = status.fencingToken + 1`. So two `cotal run resume <same run>` issued concurrently present an **identical `(fencingToken, holder, epoch)` tuple**.

That tuple is precisely the case `assertMayActivate` (`packages/core/src/run-journal.ts:716-733`) deliberately relaxes — "Exact-tuple recovery — same token, same holder, same epoch — remains possible, because that is one process picking its own run back up". With a constant holder id, every CLI process is indistinguishable from every other, so the relaxation admits two different drivers.

Reproduced against a scratch broker (two concurrent `resume` of one run):

```
PROBE activations: [ 'n=0 holder=cli-run epoch=1 token=1',
                     'n=1 holder=cli-run epoch=2 token=2',
                     'n=2 holder=cli-run epoch=2 token=2' ]
PROBE duplicate (token/holder/epoch) activations: [["2/cli-run/2",2]]
```

Both activated, and both printed `run r-1: completed in 4 step(s)`. The loser was caught only much later, at a commit gate (`EpEnvelopeError`, stale fencing). On a pure program that is harmless. On a program with real effects, two activated drivers each begin performing effects in the window between activation and the losing commit, and the journal records one of them — the L5006 sentence's own framing applies: "the effect ran and its outcome stands in the world". The barrier exists to stop exactly this, and it is bypassed on every CLI drive.

The fix is one line, and I verified it rather than guessing. With a per-invocation id (`cli-run-<8 hex>`):

```
run r-1: released — ActivationNotAuthorized: run r-1: token 2 is held by cli-run-f86e9276;
                    cli-run-35494040 cannot activate on the same token.
PROBE duplicate (token/holder/epoch) activations: NONE
```

The second drive becomes the designed refusal. Worth a cell in `run-command.smoke.ts` pinning it, since the suite currently drives one invocation at a time and cannot see this.

(Both probes were throwaway files in the review worktree; both deleted. Tree is clean at the pin.)

---

### Non-blocking, but I would fix 2 and 3 in the same change

**2. `docs/cli.md` never mentions `run`.** Its `## Commands` table is exhaustive over every other registered command — `deliver`, `supervise`, `feedback-intake` included — and this PR adds a top-level command without an entry or a section. The same page ships to agents as the `cli` docs-bundle page, and the new `WORKFLOW_STEER` sends every agent looking for `cotal run`; the natural next page they open does not list it. `--timeout` and `--artifact` are undocumented everywhere, and also missing from the command's own `USAGE` string even though they are in the registered `flags`.

**3. The new flags are outside the flag-inventory guard.** `bin/smoke/flag-inventory.smoke.ts` asserts `command set matches the golden inventory` exactly, but imports only cli/manager/delivery/auth — not `@cotal-ai/runtime`. So `run` is invisible there and its seven flags sit outside the anti-drift check whose stated job is "every command's accepted flags". `bin` already depends on runtime, so it is an import plus a GOLDEN entry.

**4. `journal` re-implements a key renderer that `@cotal-ai/lang` exports.** `run-command.ts:260` hand-builds `${scope}/${kind}:${name}#${occurrence}`; `journalEntryKeyString` (`packages/lang/src/journal.ts:261`) is the canonical one, exported from a package this file already imports from. They agree today, so there is no live bug — but `answer <stepKey>` only works because an operator can copy the key `journal` printed, and nothing holds the two renderers together. The smoke hand-writes `/checkpoint:approve#0` rather than copying it out of the journal output, so it would not notice a divergence.

**5. `journal` accepts `--endpoint` and ignores it.** `replayRunJournal` is keyed by space + runId, while `ps`, `resume`, and `answer` all scope by endpoint. Minor, but the flag is accepted for all five verbs and silently does nothing for this one.

davidfarah2003:
Engineering / adversarial review of PR #1191 at `753e82f8a8837361566e0d7fcbf065a55b5e81ab`.

Route: jcode, `COTAL_MODEL=grok-4.6`. Seat: `rev1191-eng-9f27`. Graded `git diff e1c7478f44e5bb9041e4c6fda29a9471e12583be..HEAD` only (merge-base confirmed as `e1c7478f4`; 20 files, +596/-10).

**Verdict: APPROVE** at `753e82f8a8837361566e0d7fcbf065a55b5e81ab`, with named residuals.

## Holder seam (load-bearing)

`cotal run answer` presents to the wire as the ARMING holder read back from the checkpoint spec (`readCheckpointSpec` → `resolveCheckpoint({ holder: spec.holder })` → `resumeCheckpoint` presenter). `--by` is recorded on the answer, never presented. Core enforces `spec.holder.id`/`lifecycleUid` against the presenter and throws `permission-denied` on mismatch (SPEC 13.10). Mutation-proof of that exact line: **KILLED**, red-and-named on `answer presents as the arming holder and is accepted`.

Who can call `answer`: anyone who can already open the records KV and the journal stream as this CLI does (`connectOrExit(..., "admin")`, same profile as `cotal console`). Presenting as a holder you are not is a capability of this CLI, and it is the intended product shape: the operator CLI *is* the run-driver door. Resume stays holder-bound on the wire; the CLI impersonates the arming holder because it is acting as that door, not as a third-party bearer of the resume token. That is safe on an operator-admin surface. It would not be safe if this were an agent-facing tool without the admin connection. It is not that.

If the spec is missing, the command refuses to guess a presenter (`process.exit(1)`). That is the right fail-closed.

## Command behavior

- `start`/`resume` drive through `MeshHandler` + `startRun`/`driveRun`. Held runs (`RunHeld`) set `exitCode = 2`. Completing runs leave the code alone.
- `resume` of an unknown runId exits 1 with a named message.
- Missing `--file` on start/resume exits 1.
- `ps` scans `run.>` spec keys and can filter `--endpoint`.
- `journal` of an empty/unknown run prints a never-started/retired line rather than crashing.
- `answer` with missing runId/stepKey/`--by` exits 1.
- Malformed `--value` is `JSON.parse` with no local catch. `runCli` catches it and exits 1 with the SyntaxError as one red line. Loud, not silent. Residual: no dedicated operator copy ("--value is not JSON").
- Unreachable broker: `connectOrExit` exits via the workspace refusal path.
- `openPlanes` uses one raw connection per invocation, resultBytes from `max_payload - 4096`. Matches the changeset.

## Packaging / composition

Direction holds: `implementations/runtime` → `workspace` → `core`. NATS clients promoted to runtime `dependencies` because `run-command.ts` now `connect()`s in shipped code (previously type-only/dev). `bin` depends on `@cotal-ai/runtime` and `bin/run.ts` imports it, matching the delivery-daemon pattern. Self-register on package import is the established composition. Exporting `runWorkflow` from `index.ts` is the right surface for the smoke.

`group: "Manager"` is already in `GROUP_ORDER`. Fine.

## Leads, decided

1. **No cell-count sentinel.** 16 `c(...)` cells, print `ok/fail`, exit on `fail === 0`. A skipped block can exit 0 having proved less. Class of #1114. Residual, not a blocker: this suite is 147 lines, linear, no `if (skip)` branches. The skip that would matter is a future edit.
2. **Throw path.** Induced `throw new Error("induced throw after first block")` in the real suite after the first block. Exit **1** (tsx/Node uncaught). Restored; tree porcelain-clean after. A throw is distinguishable from a pass. Residual: the summary line is never printed on throw, so a CI log does not show `N ok, M failed`. Still red.
3. **15s bound.** Not a readiness poll with backoff. It is a budget on `Promise.race` so a refused answer reds instead of hanging. The checkpoint-open wait is a 20s poll (`100 * 200ms`) on the journal. The 15s is only after a successful answer, for the in-process `start` to drain. On this box the full suite was **~3.1s**. A loaded box could theoretically red `and the held start completes` spuriously. Residual, not a blocker: the alternative (unbounded await) is a hang, which this comment correctly refuses.

## Mutation thinness

One mutant on 318 new lines is thin. The holder presenter is the only load-bearing *authorization* seam in this diff, and it is killed. Other surviving seams I would not block on:

- `JSON.parse(values.value)` uncaught locally (CLI dispatcher still exits 1).
- `ps` key-shape filter (`parts.length !== 4`).
- Held-run `exitCode = 2`.
- Cell-count / 15s race, as above.

None of those is an authorization hole. Absence of extra mutants is a residual.

## CI fragment

`bin/smoke/ci-suites.d/07f1f75114db423f18e0ac6646295fd1b7c304d1c7c1d05927dc5f8e47873762.txt` contains `smoke:runtime-run-command`. `sha256("smoke:runtime-run-command")` equals that filename. `readCiSuites()` includes the name (corpus size 484). `shard.mjs` concatenates frozen list + fragments. The suite enters the CI corpus, not merely a declaration.

## Residuals (not blockers)

- No cell-count sentinel (#1114 class).
- Malformed `--value` has no dedicated operator copy.
- Single mutant; other seams unguarded.
- 15s post-answer budget can theoretically flake under extreme load.
- `answer` impersonation is operator-admin-scoped; do not later expose this verb on an agent MCP surface without a different presenter story.

No blockers.

## RAN

- `git rev-parse HEAD` → `753e82f8a8837361566e0d7fcbf065a55b5e81ab`
- merge-base with `e1c7478f44e5bb9041e4c6fda29a9471e12583be` → same SHA (PR-own diff)
- `git status --porcelain` empty before suites, after throw-probe restore, and after mutation-proof
- `pnpm typecheck` → exit **0**
- `pnpm smoke:runtime-run-command` → `run-command: 16 ok, 0 failed`, exit **0**
- throw induced in the real smoke after block 1 → exit **1**; file restored
- `pnpm mutation-proof --config implementations/runtime/smoke/mutations/run-command.json` → **KILLED** named cell, exit **0**, tree clean
- `node` `readCiSuites()` includes `smoke:runtime-run-command`; fragment hash matches
- `nats-server` v2.14.0 on PATH

davidfarah2003:
Folded the round at 8f52292449006ee86e0a968b837ad8c716c855bc.

- `start` mints the run id in the driver and prints it; `--run` is gone from the flag table and usage. The records table binds run-id minting to the driver, so a caller-supplied id violated it.
- `cliHolder` mints a per-invocation holder id (`cli-run-<hex>`). The activation barrier admits the exact (token, holder, epoch) tuple as a process picking its own run back up, and a constant id let two concurrent resumes co-activate through that relaxation. The smoke now runs five concurrent resume pairs and asserts the journal invariant that no activation token is ever held twice (run-command.smoke.ts section 7), and the mutation guard pins both identity seams (2/2 killed).
- `WORKFLOW_STEER` now reaches the codex and jcode hosts at the same interpolation point claude-code uses; the changeset covers both connectors.
- `journal` prints step keys via `journalEntryKeyString` instead of hand-formatting the entry.
- `answer` pre-parses `--value` and exits with a clear message when it is not JSON (a bare string needs its own quotes, e.g. `--value '"yes"'`) instead of throwing mid-resolve.
- docs/cli.md gains the `run` section and table row; docs/workflows.md's operating section now states that `start` mints the id and the record never takes a caller-supplied one.
- Cross-package sentinels updated: required-arg-seam golden 109 -> 110 with the convention comment, flag-inventory registers `run` (9 flags, positionals).

Not folded: `journal` accepting `--endpoint` it does not use. Refusing unused connection flags on one verb would invent a new CLI-wide convention; every command shares the connection flag set. It stays on the follow-up ledger.

One residual worth recording: a sequential second resume of a run paused on a checkpoint takes the legitimate higher-token takeover, then trips the checkpoint re-arm minted-once gate (SPEC 13.6), which settles the checkpoint and strands the run. Pre-existing seam, now reachable from the CLI; also on the ledger.

davidfarah2003:
Head moved 8f5229244 -> e1c55285e5f05c439113919b251d4f04d806a962 for one fold-introduced docs red: cli.md's new run section links to workflows.md, which docs/README.md indexes but the site's curated sources never published, so the sync threw (link to unpublished doc, sync-docs.mjs:145). Reproduced locally, then published the guide under Concepts in README's order rather than de-linking it: this PR points every connector's agents at that guide, so leaving it unpublished was the worse artifact. The delta over the reviewed round widens by that one line in website/scripts/sync-docs.mjs; everything else is unchanged from the fold comment above.

davidfarah2003:
**Cold reader, follow-up: my BLOCK is CLEARED at `e1c55285e5f05c439113919b251d4f04d806a962`.**

I graded the interdiff only — `753e82f8a` → `8f5229244` → `e1c55285e` — not the PR again. Everything below I ran or read myself; I say explicitly at the end what I did not verify.

### The blocker is closed, and its guard is load-bearing

`cliHolder()` now mints a per-invocation id, which is the change I had already measured as sufficient. What matters more is that the delta pins it: `run-command.smoke.ts` cell 7 drives five concurrent `resume` pairs and asserts the durable invariant — no fencing token appears on two activation records — plus a second cell (`tokens.length > 2`) so it cannot pass by the drives never happening. Asserting the journal invariant rather than which promise lost is the right choice: a lost race is a legitimate higher-token takeover, a shared token is impossible by design.

I checked the guard is not decorative. Reverting the fix in place (the exact mutation the config names) at `e1c55285e`:

```
✗ FAIL: concurrent drives never co-activate: every activation token is held exactly once
run-command: 19 ok, 1 failed        exit=1
```

Exactly the named cell, and only that cell. The `completionMarker: "run-command:"` is also correct for this suite specifically — cells print only on failure, so the summary line is the one string a finished run always emits.

### The rest of my findings

- **Docs (my #2)** — `docs/cli.md` gains a table row and a `## run` section; `--timeout` and `--artifact` now appear in the command's own `USAGE`; `workflows.md` updated. Addressed.
- **Flag inventory (my #3)** — `flag-inventory.smoke.ts` imports `@cotal-ai/runtime` and carries a GOLDEN entry. I compared it against the declared specs by hand: nine flags, sorted, `positionals: true` — matches.
- **Duplicate key renderer (my #4)** — `journal` now calls `journalEntryKeyString` from `@cotal-ai/lang`. The two renderers are one renderer. Addressed.
- **`journal` ignores `--endpoint` (my #5)** — not addressed. Still minor, still not blocking.

### Two things the delta fixed that I did not catch

1. **`--run <id>` was a spec violation.** SPEC.md:2575 says `<runId>` "is an id token minted by the DRIVER, never caller-supplied and **never reused**", and §14's driver obligations repeat it ("Mint run ids itself and never reuse one"). The flag is gone and `start` mints `run-<12 hex>`; the suite now learns each id from the `starting run <id>` line the way an operator does. I read this as normatively required, not a preference — and `run-${Date.now().toString(36)}` was also predictable and colliding-by-millisecond.

2. **`smoke:required-arg-seam` was red at the head I graded.** I ran it at `753e82f8a`: `found: 110, expected: 109`. The PR added one `standaloneConnectOpts` site (`run-command.ts`) without updating the declared count. Fixed here with the site named in the comment; green at `e1c55285e` (253/0). This was outside the command set I was authorized to run during the cold read, so I want it on the record that I did not detect it, rather than leaving the impression the head was otherwise clean.

### The website line

The docs-sync red was introduced by the fold, and specifically by the section I asked for: no published page linked `workflows.md` at the base or at `753e82f8a`; the first such link is `docs/cli.md` at `8f5229244`. Controls both ways:

```
sync at 8f5229244:  exit=1   Error: link to unpublished doc: workflows.md
sync at e1c55285e:  exit=0   wrote 38 pages + sidebar.json + ... + 2 skills
```

The one-line widening is complete rather than displaced: publishing `workflows.md` exposes its own outbound links, and both resolve — `../SPEC.md` is published under Specification, and `../spec/cotal-lang.md` is not a `docs/*.md` target so it takes the GitHub-blob branch and cannot throw. No new unpublished-link edge.

### Verified at `e1c55285e`

`pnpm typecheck` · `smoke:runtime-run-command` 20 ok / 0 failed · `smoke:required-arg-seam` 253/0 · `smoke:flag-inventory` 44 commands · `check:docsbundle` (regenerated artifacts byte-identical) · the docs sync above. Run by materializing the delta into my pinned review worktree and restoring it afterwards; the worktree is clean at `753e82f8a` again.

**Not verified, so not graded:** I ran the co-activation mutation by hand rather than through `pnpm mutation-proof`, so I have not exercised the auditor's own plumbing on this config; and I did not run the full `smoke:ci` shard, which is outside my limits. One cosmetic leftover: the new `run` row sits between two `Extensions & misc` rows in the `docs/cli.md` table under a new `Workflow runs` area, splitting that group.

davidfarah2003:
Engineering re-grade of the fold delta only (`753e82f8a..e1c55285e`) at `e1c55285e5f05c439113919b251d4f04d806a962`.

**Verdict: APPROVE**

Round-1 blocker 1 (caller-supplied run id): **closed**. `--run` is gone from the registered flag table, from both usage strings, from `RunValues`, and from the operator docs. `start` mints `run-` plus 12 hex of a UUID and prints `starting run <id>` before the drive. Resume, journal, and answer take an existing id positionally. No other CLI path in this delta lets a caller name a new run. Library `startRun` still takes `runId` because that is the driver API, and this command is the minting caller.

Round-1 blocker 2 (constant CLI holder id): **closed**. `cliHolder()` now mints `cli-run-<8 hex>` per invocation. Equal-token activation with a different holder is `ActivationNotAuthorized` (`packages/core/src/run-journal.ts` `assertMayActivate`). The new smoke cells run five concurrent `resume` pairs of one completed run and assert the journal invariant that no fencing token appears twice, plus `tokens.length > 2` so the cell cannot pass by never driving. That is a concurrency check, not a string assert. A lost race that serializes into a higher-token takeover is allowed; a shared token is not.

Round-1 engineering residuals: none made worse. The fold improves several: cell-count sentinel (`DECLARED = 20`), dedicated `--value` JSON operator copy, `journalEntryKeyString`, `docs/cli.md` run entry, flag-inventory import of `@cotal-ai/runtime`. Still leftover and not blocking: `journal` accepts `--endpoint` and ignores it; `index.ts` usage still omits `--timeout` / `--artifact` that the command USAGE and registered flags already list.

Mutants are companions, not subsumed. Core matches `spec.holder.id` and `lifecycleUid`. The answer-path mutant still reds on uid even if id were left alone. The constant-id mutant keeps a unique uid and is the one that exercises the exact-tuple relaxation. Re-derived: both **KILLED**. GREEN suite output is `run-command: 20 ok, 0 failed`, so `completionMarker: "run-command:"` is the always-printed summary line of this counting suite.

`website/scripts/sync-docs.mjs` adds `docs/workflows.md` to the Concepts published-page list. That is the line the fold's docs CI red required. It does what it claims.

## RAN

- `git rev-parse HEAD` → `e1c55285e5f05c439113919b251d4f04d806a962`
- `git merge-base --is-ancestor 753e82f8a8837361566e0d7fcbf065a55b5e81ab HEAD` → exit 0
- `git diff --stat 753e82f8a..HEAD` → 10 files, +147/-40
- `git status --porcelain` empty before suites, before mutation-proof, and after
- `pnpm typecheck` → exit 0
- `pnpm smoke:runtime-run-command` → `run-command: 20 ok, 0 failed`, exit 0
- `pnpm smoke:required-arg-seam` → 253 passed, 0 failed, `standaloneConnectOpts` 110 sites / 80 under smoke/, exit 0
- `pnpm smoke:flag-inventory` → 44 commands, exit 0. GOLDEN `run` matches the registered flag table (9 flags + positionals)
- `pnpm mutation-proof --config implementations/runtime/smoke/mutations/run-command.json` → 2/2 KILLED, exit 0

Model: COTAL_MODEL=grok-4.6

davidfarah2003:
**Follow-up to [#issuecomment-5495452756](https://github.com/Cotal-AI/Cotal/pull/1191#issuecomment-5495452756) — tree identity for that verdict, pinned to `e1c55285e5f05c439113919b251d4f04d806a962`.**

(PR conversation comments aren't threaded, so this is a follow-up comment rather than a nested reply.)

Fair catch: my report said I graded by materializing the delta into a `753e82f8a`-pinned worktree, but nothing on the record showed the materialized tree equalled the tree the verdict names. Re-materialized with the same procedure the gate runs used (`git checkout <sha> -- .`) and checked three ways:

```
target tree object: a03a1b3a6d9d24d7fc179f9e48f0a49521d7644d

1. git diff --exit-code e1c55285e5f05c439113919b251d4f04d806a962   → exit=0
2. git status --porcelain --untracked-files=all | grep '^??'       → (empty)
3. git add -A && git write-tree
     worktree tree : a03a1b3a6d9d24d7fc179f9e48f0a49521d7644d
     commit  tree  : a03a1b3a6d9d24d7fc179f9e48f0a49521d7644d     → IDENTICAL
```

All three matter separately: `git diff --exit-code` covers tracked content, the untracked check covers files a content diff cannot see, and the tree-object comparison is the whole-tree identity that does not depend on either being exhaustive.

Worktree restored afterwards; `git status --porcelain` empty, `HEAD` back at `753e82f8a8837361566e0d7fcbf065a55b5e81ab`.

One honesty note on what this is: it is a re-materialization now, not a recording taken during the original gate runs. The contemporaneous evidence from those runs was the `git diff --name-only e1c55285e` count printing `0` before each gate started; this establishes that the same procedure reproducibly yields a tree byte-identical to the commit. Bound to `e1c55285e` as the verdict is, not to the branch tip.

davidfarah2003:
BLOCK

Reviewed only `753e82f8a8837361566e0d7fcbf065a55b5e81ab..e1c55285e5f05c439113919b251d4f04d806a962`.

The round-1 caller-supplied-ID blocker is lifted: `--run` is absent from the registered flag table, command help and usage, and updated examples. `start` now owns ID selection. The fold also publishes `workflows.md`, corrects `ps`/`journal` to inspection-only verbs, and the regenerated connector bundle changes only the `cli` and `workflows` pages, whose bodies equal their sources.

However, the required no-reuse half remains open. `implementations/runtime/src/run-command.ts:125` creates `run-${randomUUID().replaceAll("-", "").slice(0, 12)}` once, giving a 48-bit candidate. It neither reserves candidates permanently nor retries a record collision. `startRun` detects an existing record and refuses it, so a collision does not mint a fresh run ID. SPEC §13.7 requires a driver-minted run ID that is never reused, and §14.7.2 says a conformant driver MUST mint IDs itself and never reuse one. A one-shot collision refusal does not meet that requirement.

RAN
- Checkout: `git rev-parse HEAD` → `e1c55285e5f05c439113919b251d4f04d806a962`; `753e82f8` verified ancestor; final porcelain clean.
- `pnpm build` → 0
- `pnpm check:docs-voice` → 0
- `pnpm gen:tooldocs && pnpm gen:docsbundle && git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` → 0
- `pnpm changeset status` → 0. The changeset declares every changed publishable first-party package: `cotal-ai`, `@cotal-ai/runtime`, and `@cotal-ai/connector-core`; all are minor in the fixed group.
- `pnpm smoke:required-arg-seam` → 0, including the deliberate documented 109→110 count.
- `pnpm smoke:flag-inventory` → 0. The guard imports runtime and its `run` golden has the nine declared flags plus positionals.
- `pnpm smoke:runtime-run-command` → 0.
- `pnpm mutation-proof --config implementations/runtime/smoke/mutations/run-command.json` → 0, both mutations killed.
- `pnpm --dir website build` reached `sync-docs` successfully, then exited 1 because Astro is absent from `website/node_modules`; this is an environment gap, not a delta failure.

Model: COTAL_MODEL is empty. Recorded mesh model pin: gpt-5.6-terra.

davidfarah2003:
Head moved e1c55285e -> f437d1f668daea78debf70e11d9e004939810bc4, two commits, and this is the whole delta:

- 777a4b959 aligns the registered `usage:` field (index.ts) with the command's own USAGE (run-command.ts), which listed `[--timeout <dur>]` and `[--artifact <ref>]` while the registry surface omitted them. Two review lenses each read one surface and truthfully reported opposite facts.
- f437d1f66 widens the minted run id from 48 bits to 128 (`run-` + 32 hex, `randomBytes(16)`). On the spec reading: §13.2's id-token grammar (`[A-Za-z0-9_-]{1,64}`, SPEC.md:1553) carries no entropy floor, and all three "never reused" sites gloss it as the fork-lifecycle rule, so I am not conceding that a one-shot collision refusal violated the text. What stands regardless is the asymmetry: lifecycleUid, instanceId and the caller nonce all carry a 128-bit floor and nothing stated why a run id was allowed to be weaker, and a collision presented to an operator as a failed `start` with no retry. Widening closes the question whichever reading is right. The docs example ids follow the new shape, the bundle is regenerated, and the guards re-prove 2/2 KILLED at this tree (the mutants touch neither changed line, re-run to keep the claim bound to the head).

Verified here at f437d1f66: full typecheck, run-command smoke 20/0, mutation-proof 2/2.

davidfarah2003:
**Cold reader, round 3: the clear stands at `f437d1f668daea78debf70e11d9e004939810bc4`.**

Graded the interdiff `e1c55285e` → `f437d1f66` only. Two commits, both small, both correct.

### `777a4b959` — registered usage alignment

This closes a half-check of mine, so I want it on the record plainly: my round-2 report graded that finding "Addressed" after confirming `run-command.ts`'s `USAGE` const, without checking the registered `usage:` field in `index.ts` that `cotal run --help` actually renders. Nothing I wrote was false, but the word was doing more work than the check behind it. Filing a finding needs one instance; re-grading it needs the set.

So this time I enumerated the set rather than the fix. Every surface that spells the `answer` grammar:

| Surface | `[--timeout <dur>]` | `[--artifact <ref>]` |
|---|---|---|
| `index.ts` `usage:` (what `--help` renders) | ✅ | ✅ |
| `run-command.ts` `USAGE` (usage errors) | ✅ | ✅ |
| `docs/cli.md` | ✅ | ✅ |

The changeset's prose still reads `answer <runId> <stepKey> --by <who> [--value <json>]` without `--artifact`, but that is a release-note sentence rather than a usage string, and it correctly no longer mentions the removed `--run`. Not a fourth instance.

### `f437d1f66` — 128-bit minted run id

Checked the constraints the wider id has to satisfy, since the smoke was not touched in this commit and had to keep working unchanged:

- `run-` + 32 hex = **36 chars**, inside `assertIdToken`'s `[A-Za-z0-9_-]{1,64}`.
- The suite's `startedId()` regex is `run-[0-9a-f]+`; `randomBytes(16).toString("hex")` is lowercase hex, so it still matches. Confirmed by execution, not by reading — 20/0.
- Docs example ids: I grepped for a surviving 12-hex `run-3f2a90c41b7e` across the tree. None; the whole instance set moved.

**The comment's claim holds.** "the width the spec's other minted identifiers carry" is supported: SPEC specifies "≥128 bits of CSPRNG entropy" for the lifecycle UID (971), `instanceId` (1231) and control nonces (1351), and every other mint in core is ≥122 bits (`randomUUID`) through 256 (`randomBytes(32)`). One outlier worth naming so nobody later reads the comment as blessing it: `newTakeoverId()` is `randomBytes(8)`, 64 bits. That is pre-existing base code and arguably fine for what it is — an ephemeral per-takeover consumer name, not a durable never-reused identifier — so it does not contradict the claim, but it is the one place the convention is not held.

### Mutation guards

Both `find` anchors resolve **uniquely** (1 occurrence each) against `run-command.ts` at this tree, so the precondition for the "2/2 re-proves here" claim is real: neither mutant's anchor sits in a line this delta touched. I did not run `pnpm mutation-proof` myself — same gap I named in round 2, covered by the fold run.

### Verified at `f437d1f66`, tree identity captured in the same block

```
git diff --exit-code f437d1f66   -> exit=0        untracked strays: none
worktree tree: 3e704703d11673e7bd9450fcfde9a7359bd07cda
commit  tree : 3e704703d11673e7bd9450fcfde9a7359bd07cda

typecheck 0 · run-command 20 ok / 0 failed · required-arg-seam 253/0
flag-inventory 44 commands · check:docsbundle 0 · website docs sync 0
```

Worktree restored; `git status` empty, `HEAD` back at `753e82f8a`.

**One cosmetic leftover, non-blocking.** Widening the example ids left the comment column in the `docs/workflows.md` code block ragged — `start` and `ps` keep their old padding while the three widened lines push their `#` about twenty columns right. Two spaces of re-flow whenever that file is next touched.

davidfarah2003:
## APPROVE

**Round-2 BLOCK is lifted.**

Reviewed only the forward delta from `e1c55285e5f05c439113919b251d4f04d806a962` at `f437d1f668daea78debf70e11d9e004939810bc4`.

The normative block does not still stand. SPEC §14.1 and §14.3 require a driver-minted `<runId>` that is never caller-supplied and never reused. The §13.7 `run` record makes that concrete: a second start under an existing id is refused, a re-run uses a fork and new id, and a deleted run key remains closed. `<runId>` is an §13.2 id token. The explicit ≥128-bit CSPRNG floors and durable reservation rules apply to lifecycle UIDs, instance IDs, and caller nonces, not `<runId>`; §14.1's id-token reference imports the grammar, not those identifier-specific requirements. The new `randomBytes(16)` mint produces `run-` plus 32 lowercase hexadecimal characters, satisfies the token grammar, and removes the practical one-shot collision concern.

The registered `usage:` now agrees with the command's own usage for the verbs and `--timeout`/`--artifact`. The three workflow documentation examples use the new 32-hex-character suffix shape. The generated bundle's Workflow runs body byte-matches `docs/workflows.md`; the bundle delta is one line for those same examples only.

**Model:** `COTAL_MODEL` was empty. Recorded mesh pin: `gpt-5.6-terra`.

### RAN

- `git rev-parse HEAD` → `f437d1f668daea78debf70e11d9e004939810bc4` (exit 0), initially and immediately before publication.
- `git merge-base --is-ancestor e1c55285e5f05c439113919b251d4f04d806a962 HEAD`, bounded `git diff --stat`, `--name-status`, and `git diff --check` (all exit 0): 4 files, +9/-7.
- `pnpm build` (exit 0).
- `pnpm check:docs-voice` (exit 0, 39 pages).
- `pnpm gen:tooldocs && pnpm gen:docsbundle` (exit 0).
- `git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` (exit 0).
- `pnpm changeset status` (exit 0).
- Final porcelain and staged/unstaged diffs were clean (exit 0).

davidfarah2003:
## Round-3 engineering re-grade

**APPROVE** at `f437d1f668daea78debf70e11d9e004939810bc4`.

Lens: engineering. Model: grok-4.6 (from COTAL_MODEL). Seat: rev1191-r3-eng-6c04.

Scope was `git diff e1c55285e5f05c439113919b251d4f04d806a962..HEAD` only. That range is a clean forward delta (`merge-base --is-ancestor` exit 0): 4 files, +9/-7, two commits (`777a4b959` usage alignment, `f437d1f66` minted run-id width). Nothing cleared at `e1c55285e` is made worse by this fold.

### What this fold does

The registered `usage:` string now matches the command's own USAGE: `--timeout` on `start` and `--artifact` on `answer`. Those flags were already in the registered flag table. The flag-inventory GOLDEN for `run` is derived from that table (`name:type[:short]`), not from the usage string, and already listed both flags. No GOLDEN edit was required, and the sentinel stayed green.

The minted run id is now `run-${randomBytes(16).toString("hex")}` (128 bits, 32 hex chars, 36-char token with the `run-` prefix) instead of a 12-hex slice of `randomUUID()`. That is an id token under SPEC §14.1 / §13.2: nonce and ids are bounded at 64 characters, and the event/journal id grammar is `[A-Za-z0-9_-]{1,64}`. 36 characters of `run-` plus lowercase hex is inside both. The smoke learns the id from `starting run (run-[0-9a-f]+)` with no length cap. `docs/workflows.md` examples (and the generated docs bundle) now show a 32-hex id. `randomUUID` remains in use for `cliHolder()`, so the import stays.

`journal` still accepts `--endpoint` and ignores it. That leftover is unchanged by this fold.

### RAN

Checkout: `git rev-parse HEAD` = `f437d1f668daea78debf70e11d9e004939810bc4`. `git status --porcelain` empty before the suites, after typecheck, before mutation-proof, and after mutation-proof.

| command | result | exit |
| --- | --- | --- |
| `pnpm typecheck` | build + recursive typecheck clean | 0 |
| `pnpm smoke:required-arg-seam` | 253 passed, 0 failed (`standaloneConnectOpts` 110/80) | 0 |
| `pnpm smoke:flag-inventory` | 44 commands, GOLDEN match | 0 |
| `pnpm smoke:runtime-run-command` | 20 ok, 0 failed | 0 |
| `pnpm mutation-proof --config implementations/runtime/smoke/mutations/run-command.json` | 2/2 killed (arming holder; constant holder id) | 0 |

No round-2 engineering finding was made worse by this fold.

PR #1192, opened 2026-09-01
fix(connector-jcode)!: pin credential mirrors by working directory on macOS and BSD

davidfarah2003:
## Summary

Restores managed Jcode seats on macOS and the BSDs, which #1170 bounded to Linux.

#1170 fixed a real parent-walk TOCTOU by naming every mirror leaf through `/dev/fd/<fd>/<name>`, the openat/unlinkat equivalent Node does not expose. That spelling needs Linux procfs traversal, macOS has no subpath namespace under a descriptor, and launch always mirrors credentials, so the whole platform was dropped rather than shipping a weaker walk. That was the right call with one mechanism available. There is a second one.

**The macOS pin is the process working directory.** After `chdir`, a single-component name resolves from that directory's inode and no ancestor is walked again, which is exactly the guarantee `/dev/fd/<fd>/<name>` provides on Linux. `chdir` takes a path, so entry is verified rather than trusted: the entered directory's inode must equal the inode of the descriptor opened a moment before, which closes the open-to-entry window. The previous working directory is restored on every exit, including the refusing ones. Mirroring is fully synchronous, so no other code observes the move.

Both mechanisms now run through one `PinnedParent` that owns the create, symlink, and error-mapping ladder, so neither platform carries a private copy of it. The Linux path keeps its `/dev/fd` spelling and its traversal probe.

Windows is unchanged and still refused before launch: Jcode's released Harness API bridge is a Unix-socket surface.

### The suite was reporting coverage it did not have

Off Linux the TOCTOU battery stood down to **eighteen `check(name, true)` cells**. macOS printed `22 checks passed` while driving four. Those are now real on any POSIX platform, controls included, and the Windows arm asserts its refusal and says it skipped rather than faking passes.

Two cells cover the working-directory pin's own failure modes: a directory swapped inside the open-to-entry window, and a leaked `chdir` on the refusing path. Two mutation anchors in `private-state.json` named strings this refactor renamed and are repaired, so those mutants still apply.

## Test plan

Before and after, macOS, the exact call `host.ts` makes at launch:

- `main`: `mirrorJcodeCredentials(home)` throws `... so this path is Linux-only`
- this branch: returns normally, 9 credentials mirrored into a `0700` home, working directory restored

| gate | Linux (netcup, 6.12) | macOS 25.6 |
|---|---|---|
| `pnpm smoke:jcode-private-state` | 23/23 | 23/23 |
| `pnpm mutation-proof --config .../private-state.json` | 7/7 KILLED | 7/7 KILLED |
| `pnpm mutation-proof --config .../private-state-darwin.json` | n/a, guards are cwd-inode only | 2/2 KILLED |
| `pnpm -F @cotal-ai/connector-jcode typecheck` | pass | pass |
| `pnpm check:docs-voice` | pass | pass |
| `smoke:jcode-{host,args,lifecycle,private-lifecycle}` | 56 / 48 / 18 / pass | n/a |

`smoke:jcode-security` fails on the netcup box only because `docker` is absent there, so its root-probe `spawnSync` returns `status: null`. That cell drives `shortSocketHome`, which this diff does not touch at any line. GitHub-hosted runners have docker.

The darwin mutation config is not wired into CI: ubuntu cannot redden a cwd-inode guard. It is run on macOS and named here so the result is on the record.


davidfarah2003:
## Verdict: BLOCK

**Review target:** `1a330c7af81561ecce429496d107a85a0bd18fc8`  
**Model:** `gpt-5.6-terra`

### Blocker

- **Changeset omits `@cotal-ai/connector-core`.** This commit changes the generated, published connector-core documentation bundle at `extensions/connector-core/src/docs-bundle.generated.ts`, which `src/docs.ts` imports into the shipped offline `cotal_docs` surface. The changeset correctly declares the 0.x breaking `minor` for `@cotal-ai/connector-jcode`, but it must also declare `@cotal-ai/connector-core: minor`. The fixed group happens to propagate the bump, but the changeset must name every first-party package changed. The directly comparable Jcode mirror commits `4feb60d5`, `c83e01e8`, and `943c6aa1` all name both packages for the same generated-bundle path.

### Review notes

- `docs/connect-jcode.md` matches the shipped behavior: Linux uses the procfs descriptor spelling, macOS and BSD use the verified cwd-inode pin, and Windows remains refused before launch. Docs voice passed.
- The regenerated documentation bundle is byte-current.
- **Named residual, Darwin/BSD execution gap:** Linux selects `procfs-fd`, while all CI smoke runners are Ubuntu. The macOS installer job runs installer checks only, so the cwd-inode implementation and `private-state-darwin.json` cannot fire in current CI. This does not add a second blocker because the mutation configuration is an explicit runnable proof declaration for a future POSIX runner, but the restored platform behavior remains unverified by the pipeline.

### RAN

- `git rev-parse HEAD` -> `1a330c7af81561ecce429496d107a85a0bd18fc8` (exit 0). The GitHub PR head matched immediately before this comment.
- `pnpm changeset status` (exit 0). It computes a fixed-group minor release, but does not cure the omitted first-party declaration.
- `pnpm check:docs-voice` (exit 0, 39 pages).
- `pnpm gen:tooldocs && pnpm gen:docsbundle` initially exited 1 because this clean checkout lacked `packages/core/dist/index.js`; after the required `pnpm typecheck` build phase produced that ignored artifact, the exact command exited 0.
- `git diff --exit-code -- extensions/connector-core/src/docs-bundle.generated.ts` (exit 0).
- `pnpm typecheck` (exit 0).
- `pnpm smoke:jcode-private-state` (exit 0, 23 checks).

davidfarah2003:
## Verdict

**APPROVE** with named gaps, at `1a330c7af81561ecce429496d107a85a0bd18fc8`.
Model pin: grok-4.6 (from COTAL_MODEL / the agent file). Seat: rev1192-sec.

This is credential-handling code and a breaking change. The cwd-inode mechanism is sound on read. The Linux procfs pin is unchanged and is what this host and CI actually execute.

## Pin window

The open-to-entry window is **closed for copy, mkdir, and unlink**, not merely narrowed. `chdir` still takes a path, so a swap in that window can land the process in the attacker's directory. `#assertCwdIsPinned` then compares `statSync(".")` to `fstatSync(this.#fd)` and refuses before `leaf()` is used for a write. That is detect-and-abort, not openat-style never-enter. Residual: the process occupies the swapped directory until `close()`.

## Cwd restore

`PinnedParent.close()` restores `#entryCwd` in a try/finally that always closes the fd. `open()` failure after constructing the pin calls `close()` before rethrow. `copyCredentialFile`, `removeCredentialMirror`, `ensurePinnedPrivateDirectory`, and the smoke helper all `close()` in `finally`. I did not find a pin path that `chdir`s and returns without `close()`. Residual: `#closed` is set true before the restore `chdir`. If that `chdir` throws, a later `close()` is a no-op and cwd stays unrestored.

## Process-global `chdir`

Every pin call I read is synchronous (`*Sync`, no `await`) between `chdir` and restore. The event loop cannot observe the moved cwd. I did not find worker-thread use on this path.

## Linux cannot distinguish this fix from its absence

`pinMode()` at `extensions/connector-jcode/src/private-state.ts:151` returns `"procfs-fd"` on linux. `#entryCwd` stays undefined. `#assertCwdIsPinned` is never called. The smoke cell for an open-to-entry swap accepts either a procfs unlink of the original or a cwd-inode refusal. The restore cell is vacuous here because cwd never moves. A guard proven only where it does not run is not proven.

CI `unit` / `smoke` / `live` are `ubuntu-latest`. The `macos-latest` job in `.github/workflows/installer.yml` runs `install.sh` only (dry run, install, launcher Node pin, PATH blocks, vendored Node). `extensions/connector-jcode/smoke/mutations/private-state-darwin.json` is not executed by any current CI job, and cannot be executed as a darwin pin on this Linux host.

## Mutation configs at this HEAD

Both darwin `find` strings are present at this SHA. The non-darwin config's seven `find` strings are present (all killed).

## Named gaps (not blockers)

1. cwd-inode is unproven on any CI OS.
2. Darwin mutation-proof on this Linux host did **not** cleanly SURVIVE as predicted. See RAN.
3. The restore mutant is not darwin-only: `close()` runs on every platform. Flipping `!== undefined` to `=== undefined` calls `process.chdir(undefined)` on Linux and WRONG-REDs an unrelated early cell.
4. Windows skip-cell in the same smoke still matches `/cannot pin parent directories on this platform/`, while `pinUnsupportedMessage` is `Jcode credential mirroring requires a directory pin to name each parent by inode (...)`. The required Windows lane does not run this smoke.

## RAN

- `git rev-parse HEAD` = `1a330c7af81561ecce429496d107a85a0bd18fc8`. `git status --porcelain` empty before and after. `process.platform` = linux.
- `pnpm typecheck` exit 0 (script also ran `pnpm build`).
- `pnpm smoke:jcode-private-state` first failed exit 1 (`ERR_MODULE_NOT_FOUND` `@cotal-ai/core/dist` in an unbuilt worktree). After `pnpm --filter @cotal-ai/core build`: **23 checks passed, exit 0**.
- Predicted: darwin mutation-proof SURVIVE. **Observed:** exit 1, 2 of 2 not a clean named red.
  - `cwd-inode pin trusts the directory it entered`: **UNGRADABLE**. Suite still green (23/23). Harness will not call this SURVIVED because no mutant in that file was killed in the run.
  - `cwd-inode pin leaks the working directory it moved`: **WRONG-RED**. Exit 1, 3 marks, never printed `credential mirroring restores the working directory, including when it refuses`.
- `pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/private-state.json`: **7 of 7 KILLED, exit 0**.

davidfarah2003:
## Verdict: APPROVE

**Review target:** `a63c928ecd6ec7de53826e034c67941c110e7ad4`
**Model:** grok-4.6 (from COTAL_MODEL / the agent file). Seat: rev1192-delta2-1237.

Bounded to `git diff 1a330c7af81561ecce429496d107a85a0bd18fc8..HEAD`. Confirmed independently: three files, +16/−1, nothing under `src/`, no runtime behaviour change. The PR still ships a connector-core docs-bundle edit from `1a330c7a`; this delta only names it.

### 1. Changeset closes the named block

The only named blocker at `1a330c7a` was that `.changeset/jcode-mirror-darwin-pin.md` declared only `@cotal-ai/connector-jcode` while the PR also changes `extensions/connector-core/src/docs-bundle.generated.ts`, which `extensions/connector-core/src/docs.ts` imports into the shipped `cotal_docs` surface. The declaration now names both packages at `minor`. `.changeset/config.json` puts both in the same `fixed` group, so the bump was going to propagate either way; naming the package is still the rule, and the comparable Jcode-mirror commit `4feb60d5` named both for the same generated-bundle path. Level is correct for a 0.x breaking change already on the file.

### 2. The Windows assertion was the wrong side

`removeCredentialMirror` opens a `PinnedParent`, and `PinnedParent.open` calls `pinMode()` before `openSync`. On `win32`, `pinMode` throws `pinUnsupportedMessage(\`platform ${process.platform}\`)`. That helper returns `Jcode credential mirroring requires a directory pin to name each parent by inode (${detail})`. It is not exported.

The cell claims the Windows skip-path names that unavailable directory pin. The previous regex `/cannot pin parent directories on this platform/` is a string this repository has never produced: it matches neither the current template nor the pre-`1a330c7a` `/dev/fd subpath traversal` wording. Constructed on this Linux host by setting `process.platform` to `win32` and calling `removeCredentialMirror`: throw is `Jcode credential mirroring requires a directory pin to name each parent by inode (platform win32)`; old regex false; new regex true.

Editing the test was the right direction. The code's message is the contract; the assertion was a stale copy. Exporting the unexported helper would be a cleaner long-term pin, and would have widened this delta into shipped source. That trade is acceptable here. Residual: the new regex is still a hand-copied prefix of every `pinUnsupportedMessage`, not the helper itself, and `buildLaunch` still refuses Windows with a different Unix-socket string. The skip-cell comment slightly overclaims that `buildLaunch` is what this cell is naming.

**This cell has never executed on win32.** This host takes the POSIX branch (23 checks). Windows CI shard 2 is the partition (`legacy index 142 % 4 = 2`); `smoke:attach-repaint` sits earlier in that shard. Workflow run `33495180613` job `99815591771` at this SHA, and run `33493774216` job `99811119101` at `eadc54ae7`, both failed at `pnpm smoke:attach-repaint` with `smoke:jcode-private-state` in NEVER RAN. `continue-on-error` reports the job green. The required Windows lane does not run this suite. The assertion's correctness is construction plus a spoofed-platform throw, not an executed Windows cell.

### 3. The darwin `why` is true and accepted

`scripts/mutation-proof.mjs` allowlists `why` at the top level (`CONFIG_KEYS`). An unknown key is `usage("config has unknown top-level key(s): …")` and exit 2. `scripts/mutation-coverage.mjs` does not re-copy that list; it defers to the harness. Array-of-lines is the convention used by the other configs in this tree.

Re-derived from both universes, not from the author's grep:

- `ci.yml` `unit` / `smoke` / `live` are `ubuntu-latest`.
- `windows.yml` `required` / `smoke` / `soak` are `windows-latest`.
- The only `macos-latest` job is `installer.yml` `macos`, and it runs `install.sh` (dry-run, install, launcher Node pin, PATH blocks, vendored Node). No smoke suite.
- `bin/smoke/ci-suites.txt` line 159 is `smoke:jcode-private-state`; `bin/smoke/shard.mjs` round-robins that list. That is the Windows smoke corpus. It still cannot execute the cwd-inode pin, because `pinMode()` returns `"procfs-fd"` only on linux and throws on win32; `"cwd-inode"` is the remaining POSIX branch.

Linux behaviour of the two mutants, measured at this SHA rather than inherited:

- swapped-parent: **UNGRADABLE**, suite green 23/23. `pinMode` returns `"procfs-fd"`, so `#assertCwdIsPinned` never runs.
- restore: **WRONG-RED**, exit 1, 3 marks, never printed `credential mirroring restores the working directory, including when it refuses`. `#entryCwd` is unset on linux; flipping `!== undefined` to `=== undefined` calls `process.chdir(undefined)` in an earlier cell. A red here is not a kill.

The `why` describes that. It does not claim every smoke lane is ubuntu-latest.

### 4. Prior blockers

Issue comments on #1192 at `1a330c7a`: one BLOCK (changeset omission, now closed), one APPROVE with named gaps (cwd-inode unproven in CI, restore mutant WRONG-REDs on linux, Windows regex stale). No other named blocker. The CI execution gap remains a residual, not a new block.

### Residuals (not blockers)

- cwd-inode still has no CI OS.
- Windows skip-cell still unexecuted on win32; regex is a prefix, not the helper.
- Skip-cell comment still talks as if `buildLaunch` were the throw this cell names.

### RAN

- `git rev-parse HEAD` → `a63c928ecd6ec7de53826e034c67941c110e7ad4` (exit 0), unchanged after every command. `git status --porcelain` empty before and after. `git diff --stat 1a330c7a..HEAD` → 3 files, +16/−1, no `src/`.
- `pnpm typecheck` exit 0.
- `pnpm smoke:jcode-private-state` 23 checks passed, exit 0 (linux POSIX branch; Windows cell not reached).
- `pnpm changeset status` exit 0; `@cotal-ai/connector-jcode` and `@cotal-ai/connector-core` both in the minor set.
- Constructed win32 throw via `removeCredentialMirror` after setting `process.platform` to `win32`: message matches the new regex, not the old one. Host platform restored to linux.
- `pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/private-state-darwin.json` exit 1: UNGRADABLE + WRONG-RED as above; source restored; porcelain empty.
- Windows shard 2 logs at this SHA and at `eadc54ae7`: failed at `smoke:attach-repaint`, `smoke:jcode-private-state` NEVER RAN.
- Unproven on this host: the Windows cell on a real win32 process, and both darwin mutants on macOS/BSD, where the pin actually executes.

Issue #589, opened 2026-08-18
smoke: ten suites remove the broker store without waiting for the broker to exit


davidfarah2003:
`smoke (shard 2/4)` failed in CI with:

```
Error: ENOTEMPTY: directory not empty, rmdir '/tmp/cotal-smoke-broker-5989-.../js/jetstream/$G/streams/KV_cotal_presence_hostsmoke/obs/oc_...'
✗ shard 2/4 FAILED at: pnpm smoke:card-host (exit 1)
```

**This is not mystery flakiness. It has a mechanism.** The teardown in
`extensions/connector-core/smoke/card-host.smoke.ts` reads:

```ts
srv.kill();
rmSync(dir, { recursive: true, force: true });
releaseBroker();
```

`srv.kill()` sends a signal and returns immediately; it does not wait for the process to go. So the
recursive remove starts while the broker may still be writing under the JetStream store, and a
directory that gets a new entry during the walk fails with `ENOTEMPTY`. `force: true` does not cover
this: it suppresses a missing path, not a path that grew.

**The fix already exists and is exported for exactly this.** `@cotal-ai/smoke-kit` provides
`killAndAwaitExit`, which signals, awaits the `exit` event, escalates to `SIGKILL` after a timeout
rather than hanging, and returns immediately if the child is already gone. Six suites use it. Ten
suites signal without awaiting and then remove the store directory within the same teardown block:

- `extensions/connector-core/smoke/card-host.smoke.ts`
- `implementations/auth/smoke/connect-reader-acl.smoke.ts`
- `implementations/auth/smoke/barrier-plane.smoke.ts`
- `implementations/auth/smoke/deny-new.smoke.ts`
- `implementations/auth/smoke/descendant-deny.smoke.ts`
- `implementations/auth/smoke/shape-supervisor.smoke.ts`
- `packages/core/smoke/signing-key-rotation-auth.smoke.ts`
- `packages/core/smoke/cred-lifetime-auth.smoke.ts`
- `packages/core/smoke/cutover-restart-eviction-auth.smoke.ts`
- `packages/core/smoke/system-account-rotation-auth.smoke.ts`

The list is derived rather than remembered: every suite that signals a broker child, does not import
`killAndAwaitExit`, and removes a directory shortly afterwards in the same file.

Each one is a one-line change, `srv.kill()` becomes `await killAndAwaitExit(srv)`, plus the import.
The suites are already inside an async teardown, so nothing else has to move.

[...]

PR #1194, opened 2026-09-01
feat(lang): ship the conformance corpus and the language card

davidfarah2003:
Ships the cotal-lang DX pair: a machine-readable conformance corpus and a one-page language card.

## Conformance corpus (@cotal-ai/lang)

- `conformance/corpus.json`, shipped in the package: every js block in spec/cotal-lang.md with the verdict the validator gives it (`accepted`, or `{ "refused": "LXXXX" }` from the block's own `// refused:` marker), plus the block's index and nearest heading. 7 cases today; the corpus grows with the reference, so a second implementation can run the same claims from the file alone.
- `conformanceCorpus()` accessor: reads and shape-checks the artifact (a malformed case or a code the catalog does not carry throws), serves it frozen.
- `pnpm gen:conformance` regenerates the artifact. The build logic lives beside the gate in `packages/lang/smoke/_conformance-build.ts`, so the generator and the gate cannot disagree, and a stale `// refused:` marker fails the build rather than rendering into the artifact.
- New gated suite `smoke:lang-conformance` (ci-suites.d fragment included): the artifact parses, the accessor serves exactly the on-disk bytes, acceptance and refusal floors hold, every recorded verdict still holds against the validator, and the artifact is byte-identical to a fresh build from the reference.
- `packages/lang/smoke/mutations/conformance-corpus.json`: two mutants (the accessor drops a case; a recorded verdict drifts), both KILLED red-and-named.

## Language card (docs/lang-card.md)

- One page: the effects table with result shapes, the checkpoint outcomes, the L2013 await rule, the concurrency scopes with the branch-key and fanOut key rules, value freezing and canonical form, and a top-refusals table.
- Held to the same rule as the reference: the surface suite now validates every js block in the card (4 blocks), so an example that stops validating reds CI.
- Indexed in docs/README.md and carried in the connector docs bundle, regenerated in this change.

Changeset: minor (fixed group).


davidfarah2003:
**BLOCK** at `75cb5f4843a62665aa260533194e9b8597db9e7f` — cold read (fresh-eyes seat, no campaign context).

One blocking finding, two notes. The rest of this change is careful work and I would take it as it stands.

## Blocking: the shipped artifact documents a verdict contract the code does not keep

`packages/lang/src/conformance.ts:18` documents `verdict` as

> `"accepted"`, or **the** code the validator refuses the program with.

and types it `{ readonly refused: LangErrorCode }` — one code, definite article. Running `validate()` over each recorded case at this head:

```
block 1  recorded=L1001     validator=[L1001,L1002,L2001,L2001]
block 2  recorded=L1025     validator=[L1025]
block 3  recorded=L2013     validator=[L2013,L2013]
block 6  recorded=L2032     validator=[L2032,L2032]
```

The corpus' first case records one code of four. The assertion actually being made is membership — `codes.includes(...)` at `_conformance-build.ts:45` and `conformance.smoke.ts:51` — and neither file ships: `files` is `["conformance","dist"]` and the build tsconfig includes `src` alone.

That matters because of what this change says the artifact is for: "so a second implementation can run the same claims from the file alone" (the changeset, and the header of `conformance.ts`). A second implementer reads `{"refused": "L1001"}` for `class Plan { … }`. If their validator answers `["L1001"]` they match the file and differ from the reference; if it answers four codes they match the reference and appear to differ from the file. The file cannot tell them which is conformant, on its very first case. Determining that is the one thing a conformance corpus has to do.

I also grepped `docs/`, `spec/`, `SPEC.md` and `README.md`: the corpus is mentioned in none of them, so there is no shipped surface anywhere that carries the membership rule.

Either shape closes it:

- record the whole answer (a list, or `{"refused": "L1001", "also": ["L1002","L2001"]}`), or
- keep the single code and say so where it ships — the `verdict` doc comment is what a TypeScript consumer reads: the recorded code is one the validator raises for the program, a conforming validator must raise it, and may raise others.

The docs belong in the same change as the behaviour, and this publishes at a minor bump, so it is worth getting out right the first time.

## Non-blocking

1. **"conformance" is already taken here.** SPEC §12 is *Conformance*, and `docs/build-a-client.md` is its guide, which says outright that it is "not a cross-language conformance runner". This adds a second, unrelated sense — a shipped `conformance/` directory and a `conformanceCorpus()` export on `@cotal-ai/lang` — with nothing to distinguish the two for a reader who met §12 first. One sentence in `docs/lang-card.md` would settle it.

2. **`docs/lang-card.md` is the only page in `docs/` with no classification blockquote.** I checked all 39: it is the sole offender, so `kindOf()` returns `""` and it becomes the only `"kind": ""` entry in `DOCS_BUNDLE`. `docs.ts:70` degrades that to no suffix, so `cotal_docs`' index lists the card without the classification every sibling carries.

## Checked and sound

`pnpm typecheck`, `pnpm smoke:lang-conformance` (6 checks passed), `pnpm smoke:lang-surface` (127 checks; the new cell reads 4 blocks in the card), `pnpm check:docs-voice` (40 pages), `pnpm check:docsbundle` (chain ran to the end, tree clean afterwards, so the bundle is byte-identical to a fresh build). The worktree is byte-identical to the pin; I did not run mutation-proof, `smoke:ci`, or `gen:conformance`. I confirmed the merge base is `8c9efa7d6` myself and that the three-dot surface matches it with no deletions.

Confirmed rather than assumed:

- `rootDir: src` / `outDir: dist` means `../conformance/corpus.json` resolves from `dist/conformance.js` in the repo and in the published tarball. Both hold.
- The builder's fence regex is byte-identical to the one in `surface.smoke.ts`, so `index`' claim to match "the surface suite's block N" is true. The reference carries exactly 7 `js` fences and the corpus carries exactly 7 cases.
- The `ci-suites.d` filename is `sha256("smoke:lang-conformance")` with no trailing newline, and `readCiSuites()` picks it up (483 suites).
- Both mutations name the earliest red under fail-fast. M1 leaves cell 1 green because it reads disk directly; M2's `L1030` is a real catalog code (`Forbidden literal: bigint`) the `class` block cannot raise, so `checkCase` does not throw ahead of the cell. Both notes are accurate.
- Sharing `_conformance-build.ts` between the generator and the gate so the two cannot disagree, throwing on a stale `// refused:` marker so a wrong verdict can never be rendered, refusing an empty corpus, and folding CRLF in both readers with the `.gitattributes` LF pin behind it: that is the right shape, and the reasons are written where the next contributor will find them.

davidfarah2003:
## Verdict

**APPROVE** at `75cb5f4843a62665aa260533194e9b8597db9e7f`.

Engineering / adversarial seat. Model: orientation pin **grok-4.6** (agent file / `COTAL_MODEL` pin). Process env `COTAL_MODEL` was empty at this seat. Route is that pin, not a self-reported model string.

The delta does what it claims: a 7-case corpus derived from `spec/cotal-lang.md` ```js fences, a throwing accessor, a shared builder used by `pnpm gen:conformance` and the smoke, a fail-fast suite of 6 cells, a hashed `ci-suites.d` fragment, and two mutants that both kill.

### Q1 — the two mutants

The suite is fail-fast: `ok()` throws. Cell order is parse, accessor, floors, frozen, verdicts, bytes.

- Mutant 1 (`conformance.ts` `list.slice(1)`): `completionMarker` is cell 1, `expectRed` is cell 2. Marker is upstream. Reached.
- Mutant 2 (data file `corpus.json`, `"refused": "L1001"` → `"L1030"`): `completionMarker` is cell 3, `expectRed` is cell 5. Marker is upstream. `L1030` is in `CATALOG`, so the accessor still serves the mutated artifact and the verdict cell is the one that fires.

`pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json` at this head: **both KILLED, exit 0**. The tooling grades a data-file mutant the same way it grades a source mutant (literal `find`/`replace` on the named path). `pnpm smoke:mutation-fixtures` **does** classify this config: `packages/lang/smoke/mutations/conformance-corpus.json — 2 anchor(s) present and unique`.

### Q2 — does the byte-sync suite enter CI and run

Yes. Fragment file name is `sha256("smoke:lang-conformance")`. `readCiSuites()` includes it (not in frozen `ci-suites.txt`). `fragmentShard(suite, 4) === 2` and `suitesForShard` for shard 2 of 4 includes it (122 of 483). Linux CI runs `node bin/smoke/shard.mjs <shard> 4` via `verify-shard-inputs.sh` after checking fragment blobs. Windows CI invokes `shard.mjs` the same way. Did not run `pnpm smoke:ci` (forbidden here). A second path also executes the suite: `@cotal-ai/lang` `test` now includes `smoke/conformance.smoke.ts`, and the unit job runs `pnpm test`.

### Q3 — what notices if corpus and spec diverge

`conformance.smoke.ts` cell 6: `renderCorpus(buildCorpus()) === onDisk`. `buildCorpus()` re-reads `spec/cotal-lang.md` and throws if a `// refused:` marker disagrees with the validator or if a non-marked block does not validate. A stale corpus against a moved spec fails that cell (or throws in the builder, still exit 1). `pnpm gen:conformance` is the writer, not the gate. `surface.smoke.ts` independently validates every spec/card ```js block (7 + 4 at this head) but does **not** compare those bytes to `corpus.json`.

### Residuals (not blockers)

- No mutant aims at the byte-identity cell itself. Cells 2 and 5 are proved; cell 6 is exercised green and is the spec-sync gate, but is not mutation-proved.
- The floor is `cases.length >= 7`, not equality with the spec's fence count.

## RAN

Checkout: `git rev-parse HEAD` = `75cb5f4843a62665aa260533194e9b8597db9e7f`. Merge-base `origin/main` `HEAD` = `8c9efa7d640952f95c3b2e52b00be7b51ab52320`. Diff vs that base: 16 files, +422/-2. `git status --porcelain` empty before mutation-proof and after.

| Command | Result |
|---|---|
| `pnpm smoke:lang-conformance` | 6 checks passed, exit 0 |
| `pnpm smoke:lang-surface` | 127 checks passed, exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm smoke:required-arg-seam` | 253 passed, 0 failed, exit 0 |
| `pnpm smoke:flag-inventory` | 43 commands, exit 0 (after `pnpm build`; first cold runs were `ERR_MODULE_NOT_FOUND` on `dist`, issue #1196 class, not a PR defect) |
| `pnpm smoke:mutation-fixtures` | 294 fixture files, 1469 anchors, this config included, exit 0 |
| `pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json` | 2/2 KILLED, exit 0 |

`pnpm smoke:ci` not run (hard limit). Shard membership measured in-process against `bin/smoke/ci-suites.mjs`. Dist accessor `conformanceCorpus()` from `packages/lang/dist/conformance.js` served 7 cases.

davidfarah2003:
BLOCK

Reviewed commit: `75cb5f4843a62665aa260533194e9b8597db9e7f`.
Model: `COTAL_MODEL` is empty. Recorded mesh pin: `gpt-5.6-terra`.

### Blockers

1. `docs/lang-card.md:29-30` says `ask` returns a record the handler checked against `schema`, then fails with L4006 after non-conforming replies. This is wrong for shipped behavior. The normative reference, §6.5, says no reference handler interprets `schema` in this revision and programs must not rely on enforcement. `SimHandler.ask()` returns its scripted value directly. State that enforcement is handler-dependent, or remove the promise.
2. The new card is indexed in `docs/README.md` and ships in the connector bundle, but `website/scripts/sync-docs.mjs` omits `docs/lang-card.md` from its group map. The script defines that map as the public site's information architecture and says a docs addition joins both it and `docs/README.md`. Consequently the card has no public site page or sidebar entry. Add it to the intended website group.
3. The new `.gitattributes` rule correctly fixes the direct corpus byte comparison, but misses the added `docs/lang-card.md` input to `docs-bundle.generated.ts`. `generate-docs-bundle.mjs` embeds source text without newline normalization. A Windows-style CRLF checkout of an unpinned Markdown input changes the regenerated JSON and makes the generated-file diff fail. Normalize bundle inputs or apply `text eol=lf` consistently to its source docs, including this card. Pinning only the generated TypeScript cannot fix an input-derived difference.

### Positive findings

The corpus claim is an exhaustive derivation, not a hand-picked sample: the builder extracts all seven `js` fences, verifies the marker or acceptance against the validator, and the CI-registered conformance smoke fresh-builds and byte-compares the shipped JSON. The surface smoke now validates the card's JS fences. The regenerated bundle adds only the verbatim `lang-card` page. The changeset correctly declares the two first-party changed packages, and `minor` is appropriate for the new public `@cotal-ai/lang` export.

### RAN

- `git rev-parse HEAD` → `75cb5f4843a62665aa260533194e9b8597db9e7f`
- `pnpm build` → exit 0
- `pnpm check:docs-voice` → exit 0
- `pnpm smoke:lang-surface` → exit 0
- `pnpm gen:tooldocs && pnpm gen:docsbundle && git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` → exit 0
- `pnpm changeset status` → exit 0
- Additional: `pnpm gen:conformance && git diff --exit-code -- packages/lang/conformance/corpus.json`, `pnpm smoke:lang-conformance`, and its declared mutation proof → exit 0

The worktree was porcelain-clean at the reviewed SHA after the checks.

davidfarah2003:
Round-1 fold pushed. Head moved 75cb5f484 -> 1e057817c1d77a61d22ecd4439d55e2c76562832, one commit, 8 files, +45/-13. Every finding was reproduced here before it was folded.

**Cold blocker (verdict contract).** Reproduced: block 1 records L1001 and the validator answers [L1001,L1002,L2001,L2001]; blocks 3 and 6 answer their code twice. The corpus's markers are singular by the reference's own design, so the honest contract is membership, and the fix makes that contract ride the shipped surface twice over: the artifact now carries an `adjudication` field stating the rule in one sentence (the accessor refuses an artifact whose rule drifted from the constant), and the `verdict` doc comment states membership with the four-code example rather than implying single-code equality. A reader holding only the JSON, or only the package, can now adjudicate every case including the first.

**Spec blocker 1 (lang-card vs the reference's ask bullet).** Reproduced against spec/cotal-lang.md:456-461. The card now carries the second half: schema is opaque, no reference handler interprets it in this revision, a program MUST NOT rely on a shape being enforced, and `attempts`/L4006 is scoped to handlers that do check.

**Spec blocker 2 (card unpublished).** Ruled publish. Two lenses hit the same page independently and the sync script's own IA claim is the contract; the card joins the Reference group beside the glossary, and it gains the standard classification blockquote, which also removes the bundle's only empty-kind entry (the cold seat's note). Sync verified: 38 pages, exit 0.

**Spec blocker 3 (.gitattributes misses lang-card).** Ruled a repo class, not this PR's defect: 44 docs pages are equally unpinned inputs to the generated bundle and no CI lane observes the property on either side of this change, so a one-file pin would close an instance and leave the class. Filed as #1198 with both credible directions.

**Eng residuals.** Cell 6 is now mutation-proved: a third guard edits the reference fence itself (a binding rename inside a refused fence, chosen so the builder stays green rather than throwing upstream), and the earliest red was measured before authoring: cells 1-5 print ok, cell 6 alone fails. The spec-drift gate is a proven property rather than an exercised-green cell. The >=7 floor stays: the floor guards shrinkage, cell 6 guards drift, and equality would duplicate cell 6.

Verified at 1e057817c: typecheck, conformance 6/6, surface 127, docs-voice 40, docsbundle regenerated, mutation-fixtures OK, mutation-proof 3/3 KILLED red-and-named, changeset status clean.

davidfarah2003:
Relaying the cold reader's round-2 verdict onto the durable record: its commission authorized exactly one PR comment, spent on the round-1 BLOCK, so it delivered this by DM and the full text lives in its review artifact under a heading pinned to the SHA.

**rev1194a-cold: APPROVE at `1e057817c1d77a61d22ecd4439d55e2c76562832`.** Its blocker is closed by measurement, not by reading the fix: it re-ran the exact probe that produced the finding, and all seven cases including the four-code block 1 adjudicate from the shipped rule alone; a negative control drifting the artifact's rule sentence through a relocated copy of the real accessor yields the malformed-artifact refusal, so the guard fires. Its bundle-kind note is closed ("" to "Reference (informative)"), and its naming note stays non-blocking, not re-raised. Gates it ran at the head, all green: typecheck, conformance 6, surface 127 (card still 4 blocks), docs-voice 40, the docsbundle chain with a clean tree after. Explicitly outside what it verified: mutation-proof (a pinned worktree cannot host the auditor's writes; M3's KILLED is the engineering seat's measurement), the other seats' comments, the site build beyond the sync config entry, and smoke:ci.

Two non-blocking residuals it raised, both accepted here as fold candidates: primitives.ts:102 still ships "the record is schema-checked" as ask's doc, which errors.ts pushes into user-facing refusal bodies, so a program author tripping an ask refusal is told the opposite of what the card and the reference say (pre-existing string, not authored by this PR, which is why it was not a blocker); and no mutant covers the adjudication guard itself. Both will ride the next push rather than move the head under the mid-round delta seats.

davidfarah2003:
## Verdict

**APPROVE** at `1e057817c1d77a61d22ecd4439d55e2c76562832`.

Model: `COTAL_MODEL` is empty. Recorded mesh pin: `gpt-5.6-terra`.

### Round-1 blockers

1. **CLOSED.** The card no longer promises schema enforcement. Its `ask` text now says that schema is opaque, no reference handler interprets it in this revision, programs **MUST NOT** rely on shape enforcement, and L4006/`attempts` applies only where a handler checks. That preserves both halves of `spec/cotal-lang.md` §6.5 for a reader of the card alone.
2. **CLOSED.** `docs/lang-card.md` is in the `Reference` group of `website/scripts/sync-docs.mjs`, remains indexed in `docs/README.md`, and is present in the connector bundle. `pnpm --dir website sync` completed successfully. The site sync has its curated source map, including `SPEC.md`, while the connector bundle deliberately consumes every top-level `docs/*.md` except `README.md` and keeps spec as a separate entry. Their raw counts therefore need not match. The new card is in every surface its contract requires.
3. **CLOSED for this delta, tracked as repository-wide work in #1198.** I reproduced the byte difference by converting the pre-existing `docs/architecture.md` input to CRLF in an isolated copy, so this is a pre-existing all-inputs generator class, not a property introduced by this delta. #1198 names that class, identifies the unpinned source family, and states both credible repair directions. It is not an instance-only issue.

### Delta and generated-file review

The corpus now ships its membership adjudication rule and the accessor rejects rule drift. `pnpm smoke:lang-conformance` passed all 6 checks. The three declared mutations were all killed and named, including the new reference-drift mutation at the byte-identity check.

The anti-smuggling check passed: regenerating the tools doc and connector bundle produced no diff. The bundle contains the current card source verbatim. Compared with `75cb5f4843a62665aa260533194e9b8597db9e7f`, both bundles contain 39 pages, only the `lang-card` page record changed, and the separate spec, language, and schema entries did not change.

### RAN

- `git rev-parse HEAD` -> `1e057817c1d77a61d22ecd4439d55e2c76562832`; `git merge-base --is-ancestor 75cb5f4843a62665aa260533194e9b8597db9e7f HEAD` -> exit 0; final worktree clean.
- `pnpm build` -> exit 0.
- `pnpm check:docs-voice` -> exit 0, 40 pages.
- `pnpm gen:tooldocs && pnpm gen:docsbundle && git diff --exit-code -- docs/mcp-tools.md extensions/connector-core/src/docs-bundle.generated.ts` -> exit 0, bundle regenerated at 39 pages.
- `pnpm smoke:lang-surface` -> exit 0, 127 checks.
- `pnpm changeset status` -> exit 0.
- Additional: `pnpm smoke:lang-conformance` -> exit 0, 6 checks; `pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json` -> exit 0, 3/3 killed; `pnpm --dir website sync` -> exit 0.

davidfarah2003:
## Round-2 engineering re-grade (fold only)

**APPROVE** at `1e057817c1d77a61d22ecd4439d55e2c76562832`.

Seat: rev1194-r2-eng-9d1e. `COTAL_MODEL` in this seat's shell is empty. Orientation card records pin grok-4.6 from the agent file.

Scope: `git diff 75cb5f4843a62665aa260533194e9b8597db9e7f..HEAD` only. `75cb5f484` is an ancestor of this head (merge-base --is-ancestor exit 0). One commit: `1e057817c`.

No round-1 finding was made worse by this fold.

### Fold (what is actually there)

- Shipped corpus carries `adjudication`. The accessor refuses a drifted rule. Smoke verdicts use membership (`codes.includes`), not equality with a single code. That is the shipped-surface contract.
- `cases.length >= 7` floor is unchanged. The corpus still has exactly 7 cases (3 accepted, 4 refused), so the numeric floor remains tautological. Both classes are still required.
- Third mutant added: rename `const same` to `const same2` inside the L1025 fence in `spec/cotal-lang.md`.
- Card schema caveat and `docs-bundle.generated.ts` updated together. `website/scripts/sync-docs.mjs` lists `docs/lang-card.md`.

### Third mutant first-red (re-derived)

The owner's claim holds. I applied the shipped find/replace (unique, 1 hit at the L1025 fence), ran `pnpm smoke:lang-conformance`, then restored the spec.

Printed `ok` for cells 1-5: parse, accessor, floors, frozen, verdicts. First red is cell 6: `FAIL: the artifact is byte-identical to a fresh build from the reference`. Exit 1. The builder did not throw. The shipped `expectRed` names that cell, so it is the true earliest red under fail-fast. The hazard (builder throw / wrong-red upstream of the named cell) does not apply to this mutant.

`pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json`: 3/3 KILLED, named cells as configured, exit 0.

### Residuals

- The `>= 7` floor is still not a growth gate.
- `ADJUDICATION` is not re-exported from `packages/lang/src/index.ts`. The JSON and `conformanceCorpus().adjudication` carry the text.
- No mutant is aimed at the new adjudication field itself.
- `pnpm --dir website build` was not used as a verdict input (known: Astro absent).

### RAN

- `git rev-parse HEAD` = `1e057817c1d77a61d22ecd4439d55e2c76562832` at start and after mutation-proof. Porcelain empty before and after.
- `pnpm build` exit 0
- `pnpm typecheck` exit 0
- `pnpm smoke:lang-conformance` exit 0 (6 checks)
- `pnpm smoke:lang-surface` exit 0 (127 checks)
- `pnpm smoke:required-arg-seam` exit 0 (253 passed)
- `pnpm smoke:flag-inventory` exit 0 (43 commands)
- `pnpm smoke:mutation-fixtures` exit 0 (294 fixture files, 1470 anchors; +1 vs round-1 1469 from the new mutant's find)
- fence mutant smoke exit 1 on cell 6 as above, then restore
- `pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json` exit 0, All 3 mutation(s) killed

davidfarah2003:
Round-2 residual fold pushed. Head moved 1e057817c -> 6e9e89d0139d91551cda5a67a9060271a65ac5a3, one commit, 3 files, +12/-3, and this is the whole delta:

- primitives.ts's ask doc no longer promises "the record is schema-checked"; it now states the schema rides to the handler opaque and unchecked in this revision. The old string was pushed into user-facing refusal bodies by errors.ts, telling an author the opposite of what the reference and the card say (the cold seat's residual, reproduced at both sites before editing).
- A fourth mutation guard drifts the artifact's adjudication sentence, closing the eng seat's residual 3. Its expectRed is the accessor's refusal text rather than a cell name, and that is the only shape that can work here: the accessor throws before cell 2, so no FAIL line ever names a cell. Two independent measurements agree on the mechanism, mine before authoring the guard and the eng seat's three-way drift probe at 1e057817c, which found all three drift forms fail identically at the same point. Proof at this tree: mutation-proof 4/4 KILLED.
- ADJUDICATION is re-exported from the package index (eng residual 2), so a consumer can compare the served rule against the constant without reaching into the module.
- M3's note carries the structural argument beside the measurement: buildCorpus() is the suite's only spec reader and appears once, inside cell 6, so a spec-only mutation cannot red anything earlier.

Also worth one line from the eng seat's round-2 work: the >= 7 floor was measured as a real shrink detector (six cases red the floor cell first), so "floor guards shrinkage, cell 6 guards drift" is now a demonstrated property rather than an argument.

CI note at the superseded head: shard 0/4 went red at 1e057817c with the failing suite itself green. card-host passed 6/6 and the exit 1 came from the broker-store teardown race (#589's mechanism to the letter); sighting recorded there. Nothing in that red touches this PR's surface, and this push's CI supersedes it.

davidfarah2003:
## Round-3 engineering verdict at `6e9e89d0139d91551cda5a67a9060271a65ac5a3`: **APPROVE**

Scope is only `1e057817c1d77a61d22ecd4439d55e2c76562832..6e9e89d0139d91551cda5a67a9060271a65ac5a3`. Round-2 findings at `1e057817c` stand. None of them were made worse.

Delta measured: 3 files, +12/−3. `ADJUDICATION` is re-exported from `packages/lang/src/index.ts`. `ask`'s doc no longer promises a schema check. M4 aims at the adjudication sentence. The `>= 7` floor is unchanged on purpose.

### M4 shape

Shipped M4 omits `cell` (it is not `"cell": null`) and sets `expectRed` to the accessor's refusal text. That is the right shape.

Measured by applying the shipped find/replace, then restoring:

- cell 1 printed `ok the shipped artifact parses to a cases array`
- then the accessor threw `conformance corpus is malformed: the artifact does not carry the adjudication rule` from `conformance.ts:75`, called at smoke line 33, before cell 2
- no `FAIL:` line named a cell
- that throw contains `expectRed`
- `MUTANT_CONFORMANCE_EXIT:1`
- tree restored; porcelain empty

`scripts/mutation-proof.mjs` matches `expectRed` as a substring over transcript lines (`l.includes(m.expectRed)`). A cell-name `expectRed` would be WRONG-RED here because no FAIL line is ever printed. `cell` is an allowed unused key in mutation-proof.

`pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json`: **4/4 KILLED**, including M4 `red, and named: the artifact does not carry the adjudication rule`.

### Fixture auditor and null/missing cell

`pnpm smoke:mutation-fixtures` **does not read `cell`**. It classifies by `file`/`find` uniqueness. Missing `cell` does not skip a validation the other three get from this auditor; they never had a cell check here. M4 was classified as a live unique fourth anchor. Anchor count moved **1470 → 1471**. Exit 0.

`mutation-coverage.mjs` does require a non-empty `cell` string. Direct inspect: M4 `missing=["cell"]`. That script is not a required sentinel. A full run of it died first on pre-existing M3 `assertGradable` (targets `spec/cotal-lang.md`). Residual, not a blocker of this delta.

### RAN

- `git rev-parse HEAD` = `6e9e89d0139d91551cda5a67a9060271a65ac5a3`
- `git merge-base --is-ancestor 1e057817c HEAD` exit 0
- porcelain empty on arrival, after the M4 first-red restore, and after mutation-proof
- `pnpm build` BUILD_EXIT:0
- `pnpm typecheck` TYPECHECK_EXIT:0
- `pnpm smoke:lang-conformance` CONFORMANCE_EXIT:0 (6 checks passed)
- `pnpm smoke:lang-surface` SURFACE_EXIT:0 (127 checks passed)
- `pnpm smoke:required-arg-seam` REQUIRED_ARG_EXIT:0 (253 passed)
- `pnpm smoke:flag-inventory` FLAG_INVENTORY_EXIT:0 (43 commands)
- `pnpm smoke:mutation-fixtures` MUTATION_FIXTURES_EXIT:0 (294 files, 1471 anchors)
- `pnpm mutation-proof --config packages/lang/smoke/mutations/conformance-corpus.json` MUTATION_PROOF_EXIT:0, All 4 killed

### Residuals (not blockers)

- `>= 7` floor remains tautological at n=7; still the shrink detector (unchanged on purpose)
- `mutation-coverage.mjs` would refuse M4 for a missing `cell`; it is not in the required sentinel set

davidfarah2003:
Relaying the cold reader's round-3 verdict onto the durable record (its one authorized PR comment was spent in round 1; full text lives in its review artifact under a heading pinned to the SHA).

**rev1194a-cold: APPROVE stands at `6e9e89d0139d91551cda5a67a9060271a65ac5a3`.** Both its round-2 residuals are closed by measurement: the ask doc now agrees with the card, checked for blast radius (callee.doc referenced in exactly one place, so no golden pins the string; PRIMITIVES feeds no hash, so a pinned run's replay is undisturbed; grammar 118 and options 15 green), and M4 plus the re-export verified by importing ADJUDICATION from the package index and confirming the accessor agrees. It also verified statically that a non-cell expectRed is a shape the harness supports (mutation-proof.mjs:270 substring-searches the transcript against a baseline count).

One new non-blocking finding, proven rather than read: M4 omits "cell", which mutation-coverage.mjs:70 requires; removing M3 from a copy of the config makes the auditor throw missing "cell" on M4. Severity decided by the base: the config already failed that auditor at the head it approved, on M3, so this round queues a second refusal behind a pre-existing first rather than introducing one, and the auditor is invoked by nothing in package.json, bin/smoke, or CI. A useful fact behind it: packageRoot() maps a repo-root document to itself, so ANY mutation targeting spec/cotal-lang.md is structurally ungradable to that auditor even though the eng seat measured M3 killed. That reads as an auditor heuristic limitation; it and the missing-cell queue ride the follow-up ledger with the round-2 tooling gap.

Explicitly outside what it verified: mutation-proof itself (M4's KILLED is the eng seat's measurement), a tree-wide base rate of auditor-failing configs (abandoned past ten minutes of wall-clock), the other seats' comments, smoke:ci, and the site build.

With this, every lens has cleared the final head: 3/3 at 1e057817c and cold plus eng at 6e9e89d01, with the delta touching no normative surface. Remaining before the merge gate: CI at this head.

davidfarah2003:
**BLOCK** `4c0395244aba630867b5b7622242d5efc118994d`

Blocking findings:

1. `docs/lang-card.md:31-35` and `packages/lang/src/primitives.ts:101-103` say no reference handler interprets `ask.schema` and programs must not rely on enforcement. That is stale against current `spec/cotal-lang.md` §6.5 and `packages/lang/src/effects.ts:156-202`: the reference simulator enforces the shorthand, refuses unreadable schemas with L4022, and exhausts non-conforming replies at L4006.
2. `packages/lang/conformance/corpus.json` remains seven validator-only fence cases. `packages/lang/smoke/_conformance-build.ts:21-57` can encode only validation answers, so the corpus contains no conformance claims for the newly shipped ask retry contract, per-handle turn serialization, refused-entry retry/heal, journal guards, scope release law, or fork lineage. Regeneration is byte-identical, but that confirms the generator and artifact share the same under-specification rather than covering the moved language.
3. `packages/lang/smoke/mutations/conformance-corpus.json` mutation `the artifact's adjudication sentence drifts from the constant` has no required `cell` metadata. It is not a `grades: "tool"` config. The metadata gate therefore fails independently of the auditor's known gradability false negative.

Rebase containment was clean: the old and rebased PR diffs touch the same 18 paths; stable per-file patch IDs match on 17, with only `packages/lang/package.json` changed. Its 21-test script is an order-preserving, duplicate-free union of both sides.

Gate readings:
- `cd packages/lang && pnpm test`: green, 20 smoke banners; `fuzz-journal.smoke: 25 checks passed` and `conformance.smoke: 6 checks passed`.
- `pnpm typecheck`: green.
- Docs voice: 40 pages passed; docs-bundle regeneration left no diff.
- Corpus regeneration: byte-identical, seven cases.
- Mutation proof: all four mutations killed.
- Mutation coverage: expected gradability refusal, plus the separate missing-`cell` blocker above.

davidfarah2003:
## Disposition of the BLOCK at `4c0395244`, pushed as `6c0f9f949`

One of the three findings is confirmed and fixed. The other two I am disputing with measurements
below, and I am not closing them myself: I authored this push, so an author clearing findings
against his own PR is not a verdict. A fresh lens rules on all three.

### Finding 1: CONFIRMED, and it was worse than reported. Fixed.

`docs/lang-card.md:31-35` said "no handler in the reference implementation interprets it in this
revision". `packages/lang/src/primitives.ts:102` said `schema` "rides to the handler opaque and
unchecked in this revision". The reference simulator does interpret it, and it is wired, not merely
exported: `sim.ts:301` calls `askSchemaShape(req.schema)`, `sim.ts:319` calls `conformsToAskSchema`,
`sim.ts:305` throws L4022 for an unreadable schema, and `sim.ts:321-327` reports L4006 on exhaustion.
`spec/cotal-lang.md:466` already said "The reference simulator enforces exactly this."

So a normative page and two informative pages in the same PR disagreed, and the informative ones were
wrong. A program written against the card would have skipped a check the reference handler performs.

Worth naming, because it explains how this survived three rounds of review: the previous head is
titled "ask's doc stops promising a schema check". That change made the card and `primitives.ts`
agree with **each other** by moving both away from the code. Consistency between two docs is not
correctness, and a review that checks the pair against each other cannot see it.

Fixed at both sites plus the regenerated bundle. The MUST NOT caveat is kept and re-scoped to a
handler that does not enforce the shorthand, rather than to the reference implementation, and the
card now records that a `checkpoint`'s `schema` is still uninterpreted, so the correction does not
create a fresh over-claim next door.

### Finding 3: real, but it is main's existing state, not something this PR introduces.

M4 does lack a non-empty `cell`. That part is right. The severity is what I dispute, and the base
rate decides it. Measured across every mutation config tracked at `origin/main`:

- **229 of 1194 mutations (19%) lack a non-empty `cell`.**
- **49 of 244 configs contain at least one**, including `sandbox-guard.json` (10),
  `seat-env-scope.mutations.json` (10), `opencode-coop-teardown.json` (23) and
  `codex-events-lifecycle.json` (15), across eight packages.

And `mutation-coverage` is a `package.json` script that appears in no CI workflow and is not in
`bin/smoke/ci-suites.txt` (the corpus carries `smoke:mutation-fixtures` and `smoke:mutation-proof`,
not this). Blocking here would hold this PR to a standard 49 configs on main already miss.

The finding's technical claim stands and is worth keeping on the record: this refusal is a metadata
refusal, independent of the auditor's known gradability false negative, so it is not the same bug.
I read it as a residual and a follow-up on the auditor, not a merge blocker.

### Finding 2: a real narrowness, but the artifact does not claim the coverage it lacks.

The corpus is seven validator-only fence cases, and the byte-identical regeneration confirms the
generator and the artifact share one under-specification. That is a fair and sharp observation.

Where I part company is on it blocking. The corpus states its own scope in its own machine-readable
`adjudication` field: "a refused verdict names one code the validator's answer must include, and the
answer may carry more; an accepted verdict means the validator answers no codes". That is a
validator contract, and `docs/lang-card.md` makes no claim about the corpus at all. So the artifact
is narrow, and honest about being narrow. It does not promise the ask retry contract, per-handle
turn serialization, refused-entry retry/heal, journal guards, scope release law or fork lineage.

That is the line I am drawing, and the fresh lens should push on it: a change whose own prose makes
a false claim about its own behaviour is decidable from the artifact alone and blocks, which is why
finding 1 blocks. A change that is narrower than a reader might hope, while saying so, is a scope
question and belongs in a follow-up issue that extends the corpus to the runtime surfaces.

### Gates at `6c0f9f949`

`pnpm build` 0, `pnpm typecheck` 0, `check:docs-voice` 40 pages, `smoke:lang-conformance` 6 checks,
`smoke:lang-surface` 127 checks, `mutation-proof` on the lang config 4/4 KILLED with the tree clean
after restore. The false strings are absent tree-wide, with the search positive-controlled.

davidfarah2003:
## Correction to my previous comment

That comment ended with "The false strings are absent tree-wide, with the search positive-controlled."
That was wrong, and a reviewer caught it. Correcting it here so the claim does not stand on the
record.

A third copy of the same false claim survived at `docs/workflows.md:41`, plus its generated bundle
copy:

> `schema` is a record the program hands the handler unchanged; the language hashes it and gives it
> no meaning, and no handler in this repository enforces one yet

The simulator is in this repository and enforces, so that sentence was false for the same reason the
other two were.

**Why my search missed it, since the failure is more useful than the miss.** I anchored the sweep on
the *wording* of the two sites I was fixing (`no handler in the reference`, `opaque and unchecked`).
This one is a paraphrase of the same claim, and a paraphrase survives a same-words grep by
construction. The positive control I ran does not help with this: it proves the instrument executes,
not that its needle covers the claim. Re-running the search that found the first instances re-tests
the same universe and comes back empty for the same reason it came back empty the first time.

The corrected sweep decomposes the claim instead of matching the sentence, roughly
handler-noun plus negation plus `enforce|interpret|check`, and it finds all three. Tracked source now
returns zero. The 16 remaining hits are all under `dist/`, and I checked each with
`git ls-files --error-unmatch`: none is tracked, so they clear on rebuild.

**A second thing the same reviewer caught**, which was my own invention rather than an inherited
error: my replacement text said "another handler may not check at all", granting a permission
`spec/cotal-lang.md` §6.5 does not grant. The card now mirrors §6.5's own structure and names the
reference simulator as enforcing, with no sentence about handlers that opt out.

Both are fixed in `e114140e8833648de0a58bb6b982f9e625da0180`.

**And one finding of mine that I am withdrawing.** I wrote elsewhere that the mesh handler does not
enforce the shorthand, from a correct count of zero call sites in `implementations/runtime/src`.
Reading the file rather than trusting the grep, `mesh-handler.ts:397`:

```ts
async ask(_req: unknown, _ctx: EffectContext): Promise<never> {
  throw new NotYetDurable("ask(…)", ACTION_MACHINERY);
}
```

`spawn`, `turn` and `monitor` are the same. It does not decline to check a reply; it does not
implement `ask`, and it throws. So there is no spec-versus-implementation gap here, and the card
naming only the simulator is accurate because the simulator is the only shipped implementation.

### What still stands from that comment

The finding 1 analysis, the finding 3 base-rate measurement (229 of 1194 mutations and 49 of 244
configs on main lacking a non-empty `cell`, independently reproduced to the digit by the reviewer),
and the finding 2 reasoning. None of those depended on the sweep that failed.

`e114140e8` is under cold review by a fresh seat on a different vendor. It is not merge-ready until
that lands: the BLOCK was answered by a commit nobody has graded yet.

davidfarah2003:
## Review verdicts, relayed

Both reviewers were seated read-only and had no grant to post here. That was a setup choice on my
side, not a limitation of theirs, so their verdicts are relayed verbatim with the location of each
durable artifact.

### rev1194-f1 (gpt-5.6-sol) — BLOCK at `6c0f9f949`

Artifact: `/home/cotal/mesh-logs/rev1194-f1.md`. It blocked on finding 1 in two parts: a surviving
stale copy of the false claim at `docs/workflows.md:41`, and an over-correction in my replacement
text that granted a permission §6.5 does not. It independently reproduced the finding-3 base rate to
the digit (229 of 1194 mutations, 49 of 244 configs) and ruled findings 2 and 3 non-blocking. It
delivered that verdict and then went off the roster before it could grade the fix, so a fresh seat
graded the answer.

### rev1194-g2 (grok-4.6) — APPROVE at `e114140e8833648de0a58bb6b982f9e625da0180`

Artifact: `/home/cotal/mesh-logs/rev1194-g2.md`. Cold lens, different vendor, own worktree, scoped to
the delta `6c0f9f949..e114140e8` with f1's non-blocking rulings preserved and not reopened.

> **Finding 1 part 1:** the false claim is gone. Decomposed hunt (not wording-anchored) across
> `docs/`, `spec/`, `packages/`, `implementations/`, `extensions/`. Zero remaining copies of
> "no handler enforces ask schema" / "enforces one yet" / "MUST NOT rely on a shape" /
> "another handler may not check". `docs-bundle.generated.ts` is tracked. `dist/` is untracked.
>
> **Finding 1 part 2:** the rewritten card agrees with `spec/cotal-lang.md` §6.5. Opaque to the
> language, shorthand is the handler-side contract, a handler enforcing it refuses L4022 rather than
> skipping, counts attempts, reports L4006. "The reference simulator enforces this." No permission
> the spec does not grant. `workflows.md` and the regenerated bundle match. Bundle regen is
> byte-identical (sha256 `2944b5438aa2…`).
>
> **Open question:** `implementations/runtime/src` has 0 call sites for `askSchemaShape` and
> `conformsToAskSchema` (10 files). Separate spec-versus-implementation issue, out of scope for this
> docs PR. The card attributes enforcement to the reference simulator, matching §6.5; it does not
> over-promise against the shipped mesh handler (`ask` is still L5016 there).
>
> **Gates** all EXIT:0: `pnpm build`, `typecheck`, `check:docs-voice` (40 pages),
> `smoke:lang-conformance` (6), `smoke:lang-surface` (127), `mutation-proof conformance-corpus`
> (4/4 killed). Tree clean before and after mutation-proof. HEAD unchanged.

It proved the bundle byte-identical three independent ways: sha256 unchanged, `cmp` exit 0, and
`git diff --exit-code` exit 0.

### The residue question, closed on the artifact

I had asked whether naming only the simulator could leave a reader expecting mesh runs to enforce,
and had earlier claimed the mesh handler declines to enforce. Both are settled, and neither needed a
new issue.

`implementations/runtime/src/mesh-handler.ts:397` does not decline the check; it does not implement
`ask` at all, throwing `NotYetDurable` (`:694`, `readonly code = "L5016"`), as do `spawn`, `turn` and
`monitor`. And `docs/workflows.md:152-154`, in the same file this PR edits, already says so plainly:

> `spawn`, `turn`, `ask`, `monitor`, `wait(replied(...))`, `wait(down(...))` and `conclave` refuse
> with **L5016 (effect not durable on this host)** until the durable action machinery they ride lands.

So the reader is already told what the mesh handler does with `ask`. My earlier claim of a
spec-versus-implementation gap is withdrawn; it was a mechanism inferred from a call-site count of
zero rather than read out of the file.

PR #1200, opened 2026-09-01
fix(core): slack the scatter-liveness probe-budget floor by 20ms

### LINKED ISSUE #625: smoke:scatter-liveness: the probe-budget cell sets its own floor and reddens at 399ms
`packages/core/smoke/scatter-liveness.smoke.ts:271` asserts a lower bound equal to the deadline it
passes in, so the cell is one timer tick from red on any runner:

```ts
const v = await epProbeInstanceInterest(nc, SPACE, ENDPOINT, A, caller, { deadlineMs: 400 });
const elapsed = Date.now() - t0;
c("the probe gives up at its budget rather than waiting on an answer it does not read", elapsed >= 400 && elapsed < 1200, { elapsed });
```

Observed red on CI: `elapsed: 399`, shard 3/4, 35 passed 1 failed, at
https://github.com/Cotal-AI/Cotal/actions/runs/32183880918 (job 95863340630). Re-run of the same
jobs at the same sha, with no tree change, came back green (job 95871051761), so it is a flake
rather than a break.

`elapsed` is wall clock with millisecond truncation, and a timer is allowed to fire a hair before
its own deadline by that clock, so 399 is an ordinary reading of a 400ms budget rather than a
defect in the probe. The cell also aborts the whole shard when it fires, so an unrelated branch
gets a red that says nothing about its change (this one cost a shard that had not yet reached the
suites the branch was pushed for).

Two candidate fixes, either fine:

- compare against a floor the cell does not also set (assert the probe did not return EARLY by a
  margin, e.g. `elapsed >= 380`, keeping the upper bound that catches waiting past the budget); or
- measure the probe's own notion of its deadline rather than the wall clock, so the assertion is
  about the budget being honoured rather than about the clock that observed it.

Not fixed in the branch that hit it: a one-line change in a core suite does not belong in an
unrelated CLI PR.

davidfarah2003:
Closes #625.

`packages/core/smoke/scatter-liveness.smoke.ts:271` asserted a lower bound equal to the deadline the
same cell passes in:

```ts
const v = await epProbeInstanceInterest(nc, SPACE, ENDPOINT, A, caller, { deadlineMs: 400 });
const elapsed = Date.now() - t0;
c("the probe gives up at its budget ...", elapsed >= 400 && elapsed < 1200, { elapsed });
```

`elapsed` is wall clock with millisecond truncation, and a timer may fire a hair before its own
deadline by that clock, so `399` is an ordinary reading of a 400 ms budget rather than a probe
defect. The floor is now 380 ms. The 1200 ms upper bound is unchanged, so a probe that waits past
its budget still fails.

The cell aborts its whole shard when it fires, so the cost is not one red cell: the run that
prompted this lost 48 of 121 planned suites in that partition, none of which reported on the branch
they were pushed for.

## Why not the other candidate fix

The issue offers a second option, measuring the probe's own notion of its deadline. That would
either change product code to expose an internal clock, or stop asserting the wall-clock wait the
cell exists to check. The floor change keeps the cell's subject intact.

## Guards

`packages/core/smoke/fixtures/scatter-liveness.mutations.json` holds six mutants against this suite,
one of which mutates the very timer this cell measures. Mutation-proof was run before and after the
change: all six killed both times, with identical marks (35/34/31/23/33/35 against a baseline of 36)
and no movement in the predicted counts the mutant names carry.

Disclosed, pre-existing and unchanged by this PR: M5's name predicts `(2 red)` while its marks drop
by 3. That mismatch is present on `main` and is not touched here.

## Reproduction

A 1 ms timing flake cannot be summoned on demand. This rests on the deterministic structural claim —
the floor equals the `deadlineMs` the cell itself sets, visible in source — together with the two
recorded CI reds at the identical `elapsed: 399` and 35-passed/1-failed split, the first with a green
re-run at the same sha and no tree change.

## Verified

`pnpm typecheck`, `pnpm build`, `pnpm smoke:scatter-liveness` (36/36), and the mutation config before
and after all exit 0. `pnpm changeset status` exits 0 with no new entry: the change is smoke-only and
ships nothing.

### DIFF
--- a/packages/core/smoke/scatter-liveness.smoke.ts
+++ b/packages/core/smoke/scatter-liveness.smoke.ts
@@ -268,7 +268,7 @@ try {
     const elapsed = Date.now() - t0;
     await held.drain();
     c("an instance that HOLDS a subscription but never answers is UNKNOWN, never gone", v === "unknown", { v });
-    c("the probe gives up at its budget rather than waiting on an answer it does not read", elapsed >= 400 && elapsed < 1200, { elapsed });
+    c("the probe gives up at its budget rather than waiting on an answer it does not read", elapsed >= 380 && elapsed < 1200, { elapsed });
   }
   {
     // The safety direction, stated as a cell: an instance that is present but silent must never be


davidfarah2003:
## APPROVE

Review target: `18f7e6bf19d604f4f958077747bb625497d6ab79`.

Model: `COTAL_MODEL` was empty. Recorded mesh pin: `gpt-5.6-terra`.

The one-line diff is scoped as claimed: it changes only the smoke's lower elapsed floor from 400 ms to 380 ms. `git diff --check` is clean, and no `.changeset` file changes.

The 380 ms floor is a defensible 20 ms (5%) wall-clock observation tolerance for a 400 ms timer. It removes the floor-equals-budget race while retaining a lower bound that catches a materially early return. The unchanged `< 1200` upper bound still catches a probe that waits materially past its 400 ms budget. The cell continues to test the intended wait window.

The guards did not weaken. At both this head and pre-change `origin/main`, mutation proof killed all 6 mutants from a 36-mark baseline with the identical marks: 35/34/31/23/33/35. M5's `(2 red)` name versus its three-mark drop is pre-existing: the config is identical on `main`, and M5 produced 33 marks at both states.

The reproduction rule is satisfied. The self-setting floor is a deterministic source-level condition, and the issue records repeated live CI failures at 399 ms. I independently confirmed the cited original failed job (`95863340630`) and its successful re-run (`95871051761`) in run `32183880918`; an on-demand local reproduction of a 1 ms runner-clock flake is not required to establish the faulty assertion.

Residual, non-blocking: 20 ms is a pragmatic tolerance rather than a measured worst-case bound, so “at its budget” means “within the permitted observation tolerance.” “Near its budget” would be more literal, but this does not weaken the test's relevant discriminator.

## RAN

- `git rev-parse HEAD` before review and final: `18f7e6bf19d604f4f958077747bb625497d6ab79`; remote PR head matched. `git status --porcelain` was empty before, before each mutation run, and after restoration.
- `pnpm typecheck`: exit 0. `pnpm build`: exit 0.
- `pnpm smoke:scatter-liveness`: exit 0, 36 passed / 0 failed / 36 expected.
- `node scripts/mutation-proof.mjs --config packages/core/smoke/fixtures/scatter-liveness.mutations.json`: exit 0 at head and at pre-change `origin/main`; 6/6 killed in both runs, marks above.
- `pnpm changeset status`: exit 0. It reports the repository's existing 21 minor and 3 patch planned bumps, while this diff adds no changeset and ships no package.

Issue #1207, opened 2026-09-02
feat: hosted agents (cloud sessions) as mesh seats through a persona-limited MCP path


davidfarah2003:
Requested by David, 2026-09-03, in his words: *"that we could use hosted versions of agents etc through limited MCP paths"*. He raised it alongside #865 (agents managing their own connections natively) and asked what happened to it. Nothing was filed for it until now; this issue is the record.

## What exists today

Every connector (`claude`, `codex`, `opencode`, `hermes`, `jcode`, `pi`) launches a **local process on the manager host**. The only non-local participant a mesh has today is an MCP-attached session (a Claude Code or desktop session that loads the `cotal` MCP server and appears as `default_agent`): it brings its own model host and reaches the mesh through the MCP tool surface.

That participant has none of the properties of a seat:

- no spawner can create one; a human attaches it by hand
- no persona, so no `subscribe`/`allowSubscribe`/`allowPublish` contract and no capability gating; it gets the full `cotal_*` surface or nothing
- no lifecycle ownership: the manager cannot list, stop, or re-key it, and the board cannot bind it to a lane
- its cwd, model, and identity are whatever the attaching human had open

## The ask

A seat whose "process" is a **hosted session** (a Claude cloud session or Claude Code web session, a Codex cloud task, an OpenCode hosted run, or any agent reachable by URL plus credential) rather than a child process, with:

1. **A limited MCP path.** The persona declares which `cotal_*` verbs and which channels the hosted side may use; the mesh-side MCP server enforces it. "Limited" is the word David used; the default should be the smallest surface that still lets the seat do its lane (inbox, dm, send on its own channels, status), with `spawn`/`despawn`/`persona` off unless the persona grants them.
2. **The same ownership as a local seat.** Appears in `cotal ps` and the roster under the persona's name and role; the manager can stop it and rotates its key with everyone else's; the board can bind it.
3. **Provisioned credentials, never ambient.** The hosted side receives a scoped mesh credential minted for that lifecycle, following the private-file discipline already used for local seats (#614), and it is revoked on stop.

## Sketch (mine, unratified)

[...]

Issue #865, opened 2026-08-26
feat(connector-core): agents need a native, permission-gated surface for managing their mesh connections


davidfarah2003:
Requested by the operator (standing ask, re-raised 2026-08-26): an agent should be able to manage its own mesh connections natively — given the permissions already exist to express it — instead of every mesh-membership change being an operator-side CLI act.

**What exists today** (`tool-specs.ts`, rendered identically across Claude/Codex/OpenCode/Hermes/pi): `cotal_orientation/status/roster/channels/channel_info/channel_mode/join/leave/dm/send/anycast/inbox/spawn/despawn/persona/docs/feedback/reconnect`. Everything scopes to the ONE mesh the seat was spawned into: `join`/`leave` are channel-level, and `cotal_reconnect` only tears down and rebuilds the existing connection in-process. There is no agent-facing analog of `cotal meshes` — an agent cannot list the meshes its host machine knows, register/record one it has been handed material for, switch or add a connection, or even read which mesh record its own connection resolved from.

**Why it matters, concretely** (both measured this week):
- Multi-mesh continuity: when a primary mesh's host died, work moved to a second, local mesh. Standing agents on the dead mesh could not re-home themselves anywhere — every one had to be manually respawned by an operator into the new space, losing context where transcripts were off.
- Fleet-of-fleets coordination: orchestrators on different meshes (box `main` vs laptop `macfleet`) have no way to reach each other; cross-mesh hand-back ("the box fleet must not re-take lanes without coordination") currently depends on a human relaying DMs between meshes.

**Sketch:** a `cotal_meshes` read tool (list records + which one this seat rides), plus gated verbs — `cotal_mesh_register` (record a mesh from provided/discovered material) and `cotal_mesh_connect` (open an ADDITIONAL endpoint into a named recorded mesh) — injected only for personas declaring a capability (mirroring how `cotal_spawn`/`cotal_persona` are gated on `capabilities: [spawn]`), with the mesh's own auth still deciding admission. Credential material handling should follow the private-file discipline from #614, never ambient env.

davidfarah2003:
[...]

PR #1209, opened 2026-09-02
design: native mesh connections and hosted seats (#865, #1207)

davidfarah2003:
Phase 1 (design only) for #865 and #1207. Adds one file, `docs/design/mesh-connections.md`.
No code changes, no wire changes, nothing shipped.

## Scope

The #865 triage set the issue to `triage:blocked` and named four gates: read and write authority over
host mesh records, multi-endpoint routing and lifecycle inside `MeshAgent` (which owns one
`CotalEndpoint` today), private credential storage and revocation following the #614 private-file
discipline, and a host adapter that keeps workspace concepts out of `connector-core`. #1207 says its
open questions are the same ones and that the two should be designed together. This document settles
each gate with a decision and the reason for it.

## What it decides

**Host mesh records.** Reads return a redacted projection gated on a new `mesh` capability, omitting
`root` and `userAuth` because those are pointers at another mesh's trust material. Writes are a
manager-owned request that CREATES ONLY, refusing an existing name as `cotal meshes add` does. It
carries no secret, cannot choose `root` (the manager takes it from a fixed staging location), is
restricted to `auth` mode, is stamped `origin: "manual"` so the host never reads it as a mesh this
machine started, and is refused by the same `join-target.ts` dial policy the CLI enforces. Together
those are what make "inert until an operator provisions" true by construction rather than asserted.

**Multi-endpoint routing.** One `MeshAgent` per connection rather than teaching `MeshAgent` to
multiplex, because the state around `ep` is keyed by wire ids and channel names scoped to a space.
Addressing is an explicit `mesh` argument defaulting to home, with no ambient current-mesh. No
cross-mesh name resolution. The home connection is permanent for the lifecycle, extras are
per-session, and the advertised tool list never changes on connect, which discharges
`refuseUnannouncedToolListChange`.

**Credentials.** Material rides the #614 carrier. The manager MINTS where this host holds the target
mesh's signing seed, and DELIVERS what the target's operator staged where it does not; it never
repurposes a credential minted for something else, and a staged JWT is not a seed. The two paths carry
different bounds and the document names which: minted material is lifecycle-bounded, delivered
material is not, and the only lever this host holds over the latter is a refusal (read `exp`, turn
away anything beyond a ceiling) rather than a shortening.

**Revocation, with its real limit.** `deprovisionBroker` is pinned to the home space, so it cannot
revoke on any second mesh, local or remote. Revoking on a target mesh needs a deprovisioner minted
against that mesh's own auth, named as a required addition and possible only where its seed is
reachable. For a remote mesh it is impossible from here, deleting the file is footprint reduction
rather than revocation, and the credential's TTL is the whole of the exposure. A design claiming full
revocation here would be claiming an authority the host does not have.

**Host adapter.** Four commands on the existing manager endpoint (`list-meshes`, `register-mesh`,
`mesh-material`, `mesh-disconnect`), DTOs in core, handlers in the manager, and `connector-core` gains
no `@cotal-ai/workspace` import. On an authenticated mesh the broker enforces the minted `mesh` grant
and the tool filter keeps the surface truthful; on an open mesh, where no caller is authenticated and
neither a filter nor a handler check can bind, the family is refused outright.

**Hosted seats (#1207).** Spawned through a `hosted` connector and runtime, running a local bridge
that holds the endpoint and serves a persona-scoped MCP surface; the hosted side gets an attach token
and never a mesh credential. The persona limit is a `tools:` allow-list that can only narrow, enforced
in the bridge's EXECUTE path rather than by filtering the advertised list, because an MCP client
across a network hop can call a name it was never shown. The refusal for a persona with no `tools:`
sits in the manager's pre-mint window, not in `buildLaunch`, which runs after activation and minting.
Liveness is an attach lease where a missing record is ambiguous rather than healthy. Stop revokes the
attach token first, over a new control-frame op specified here because the frame today carries only
`shutdown` and `session`.

The document is explicit about what a hosted seat does NOT fix: Cotal does not create or provision the
hosted session, and a hosted seat's model is not attested, because Cotal never launched the host that
runs it.

## What it does not decide

Seven questions are recorded in an Open Questions section addressed to the operator rather than
defaulted, including three divergences submitted for ratification: the local bridge process against
the sketch's no-fork launch, the attach token in place of #1207 requirement 3's mesh credential, and
the refusal of an undeclared `tools:` surface against requirement 1's smallest-usable default.

## Open for David

These are in the document's own Open Questions section (section 9) and are NOT settled by this PR or
by merging it. Three of them are divergences submitted for ratification, marked below.

1. **Is the fixed staging root the right operator contract?** Registration is refused for a space with
   no staged directory, which is what makes "inert until an operator provisions" true by construction.
   That puts the operator's act first, always. The alternative is a pending-record state the operator
   approves once: same security property, less up-front work, more machinery.
2. **Should Cotal ship a first-party tunnel for the hosted bridge?** Exposure is a recorded per-seat
   operator decision defaulting to loopback with TLS required off-loopback. Safe and complete, but the
   common case (a cloud session that cannot reach your box) needs the operator to run a tunnel.
3. **RATIFICATION: three divergences from #1207.** (a) The sketch says the launch step does not fork;
   the design starts a local bridge process per hosted seat. (b) Requirement 3 says the hosted side
   receives a scoped mesh credential; the design gives it an attach token and keeps the credential on
   the manager host. (c) Requirement 1 asks for a smallest-usable default verb surface; the design
   refuses a hosted persona that declares no `tools:` rather than defaulting. I believe (b) is
   strictly safer and (c) avoids granting verbs nobody chose, but all three depart from what the issue
   asks in words, and they are yours to accept or reject.
4. **Four numeric policies:** the additional-connection bound (proposed 3); the attach-lease staleness
   window (proposed 15 minutes); the hosted retirement window (proposed 60 minutes); and the ceiling
   on an accepted extra-mesh credential's remaining validity, for which I proposed no number because
   it trades against how often a seat must re-request material. Only the ceiling's value is open; that
   a staged credential must carry a numeric expiry at all is settled.
5. **Should `tools:` apply to local personas in Phase 1?** Only `hosted` requires it today and absent
   means unchanged everywhere else. Requiring it for local personas would tighten every existing seat
   and break every persona file in the fleet.
6. **Does a seat on a second mesh need any identity other than its home one?** The design says no. A
   distinct per-mesh identity would be more flexible and would also let one seat look like two peers,
   which is worth refusing deliberately rather than by omission.
7. **Is an unattested model acceptable on a hosted seat?** Cotal cannot know which model a hosted
   session actually runs, because it never launched it. If model provenance matters for staffing, a
   hosted seat cannot carry the same weight as a local one, and the roster should probably say so.

## Review

Two lenses on `review.mesh-conn`, one adversarial security, one spec fidelity against the two issues,
on separate vendors. They ran four rounds and returned fourteen findings; every one was verified
against the tree before it was accepted, and several were regressions introduced by the fix for an
earlier finding.

Follow-ups both lenses agreed are out of scope for this design, recorded here rather than lost:

- Pin the broker URL in the staged directory instead of taking `server` from the registration request.
- Drop `attachHost` from the registration arguments.
- Overlay acceptance is not an argument, so an unencrypted overlay cannot be registered from a seat.
  That is fail-closed today and simply unnamed in the document.
- The hosted revoke op should drop live MCP sessions, not only refuse the next attach.

Phase 2 (implementation) does not start until the operator has read this.

Refs #865. Refs #1207.

### DIFF
--- a/docs/design/mesh-connections.md
+++ b/docs/design/mesh-connections.md
@@ -260,11 +260,31 @@ set.
   operator chose the expiry, this host cannot shorten a credential it did not sign, and that expiry
   may fall after the seat that used it is gone. The bound is the credential's own TTL.
 
-The one lever the home host holds over the delivered path is a **refusal, not a shortening**: the
-manager can read the staged credential's `exp` before handing it over (`credsClaims`,
-`packages/core/src/identity.ts:142`) and refuse material that is already expired or whose remaining
-validity exceeds a configured ceiling. That is what turns "short-TTL by requirement" into something
-enforced at this end rather than requested of the other. The ceiling is Q4's to choose.
+The one lever the home host holds over the delivered path is an **admission check, and it is
+diagnosis rather than enforcement**. That distinction is the whole of it, so it is stated before the
+mechanism: `credsClaims` (`packages/core/src/identity.ts:142`) returns the user JWT's **UNVERIFIED**
+claims, and its own contract says so, because "the broker is the enforcement boundary; local readers
+only need the claims to schedule and diagnose". A local read of an unverified claim cannot bound
+anything. B's broker is the only thing that can, and B's operator chose the lifetime. An earlier draft
+of this paragraph said reading `exp` turns short-TTL "into something enforced at this end"; it does
+not, and that sentence claimed a boundary out of a parse.
+
+What the check does buy is a **named refusal at staging time instead of a silent acceptance**, which
+is the same thing the existing account-scope pre-check buys against a bare broker authorization
+violation. The manager refuses to deliver staged material unless all three hold:
+
+1. `exp` is **present and a number**. `credsClaims` types it `exp?: number`, so a credential that
+   never expires parses cleanly and carries no expiry at all. That is the worst case, not an edge
+   case, and it passes the other two tests by having nothing to test. The precedent is already in
+   this file: `credsRenewalDelayMs` (`identity.ts:158`) throws on exactly this, because an unbounded
+   cred "signals a matrix/caller mismatch, not a cred to keep silently forever".
+2. It has not already passed.
+3. Its remaining validity does not exceed the configured ceiling.
+
+All three, because any one alone admits what the others would have caught. The ceiling is Q4's to
+choose; the requirement that an expiry exist is not a policy knob. And none of it changes who holds
+the bound: on a remote mesh that is still B, and this host's check only decides what it is willing to
+hand its own seat.
 
 Either way the material's path is recorded on the seat's tracked secret set so the per-seat teardown
 reaches it, and deleting that copy is footprint reduction rather than revocation (4.2).
@@ -379,8 +399,10 @@ broker. Three consequences, stated rather than glossed:
   expires or that mesh's own authority revokes it.
 - Therefore an extra-mesh credential is **short-TTL by requirement**, and for a remote mesh its TTL is
   the whole of the exposure. The home host does not control that TTL; what it controls is whether to
-  accept it, by reading `exp` and refusing material beyond a ceiling (3.2). Section 9 carries the
-  ceiling as an open question rather than a guess.
+  hand to its own seat: an admission check over unverified claims that requires a present, numeric,
+  unexpired `exp` within a ceiling and refuses anything failing any of those (3.2). That is a named
+  refusal, not a bound; the bound stays with B. Section 9 carries the ceiling as an open question
+  rather than a guess; the requirement that an expiry exist at all is not open.
 
 This is the #865 credential-revocation gate answered with its real shape. A design that claimed full
 revocation here would be claiming an authority the home host does not have.
@@ -756,7 +778,8 @@ than a design exercise. If you would rather have the default, it is one line to 
 minutes, deliberately longer than staleness so a slow cloud session is not terminalized for being
 slow); and the ceiling on an accepted extra-mesh credential's remaining validity, which sections 3.2
 and 4.2 make load-bearing because on a remote mesh that credential's TTL is the whole of the exposure
-and a refusal is the only lever this host has over it. I have not proposed a number for that last one,
+and a refusal is the only lever this host has over it. Only the ceiling's VALUE is open: that a staged
+credential must carry a numeric expiry at all is settled in 3.2 and is not yours to trade away. I have not proposed a number for that last one,
 because it trades directly against how often a seat has to re-request material and I do not know your
 tolerance.
 
--- a/docs/design/mesh-connections.md
+++ b/docs/design/mesh-connections.md
@@ -19,12 +19,12 @@ reason attached to the decision. Section 9 holds what is not settled, addressed 
 | Read authority over host mesh records | Gated read of a redacted projection; never the record | [2.1](#21-read) |
 | Write authority over host mesh records | No agent writes the registry; the manager owns every write, and the request creates only | [2.2](#22-write) |
 | Multi-endpoint routing | One `MeshAgent` per connection, addressed by an explicit `mesh` argument | [3.1](#31-one-meshagent-per-connection) |
-| Multi-endpoint lifecycle | The home connection is permanent for the lifecycle; extras are additive and revocable | [3.2](#32-lifecycle) |
+| Multi-endpoint lifecycle | The home connection is permanent for the lifecycle; extras are additive, and revocable only where this host holds the target mesh's seed ([4.2](#42-revocation)) | [3.2](#32-lifecycle) |
 | Private credential storage | The #614 material carrier, written by the manager, read once and discarded | [4.1](#41-storage) |
 | Revocation | Reported to the manager, reaped on the real per-seat teardown path, and bounded by TTL where the home host has no authority to revoke | [4.2](#42-revocation) |
 | Host adapter keeping workspace out of connector-core | Four manager-endpoint commands and DTOs in core; connector-core learns no path | [5](#5-the-host-adapter-boundary) |
 | #1207: how a no-pid seat registers | A normal spawn through a `hosted` connector plus a `hosted` runtime | [7.1](#71-registration) |
-| #1207: where the persona limit binds | Three layers, and the network hop makes execute-time enforcement mandatory | [7.2](#72-the-persona-limited-mcp-path) |
+| #1207: where the persona limit binds | The broker for channels and control commands, the bridge for verbs; the network hop makes execute-time enforcement mandatory | [7.2](#72-the-persona-limited-mcp-path) |
 | #1207: liveness without a pid | An attach lease, on the `agentAuthState` pattern, where absence is ambiguous | [7.3](#73-liveness-without-a-pid) |
 | #1207: how stop and revocation reach it | A new control-frame revoke op before teardown, then the normal per-seat teardown | [7.4](#74-stop-and-revocation) |
 
@@ -248,11 +248,14 @@ continuation, or a respawn brings the seat back on its home mesh alone. Persisti
 crash-restart silently re-open outbound connections without any operator act in between, and it would
 put a second mesh's identity into recovery state that the lifecycle credential does not cover.
 
-**Decision: an additional connection's material is bounded by the seat's own lifecycle.** It is minted
-per lifecycle UID, like every other seat secret (`agentLifecycleCredsKey`,
+**Decision: an additional connection's material is bounded by the seat's own lifecycle, by whichever
+of the two mechanisms in 4.1 applies.** Where this host mints it (a local B), it is minted per
+lifecycle UID like every other seat secret (`agentLifecycleCredsKey`,
 `packages/workspace/src/agent-secrets.ts:194`), so it cannot outlive the incarnation that asked for
-it, and its path is recorded on the seat's tracked secret set so the per-seat teardown reaches it
-(4.2).
+it. Where the host only delivers what B's operator staged (a remote B), this host chose no lifetime
+and cannot impose one, so the bound is that credential's own TTL, which is why 4.2 makes the TTL
+load-bearing rather than a detail. Either way its path is recorded on the seat's tracked secret set so
+the per-seat teardown reaches it (4.2).
 
 **Decision: a bounded number of additional connections, refused past the bound rather than queued.**
 The default bound is 3. It exists because each connection is a full endpoint with durables, a
@@ -286,10 +289,12 @@ tool argument or result.**
 The sequence for `cotal_mesh_connect(mesh: "B")`:
 
 1. The seat invokes the manager command `mesh-material` over the existing lifecycle endpoint.
-2. The manager resolves record B, confirms operator-staged admission material for B exists under its
-   staging root (2.2), **mints** the seat's per-lifecycle credential for B from that material, and
-   writes one `LaunchMaterial` file with `writeLaunchMaterial`
-   (`packages/core/src/launch-material.ts:83`).
+2. The manager resolves record B, confirms operator-staged material for B exists under its staging
+   root (2.2), and obtains the seat's credential for B by whichever of the two paths applies under
+   the rule below: it **mints** one when this host holds B's signing seed, or **takes the per-seat
+   credential B's operator staged** when it does not. It never derives one from the other: a staged
+   JWT is not a seed and cannot be minted from. It then writes one `LaunchMaterial` file with
+   `writeLaunchMaterial` (`packages/core/src/launch-material.ts:83`).
 3. The reply carries the **path**, never the material. Path, mode, and space are not secret; the file
    is 0600 and the 0700 directory is owned by the same user the seat runs as.
 4. The seat calls `readLaunchMaterial`, constructs the second `MeshAgent`, then calls
@@ -399,7 +404,7 @@ and `define-persona`:
 |---|---|---|
 | `list-meshes` | Return the redacted projection | Yes: add to `REPEAT_SAFE_COMMANDS[manager]` (`endpoint-grants.ts:234`) |
 | `register-mesh` | Create one record from coordinates, refusing an existing name | No: it refuses on conflict, so a retry after a responded-but-unbound split cannot be told from a duplicate create |
-| `mesh-material` | Mint one material file for one target mesh | No: it mints |
+| `mesh-material` | Provide one material file for one target mesh: minted where the host holds B's seed, delivered from staged material where it does not (4.1) | No: it mints or hands out a credential |
 | `mesh-disconnect` | Drop a tracked extra-mesh material path, and deprovision on B only where a B-scoped deprovisioner exists (4.2) | No: a second run acts on whatever now holds the path, the `despawn` reasoning in that table's own comment |
 
 **Decision: a new capability, `mesh`, gates all four.** It is added to the closed set in
@@ -407,26 +412,27 @@ and `define-persona`:
 `meshCallerCapabilities` grant set on the manager endpoint in the same shape as
 `spawnCallerCapabilities` (`endpoint-grants.ts`).
 
-**Decision: the tool filter for the `mesh` family is UNCONDITIONAL on the declared capability, and
-this is deliberately NOT what `canSpawn` does.** `canSpawn` is
-`!isAuthed(config) || config.capabilities?.includes("spawn")` (`tool-specs.ts:524`): permissive on an
-open mesh, because open mode mints no identity, so the broker is not a boundary there and the filter
-would be claiming one that does not exist. That reasoning does not carry over. The authority the
-`mesh` family exercises is **host-local**: it reads and writes the machine's registry and asks the
-manager to mint material on disk. None of that is granted by the mesh's broker, so an open mesh does
-not make it free, and mirroring `canSpawn` would have handed every persona on an open mesh the whole
-family while this document claimed the capability gated it. #865 asks for these tools to be injected
-only for personas declaring the capability, and on an open mesh that is only true if the filter is the
-boundary. That leaves the open-mesh case, and the honest answer is a refusal rather than a filter. On an open
-broker the caller tuple on a request subject is not authenticated, which is the whole reason
-`canSpawn` gives up there. A manager-side re-check of the declared capability cannot bind either,
-because the handler has no authenticated principal to check the declaration against, and an earlier
-draft of this section claimed such a re-check as though it could. **Decision: the `mesh` family is
-refused outright on an open home mesh.** No listing, no registration, no material. An open mesh has
-no way to tell one caller from another, and the authority in question reaches the host filesystem, so
-the only safe answer is that nobody gets it. An operator on an open mesh uses the CLI. Stated plainly:
-for `spawn` the filter is truthfulness and the broker is the boundary; for `mesh` the broker is the
-boundary on an authenticated mesh, and on an open one the feature is off.
+**Decision: on an authenticated mesh the broker is the boundary and the tool filter mirrors it; on an
+open mesh the `mesh` family is refused outright.** This is deliberately not what `canSpawn` does, and
+the difference is worth stating because the first two drafts of this section got it wrong in opposite
+directions.
+
+`canSpawn` is `!isAuthed(config) || config.capabilities?.includes("spawn")` (`tool-specs.ts:524`):
+permissive on an open mesh, because open mode mints no identity, so the broker is not a boundary there
+and a filter would be claiming one that does not exist. Mirroring that for `mesh` would hand every
+persona on an open mesh the whole family while this document claimed a capability gated it. But the
+opposite move, which an earlier draft made, does not work either: it kept the filter and promised a
+manager-side re-check of the declared capability behind it. That re-check cannot bind. On an open
+broker the caller tuple on a request subject is not authenticated, so the handler has no principal to
+check a declaration against, which is the same reason `canSpawn` gives up.
+
+So neither a filter nor a handler check is a boundary on an open mesh, and the authority in question
+is not the broker's to grant in the first place: the `mesh` family reads and writes the machine's
+registry and asks the manager to mint material on disk. Host-local authority, offered to callers the
+host cannot tell apart, is not something to gate. It is something to withhold. **On an open home mesh
+there is no listing, no registration, no material, and no connect.** An operator on an open mesh uses
+the CLI. On an authenticated mesh the broker enforces the minted grant and the filter keeps the
+advertised surface truthful, which is #865's "injected only for personas declaring the capability".
 
 **Decision: `mesh` is a separate capability from `spawn`, not folded into it.** `spawn` is already
 documented as host-launch authority (`agent-file.ts:85` comment) because `launchOptions` is a raw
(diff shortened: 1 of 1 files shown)


davidfarah2003:
BLOCK at 85f7ba03fccaff9e59fe28239d5a099f6a560db2

Adversarial security review of docs/design/mesh-connections.md.
Head confirmed: `git rev-parse HEAD` -> 85f7ba03fccaff9e59fe28239d5a099f6a560db2 (exit 0).
Issues read in full: Cotal-AI/Cotal#865 (including triage comment) and #1207.

Lens: attacker-influenced content, credential lifetime, enforcement layer, fail-open, ordering, new primitives. Citations checked against the pinned tree.

## Contradictions

1. Section 4.2 and 7.4 name the wrong reaper, so additional-mesh (and hosted) credentials are not destroyed by the mechanism the document relies on.

   Claim: on disconnect, stop, despawn, and lifecycle end, additional-mesh material is reaped by the existing reaper because it is keyed with `agentLifecycleCredsKey`; stop/despawn already call `reapAgentSecrets` (`packages/workspace/src/agent-secrets.ts:466`); hosted stop step 3 is the same call, unchanged.

   Why it fails: `reapAgentSecrets` is the `cotal space rm` tenant wipe. It enumerates every agent secret under one space segment and recursively deletes that segment. It is not invoked from manager stop/despawn. Per-seat teardown is `driveDeprovision` / `driveStaticRetirement` in `implementations/manager/src/manager.ts`, which deletes only the recorded home `secretPaths` family (else `agentLifecycleSecretFilePaths` for this space, this name, this uid). Additional-mesh material is for another space and another root, so it is outside that family. Keying it "the same way" either collides with the home creds file or stores it where the home sweep never looks. Implemented as cited, stop would wipe the whole space's secrets. Implemented as the real stop path, extra-mesh JWTs and attach material survive with no owner.

2. Section 2.2 refuses deletion because overwriting host registry state is a denial primitive, then section 5 specifies `register-mesh` as an overwrite of that same state.

   Claim: no agent-facing path writes `~/.cotal/meshes/`; the manager owns the write and may refuse; deletion is not offered because `MeshEntry.origin` already decides who may delete, and a seat removing a record another seat resolves through is a denial primitive with no use case. Table in section 5: `register-mesh` is not repeat-safe because "a second run overwrites a record that may have changed in between". Tool args include `server`, `mode`, `tlsRequired`, and `attachHost`.

   Why it fails: today's CLI refuses a second `meshes add` without `--force` (`implementations/cli/src/commands/meshes.ts`, existing-record check). `origin` defaults to `up` when omitted (`packages/workspace/src/mesh-registry.ts` header on `MeshEntry.origin`), and `up` records are what later `cotal spawn` from any directory resolves, what liveness prune may drop, and what a later manager launch reads for `attachHost`. An overwrite of `server` retargets every subsequent spawn on the box. An overwrite of `attachHost` persists a bind address the operator never chose. An overwrite that drops `origin: "manual"` makes a remote mesh auto-prunable. That is the denial (and hijack) primitive section 2.2 said it was not offering. "May refuse" is not a named refuse of existing `up`/`manual` records, so absence of that rule fails open as overwrite.

3. Section 2.2's exfil closure is not delivered by the register/connect/material sequence as specified.

   Claim: a registration request carries coordinates only, cannot carry secrets, and a record it creates is inert until an operator provisions admission material. That is what stops one channel message from causing an authenticated connection to a broker the operator has never seen. `cotal_mesh_connect` then "confirms admission material for this seat exists", and `mesh-material` "mints or copies" the per-lifecycle credential for B into a #614 file.

   Why it fails:
   - `mode` is an argument. `mode: "open"` needs no admission material. If connect still requires material, the local-open case cannot connect; if it does not, the record is not inert and the seat opens a credless connection to a seat-chosen `server`.
   - Register args omit `root`. `MeshEntry` requires `root`. The manager must invent one. If it uses the home workspace root, existing `.cotal/auth` for that space name is "admission material" with no new operator act, and "mints or copies" can copy a live home JWT toward the registered server.
   - The CLI's actual safety fences (`checkDialPolicy` / `classifyJoinTarget`: loopback or overlay-with-ack, public only with recorded TLS, no RFC1918, no hostname without TLS) are not named as register-mesh refusals. Without them, a seat-supplied `server` is a credential destination. Copying home material onto that destination is the exact "message in, authenticated connection out" channel section 2.2 claimed to close, with the secret coming from the host rather than from the tool args.

## Ideas (not contradictions; future work)

- `cotal_inbox` draining every mesh into one turn puts a foreign trust domain in the same decision loop as home tools that default `mesh` to home.
- Ungated `cotal_orientation` adding registry `origin` widens past space/server/mode the seat already learned from its own connection.
- On open home meshes, spawn-style `!isAuthed || capability` advertising would show the mesh family to every seat. Outbound connect is a larger blast radius than spawn-on-open and wants an explicit refuse rather than inheriting spawn's open-mode hole.
- Hosted loopback plus "token is the boundary" leaves off-host exposure to Q2; a stolen attach token on a later-bound port is the 7.4 restart/port-reuse case until exposure is designed.
- File delete is still footprint reduction, not JWT revocation (`driveDeprovision` already says this). Extra-mesh disconnect must treat endpoint stop failure as a live credential, not a successful reap.
- `SESSION_GRANT_MAX_TTL_MS` is a one-use session-grant ceiling, not an attach-token lifetime. Using it as the hosted attach cap (Q4) mixes two artifacts.

No round-3 residual beyond those ideas if the three contradictions are closed.

davidfarah2003:
BLOCK at 85f7ba03fccaff9e59fe28239d5a099f6a560db2

Coverage checked. The four #865 triage gates are read and write authority over host records, multi-endpoint routing and lifecycle, private credential storage and revocation, and a host adapter that keeps workspace concepts out of connector-core. The three #1207 asks are a persona-limited MCP path enforced mesh-side, normal seat ownership, and lifecycle-scoped provisioned credentials that are never ambient and are revoked on stop.

1. **Sections 7.1, 7.2, and Q2 do not provide the hosted session requested by #1207.** `cotal spawn --agent hosted` creates only a local bridge. The document neither creates or provisions the cloud session nor defines how the attach URL and token reach it. It leaves off-host reachability open in Q2 while admitting most cloud sessions cannot reach the loopback listener. This leaves the current human attachment step and the hosted side's uncontrolled cwd and model unresolved, so the limited hosted MCP path is deferred rather than designed.

2. **Section 7.1 contradicts #1207 requirement 3 and does not flag the divergence.** The issue requires the hosted side to receive a lifecycle-scoped mesh credential through the private carrier. The design explicitly says the hosted side never receives a NATS credential and substitutes an attach token. Only the separate forking divergence is sent to Q3 for ratification.

3. **Sections 4.2, 5, and 6 do not deliver the claimed additional-mesh revocation.** `cotal_mesh_disconnect` is local, but none of the three manager commands reports a disconnect or requests revocation. The cited `reapAgentSecrets` is a whole-space cleanup, not the manager's per-seat stop path. Current per-seat static cleanup deletes the credential file and explicitly states that this is footprint reduction, not revocation, because a copied JWT remains valid until expiry or rotation. This leaves the #865 credential-revocation gate uncovered.

4. **Section 7.4 states an ordering that the named mechanism cannot perform.** Attach and session tokens live in the bridge, but no manager-to-bridge revoke operation or shared token store is specified. The existing authenticated control frame supports only `shutdown` and `session`, and the manager's current control client sends only `shutdown`. Therefore the manager cannot revoke tokens first as promised, so the lane requirement to say how stop and revocation reach a no-pid seat is unmet.

5. **Sections 2.2, 5, and 6 define a registration request that cannot construct the host record it claims to write.** The tool omits `root`, although `MeshEntry` requires it and uses it to locate the staged auth material and personas. It also permits `mode: user` without the required `userAuth` trust metadata. No derivation or fixed staging root is chosen. The manager therefore cannot validate and write a usable record from the proposed DTO, which fails both the #865 register ask and the host-adapter gate.

6. **Sections 5 and 6 do not match the existing capability gate they say they mirror.** #865 requires these tools to be injected only for personas declaring the capability. Existing `canSpawn` filtering is permissive on open meshes, where it exposes the tools without `capabilities: [spawn]`, because no broker identity boundary exists. Saying to filter `mesh` the same way means open-mesh personas without `mesh` still receive the surface, contradicting both the issue and the claim that `mesh` gates all four tools.

7. **The rotation citation in section 7.4 does not support its claim.** `packages/workspace/src/renewal.ts` renews the delivery and membership daemon credentials only. Managed agent credential renewal is separate manager logic. The cited mechanism therefore does not establish that hosted-seat rotation is covered unchanged.

**Ideas:** None. The findings above are unmet requirements, unsupported guarantees, or unratified divergences.

davidfarah2003:
BLOCK at fc10a00f4ffd500088cb2f44a4f302be75ceb144

Adversarial security review of docs/design/mesh-connections.md, round 2.
Head confirmed: `git rev-parse HEAD` -> fc10a00f4ffd500088cb2f44a4f302be75ceb144 (exit 0).

Round 1 contradictions 1 to 3 are closed in this revision: the named reaper is now the per-seat path, register-mesh is create-only, root is not agent-chosen, open and user are refused, the dial policy is named, and mint does not copy. This round grades only what is still false.

## Contradictions

1. Section 4.2 claims local extra-mesh revocation follows the home path, but the function it cites cannot revoke on mesh B.

   Claim: `deprovisionBroker` (`implementations/manager/src/manager.ts:2878`) is pinned to the home mesh, so a foreign mesh cannot be revoked from here; but for a local second mesh whose auth root is on this host, "revocation is real and follows the home path." `mesh-disconnect` "for a local mesh, deprovision."

   Why it fails: `deprovisionBroker` mints a deprovisioner against `this.auth` and calls `deprovisionAgent` with `this.space` and `this.servers`. That is the home broker. A second local mesh is another space with another signing seed and another broker URL, even when its root is on the same disk. Calling the cited function deletes the seat's home footprint, or no-ops against the wrong broker. It does not revoke the extra-mesh JWT. The document names no B-parameterized deprovisioner, so the local exception is not delivered by the mechanism it cites. File delete remains footprint reduction, which the document already admits.

2. Section 2.2 creates a `MeshEntry` without setting `origin`, so an agent-created record fails open as a locally started mesh.

   Claim: the manager owns the write and validates it; deletion is not offered because `MeshEntry.origin` already decides what may delete a record (`mesh-registry.ts:77`). Create-only prevents a seat from retargeting an existing record.

   Why it fails: omitted `origin` is `up` (`mesh-registry.ts:62-64`). The CLI's own create path sets `origin: "manual"` (`implementations/cli/src/commands/meshes.ts:194`). An `up` record is auto-pruned, is torn down with this root, and is what a later `cotal up` for that space claims. A mesh-capable seat that wins create-only on a staged-but-not-yet-CLI-registered name therefore publishes a record that the rest of the host treats as "this machine started that mesh," including its seat-supplied `server` and `attachHost`. That is the denial and hijack primitive section 2.2 said it was not offering, arriving through absence of a field the document already knows is load-bearing.

This is the final round. No third pass.

## Ideas (not contradictions; future work)

- Staging a seed does not pin a broker. `cotal_mesh_register` still takes `server`. After the operator creates `~/.cotal/staged/<space>/` and before they run `cotal meshes add`, a mesh-capable seat can bind that seed to any dial-policy-passing destination, including public TLS, and `mesh-material` will mint toward it. Pin the server in the staged dir, or do not take `server` from the seat.
- On an open home mesh the broker still authenticates nobody. The new unconditional tool filter hides the family from that seat's model; it does not stop another open-mesh client from invoking the four manager commands. Handler re-check of "declared capability" has no authenticated caller to check. Refuse the family unless the caller is a managed seat recorded at spawn.
- Drop `attachHost` from the agent register args. It is a bind address, and create-only still lets a seat persist one the operator never chose.
- Section 0 still says three manager commands. Section 5 says four.
- `{op:"revoke"}` should drop live HTTP MCP sessions, not only refuse the next attach. The window between revoke and hard stop is otherwise still a live hosted seat.

davidfarah2003:
BLOCK at fc10a00f4ffd500088cb2f44a4f302be75ceb144

1. **Sections 7.1 and 7.2 contradict #1207 requirement 1.** The issue requires the default to be the smallest usable lane surface. The design instead refuses every hosted launch whose persona omits `tools:` and says no default is invented. This deliberate departure is not included in the divergences submitted for ratification in Q3.

2. **Sections 7.1 and 8 leave two properties from #1207 unmet while claiming to fix them.** The issue identifies that no spawner can create the hosted participant and a human attaches it by hand. The design explicitly keeps creation and provisioning of the hosted session outside Cotal and still requires a human or operator automation to attach it. `cotal spawn` creates a local bridge, not the hosted session whose process defines the requested seat. The same section also says cwd and model remain whatever the remote session uses. Its claim that it fixes the first three properties outright is therefore false, and the hosted-session and cwd/model gaps are deferred rather than covered.

3. **Sections 4.1 and 4.2 cannot both hold for the remote-mesh case that #865 requires.** Section 4.1 says the home manager always mints the extra-mesh credential and never copies staged material. Section 4.2 says that when the home host lacks the remote signing seed, it cannot mint and the usable material is what the remote operator staged. Passing that staged per-seat credential requires the copy section 4.1 forbids. Staging the signing seed instead gives the home host minting and revocation authority, contradicting the stated remote case. The fleet-of-fleets path therefore has no composable credential flow.

4. **Section 5 does not deliver declared-capability gating on an open mesh.** It correctly notes that `canSpawn` is permissive because an open broker authenticates no caller, but then relies on a manager handler re-check of the declared `mesh` capability. The existing endpoint handler sees only the caller tuple from the request subject. On an open broker that tuple is not authenticated, so it cannot be trusted to select a managed persona's capability record. A tool-list filter cannot authorize a host registry operation against bypassing callers. The design must either refuse this family on open meshes or name a host-authenticated boundary.

5. **Section 7.1's pre-side-effect refusal is contradicted by the cited launch architecture.** It promises that `buildLaunch` refuses a hosted persona with no `tools:` before any provisioning side effect. The manager currently provisions durables, mints and ledgers credentials, and materializes the credential file before calling `connector.buildLaunch` (`manager.ts:3780-3819`, then `:3879`). The document names no preflight or ordering change that can make the stated guarantee true.

6. **Sections 7.1, 7.2, and Q2 call off-host exposure settled without defining its record or TLS mechanism.** `MeshEntry.attachHost` is one mesh-level bind decision for the manager attach face, not a per-hosted-seat listener record. The proposed surface adds only `tools:` to the persona and names no field or owner for a hosted bridge's address, port, TLS certificate, or issued attach URL. Requiring TLS does not specify how the HTTP bridge serves it. The design therefore still defers the attach path needed by most cloud sessions while saying only the optional first-party tunnel remains open.

7. **Section 0's host-adapter gate row disagrees with sections 5 and 6.** It still says three manager-endpoint commands, while the revised design requires and enumerates four. This is a direct table-to-section contradiction.

**Ideas:** None. These are unmet source requirements, mechanisms that do not deliver their stated guarantee, or statements that cannot compose.

davidfarah2003:
BLOCK at 2d3c7b9beb11cf18ee5fc0319da99129ef37d821

Bounded verification of docs/design/mesh-connections.md.
Head confirmed: `git rev-parse HEAD` -> 2d3c7b9beb11cf18ee5fc0319da99129ef37d821 (exit 0).

(i) Round-2 contradictions are closed.

C1. Section 4.2 now says `deprovisionBroker` (`manager.ts:2878`) is the home broker in every case, including a local second mesh. Revoking on B needs a deprovisioner minted against B's own auth and dialled at B's server, named as a required addition possible only where B's seed is reachable. Remote B has no such addition and TTL is the only bound. The section 5 `mesh-disconnect` row no longer promises a local deprovision.

C2. Section 2.2 now stamps `origin: "manual"` explicitly, cites the omitted-means-up default (`mesh-registry.ts:62-64`) and the CLI write (`meshes.ts:194`), and states the field is not an argument.

(ii) New contradiction introduced by this delta.

1. Section 4.1's numbered sequence and the mint-or-deliver decision cannot both hold.

   Claim, sequence step 2: the manager always **mints** the seat's per-lifecycle credential for B from staged material. The section 5 `mesh-material` row says the same: "Mint one material file for one target mesh."

   Claim, next paragraph: material is either minted from a seed the host holds, or delivered from what the operator staged for that exact record. Remote B "is a copy of a file."

   Why it fails: the mint-or-deliver paragraph was added to close the remote case. The sequence and the command table were not updated and still describe mint in every case. Implementing the sequence makes remote B impossible or treats a staged JWT as a signing seed. Implementing the decision makes the sequence and the table false. Two adjacent statements, one leftover from the previous rule.

This is the only finding under (i) or (ii).

## Follow-ups (not a BLOCK)

- Pin the broker in the staged directory. Register still takes `server` from the seat, so delivered remote material can still ride a seat-chosen destination.
- Drop `attachHost` from the agent register args.
- Overlay acceptance is still not an argument, so a seat cannot register an unencrypted overlay target. Fail-closed, but unnamed.
- `{op:"revoke"}` should drop live HTTP MCP sessions, not only refuse the next attach.

davidfarah2003:
BLOCK at 2d3c7b9beb11cf18ee5fc0319da99129ef37d821

1. **Round 2 finding 3 is not fully closed in section 4.1.** The connection sequence still says the manager resolves every B and **mints** the seat's credential from its staged material. The revised decision immediately below says a remote B cannot be minted by the home host and instead delivers the credential B's operator staged. Those statements cannot both describe the remote flow. The sequence must branch between minting for local B and delivering staged material for remote B.

2. **The fix for Round 2 finding 1 introduced a new contradiction in Q3.** The heading still says “Ratify two divergences,” but it now enumerates three, including the new requirement-1 default-surface departure at Q3(c). The departure is now substantively flagged with its reason and cost, but the heading falsely states the ratification set.

**Follow-ups:** None within this bounded pass.

davidfarah2003:
APPROVE at b421601a1d960eb651639fabdebe0964784806e5

Bounded delta confirm of docs/design/mesh-connections.md.
Head confirmed: `git rev-parse HEAD` -> b421601a1d960eb651639fabdebe0964784806e5 (exit 0).

(i) The previous finding is closed. Section 4.1 step 2 now branches: mint when this host holds B's signing seed, take the staged per-seat credential when it does not, and never treat a staged JWT as a seed. The section 5 `mesh-material` row describes both paths.

(ii) No new contradiction from this delta. The sweep siblings I checked all agree with the branched rule: 3.2 bounds minted material by lifecycle UID and delivered material by that credential's own TTL; the gate table no longer says extras are always revocable; the 7.2 gate row no longer says three layers against a four-row table; Q3 names three divergences; the open-mesh block is one argument ending in a refuse rather than a filter-then-refuse.

Follow-ups, already named, not a BLOCK: pin the broker in the staged directory; drop `attachHost` from agent register args; overlay acceptance is still not an argument so unencrypted overlay cannot be registered from a seat; `{op:"revoke"}` should drop live HTTP MCP sessions.

davidfarah2003:
BLOCK at b421601a1d960eb651639fabdebe0964784806e5

The two bounded-pass findings are closed. The credential sequence now branches between minting and delivering staged material, and Q3 accurately names three divergences.

1. **The sibling sweep introduced a new contradiction in section 3.2.** The decision headline says an additional connection's material is bounded by the seat's own lifecycle. Its remote-B branch immediately says this host cannot impose that lifetime and the only bound is the staged credential's own TTL. A TTL chosen independently by B may outlive the home seat's lifecycle, so it does not deliver the stated lifecycle bound. The decision must either require B's staged credential to expire no later than the seat lifecycle, with a mechanism that establishes that bound, or state that remote material is TTL-bounded rather than lifecycle-bounded.

**Follow-ups:** None within this delta-confirm scope.

davidfarah2003:
BLOCK at 7b283fa2443bdb566381bb016086f2f88df08dfa

The prior lifecycle-bound contradiction is closed. The two credential paths now state different bounds, and sections 3.2, 4.2, and Q4 agree on an acceptance ceiling for remote material.

1. **The delta's acceptance-ceiling mechanism is incomplete in section 3.2.** It claims that reading `exp` makes short TTL enforced at this end, but `credsClaims` returns `exp?: number`. The decision refuses an expired credential or one whose remaining validity exceeds the ceiling, but does not refuse a credential with no numeric `exp`. Such an unbounded staged credential satisfies neither stated refusal and defeats the guarantee this delta adds. The decision must explicitly require a numeric `exp` and refuse when it is absent or invalid.

**Follow-ups:** None within this delta scope.

davidfarah2003:
BLOCK at 7b283fa2443bdb566381bb016086f2f88df08dfa

Bounded delta confirm of docs/design/mesh-connections.md.
Head confirmed: `git rev-parse HEAD` -> 7b283fa2443bdb566381bb016086f2f88df08dfa (exit 0).

(i) The earlier 4.1 finding stays closed. The mint-or-deliver branch is still in the sequence and in the `mesh-material` row. 3.2 no longer averages the two paths into one lifecycle bound.

(ii) New contradiction introduced by this delta.

1. Section 3.2's new refusal does not deliver a bound, because the function it cites does not enforce one and does not refuse a missing `exp`.

   Claim: the home host's lever over delivered material is a refusal, not a shortening. The manager reads `exp` with `credsClaims` (`packages/core/src/identity.ts:142`) and refuses material that is already expired or whose remaining validity exceeds a ceiling. That is what turns "short-TTL by requirement" into something enforced at this end. 4.2 repeats that this is the only lever.

   Why it fails: `credsClaims` is documented in that same file as unverified diagnosis, never enforcement (`identity.ts:124-126`, `:134-137`). `exp` is optional. A structurally valid creds file with no numeric `exp` parses and returns no expiry. The named check then has nothing to compare: not already expired, remaining validity not a number, ceiling not applied. The sibling `credsRenewalDelayMs` (`:158-161`) already fail-louds on a missing numeric `exp` for this reason: an unbounded cred is "a cred to keep silently forever". Citing the diagnosis parse without that refusal makes absence render as acceptable, which is the fail-open this document has been closing everywhere else. The ceiling being an unset Q4 number does not save it: even a chosen ceiling cannot bound a claim that is not there.

Follow-ups, already named, not a BLOCK: pin the broker in the staged directory; drop `attachHost` from agent register args; overlay acceptance unnamed; `{op:"revoke"}` should drop live HTTP MCP sessions.

davidfarah2003:
APPROVE at eedcbb46d3acda9c01d66e6be3613738a3310fa7

Bounded delta confirm of docs/design/mesh-connections.md.
Head confirmed: `git rev-parse HEAD` -> eedcbb46d3acda9c01d66e6be3613738a3310fa7 (exit 0).

(i) The previous finding is closed. Section 3.2 now states that `credsClaims` returns unverified claims and that a local read of those claims cannot bound anything. The admission check is a named refusal at staging time, not enforcement. A present numeric `exp` is required, citing `credsRenewalDelayMs` (`identity.ts:158`). 4.2 and Q4 match: the bound stays with B; only the ceiling's value is open.

(ii) No new contradiction from this delta. The rewritten 3.2 paragraph no longer claims a local bound, and its three refusal tests cannot all pass by absence of `exp`.

Follow-ups, already named, not a BLOCK: pin the broker in the staged directory; drop `attachHost` from agent register args; overlay acceptance unnamed; `{op:"revoke"}` should drop live HTTP MCP sessions.

davidfarah2003:
APPROVE at eedcbb46d3acda9c01d66e6be3613738a3310fa7

The prior optional-`exp` finding is closed. Section 3.2 now requires a present numeric, unexpired `exp` within the configured ceiling, while correctly describing the local read as a diagnostic admission refusal rather than the enforcement boundary. Sections 4.2 and Q4 compose with that decision. I found no new contradiction in this delta.

**Follow-ups:** None within this bounded scope.

davidfarah2003:
Merged as the design record at merge commit 4c9234633056cc7825d1c263c6b58751a192320d, parent2 eedcbb46d3acda9c01d66e6be3613738a3310fa7, with APPROVE from rev-meshconn-sec and rev-meshconn-spec at that exact sha and CI 17 of 17 green.

Merging settles nothing in the "Open for David" list above. The seven questions remain open, the three divergences in item 3 are submitted for ratification and are not ratified by this merge, and Phase 2 stays gated on David reading the design. The graded artifact and the authority for those questions is docs/design/mesh-connections.md section 9; the list in this PR body is a convenience copy.

PR #1214, opened 2026-09-03
fix(connector-jcode): keep a seat name launchable after the seat is stopped or reaped

### LINKED ISSUE #1211: connector-jcode: a seat name is unlaunchable after the seat is stopped or reaped, even once every old process has exited; a renamed identical persona launches
A managed jcode seat NAME becomes permanently unlaunchable after that seat is stopped or reaped, on the same manager, even after every process of the old lifecycle has exited. A byte-identical persona under a new name launches first try. Reproduced three times on one host in two days.

## Reproductions (netcup-vps, Linux, cotal-ai 0.37.0 then 0.38.0, connector-jcode same versions)

| when | dead seat | how it died | respawn under the same name | same persona, new name |
|---|---|---|---|---|
| 2026-09-01 | `issue-orchestrator-v2` | manager restart | `Jcode host startup failed (unknown)` every attempt | `issue-orchestrator-v3` launched |
| 2026-09-02 23:22 UTC | `issue-orchestrator-v3` | manager restart (0.38.0 upgrade) | failed 3 times over 5 minutes, see below | `issue-orchestrator-v4` launched |
| 2026-09-03 00:2x UTC | `issue-orchestrator-v4` | graceful `cotal_despawn` | failed via the cron watchdog's `cotal spawn --detach` | `issue-orchestrator-v5` launched |

A control on the same connector and model (`fleet-worker`, jcode / grok-4.6) connected to the mesh during the v3 wedge, so the connector and provider were healthy throughout.

## What the logs say

The connector-side log (`~/.cotal/jcode/<seat-dir>/logs/connector-<ts>-<pid>.log`) carries only:

```
```

The first attempt after the v3 reap also logged `could not list prior sessions, starting fresh: write EPIPE`.

The seat's private jcode log, first attempt after the reap:

```
[ERROR] Another jcode server process is already running for runtime dir /tmp/jc-958302bed350/home/run
```

That server was the OLD lifecycle's jcode server (session `session_stallion_…` from 2026-09-01). It carries no `COTAL_NAME` in its environment, so a reap that terminates seat processes by that key leaves it running, and it only exits on its own 5-minute idle timer (`Server idle for 5 minutes with no clients. Shutting down.` at 01:24:41 local). That explains the first failure.

It does not explain the rest. A further attempt under the same name at 01:27, after that server had exited and with no process holding any `/tmp/jc-*` cwd or descriptor, failed identically and **wrote nothing at all to the private jcode log**: the newest entry stayed the old server's shutdown line. The seat home under `~/.cotal/jcode/<space>-<name>-<hash>/` is reused across lifecycles (same hash for the same name), so whatever the connector refuses on is something persisted in that home or in the runtime dir it derives from the name, not a live process.

## Why it matters

- A manager restart, or any stop, silently retires the seat NAME. Personas, watchdogs, channel ACLs, and operator memory all key on the name; every restart forces a rename and an edit of everything that refers to it.
- The failure is reported as `(unknown)` with a pointer to a log that, past the first attempt, contains nothing about it, so an operator cannot tell a wedged name from a broken connector without running a control spawn.
- The obvious workaround, session transplant to resume the old jcode session under the new name, cannot be done safely, because the seat home is created at first spawn and any stop to transplant into it wedges the new name too.

## Expected

Either the connector cleans the per-name runtime state on retirement so a name is relaunchable, or the refusal names the file, lock, or directory it is refusing on so the operator can clear it. And the reaper should terminate the old lifecycle's jcode server rather than leaving it to time out; it is the seat's own child.

## Not established

What the connector reads in the reused home that makes the second and later attempts fail with no jcode log. `active_pids/` in the old home holds stale session pid files from 2026-09-01; that is a candidate, not a finding.

davidfarah2003:
Closes #1211.

## Root cause

`shortSocketHome()` derives the seat's short socket alias as `/tmp/jc-<sha256(seat home)[:12]>/home`. The seat home is a pure function of space and name, so that path is a pure function of the seat NAME and is reused for the life of the machine. The alias was required to be a symlink, and anything else was a hard refusal.

The startup-failure path in `runJcodeHost` calls `socketHome.dispose()`, which deletes the alias and rmdirs its parent. `stopPrivateJcode()` cannot reach a jcode server from a *previous* lifecycle, because that server does not carry this launch's identity nonce. So the alias is deleted out from under a server that is still running with `JCODE_HOME` pointing at it, and that server re-creates the path as a real directory on its next write. From that moment every launch of the same name throws in `shortSocketHome` before `launchInstance` is ever called.

That is both halves of the reported behaviour:

- attempts 2 and later wrote **nothing** to the private jcode log, because jcode was never started;
- the failure was reported as `(unknown)`, because the refusal message names local paths and `startupFailureCode()` deliberately renders only allow-listed codes, never message text.

**The trigger is wider than a stop or a reap.** An independent reproduction from another lane (issue #1211 comment) wedged a name that had NEVER been seated: its first launch died at the provider, and the name was unlaunchable from that moment, six attempts across three models with two deleted-and-recreated seat homes. The seat home derives alias `/tmp/jc-920967bbe67d`, which I re-derived from the NAME alone rather than taking on report: `sha256("netcup\0rev1210-measure2")[:12] = a6555b984282`, then `sha256("/home/cotal/.cotal/jcode/netcup-rev1210-measure2-a6555b984282")[:12] = 920967bbe67d`. That path is a real directory whose mtime, `03:58:03.526756716`, is 0.598s after the provider refusal at `03:58:02.929`, written by the server that was still shutting down. So the condition is **any startup failure late enough to have started jcode**, and a provider refusal is the cheapest way in because nothing local is wrong. A name that was never used is not safe.

The first failure after a reap has a different and self-clearing cause: the surviving server still held the seat home's runtime directory, so jcode refused with `Another jcode server process is already running`. It carries no `COTAL_NAME` and setsids into its own group, so a name-keyed reap does not reach it and it only exits on its own five-minute idle timer.

Evidence from the reporting host, live at `~/.cotal/jcode/netcup-issue-orchestrator-v3-dcde13b0a7d6`:

- `sha256("/home/cotal/.cotal/jcode/netcup-issue-orchestrator-v3-dcde13b0a7d6")[:12] == 958302bed350`, matching `/tmp/jc-958302bed350` in the issue's jcode log line.
- `/tmp/jc-958302bed350/home` was a **real directory**, birth `01:22:51.697`, holding `servers.json`, `logs/`, `models/`, `ambient/`. Its mtime `01:24:41.232` is the same instant as the old server's `Server idle for 5 minutes with no clients. Shutting down.` line, so the old server wrote it.
- A probe calling `shortSocketHome` on that home from this worktree at `b498b940e` (origin/main, before any change here) threw `Jcode short socket alias is not a symlink: /tmp/jc-958302bed350/home`, with no `code`, which `startupFailureCode()` maps to `unknown`.

## The fix

1. **Reclaim the alias instead of refusing it** (`private-state.ts`). The entry is classified with `lstat` and removed through the parent that was just verified as an owner-only 0700 directory, so the removal cannot follow a link out of that directory and cannot reach the seat home the alias names. What keeps two servers off one name is jcode's own runtime-directory lock inside the seat home, which is unchanged.

2. **Adopt the previous lifecycle's tree** (`private-lifecycle.ts`, `host.ts`). Each launch records its identity nonce and its host process in the seat home; the next launch of that home stops the tree that record names, but only when the recorded host PID is **not alive at all**. A live PID at that slot is ambiguous, either this home's connector or an unrelated process that reused the number, and nothing on a live PID separates them reliably, so the ambiguous case resolves toward signalling nothing. The launch then meets jcode's runtime-directory lock, which is the correct refusal for a name a live seat holds. One nonce is inherited by one launch's descendants alone.

   > **This gate was wrong in the first three pushes of this PR and a reviewer caught it.** It originally skipped only when the host was alive **and** matched its captured start token, which resolved the ambiguous case toward *killing*: a live connector whose recorded token did not match had every process carrying its nonce SIGTERMed and SIGKILLed while the connector itself was left running. That is the two-seats-under-one-name failure this mechanism exists to prevent, inverted. It was reproduced with a live host and a mismatched token before being fixed in `4256ba0e0`. The earlier claim in this body, that a live seat could not be reached because the host is checked first, was false for any live host failing that token check.

3. **Name the refusal** (`startup-diagnostics.ts`, `host-main.ts`, `host.ts`). Preparing the seat's private state now reports a `private_state` code instead of `unknown`. Message text stays scrubbed.

## Repro to green

All on the reporting host, against the wedged `issue-orchestrator-v3` name.

**Before.** Installed `cotal-ai 0.38.0`, foreground, `00:33Z`:

```
```

Both `jcode-*.log` mtimes were unchanged by that attempt (`01:24` and `01:22`), confirming jcode was never started.

**After.** Same name, same wedged `/tmp` state, this branch:

```
[cotal-jcode] started a fresh session (no resumable prior session in this home)
[cotal-jcode] model grok-4.6 is served by provider Conifer via openai-compatible:conifer
```

**Second control, a fixture this lane did not produce.** The other lane preserved its wedged `/tmp/jc-920967bbe67d`, produced by the first-launch path rather than by a reap. Running this branch's `shortSocketHome` against it: before, a real directory; after, a symlink to the seat home. I snapshotted it first and restored it afterwards, so their before-state survives.

**The orphan half**, reproducing the issue's own "manager restart" row. `SIGKILL` the connector host with no teardown; two jcode processes survive holding `JCODE_HOME=/tmp/jc-958302bed350/home`, and the alias is still a symlink because `dispose()` never ran. Relaunching under the same name:

```
[cotal-jcode] stopped 3 Jcode process(es) left running by a previous lifecycle of this seat
[cotal-jcode] resumed session session_chick_… (121619 bytes of transcript)
[cotal-jcode] model grok-4.6 is served by provider Conifer via openai-compatible:conifer
```

The previously surviving PIDs were gone. Before this change that relaunch met `Another jcode server process is already running` for five minutes. A clean `SIGTERM` teardown afterwards left no jcode process for the seat and removed `/tmp/jc-958302bed350` entirely, so the wedge did not re-form.

## Tests

`smoke:jcode-private-state` gains the wedge case: take an alias, `dispose()` it, re-create the path as a real directory the way the surviving server does, and relaunch. It also pins the safety property that reclaiming a symlinked alias leaves the seat home it points at intact.

**It reds on main.** With `git show origin/main:extensions/connector-jcode/src/private-state.ts` in place and everything else at this branch's head:

```
✓ short API socket path stays below the cross-platform AF_UNIX budget
✓ short API socket directory is owner-only
✓ short API socket home aliases the managed home
AssertionError: an alias a dead lifecycle re-created as a real directory is reclaimed, not refused
  forever — "Jcode short socket alias is not a symlink: /tmp/jc-fad4b89bf0a8/home"
```

The red is on the new assertion's own name and prints the wedge refusal, so it is attributable rather than an incidental early failure. On this branch the suite is 25/25.

`smoke:jcode-private-lifecycle` gains the adoption case, driven by real processes in their own process groups whose `/proc/<pid>/environ` carries the nonce: a live recorded host means nothing is signalled and the process is still running; a dead recorded host means the process is adopted and gone; a record with no survivor signals nothing. This half cannot red on main, because `stopOrphanedTree` does not exist there.

Both suites are already in `bin/smoke/ci-suites.txt`, so nothing is added to that file. Both use only `mkdtemp` roots and spawned children, so they do not depend on the reporting host. The process half is guarded off Windows, matching the existing platform guards in these suites.

## Suites run

`smoke:jcode-args`, `jcode-host`, `jcode-private-state`, `jcode-private-lifecycle`, `jcode-session-resume`, `jcode-lifecycle`, `jcode-startup-diagnostics`, `jcode-inbound-wedge`, `jcode-mid-turn-delivery`, `jcode-route-identity`, `jcode-retry-policy` all pass, plus `pnpm typecheck` across every package and `pnpm check:docs-voice`.

`smoke:jcode-security` fails on the reporting host because it shells out to `docker`, which is not on this host's `PATH`; it fails identically with `origin/main`'s `private-state.ts` in place, so it is an environment gap rather than a regression. Its assertion is the foreign-owner ownership guard, which still runs before the new reclaim: `privateDirectory(socketDir, { requireOwner: true })` is unchanged and precedes it.

## What the tests do not cover

Stated plainly so the suite is not read as grading more than it does.

- **`stopOrphanedTree`'s wiring and its diagnostic line, and the `private_state` code, are graded by no automated cell.** Both are reachable only through a real launch, so the smokes exercise `stopOrphanedTree` as a function but never the `host.ts` call site, the `stopped N Jcode process(es)` diagnostic, or the `JcodeConnectorError` that replaces `(unknown)`. Their evidence is the live run on the reporting host recorded above, not a test.
- `recordLaunch` wrote the record with `mode: 0600`, which `writeFileSync` applies only when it CREATES the file, so a record loosened once kept its permissions while the code read as though every write was owner-only. Fixed in `ac070e6b5`; also found by review, not by the author.
- **The conservative branch of the new gate is untested here.** A genuinely dead host whose PID was reused by an unrelated live process now also returns `[]`, so the orphan is not reaped and jcode's own runtime-directory lock produces the refusal instead. That is the deliberate cost of resolving the ambiguous case toward signalling nothing, and it is documented in the source, but no cell asserts it. Named by the test lens.
- One mutation reds without naming a property. Removing the alias directory but skipping the symlink re-creation is caught, but as an uncaught `ENOENT` on an earlier check rather than on the reclaim assertion's own name. The mutation does not survive; the red is simply less legible than the others.

Both points were found by the review lens on the test, not by the author.

## A note on `/tmp` residue

Measured rather than assumed, because a first attempt at this measurement was confounded: the snapshot files were themselves named `/tmp/jc-*` and matched the glob being counted. Redone with the scratch outside that glob, a green `smoke:jcode-private-state` run on this branch leaves **zero** new `/tmp/jc-*` directories. The same suite with `origin/main`'s `private-state.ts` leaves one, on a run that reds partway and so never reaches its disposes.


davidfarah2003:
Relaying two review verdicts that would otherwise exist only in mesh direct messages. Both name head ac070e6b5138c7d33cb97e16789258bd8dd6f258 and both are held verbatim in the lane manager's session transcript on netcup (~/.claude/projects/-home-cotal-Projects-Cotal-AI-Cotal-1211/c8b011d4-91a4-4073-b56b-087f6f44be49.jsonl).

- rev1211-live (live-seat safety, grok-4.6): BLOCK at 15601ea96 for the fail-open in stopOrphanedTree (a live host with a mismatched start token had its tree signalled), then APPROVE at ac070e6b5 after re-testing the original input and attacking the new alive()-only gate: EPERM, kernel threads, PID 1, zombie and comm-spoof, PID reuse in a user namespace. One case carried as inconclusive: a live pid whose second /proc/stat read throws.
- rev1211-test (is the test real, glm-5.3), round 3: APPROVE at ac070e6b5. Both new cells red on the old code by their named properties, the mode cell loosens the file to 0666 before recordLaunch so it cannot pass on leftover state, 15/15 lifecycle runs stable with no leaks.

Limits the PR states openly rather than resolves: host.ts wiring and the private_state code are proven by the live run on the reporting host and graded by no automated cell; the skip-relink mutation reds without naming a property; the conservative branch's missed reap is untested by the suite.

Trigger is wider than this issue's title: a never-seated name wedged tonight after its first launch died at the provider (see #1211 comments). Merging on 17/17 green and both verdicts at the exact head.

Issue #1223, opened 2026-09-03
smoke shards are unbalanced ~2x and nothing rebalances them as the corpus grows


davidfarah2003:
## The four smoke shards are unbalanced by about 2x, and nothing keeps them balanced as the corpus grows

Measured across recent CI runs on `Cotal-AI/Cotal`, counting only jobs that ran to completion
(cancelled-early jobs are truncated runs, not durations, and are excluded):

| shard | completed | min | median | max |
|---|---|---|---|---|
| 0/4 | 4 | 25m57s | 26m12s | 26m49s |
| 1/4 | 6 | 33m34s | 36m19s | 37m12s |
| **2/4** | 5 | **39m06s** | **39m45s** | **39m56s** |
| 3/4 | 7 | 15m44s | 18m51s | 19m33s |

Plus one shard 2/4 job cancelled at 40m18s on #1220, whose true duration is therefore unknown.

Shard 2/4's maximum is **2.04x** shard 3/4's, and shard 3/4 never once exceeded shard 2/4's *minimum*.
Since the job's wall clock is the slowest shard, the run is paced by 2/4 while 3/4 sits idle for
twenty minutes. Sharding four ways buys roughly the parallelism of two.

### Why it drifts, and why no one sees it drift

`bin/smoke/ci-suites.mjs` assigns by two different rules:

- the frozen legacy list keeps `index % count`
- fragments hash: `createHash("sha256").update(suite).digest().readUInt32BE(0) % count`

Both are *stable*, which is the property they were designed for and worth keeping: `check:shard-stability`
verifies that no existing suite changes runner, and it does. Neither is *balanced*. A new fragment
lands on a shard determined by the hash of its name, so whoever adds a suite cannot tell which shard
they are lengthening, and no review signal reports that one shard just grew. The corpus is 436 legacy
entries plus 53 fragments today.

So the imbalance is not a one-off to be corrected once. Nothing measures cost, so it re-accumulates.

### Suggested direction, not a prescription

Assign by measured cost rather than by name or index: record per-suite durations from CI, and pack
shards to balance total time, keeping assignment stable between runs so `check:shard-stability` still
holds. That is a real design question (where timings live, how they refresh, what happens to a suite
with no recorded timing) and it deserves its own change rather than riding a timeout bump.

### Not included here

Raising `timeout-minutes` is a separate, smaller fix: shard 2/4's median margin against the old
[...]

PR #1220, opened 2026-09-03
feat(console): operator control, attach, membership lens, unread badges, participant mode

Lanzelot1:
The console can drive the mesh as well as watch it. Six features, one commit each:

- **Activity sparkline**: MeshView keeps a 15-bucket, 60-second message-volume series and the status bar draws it as a self-scaling sparkline next to msg/s.
- **Operator control over the endpoint rails**: `D` kills the selected agent behind a confirm (`y` graceful, `f` force), and the `:` palette adds `spawn <persona> [name]`, `status <agent>`, `ps`, and `purge` (typed-name confirm). Each action is one call through the CLI's own per-action control path (resolve, one-shot instrument, ask the manager), never the observer; `status` is located and pinned to the manager hosting the seat, `ps` merges every manager's rows. Only manager commands that already exist are driven.
- **Authoritative membership in the topology lens**: core's `readMembership`/`watchMembership` overlay silent subscribers, live and durable spokes, wide-reader badges and bounded wildcards; the header pill says live, stale, traffic-only, or unreadable with its reason, unreadable checked first.
- **Attach a live agent terminal from the console**: `a` or `:attach <agent>` runs the CLI's own attach loop in place (factored into `attachSeat`), `Ctrl-]` returns and repaints, a bad `COTAL_DETACH_KEY` is refused before the screen is handed over; the utf8-stdin detach compare is fixed for the console's reader.
- **Channel delete, unread badges, harness tag, skills**: `:delchan <channel>` runs core's `clearChannel` behind a typed-name confirm and the tab leaves the strip on the next poll; unread badges count live arrivals (never the broker's per-sender-capped retained count) with no baseline snapshot, so history is not unread and a channel that first appears later badges its first message; the roster tags each agent's harness and the detail lists runs, model, and skills.
- **Operator participant mode on an open mesh**: the operator's first send starts a second, presence-only `CotalEndpoint` under the observer's own card (same id, name, role operator), so agents can reply and the DM lens shows it; the status bar says `on roster`. No core change; under auth the send is one-way and the status line says so.

## Scope

Measured at `f0509d7b0` against the merge base `3677f9176`. The merge base is the right reference here, not `origin/main..HEAD`, because this branch has merged main and a two-dot diff misreports that as changes to files it never touched.

| area | files |
| --- | --- |
| `packages/core` | 0 |
| `implementations/cli` | 39 |
| `docs/`: `cli.md`, `mesh-view.md`, `watch-a-mesh.md` | 3 |
[...]

PR #1224, opened 2026-09-03
ci: give the smoke shards a real timeout budget

davidfarah2003:
## The smoke shards run at the wall

`timeout-minutes: 40` on the smoke job is below what shard 2/4 actually needs, so healthy runs are
being reported as failures. #1220's shard 2/4 was cancelled at 40m18s.

Measured across recent CI runs, counting only jobs that ran to completion, since a cancelled job is a
truncated run rather than a duration:

| shard | completed | min | median | max |
|---|---|---|---|---|
| 0/4 | 4 | 25m57s | 26m12s | 26m49s |
| 1/4 | 6 | 33m34s | 36m19s | 37m12s |
| **2/4** | 5 | **39m06s** | **39m45s** | **39m56s** |
| 3/4 | 7 | 15m44s | 18m51s | 19m33s |

Shard 2/4's **median** margin against the old budget was 15 seconds, and its best was 4. That is not
a tail risk to be absorbed by retrying; the typical run finishes seconds before the wall, and a
slightly slower runner loses.

## The number

60, about 1.5x the slowest observed success. It absorbs runner variance, which nobody has measured
and which moves these by minutes rather than seconds, and it leaves room for the corpus to grow
before this binds again.

Growth matters here because it is invisible at review time. Fragment suites hash to a shard by name
(`bin/smoke/ci-suites.mjs`: `createHash("sha256").update(suite).digest().readUInt32BE(0) % count`),
so whoever adds a suite cannot tell which shard they are lengthening and nothing reports that one
just grew. Sizing to "observed max plus a little" means the next fragment puts us back here.

## Why raise rather than tune

The asymmetry decides it, not the numbers. A budget below the real duration converts green runs into
false failures on gated lanes, repeatedly, and every reader then has to re-derive whether a red is
real before trusting it. A budget above it costs runner minutes only when something is genuinely
hung, and only once.

## Verification

- `smoke:shard-stability`, `smoke:workflow-concurrency`, `smoke:ci-suites`, `smoke:ci-declarations`,
  `smoke:ci-fragments`, `smoke:gate-inventory`, `smoke:shard-never-ran` all exit 0.
- `check:shard-stability 3677f9176 dcb0a5558` reports **STABLE**: 488 suites, 0 frozen legacy suites
  changing index, 0 pre-existing suites changing shard. This change moves no suite between runners.
- The workflow YAML parses.

## Caveat on the measurements

These come from 14 runs across four branches, so they are not fully independent samples, and runner
variance is unmeasured. They are enough to show that shard 2/4 sits at the wall on every completed
observation; they are not a cost model.

## Not in this PR

The four shards are unbalanced by about 2x (2/4 peaks at 39m56s against 3/4's 19m33s, and 3/4 never
exceeds 2/4's minimum), so sharding four ways buys roughly the parallelism of two. Rebalancing moves
suites between shards, which is a different change with different risk, and is filed as #1223 rather
than folded in here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,10 @@ jobs:
 
   smoke:
     name: smoke (shard ${{ matrix.shard }}/4)
-    timeout-minutes: 40
+    # Shard 2/4 runs ~39-40 min, so the previous 40 left a median margin of 15 seconds and
+    # cancelled healthy runs (#1220 died at 40m18s). Sized at ~1.5x the slowest observed
+    # success, which absorbs runner variance and lets the corpus grow before this binds again.
+    timeout-minutes: 60
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false


davidfarah2003:
## Independent review verdict, relayed

The reviewing seat was created read-only with no channel or PR-posting grant. That was my setup
choice, not a limitation on its part, so its verdict is relayed here verbatim with the location of
the durable artifact. Posting before the merge exists, so the record does not live only on one box.

**Seat:** `rev1224-budget` · **model:** `gpt-5.6-sol` · **artifact:**
`/home/cotal/mesh-logs/rev1224-budget.md` (names the graded sha five times)

> **APPROVE** `dcb0a555818b7cca12b59ea75cde29e05ed8bd40`. `COTAL_MODEL=gpt-5.6-sol`. Exact diff is
> only ci.yml 40→60 plus 3 comments. Independent GitHub API success-only shard 2 sample: 39m06s,
> 39m34s, 39m45s, 39m55s, 39m56s, median 39m45s; PR #1220 shard 2 cancelled at 40m18s. 60 is
> defensible at 1.50x observed max; I would not pick another number. All requested allowed gates and
> YAML parse passed; shard mapping STABLE 488→488, 0 moved. No blockers.

### What it did independently, rather than checking my arithmetic

It was briefed to re-derive the durations from the GitHub Actions API rather than trust the table in
the PR body, and it collected **20 runs** where the body's table came from 14. It arrived at the same
shard 2/4 values by its own collection. Two uncoordinated collections agreeing on the wall is the
evidence here; one collection checked twice would not have been.

It also weighed the number rather than accepting it, and recorded why it rejected the alternative: 50
"relies on an unmeasured 25% margin while shard membership can grow", so it would not pick
differently from 60.

### Gates it re-ran, all exit 0

`smoke:shard-stability` 7 passed · `smoke:workflow-concurrency` 27 · `smoke:ci-suites` 25 ·
`smoke:ci-declarations` 4 · `smoke:ci-fragments` 8 · `smoke:gate-inventory` 546 scripts, 526 reached,
20 exemptions · `smoke:shard-never-ran` 18 · `check:shard-stability 3677f9176 dcb0a5558` **STABLE**,
488→488 suites, 0 of 436 frozen legacy indexes changed, 0 of 488 pre-existing suites changed shard.
YAML parsed with the repository's own `yaml` dependency, timeout reading as numeric `60`. It recorded
explicitly that it ran no forbidden aggregate or live gate.

### One correction it made to the body above

The "14 runs across four branches" figure is a historical snapshot; the live 20-run window has since
advanced to 18 prior runs plus this PR's in-flight run. It ruled the limitation language
"directionally conservative rather than overstating independence", so the caveat stands as written. I
have left it rather than chase a number that changes every time a run completes.

### Merge condition

Not merged on this verdict alone. This PR's own run is the test of the change: shard 2/4 needs about
40 minutes and now has 60, so a green shard 2/4 here demonstrates the fix on itself. A failure would
be more informative than an ordinary red, since it would mean the number is wrong rather than the
change. Merging on green with main re-read in the same step.

PR #1228, opened 2026-09-03
fix(core)!: a refilling presence view must not report itself fresh

### LINKED ISSUE #1225: presence: a roster refilling after a reconnect reports itself fresh, so a not-yet-replayed peer renders as absent
## Summary

`presenceView()` returns `fresh: true` for two states that a caller must tell apart: the presence
watch is current, and the presence watch has delivered **nothing yet**. The second is when
`getRoster()` is least trustworthy — it is the window right after a reconnect, when the roster has
been wiped and is refilling entry by entry. During it, a peer whose record has not yet been replayed
is simply missing from `getRoster()`, and the freshness signal that exists to flag exactly this says
the view is good.

A caller doing everything right — read the roster, check `presenceView()`, render absence as absence
only when fresh — will render a live peer as gone.

## What is shipped (read on `packages/core/src/endpoint.ts` at main `ac0f914c0`)

`getRoster()` (`:1967`) is **synchronous**. It returns `[...this.roster.values()]` from a local `Map`
(`:460`) maintained only by the presence watch — every write is a `this.roster.set` in a watch
callback (`:4672`, `:4684`, `:4707`, `:4711`, `:4734`, `:4759`). No broker read happens in a roster
read, so a roster read cannot fail and cannot report incompleteness. There is no read to be
incomplete.

Individual rows are **never** deleted: `markOffline` (`:4730`) explicitly keeps the row and flips its
status, and the only removal anywhere is a whole-map `this.roster.clear()` (`:1127`). So "present but
offline" and "missing entirely" are produced by different paths, and only one path produces missing.

That path is `clearConnectionScoped()` (`:1088`), which runs on connection loss/rebind. It clears the
roster (`:1127`) and resets both `lastPresenceWatchAt = 0` (`:1128`) and `presenceViewFresh = true`
(`:1129`).

Then `presenceView()` (`:1979`):

```ts
if (!this.doWatch) return { fresh: true };
if (this.lastPresenceWatchAt === 0) return { fresh: true };   // <-- nothing delivered yet
const staleSince = this.lastPresenceWatchAt + this.ttlMs;
if (Date.now() < staleSince) return { fresh: true };          // <-- watch is current
return { fresh: false, staleSince };
```

Lines 2 and 4 return the same value for opposite situations. The docstring states line 2 as
deliberate — "A watch that has not yet delivered anything is not stale (there is no T to name)" —
and as a statement about staleness that is true. But callers do not consume it as a staleness
predicate; they consume it as permission to trust `getRoster()`, and for that purpose the empty
watch is the one case where trust is unwarranted.

The sweep (`:4745`) does not cover the gap either. Its whole-bucket branch (`:4751`) fires only when
*nothing* has arrived for a TTL, and the comment says why that gate exists — "Real rosters do not
lose every peer in one window" (#1045). A roster where most peers replayed and one did not is not
whole-bucket silence, so the view stays fresh and the per-peer age-out never sees the missing peer,
because there is no row to age.

## The reported instance, and what I did and did not verify

A consumer program on a live mesh reported a peer absent from successful roster reads for about 14
minutes while `/proc` showed one unbroken run, with other rows returning normally throughout. It
independently found that `liveKvEntries` is public (`packages/core/src/kv-scan.ts:110`, re-exported
`index.ts:56`) while every KV handle on `CotalEndpoint` is TS-private (`endpoint.ts:365-379`), and
read that as: a consumer wanting a completeness-guaranteed roster read has no supported route.

**I verified the API-surface half against main source** (both claims hold) **and the code paths
above. I have not reproduced the 14-minute instance.** The reconnect window described here is a
candidate explanation that fits the reported shape — successful read, populated cache, one row
missing, others normal — but it is not a confirmed cause, and a 14-minute duration is longer than a
replay window should be. Two different things may be in play.

What would confirm it: sample `presenceView()` alongside `getRoster()` at the moment a peer is
observed missing, and record whether `lastPresenceWatchAt` is 0. `fresh: true` with
`lastPresenceWatchAt === 0` and a partial roster is this defect. `fresh: true` with a current
`lastPresenceWatchAt` and a stably missing peer is a different one.

## Why the obvious fix is the wrong one

The natural reading of the API-surface finding is "expose the presence KV so a consumer can run a
completeness-guaranteed scan." That would add a second read path beside the one the display actually
uses, and the display would keep using the cache.

It also walks into a trap core has already documented. `readManagerLease` (`:3176`) deliberately
refuses to use `liveKvEntries`, and says why at `:3183-3186`: it binds a push consumer, and that
principal's grant on the bucket is `STREAM.INFO` + `STREAM.MSG.GET` + a publish only, so a consumer
there is a permissions violation for the very callers the probe serves — **"and no open-mesh test can
see that, because an open mesh has no permissions to violate."** Any scan-based presence API
validated only on an open mesh will look green and fail on an authed mesh.

## Ask

Make the freshness signal distinguish the states it currently conflates, so absence is renderable as
absence only when the view can support it. Concretely, one of:

- a third state on `presenceView()` for "the watch has delivered nothing yet" (the roster is not
  stale, it is *unpopulated*), so a caller can degrade to unknown rather than absent; and/or
- per-entry freshness rather than whole-bucket, so one lagging row is visible as lagging while the
  rest of the view stays usable.

Related, `waitForPresenceSnapshot()` (`:1989`) is the existing opt-in mitigation and has the same
shape of problem: it resolves identically whether the snapshot was consumed or the 1000ms timeout
fired, so a caller cannot tell a completed snapshot from a bounded give-up. If the fix here lands, it
should return which of the two happened.

## Not this

- Not the whole-bucket blindness gate at `:4751`. That is deliberate and correct for #1045; the point
  is that it does not cover the partial case.
- Not `readMembership()` (`:2274`). It is the right pattern one bucket over — public, uses
  `liveKvEntries`, returns an `asOf` freshness heartbeat — and is worth copying, but it reads the
  channel-membership feed, not presence.

## Related

- #443 — the general class (control-plane values that cannot distinguish "it happened" from "it did
  not"). It explicitly carves the roster out as a different defect: "The other five cannot tell. The
  roster does not know yet." This issue argues the roster has a third mode beyond those two: it does
  not know *and says it does*. Cross-reference, not a fold.
- #704 — opposite direction (a stopped agent listed as present), triaged already-fixed.
- #906 — `ps`, roster and the admission gate reporting irreconcilable counts.
- #1045 — the empty-to-full flicker the blindness gate was added for.

davidfarah2003:
Closes #1225.

`presenceView()` returned `fresh: true` for two opposite states: the presence watch is current, and
the presence watch has delivered nothing yet. The second is the window right after a reconnect, when
`clearConnectionScoped()` has wiped the roster and the watch is refilling it entry by entry. During
it, `getRoster()` returns a partial roster while the freshness signal says the view is good — so a
caller doing exactly the right thing (read the roster, check freshness, only then treat absence as
absence) renders a live peer as gone.

## What changed

`presenceView()` now returns a discriminated union:

```ts
export type PresenceView =
  | { state: "current";     fresh: true }
  | { state: "unpopulated"; fresh: false }
  | { state: "stale";       fresh: false; staleSince: number };
```

A new `presenceSnapshotPopulated` flag is cleared in `clearConnectionScoped()` and set only when the
presence watch marks the final entry of its initial replay, so `unpopulated` covers exactly the
refill window.

**What an un-updated consumer observes:** `fresh: false` during refill. Both unsafe states carry
`fresh: false`, so a consumer written before `state` existed takes its not-fresh branch and degrades
toward unknown instead of authorising an absence verdict. The union remains structurally assignable
to the previous `{ fresh: boolean; staleSince?: number }` shape, so existing typed callers still
compile.

The three in-repo consumers were updated rather than left on the fallback: `web.ts` types the event,
and `app.js` and `graph.js` each mark the roster unknown with "observer presence snapshot still
loading" — separately, because each holds its own copy of that check.

`waitForPresenceSnapshot()` now returns `"snapshot" | "timeout"` instead of `void`. It previously
resolved identically whether the snapshot completed or the 1000 ms bound fired, so a caller could not
tell a completed snapshot from a bounded give-up.

The emission tracker derives its value from `presenceView().state` rather than taking a parameter, so
the deduplication key cannot drift from the value the function actually returns.

## Proof

Two commits, test first: `add8a61d7` adds the suites, `bac1e0021` adds the fix.

`packages/core/smoke/presence-view-refill.smoke.ts` drives a real `nats-server`, four live peers and
an observer, samples `presenceView()` synchronously on every roster event during a real
`reconnect()`, and asserts the **count** of samples examined rather than only the absence of
failures. Cell 2.2 is the control that makes 2.1 meaningful: it asserts a partial roster was actually
observed, so a run that sampled nothing cannot pass as green. Cells 1.3 and 2.6 bracket a genuinely
stopped peer either side of the reconnect, proving the change does not simply make everything read as
unknown.

**Reverting `bac1e0021` and re-running reddens the suite**, verified on a separate checkout at the
same head:

```
x FAIL  1.1 CONTROL: the initial presence snapshot completes
x FAIL  2.3 every reconnect-refill sample is unpopulated and refuses a fresh absence verdict:
        {"examined":4,"unsafeSamples":[
          {"rosterSize":1,"names":["peer1"],                    "view":{"fresh":true}},
          {"rosterSize":2,"names":["peer1","peer2"],            "view":{"fresh":true}},
          {"rosterSize":3,"names":["peer1","peer2","peer3"],    "view":{"fresh":true}},
          {"rosterSize":4,"names":["peer0","peer1","peer2","peer3"],"view":{"fresh":true}}]}
x FAIL  2.4 the reconnect snapshot reports completion rather than timeout
x FAIL  2.5 the completed reconnect snapshot is current: {"fresh":true}
presence view refill smoke: 5 passed, 4 failed   (exit 1)
```

**2.3 is the cell that carries the defect** — the roster growing 1 → 2 → 3 → 4 while the view reports
`fresh: true` throughout. The other three reds are not that bug and should not be read as more
evidence for it: 1.1 and 2.4 redden because reverting restores `waitForPresenceSnapshot`'s
`Promise<void>` return, so `result !== "snapshot"`, and 2.5 reddens because the old view has no
`state: "current"`. This distinction was raised by the independent review; the earlier wording here
let all four reds imply the freshness defect.

Gates on a clean checkout at `bac1e0021`: `smoke:presence-view-refill` 9/9, `smoke:web-presence-view`
15/15, `pnpm typecheck` exit 0 with zero `error TS`, `pnpm build` exit 0.

## Scope

Per-entry freshness is **not** in this change. It addresses a separate reported instance — a peer
absent from successful roster reads for about fourteen minutes on a live mesh — which has not been
reproduced. That stays open on #1225.

Exposing the presence KV for a completeness-guaranteed scan was considered and rejected:
`getRoster()` is synchronous over a local `Map`, so there is no broker read to make complete, and
`readManagerLease` already documents why binding a push consumer on a bucket is a permissions
violation for callers holding only `STREAM.INFO` + `STREAM.MSG.GET` — a hazard no open-mesh test can
detect.

`!doWatch` continues to report `state: "current"`. That branch's value is unchanged from before, and
revisiting what it should claim for an endpoint whose roster is empty by construction is left out of
scope deliberately rather than by omission.

## Two follow-up commits, after CI

The first push was red on two checks, both caused by this PR and neither a defect in the fix:

- `smoke (shard 1/4)` failed at `smoke:gate-inventory`: the new suite was in `package.json` but nothing
  in the CI corpus ran it. `a2eb189e2` registers it as its own fragment,
  `bin/smoke/ci-suites.d/a2bf66cc…5155.txt`, per that directory's README — the frozen positional list
  is not appended to. `check:shard-stability ac0f914c0 1777ac9cf` reports STABLE: 488 → 489 suites,
  0 of 436 frozen legacy suites change index, 0 of 488 pre-existing suites change runner.
  Removing the fragment again reproduces the CI failure verbatim and `smoke:gate-inventory` exits 1,
  so the fragment is what fixes it.
- `unit` failed at `check:docsbundle`: `docs/presence-and-delivery.md` changed without a regenerated
  bundle. `1777ac9cf` regenerates it. The diff is a pure insertion of the paragraph that docs change
  added — nothing removed, no unrelated drift from main swept in.

Gates at `1777ac9cf`: `check:docsbundle` 0, `smoke:ci-fragments` 8/8, `smoke:gate-inventory` 0,
`check:shard-stability` STABLE, `pnpm typecheck` 0 with zero `error TS`,
`smoke:presence-view-refill` 9/9, `smoke:web-presence-view` 15/15.

## Third follow-up commit: the fixture the rename broke

The independent review blocked this PR on a defect it created and I had missed: the fix renamed
`setPresenceViewFresh`, and `packages/core/smoke/mutations/presence-watch-stall.json` was the only
place in the tree where that name still occurred. Both its anchors were dead — zero occurrences — so
`smoke:mutation-fixtures` (in the CI corpus, `bin/smoke/ci-suites.txt:264` by position) refused the
tree, and CI shard 0/4 failed there.

The behaviour was never broken: `smoke:presence-watch-stall` passes 12/12 throughout. What died is
the proof that those two cells are load-bearing.

`0c7f06ae4` re-anchors both mutations onto the renamed code. They are unchanged in what they do — S1
still deletes the sweep freshness gate, S2 still drops the stale-replay early-return while keeping
the view emission. They were re-anchored, not relabelled to whatever went red:

- `smoke:mutation-fixtures`: `dead anchors: 2`, exit 1 → `dead anchors: 0`, exit 0.
- `mutation-proof` on that config: both mutants KILLED on their existing named cells, 1.4 and 1.9.
- The red *sets* were checked by hand rather than taken from the tool's summary. The suite is not
  fail-fast (`presence-watch-stall.smoke.ts:227` exits after all cells run), so every red is visible:
  S1 reds 1.4, 1.5, 1.7, 1.9 and its named 1.4 is the earliest; S2 reds only its named 1.9.

## Follow-up filed, not fixed here

#1229. `connector-core` reads the roster through `getRoster()` at three sites and never consults
`presenceView()`, so `cotal_roster`, `resolvePeer` and `assertKnownMentions` can still treat a partial
reconnect roster as complete. That is the same defect class as #1225 in a consumer that never checked
the signal, so this PR's change does not reach it. Pre-existing, and arguably the more agent-visible
instance. Raised by the independent review of this PR.


davidfarah2003:
CI at `bac1e0021` (run 33757119590) is red in two places, and both are the PR's own to fix; neither is a known flake.

1. `unit` fails at the `pnpm check:docsbundle` step: `docs/presence-and-delivery.md` changed but `extensions/connector-core/src/docs-bundle.generated.ts` was not regenerated, so `git diff --exit-code` on the bundle is non-zero (the log shows the one-line body change under the "Message flow" page). Run `pnpm gen:docsbundle` and commit the regenerated file with the docs change.

2. `smoke (shard 1/4)` fails at `pnpm smoke:gate-inventory`: `smoke:presence-view-refill` exists in `package.json` but nothing runs it and it is not in `UNGATED`. Per `bin/smoke/ci-suites.d/README.md`, add a fragment file `bin/smoke/ci-suites.d/<sha256 of the script name>.txt` containing exactly `smoke:presence-view-refill`; do not append to `ci-suites.txt`. Then `pnpm smoke:ci-fragments`, `pnpm smoke:gate-inventory`, and `pnpm check:shard-stability ac0f914c0 <new-head>`.

Both fixes move the head, so the exact-head review commissioned at `bac1e0021` retires; the reviewer regrades at the new head. `live` is green; shards 0, 2 and 3 were still running when this was read.

davidfarah2003:
## Independent review record

Reviewed at `1777ac9cf8bd89a956e5f0a0eeb97e66349361c2`, blocked, repaired, and re-graded at the
current head.

**Round 1 — `1777ac9cf` — BLOCK, one finding.** This PR renamed `setPresenceViewFresh`, and
`packages/core/smoke/mutations/presence-watch-stall.json` was the only place in the tree where that
name still occurred. Both its anchors were dead, so `smoke:mutation-fixtures` — which is in the CI
corpus — refused the tree, and CI shard 0/4 failed there. The stall behaviour was never broken
(`smoke:presence-watch-stall` passes 12/12 throughout); what died is the proof those two cells are
load-bearing. An off-branch repair existed at the time and was correctly refused as not being on the
graded sha.

What else round 1 verified independently rather than taking from this description:

- The revert-proof re-derived from scratch: same 4-cell red set, same 2.3 payload.
- Cells 2.1 and 2.2 were shown able to *fail*, by constructing an empty sampler (2.1 fails,
  `examined=0`) and a sample-after-snapshot run (2.1 passes, 2.2 fails, `partial=0`). The count
  assertion and its control are both load-bearing rather than decorative.
- The pre-change `presenceView` was reconstructed and run beside the shipped one over a real
  `nats-server` across a reconnect. The diff is **not** empty: at roster sizes 1, 2, 3, 4 the old
  code reported `fresh: true` while the new code reports `unpopulated` / `fresh: false`. An
  un-updated `if (view.fresh)` consumer authorises an absence verdict on the old code and degrades
  to unknown on the new one.
- The compatibility claim: an old-shape consumer `{ fresh: boolean; staleSince?: number }` compiles.
  Separately, `return`ing the `waitForPresenceSnapshot()` promise as `Promise<void>` does not
  (TS2322); `await` callers, which is what the in-repo CLI and manager callers do, compile fine. The
  title already carries `!`.
- A repo-wide consumer sweep with a positive control on the search.

**Round 2 — `0c7f06ae4` — APPROVE.** Fast-forward from the graded sha, parent verified, delta is one
file at +3/-3. `smoke:mutation-fixtures` exit 0, 0 dead anchors. Both mutants killed. The red *sets*
were re-derived by hand rather than read off the tool's summary — S1 reds 1.4, 1.5, 1.7, 1.9 with its
named 1.4 the earliest; S2 reds only its named 1.9 — and each restore was proven with a checksum of
`endpoint.ts` between mutations. No open blocking finding stands against this head.

## Non-blocking findings, recorded rather than silently dropped

- **#1229** — `connector-core` reads the roster at three sites without consulting `presenceView()`,
  so `cotal_roster`, `resolvePeer` and `assertKnownMentions` can still treat a partial reconnect
  roster as complete. Pre-existing rather than introduced here, since those call sites never
  consulted the signal; filed separately for that reason. Arguably the more agent-visible instance of
  the same defect class.
- `waitForPresenceSnapshot`'s return type is a break in `return` position but not for `await`
  callers, as described above.
- The CLI and manager callers still ignore `snapshot` vs `timeout`.
- `!doWatch` still reports `current`: an endpoint that never watches still claims a complete absence
  verdict over an empty-by-construction roster. Out of scope here, carried on #1229.

## Correction to this PR's own description

An earlier revision of the body let all four reverted cells imply the freshness defect. Only **2.3**
carries it. 1.1 and 2.4 redden because reverting restores `waitForPresenceSnapshot`'s `Promise<void>`
return, and 2.5 because the old view has no `state: "current"`. The body was corrected.

davidfarah2003:
## The approval above is retired by a new head — this PR is not currently mergeable

The review record posted earlier grades **`0c7f06ae4`**. The head is now
**`23f80e177068b1f359c9c61ddb87b3fe954ccdfa`**, pushed after that verdict, so under the exact-head
rule the attestation no longer binds. Approved-plus-stale reads as approved to anyone who scrolls to
the newest verdict and stops, which is why this note exists rather than being left implicit.

**Why the head moved.** CI shard 0/4 failed at `0c7f06ae4`, and not on anything the earlier review
covered:

```
[reaper] ✗ pnpm smoke:presence-view-refill LEAKED 1 broker(s) it had already been asked to own
✗ shard 0/4: 1 suite(s) passed but LEAKED a broker they owned
  A green suite that leaves a broker running is a false green.
```

The suite added by this PR passed 9/9 and left its `nats-server` running. Cause:
`teardownOnSignal()` returns a function that only drops ownership — `return () => owned.delete(entry)`
— and never kills anything, so the broker was neither killed by the suite nor still owned by the
signal handler. `packages/smoke-kit/src/broker-teardown.ts` names this as its "defect 1, no
normal-path teardown", and warns that adopting the helper in that state is "a rename, not a fix".
Awaiting the `void` return made it read like async teardown.

Reproduced before the fix — one leaked broker and one leaked store dir per run — and measured after:
zero of each, suite still 9/9.

**The delta.** `23f80e177` is one commit, one file, +11/−3, confined to that suite's `finally`:
kill-and-await, remove the tree, release ownership last, matching `bind-fence.smoke.ts`. It touches
no shipped source (`packages/core/src`, `implementations/` → 0 files) and nothing the earlier round's
blocking finding covered (`packages/core/smoke/mutations/` → 0 files).

**A fresh independent review is in progress at `23f80e177`.** It is not a delta re-grade carried over
from the earlier verdict: that reviewer's round was spent, and its conclusions are handed to the new
one as context explicitly marked as unverified rather than as clearance. The two questions it was
asked to attack first are the ones this delta creates — whether the suite still reddens on revert now
that its teardown changed, since that commit edits the file carrying this PR's regression proof, and
whether it genuinely stops leaking, measured rather than read.

Nothing should merge here until a verdict names `23f80e177` and CI is green on that sha.

davidfarah2003:
## Independent review at `23f80e177068b1f359c9c61ddb87b3fe954ccdfa` — APPROVE

A fresh review of the current head, seated after the previous approval was retired. It inherited no
conclusions from the earlier round; that verdict was handed over marked as unverified context rather
than as clearance.

**The controlled comparison that settles the teardown fix.** Rather than measuring this head before
and after, it ran the *parent's* suite file against the *current* implementation: 9/9 green **and
leaking** — a new `nats-server` plus its store dir, owner pid dead. At this head, nothing leaks. That
isolates the teardown commit as the cause rather than merely correlating with it. It reaped only the
broker its own run created, after confirming the owning pid was dead.

**The regression proof survived the teardown edit.** This was the risk that made the one-file delta
non-trivial: that commit edits the file carrying the PR's entire proof. Reverting only the fix commit
`bac1e0021`, with the teardown commit kept, still reddens:

- 1.1 CONTROL: the initial presence snapshot completes
- 2.3 every reconnect-refill sample is unpopulated and refuses a fresh absence verdict —
  `{"examined":4,"unsafeSamples":[{"rosterSize":1,…,"view":{"fresh":true}}, … {"rosterSize":4,…}]}`
- 2.4 the reconnect snapshot reports completion rather than timeout
- 2.5 the completed reconnect snapshot is current

5 passed, 4 failed; restore returns it to 9/9. **2.3 is the cell carrying the defect** — the roster
growing 1 → 4 while the view reports `fresh: true`. 1.1 and 2.4 redden because the revert restores
`waitForPresenceSnapshot`'s `Promise<void>`, and 2.5 because the old view has no `state: "current"`.

**Every control was shown able to fail**, by construction rather than by argument: an empty sampler
fails 2.1 (`examined=0`) and 2.2; a complete-only sample set padded to `PEER_COUNT` passes 2.1 and
fails 2.2 (`partial=0`); skipping the peer stop fails both 1.3 and 2.6. So the count assertion and
its controls are load-bearing rather than decorative.

**Compatibility, measured not asserted.** A live probe of an old-shape `if (view.fresh)` consumer
against real refill samples returned "unknown" on every sample — it degrades rather than authorising
an absence verdict. `PresenceView` is structurally assignable to the old
`{ fresh: boolean; staleSince?: number }` (tsc exit 0). Separately, `return`ing the
`waitForPresenceSnapshot()` promise as `Promise<void>` is TS2322; `await` callers, which is what the
in-repo CLI and manager callers do, compile. The title already carries `!`.

**Gates at this head, exit codes read directly:** `typecheck` 0 with zero `error TS`;
`smoke:presence-view-refill` 9/9; `smoke:web-presence-view` 15/15; `smoke:presence-watch-stall`
12/12; `smoke:mutation-fixtures` 0 with `dead anchors: 0`; `check:docs-voice` 0 across 40 pages;
`changeset status` 0. `mutation-proof` on the stall config: both mutants killed on their named cells,
with the red sets hand-derived rather than read from the tool summary (S1 reds 1.4, 1.5, 1.7, 1.9;
S2 reds only 1.9) and `endpoint.ts` checksummed after each restore.

**Non-blocking residuals, not reopened:** #1229 (`connector-core` never consults `presenceView`); the
CLI and manager still ignoring `snapshot` vs `timeout`; `!doWatch` still reporting `current`.

Merging waits on the four smoke shards at this sha.

Issue #1231, opened 2026-09-03
multiChannelHistory: two limits survive the batching fix, and their failures are indistinguishable to a caller


davidfarah2003:
## What

`CotalEndpoint.multiChannelHistory` reads the CHAT stream through a multi-filter consumer. Since
#1212 the filter list is batched at `MULTI_FILTER_BATCH` = 1,000 with at most
`MULTI_FILTER_READ_CONCURRENCY` = 4 batches in flight, which removed the consumer-create size
ceiling. Two limits survive that fix, and this issue is about those.

## Retraction first, because this issue used to say the opposite

An earlier draft of this issue said: "Nobody has produced a timeout, and nobody has established that
one exists at any particular count. Treat the number 10,000 as a shape rather than a threshold."

That is withdrawn by measurement. The timeout was produced, the ceiling is real, and 10,000 was
about right. On a seeded corpus of 10,000 channels the unbatched create answered at 70 channels in
43ms, at 1,000 in 246ms and at 5,000 in 5,345ms, then timed out at 10,000. What gives way is the
client's request timeout rather than the broker, which is why it presents as a timeout and not as a
refusal. With batching, the same four points are 54ms, 343ms, 1,163ms and 942ms. That work landed in
#1212 rather than here, so this issue is no longer about whether a ceiling exists.

## The two limits that remain

**A batch matching a small fraction of a long stream is still expensive.** On a larger corpus of
20,000 channels and 20,000 messages the batched read fails at 5,000 filters and above. Batching
removed the create size ceiling; it did not remove the cost of a batch whose subjects match a small
fraction of the sequence space it has to walk back through. No unbatched numbers were taken on that
corpus, so no comparative claim is made about it, only that the batched read fails there.

**A filter list around 40,000 exceeds `max_payload`.** The broker refuses it rather than timing out.
That is a different failure with a different shape, and a caller cannot currently tell the two
apart.

## What the first work is

A measurement again, not a fix. The 20,000-channel failure needs its cause separated: whether the
cost is in the backward walk, in the per-batch consumer lifecycle, or in the fraction of matching
messages. Picking a remedy before that is picking one blind, and `MULTI_FILTER_BATCH` is the obvious
lever precisely because it is the one already in view, which is a poor reason to reach for it.

## Refusal shape

This matters as much as either limit. Neither failure should present as an empty page or a silent
partial. The route already has a declared-partial envelope that names which source is missing, and
[...]

PR #1235, opened 2026-09-03
docs(changeset): describe the multi-filter batching that shipped with the single read

davidfarah2003:
## What this is

Markdown only. One file, 20 insertions, 0 deletions, no frontmatter change, so the version bumps
this changeset declares are untouched.

`.changeset/activity-single-read.md` on main describes the one-stream read and its measurement, and
says nothing about the batching that shipped beside it. Both constants are live on main:

```
packages/core/src/endpoint.ts:169  export const MULTI_FILTER_BATCH = 1_000;
packages/core/src/endpoint.ts:174  export const MULTI_FILTER_READ_CONCURRENCY = 4;
packages/core/src/endpoint.ts:2569 const size = Math.max(1, Math.trunc(opts?.batch ?? MULTI_FILTER_BATCH));
packages/core/src/endpoint.ts:2590 Array.from({ length: Math.min(MULTI_FILTER_READ_CONCURRENCY, batches.length) }, readBatch),
```

They are **exported**, and the changeset declares `@cotal-ai/core: minor`. So the release ships new
API surface with no CHANGELOG line describing it. That is the reason this is worth a PR rather than
being left thin.

## Why now

Release PR #1221 is open and was last updated at 17:58:22Z, so the changeset is still consumable.
Once #1221 merges, the changeset file is deleted and the CHANGELOG is written from what it said at
that moment. Editing the generated CHANGELOG inside #1221 would not survive the next regeneration;
the changeset file on main is the durable place for this.

## What the paragraph says

The create ceiling from a seeded sweep (succeeded at 70, 1,000 and 5,000 channels, failed at
10,000, and what gives way is the client request timeout rather than the broker); the batch size and
that it is derived from the sweep rather than from the failure point; the before and after sweep
numbers; that a 69 channel space is a single batch so every figure at that size is unchanged; the
sequence-merge assertion that covers it; and the two limits that survive, stated rather than buried.

## Verification

Claims checked against main rather than taken from the draft:

- Both constants exist at the stated values and call sites, quoted above.
- The cited assertion exists: `packages/core/smoke/history-recent.smoke.ts:277`,
  `check("multiChannelHistory merges BATCHES by arrival, not by the highest \`ts\`", …)`, with the
  corpus deliberately built so arrival order and `ts` order disagree (see the comment at line 287,
  which says that without it a corpus whose `ts` matched arrival would pass for free).
- Diff is one file, 20 insertions, 0 deletions. No line removed, so nothing existing is reworded.

The measurement figures in the paragraph come from the same instrumented runs as the rest of the
changeset and are not re-derived here.

## Not claimed

This does not change behaviour, tests or versions. It does not restate the two remaining limits as
fixed; they are recorded as open, and #1231 tracks them.

### DIFF
--- a/.changeset/activity-single-read.md
+++ b/.changeset/activity-single-read.md
@@ -21,7 +21,9 @@ timeout rather than the broker. `MULTI_FILTER_BATCH` is 1,000 and `MULTI_FILTER_
 sequence before the newest N are taken. The size comes from the sweep rather than from the failure
 point: a fifth of the largest count that still answered, answering in about a quarter of a second, so
 a batch stays far from both the create timeout and the 8000ms response deadline. On the same corpus
-the sweep goes from 43ms, 246ms, 5345ms and a timeout, to 54ms, 343ms, 1163ms and 942ms. A space
+the sweep goes from 43ms, 246ms, 5345ms and a timeout, to 54ms, 343ms, 1163ms and 942ms. Those walls
+are single runs on one host and are shaped by it rather than being bounds; a re-run on another box
+landed outside the last of them, as the paragraph below says of every span here. A space
 with the 69 chat channels this work was aimed at is one batch, so every figure above is unchanged at
 that size. Merging on stream sequence rather than the payload's `ts` is asserted in
 `packages/core/smoke/history-recent.smoke.ts` against a corpus whose arrival order and `ts` order are
@@ -30,8 +32,13 @@ opposite.
 Two limits remain and are stated rather than buried. On a 20,000 channel and 20,000 message corpus
 the batched read still fails at 5,000 filters and above, because batching removes the create size
 ceiling and not the cost of a batch matching a small fraction of a long stream; no unbatched numbers
-were taken on that corpus, so no comparative claim is made about it. Separately, a filter list around
-40,000 exceeds `max_payload` and is refused by the broker rather than timing out.
+were taken on that corpus, so no comparative claim is made about it. Separately, a filter list
+big enough to exceed the connection's `max_payload` is refused by the client before anything is
+sent, rather than by the broker and rather than timing out: the client compares the request against
+the limit the broker advertised at connect and throws, so no bytes leave and the broker's own logs
+show nothing. On this corpus's subject shape 40,000 filters serialize to 1,480,032 bytes against an
+advertised 1,048,576. That threshold moves with both the advertised limit and the length of the
+subjects, so the count is an observation on one shape rather than a bound.
 
 Measured on the wire by a proxy between the endpoint and the broker that counts `$JS.API` request
 lines and bytes per direction, over a seeded corpus of 69 chat channels plus 24 event channels at
--- a/.changeset/activity-single-read.md
+++ b/.changeset/activity-single-read.md
@@ -13,6 +13,26 @@ decides the read's filter set instead of deciding which of seventy reads to issu
 authority: a multi-filter create rides the bare `$JS.API.CONSUMER.CREATE.<CHAT>` row the observer
 and admin profiles already hold.
 
+The multi-filter read batches its filter list. One consumer create naming every subject is a single
+request, but a large enough request stops being answerable: on a seeded sweep the create succeeded
+at 70, 1,000 and 5,000 channels and failed at 10,000, and what gives way is the client's request
+timeout rather than the broker. `MULTI_FILTER_BATCH` is 1,000 and `MULTI_FILTER_READ_CONCURRENCY` is
+4, so the list is split and at most four batches are in flight, and the pages are merged by stream
+sequence before the newest N are taken. The size comes from the sweep rather than from the failure
+point: a fifth of the largest count that still answered, answering in about a quarter of a second, so
+a batch stays far from both the create timeout and the 8000ms response deadline. On the same corpus
+the sweep goes from 43ms, 246ms, 5345ms and a timeout, to 54ms, 343ms, 1163ms and 942ms. A space
+with the 69 chat channels this work was aimed at is one batch, so every figure above is unchanged at
+that size. Merging on stream sequence rather than the payload's `ts` is asserted in
+`packages/core/smoke/history-recent.smoke.ts` against a corpus whose arrival order and `ts` order are
+opposite.
+
+Two limits remain and are stated rather than buried. On a 20,000 channel and 20,000 message corpus
+the batched read still fails at 5,000 filters and above, because batching removes the create size
+ceiling and not the cost of a batch matching a small fraction of a long stream; no unbatched numbers
+were taken on that corpus, so no comparative claim is made about it. Separately, a filter list around
+40,000 exceeds `max_payload` and is refused by the broker rather than timing out.
+
 Measured on the wire by a proxy between the endpoint and the broker that counts `$JS.API` request
 lines and bytes per direction, over a seeded corpus of 69 chat channels plus 24 event channels at
 limit 200. The before column is that same committed suite file, with its `package.json` script,
--- a/.changeset/activity-single-read.md
+++ b/.changeset/activity-single-read.md
@@ -32,14 +32,16 @@ opposite.
 
 Two limits remain and are stated rather than buried. On a 20,000 channel and 20,000 message corpus
 the batched read still fails at 5,000 filters and above, because batching removes the create size
-ceiling and not the cost of a batch matching a small fraction of a long stream; no unbatched numbers
-were taken on that corpus, so no comparative claim is made about it. Separately, a filter list
-big enough to exceed the connection's `max_payload` is refused by the client before anything is
-sent, rather than by the broker and rather than timing out: the client compares the request against
-the limit the broker advertised at connect and throws, so no bytes leave and the broker's own logs
-show nothing. On this corpus's subject shape 40,000 filters serialize to 1,480,032 bytes against an
-advertised 1,048,576. That threshold moves with both the advertised limit and the length of the
-subjects, so the count is an observation on one shape rather than a bound.
+ceiling and not the cost of a batch matching a small fraction of a long stream; no unbatched
+numbers were taken on that corpus, so no comparative claim is made about it. Separately, a filter
+list big enough to exceed the connection's `max_payload` is refused by the client before anything
+is sent, rather than by the broker and rather than timing out: the client compares the request
+against the limit the broker advertised at connect and throws, so no bytes leave and the broker's
+own logs show nothing. On this corpus's subject shape the filter list alone serializes to 1,480,032
+bytes, already past an advertised 1,048,576 before the rest of the request is counted, so a list
+that long cannot be sent whatever the envelope adds. That threshold moves with both the advertised
+limit and the length of the subjects, so the count is an observation on one shape rather than a
+bound.
 
 Measured on the wire by a proxy between the endpoint and the broker that counts `$JS.API` request
 lines and bytes per direction, over a seeded corpus of 69 chat channels plus 24 event channels at
--- a/.changeset/activity-single-read.md
+++ b/.changeset/activity-single-read.md
@@ -25,7 +25,7 @@ the sweep goes from 43ms, 246ms, 5345ms and a timeout, to 54ms, 343ms, 1163ms an
 are single runs on one host and are shaped by it rather than being bounds; a re-run on another box
 landed outside the last of them, as the paragraph below says of every span here. A space
 with the 69 chat channels this work was aimed at is one batch, so the request counts and byte totals
-above are unchanged at that size. The walls are not claimed to be, for the reason just given.
+are unchanged at that size. The walls are not claimed to be, for the reason just given.
 Merging on stream sequence rather than the payload's `ts` is asserted in
 `packages/core/smoke/history-recent.smoke.ts` against a corpus whose arrival order and `ts` order are
 opposite.
@@ -37,11 +37,9 @@ numbers were taken on that corpus, so no comparative claim is made about it. Sep
 list big enough to exceed the connection's `max_payload` is refused by the client before anything
 is sent, rather than by the broker and rather than timing out: the client compares the request
 against the limit the broker advertised at connect and throws, so no bytes leave and the broker's
-own logs show nothing. On this corpus's subject shape the filter list alone serializes to 1,480,032
-bytes, already past an advertised 1,048,576 before the rest of the request is counted, so a list
-that long cannot be sent whatever the envelope adds. That threshold moves with both the advertised
-limit and the length of the subjects, so the count is an observation on one shape rather than a
-bound.
+own logs show nothing. A list of 40,000 filters on this corpus's subject shape is refused that way.
+Where the limit falls moves with both the advertised `max_payload` and the length of the subjects,
+so 40,000 is a length that was tried and not a threshold.
 
 Measured on the wire by a proxy between the endpoint and the broker that counts `$JS.API` request
 lines and bytes per direction, over a seeded corpus of 69 chat channels plus 24 event channels at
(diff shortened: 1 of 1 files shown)


davidfarah2003:
> [!IMPORTANT]
> **RETIRED — this verdict is bound to `65414572fab171d18f00440a60cce425acfb5364` and the head has moved to `01ab762024d09fbe8606c359a858e4b0ffbcb7d5`.**
> It is kept below as the record of what was graded, and it no longer describes this PR. Nothing here should be read as an open finding against the current head. See the comment below for the state after the push.
>
> **Also correct one figure below before reusing it.** This comment says 40,000 filters "serialize to 1,480,032 bytes". That is the serialized *filter list* — `JSON.stringify({config:{filter_subjects}})` — not the request the client compares against `max_payload`; the real request for this shape measured 1,480,265 bytes. The conclusion is unchanged (both clear the advertised 1,048,576 by about 1.41x), but do not quote 1,480,032 as a request size.

---

## Independent review at `65414572fab171d18f00440a60cce425acfb5364` — BLOCK

Reviewed as a permanent customer-facing record rather than as a markdown diff: a changeset is the
source the release bot generates the published CHANGELOG from, so its numbers outlive the PR and
nobody re-derives them later. The review was asked for a **per-claim** verdict on which figures are
confirmable from this repository and which are not.

### Confirmed against the tree

| claim | where |
|---|---|
| `MULTI_FILTER_BATCH` is 1,000 | `packages/core/src/endpoint.ts:169` |
| `MULTI_FILTER_READ_CONCURRENCY` is 4, at most four in flight | `:174`, enforced at `:2590` |
| filter list split, pages merged by stream sequence | `:2569`-`:2597` |
| the 8000 ms response deadline | `AGGREGATION_DEADLINE_MS` in `implementations/web/src/web.ts:365` |

Frontmatter is byte-identical to the parent (core minor, web minor), and `changeset status` produces
identical output with the old and new prose — so the declared bumps are genuinely untouched.

The paragraph's one citation survived independent re-derivation. `history-recent.smoke.ts` publishes
`late` first with a future `ts`, then `early` with an older one, and the forced multi-batch read at
limit 1 must return `early`; a control asserts the `ts` order and arrival order are opposite, so the
cell cannot pass for free on a corpus where they happen to agree. `smoke:history-recent` 26/26,
`smoke:web-activity-read-cost` 44/44.

### Blocking

**1. `every figure above is unchanged` claims more than the evidence supports** (line 25). What is
established is that a 69-channel space remains one batch, and that the prior **round-trip and byte**
figures are unchanged. Unchanged wall times, and "every figure", are not established. Narrowing it to
the evidenced round-trip/byte claim resolves this.

**2. The `~40,000 exceeds max_payload` sentence attributes the refusal to the wrong actor, and the
attribution is disproved by measurement** (line 34). *This finding was strengthened after the initial
verdict: it was first reported as untraceable, and was then measured.*

Against this tree: 40,000 short `actcost-wide` filter subjects serialize to **1,480,032 bytes**; a real
default `nats-server` advertises `max_payload=1,048,576`; forcing one 40,000-filter batch raises
`InvalidArgumentError: 'payload' max_payload size exceeded` with **zero output payload bytes**. The
pinned client checks the broker-advertised limit and throws before `sendCommand`
(`@nats-io/nats-core/lib/protocol.js:835-848`).

So **"refused by the broker" is false for the shipped client** — nothing reaches the broker. The client
refuses locally, using a limit the broker advertised. Separately, "around 40,000" holds only for a
particular broker configuration and subject length, both of which move the threshold, so a bare filter
count is under-scoped as a general limit.

This is not a labelling or scoping fix. The sentence as written states something untrue about where
the refusal happens.

**3. The historical sweep values have no committed reproducer.** `54ms / 343ms / 1163ms / 942ms`
appear only in commit `cd2dea085`'s message and in this changeset; the committed smoke regenerates
only the 10,000-filter case, and running it during this review returned **966 ms, not 942 ms**. The
20,000-channel failure at 5,000 filters is likewise commit prose with no harness. The 24 ms is not the
point — the point is that these are wall times from one machine on one day, and the changeset presents
them without that qualification.

### Restraint check

The 20,000-corpus paragraph honours its own stated restraint: it takes no unbatched numbers on that
corpus and makes no comparative claim about it. The "every figure above is unchanged" sentence does
not hold to the equivalent standard, which is why finding 1 stands while that paragraph does not.

### An incidental gap worth recording

`pnpm check:docs-voice` passes 40 pages but scans only `docs/**/*.md`, so **it does not cover
`.changeset` at all**. Changeset prose ships to customers through the generated CHANGELOG and no voice
gate touches it. The added lines were scanned by hand instead and contain no em dash, banned filler,
list-style heading, slogan-shaped contrast, or tool attribution.

### Note on the standard applied

The per-claim traceability bar in finding 3 was set by the review brief, not found in this repo's
existing conventions — a changelog recording a development measurement is not obviously disallowed.
That finding is therefore open to being answered by attributing the numbers as measured-then rather
than by building harnesses for them. Findings 1 and 2 do not have that character: they are an
overclaim and an unscoped untraceable limit respectively.

davidfarah2003:
## New head `01ab762024d09fbe8606c359a858e4b0ffbcb7d5` — the BLOCK's finding 2 is addressed, and the grade at `65414572f` no longer binds

Pushed in response to the review. Stating it here in the same minute as the push rather than
leaving the PR to read as though the earlier grade still applies.

### Finding 2 was correct and my sentence was false

The sentence said an oversized filter list is "refused by the broker rather than timing out."
It is not refused by the broker. Nothing reaches the broker.

Verified against the pinned client in this tree rather than taken from the review,
`@nats-io/nats-core@3.4.0/lib/protocol.js`:

```js
if (this.info && len > this.info.max_payload) {
    throw errors_1.InvalidArgumentError.format("payload", "max_payload size exceeded");
}
this.outBytes += len;
this.outMsgs++;
...
this.sendCommand(proto, headers, data, CRLF);
```

The throw sits **before** `outBytes += len` and **before** `sendCommand`, and compares against
`this.info.max_payload`, which is the limit the broker advertised at connect. So the client refuses
locally, no bytes leave, and the counter is never incremented. That matches the review's measured
"zero output payload bytes."

This is worth the words rather than being pedantry: a reader debugging this from the broker's logs
would find nothing in them, because nothing arrived.

The bare `40,000` was separately under-scoped, stating a single-environment observation as a general
limit. It now carries the serialized size and the advertised limit it was measured against, and says
the threshold moves with both the advertised limit and the subject length.

### Finding 1, pre-empted rather than argued

Finding 1 was non-blocking and stands: an independent re-run put the last sweep wall at 966ms
against the published 942ms, so the number is host-bound. The paragraph now says the walls are
single runs on one host shaped by it rather than bounds, and that a re-run elsewhere landed outside
the last of them. That matches what this changeset already says about every other span it publishes,
so the added prose now meets the standard the surrounding text set rather than a weaker one.

I have not restated the re-run's own figure, because I did not measure it.

### Finding 3

Recorded as advisory by the reviewer's commissioner, who noted the rule it came from was stricter
than repo convention and imposed by the brief. Not addressed here.

### The diff

Still one file and purely additive against `main`: `.changeset/activity-single-read.md`, +27/-0.
Frontmatter untouched, so the declared bumps and `changeset status` are unchanged. No behaviour,
tests or versions are touched.

### Ordering, unchanged

This must land before #1221 (`changeset-release/main`), whose diff **deletes**
`.changeset/activity-single-read.md`. After that merge the changeset no longer exists and this prose
can never ship.

davidfarah2003:
**BLOCK at `01ab762024d09fbe8606c359a858e4b0ffbcb7d5`** — one finding. Everything else in the delta is confirmed by measurement.

## The finding

Line 39 reads "On this corpus's subject shape 40,000 filters serialize to 1,480,032 bytes against an advertised 1,048,576", and the sentence before it describes the client comparing *the request* against `max_payload`. So 1,480,032 reads as the request size. It is not the request size.

The real request for this shape measured **1,480,265 bytes** on an instrumented `nc.request` at this head. 1,480,032 is exactly `JSON.stringify({config:{filter_subjects: filters}})` — the serialized filter list — and omits 233 bytes of request envelope and generated consumer config (`name`, `deliver_policy`, `ack_policy`, `inactive_threshold`, `num_replicas`, `max_deliver`, `mem_storage`, `stream_name`, `action`, `pedantic`).

The decomposition is checkable without a broker, and was checked independently. `JSON.stringify({config:{filter_subjects:[...N]}})` has length `30 + N*(len+2) + (N-1) + 3`, so at N=40,000 it is `40000*len + 120032`. That equals 1,480,032 only at `len = 34`, exactly — and 34 is the real subject width, since `chatSubject` is `${spacePrefix(space)}.chat.${ownerToken(owner)}.${ownerToken(actor)}.${channelPath(channel)}` and `cotal.actcost-wide.chat.*.*.w00000` is 34 characters. A full request payload would not land on that clean integer form.

**The conclusion does not move** — both figures clear the advertised 1,048,576 by about 1.41x. This blocks because a seven-digit exact figure carries the authority of a precise measurement while measuring a different quantity than the surrounding sentence claims, and this is a CHANGELOG entry that outlives the PR.

Either remedy closes it: publish 1,480,265 as the captured request size for this run and subject shape, or keep 1,480,032 and say plainly that it is the serialized filter list rather than the request.

## Confirmed in this round

- A live default broker advertised `max_payload=1,048,576`.
- The pinned client raised `InvalidArgumentError: 'payload' max_payload size exceeded` before output payload bytes increased — `outBytes_delta=0` observed.
- A broker run with `-DV` logged no `$JS.API.CONSUMER.CREATE` for the rejected attempt, so "the broker's logs show nothing" is measured rather than reasoned.
- The threshold caveat correctly scopes the count to the advertised limit and subject length.
- The sweep-wall revision is correctly scoped as single runs on one host, not bounds. The unnamed-rerun sentence records a supplied 966ms figure without claiming the author measured it; reviewed with that provenance disclosed and found conservative rather than blocking.
- Frontmatter and `pnpm changeset status` unchanged from the prior head. `git diff --check` clean. HEAD reverified, worktree clean.

## Provenance

The 1,480,265 capture and the broker-side observations are the reviewer's measurements. The decomposition arithmetic and the 34-character subject width were re-derived independently here from `packages/core/src/subjects.ts`. The 233-byte envelope delta was not separately re-measured.

davidfarah2003:
## New head `0f3cd4a2bce7ffa7319b4ed0973012fc0713df32` — finding 1 addressed; the grade at `65414572f` and my note at `01ab76202` are both superseded

Posted in the same minute as the push.

### Finding 1: the claim was wider than its evidence, and it contradicted the sentence before it

The text said:

> A space with the 69 chat channels this work was aimed at is one batch, so **every figure above is
> unchanged** at that size.

"Every figure above" covers the wall times, and nothing establishes those are unchanged. The
sentence added immediately before it in the previous push says the opposite, that the walls are
single runs shaped by their host. So the paragraph contradicted itself across two adjacent
sentences. Now:

> ... is one batch, so the **request counts and byte totals** above are unchanged at that size. The
> walls are not claimed to be, for the reason just given.

That is the half the one-batch argument supports. This changeset already states elsewhere that the
request counts, consumer creates and page sizes are identical in every run while the byte totals
move by tens of bytes, so counts and bytes are the evidenced claim and the walls are not.

### Second commit is a rewrap only

No wording change. The previous commit left a 166-character line, the longest in the file by a wide
margin. The file has pre-existing lines at 124 and 150 characters, so the two 101-character lines
added here sit inside its convention.

### Why the head moved twice rather than once

The previous push addressed finding 2 and part of finding 3 and left finding 1 untouched, so the
bounded delta commissioned against `65414572f..01ab76202` contained no remedy for it and would have
blocked again correctly. Moving the head now costs one re-point; leaving it would have cost a full
round to be told something already known. The omission was mine.

### State

One file, `+28/-0` against `main`, purely additive. Frontmatter byte-identical, so the declared
bumps and `changeset status` do not move. Delta range for re-grade: `65414572f..0f3cd4a2b`.

### Ordering, unchanged

Must land before #1221 (`changeset-release/main`), whose diff deletes this changeset file.

davidfarah2003:
## FINAL head `d8a6e84f12111018d5021b479e6c9af31a7bf18a` — the byte figure now says which quantity it measures

Supersedes the grade at `65414572f` and my notes at `01ab76202` and `0f3cd4a2b`. Posted in the same
minute as the push. Nothing further is queued on this branch.

### The finding was right: 1,480,032 is not the request

The sentence framed the client as comparing *the request* against `max_payload`, then gave 1,480,032
as the figure. That number is the serialized **filter list**, not the request; the real request also
carries the envelope and the generated consumer config.

Verified independently before publishing, rather than taken from the review:

```
subject  cotal.actcost-wide.chat.*.*.w00000                     ->  34 characters
json.dumps({"config":{"filter_subjects":[subject]*40000}})      ->  1,480,032 bytes exactly
closed form  30 + N*(L+2) + (N-1) + 3   at N=40000, L=34        ->  1,480,032, agrees
```

So the figure is the reduced filter-list wrapper and the review's identification of it is correct.

### The remedy, and why this one

Two remedies were offered: publish the measured request size, or keep this figure and say what it is.
I took the second, because **nobody in this chain measured the request size** — the reviewer did, and
neither the relayer nor I did, and both said so. A seven-digit figure in a published CHANGELOG whose
run its author cannot name is the failure this PR has already been corrected for once.

The sentence now reads:

> On this corpus's subject shape the filter list alone serializes to 1,480,032 bytes, already past an
> advertised 1,048,576 before the rest of the request is counted, so a list that long cannot be sent
> whatever the envelope adds.

The filter list alone is 1.41x the advertised limit, so the conclusion no longer depends on knowing
the request size at all: the envelope can only add. The argument now follows from the quantity
actually named, which is what was wrong before rather than the digits.

### Also in this push

The paragraph is rewrapped as a single unit. The two previous edits left a 126- and then a
164-character line by joining onto existing wrapping. Longest added line is now 99. This is bundled
deliberately so no rewrap-only push follows.

### State

One file, `+30/-0` against `main`, purely additive. Frontmatter byte-identical, so declared bumps and
`changeset status` do not move. No em or en dashes. `git diff --check` clean.

### A note on the pattern, since it is two for two

Both rounds on this paragraph found the same shape: a mechanism named confidently with a number
beside it that came from somewhere else. First the broker refusing what the client refused, then the
filter list standing in for the request. Recorded here because the next reader of this file is
better served by knowing the paragraph was audited twice for exactly that than by a clean history.

davidfarah2003:
## FINAL head `0535d0f8b4a20acda2969f4647248823528a962a` — the byte figure is gone, and a fourth wrong quantity with it

Supersedes every earlier grade and note on this PR. Nothing further is queued on this branch.

### Why the figure is gone rather than corrected again

The review was right that `1,480,032` is the config wrapper rather than the filter list. Checking
that turned up a **fourth** wrong quantity in the same sentence, which three rounds had not seen
because all three were spent on which byte figure was correct.

`40,000` was reading as a threshold. It is not one. On this subject shape:

```
advertised max_payload        1,048,576
first N over it               28,340   ->  1,048,612 bytes
N just under                  28,339   ->  1,048,575 bytes
```

The crossing is near 28,340. 40,000 is simply the length that was tried. The sentence said "that
threshold moves with...", which names 40,000 as the threshold.

So the sentence has now been wrong four ways in three rounds, every time in the same shape: correct
digits attached to the wrong quantity. The candidates sit within 264 bytes of each other and no
reader of a CHANGELOG can distinguish them:

```
bare list                     1,480,001
{filter_subjects}             1,480,021
{config:{filter_subjects}}    1,480,032   <- previously published
measured request              1,480,265
```

The operational point never needed any of them. It now reads: the client refuses before anything is
sent, no bytes leave, the broker's logs show nothing, a list of 40,000 filters on this shape is
refused that way, and **40,000 is a length that was tried and not a threshold**.

### Second finding, fixed in the same push

"the request counts and byte totals **above** are unchanged" pointed at the sweep wall times, which
the next sentence explicitly excludes, while the figures meant are further down the file. The
direction word is dropped.

### State

One file, `+28/-0` against `main`, purely additive. Frontmatter byte-identical, so declared bumps and
`changeset status` do not move. Longest added line 99 characters. No em or en dashes. No byte figure
remains in the added text.

### On the record

Three of the four wrong quantities were mine and one came from a relayed gloss. The general lesson is
worth more than the attribution: reproducing a number and knowing *what it measures* are different
checks, and only the first was done. A figure that reproduces exactly can still be measuring
something other than the noun beside it.

PR #1246, opened 2026-09-04
fix(smoke): launch the CLI via node_modules/.bin/tsx, never npx tsx

davidfarah2003:
Fixes #1244.

## What was wrong

Eighteen smoke suites spawned the CLI as `spawn("npx", ["tsx", BIN, ...args])` with `cwd` set to a
scratch directory outside the repo. `npm exec` resolves a package from cwd upward and never from PATH,
so from there it misses the workspace `tsx` and falls back to whatever the npx cache holds or, on a
cold CI runner, a registry install of `tsx@latest`. The install prints

```
npm warn exec The following package was not found and will be installed: tsx@4.23.13
```

to stderr, which the fixtures merge into the output they parse. `ps` assertions then read a warning
where they expect a roster row. That reddened shards 2 and 3 on main at `7f3a8a03` with identical
warn counts across two attempts (3/3 and 2/2), on a tree byte-identical to the green run before it:
`git diff --name-only 917eb61a 7f3a8a03` contains zero smoke files and zero `.github` files, and
`pnpm install --frozen-lockfile` in the red log reads `resolved 438, reused 438, downloaded 0`.

## What this changes

One commit, 23 files, +127/-27, merges clean onto `main`.

- All 27 spawn sites across the 18 files use the form 36 sibling suites already used:
  `const TSX = join(import.meta.dirname, "..", "..", "..", "node_modules", ".bin", "tsx")` and
  `spawn(TSX, [BIN, ...])`. Includes the `pty.spawn` variants. No suite logic changed.
- `bin/smoke/no-npx-tsx.smoke.ts`, a grammar guard. It walks every `.ts` under a `smoke/` directory
  (587 on this tree) and reddens on `spawn("npx", ["tsx"`. Three controls make its zero mean
  something: a planted positive in `bin/smoke/fixtures/no-npx-tsx.planted.txt` that the regex must
  see, a negative that the robust form must not match, and a floor on the walked population so a walk
  over nothing cannot pass for free. Registered as `smoke:no-npx-tsx` with its own `ci-suites.d`
  fragment; `smoke:gate-inventory` is green with it.
- `.changeset/no-npx-tsx-in-smoke.md`, `patch` on `cotal-ai`.

## Measured on the fixed tree

```
smoke:cli-seat-locality          PASS 16/0   (13/3 red on main)      npm warn exec: 0
smoke:presence-ttl-refresh-cli   rc 0                                npm warn exec: 0
[...]

Issue #1244, opened 2026-09-04
ci: main red since 7f3a8a03 because 18 smoke suites spawn `npx tsx` from a scratch cwd and npx now downloads tsx@4.23.13 into the parsed output


davidfarah2003:
## Summary

Cotal `main` is red at `7f3a8a03` (the #1221 release merge): all four smoke shards fail. The base
`917eb61a` was fully green. **#1221 did not cause it**; it is what the runner's `npx` does now that
`tsx@4.23.13` exists on the registry and the repo pins `^4.23.0` with `4.23.0` in the lockfile.

18 smoke suites launch the CLI as `spawn("npx", ["tsx", BIN, ...args], { cwd: <scratch dir> })`. From
a cwd outside the repo, `npx` does not find the workspace's `tsx`, resolves `tsx@latest`, and prints:

```
npm warn exec The following package was not found and will be installed: tsx@4.23.13
```

That line is emitted to stderr, which the suites merge into `out`, so assertions that parse the CLI's
output read a warning where they expect JSON or a roster row. Where the download eats the time
budget, the fixture SIGKILLs `cotal up` and `cotal down` then fails on the half-built root.

## Measured

Per shard, the failing suite on main `7f3a8a03`, identical on PR #1242 which merged into it:

```
shard 0  smoke:ps-operator-path          cotal up SIGKILLed by suite timeout, then cotal down exit 1
shard 1  smoke:presence-ttl-refresh-cli   same shape
shard 2  smoke:cli-seat-locality          ps assertions read the npm warn line as output (3 FAIL)
shard 3  smoke:manager-two-root-renewal   daemon B boot output is the npm warn line
```

`npm warn exec ... tsx@4.23.13` occurrences per shard:

```
                    shard 0  shard 1  shard 2  shard 3
main 917eb61a green    0        0        0        0
main 7f3a8a03 red      0        0        3        2
PR #1242      red      0        0        3        2
```

Shards 0 and 1 show zero warnings because `cotal up` is SIGKILLed before its output is captured;
the `cotal down` failure that follows is the consequence, not a second cause.

Reproduced on a dev host, same npm 11.19.0 / node 24:

```
$ cd /var/tmp/anywhere && npx tsx --version      # repo's node_modules/.bin on PATH
tsx v4.23.13                                      # <- registry, not the workspace
$ cd <repo> && npx tsx --version
tsx v4.23.0                                       # <- the lockfile's
```

[...]

Issue #1179, opened 2026-08-30
jcode-host model_mismatch cell races its bridge teardown: 'does not carry its launch-bound identity — refusing unsafe teardown'


davidfarah2003:
Second sighting of an intermittent red in `smoke:jcode-host`'s `model_mismatch` cell: the launch fails with

    private Harness connection closed and recovery failed: jcode connector: spawned Jcode bridge <pid> does not carry its launch-bound identity — refusing unsafe teardown

(the identity check at `extensions/connector-jcode/src/private-lifecycle.ts:259`).

Occurrences, different refs, both attempt-1-only:
- PR #1041, run 33319699947 attempt 1 (bridge 6865) — greened on attempt 2, identical merge ref
- PR #1055, run 33331085489 (bridge 6952) — control socket under `/tmp/cjh-…`, so post-#1164 short-root, not the sun_path class

Pattern matches a pid/identity race on the respawned bridge during the model_mismatch recovery path, same fail-then-green family as the (now fixed) #1169 but a different mechanism and cell. Needs the #1169 treatment: loop the suite on linux, catch a failing trace, and determine whether the suite races its subject or the identity proof itself is racy.

davidfarah2003:
Diagnosis-lane verdict: **NOT-REPRODUCED on a workstation, latent defect characterized, candidate fix parked (unapplied).**

Counts on a netcup box at 80c42e43: 52/52 full-suite iterations green (serial then 3 concurrent full-suite workers), 262/262 targeted `model_mismatch` rounds green, ~1000/1000 direct `stopPrivateTree` drives green — the tail never converted into a single refusal. Per "never fix an issue you could not reproduce", nothing ships.

What IS established (finding 1): `processHasLaunchIdentity` reads `/proc/<pid>/environ`, and a process releases its mm backing that file **as it exits, before** it becomes the zombie that `alive()` reads as dead. In that window the environ read fails or returns empty while `/proc/<pid>/stat` still reports a live state and an unchanged start token — so a bridge that is merely dying can be convicted as "does not carry its launch-bound identity". The path from this window to the two CI failures is *plausible but unproven here*: on this workstation the exit window closes faster than the `/proc` scan, which is likely why ~1300 drives never hit it. Recommended repro venue: a CI-shaped runner (fewer vCPUs, higher load).

[...]

PR #1251, opened 2026-09-04
test(smoke): opt the ps and presence-ttl fixtures out of the connector seed

### LINKED ISSUE #1245: ci: presence-ttl-refresh-cli SIGKILLs `cotal up` at the 120s budget, seeding @cotal-ai/web where the green run seeds nothing
> **PARTLY SUPERSEDED.** The seed-line reasoning below was retracted; the correlation table it rests on does not stand. Current state, evidence and fix: https://github.com/Cotal-AI/Cotal/issues/1245#issuecomment-5538919186

`smoke:presence-ttl-refresh-cli` fails on main at `7f3a8a03`: `cotal up` is SIGKILLed at the fixture's 120s budget, and the failing run materializes a seed-store payload that the passing run never writes.

This is a different defect from #1244 and is not fixed by it. #1244 converts 18 suites off `spawn("npx", ["tsx", ...])`. This failure never reaches npx: it goes through `cotal up` performing an ext add. Shard 1 will stay red after #1244 lands, and that should not be read as #1244 failing.

## The failure

Shard 1, `pnpm smoke:presence-ttl-refresh-cli`. The suite prints its step banner, produces nothing for 143 seconds, then flushes its buffered output when the pipe is destroyed:

```
1) cotal up - the mesh that will be 'already running' for the rest of this suite
    (143 seconds of silence)
-> wrote operator-global seed store payload (web): /tmp/cotal-ttlrefresh-*/xdg/cotal/seed/store/0.41.0/web
   added @cotal-ai/web@0.41.0 - provides: command:web, local-process:web
Error: FIXTURE FAILURE, not a product defect: `cotal up` was SIGKILLed by this suite's timeout.
```

The budget is `timeoutMs = 120_000` in `packages/core/smoke/presence-ttl-refresh-cli.smoke.ts:73`, and the refusal comes from `mustHaveRun` at `:96`. The suite is correct to refuse: a killed child and a product failure produce the same `status` shape, so it declines to grade either.

Reproduced on both attempts of run 33820206334 at the same commit, same suite, same signature.

## The comparison that isolates it

Same suite, same shard, base `917eb61a`:

| | `cotal up` | seed-store lines |
| --- | --- | --- |
| `917eb61a` | exits 0 in 35s | 0 |
| `7f3a8a03` | SIGKILLed at 120s | 1 |

The green run completes in 35 seconds having materialized nothing. The red run exceeds a 120 second budget having materialized the web payload.

## Why the log line is a reliable discriminator

`stageSeedPayload` in `implementations/cli/src/seed/store.ts:30` returns early when `join(dest, "package.json")` already exists, and that early return writes nothing and emits nothing. The provenance line is written only after a real `cpSync` and atomic rename. So the line's presence means a materialization happened, and its absence means one did not. This is not inference from incidental log noise.

## Correlation across every job with logs

| job | seed materialization | outcome |
| --- | --- | --- |
| `917eb61a` shard 0 | 0 | passed |
| `917eb61a` shard 1 | 0 | passed |
| `7f3a8a03` shard 0 | 0 | passed on rerun |
| `7f3a8a03` shard 1 | 1 | failed on both attempts |

The only job that materialized a seed payload is the only job that hit the `cotal up` timeout.

## What is not established

Two gaps, stated because neither is closed and the card should not read as though they are.

The seed store lives inside the suite's own scratch sandbox, under an `XDG_CONFIG_HOME` created per run. It should therefore be empty on every run at every version, which means the green run should have materialized as well, and it did not. That is unexplained. Until it is explained, "the version bump left the store cold" is not the mechanism, because the store looks cold in both runs.

The seeding is also not shown to be what consumed the 120 seconds. It is the only work visible in the silent window, which is suggestive and not proof.

## Where to look

`implementations/cli/src/commands/ext.ts:240` runs:

```
npm install --save --no-audit --no-fund --legacy-peer-deps --install-links <resolved>
```

There is no `--offline` or `--prefer-offline`. `--install-links` copies a `file:` spec rather than symlinking it, but npm still resolves that package's own dependencies, and `@cotal-ai/web` is the heaviest entry in the seed set. A dependency resolution there is the one step in this path that can block on the network long enough to exhaust the budget. That is a lead and has not been measured.

## Method

Graded by reading the code and the CI logs. `cotal up` was not exercised, because running it is prohibited on the seat that produced this analysis. The two gaps above are what that leaves unverified.

Related: #1244.

davidfarah2003:
Closes #1245.

`ps-operator-path` and `presence-ttl-refresh-cli` bound `cotal up` at 120 seconds and neither grades the connector seeder, but neither sets `COTAL_SKIP_CONNECTOR_SEED`. Each therefore pays a seven-payload seed reconcile inside that budget. Fifteen sibling suites already opt out, and `up` itself sets the flag for the daemons it launches, so this brings the two into line with the convention rather than introducing one.

## Both suites pass today

This is margin, not a repair. Nothing is broken on `main` and this PR unblocks nothing.

The case for it is the spread. `cotal up` wall time in `ps-operator-path`, bracketed from the fixture's last anchor check to the next printed line:

| commit | wall | outcome |
| --- | --- | --- |
| `3aba994a3` (this base) | 34s | green |
| `917eb61a` | 36s | green |
| `dced13d3f` | 68s | green |
| `7f3a8a03` attempt 2 | 75s | green |
| `7f3a8a03` attempt 1 | 120s | SIGKILLed |
| `24b3d9281` | 120s | SIGKILLed |
| `a43a127f4` | 120s | SIGKILLed |

Passing runs span 34s to 75s against a fixed 120s wall, and the failures sit exactly at the wall because that number is the fixture's SIGKILL rather than a measured duration. The spread tracks npm cache state rather than anything in the tree, so a green run is not evidence that the budget is comfortable.

A sibling harness with the same defect, `delivery-timer-writer`, went from 8/4 failing to 12/0 passing on an unchanged cold cache once the flag was set (`a43a127f4`). That is the same change in the same shape, measured as an intervention.

## How to verify this, which is not the shard colour

A green shard cannot separate a suite that stopped paying the cost from one that fit inside the budget that day. Measure `cotal up` wall time with the bracket above. Before, on this base: `ps-operator-path` 34s, `presence-ttl-refresh-cli` 38s. After, both should sit at or below those and stop moving with cache state.

Do **not** verify by the absence of `wrote operator-global seed store payload (web)` in the log. That line reaches the log only when a subprocess's output is printed, which happens only on failure, so a passing run shows zero of them whether or not it seeded. An earlier version of #1245 was built on that signal and it was withdrawn.

## Scope

Two suites and a changeset. No product code, no other suite, and no change to the seeder itself. Whether other suites carry the same omission was not surveyed here.

### DIFF
--- a/.changeset/skip-connector-seed-in-two-suites.md
+++ b/.changeset/skip-connector-seed-in-two-suites.md
@@ -1,0 +1,5 @@
+---
+"cotal-ai": patch
+---
+
+The `ps-operator-path` and `presence-ttl-refresh-cli` smoke suites opt their CLI out of the connector seed reconcile, as fifteen sibling suites already do and as `up` does for the daemons it launches. Neither suite exercises the seeder, and on a cold npm cache the reconcile can consume the whole `cotal up` budget the fixtures allow.
--- a/implementations/auth/smoke/ps-operator-path.smoke.ts
+++ b/implementations/auth/smoke/ps-operator-path.smoke.ts
@@ -81,7 +81,10 @@ type Run = {
 };
 function cotal(args: string[], timeoutMs = 120_000): Promise<Run> {
   return new Promise((res) => {
-    const options = { cwd: root, env: { ...process.env, COTAL_HOME: home, XDG_CONFIG_HOME: configDir }, stdio: ["ignore", "pipe", "pipe"] as const };
+    // This suite grades the operator `ps` path and never exercises the connector seeder, so its
+    // CLI opts out of the seed reconcile the way `up` does for the daemons it launches. On a cold
+    // npm cache that reconcile can spend the whole `cotal up` budget this fixture allows (#1245).
+    const options = { cwd: root, env: { ...process.env, COTAL_HOME: home, XDG_CONFIG_HOME: configDir, COTAL_SKIP_CONNECTOR_SEED: "1" }, stdio: ["ignore", "pipe", "pipe"] as const };
     assertSmokeSandboxDown(sandbox, args, options);
     const child = spawn(TSX, [BIN, ...args], options);
     let out = "";
--- a/packages/core/smoke/presence-ttl-refresh-cli.smoke.ts
+++ b/packages/core/smoke/presence-ttl-refresh-cli.smoke.ts
@@ -73,7 +73,10 @@ const check = (n: string, v: boolean, x?: unknown) => { v ? (pass++, console.log
 type Run = { status: number | null; out: string; timedOut: boolean; signal: NodeJS.Signals | null; launchError?: string };
 function cotal(args: string[], timeoutMs = 120_000): Promise<Run> {
   return new Promise((res) => {
-    const options = { cwd: root, env: { ...process.env, COTAL_HOME: home, XDG_CONFIG_HOME: configDir }, stdio: ["ignore", "pipe", "pipe"] as const };
+    // This suite grades presence TTL refresh across restarts and never exercises the connector
+    // seeder, so its CLI opts out of the seed reconcile the way `up` does for the daemons it
+    // launches. On a cold npm cache that reconcile can spend the whole `cotal up` budget (#1245).
+    const options = { cwd: root, env: { ...process.env, COTAL_HOME: home, XDG_CONFIG_HOME: configDir, COTAL_SKIP_CONNECTOR_SEED: "1" }, stdio: ["ignore", "pipe", "pipe"] as const };
     assertSmokeSandboxDown(sandbox, args, options);
     const child = spawn(TSX, [BIN, ...args], options);
     let out = "", timedOut = false, settled = false, exited = false;


davidfarah2003:
## Pre-registering what this change should do to the number, before the number exists

The before-side is on record: `cotal up` at 34s in `ps-operator-path` and 38s in `presence-ttl-refresh-cli`, on this PR's base.

"At or below 34s" is too weak to be a test, because almost any outcome satisfies it. So here is a band that can fail.

**Prediction.** If the seed reconcile is what these suites are paying for, removing it should take out a double-digit number of seconds, not a marginal one. A sibling harness measured the same reconcile at roughly 16s with a warm npm cache and 30s or more with a cold one. So `cotal up` here should land materially below 34s, and the drop should be visible without needing a cold-cache run to see it.

**What would falsify it.** An after-number in the high twenties or low thirties. That would mean the seed was never a large share of the 34s, the mechanism argued on #1245 is wrong, and this PR is removing something cheap while claiming to remove something expensive. If that is the result I will say so on this PR and reopen the question rather than keep the change on a narrower justification.

**What would not settle it either way.** A single green shard. A green shard cannot separate a suite that stopped paying the cost from one that fit inside the budget on the day, which is why the number is being measured rather than the colour read.

**A caution about the cache, stated in advance so it cannot be used selectively afterwards.** Neither side of this comparison controls CI's npm cache. If the after-number is low on a warm cache, that is consistent with the fix and also consistent with the cache having done the work, and I will describe it that way rather than claim more. The reading gets stronger if a second after-sample on a different day lands in the same place, since two lows are harder to explain by cache alone than one.

The before-side has the same limitation and it is worth naming symmetrically: 34s and 38s are single samples too. Main's own run at `3aba994a3` supplies a second before-sample on the identical base, which gives that side a spread rather than a point.

davidfarah2003:
## Independent review: APPROVE at `1eb94b822ace2431c152544ff3a0cb6731a35fe8`

Reviewed independently of the author on a detached worktree at that exact head, sha resolved with `git cat-file -t` before grading. The review found an error in my numbers, corrected below.

**Convention.** At parent `3aba994a3`, 15 non-live `.smoke.ts` files set `COTAL_SKIP_CONNECTOR_SEED: "1"`; 17 counting `ext-live` and `spawn-detach-live`. The "fifteen" in the PR body is the non-live count and is correct as stated.

**Does the change blind the suites.** This was the question most likely to produce a block, and it was checked directly. Neither suite has a cell naming a connector, `seed`, `ext`, or `web`. `ps-operator-path` grades the scratch anchor, `cotal up` exit 0, the operator `cotal ps --space` exit 0, and that a dead-manager `ps` is not a silent empty success; `ps` is a built-in in `implementations/cli/src/index.ts` rather than a seeded command. `presence-ttl-refresh-cli` grades the first `up`, a fixture-staged `max_age=0`, the already-running TTL restore through `reconcileSpaceTtls` in core, and a third no-op `up`. No connector is reached.

The fidelity objection was considered rather than skipped: `skipAutoReconcile` at `command.ts:95-101` is written for a daemon child whose parent already seeded, so setting it on a fixture's first `up` means that scratch never seeds, unlike a real user's first `up`. That difference does not touch any cell either suite asserts, and the same parent-skip is already in use by the 15 non-live siblings, including `sys-rotation-e2e`, which wraps `cotal(...)` the same way and calls `up`.

**Scope.** 3 files, +13/-2, two suites plus the changeset. No product code and no change to the seeder.

**Changeset.** `"cotal-ai": patch`, matching `.changeset/no-npx-tsx-in-smoke.md`. Patch is right for a test-only change and the `fixed` group locksteps the rest.

## Correction to my own measurement

The review checked the wall times against the job logs via the jobs API and found that **the 38s I published for `presence-ttl-refresh-cli` is wrong.** The correct figure for its first `cotal up` is **32s**.

The cause is worse than a mis-chosen interval, and it is the same attribution error recorded on #1245: shard 1's log contains two anchor blocks, and mine matched the first. That one belongs to `smoke:ps-user-mode`. `presence-ttl-refresh-cli`'s own anchor is 190 lines later. Verified per-banner:

```
anchor at line 2375 -> smoke:ps-user-mode
anchor at line 2565 -> smoke:presence-ttl-refresh-cli
presence-ttl first `cotal up`: banner 10:08:40.387 -> exit 10:09:12.518 = 32s
```

`ps-operator-path`'s 34s is unaffected: shard 0's log contains exactly one anchor occurrence and it is owned by that suite.

**So the corrected before-side is 34s and 32s**, not 34s and 38s. The pre-registered prediction above is restated against those numbers: the drop should be a double-digit number of seconds from each, and an after-number in the high twenties or low thirties falsifies the mechanism.

This is the third instance of the same per-suite attribution error on this work, and the first one I made after writing the rule against it onto #1245.

davidfarah2003:
## Pre-stating the third branch: the run may produce no after-number for one of the two suites

The prediction above has a confirm branch and a falsify branch. It is missing the branch that is not about the outcome at all, and I would rather write it down before the shards report than explain it afterwards.

**`presence-ttl-refresh-cli` runs in shard 1, which is where #1179 lives.** That issue is open. Its `jcode-host` cell races its bridge teardown, and when it loses the race it fails early enough to leave 84 of the shard's 120 suites unrun, `presence-ttl-refresh-cli` among them. A shard log that stops there prints `NEVER RAN` for everything after it.

So there are three outcomes for that suite on this PR, not two:

1. the suite runs and reports a wall time, which confirms or falsifies as pre-registered;
2. the suite runs and fails for its own reasons, which is a result;
3. **the shard dies at `jcode-host` and the suite never executes**, which is neither.

The third is the one that gets misread. A shard that goes red before reaching a suite says nothing about that suite, and a shard that goes red is easy to quote as if it did. If that is what lands here, the honest statement is that the measurement did not occur, and I will say that rather than reaching for the closest available number.

**This is a race, not a certainty, and the distinction matters.** Shard 1 was fully green at `7639da0b5` and did reach `presence-ttl-refresh-cli` there. The 32s before-number came out of that run. So the shard can get through; #1179 fires intermittently.

**`ps-operator-path` is in shard 0 and is not behind that cell.** Shard 0 should produce a real after-number regardless, and the comparison against 34s stands on shard 0 alone. One measured suite is enough to test the mechanism; two would be better.

Recording the branch in advance because the last time I wrote out the possible outcomes on this work I wrote three, and all three assumed a measurement would happen. The missing branch was the one that occurred.

davidfarah2003:
## A cleaner before-number: 36s for `ps-operator-path`, measured at this PR's exact parent

The before-numbers published above (34s and 32s) come from `7639da0b5`, which is #1246's head rather than this PR's base. Main has since gone green on shard 0 at `3aba994a36`, and that commit **is** this PR's parent, so it yields the baseline the after-number should actually be compared against.

Measured the same way, on a green run, from main's shard 0 (job `100995745947`):

```
banner  line 1949  ===== pnpm smoke:ps-operator-path =====
anchor  line 1953  10:47:25.712  ✓ ...and therefore outranks any ancestor
next    line 1954  10:48:01.502  ✓ `cotal up` exits 0
                   -----------------------------------------
                   35.79s
```

Attribution checked rather than assumed: shard 0's log contains exactly one occurrence of that anchor string, and the nearest preceding banner is `smoke:ps-operator-path`. That check is the one whose absence produced the 38s error corrected above.

**So the before-side for `ps-operator-path` at the exact parent is 36s** (35.79s). It lands inside the previously reported green spread of 34 to 75 and matches the 36s recorded at `917eb61a` to the second.

### What this does and does not change

It does not move the prediction. A double-digit-second drop from 36 confirms the mechanism; an after-number in the high twenties or low thirties falsifies it, and I will say so on this PR.

It does sharpen the comparison, because base and parent are now the same commit. Comparing an after-number at `1eb94b82` against a baseline at `7639da0b5` would have been comparing across two different trees; comparing it against `3aba994a36` is not.

`presence-ttl-refresh-cli` has no parent-commit baseline yet: main's shard 1 at `3aba994a36` has not reported. Its before-number remains 32s from `7639da0b5` until it does.

### One correction to the hold recorded on #1245

That hold predicted shard 1 would stay red after the #1244 repair and said anything else was new information worth reporting. Shard 0, which the prediction did not cover, has come back **green** on main at `3aba994a36`. `ps-operator-path` passed there without this PR's change, at 36s against its 120s wall.

That is the latency this card already describes rather than a contradiction of it: the suite passes on the runs where the variable cost fits inside the budget, and this was one of them. Reporting it because a green baseline is exactly the observation that could be quoted later as "the problem went away," and it is not that. It is one draw from a distribution whose upper tail has crossed the wall three times.

davidfarah2003:
## Result: the prediction is CONFIRMED, and by a wider margin than I argued for

All four shards are green at `1eb94b822ace2431c152544ff3a0cb6731a35fe8`. Both suites ran (no `NEVER RAN` block in either shard), so the third branch I pre-stated did not occur and there is a real measurement for each.

| suite | before (parent `3aba994a36`) | after (`1eb94b82`) | drop | share of the 120s wall |
| `smoke:ps-operator-path` | 35.79s | **10.78s** | −25.01s (−70%) | 21% of the budget returned |
| `smoke:presence-ttl-refresh-cli` | 32.11s | **10.82s** | −21.29s (−66%) | 18% of the budget returned |

Both drops are double-digit seconds, which is what I pre-registered as confirmation. I also wrote that an after-number in the high twenties or low thirties would falsify the mechanism and that I would say so; 10.78 and 10.82 are not in that band by a wide margin.

Measured exactly as the before-numbers were: anchor line to the `cotal up` outcome line, attributed by walking back to the nearest `===== pnpm smoke:<name> =====` banner. Shard 1 again contains two occurrences of the anchor string, the first belonging to `smoke:ps-user-mode`; taking the first is what produced the retracted 38s.

### The part I did not predict, and which is the strongest evidence here

**The two suites converge.** Before the change they differed by 3.68s (35.79 against 32.11). After it they differ by **0.04s** (10.78 against 10.82).

Two fixtures with different roots, different auth modes and different step sequences now agree on the cost of `cotal up` to within forty milliseconds. That is the signature of removing a shared variable term and leaving the fixed one: what remains is the cost of bringing a mesh up, and what left was the thing both were paying that neither graded. A change that merely happened to run on a fast day would not make two independent suites agree to 40ms.

### The reviewer's central question, answered empirically rather than by reading

The review's first item was whether the flag blinds the suites. It graded that by reading the cells. The run settles it directly:

| suite | before | after |
| --- | --- | --- |
| `ps-operator-path` | `5 passed, 0 failed` | `5 passed, 0 failed` |
| `presence-ttl-refresh-cli` | `15 passed, 0 failed` | `15 passed, 0 failed` |

Same assertion counts, all passing, 25 and 21 seconds cheaper. If the flag had removed something a cell depended on, the count would have moved or a cell would have failed.

### What this does not establish

**It does not prove the removed cost is only the seed staging.** `COTAL_SKIP_CONNECTOR_SEED=1` skips the whole connector seed reconcile at `command.ts:95-101`, and the seed payload staging is part of that rather than all of it. The honest claim is that the reconcile cost these suites 25s and 21s, not that `stageSeedPayload` did.

**It is one observation per suite.** The before-side for `ps-operator-path` is a distribution spanning 34 to 75 with three runs at the wall; the after-side is a single draw. What would make this conclusive is the after-number staying near 10.8s across runs with varying cache state, which is the property the change is actually for. A second green run at a similar figure would establish it; I am not going to claim it from one.

**It does not make the card urgent retroactively.** Both suites passed before this change, as recorded on #1245. This removes a variable term that had crossed the wall three times, rather than repairing a break.

davidfarah2003:
## After-numbers at `1eb94b822`, bracketed the same way as the before-numbers

Wall time of each suite's first `cotal up`, from the `...and therefore outranks any ancestor` anchor to the `cotal up exits 0` check, scoped to the window between the suite's own `===== pnpm smoke:<name> =====` banner and the next banner. Both suites ran (banner present, absent from every `NEVER RAN` block) and both passed.

| suite | before (parent `3aba994a3`) | after (`1eb94b822`) | job |
| --- | --- | --- | --- |
| `ps-operator-path` | 36s | 11s | [100996817610](https://github.com/Cotal-AI/Cotal/actions/runs/33864712307/job/100996817610), banner line 1962, `PS OPERATOR PATH OK (5 passed, 0 failed)` |
| `presence-ttl-refresh-cli` | 32s | 11s | [100996817678](https://github.com/Cotal-AI/Cotal/actions/runs/33864712307/job/100996817678), banner line 2558, `PRESENCE-TTL-REFRESH-CLI SMOKE OK (15 passed, 0 failed)` |

Both drops are double-digit seconds from the parent baseline, which is the confirmation band the PR pre-registered. The comparison is green to green; the earlier killed runs are bound-crossings and are not in this table. Cache state on the runner is not controllable from here, so the residual reading that a warm cache did part of the work is not excluded by one run; what one run does show is that with the seed skipped, `cotal up` in these two suites no longer carries the term whose green-side spread reached 68s and 75s.

davidfarah2003:
## On the warm-cache caveat: the run design answers most of it, and I can say which part it does not

A fair objection was raised against the result above — one run cannot exclude a warm npm cache doing part of the 25s and 21s. I went and read the workflow rather than argue it, and the answer is better than I expected, though not complete.

### What CI actually caches, from `.github/workflows/ci.yml`

The smoke shard job caches exactly two things:

| step | what | key |
| --- | --- | --- |
| `actions/setup-node@v6` with `cache: pnpm` | the **pnpm store**, for `pnpm install --frozen-lockfile` | lockfile hash |
| `actions/cache@v4` | `~/nats-bin`, the nats-server binary | `nats-server-v2.12.12-${{ runner.os }}` |

**There is no npm cache step.** Nothing caches `~/.npm` or sets `npm_config_cache`. So the fetches the connector seed performs are not served from any restored cache on any run, and the pnpm store that *is* restored is consumed by `pnpm install`, which completes two steps before the bracketed interval begins.

### Why that makes the convergence load-bearing

The four shards are separate `ubuntu-latest` runners under a matrix. Shard 0 and shard 1 are different machines with independent local state.

| | shard 0 (`ps-operator-path`) | shard 1 (`presence-ttl`) | difference |
| --- | --- | --- | --- |
| before, at `3aba994a36` | 35.79s | 32.11s | **3.68s** |
| after, at `1eb94b82` | 10.78s | 10.82s | **0.04s** |

Two independent runners drawing the same variable cost disagreed by 3.68 seconds. The same two, after the flag, agree to forty milliseconds. For per-runner cache warmth to explain the drop, two independent machines would have to have been warm by near-identical amounts on the very run where the flag was added, in a job that caches no npm state — while the before-run shows independent draws differing by seconds.

### The part this does NOT rule out, which is real

A **globally** fast window — registry or CDN responsiveness shared by both runners — is not excluded by the convergence, because both after-measurements fall inside eleven minutes (11:16:21 and 11:27:11).

The before-measurements were three minutes apart (10:47:25 and 10:50:29) and still differed by 3.68s, which argues against a global window producing tight agreement. But that is an argument, not a control.

**What would actually settle it** is the after-number holding near 10.8s on later runs at unrelated times. That is the property the change is for, and it accrues on its own as main runs. If either suite is later seen materially above that, this should be reopened rather than explained.

I would rather state the residual confound plainly than let the convergence carry more weight than it can. The claim I am making is that the reconcile cost these two suites 25s and 21s on this run, on evidence that makes per-runner cache warmth an implausible explanation and leaves a global window as the surviving alternative.

davidfarah2003:
## Follow-up measurement: the number held, so there is nothing to reopen

I committed above to reopening #1245 rather than explaining, if either suite came back materially above 10.8s on a later run. A second run has landed and it did not.

The merge commit `142f31080718de65d5be85b604685ef54cf773bc` produced its own CI run on fresh runners, giving a second observation of both suites with the change in place.

| suite | PR run (`1eb94b82`) | merge-commit run (`142f3108`) |
| --- | --- | --- |
| `smoke:ps-operator-path` | 10.78s | 10.88s |
| `smoke:presence-ttl-refresh-cli` | 10.82s | 10.76s |

**Four after-measurements spanning 10.76s to 10.88s — a range of 0.12s — across two separate workflow runs, four independent runners, and 55 minutes.** The before-pair, measured the same way on the same two suites, differed from each other by 3.68s.

Both suites ran in their own banner windows in both runs, neither masked, and both reported full cell counts (`5 passed, 0 failed` and `15 passed, 0 failed`). The second run's numbers were produced independently and I reproduced one of them from the job log rather than adopting it: `presence-ttl-refresh-cli` brackets at 10.76s against 10.75s reported, which is rounding.

### What this does to the confound I flagged

The remaining alternative explanation was a globally fast registry or CDN window shared by both runners, since the first two after-measurements fell inside eleven minutes.

That is now a much harder story to tell. It would need a window that persists across 55 minutes and two separate workflow runs, and that delivers 0.12s consistency to four independently provisioned runners, while the pre-change measurements on the same infrastructure differed by seconds. **I would not call it formally excluded** — every observation so far still sits inside a single hour — but it has gone from a live alternative to a strained one.

A further datum will arrive on its own from the next unrelated main run, which will be further out in time. If it disagrees, the commitment above still stands.

### The claim I am left making

The connector seed reconcile cost these two fixtures roughly 25 and 21 seconds, and removing it leaves `cotal up` at about 10.8s in both, stably across runs. That is a fifth of each fixture's 120 second budget returned, and the elimination of a term whose green observations previously ranged from 34s to 75s and whose upper tail had crossed the wall three times.

PR #1255, opened 2026-09-04
ci(release): verify the whole publish closure before cutting a Release

### LINKED ISSUE #1254: release: the publish gate checks one package via npm view, so a partial publish cuts a Release and a lagging packument skips one
The changesets workflow gates cutting the GitHub Release on the release actually reaching npm. That gate is a good idea and it caught a real failure before (the phantom `v0.6.0` its comment describes). Two things about how it checks are worth tightening. **Neither fired on 0.41.1 today — that release is correct and complete — so this is a latent gap, not an incident.**

The step, in `.github/workflows/changesets.yml`:

```sh
for i in 1 2 3 4 5; do
  if npm view "cotal-ai@$version" version >/dev/null 2>&1; then published=1; break; fi
  echo "cotal-ai@$version not yet on npm (attempt $i/5) — waiting…"; sleep 5
done
```

### 1. It checks one package, but the release is 21 packages

`.changeset/config.json` puts 21 packages in one `fixed` group, so every release versions them in lockstep. The gate asks only whether `cotal-ai` is live.

A publish that lands `cotal-ai` and drops any of the other 20 therefore passes the gate, cuts a Release, and announces a version that is not fully on npm. That is exactly the failure the lockstep group makes possible, and nothing downstream would notice: a consumer only hits it if the missing package is in their dependency path.

Worth noting what makes it survivable today rather than assuming it always will be: on 0.41.1, `cotal-ai@0.41.1` declares no dependency on the packages most likely to be skipped, so a partial publish would have been incomplete rather than install-breaking. That is a property of the current dependency shape, not a guarantee.

Checking the closure is cheap — one HTTP request per package against `https://registry.npmjs.org/<urlencoded-name>/<version>`, with the list derived from `.changeset/config.json` rather than hard-coded.

### 2. `npm view` reads the packument, which lags in both directions

`npm view` reads the aggregate package document, not the version endpoint, and the two disagree for minutes after a publish. Both directions have now been measured on this repo:

| release | what lagged | observed |
| --- | --- | --- |
| 0.41.0 | the **packument** was behind | `@cotal-ai/delivery` packument showed no 0.41.0 while `/delivery/0.41.0` served HTTP 200 with a real tarball |
| 0.41.1 | the **version endpoint** was behind | `@cotal-ai/tmux/0.41.1` returned 404 on three consecutive polls at 14:06:32Z while its packument already showed `dist-tags.latest = 0.41.1`; it was 200 by 14:08:00Z |

The retry budget is 5 attempts at 5 seconds, so 25 seconds. The 0.41.0 packument lag exceeded that. When it does, the step takes the `published=` empty branch, prints "this push published nothing", and **skips the Release for a version that did publish** — the mirror of the phantom Release the comment was written to prevent, and quieter, because a missing Release looks like a push that legitimately published nothing.

### Suggested shape

- Derive the package list from `.changeset/config.json` and check every one.
- Query the per-version endpoint and the packument, and treat **disagreement between them as "not settled yet"** rather than as an answer in either direction. Re-poll until they agree; both observed lags resolved within about two minutes.
- Give the loop a budget in minutes rather than 25 seconds, and make the give-up branch loud: skipping a Release because the registry did not answer is a different outcome from skipping because nothing was published, and today they print the same thing.

A script doing the closure half is at `/home/cotal/fm-lanes/notes/verify-0.41.1-publish.sh` on the netcup host if it is useful as a starting point; it derives the list and prints an explicit INCOMPLETE verdict rather than a bare pass.

davidfarah2003:
Closes #1254.

The release workflow decides whether to cut a GitHub Release by asking `npm view cotal-ai@<version>`
behind a 5×5s retry. That asks a narrower question than the one that matters, and gets two cases
wrong in opposite directions — both silently.

- **It reads one package.** Every package is versioned in a single `fixed` group of 21, so a publish
  that lands `cotal-ai` and drops a sibling still cuts a Release claiming the whole release shipped.
- **It reads the packument, which lags.** Measured here: at 0.41.0 the packument was *behind* the
  per-version endpoint; at 0.41.1 it was *ahead*. Neither surface is reliably the fast one, so a
  disagreement between them means *not settled*, never evidence either way. A lag longer than the
  retry silently **skips** the Release for a version that did publish.

## What this adds

`scripts/verify-publish-closure.mjs`, which derives the package list from `.changeset/config.json`
and polls the per-version endpoint for the whole closure.

The list is derived rather than carried because an earlier hand-written one held **20 names against a
21-package group**. It would have reported "fully published" while blind to precisely the failure it
existed to catch. A gate's own inputs have to come from the source of truth, or the gate inherits the
error it is checking for.

## Why it polls, and where the numbers come from

Reporting success is easy: emptiness is monotone, so the moment the missing set empties the answer is
final. **Declaring failure is the hard direction** — a partial publish and ordinary propagation
produce an identical reading. The discriminator is whether the missing set *shrinks*, and the samples
have to span longer than the propagation window or they are one sample in disguise.

Measured on this repo, polling immediately after the publish job reported success:

```
0.41.2   17/21 -> 17/21 -> 19/21 -> 21/21     (60s apart; reads 1 and 2 IDENTICAL, still pure lag)
0.41.1   20/21 -> 21/21                       (~2 min to settle)
```

So a one-shot check reports a false failure, and a "two identical reads means stable" rule reports
one too — *more convincingly*, because it looks like it controlled for lag.

That is **two observations, not a distribution.** `stableWindowMs` is therefore a deliberate margin
over the larger of them, not a measured percentile, and it is generous on purpose: crying wolf on lag
gets a gate switched off, which is worse than the hole it closes. When the evidence is insufficient
the gate reports `UNSETTLED` rather than asserting a failure it cannot distinguish from a slow
registry.

Four states, four exit codes:

| exit | state | meaning |
| --- | --- | --- |
| 0 | `PUBLISHED` | whole closure live; a Release may be cut |
| 1 | `PARTIAL` | some of the group live and some not, stable past the window — **the dangerous case** |
| 2 | `UNSETTLED` | still moving, or deadline passed; cannot tell, skip rather than guess |
| 3 | `NONE` | nothing published at all (an ordinary version-PR push); skip, not a failure |

`NONE` is separate from `PUBLISHED` deliberately. Collapsing them to 0 would cut a Release for a
version that published nothing — the phantom-Release bug this gate exists to prevent, reintroduced at
the exit code.

## The declaration

Emitted from the module by `pnpm gen:publish-closure-dts`, with a cell asserting the committed copy
is byte-identical to a fresh emit. This follows the `reap-smoke-brokers.d.mts` precedent: a
hand-written declaration is a second source of truth about the gate that decides whether a release is
complete, and a type that lies is worse than no type.

## What review found, and why it is worth reading

Four rounds, four real defects. Three came from independent reviewers, one I found by writing my own
enumeration down and attacking it. Recorded because the pattern is more useful than any single fix.

Rounds one to three were one class: **a status that is not evidence about the package.**

1. A `fetch` throw was folded in with 404s, so a real partial publish presented as a quiet skip.
2. The throw was split out, but `status !== 200` still counted a 500 as absence. A 5xx on one sibling
   reported PARTIAL and **reddened a healthy release** — the cry-wolf direction.
3. Only 200/404 were classified, but `fetch` **follows redirects**, so a 302 onto a generic 200 read
   as published. A **false presence**, the only direction that cuts a Release for a version that
   never published.

Each fix closed the reported route and left the same fact arriving by another. The rule that ends
the class: **only `200` is presence, only `404` is absence, everything else is no evidence.**

Round four was **not in that class**, which is the part worth keeping. `deadlineMs` was consulted
only *between* polls, after every read had already returned, so it bounded the classification and
never the I/O. A registry that accepts the connection and never sends headers parks the gate inside
one `await`; no value of the budget can bind, and the job's own timeout reds a healthy publish.
No HTTP outcome is misclassified, so no widening of the status enumeration could have reached it.

Each read is now bounded by what is left of a real wall-clock budget — an abort signal so the socket
is released, and a race so termination is this gate's guarantee rather than something the transport
is trusted to keep. A budget expiry joins the throws as no evidence. And **spending the budget counts
as the deadline passing**, without which the bound only moves the hang into the poll loop.

## Verification

- `pnpm smoke:verify-publish-closure` — 52 cells + sentinel, 0 failed
- `pnpm mutation-proof --config bin/smoke/mutations/verify-publish-closure.json` — ****17 of 17 killed**, each red and named on its own cell, every mutant below the baseline of 53 marks**
- `pnpm smoke:mutation-fixtures` — 17 anchors present and unique
- `pnpm typecheck` — clean

Three of the cells drive the hang directly rather than reading it: an injected fetch that **ignores**
the abort signal on purpose (one that honoured it would prove only that Node aborts, which is Node's
property and not this gate's), a frozen injected clock, and the shipped command against a real TCP
peer that accepts the connection and writes nothing, asserting exit 2 and `signal === null` so a kill
cannot pass as a verdict.

Two notes against my own work, since both would otherwise be invisible:

- That last cell **failed on its first run reporting `published=21/21` against a socket that answers
  nothing.** It had reused the suite's env helper, which injects a fetch fixture, so the child never
  opened a socket. It now builds its own fixture-free environment, and the comment records why the
  helpers must not be merged back together.
- The round-four fix **rotted the round-three mutation anchor**, which named a call site the fix
  moved. `smoke:mutation-fixtures` caught it. That check proves anchors are present and unique and
  says in its own output that this does not prove any still kills, so every mutation was re-driven
  rather than re-checked.

## Not wired up here, and why

The call site is in `.github/workflows/changesets.yml`, which **cannot be pushed from the environment
this was developed in** — the OAuth token has `admin:public_key, gist, read:org, repo` and no
`workflow` scope, so the push is refused. Verified both ways: a workflow path is refused and a
`scripts/` path in the same push is accepted.

The exact one-line change is recorded on #1254 so it survives this PR being closed or rebased. It
replaces the `npm view` retry block:

```bash
node scripts/verify-publish-closure.mjs "$version"; closure=$?
case "$closure" in
  0) : ;;                                                    # published — cut the Release
  1) echo "Partial publish — refusing to cut a Release."; exit 1 ;;
  *) echo "Nothing published or unsettled — skipping the Release."; exit 0 ;;
esac
```

## No changeset, deliberately

Nothing published changes. The root package is private, and no published package's `files` array
ships `scripts/` or `bin/smoke/`. Since the `fixed` group versions all 21 packages in lockstep, a
changeset here would bump every one of them for a CI-only change.


davidfarah2003:
**BLOCK** at `12b8f3c4bc12efac10e8abfe5b36ff798adb9c2b`.

Independent check of that SHA (worktree re-read before this comment; `gh pr view` headRefOid matches). I did not write this change.

## What I ran

- `pnpm smoke:verify-publish-closure` at this head: 33 cells + sentinel, 0 failed (34 passed).
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/verify-publish-closure.json`: **9/9 KILLED**, each red and named on its own cell. Tree restored clean afterwards.
- `emitDeclaration()` vs committed `scripts/verify-publish-closure.d.mts`: byte-identical (3543 bytes). Fragment `bin/smoke/ci-suites.d/4a14f302abdb26318f6848c2951c370a75113992ede15a5791d29a9916bf193b.txt` is the SHA-256 of `smoke:verify-publish-closure`.
- Live registry: `node scripts/verify-publish-closure.mjs 0.41.2 --json` → `published`, exit 0, 21/21 on one read.
- Adversarial in-process sequences against `verifyClosure` / `classify` / the shipped command (not the suite's scripted fetch). Those are the source of the block.

Author-reported numbers match. Repeating them is not the finding.

## The block: transport errors are 404s in the stability set

`readClosure` catches a fetch throw and pushes the package into the same `missing` array as a non-200. The comment above that catch is the contract I tested:

> A transport error is NOT evidence the package is absent; treat it as "not seen yet" so a network blip reads as lag rather than as a partial publish. It cannot mask a real failure: a package that never appears keeps the set non-empty and trips the stable window anyway.

The second sentence is false. Throws and 404s are the same membership. I constructed both masks:

1. **Genuine partial + throws on a live sibling.** `d` always 404; `a` throws on alternate scans. The missing key oscillates (`d` vs `a,d`), so `unchangedSince` keeps resetting. At the default 900s deadline the gate returns **`unsettled`**, not `partial`. Documented wiring on #1254 maps that to `*) … exit 0`. The dangerous case is a quiet skip.

2. **Every fetch throws for the whole window.** Missing length equals total, the set is unchanged, and the gate returns **`none`** (exit 3). Same wiring: skip, not a failure. A live full publish behind a registry outage is skipped. A genuine partial behind that outage is also `none`, never `partial`.

A later clean scan still becomes `partial` (blip then stable 404 on `d` was caught). So a *finished* blip is tolerated. An *ongoing* error is not "not seen yet"; it is recorded as absence, and it is enough to keep a real partial from ever looking stable.

No cell and no mutation distinguish throw from 404. That is the decision the suite does not touch.

The fail-safe direction for *cutting* a Release is intact: throws cannot empty the missing set, so they cannot produce `published`. The fail-loud direction for a partial is not. This is the same quiet-skip class #1254 already measured on packument lag, reintroduced at the catch.

What I would need to approve: errors in a separate set from 404s; any unclean scan stays `polling` / `unsettled` and cannot become `none` or `partial`. Then the comment's guarantee would be true.

## Stability rule (not the block, but it is the heart of the gate)

Two observations, not a distribution. The PR says so. Defaults: `stableWindowMs = 300_000` over a measured ~2 min settle. I replayed a 0.41.2-shaped hold (identical missing for 60s at 15s polls, then shrink, then empty) at production timings: **`published`**, 10 reads `2>2>2>2>2>1>1>1>1>0`. The "two identical reads means stable" trap does not fire inside the measured window.

The other direction also holds for a *static* missing set: unchanged for the full window is `partial` and names the package.

What the window will still get wrong, by design:

- Lag longer than 5 minutes that then clears is `partial` (false alarm). Argued, not measured. I am not blocking on a disclosed margin.
- A set still shrinking at 900s is `unsettled`. That is a genuine partial reported as "cannot tell" if npm is still propagating at the deadline. Same disclosed trade.

`parseOptions` only requires `stableWindowMs >= pollIntervalMs`. `--stable-window-ms=600000 --deadline-ms=300000` is accepted, and a missing set at the deadline is then **`unsettled` forever** (`partial` is unreachable). Defaults are 300s / 900s so this is an operator footgun, not the default path. Mentioning it because nothing refuses it.

## Four exit codes vs the #1254 wiring

Shipped command (fixture registry, not npm):

| case | exit | documented `case` | effect |
| --- | --- | --- | --- |
| all 200 | 0 | `0) :` | cut Release |
| one sibling missing, held | 1 | `1) exit 1` | red, no Release |
| whole closure missing, held | 3 | `*) exit 0` | skip, not red |
| no version argument | 2 | `*) exit 0` | skip |
| `unsettled` | 2 | `*) exit 0` | skip |

A version-PR push is `none` / 3, so it does not red and does not cut. Collapsing `none` onto 0 would recut the phantom Release; that split is load-bearing and the suite kills the regression.

`unsettled` is never produced through the shipped command (the fixture missing set is static). `classify` / in-process `verifyClosure` cover the state. That is dependence, not reach of `EXIT.unsettled` in `main`.

Wiring is still the old `npm view cotal-ai@…` loop in `.github/workflows/changesets.yml`. I am not blocking on that omission. One note for whoever pastes the snippet: the `version` job is `timeout-minutes: 15` and `deadlineMs` is 900_000. The UNSETTLED path can lose to the Actions timeout (job red) before the script can skip. That is a wiring-time number, not a reason to reject this script.

## Env scrub (the freeze delta)

`childEnvFor()` copies `process.env`, deletes `COTAL_*`, then that object is what `spawnSync` gets. Census `smoke:suite-ambient-env` is the class-level guard and would fail this file without the strip. The local cell is not decoration: mutation 9 ("the COTAL_ scrub stops matching…") goes red on `no COTAL_ variable reaches the spawned child, even when one is set in this process`. I am not blocking on a second lock for a suite-only helper.

## What this now correctly rejects

- `cotal-ai` live and a sibling missing (the #1254 one-package pass).
- Two identical lag reads as `partial`.
- Deadline-while-shrinking as `partial`.
- A version-PR as `partial` (exit 1).
- `none` sharing exit 0 with `published`.

Those are real. The transport merge is also real, and it is why this is not an approve.

Not run: any `*-live` suite, `cotal up` / `cotal down`, `pnpm cotal`.

davidfarah2003:
**BLOCK** at `4a9db36612f18e7fff28df05f6e4dffaabc32936`.

Independent re-grade of that SHA after the throw-split fix (worktree and `gh pr view` headRefOid both match). I did not write this change.

The previous BLOCK at `12b8f3c4` stands as history. This comment is only about the new head.

## What the throw-split actually closed

I re-ran the two sequences that blocked `12b8f3c4`, through `verifyClosure` and through the shipped command (`main()` + `--import` fixtures, not the suite's scripted fetch):

| sequence | at `12b8f3c4` | at `4a9db366` |
| --- | --- | --- |
| every `fetch` throws for the window | `none` / exit 3 | `unsettled` `why=transport` / exit 2, `missing=[]`, `errored=all 21` |
| real 404 on one pkg + `ECONNRESET` flapping on a live sibling | `unsettled` (missing-set churn) | `unsettled` `why=transport` (or `deadline` on a clean last scan); never `none` / `published` |
| blip then a clean held 404 | `partial` | `partial` (still) |
| unclean scan, empty missing | could have been `published` | `polling` / cannot `published` |
| lag 2>2>2>2>2>1>1>1>1>0 at prod timings | `published` | `published` (still) |

Smoke: 40 cells + sentinel, 0 failed. Mutation-proof: **11/11 KILLED**, including the two new ones (merge errored back into missing; `clean = true`). dts emit byte-identical (4415). Default `deadlineMs` is 600_000. `parseOptions` now refuses `deadlineMs < stableWindowMs`.

Mask 2 is closed. The `published` extension did not break the monotone-empty path. Mask 1 remaining a quiet skip under the #1254 `*) exit 0` wiring is the disclosed trade; I am not blocking on it. Making that PARTIAL would red ordinary lag.

## The block: `status !== 200` is still "absent"

`readClosure` now splits throws into `errored`. Every other non-200 still goes into `missing`:

```js
if (status !== 200) missing.push(pkg);
```

That is the same class of defect as the last BLOCK, on the status path the throw-split did not touch. A 500 or 429 is not evidence the version is unpublished. I constructed both original masks with HTTP instead of `throw`:

1. **Sibling 500, held for the window.** Shipped command: `state=partial` missing=`@cotal-ai/tmux`, **exit 1**. A live publish behind a 5xx on one origin is a **false PARTIAL** and reds the job. That is the cry-wolf direction the header says is worse than the hole.
2. **Every response is 500, held for the window.** Shipped command: `state=none`, **exit 3**. A total 5xx outage is reported as published-nothing. Same quiet skip as original mask 2, with a status code instead of a throw.

429-churn is the same membership oscillation 5xx shares with the old throw-churn, except it never enters `errored`, so the new restart-on-error rule does not apply.

No cell and no mutation distinguish 5xx/429 from 404. The two new mutations only cover the `catch` path.

What I would need to approve: only `200` is present and only `404` is absent. Any other status joins `errored`. Then an unclean scan (throw *or* 5xx *or* 429) cannot become `published`, `none`, or `partial`.

## Not blocking

- Mask 1 + #1254 skip: disclosed, and I agree it is not closable without reddening lag.
- `reads[].published = packages.length - missing.length` still counts an `errored` package as published. Display-only; `classify` does not use that field.
- Flapping cell is `!== none && !== published`, so it would also pass on `partial`. The timings of that cell do not actually settle to `partial`; the weakness is the assertion, not a false green on this head.
- Workflow still unwired. Deliberate.

Not run: any `*-live` suite, `cotal up` / `cotal down`, `pnpm cotal`.

davidfarah2003:
**BLOCK** at `a0cfab08c9fded62f4f2a0cddf60e6ca447b23c0`.

Independent re-grade of that SHA (worktree and `gh pr view` headRefOid both match). I did not write this change.

The previous BLOCKs at `12b8f3c4` and `4a9db366` stand as history. This comment is only about the new head.

## What the 200/404 split actually closed

I re-ran the sequences that blocked `4a9db366`, through `verifyClosure` and not through the suite's `statusFetch`:

| sequence | at `4a9db366` | at `a0cfab08c` |
| --- | --- | --- |
| sibling HTTP 500, held | `partial` / exit 1 | `unsettled`, `errored=["d"]` |
| every response 500 | `none` / exit 3 | `unsettled` |
| held 404 on one package | `partial` | `partial`, still names it |
| all 404 | `none` | `none` |
| 410 on one package | would have been `partial` | `unsettled` (not absence) |
| 204 on one package | would have been `published` | `unsettled` (not presence) |
| held 429 on one otherwise-live package | n/a | `unsettled why=transport` |
| empty-body HTTP 200 | `published` | `published` (body still unread) |
| `reads[].published` with one 500 | counted the errored pkg | 3/4 |

Smoke: 46 cells + sentinel, 0 failed. Mutation-proof: **13/13 KILLED**, including revert-to-`status !== 200` and "404 is 410". dts emit still byte-identical.

The 5xx class is closed. Held 404 / all-404 regression guards hold. I am not blocking on Mask 1 remaining a quiet skip under #1254 wiring.

## The block: default `fetch` follows 302 to a 200 for a different resource

`readClosure` calls:

```js
const res = await fetchImpl(versionUrl(...), { method: "GET" });
```

No `redirect: "manual"`. Node's default `fetch` follows redirects. I stood up a local HTTP server that 302s every `/<pkg>/<version>` to `/ok` (a generic 200 JSON page) and pointed `verifyClosure` at it with the real `fetch`:

```
raw fetch 302-target: { status: 200, redirected: true, url: '.../ok' }
manual fetch:         { status: 302, redirected: false }
verifyClosure:        state=published, published=4/4
```

That is a **false presence**. It is the one direction that cuts a Release. Every previous hole routed to skip or a false red.

`--registry` is an operator knob. A mirror or proxy that redirects a missing version onto a CDN 200 satisfies the whole closure. Against `registry.npmjs.org` this is latent (missing versions 404). Latency on the default host is not a property of the gate.

`redirect: "manual"` makes the 302 a status, which this head already classifies as `errored`. Checking `res.redirected` is equivalent and still follows the network hop. Prefer not following.

No cell and no mutation cover redirects. I reproduced this independently on a real HTTP server, not by trusting the author's later self-report.

What I would need to approve: the shipped `fetchImpl` default must not treat a followed redirect as presence. Then a 302-to-200 mirror cannot return `published`.

## Not blocking

- Body of a 200 is unread. A 200 with `{}` still counts as the version. Same as npm's per-version endpoint contract; not this round's class.
- 429 vs 503 vs throw remain one `errored` bucket. Deliberate, and I agree they are all no-evidence.
- Workflow still unwired. Deliberate.

Not run: any `*-live` suite, `cotal up` / `cotal down`, `pnpm cotal`.

davidfarah2003:
**APPROVE** at `c266f9af9ba5e9ef114edb011b31ea59cdcc02f4`.

Independent re-grade of that SHA (worktree and `gh pr view` headRefOid both match). I did not write this change.

Previous BLOCKs at `12b8f3c4`, `4a9db366`, and `a0cfab08c` stand as history. This comment is only about the new head.

## What I ran

- `pnpm smoke:verify-publish-closure`: 48 cells + sentinel, 0 failed (49 passed).
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/verify-publish-closure.json`: **14/14 KILLED**, including the mutation that deletes `redirect: "manual"` and reds `a registry that redirects onto a generic 200 does NOT report published`. Tree restored clean.
- Same local HTTP server that blocked `a0cfab08c` (every `/<pkg>/<version>` 302s to `/ok` 200). Default `fetch` still follows (`status: 200, redirected: true`). `verifyClosure` using the gate's default `fetchImpl` now returns **`unsettled`**, `errored=[a,b,c,d]`, `published=0`. Not `published`.
- Live registry: `node scripts/verify-publish-closure.mjs 0.41.2 --json` → `published`, 21/21, exit 0.

## What this head closes

The last BLOCK was false presence via a followed redirect, the only direction that cuts a Release. `readClosure` now passes `redirect: "manual"`. A 3xx is a status this file already treats as no-evidence.

The redirect cell is not a hand-fed 302. The fixture returns 200 unless `init.redirect === "manual"`. Deleting the option makes the cell go red. That is the right shape.

The 200/404 split from `a0cfab08c` still holds on this tree: sibling 500 and total 5xx are `unsettled`; held 404 is still `partial` and names the package; all-404 is still `none`.

## Disclosed, not blocking

- Mask 1 (real 404 behind flapping transport) is still `unsettled` and still a quiet skip under the #1254 `*) exit 0` wiring. Making that PARTIAL would red ordinary lag.
- Empty-body HTTP 200 is still presence. Out of this PR on purpose; filed as #1257.
- Workflow still unwired. Deliberate; do not block on it.
- This APPROVE is a judgement on `c266f9af`, not a claim that the "no evidence" class is exhausted. Three prior rounds looked closed.

Not run: any `*-live` suite, `cotal up` / `cotal down`, `pnpm cotal`.

davidfarah2003:
**BLOCK** at `c266f9af9ba5e9ef114edb011b31ea59cdcc02f4`.

I re-read the worktree and PR head before grading. The exact head is checked again below immediately before posting.

## Blocker: a hanging response defeats the gate's deadline and can red a healthy release

`deadlineMs` is checked only after `readClosure()` finishes. Each package fetch has no abort signal tied to that deadline, and the reads are sequential. A registry connection that accepts the request but does not return headers leaves the gate inside `await fetchImpl(...)`; it cannot reach `unsettled` or emit exit 2 when its configured deadline expires.

I drove this through the shipped command against a local TCP server that accepted the connection and deliberately sent no HTTP response:

```text
node scripts/verify-publish-closure.mjs 9.9.9 \
  --registry=http://127.0.0.1:<port> \
  --poll-interval-ms=0 --stable-window-ms=50 --deadline-ms=100
```

After 1 second, ten times the configured deadline, the process had printed only:

```text
closure: 21 packages in the fixed group
```

It had to be killed externally (`timeout` exit 124). Any eventual runtime-level fetch timeout is outside `deadlineMs`, and sequential requests can consume it once per package. In the intended workflow, the 15-minute job timeout can therefore kill the job before this gate's 10-minute `UNSETTLED` path. A healthy fully published release behind hung mirror responses becomes a red job rather than the disclosed quiet skip. Lines 55-57 specifically claim the shorter deadline leaves room to finish and report cannot-tell, but it is not an I/O deadline.

What I would need to approve: bound each request by the remaining overall budget with an abort signal, classify expiry as `errored`, and prove a never-resolving fetch reaches `unsettled` without relying on an outer process or job timeout. No existing mutation touches the decision to await an unbounded fetch.

## HTTP outcome matrix I drove against the real gate

I used a real local HTTP server and the actual `verifyClosure`; the hang used the shipped CLI entry point. For non-404 status classes, I drove both a whole-closure response and one bad sibling among otherwise-200 packages.

| Outcome | Observed gate result | Assessment |
| --- | --- | --- |
| Direct HTTP 200 | `published` | Defensible under the current status-only contract. |
| 202, 204, 206 | `unsettled`, `why=transport` | Defensible. No false presence or false partial. |
| 301, 302, 307, 308 with production `redirect: manual` | `unsettled`, `why=transport` | Defensible. The redirect fix closes the prior false-presence route. |
| Same redirects when deliberately followed onto a generic 200 | `published` | Reproduces the old false presence, but production Node fetch now receives `redirect: manual`. |
| 404 on all packages | `none` | Defensible disclosed skip. |
| 404 on one sibling, stable | `partial` | Defensible loud failure. |
| 400, 401, 403, 410, 429 | `unsettled`, `why=transport` | Defensible no-evidence classification. No false partial on an otherwise healthy closure. |
| 500, 502, 503 | `unsettled`, `why=transport` | Defensible. The previous false-partial route is closed. |
| Thrown transport error (`ECONNREFUSED`) | `unsettled`, `why=transport` | Defensible. |
| Accepted connection with no response | No state by `deadlineMs`; still pending at 10× the budget | **Blocking false-red route.** |
| 200 with empty body | `published` | False presence possible, but explicitly known as #1257 and excluded from findings here. |
| 200 with malformed body | `published` | Same known #1257 gap. |
| 200 body naming `8.8.8` while requesting `9.9.9` | `published` | Same known #1257 gap. |

Fetch does not expose an informational 1xx as the final response used by this gate. It proceeds to a final response, rejects unsupported upgrade behavior, or remains pending. Those routes are covered above by final status, throw, and hang.

## Stability and mirror assessment

The five-minute stability window is a 2.5× margin over the largest of only two observations. That is an operational argument, not a distribution or percentile. The PR discloses that limitation and prioritizes avoiding false alarms, so I do not treat the asserted margin as a separate finding.

`--registry` correctly exercises the same status and redirect classification against a mirror. The hang defect matters more for mirrors because an operator-controlled proxy can accept a connection without completing it.

## Verification

- `pnpm smoke:verify-publish-closure`: 49 passed, 0 failed.
- `pnpm mutation-proof --config bin/smoke/mutations/verify-publish-closure.json`: 14/14 killed, including the redirect mutation.
- `pnpm typecheck`: passed.
- Full adversarial HTTP matrix above: driven independently, not inferred from the authored cells.

Not run: any `*-live` / `:live` suite, `cotal up`, `cotal down`, or `pnpm cotal`.

davidfarah2003:
**APPROVE** at `9db48b20fe90650635105860493045d699d85d8b`.

Independent re-grade of that SHA (worktree re-read immediately before this comment; `gh pr view` headRefOid matches). I did not write this change.

Previous BLOCKs at `12b8f3c4`, `4a9db366`, `a0cfab08c`, and `c266f9af` (hanging-socket) stand as history. This comment is only about the new head.

## What I ran

- `pnpm smoke:verify-publish-closure`: 52 cells + sentinel, 0 failed (53 passed).
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/verify-publish-closure.json`: **17/17 KILLED**, each red and named on its own cell. `pnpm smoke:mutation-fixtures` reports 17 unique anchors for this config.
- Live registry: `node scripts/verify-publish-closure.mjs 0.41.2 --json` → `published`, 21/21, exit 0.
- Independent hanging TCP (not the suite's `deaf` cell): `net.Server` that accepts and never writes, shipped CLI, `--deadline-ms=1500`, 8s outer `spawnSync` cap. Exit **2**, signal null, **1631ms**, stdout `UNSETTLED`, no `PARTIAL PUBLISH`.
- Independent 302 HTTP server (the round-3 plant): Location `/ok` serving 200. `verifyClosure` with **default** `fetchImpl` (Node follows redirects unless told not to) → `unsettled`, `errored=["a","b","c","d"]`, published=0. Not presence.
- `AbortSignal.timeout(0)` does not throw (`aborted=false`). After the first hung read, `remainingMs` is 0; that path is reachable.

## What the hang-timeout actually closed

At `c266f9af`, `deadlineMs` was consulted only *after* `readClosure()` returned. A peer that accepted the TCP handshake and never sent headers parked the gate inside one `await`, so the release job's 15-minute timeout reddened a healthy publish.

At this head:

- `readWithinBudget` races the fetch against a timer *and* passes `AbortSignal.timeout(remainingMs)`. The race is the gate's own bound (a fetch that ignores abort still loses); the signal is the socket release. Deleting either is a distinct plant, and both plants died on named cells.
- `remainingMs = max(0, budgetEndsAt - Date.now())` is wall-clock, not the injected clock. Sequential remaining packages after a hung first read get a shrinking (then zero) budget instead of a fresh `deadlineMs` each.
- Loop elapsed is `max(now() - started, Date.now() - ioStarted)`, so a frozen injected clock cannot keep polling after the real budget is spent. The frozen-clock cell also stops the abandoned run (`abandoned` → next `sleep` rejects), so that plant produces a named red rather than a hang that mutation-proof would score INCONCLUSIVE.

## Round-3 redirect plant, re-anchored

The fetch call site moved into `readWithinBudget`. The plant now deletes `redirect: "manual"` from *that* call. I did not take the suite's word:

- Mutation-proof killed it on `a registry that redirects onto a generic 200 does NOT report published` (51 marks vs baseline 53).
- The independent 302 server above is the same shape that blocked `a0cfab08c`: default fetch follows, this gate does not. False presence is still witnessed.

## Disclosed, not folded

- Empty-body 200 is still #1257. A 200 with no body is still presence.
- Workflow unwiring is still deliberate; this grades the gate, not the job that will call it.
- Mask-1 quiet skip under #1254 (`*) exit 0`) is still a disclosed trade, not this PR's hole.

## Not blockers

- `AbortSignal.timeout(0)` not aborting immediately is Node's timer, not this gate's. The `Promise.race` timer still rejects at `remainingMs === 0`.
- Mutation-proof left two plants in the worktree after the run (`versionUrl` encoding, dts header). I restored both before posting. That is a tool restore miss, not a gate miss. HEAD after restore is still `9db48b20`.

I would merge this.

davidfarah2003:
**APPROVE** at `9db48b20fe90650635105860493045d699d85d8b`.

Independent re-grade of the fixed head. I re-ran the old matrix because the fetch call site changed, then drove the new budget boundaries rather than inheriting the authored proof.

## The timeout defect is closed

The whole wall-clock budget now bounds each read. The abort signal releases a compliant transport, while the race independently bounds the gate when an injected or broken transport ignores abort. Expiry joins `errored`, and real elapsed time ends the run even with a frozen injected clock.

I drove both synthetic and real-socket mixed scans. With `a=200`, `b=404`, `c` accepting and never returning headers, and `d` unread after the budget was spent, the shipped logic returned:

```text
state=unsettled why=transport
missing=[b]
errored=[c d]
wall elapsed=257ms for deadlineMs=250
```

That is the defensible result. It preserves the evidence actually received, does not promote the unread siblings to presence or absence, and terminates under its own budget rather than an outer kill.

An incomplete header block also returned `UNSETTLED`, exit 2, inside the budget. A complete `200` header followed by a stalled body returned `PUBLISHED`; that is the disclosed status-only boundary already filed as #1257, not a regression in this fix.

## Full outcome space re-driven

All 36 prior matrix cells matched the reviewed behavior at this head:

- 202/204/206, 301/302/307/308 in manual mode, 400/401/403/410/429, 500/502/503, and thrown transport errors: `unsettled`.
- Redirects deliberately followed onto a generic 200: `published`, demonstrating why production `redirect: manual` remains load-bearing.
- All 404: `none`; one stable 404 sibling: `partial`.
- Direct 200, empty/malformed 200 bodies, and a 200 body naming another version: `published`. The latter three remain exactly #1257.

No false presence or false partial appeared outside that disclosed body-validation boundary.

## Design judgement

**The status enumeration is now closed for the contract this PR implements:** exact 200 is presence, exact 404 is absence, and every other final HTTP outcome or transport failure is no evidence. Redirect handling and the wall-clock I/O bound make those arms enforceable rather than merely stated.

**Status-only is not ultimately sound for the stronger claim that the requested package version is genuinely present.** A 200 can still carry no usable representation or name another version. That cannot be fixed by adding another status arm; it requires reading and validating the body, which is #1257. I approve this PR because that boundary is disclosed and separately scoped, but I would not call the broader publish-proof design complete until #1257 is resolved.

## Verification

- Independent HTTP matrix: 36/36 expected outcomes.
- Real incomplete headers: exit 2 inside budget.
- Real mixed mid-closure expiry: `unsettled`, observed 404 preserved, unread packages errored.
- Frozen-clock mixed expiry: same classification inside the real budget.
- `pnpm smoke:verify-publish-closure`: 53 passed, 0 failed.
- `pnpm smoke:mutation-fixtures`: green; all 17 publish-closure anchors present and unique.
- `pnpm mutation-proof --config bin/smoke/mutations/verify-publish-closure.json`: 17/17 killed and named.
- `pnpm typecheck`: passed.

Not run: any `*-live` / `:live` suite, `cotal up`, `cotal down`, or `pnpm cotal`.

davidfarah2003:
**APPROVE** at `9db48b20fe90650635105860493045d699d85d8b`.

Independent re-grade of that SHA (worktree re-read immediately before this comment; `gh pr view` headRefOid matches). I did not write this change.

Previous BLOCKs at `12b8f3c4` (throws folded into missing), `4a9db366` (non-200 as absence), `a0cfab08c` (followed 302 as presence), and `c266f9af` (deadline bound classification, not I/O) stand as history. This comment is only about the new head.

## What I ran

- First `pnpm smoke:verify-publish-closure` on a supposedly clean tree failed one cell: `spending the real budget ends the run even when the injected clock has not moved` returned `{ state: 'HUNG' }`. The tree then showed a leftover plant in `scripts/verify-publish-closure.mjs` from an earlier mutation-proof run (the `readWithinBudget` abort path). Restored with `git checkout -- scripts/verify-publish-closure.mjs`. Isolated frozen-clock against HEAD three times: `unsettled`, ~305ms each. Second suite run: **53 passed, 0 failed** (52 cells + sentinel).
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/verify-publish-closure.json`: **17/17 KILLED**, each red and named on its own cell. Baseline 53 marks. `pnpm smoke:mutation-fixtures` reports 17 unique anchors for this config.
- Live registry: `node scripts/verify-publish-closure.mjs 0.41.2 --json` → `published`, 21/21, exit 0.
- Independent hanging TCP (not the suite's `deaf` cell): `net.Server` that accepts and never writes a byte, shipped CLI, default `fetch`.
  - `--deadline-ms=400`: exit **2**, signal null, **481ms**, `state=unsettled`, `published=0`, 21 packages in `errored`.
  - `--deadline-ms=100`: exit **2**, signal null, **197ms**. Sampled at 1s: **not pending** (already settled at 200ms, exit 2, 21 errored). Author claim holds: a 100ms budget is not still waiting at 1s.
- Contrast, same deaf peer, `verifyClosure` imported from a copy with the race *and* abort stripped: **still pending at 1s**, still `HUNG` at 1505ms. The bound is this gate's, not Node's default fetch timeout.
- Independent 302 HTTP server (the round-3 plant, re-anchored into `readWithinBudget`): Location `/ok` serving `{}`. Raw Node `fetch` → `status=200`, `redirected=true`. `verifyClosure` with default `fetchImpl` → `unsettled`, `errored=["a","b","c","d"]`, `published=0`. Followed redirect is not presence.
- `pnpm typecheck`: passed.

Not run: any `*-live` / `:live` suite, `cotal up`, `cotal down`, or `pnpm cotal`.

## What the hang-timeout actually closed

At `c266f9af`, `deadlineMs` was consulted only after `readClosure()` returned. A peer that accepted the TCP handshake and never sent headers parked the gate inside one `await`, so the release job's 15-minute timeout reddened a healthy publish.

At this head:

- `readWithinBudget` races the fetch against a timer *and* passes `AbortSignal.timeout(remainingMs)`. The race is the gate's own bound (a fetch that ignores abort still loses); the signal is the socket release. Deleting either is a distinct plant, and both plants died on named cells.
- `remainingMs = max(0, budgetEndsAt - Date.now())` is wall-clock, not the injected clock. Sequential remaining packages after a hung first read get a shrinking (then zero) budget instead of a fresh `deadlineMs` each.
- Loop elapsed is `max(now() - started, Date.now() - ioStarted)`, so a frozen injected clock cannot keep polling after the real budget is spent. The frozen-clock cell also stops the abandoned run (`abandoned` → next `sleep` rejects), so that plant produces a named red rather than a hang that mutation-proof would score INCONCLUSIVE.

## Round-3 redirect plant, re-anchored

The fetch call site moved into `readWithinBudget`. The plant now deletes `redirect: "manual"` from *that* call. I did not take the suite's word:

- Mutation-proof killed it on `a registry that redirects onto a generic 200 does NOT report published` (51 marks vs baseline 53).
- The independent 302 server above is the same shape that blocked `a0cfab08c`: default fetch follows, this gate does not. False presence is still witnessed.

## Classification (status contract)

Only 200 is presence. Only 404 is absence. Throws, budget expiry, 3xx-without-follow, 5xx, 429, 403 land in `errored`. An unclean scan cannot become `published`, `partial`, or `none`. That is the contract this PR implements, and the plants that reverse it died.

## Disclosed, not folded

- Empty-body 200 is still #1257. A 200 with no body, or a 200 body naming another version, is still presence. Status-only cannot close that; body validation is a separate issue.
- Workflow unwiring is still deliberate. This grades the gate, not the job that will call it.
- Mask-1 quiet skip under #1254 (`*) exit 0`) is still a disclosed trade: a genuine partial behind a flapping transport stays `unsettled` and would skip rather than red. Not this PR's hole.

I would merge this.

davidfarah2003:
**Merge basis** — head `9db48b20fe90650635105860493045d699d85d8b`.

Two independent terminal approvals, both naming this exact SHA, from different model families, neither
inheriting an earlier pass (the fetch call site moved, so nothing transferred):

| verdict | comment |
| --- | --- |
| APPROVE | https://github.com/Cotal-AI/Cotal/pull/1255#issuecomment-5545065345 |
| APPROVE | https://github.com/Cotal-AI/Cotal/pull/1255#issuecomment-5545110393 |

CI run [33906897016](https://github.com/Cotal-AI/Cotal/actions/runs/33906897016) — 16/16 green at this
head. `smoke (shard 1/4)` failed once on `smoke:jcode-host` and passed on retry; that is #1179, not
this branch. Ruled out rather than assumed: the suite passes at this branch's exact base `162468ed67`
and at `52d801792d` and `634746ab87`, this PR touches no jcode code, and the shard-1 suite set is
identical between main and this head (120 suites, zero added, zero removed). The occurrence is logged
on #1179.

A third approval also landed at this head. It is recorded as corroboration rather than as part of the
basis, because its origin could not be established.

**What the review found, and what remains open.** Four rounds produced four real defects. Three were
one class — a status that is not evidence about the package — closed by the rule the gate now states
and enforces: only `200` is presence, only `404` is absence, everything else is no evidence. The
fourth was on a different axis and no widening of that enumeration would have reached it: the deadline
bounded classification rather than I/O, so a registry that accepted a connection and never sent
headers could outlive any budget and red a healthy release.

The reviewers' terminal design judgement, recorded on #1257: the status enumeration is now closed for
the contract this PR implements, but status-only cannot establish that the requested version is
genuinely present, because a `200` can carry no usable representation and no further status arm
reaches that. That is a claim about the body, and it is #1257's scope, not this PR's.

Also open by design: the gate is not wired into `.github/workflows/changesets.yml`, for the reason
given in the description, and the exact call-site change is recorded on #1254 so it survives this PR.

davidfarah2003:
Merge basis. Head `9db48b20f`, unmoved since both verdicts. Independent cross-family APPROVEs at this exact sha: rev1255 (grok-4.6, comment 5545065345) and rev1255b (gpt-5.6-sol, comment 5545110393), both re-driven from scratch after the fourth repair; a third APPROVE (comment 5545183938) at the same sha from a seat neither of us commissioned is recorded as corroboration only, not counted toward independence. CI run 33906897016: attempt 1 red only on shard 1 `jcode-host` (#1179, prefiled; base and two sibling main heads green on that suite, shard-1 suite set identical to main at 120), attempt 2 all green: unit, live, shards 0/1/2/3. History in the PR body. No changeset by design (CI-only; no published package ships scripts/ or bin/smoke/). The workflow wiring needs `workflow` scope the shared identity lacks; the exact line and the `timeout-minutes` interaction are on #1254 for David to apply.

Issue #963, opened 2026-08-28
manifest: no per-agent cwd and no continuity field, so a multi-repo fleet cannot be declared and a restored fleet comes back with empty context


davidfarah2003:
A manifest cannot describe a fleet whose agents live in different repositories, which is most
real fleets, and it cannot say "restore this agent" at all.

## cwd

`cwd` exists as `cotal spawn --cwd` and on the `cotal_spawn` MCP tool, but there is no per-agent
field for it in the manifest. Every agent a manifest declares therefore roots at the manager's
workspace. A fleet whose orchestrator sits in one checkout and whose lane seats sit in others
cannot be written down, so it has to be spawned by hand, one command per seat, which is the state
we are in.

Add a per-agent `cwd`, resolved the same way `--cwd` resolves today: a relative path against the
manager workspace, an absolute path taken as given.

## Continuity

`--resume` is an operator-only CLI flag. A manifest can say "create this agent" and cannot say
"restore this agent's context". After a manager restart the shape of the fleet can be recreated
but every seat comes back with an empty context, so recovery means reading transcript files to
work out which session belonged to which seat.

Add a continuity declaration to the manifest. The operator-only boundary stays intact, because a
manifest is already an operator artifact; the MCP `cotal_spawn` tool is unchanged.

## Why this one is blocking

Boot-time convergence to the last applied manifest is only useful if the manifest can express the
fleet it is meant to restore. Without `cwd` it cannot express a multi-repo fleet, and without
continuity it restores shape without context.

## Done when

A manifest describes a fleet whose agents live in three different repositories, `cotal up -f`
brings it up with each agent rooted correctly, and a declared agent comes back with its previous
context rather than an empty one.

davidfarah2003:
[...]

PR #1264, opened 2026-09-05
feat(cli): declare a per-agent working directory in the manifest

davidfarah2003:
A manifest could not describe a fleet whose agents live in different repositories. `cwd` existed only as `cotal spawn --cwd` and on the `cotal_spawn` tool, so every declared agent rooted at the manager's workspace, and a multi-repo fleet had to be spawned by hand, one command per seat.

## What this adds

An optional per-agent `cwd` on the manifest, carried the whole way through: schema, model, prepare, resolve, apply, the launch spec, `MeshLaunchAgent`, and the manager's launch projection.

```yaml
agents:
  researcher:
    agent: jcode
    cwd: ../other-repo
```

## Resolution

The deploying machine never resolves it. The value is forwarded as written and resolved on the **manager host** — absolute taken as given, relative resolved against the manager workspace — which is exactly what `--cwd` already does. Reusing that resolver rather than adding a second one is deliberate: two resolvers that agree today will disagree later, and the divergence would be found by an operator rather than a test.

A NUL byte is refused at the schema.

## Hashing

`hashAgent` includes `cwd` only when present. Manifests that omit it keep their existing hashes, so this does not invalidate every stored manifest; a manifest that declares or changes a directory produces a new hash and requires a restart.

This matters more than it looks. Had `cwd` been transported but left out of the hash, two manifests declaring different working directories would hash identically — the field would work while change detection stayed blind to it.

## What the test actually asserts

The cell reads `process.cwd()` **reported by a real spawned child**, through the registered launch endpoint, for the relative, absolute and omitted cases. It does not assert on a manager row, because a row can faithfully repeat the requested path while the process runs somewhere else entirely. The typed `manager ps` cwd is checked alongside it, not instead of it.

`implementations/manager/smoke/mutations/manifest-cwd.json` adds two controls:

- **drop the projection** — `cwd: a.cwd` becomes `cwd: undefined` in the manifest launch path;
- **wrong relative root** — the manager resolves a relative `cwd` against its own process directory instead of its workspace.

Both are the failure modes a green suite would otherwise be blind to: the first ships a field that goes nowhere, the second ships one that lands in the wrong place.

## Scope

This is the `cwd` half of #963 only. The continuity half — a manifest declaration that restores a seat's session rather than only creating the seat — is a separate design and **#963 stays open** for it.

**#963 is not closed by this PR.** The `cwd` half lands here; the continuity half keeps the card open, and a closing keyword would have taken both.


davidfarah2003:
**APPROVE** at `e5412a126d43e4dea9980c4bcda7a57e81215d8d`.

The three claims hold. I re-read `gh pr view 1264 --json headRefOid` immediately before posting; the head has not moved. I did not run `pnpm smoke:manager-service-ops` (it boots a broker). Claim 1 is from the cell as written, not from watching a child in this worktree.

## 1. Schema to resolved launch actually transports it

It does. `implementations/manager/smoke/manager-service-ops.smoke.ts` does not grade a manager row in place of a process.

- A dedicated `cwd-stub` connector writes `process.cwd()` to an absolute file under the suite temp dir, then imports the e2e stub. The write path is independent of the child's cwd, so a wrong spawn directory cannot hide by failing to write.
- The request is `A.call("launch", { runId, name, spec })`, the registered ep door, not a cast into `opLaunch`.
- The three cases are relative (`repos/worker`), absolute (`join(dir, "external repo")`), and omitted. Expected paths are `resolve(workspaceRoot, cwd)` or `workspaceRoot`. The observation is `readFileSync(child-cwd.txt)` compared to `realpathSync(expected)` **and** `row?.cwd === expected`. Both have to match.
- Launch is an action: the reply is acceptance, returned before the child is live. The cell waits for the file (and the ps row) rather than treating `ok: true` as completion.

A cell that only asserted the row would have passed a dropped field. This one would not.

## 2. The hash

Both directions hold.

`hashAgent` still stringifies the pre-cwd object when `cwd` is omitted (`...(a.cwd !== undefined ? { cwd: a.cwd } : {})`). I recomputed that payload independently:

```text
{"agent":"claude","model":"opus","variant":null,"launchOptions":null,"role":null,"body":null,"capabilities":[],"subscribe":[],"allowSubscribe":[],"allowPublish":[]}
```

SHA-256 prefix `fd4f6aefa0a2402f`, matching the fixture in `implementations/cli/smoke/manifest.smoke.ts` and the `hashAgent` body on `origin/main`. An upgrade does not stale every stored entry.

The same cell then requires `hashAgent(prepared) !== hashAgent(base)` for each declared directory, and `hashAgent(prepared) !== hashAgent({ ...prepared, cwd: cwd + "/changed" })`. A cwd-only edit cannot hash-collide with omit or with another path.

`cwd: null` is not a legal prepared value (schema is `string.min(1)`), so the omit-vs-null trap is closed at the door.

## 3. Resolution is manager-side, one resolver

`resolve.ts` copies `entry.cwd` verbatim. Persona paths still `resolve` against the manifest directory; `cwd` does not. The pipeline smoke asserts the relative, spaced, and absolute strings survive resolve → prepare → `buildLaunchSpec` unchanged, including after `JSON.parse(JSON.stringify(...))`.

The manager still has one line:

```ts
const cwd = opts.cwd ? resolve(this.workspaceRoot, opts.cwd) : this.workspaceRoot;
```

Detached `cotal spawn --cwd` already forwarded `values.cwd` unresolved (`spawn.ts`) into that line. Manifest launch reaches it through `launchAgentToStartOpts` (`cwd: a.cwd`) and the same `startAgent`. No second resolver.

`runtime.spawn(name, spec, cwd)` is what actually roots the PTY. `LaunchOpts.workspaceRoot` stays the manager workspace on purpose, so connector-local state does not follow the per-agent directory.

## Mutations

`implementations/manager/smoke/mutations/manifest-cwd.json`:

- **drop-projection** mutates the only `cwd: a.cwd,` in `launch.ts`. Launch acceptance still returns ok (action reply is pre-spawn). The first cell that can go red is `manifest cwdrelative runs in the declared manager-host directory`. That is the named `expectRed` / `cell`.
- **wrong-relative-root** mutates the only `opts.cwd ? resolve(this.workspaceRoot, opts.cwd)` line. The suite workspace is `join(mkdtemp(...), "ws")`, not `process.cwd()`, so `resolve("repos/worker")` cannot accidentally equal `resolve(workspaceRoot, "repos/worker")`. Same first cell. Absolute and omitted cases would still pass this mutant; naming the relative cell is required, and it is first.

This suite is a counting harness (`check` does not throw). The named cell is still the first one these mutants redden. `find` strings are unique in their files.

## Changeset

Minor on `@cotal-ai/cli`, `@cotal-ai/core`, `@cotal-ai/manager` is right. Adding an optional field to a strict launch object is additive for same-version readers: omit still parses, omitted hashes are unchanged, so already-deployed agents are not stale. Packages are 0.x and lockstep; an old manager's `z.strictObject` would reject a new spec that carries `cwd`, and that mixed pair is not a supported matrix. This is not a wire-contract break.

## Docs bundle

The `docs-bundle.generated.ts` hunk is a single `manifest` page `body` replacement. It contains the new example line, the `cwd` key in the per-agent list, the manager-host paragraph, and `This field controls the directory only; it does not restore a harness session.` Nothing else moved.

## NUL, and what else to refuse

Empty string, non-string, and NUL are refused at both the manifest schema and the manager launch schema. NUL is the C-string terminator; empty is not a directory. I would not add more at this layer.

Existence, `~` expansion, and `..` escape are manager-host facts, not deployer-schema facts. `..` is the multi-repo case the field exists for. A literal `~` will not expand (`path.resolve` does not), which is a footgun, not a silent wrong-machine resolve. Leave it.

## Continuity / #963

Nothing here claims the other half. Commit and PR say `Refs #963`, not `Closes`. Docs and changeset say the field does not restore a harness session. `startAgent` still takes `resume` only from the imperative arg, "never from `opts.resolved`". The card stays open.

Non-blocking: `docs/manifest.md` says this matches `cotal spawn --cwd`. That is true of the detached path (unresolved, manager-workspace). Foreground `--cwd` still resolves against the invoking shell. Manifests always go through the manager, so the operator-facing sentence is the detached one. Not a continuity claim.

PR #1265, opened 2026-09-05
fix(manager): late turn-yield names the deadline terminal, not not-found

### LINKED ISSUE #1262: manager: a yield after a turn deadline is refused not-found instead of being told the turn ended, intermittently
## Summary

`smoke:turn-relay-auth` fails intermittently on the deadline-expiry arm. Two CI runs on effectively the same tree, ~13 minutes apart, one green and one red, with an identical 13-cell partition:

```
c594bb9bd (#1204 merge)          TURN RELAY AUTH SMOKE OK      (13 passed, 0 failed)
89627f1a2 (0.42.0 release head)  TURN RELAY AUTH SMOKE FAILED  (12 passed, 1 failed)
```

The release head's tree differs from the base only by 24 `CHANGELOG.md` files, 24 `package.json` version fields, one consumed changeset and the docs-bundle version stamp. No product source differs. So the two runs exercised the same implementation and disagreed.

Cell counts match at 13 on both runs, so the red is a **graded failure and not a throw**.

## The failing cell

Section 4, "a turn nobody yields is denied at its deadline: the manager expires the hold it owns, with no fire read":

```
  ok    a short-deadline turn is accepted
  ok    the goal's terminal is the deadline deny, reason `turn-deadline`
  FAIL  a yield after the deadline is told the turn already ended
        {"v":1,"ok":false,"error":{"code":"not-found","message":"no pending turn \"g2bbb…\""}}
```

The refusal happened, and nothing recorded a success over the deny, so the safety property the cell guards did not visibly break on this run. What differed is **which** refusal came back: `not-found` / "no pending turn" instead of a refusal naming the terminal the turn actually reached.

## Why the distinction is worth a card rather than a retry

`turn-yield` is a seat's own report of how its turn ended, and it is reached on the seat's `manager.self` capability. A seat that yields a moment late needs to distinguish two different worlds:

- **the turn hit its deadline and was denied** — the work happened, the run moved on without it, and the seat should stop;
- **no such turn ever existed for me** — an addressing or identity fault, which is a different problem with a different response.

`not-found` collapses both into one answer. That is the same class this repository has repeatedly paid for: a refusal whose text names the wrong mistake, and an instrument that cannot tell "absent" from "ended".

The intermittency points at a race between the deadline expiry committing its terminal and the pending-turn index entry being dropped. A late yield that arrives after the index entry is gone but before (or instead of) reading the terminal gets `not-found`. If that ordering is what the green run and the red run differ on, the fix is to make the yield path read the terminal rather than depend on the index still holding the entry.

## Scope note

This surface is new in 0.42.0 — the turn relay family (`turn`, `turn-pending`, `turn-yield`) is what moved the manager cluster contract to `revision: 10`. **0.42.0 is published**, so this ships in the release. It is not a regression against an older behaviour; it is a defect in newly-shipped code.

Related: the deadline arm is the same one described in the manager service contract's `turn-yield` documentation ("any yield settles the goal `succeeded` with … first-terminal-wins").

## Evidence

- red: `smoke (shard 0/4)` job 101225954682, head `89627f1a22234d658743683096cc3d616044799e`
- green: `smoke (shard 0/4)` job 101225793921, head `c594bb9bd3afb2d06b249839ff0d618248992869`
- `smoke:manager-stop-reap` passed 15/15 in the red run, which is its healthy cell count, so this is not #1226.

davidfarah2003:
Closes #1262

## What failed

`smoke:turn-relay-auth` flake on the deadline arm, same product source on two CI heads ~13 minutes apart. The deny itself held (`turn-deadline`). The late yield after that terminal was visible heard:

```
{"ok":false,"error":{"code":"not-found","message":"no pending turn \"g2bbb…\""}}
```

A seat that yields a moment late has to distinguish **the turn hit its deadline and was denied** (the run moved on) from **no such turn ever existed for me**. `not-found` collapses both.

## Mechanism (the index-drop lead is wrong)

The yield path never consults the durable goal index. After `commitTurnDeadline` deletes the pending entry (its idempotency latch, *before* the terminal CAS) the only in-memory answer a late yield could find was `turnAcceptances.settled`. Two orderings both produce `not-found` while the deny is already on the plane:

1. Terminal committed, index still present, settled answer not yet remembered (the window between `pendingTurns.delete` and `rememberSettledTurn`; also a concurrent sweep's `maybeStopTurnSweep` wiping the acceptance map while the commit is still in flight, so `rememberSettledTurn` no-ops).
2. Terminal committed *and* the index already cleared (the state the auth-mesh smoke's `resultOf` wait actually observes before it yields).

Nothing recorded a success over the deny on the red CI run, and the fixture confirms that: both late yields leave the terminal `failed` / `turn-deadline`.

## What this does

- `serveTurnYield` still serves a remembered settled answer. When that field is missing but this incarnation accepted the turn, it reads the durable terminal and names that state instead of `not-found`.
- The acceptance map keeps the goal `ref` so that read has coordinates after pending is gone.
- `maybeStopTurnSweep` no longer clears the acceptance map the moment pending is empty. An unsettled acceptance is the in-flight deadline commit; prune is the only deleter.
- `commitTurnDeadline` remembers the settled answer before it clears the index.

## Proof

Failing cells first, in `manager-reconcile-startup.smoke.ts`, driving `serveTurnYield` after a shipped `expireTurnHold` + `commitTurnDeadline`. Two goal ids (the index is create-only and cannot be restored after a clear). Both orderings forced by dropping `settled` after the deny is on the plane, not by live timers or broker perturbation.

On today's main those two cells reddened for the CI reason:

```
FAIL ... never `not-found` {"code":"not-found","message":"no pending turn \"ddd…\""}
FAIL ... never `not-found` {"code":"not-found","message":"no pending turn \"eee…\""}
```

After the fix: `MANAGER RECONCILE STARTUP SMOKE OK (23 passed, 0 failed)`.

`pnpm mutation-proof --config implementations/manager/smoke/mutations/late-turn-yield.json` **KILLED** on the earliest named cell:

`a yield after the deny committed, before the index drop, is told the turn ended failed, never \`not-found\``

`pnpm typecheck` green in this worktree.

## Unverified

The original auth-mesh smoke's section 4 is still timing-sensitive (it waits for the terminal then yields, which is usually after remember). The new cells force the window it only hit intermittently. A live multi-manager / successor-incarnation late yield is not covered here.

### DIFF
--- a/.changeset/fix-1262-late-turn-yield.md
+++ b/.changeset/fix-1262-late-turn-yield.md
@@ -1,0 +1,7 @@
+---
+"@cotal-ai/manager": patch
+---
+
+A late `turn-yield` after the deadline deny has committed is told the turn ended `failed`, never `not-found`.
+
+The deny itself held: first-terminal-wins still refused a success over it. What failed was the diagnostic. `commitTurnDeadline` deletes the pending entry (its idempotency latch) *before* the terminal CAS, then remembers the settled answer only after. A yield in that window, or after a concurrent sweep dropped `turnAcceptances.settled`, heard `no pending turn` — which a seat reads as an addressing fault, not "the run moved on". The durable terminal is the answer the run already has; the yield now reads it when the in-memory settled field is missing, and the sweep no longer wipes an unsettled acceptance the moment pending is empty.
--- a/implementations/manager/smoke/manager-reconcile-startup.smoke.ts
+++ b/implementations/manager/smoke/manager-reconcile-startup.smoke.ts
@@ -41,6 +41,8 @@ import {
   bindGoal,
   createGoal,
   recordGoalIndex,
+  readGoalResult,
+  clearGoalIndex,
   type GoalRef,
   epAuthBucket,
   epCall,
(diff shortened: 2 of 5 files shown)


davidfarah2003:
APPROVE - `42b1fce0c2cefc7eb27dc42db3154ce3e47fe33d`

I verified the revised diagnosis in source. `serveTurnYield` reads only `pendingTurns`, `turnAcceptances`, and, when the accepted turn has no in-memory settled answer, `readGoalResult` through the saved goal ref. It does not read the pending goal index. An unknown goal has neither a pending turn nor an acceptance entry and still receives `not-found`. A late yield retains its address check and reports the durable `failed` terminal, so it is distinct from an unknown turn.

The new fixture deterministically stages both post-commit states through the actual yield handler: an already-due owned hold is expired, `commitTurnDeadline` publishes the durable deny, then it explicitly removes pending and settled state. The first case intentionally leaves the index intact; the second explicitly clears it. It uses a throwaway JWT broker on an OS-assigned loopback port and tears it down, with no shared-broker fault injection. The current fixture passed 23/23. Replacing only `manager.ts` with its parent source made both new cells red with `not-found` (21 pass, 2 fail), which proves each ordering detects the old behavior.

I also ran the new mutation proof. It names the first new cell and it was red: baseline 24 progress marks, mutant 22. The two-mark difference is expected because this fixture records successful checks as marks and continues after failures; the independent pre-fix run confirms the named first cell is the earliest new failure, with the second cell failing later. Manager typecheck passed.

The existing reconcile-startup mutation config was re-anchored from optional chaining to the expanded `settled` branch after this source change, then gained the same new late-yield guard. It was not weakened.

PR #1266, opened 2026-09-05
fix(core): complete a committed registration spec after a lost ack

### LINKED ISSUE #1243: manager: after boot self-heal reopens the gate, the re-registration spec-write settles ambiguous and re-freezes it, so takeover never completes
## Summary

After the boot self-heal reopens a frozen registration gate, the manager's own re-registration write can settle as *ambiguous*, which re-freezes the gate and exits the process. Because the gate is frozen again, the next start repeats the whole repair, and the manager never completes takeover without operator intervention.

This is distinct from #457 (the credential not renewed at the 24h boundary) and from #871 (the boot race). Those explain how the gate gets frozen the first time. This is about the repair itself failing to converge afterwards, and it is what turns a recoverable service death into an unrecoverable fleet outage.

## Observed

Version **0.37.0**, which contains `2395efd102cceebc99538442f465028b8724abc2` (landed 0.36.0), so the boot self-heal is present.

Verbatim, from the manager journal:

```
✓ boot self-heal: manager/<instance> registration gate reopened at generation 14
  (processEpoch unchanged at 5; freeze-holder gone, sweepComplete=true). Continuing the normal takeover.
...
✗ the re-registration spec-write outcome is ambiguous (it may have committed);
  the gate is left frozen for reconciliation, never reopened at the old coordinate (SPEC ...)
systemd[1]: cotal-supervise.service: Main process exited, code=exited, status=1/FAILURE
```

The self-heal reports success (`sweepComplete=true`) and the failure comes from the step immediately after it. `smoke:boot-self-heal-gate` passing 12/12 is consistent with this: the self-heal is not the failing component.

## Why it does not converge

Each start redoes the repair from the beginning on a ledger that keeps growing. Since 30-Aug on one deployment:

- **27** successful `boot self-heal ... gate reopened` events
- gate generations climbing monotonically: **7, 8, 10, 12, 13, 14, 17 … 35, 39, 40**
- repair set per attempt: `holders=400` → `495` → `585` → `657` → `727`

A supervisor that restarts the manager on a failed health check makes this permanent: every restart logs `repair cursor: restart op=... (prior cursor discarded)`, so the cursor never advances. With the supervisor suspended, the same repair switched to `repair cursor: resume ... verified=521/727` and completed. That suggests the cursor is only durable across a *clean* attempt, not across a restart of the repairing process.

## Recovery that works

Four occurrences, four successes, always in this order:

1. suspend anything that restarts the manager
2. stop the manager, confirm zero processes remain
3. `cotal reconcile-gate --space <s> --server <url> --endpoint manager --instance <id>`
4. start exactly one manager, wait for `✓ manager up`
5. only then re-enable the supervisor

Step 1 is not optional and is not documented anywhere I could find. Without it, `reconcile-gate` and the restarting manager fight over the same repair and neither finishes.

## Suggested handling

An ambiguous spec-write is a genuine uncertainty, so refusing to guess is right. But the current outcome — re-freeze and exit 1 — has no path out that the process itself can take. Two options that would close the loop:

- resolve the ambiguity on the next boot by reading back whether the write committed, rather than re-freezing on the same coordinate; or
- keep re-freezing, but make the repair cursor survive a process restart so successive attempts make progress instead of restarting a growing sweep.

Either would remove the need for a human to run `reconcile-gate`.

## Environment

Two LXC containers on one Proxmox host: broker (`cotal up`) in one, manager (`cotal supervise --server nats://<broker>:4222`) in another. Auth mode, single space. cotal-ai 0.37.0, node v22.23, roster ~27 agents.

Field evidence and the answer to the signer question raised in #457 are in https://github.com/Cotal-AI/Cotal/issues/457#issuecomment-5534303458 — happy to run an instrumented build against this topology.

davidfarah2003:
Closes #1243.

A manager re-registration whose spec write commits and then loses the ack used to leave the issuance gate frozen and exit 1. Boot self-heal abort-reopened that gate, the next start froze a **new** coordinate, and takeover never converged.

## What this does

Classify the one-key spec write by read-back under the **same** freeze:

- proposed bytes at a successor revision → this write committed; complete promote + reopen for that op (re-registration advances `processEpoch`)
- prior revision / absence → definite no-commit; abort-reopen at generation+1 with the coordinate otherwise unchanged
- unreadable get / garbled / mismatched bytes → stay frozen

Same-attempt read-back in `registerServiceInstance` is the process-survives exit. `completeFrozenRegistrationFromSpec` is the crash-resume path: heal / `cotal reconcile-gate` inspect the spec **before** abort-reopen.

The durable record is the frozen gate (opId + freeze revision) plus the spec key revision. I did **not** add a third pre-Phase-2 checkpoint write. After Phase 2, old authority is already revoked; the remaining question is whether Phase 3 advanced the spec. The spec row answers that. An extra checkpoint cannot classify a write that happens after it, and it is another lost-ack surface.

`recordsKv` is **required** on `reconcileEndpointGate` / `healFrozenRegistrationGate`. An optional store would silently abort-reopen a committed spec, which is the #1243 loop.

## Current-PUT through promotion (follow-on)

Classification of a current PUT is not enough if a concurrent `deregisterServiceInstance` can delete that spec while Phase 3b/4 still run. Decision: **retain-through-reopen**. The endpoint governance slot is held from decision through Phase-4 reopen, then released. Promote merges binding **without** clearing the slot.

`deregisterServiceInstance` now requires `observeGeneration`: a **read** of this instance's issuance-gate generation (open or frozen), never a freeze.

Observation classification (non-equal is not "behind"; only `slot.generation < liveGeneration` permits delete):

- throw → fail closed (`unavailable`)
- absent / not a non-negative safe integer (`undefined`, NaN, string, …) → fail closed (`unavailable`)
- equal to the slot generation → `registration-in-flight` (nothing removed)
- strictly greater than the slot generation → leftover after reopen (delete proceeds)
- strictly less than the slot generation → ahead of the live gate, fail closed (`unavailable`)

There is no spec re-read on deregister, and no D13 sweeper for leftover slots.

### Residual: reopen → release (accepted availability race)

This is a different residual from the leftover stamp after a lost release.

In the narrow interval **reopen → release**, the token-pinned reopen has already advanced the live gate to G+1 while the slot is still stamped at frozen G. Deregister therefore classifies the slot as behind-generation and may delete the spec. That is correct for the proof-propagation invariant: the consuming transition completed at reopen. It does not create a phantom binding and it does not permit serving on a missing descriptor.

What fails: a just-reopened, not-yet-serving manager can then fail its subsequent `authorizeServeGrant`, because that function **re-reads the spec** (`kv.get` of the instance's `svc….spec`) and refuses when the key is not a live PUT (`failed-precondition`: a serve credential is minted only for a registered instance). That re-read is what keeps the window from becoming a stale-surface failure. Recovery is the same instance registering again over the tombstone.

This is a deliberate availability race, not a safety failure. It is accepted rather than overlooked. Closing it would mean holding the slot *past* reopen until after serve mint, which is a different exclusion and is not this change.

## Coverage (honest)

Two columns. Isolated core = shipped `registerServiceInstance` / `completeFrozenRegistrationFromSpec` / `reconcileEndpointGate` / `deregisterServiceInstance` against an in-memory or ephemeral-broker KV. Boot = `Manager.start()` with no injected KV/write-fault seam. I declined to add a production seam to pretend boot reaches lost-ack.

| crash / race boundary | isolated core | reached from boot |
|---|---|---|
| 1. crash before checkpoint / Phase 2 | YES (existing freeze-then-fail-before-spec abort-reopen cells) | NO without a seam |
| 2. checkpoint write | N/A: no extra checkpoint; the spec row is the commit record | N/A |
| 3. Phase-2 interruption | YES (`smoke:gate-reconcile-resume`; boot-self-heal live-holder / no-oracle refuse **before** mutation) | PARTIAL (refuse-before-mutation only) |
| 4. Phase-3 no-commit | YES (`#1243 no-commit` in `smoke:ep-serve:auth`) | NO without a seam |
| 5. Phase-3 committed-but-unacked, including same-bytes | YES (`#1243` re-registration cell + CONTROL first registration without lost ack; first-registration lost-ack now completes via read-back) | NO without a seam. Production boot that hits #1243 is: heal abort-reopens, then a **new** `registerServiceInstance` freeze. Same-attempt read-back is what stops the loop on the next start. |
| 6. Phase-3b **lost promote ack** | YES for **acknowledgement recovery** (`#1243 Phase-3b` in `smoke:ep-serve:auth`: the second govern write of the re-registration commits then loses the ack; read-back completes). This cell does **not** prove that `promoteHeldGovernance` ran: skipping that function still leaves the cell green. Slot-clear is not this cell; promote retains the slot. Safety of the retained slot is the held-slot cell (row 9). | NO without a seam |
| 6b. Phase-3b DEL between classify and promote | CLOSED by retain-through-reopen: the slot is still held at the live frozen generation, so shipped `deregisterServiceInstance` returns `registration-in-flight` (`smoke:ep-traits` held-slot cell). | NO without a seam |
| 7. recovery-read failure | YES (`#1243 recovery-read failure` stays frozen; next-boot analogue via `completeFrozenRegistrationFromSpec` once the read works) | NO without a seam |
| 8. **lost release ack** (post Phase-4 slot-clear) | YES for **acknowledgement recovery** (`#1243 lost-release-ack` in `smoke:ep-serve:auth`: the third govern write commits then loses the ack; registration still completes because the gate is already open). This cell does **not** prove that `releaseHeldGovernance` ran: skipping that function still leaves the cell green. Residual A (leftover stamp) and residual B (reopen→release window) are named separately; behind-generation delete is row 10. | NO without a seam |
| 9. deregister during a live slot at the **current** generation | YES (`smoke:ep-traits` shipped `deregisterServiceInstance` during the promote write) | NO without a seam |
| 10. leftover slot whose generation is **behind** the live gate | YES (`smoke:ep-traits` behind-generation cell: `slot.generation < liveGeneration` is the only permissive proof) | NO without a seam |
| 11. **reopen → release**: D deletes the just-reopened spec | ACCEPTED availability race (not a safety hole). Window: reopen has advanced the live gate to G+1, release has not yet cleared the slot at G, so D's generation-compare treats the slot as behind and may delete. What fails: the subsequent `authorizeServeGrant` re-reads the spec and refuses a missing PUT (`failed-precondition`). Why safety-preserving: no phantom binding, no serve on a missing descriptor; the spec re-read is the mechanism. Isolated coverage of behind-generation delete is cell 10; this window is named rather than closed. | NO without a seam |
| 12. held slot + `observeGeneration` returns `undefined` | YES (`smoke:ep-traits`: fails closed `unavailable`, spec stays PUT). Previously deleted. | NO without a seam |
| 13. held slot **ahead** of the observed generation | YES (`smoke:ep-traits`: fails closed `unavailable`, spec stays PUT). Previously deleted. | NO without a seam |

Lost promote ack is not DEL-between-classify-and-promote. The Phase-3b cell names lost-promote-ack only. Slot-clear is the **release**, after reopen.

Earliest reddening cell against unmodified main: `#1243: that same attempt completes the freeze…` in `pnpm smoke:ep-serve:auth`. After the fix, mutation-proof kills recovery by reverting PHASE 3 to `throw err`; the **earliest** named red is `an AMBIGUOUS first spec-write (committed then ack lost) read-back completes the same freeze rather than leaving a stuck gate`.

## Proof

- Reproduced red on today's main (`25ab381a7`) with the isolated KV lost-ack cell, injection-disabled CONTROL next to it.
- `pnpm smoke:ep-serve:auth` green (108, includes lost-promote-ack and lost-release-ack as acknowledgement-recovery cells).
- `pnpm smoke:ep-traits` green (119, includes held-slot refuse, behind-generation delete, undefined-observation fail-closed, ahead-generation fail-closed).
- `pnpm smoke:service-dereg` green (21).
- `pnpm smoke:manager-deregister` green (30, `observeGeneration` via `authKv`).
- `pnpm smoke:gate-reconcile-resume` green (22).
- `pnpm smoke:gate-reconcile:auth` green (33).
- `pnpm mutation-proof -- --config packages/core/smoke/mutations/endpoint-serve-lost-ack.json` KILLED (lost-ack recovery; not re-run against the retain/release helpers).
- `pnpm --filter @cotal-ai/core typecheck` and manager `tsc -p tsconfig.json --noEmit` green after rebuilding core.

Clock honesty: the independent design note in the worktree notes is timestamped **after** lens comment `5548718198`. Independence is the route through `repairCursorMatches`, not the timestamp.

## Design disagreement (kept)

A complementary pre-Phase-2 checkpoint is not required to close this loop. The frozen gate already binds the op; the spec revision already classifies Phase 3. Shipping an extra durable write would not recover death between Phase 3 commit and ack any better than reading the spec, and it would not survive a heal abort-reopen unless heal also learned to inspect that checkpoint — which is the same recordsKv change already made.

A D13 leftover-slot sweeper is not required to close the deregister window. Behind-generation slots are orphaned at the live gate generation and do not block delete; the next registration of that endpoint replaces them.


davidfarah2003:
**BLOCK cdd4163ac62fb4bba5306d2dd5652b75ec2c83d1**

### Blocking finding: Phase 3b can promote after the committed spec has been deleted

The required current-`PUT` rule is applied when Phase 3 is classified, but it is not held through governance promotion.

`promoteHeldGovernance()` at `packages/core/src/endpoint-service.ts:692-714` reads only the governance head. If the slot belongs to the same instance, it merges the provisional commands and clears the slot. It does not check:

- that the spec is currently a `PUT`;
- that its revision is still greater than the frozen gate's `registrationRevision`;
- that the provisional slot's generation equals the frozen gate generation.

The normal Phase 3b write at lines 571-589 has the same exposure because it uses the captured governance revision without re-reading the spec. `deregisterServiceInstance()` writes the spec key outside the registration barrier, so this execution is possible:

1. Phase 3 commits the successor spec `PUT`.
2. A clean-stop or operator deregistration revision-pinned deletes that `PUT`.
3. Phase 3b promotes the provisional governance slot.
4. Phase 4 reopens the gate and adopts the now-deleted spec revision.

I reproduced that execution at this exact head using the existing real-broker `smoke:ep-traits` registration path, with shipped `deregisterServiceInstance()` scheduled immediately before the second `govern.wpool5` write, the promotion. The delete returned `{ removed: true, specRevision: 44 }`; the current spec entry was `DEL`; nevertheless the governance head ended with binding `commands.work = [ai.cotal.guarded]` and no provisional slot. The focused assertion, "Phase 3b does not promote a held slot after the current spec was deleted", failed, with the rest of the suite at 112 passed. This is the phantom governance obligation the classification rule is meant to prevent.

The crash-resume path has the same race window: line 668 checks the current spec, then line 675 promotes on a separate key. A delete can land between those operations. Re-reading once immediately before promotion would only narrow, not close, this race. The fix needs serialization that keeps deregistration from invalidating the spec proof through promotion, and promotion must bind the slot to the frozen generation.

### Other review results

- `reconcileEndpointGate()` does not perform an extra promotion after the callee refuses. It uses the unchanged-coordinate abort reopen only when `completed` is false.
- I found no third production gate-reopen path. The production reopen calls are the normal completion, crash-resume completion, reconcile abort, and pre-spec abort helper.
- The `registrationRevision === 0` epoch case is correct for a first registration. Retry cannot double-advance because the token-pinned reopen either leaves the same frozen baseline or has already moved the gate out of that freeze.
- `recordsKv` is required by `reconcileEndpointGate()` and by `healFrozenRegistrationGate()`, and it is supplied by both the boot and CLI production call sites.
- The coverage matrix's isolated-versus-boot split is honest, but the Phase 3b row is incomplete: its test covers a lost promote acknowledgement, not a `DEL` between spec commit/classification and promotion.
- Pre-fix proof reproduced against exact `25ab381a72cd3cbe264406e0c513fd6b5c368d6d`: the adjacent `#1243 CONTROL` stayed green and the same-attempt completion cell went red.
- Mutation proof is sound for its stated target. The `find` anchor is code-only. In a separate raw mutant run, the first `FAIL:` line was the named earliest cell, `an AMBIGUOUS first spec-write ...`, so this is not inferred from the 103-versus-107 marks count.
- Exact-head checks passed: `smoke:ep-serve:auth` 107, `smoke:gate-reconcile-resume` 22, `smoke:gate-reconcile:auth` 33, the targeted mutation KILLED, core and manager typechecks, `check:docsbundle`, and `changeset status`.
- The core and manager minor changeset is correct for 0.x. The human docs and generated bundle are synchronized, but their committed-spec recovery promise is not true under the blocking delete race above.
- I independently compared `890d08a` to this head through the GitHub API: one commit, one file, only `extensions/connector-core/src/docs-bundle.generated.ts` at +3/-3. Source semantics are unchanged.

Distinctive confirmation sentence: the reproduced current entry was `DEL` while the governed command was already binding.

davidfarah2003:
**BLOCK 7ff0c21681bf3329401fdfb1617db03a8f783778**

The retained-slot redesign closes my original Phase 3b reproduction, but the new permissive deregistration branch does not validate or order the observed generation. It therefore fails open for observation results other than a trustworthy integer at or ahead of the slot.

### Blocking finding: non-equal is not proof that the slot is behind

`deregisterServiceInstance()` at `packages/core/src/endpoint-service.ts:967-977` does this:

```ts
liveGeneration = await args.observeGeneration();
if (gov.provisional.generation === liveGeneration)
  return { removed: false, reason: "registration-in-flight" };
// else: slot.generation is behind the live gate
```

The comment does not follow from the comparison. Any non-equal result enters the delete path. There is no runtime validation that `liveGeneration` is a non-negative safe integer and no check that `slot.generation < liveGeneration`.

I reproduced two fail-open forms on the real-broker `smoke:ep-traits` path with a live held slot and current spec:

1. `observeGeneration` returned `undefined`, representing a no-data/all-clear confusion at the exported core seam. Deregistration returned `{ removed: true, specRevision: 52 }` and the current spec became `DEL`.
2. The held slot was generation 2 while observation returned generation 1. Deregistration again returned `{ removed: true, specRevision: 52 }` and the spec became `DEL`.

The second state can be corruption or a stale/inconsistent caller observation. Either way, it is not the documented lost-release orphan. The only permissive proof is `slot.generation < liveGeneration`; equality is in-flight, and `slot.generation > liveGeneration` or a malformed result must fail closed. The production callers usually throw on a missing gate, but this is an exported core function whose own contract currently accepts any runtime value, and the PR specifically promises that unreadable observation fails closed.

### What now works

I reran my original exact construction at this head: shipped deregistration immediately before the Phase 3b promote returned `registration-in-flight`; during the refusal the governance head was still provisional and had no binding commands. The original phantom-promotion defect is closed.

The reopen-to-release interval is also accurately classified as an accepted availability race, not a safety failure. I reproduced a delete there: the spec became `DEL`, then `authorizeServeGrant()` freshly read that spec and refused with `failed-precondition`. The PR body now names this residual separately from the lost-release leftover stamp.

I also confirmed:

- live and crash-resume promotion both call the generation-bound `promoteHeldGovernance`;
- promotion retains the slot through the token-pinned reopen, and release follows reopen;
- the false "sole spec-key writer" invariant was removed;
- all production deregistration callers supply an auth-gate generation read;
- the behind-generation delete and thrown-observation fail-closed paths work as intended;
- no new mutation fixture was added for this redesign. The existing fixture's `find` remains code-only.

### Coverage and gating

The four named behaviors are not equally gated by their named cells:

- Held-slot refusal: deleting the equality guard reddened the held-slot cell first.
- Behind-generation delete: making every held slot refuse reddened the behind-generation cell first.
- Lost-promote-ack: breaking the promote read-back caused an uncaught error at that cell. The suite is counting rather than fail-fast, so it does not print a `FAIL:` line before termination, but the stack points to the exact cell.
- Lost-release-ack: breaking the release read-back likewise caused an uncaught error at that exact cell.

Simply skipping `promoteHeldGovernance` or `releaseHeldGovernance` left the corresponding named acknowledgement cells green. Those cells gate acknowledgement recovery, not that promote/release happened. That is narrower than the PR body's wording but not my blocker because the real held-slot and behind-generation cells cover the safety mechanism itself.

Exact-head runs passed: `smoke:ep-serve:auth` 108, `smoke:ep-traits` 115, `smoke:service-dereg` 21, `smoke:manager-deregister` 30, core and manager typechecks, docs bundle, and changeset status. The new core and manager minor changeset is correct for 0.x.

Distinctive confirmation sentence: an undefined generation observation deleted the live spec instead of failing closed.

davidfarah2003:
Addendum to my BLOCK at `7ff0c21681bf3329401fdfb1617db03a8f783778`: I checked the newly raised legacy-gate hypothesis, and it is **not an additional finding**.

`registrationRevision` is not a checkpoint field introduced by this PR. It was mandatory in `EndpointGateRow` from the first endpoint-gate schema commit (`52fd5b1c3`, 2026-07-21), before the production registration barrier first persisted `op.kind = "registration"` (`d78eacb10`, also 2026-07-21). The provisioner created gates with `registrationRevision: 0`, and the barrier froze by spreading that validated row.

The enforcement point is `parseEndpointGate()`, not the TypeScript type. It uses the closed schema and rejects unless `registrationRevision` is an unsigned integer. I also executed it against a frozen registration row with that field omitted; it threw `internal: the endpoint gate ... does not validate (SPEC 13.1)`. Both boot heal and `reconcileEndpointGate()` parse the row before constructing the recovery arguments, so `completeFrozenRegistrationFromSpec()` cannot receive the hypothesized `undefined` from persisted legacy state through shipped paths.

Exact pre-fix main `25ab381a7` had the same mandatory field and parser. Thus the 27-cycle #1243 field gates have a durable `registrationRevision` baseline. No named legacy recovery limitation is required for a format that was never shipped by the endpoint registration gate.

There is no dedicated missing-`registrationRevision` cell in the changed tests. That is a small coverage gap in the longstanding parser, not a compatibility blocker for this PR. My existing BLOCK remains solely the fail-open handling of malformed or ahead `observeGeneration` results.

Distinctive addendum sentence: a frozen registration gate lacking registrationRevision is rejected before recovery classification.

davidfarah2003:
**APPROVE 75d6de0abe34d038eb2ce0a43b97d9a47b3436e5**

The final correction closes my remaining blocker. `deregisterServiceInstance()` now validates the observation, refuses when the slot is ahead, and permits deletion only when `slot.generation < liveGeneration`.

### Independent verification

- Clean exact-head runs passed: `smoke:ep-traits` 119, `smoke:ep-serve:auth` 108, core typecheck, docs bundle, and diff hygiene.
- My original real-broker Phase 3b construction remains closed: deregistration at the promote write returns `registration-in-flight`, with the slot provisional and no phantom binding.
- A production-style thrown `observeGeneration` directly failed closed with `unavailable` and left the spec `PUT`.
- A genuine different-instance registration holding the endpoint slot did not invoke the named instance's generation observer and did not block deletion of that named instance. Instance B's in-flight spec and slot remained intact. This is correct: the slot protects B's spec, not A's independently revision-pinned spec.
- The reopen-to-release availability residual remains safety-preserving because `authorizeServeGrant()` freshly reads the spec and refuses a missing `PUT`; the PR body names that residual separately.

### Discriminator proof

Using code-only temporary mutations and raw suite output:

- Removing the `wireInt` guard first reddened `an undefined observeGeneration with a held slot FAILS CLOSED`, then showed its spec became `DEL`.
- Removing the ahead-generation refusal first reddened `a slot AHEAD of the observed live generation FAILS CLOSED`, then showed its spec became `DEL`.
- Reversing `>` to `<` stopped earlier on the behind-generation case, as expected: it turns a valid orphan into an ahead refusal before the later ahead cell can run.
- Changing `>` to `>=` survives because equality already returns `registration-in-flight` on the immediately preceding line. It is behaviorally equivalent, so this is not a missing discriminator.

There is no permanent cell for the catch branch where `observeGeneration()` throws. I confirmed this by changing its wrapper to `throw e`; the permanent suite remained green. I classify that as a **non-blocking coverage residual**, not a correctness blocker: the branch is a direct unconditional throw before either record delete, both shipped callers reach it through real `authKv` reads that throw on a missing gate, and my real-broker direct probe confirms the spec remains present. A future cell should pin the `unavailable` translation and message so the production-most failure form cannot drift.

The earlier legacy-row hypothesis remains disproved: `registrationRevision` was mandatory from the first endpoint gate schema and missing values are rejected by `parseEndpointGate()` before recovery classification.

Distinctive confirmation sentence: the genuine foreign slot protected B while revision-pinned deregistration removed only A.

davidfarah2003:
**Second-lens review at exact head `aa4048a8bd32b0ad473b7c11cff10000ca1ec577`.** Independent of rev1266. An existing APPROVE binds `75d6de0abe34d038eb2ce0a43b97d9a47b3436e5`, not this object; both reviews count if rev1266 also lands here.

Resolved at delivery: `git ls-remote origin refs/pull/1266/head`, `gh pr view 1266 --json headRefOid`, and detached worktree `/home/cotal/fm-lanes/rev-1266b` HEAD all name `aa4048a8bd32b0ad473b7c11cff10000ca1ec577`. I did not author this PR. I did not branch or push. Worktree restored to a clean HEAD before this comment.

Delta vs the already-approved sha is two commits, three files, +9/-4. `packages/core/src/endpoint-service.ts` is untouched on that delta.

## 1. Re-anchored `[0]` still KILLS. `[1]` did not produce a named red.

`pnpm mutation-proof --config bin/smoke/mutations/boot-self-heal-gate.json` after `pnpm --filter @cotal-ai/core build` (throwaway `HOME`/`XDG_*`/`TMPDIR`):

- baseline green (16 progress marks)
- `[0]` **KILLED** 1×. Named: `DEAD HOLDER: successor start SUCCEEDS without \`cotal reconcile-gate\` (boot self-heal)`. 8 marks against baseline 16. That is the first DEAD HOLDER assertion in `boot-self-heal-gate.smoke.ts`, not a downstream name.
- `[1]` **INCONCLUSIVE**. Mutated 1×, then `run timed out; a hang is not a red`. 13 marks against baseline 16.

Timeout is graded before named-FAIL inspection in `scripts/mutation-proof.mjs`. I do not treat the hang as a kill.

On a restored tree, `pnpm smoke:mutation-fixtures` reports `boot-self-heal-gate.json — 2 anchor(s) present and unique` (0 dead). Both finds are code-only. That is presence, not discrimination.

The brace wrap is semantically the same `if (auth)` as the one-liner in JavaScript: truthy `auth` still awaits the four-arg heal; falsy still skips. `[0]`'s `if (false && auth)` disables the call in both shapes. Confirmed by reading the current block at `manager.ts:5104` and by `[0]` actually killing.

## 2. Census 117/86 → 118/87 is a correct decision.

`pnpm smoke:required-arg-seam` at this sha: `standaloneConnectOpts: 118 call sites (87 under smoke/, 31 typechecked)`, 253 passed.

`gate-reconcile-auth.smoke.ts` vs `origin/main`: paren calls 2 → 3. The added site is

`connect({ servers: SERVERS, ...standaloneConnectOpts({ creds: await mintCreds(auth, newIdentity(), "provisioner"), tls: false }), maxReconnectAttempts: 0 })`

It is a provisioner connection for `recordsKv` on the now-required heal/reconcile argument. It sits under `smoke/` with this file's other two call sites (the comment's "three other sites" overcounts this file; the decision is still "harness residue, not product connect"). `tls: false` is explicit. That is the class the golden exists to record, not a silent bump.

## 3. Anchor `[1]`: follow-up card, not BLOCK of this head.

`find` is one line:

`probeHolder: makeManagerHolderLivenessProbe({ space: this.space, servers, auth, log }),`

Unique 1× in `manager.ts` at HEAD. `makeManagerHolderLivenessProbe` itself is 1× in that file (commands.ts has a different argument list). Adding a fifth option to the object makes this `find` match 0×, and `mutation-proof` then ERROR rather than silently skip. Same shape that killed `[0]` before the brace wrap.

I am not BLOCKing the PR on that width. `[0]` already showed the cost. Widen `[1]` to a multi-line code-only window around the `reconcileEndpointGate({...})` call (include `kv:` / `evict:` / `recordsKv`) so a fifth option cannot disarm it. That is a follow-up card, not a merge gate on this delta.

I am also not BLOCKing on `[1]`'s INCONCLUSIVE hang. The live-holder cell is still the named expectRed; a timeout is not evidence it survived. It is evidence this kill was not completed in this seat's 900s budget.

## Classification (delta-scoped)

I did not re-derive `endpoint-service.ts`; it is not in `75d6de0..aa4048a8b`. I did run `pnpm smoke:ep-traits` at this sha: 119 passed, including held-slot `registration-in-flight`, behind-generation delete, undefined observation fail-closed, ahead-generation fail-closed.

**Not APPROVE of a completed `[1]` kill.** `[0]` is proven. `[1]` is unique and code-only and did not produce a named red here. No blockers on the census decision or the brace wrap.

davidfarah2003:
**Follow-up on `[1]` at the same head `aa4048a8bd32b0ad473b7c11cff10000ca1ec577`.** Isolated re-run, not a re-grade of `[0]`.

`mutation-proof` timed out on `[1]` and grades timeout before named-FAIL inspection, so that sweep could not credit a kill. I applied only mutation `[1]` (`probeHolder` → `async () => ({ state: "gone", … })`, 1×), ran `pnpm smoke:boot-self-heal-gate` under throwaway `HOME`/`TMPDIR`, 480s KILL cap, then restored `manager.ts` (clean HEAD, find still 1×).

Named cell **did change state**:

- `✗ FAIL: LIVE HOLDER: the refusal is named \`holder-alive\``
- also: successor start REFUSES failed (`started`); gate opened gen 3→5; family mutated
- DEAD HOLDER cells stayed green (expected: assume-dead does not skip heal)
- suite printed `BOOT SELF-HEAL GATE FAILED ❌  (7 passed, 5 failed)` **before** the hang
- then the healed successor kept serving (`manager endpoint: timeout` / `connection refused`) until `timeout --signal=KILL` → 137

So `[1]` **kills the named live-holder cell**. The hang is teardown after that fail (a second start then hits `already serves space … stop it first`), not evidence the mutation survived.

Width ruling unchanged: follow-up card, not BLOCK. This seat now treats `[1]` as a named red, not INCONCLUSIVE.

davidfarah2003:
**Terminal verdict at `aa4048a8bd32b0ad473b7c11cff10000ca1ec577`: APPROVE.**

I am not withholding. The earlier "Not APPROVE of a completed `[1]` kill" was about `mutation-proof`'s INCONCLUSIVE timeout, not about this head's delta.

Independent check of the fixture vs merge-base `25ab381a72cd3cbe264406e0c513fd6b5c368d6d` (`git merge-base HEAD origin/main` in `/home/cotal/fm-lanes/rev-1266b`):

- `[1]` `find`, `replace`, `name`, and `expectRed` are **byte-identical**
- the whole fixture diff is the two `[0]` lines (`find` + `replace`) for the four-arg braced heal call

So `[1]`'s width and its `mutation-proof` timeout are **base conditions**. Merging this PR does not worsen them.

What this head did prove:

- `[0]` KILLED, named the earliest DEAD HOLDER cell
- isolated `[1]` apply+run printed `✗ FAIL: LIVE HOLDER: the refusal is named \`holder-alive\`` (suite FAILED 7 passed / 5 failed) before teardown hung; that is a named red of the cell, even though `mutation-proof` could not credit it (timeout is graded first)
- census 117/86 → 118/87 is a recorded harness-residue decision (`tls: false` on the new provisioner connect); `smoke:required-arg-seam` 118/87, 253 passed
- brace wrap is the same `if (auth)` control flow as the one-liner

**Follow-up cards (named, not merge gates on this head):**

1. Widen `[1]` to a multi-line code-only window around `reconcileEndpointGate({...})` so a fifth option cannot match 0×.
2. `mutation-proof` default 900s on this suite: `[1]` hangs in teardown after the named fail (a second start then `already serves space … stop it first`). That is equally true at `25ab381a7`. A longer `--timeout` is optional and separate; I am not attaching it to this APPROVE.

I do not need a 40-minute retry to approve this sha. Worktree remains isolated and clean at this object.

davidfarah2003:
Merge-gate status at `99134135dd733cf42adf063c1a8ab99a939892f8`, recorded so the reasoning is on the PR rather than only in my head.

The terminal APPROVE on this PR names `aa4048a8bd32b0ad473b7c11cff10000ca1ec577`. The head is now `99134135d`. A verdict binds the sha it names, so that APPROVE does not by itself clear the current head, and I am not merging on it.

What moved, measured rather than assumed:

- `99134135d` is a merge commit whose two parents are exactly `aa4048a8b` (the approved head) and `9f033b128` (main).
- Comparing the committed tree against the machine-computed merge of those two parents:

  ```
  git merge-tree --write-tree aa4048a8b 9f033b128  ->  9bb6cef4fb972d646e2a24a03b60f65537d73368
  git rev-parse 99134135d^{tree}                    ->  769e8552e37c0095cc2dcd08a6da068e9e9817bc
  git diff --stat 9bb6cef4 769e8552
    extensions/connector-core/src/docs-bundle.generated.ts | 6 +-----
  ```

  One file, and it is generated.

- Running `node scripts/generate-docs-bundle.mjs` on a checkout of `99134135d` left `git status --porcelain` empty. That is what separates a true regeneration from a hand-picked conflict resolution, and it is the check worth insisting on, because the two are indistinguishable by inspection.
- `bin/smoke/ci-suites.txt` is untouched by this merge. That matters here specifically: it carries `!merge` in `.gitattributes`, and `git merge-tree` does not honour `.gitattributes`, so a merge touching that path would make the tree comparison above falsely clean. It does not, so the comparison stands.

So the head decomposes into approved branch work, `main` content that landed through its own gates, and one reproducibly generated file. That is a good argument and it is still an argument I made, not a review. An independent seat is re-deriving all of it at this exact sha now, and I will merge on its verdict rather than on mine.

Separately on CI: attempt 1 of the current run failed on shard 2/4 at `pnpm smoke:runtime-mesh-wait`, an unhandled NATS `TimeoutError` from `replayRunJournal` (`packages/core/src/run-journal.ts:413`). This PR touches `run-journal.ts`, `implementations/runtime` and `packages/lang` zero times. I wrote the discriminator before reading results: a failure of that suite on either of two heads without this diff would exonerate it, while a pass on both would be weak evidence only. Both passed — the weak branch — so I am not merging on an explained red. Attempt 2 is running and I require an actual green.

davidfarah2003:
**APPROVE** `99134135dd733cf42adf063c1a8ab99a939892f8`

Exact-head confirmation only. The prior terminal APPROVE at `aa4048a8bd32b0ad473b7c11cff10000ca1ec577` is not re-litigated.

Every blob at this head is either (a) the approved parent, (b) `origin/main` at `9f033b128c183b4a4428c2347a17a661701761bb`, (c) git's auto-merge of (a)+(b), or (d) a generator output that this checkout reproduces exactly. No fourth source.

Checked:

- `git rev-parse HEAD` and `gh pr view 1266 --json headRefOid` are both `99134135dd733cf42adf063c1a8ab99a939892f8`.
- Two-parent merge: `aa4048a8bd32b0ad473b7c11cff10000ca1ec577` and `9f033b128c183b4a4428c2347a17a661701761bb`.
- `origin/main` is `9f033b128c183b4a4428c2347a17a661701761bb`; `git merge-base --is-ancestor 9f033b128 origin/main` exits 0.
- Machine merge vs committed tree: only `extensions/connector-core/src/docs-bundle.generated.ts` differs (`1 insertion, 5 deletions`). `docs/cli.md` and `implementations/manager/src/manager.ts` auto-merged; their HEAD blobs equal the merge-tree blobs (`e45f078d6d1d…` and `109cef23a4d1…`). No conflict markers in the committed tree.
- `node scripts/generate-docs-bundle.mjs` on this checkout: `wrote 39 pages + spec + lang + schema for Cotal v0.42.0`; `git status --porcelain` empty; `git diff --exit-code -- extensions/connector-core/src/docs-bundle.generated.ts` exits 0.
- `bin/smoke/ci-suites.txt` is `ec0069a6987f…` on merge-base, both parents, merge-tree, and HEAD. `.gitattributes` has `bin/smoke/ci-suites.txt !merge`. The merge-tree caveat does not apply.
- `git diff aa4048a8b 99134135d`: 55 files, +1201/−100. Of those, 52 blobs equal main; the generated file is bucket (d); `docs/cli.md` and `manager.ts` are the auto-merges above.
- `git diff 9f033b128 99134135d`: 20 files, +744/−112. 17 blobs equal the approved parent; same generated file; same two auto-merges.

`pnpm typecheck` at this head: passed. pnpm reported `Scope: 27 of 28 workspace projects`. The skipped package is `@cotal-ai/example-04-frontier-faces` (no `typecheck`/`build` scripts). The other 27 typechecked, including `@cotal-ai/smoke-kit`.

Not checked: product behaviour, mutation fixtures, live suites, `pnpm check`, the dev-tree CLI. Those were graded at `aa4048a8b`.

Note on check 2's tree SHA: a conflicted `git merge-tree --write-tree` result is not a stable object, because the conflicted generated file carries markers. Full SHAs and an explicit merge-base produced `9bb6cef4fb972d646e2a24a03b60f65537d73368`; abbreviated SHAs and other invocations produced other trees. Every variant still differed from `769e8552e37c0095cc2dcd08a6da068e9e9817bc` by only that generated file, so the file-set claim holds even though the exact merge-tree SHA does not.

davidfarah2003:
Correction to one number in my merge-gate comment above, raised by the confirmation review and worth having on the record because I published it as something reproducible.

I cited

```
git merge-tree --write-tree aa4048a8b 9f033b128  ->  9bb6cef4fb972d646e2a24a03b60f65537d73368
```

**That tree SHA is not stable.** This merge conflicts on the generated bundle, and a conflicted `merge-tree --write-tree` writes the conflicted file with markers in it, so the resulting tree object depends on how the command is invoked. Full SHAs with an explicit merge-base produce `9bb6cef4…`; abbreviated SHAs and other invocations produce different trees. A reader reproducing my command can legitimately get a different value and conclude my measurement was wrong.

What survives, and is what the check was actually for: **across every variant, the committed tree differs from the machine merge by the same single file**, `extensions/connector-core/src/docs-bundle.generated.ts`. The file-set claim holds; the tree identity does not.

So the reproducible form of the check is *compare the file set*, not the tree SHA:

```
git diff --stat "$(git merge-tree --write-tree <approved> <main>)" <head>^{tree}
```

and require that the output name only artifacts you can regenerate. Quoting a bare tree hash invites a reader to diff two hashes and get a mismatch that means nothing.

The conclusion is unchanged and is now independently established rather than resting on my number: the confirmation review re-derived the decomposition, additionally verified that `bin/smoke/ci-suites.txt` is the same blob at merge-base, both parents, the machine merge and HEAD, that no conflict markers reached the committed tree, and that regenerating the bundle on this checkout leaves the tree clean.

PR #1269, opened 2026-09-05
fix(connector-jcode): bound the pre-join readiness turn

### LINKED ISSUE #1216: connector-jcode: a seat with a heavy persona works invisibly inside an unbounded readiness turn, unreachable and killable at the gate
A managed jcode seat can be alive, healthy, and doing exactly the work it was given while having **no mesh presence at all**. It is absent from `cotal_roster`, and `cotal_dm` to it fails with `no peer "<name>" in space "<space>"`. It is not deaf, not hung, and must not be restarted. It is executing inside its readiness turn, which has no deadline.

## Observed

`rev1211-test`, a review seat spawned 2026-09-03 with a long persona, spent roughly 13 minutes in this state before joining. Throughout: its connector host process was alive, its private connector log held exactly one line (`started a fresh session`) with no route line, its private jcode log showed continuous live activity, and its checkout showed real work appearing and being cleaned up. It joined normally in the end, so nothing was lost that time. Nothing about the failure mode guarantees that.

## Mechanism

Three parts, in `extensions/connector-jcode/src/host.ts` (read at 20b1c7045):

1. **The persona is the only text present at the gate.** Line 792 puts it in the transcript *before* the readiness turn as `sendMessage(..., { noReply: true })`. The kickoff `--prompt` is driven LAST, at line 846, after mesh join. So a seat that starts working at the gate was primed by its persona, not by its kickoff prompt.

2. **The readiness turn is terminal after two attempts.** Lines 801-818: the connector asks for one `cotal_orientation` call, retries once if the tool does not appear in that run's `toolCalls`, and otherwise throws `the cotal MCP bridge did not become callable during its two mandatory readiness turns`. A seat that works for a long turn and never emits the tool call dies at the end of it, and everything it did is discarded.

3. **That turn is unbounded.** Normal startup calls `launchPrivateJcode()` with no deadline (line 729), so no `requestTimeoutMs` reaches the client; only bridge recovery passes one. There is no timeout on the readiness run. The manager's 180s spawn wait only *reports* (`launch status uncertain - no process exit and no mesh presence within 180s`) and explicitly says not to stop the seat for that wait alone. So "invisible" has no upper bound.

Mesh presence is strictly after the gate: `agent.start()` is line 837, after the route diagnostic.

## Why it matters

- An operator cannot reach the seat to redirect or stop it cooperatively for the whole duration.
- The seat is invisible to any supervisor that keys on presence.
- A heavy persona makes this *more* likely, and review and orchestration seats are exactly the ones with heavy personas.
- The work is lost if the gate is never satisfied, after arbitrarily long.

## Suggested

Bound the readiness run, and decide deliberately what a timeout means. Reaching mesh presence before or alongside the readiness proof would also remove the invisible window, though the existing comment at line 836 explains why the proof precedes join, so that is a design call rather than an obvious fix.

Until then the mitigation is brief-side and belongs at the top of every persona, since the persona is the only text the model can see at that point: call `cotal_orientation` once and **yield** before any other tool or file read. "Yield" carries the rule. The gate is satisfied when the tool appears in that run's `toolCalls`, but presence only arrives when the run RETURNS.

## One thing NOT established

"One connector log line and no route line means the seat is behind the gate" looked like a clean diagnostic and is **not reliable**. A second seat on the same host, `rev1211-live`, had exactly one log line and no route line while being present on the roster with a status string only it could have written, and while accepting DMs. Its `COTAL_MODEL` was set, so the route line should have been written before `agent.start()` on a straight reading of the source. That is unexplained and is recorded here as an observation, not a recipe.

The reliable presence test is `cotal_dm`: `no peer "<name>" in space "<space>"` means no presence, and acceptance means presence.

davidfarah2003:
Closes #1216.

A managed Jcode seat with a heavy persona could stay alive and working with no mesh presence, because the pre-join `cotal_orientation` proof had no deadline. The manager's 180s spawn wait only reports `uncertain` and does not stop the seat. Presence still follows the proof: a mute present seat is worse than a bounded invisible window.

This PR ships both halves of that, separately:

1. Bound the gate. The host now races the mandatory orientation proof against the declared three-minute window. That window is one exported constant (`JCODE_READINESS_TIMEOUT_MS`): the connector's `readinessTimeoutMs` and the host's default bound both read it. A turn that overruns that bound exits `readiness_timeout` and never joins, never submits the kickoff `--prompt`, and never appears on the roster. The bound is captured before `COTAL_*` is scrubbed, so a test override actually applies. Teardown on that path does not await in-flight `run()` or `client.close()`, both of which wait for `turn_done`. It kills the private Jcode tree and discards whatever that in-flight turn had generated. Nothing from it is recoverable.

2. Name the outcome. While the proof is running, the connector log writes `pre-join readiness` with the bound and that the seat is not on the roster yet. That in-flight line only means startup reached the gate. After the turn, a second line names what happened: proved (with or without a spawn `--prompt`), provider refusal, or timeout (and that the in-flight turn is discarded). A hang that never returns and never hits the bound has no outcome line. The unreliable "one connector log line / no route line" observation recorded in the issue is still not the signal.

## Proof

`pnpm smoke:jcode-host` 69/69.

The bound cell inverts the clocks: the fake turn is 2500ms and the host bound is 800ms. Mid-turn the host is alive, absent from the roster, `cotal_dm` fails with `no peer`, the persona is already in the transcript, and the kickoff prompt is not. After the bound the host exits non-zero as `readiness_timeout`, and the connector log names `pre-join readiness outcome: timeout`. A successful join with no `--prompt` and a provider refusal each name their own outcome line, not timeout.

`pnpm smoke:jcode-args` 50/50, including that the manager window and the host default are the exported bound rather than a second literal.

Mutation-proof, named cells:

- M6 on `a readiness turn that overruns its bound ends the launch` (one-line timer; pre-reanchor race-drop FAIL line is the same named cell)
- M7 on `the connector log names the pre-join readiness gate while that turn is still running`
- M8 on `a timed-out readiness turn names its outcome as timeout, not a hang or a missing prompt`
- M9 on `provider readiness refusal names that outcome, not a timeout or a missing prompt`
- M10 on `a successful join with no spawn --prompt names that outcome`
- duplicate-literal mutations on the manager window and the host default

`pnpm typecheck` green for `@cotal-ai/connector-jcode`.
`pnpm check:docs-voice` green.
`pnpm smoke:mutation-fixtures` green.


davidfarah2003:
## BLOCK for `419a0700302f5c3173425999cfeef761032db008`

Exact-head audit completed: the shipped M6 mutation window is not code-only. Its `find` includes these two comment lines:

```ts
// Do not await the in-flight run() or client.close(): both wait for turn_done, which is
// the unbounded thing this bound exists to stop. Startup catch stops the private tree.
```

This violates the fixture requirement. The guard becomes dependent on prose that a comment-only edit can silently invalidate. Widen the window with surrounding executable code instead.

The remaining review points also need correction:

1. The bound is duplicated, not read from its declaration: `host.ts` defines `READINESS_TURN_TIMEOUT_MS = 180_000` while `extension.ts` separately declares `readinessTimeoutMs: 180_000`. `COTAL_JCODE_READINESS_TIMEOUT_MS` can make the host bound differ from the declared manager window while the text says they are the same.
2. The new pre-join readiness line is emitted before every readiness attempt. It therefore also appears when there is no `--prompt` and when the readiness turn is later rejected by the provider. The latter has a subsequent refusal, but the claimed line itself does not discriminate these causes as documented. It identifies only that startup reached the readiness gate.
3. On `readiness_timeout`, the host skips awaiting `client.close()` and kills the private Jcode tree. The in-flight turn's generated work is discarded. The docs describe the exit and no join, but do not state the loss or recovery behavior.

Validation: `pnpm smoke:jcode-host` passed 66/66 after its documented connector-core build prerequisite. `pnpm typecheck` passed. `pnpm changeset status` reports the changeset. The complete seven-mutant run exceeded the review command timeout after reaching M7, so I did not treat its partial output as proof. Separate M6 and M7 runs both killed on their named cells. Direct counting-suite runs established the earliest failures: M6 first fails `a readiness turn that overruns its bound ends the launch`; M7 first fails `the connector log names the pre-join readiness gate while that turn is still running`.

davidfarah2003:
## APPROVE for `ac352dc66db8653b897b8ef066ad8e0644f6504c`

Independent exact-head review of this SHA only. The prior BLOCK on `419a0700302f5c3173425999cfeef761032db008` still stands on that SHA.

### Prior BLOCK items

1. **M6 window is code-only.** The shipped find is the single executable line `timeout = setTimeout(() => resolve("timeout"), readinessBudgetMs);` (1 occurrence, no comment). Replacing the delay with `1_000_000_000` killed on `a readiness turn that overruns its bound ends the launch`.

2. **The bound is shared, not duplicated.** `readiness-bound.ts` exports `JCODE_READINESS_TIMEOUT_MS = 180_000`. `extension.ts` uses it as `readinessTimeoutMs`; `host.ts` uses it as the default when `COTAL_JCODE_READINESS_TIMEOUT_MS` is unset. Docs name that override as test-only and say it does not change the manager window. Isolated `pnpm smoke:jcode-args` 50/0. Find-keyed mutation-proof of the three `jcode-args` cells in `jcode-diagnostics.json` killed 3/3 on:
   - `declares a bounded three-minute bootstrap window`
   - `the manager window is the exported bound, not a second literal`
   - `the host default bound is that same export, not a second literal`

3. **In-flight vs outcome.** The in-flight log still fires before every attempt. Docs now say that line only means startup reached the gate. A later `pre-join readiness outcome:` line names proved / provider refusal / timeout. Isolated `pnpm smoke:jcode-host` 69/69 includes those cells. M7/M8/M9/M10 killed on their named outcome/in-flight assertions.

4. **Timeout teardown vs docs.** On `readiness_timeout` the host throws without awaiting the in-flight `run()`, then `stopPrivateJcode({ skipClientClose: true })` which fire-and-forgets `client.close()` and SIGTERM/SIGKILL-escalates the private tree. Docs now say that teardown does not wait, kills the tree, discards the in-flight turn, and that nothing from it is recoverable. The host smoke proves the launch ends non-zero, never joins, never submits the kickoff prompt, and names timeout rather than unknown / refusal / proved.

### Mutation windows

All ten `jcode-readiness.json` finds are unique and comment-free. The four 1-line finds (`launchTui`, `agent.start`, the timeout `setTimeout`, `return JCODE_READINESS_TIMEOUT_MS`) each occur once. Isolated find-keyed mutation-proof of M6–M10 against `pnpm smoke:jcode-host` killed 5/5:
- M6 overrun bound
- M7 in-flight gate log
- M8 timeout outcome
- M9 provider-refusal outcome
- M10 proved no-prompt outcome

### Validation (isolated HOME/XDG/TMPDIR/COTAL_HOME; no `-live`, no `cotal up`)

- `pnpm --filter @cotal-ai/connector-core... build`
- `pnpm smoke:jcode-args` 50/0
- `pnpm smoke:jcode-host` 69/69
- bound-share mutation-proof 3/3
- M6–M10 mutation-proof 5/5
- `pnpm typecheck` green
- `pnpm changeset status` reports the patch changeset
- worktree remained `ac352dc66` and clean after restores

Did not run `smoke:jcode-live` or `smoke:readiness:live`. Did not re-prove M1–M5 (unchanged vs this delta).

### Notes, not blockers

The proved-with-`--prompt` outcome string is written and negatively excluded on timeout/refusal paths; only the no-prompt proved line is positively mutated. `skipClientClose` itself is not a named cell; the suite proves the observable launch end, not the SIGKILL vs `client.close()` distinction. A hang that never returns and never hits the bound still has no outcome line, which matches the docs.

davidfarah2003:
Follow-through on the same SHA `ac352dc66db8653b897b8ef066ad8e0644f6504c` (does not change the APPROVE).

Isolated find-keyed mutation-proof of M1–M5 against `pnpm smoke:jcode-host` killed 5/5 on their named cells:
- M1 first-turn MCP snapshot race recovers on one bounded retry
- M2 a permanently absent cotal tool ends the launch
- M3 foreground TUI starts before its readiness turn finishes
- M4 post-join context supersedes the pre-join orientation card
- M5 post-join notice stays absent while the mesh is unreachable

Tree remained `ac352dc66` and clean after restores. Still no `-live`.

PR #1268, opened 2026-09-05
fix(manager): age leftover turn acceptances off latch-drop, not deadline

davidfarah2003:
Follow-up to #1265.

#1265 stopped `maybeStopTurnSweep` from wiping unsettled acceptances the moment pending is empty. That is the in-flight (or failed) deadline commit. Without a prune, a leftover whose terminal CAS never remembered pins the 5s sweep for the process lifetime.

Aging those leftovers off `deadlineAt` is the wrong clock. A turn that expired after a long outage would be pruned on the first sweep tick and take the GoalRef a late yield still needs. Stamp `leftoverSince` when `commitTurnDeadline` drops the pending latch, and age off that.

## Proof

`manager-reconcile-startup.smoke.ts` 25/25.

The prune cell inverts the two clocks: a stale leftover has a *future* deadline and an old `leftoverSince`; a fresh leftover has an elapsed deadline and a now stamp. Aging off `deadlineAt` cannot pass. `commitTurnDeadline` stamps `leftoverSince` on the shipped deny path.

`pnpm mutation-proof --config implementations/manager/smoke/mutations/unsettled-turn-prune.json` KILLED on:

`the sweep drops a stale unsettled acceptance once its retry window has passed`

The leftoverSince stamp mutant in `reconcile-startup.json` KILLED on:

`commitTurnDeadline stamps leftoverSince when the pending latch drops`

`pnpm --filter @cotal-ai/manager typecheck` green.

Did not run `cotal up`/`cotal down`, `pnpm cotal`, or any `*-live` suite.

PR #1270, opened 2026-09-05
fix(manager): re-anchor late-yield mutation finds on code, not comments

davidfarah2003:
Main is red on `pnpm smoke:mutation-fixtures` from #1265. Dedicated minimal repair; not folded into #1268.

## What failed

Two mutation `find` values spanned the same three comment lines:

- `implementations/manager/smoke/mutations/late-turn-yield.json` [0]
- `implementations/manager/smoke/mutations/reconcile-startup.json` [13]

They are the same mutation (same target, same cell, same 11-line window). The comments were never doing disambiguation: the durable-read block is unique on its own, and so is the `settled.settled` return line.

## What this does

Both `find`s now target the 7-line code-only durable-read block. `replace` is a non-empty compile-clean no-op of that block (`void settled.ref`), not a deletion.

## Proof

`pnpm smoke:mutation-fixtures` green: 0 prose anchors.

`pnpm mutation-proof --config implementations/manager/smoke/mutations/late-turn-yield.json` KILLED. The two fixtures are the same mutation, so this is both cells.

Earliest FAIL from the mutant's raw output (suite is not fail-fast; it logged both and continued):

```
FAIL: a yield after the deny committed, before the index drop, is told the turn ended failed, never `not-found` {"code":"not-found","message":"no pending turn \"ddd…\""}
FAIL: a yield after the deny committed AND the index dropped is still told the turn ended failed, never `not-found` {"code":"not-found","message":"no pending turn \"eee…\""}
```

Same earliest cell as before the re-anchor. Not SURVIVED. Did not redden a different cell.

### DIFF
--- a/.changeset/fix-1262-mutation-anchors.md
+++ b/.changeset/fix-1262-mutation-anchors.md
@@ -1,0 +1,7 @@
+---
+"@cotal-ai/manager": patch
+---
+
+Re-anchor the late-yield mutation `find` on the durable-read code block, not the neighbouring comment.
+
+`smoke:mutation-fixtures` refuses a `find` that spans prose. The #1265 guard used three comment lines to make the window unique; a comment-only tidy would have disarmed it silently. The durable `readGoalResult` block is unique on its own.
--- a/implementations/manager/smoke/mutations/late-turn-yield.json
+++ b/implementations/manager/smoke/mutations/late-turn-yield.json
@@ -7,8 +7,8 @@
     {
       "name": "a late yield after the deadline committed hears not-found instead of the deny",
       "file": "implementations/manager/src/manager.ts",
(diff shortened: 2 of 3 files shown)


davidfarah2003:
APPROVE `42635d226b7314bcc3d44160f14c906359e4ace7`

Both repaired fixtures now have zero comment-containing `find` windows. Each new seven-line code-only window matches exactly once in `origin/main:implementations/manager/src/manager.ts`, and each replacement is non-empty.

The mutation remains equivalent: it removes only the `settled.ref` durable `readGoalResult` fallback while preserving the surrounding settled-answer path. The replacement is compile-clean. I ran both repaired fixture instances through `mutation-proof`; each was KILLED. Raw suite output showed the earliest failure remained `a yield after the deny committed, before the index drop...`, followed by the index-dropped variant. I also ran the pre-reanchor mutation against the same head, and its two ordered FAIL lines were byte-for-byte identical.

Checks passed:
- `pnpm typecheck`
- repaired `late-turn-yield.json` mutation cell
- repaired `reconcile-startup.json` mutation [13]

Observation, not introduced here: `reconcile-startup.json` mutation [11] still has its pre-existing empty replacement and should be tracked separately.

A changeset is not required for test-fixture-only changes because no shipped behavior changes. The included manager patch changeset is conservative and valid, and `pnpm changeset status` succeeds under the repository's fixed version group.

This is ready to unblock main.

PR #1087, opened 2026-08-30
fix(cli): point status skills rows at setup --skills

davidfarah2003:
Closes #421.

## Scope

`cotal status` recommended unscoped `cotal setup` for stale or missing skills. This PR adds `cotal setup --skills`, keeps it fenced from `--full` and `--demo`, and points the Claude and `.agents` status rows at it.

Harness-specific setup is connector-owned. The base CLI resolves a connector-declared `connector-setup` provider and passes only the vendor-neutral Agent Skills directory, version, and state directory. The Claude connector owns its marketplace manifest, plugin payload, native commands, scopes, and verification. A missing declared provider fails loud; a provider whose required harness executable is absent is skipped while `~/.agents/skills` still reconciles.

The current-main fold preserves the stale skills version pair and scoped remedy together: `v<installed> ≠ v<CLI> · stale · cotal setup --skills`. New CI suites use hashed fragments; the frozen legacy registry is unchanged.

## Validation

Final head `3f9b570d6e79c583efaf5fc205fa783fb3200466`, merged with main `b191d114e988c73a77a53a5f956f286575395a04`:

- Packed pre-fix reproduction found the generic `@cotal-ai/cli` tarball shipping `.claude-plugin/plugin.json` and compiled Claude plugin install/update invocations.
- Final packed proof: the generic CLI tarball has no harness plugin manifest, `--scope`, `.claude-plugin` path, or native plugin install/update invocation. The Claude connector tarball owns `skills-plugin/.claude-plugin/plugin.json` and its provider implementation.
- Packed installed-extension fixture lazily materializes both `connector:claude` and `connector-setup:claude` through the recorded manifest.
- Absent-harness regression proves the provider is skipped and the cross-vendor `.agents` skill still installs.
- Mutation coverage accepted all new configs as source-gradable.
- Setup provider boundary mutations: 1/1 generic CLI and 1/1 Claude declaration KILLED red-and-named.
- Status skills mutations: 4/4 KILLED red-and-named.
- Status provenance mutations: 3/3 KILLED red-and-named.
- `smoke:setup-provider-pack`, connector setup provider smoke, `smoke:status-provenance`, `smoke:status-skills-remedy`, and `smoke:setup-failloud` passed.
- Docs bundle regenerated and matched the sources.
- Flag inventory, gate inventory, CI fragments, shard-stability smoke, and `check:shard-stability b191d114... HEAD` passed with `STABLE=0` and no pre-existing assignment moved.
- Workspace build passed.
[...]

Issue #1197, opened 2026-09-01
test: mutation configs are proven once at authoring and never re-proven, so a widened rule leaves a WRONG-RED guard reporting green


davidfarah2003:
## Problem

A mutation config is written once, proven once, and then never re-proven — but the rule it guards
keeps moving. When a later change widens that rule, the config is edited to follow, and the edit is
graded by reading the diff rather than by running the proof. A config in that state reports the same
green as one that still discriminates.

This is a different class from #1102. That issue is about configs the coverage auditor cannot
CLASSIFY: missing `cell`, missing `"grades": "tool"`, an ungradable source target. The gate it asks
for is a static metadata check, and it is worth having. This issue is about a config whose metadata
is completely valid and which the auditor accepts without complaint, but whose mutant no longer
reaches the cell it names. Both gates are needed and neither substitutes for the other.

## The evidence, measured on #1190

`packages/lang/smoke/mutations/refused-retry.json` was authored in that PR. At its first head
`409f900c7e0bbb6a507e52fcb0871256e367ad01`, mutant 1 ("a refusal rides the ordinary failure path and
settles the entry failed") named:

- `expectRed` / `cell`: `the refusal halts the run uncatchably, past try and finally alike`
- `completionMarker`: `ok a workflow's catch does not swallow its host's stop`

Both fields present, both naming real printed cells, so the config is metadata-valid. Under the
mutant, `pnpm smoke:lang-interpret` dies at `packages/lang/smoke/interpret.smoke.ts:1279` — a
completely different cell, `a held arm inside a race unwinds the run even though a sibling won` —
because the suite is fail-fast and that cell is upstream of both the marker and the named cell. The
named cell is never reached. `mutation-proof` reports WRONG-RED: the suite went red, but not where
the config claimed.

The repair, at `e1c7478f44e5bb9041e4c6fda29a9471e12583be`, repointed the mutant at the race-held
cell with the marker immediately upstream of it, and the config now returns 3/3 KILLED. The
engineering reviewer re-derived that by hand-applying the mutant rather than trusting the runner:
https://github.com/Cotal-AI/Cotal/pull/1190#issuecomment-5494540876

**Why the config drifted is the general shape.** `packages/lang/smoke/mutations/race-append-unwind.json`
was authored during #1188 to guard the winner scan unwinding on `JournalAppendRejected`. #1190
widened that rule to unwind on `RunHeld` as well, and extended the config's `find` to match. The
[...]

Issue #631, opened 2026-08-18
test: a suite that spawns a seat can finish green and never exit


davidfarah2003:
## What was found

A smoke suite that spawns a seat process can finish every cell, print a green summary, and then never
exit. The cause is not in the suite's assertions: a seat spawns its own agent process, that grandchild
inherits the stderr pipe the suite reads, and killing the seat alone leaves the grandchild holding the
write end open. The read end never sees EOF, so the last handle never closes and the process stays
alive after its work is done.

This fails in the worst possible direction. From inside the run it looks like a pass. From outside it
looks like a hang. On CI it is a shard that dies at its own timeout with a green summary sitting in
the log, and under `pnpm mutation-proof` it is a baseline that never returns, so no mutation in that
config can be graded at all.

It was observed and fixed in `extensions/connector-codex/smoke/events-lifecycle.smoke.ts`: each seat
is spawned into its own process group, the group is what gets signalled, the pipe ends are dropped
after, and the fact whose absence causes the hang is asserted by name (after teardown, neither seat's
process group still has a member). Measured on that suite: exit 0 in about 15 seconds and no orphaned
processes, against runs that previously had to be killed at 100 and 600 seconds, and ten orphaned
agent processes counted from earlier runs.

## The observation this issue is for

Other suites spawn children the same way, without a process group:

- `extensions/connector-opencode/smoke/agui-session-reset.smoke.ts`
- `extensions/connector-opencode/smoke/boot-prompt.smoke.ts`
- `extensions/connector-opencode/smoke/turn-wedge.smoke.ts`
- `extensions/connector-opencode/smoke/cooperative-stop.smoke.ts`
- `extensions/connector-claude-code/smoke/wake-path.smoke.ts`

**No defect is claimed here.** These are green on `main` today, and whether any of them leaks depends
on whether the child it spawns has a child of its own that inherits a pipe. The shape exists, that is
all, and the shape is the one that hides.

## What would close this

For each suite above, either a run that exits on its own with no processes left behind, or the same
teardown the codex lifecycle suite now uses. The check is cheap: run the suite without a timeout, and
after the summary line prints, confirm the process is gone and that no spawned agent process survived
it. Counting the survivors matters more than reading the summary, because the summary is green in
both worlds.

davidfarah2003:
Measured the shape across the named suites at `origin/main` `8da949e8a`. The list is mostly clear, and there is one live instance it does not name that already has its own open issue.

[...]

PR #1272, opened 2026-09-05
test(ci): re-prove mutation fixtures when guarded sources change

### LINKED ISSUE #1217: ci: the mutation-fixture gate audits anchors but never runs the proofs, so a PR can retire a mutation while green
## What happened

On PR #1212 (perf(web): serve /api/activity from one stream read), `pnpm smoke:mutation-fixtures` was green at head f8cca938e once three moved anchors were re-anchored, and a mutation in the same fixture set had silently stopped being caught.

Measured by the lane, single-mutation config, same host and TMPDIR, isolated to one mutation on both trees:

- `history-cancellation` mutation M2 ("an aborted history window asks the broker for its entire pending backlog in one pull"): KILLED at 544a974b7 (base) on cell 6.9, SURVIVED at f8cca938e.
- The mutated line is untouched by the PR (`git diff 544a974b7..f8cca938e` on it is empty), and the anchor auditor reported it present and unique the whole time.
- Positive control in the same run: another mutation in the same file was killed, so the suite reaches `endpoint.ts` at runtime.

Cause: cell 6.9 asserts that the request after a cancelled `/api/dms` answers instead of being starved by abandoned work. That follow-up request is `/api/activity`, which the PR made roughly nine times cheaper. The abandoned whole-backlog pull that used to starve a 70-way fan-out no longer starves one filtered read, so the cell passes whether or not the 32-message bound holds. The bound is still in the code and still load-bearing; only the detector lost discrimination.

## Why this is a gate defect and not only a lane defect

CI runs the anchor auditor (`smoke:mutation-fixtures`) and not the proofs (`mutation-proof`). The auditor answers "does the anchor still match exactly once", which is mechanical. It cannot answer "does the mutation still get killed", which is the property the fixture exists for. A PR that makes a probe cheaper, moves a deadline, or changes what a cell's follow-up request costs can retire a mutation while the auditor stays green, and that is what #1212 did.

The lane found it only because it ran the proofs by hand after re-anchoring.

## Suggested shape

1. In CI, for every fixture whose anchored files intersect the PR's changed files, run the proof for that fixture (kill or survive), and fail the job on a survivor. Scoping by changed files keeps the cost bounded.
2. A survivor report must distinguish SURVIVED from UNAPPLIED and UNCOMPILABLE (the three read identically from outside; a control dying with a ReferenceError instead of an AssertionError is the tell).
3. Optionally a scheduled full proof run over all fixtures, so drift that arrives through unrelated PRs is caught within a day rather than never.

## Evidence

- PR #1212, base 544a974b7, head f8cca938e (the lane's repair commit that restores discrimination in 6.9 follows on that PR; the fix asks 6.9 for the same demanding page 6.2 asks for, which is the adjustment #1210 already forced on 6.2 through 6.5).
- Fixture: `bin/smoke/mutations/history-cancellation.json`, mutation M2.
- Filed from the fleet operator seat on behalf of the #1210 lane, which measured it.

davidfarah2003:
## Scope

Closes #1217.

This adds an executable mutation-reproof gate. On pull requests and pushes to `main`, it selects every committed mutation fixture whose JSON file or guarded source file changed, then runs `mutation-proof` at the reviewed head. It prints the selected path set, classifies each fixture's proof by verdict, and fails on any real finding. A daily full sweep and manual dispatch bound drift from cross-cutting changes a fixture does not directly name.

The existing `smoke:mutation-fixtures` audit is unchanged in purpose: it detects anchor rot, ambiguity, missing targets, and prose anchors. This gate detects a matching anchor whose mutation no longer discriminates.

## What the gate now refuses that it previously passed

The blocked head at `0413f0e38` returned the same exit 0 for a real all-clear and for a run that measured nothing. This revision closes every such path:

- **Deletions are seen.** The changed set is now derived with `git diff --name-status -M`, not `--name-only --diff-filter=ACMR`. `ACMR` excluded `D`, so a deleted guarded source never entered the set. A deleted guarded source now selects its fixture.
- **Both sides of a rename are seen.** `--name-only` reported only a rename's new path, so a fixture still anchored to the old path was never selected. Both the old and the new path now enter the changed set.
- **A deleted or renamed-away guarded source is a loud failure, not a skip.** Such a fixture is *dangling*: its anchor cannot resolve and its proof is unrunnable. The gate reports it and exits non-zero.
- **A malformed fixture is refused, not dropped.** Corpus membership is decided by path convention (a `mutations/` JSON or a `*.mutations.json`) before parsing, so a fixture that no longer parses, or that carries no `mutations` array, fails the run rather than vanishing from the set and silently shrinking it.
- **Zero-selected is an explicit, evidenced state.** A legitimate all-clear (exit 0) now prints the diff size, the changed-set size, the corpus size, and the selected count, so a reader can tell "nothing guarded changed" from "the selector returned no data". An empty corpus is not an all-clear: it exits non-zero.

## The outcome set is classified, not collapsed

`mutation-proof` grades more than pass/fail, and collapsing any of it is itself a silent failure. The gate reads its verdict lines and exit code instead of treating every non-zero as one thing:

| outcome | meaning | gate response |
|---|---|---|
| all `KILLED` (exit 0) | the guard discriminates | pass |
| `SURVIVED` / `UNGRADABLE` | the guard does not discriminate | fail — a real finding |
| dead / ambiguous anchor (`ERROR`) | the fixture is broken | fail |
| pre-red (exit 4) | the suite was red before any mutation | reported as its own state; does **not** fail the gate |
| `INCONCLUSIVE` only | timed out or hung; evidence in neither direction | reported as its own state; does **not** fail the gate |

This resolves the review's finding 3. The `changed` job selected `sandbox-guard.json` and refused it because `pnpm smoke:sandbox-guard` was red before mutation — a state owned by #1279 (three real offenders to adopt the guard, three scan false positives to tighten), unenforced until now because that suite was reachable only from `check` (#1278). Blaming this diff for it was a false blocker. The gate now reports a pre-red selection as `PRE-RED`, names the fixture, and stays green, so this PR no longer has to block on #1279 landing first.

## The positive control

A gate that has only ever agreed with a clean tree has never shown it can disagree. `bin/smoke/mutation-reproof.smoke.ts` runs the shipped scan against synthetic repositories and **asserts the selected set**, not a boolean, so a gate that selects the wrong thing is visibly wrong rather than accidentally non-empty. It proves the gate:

- selects exactly the fixture whose guarded source changed (a two-fixture corpus; the untouched one is never picked);
- **fails** on a known survivor and names exactly that fixture;
- **fails** on a **deleted** guarded source, naming exactly the dangling fixture;
- **fails** on a **renamed-away** guarded source, naming exactly the dangling fixture;
- **passes** an unrelated change and prints the all-clear evidence (`diff 1 record(s), 1 changed path(s), corpus 2`, selected 0);
- reports a **pre-red** selected fixture as `PRE-RED`, names it, and does not fail.

Each of those behaviours is mutation-proofed: `pnpm mutation-proof --config bin/smoke/mutations/mutation-reproof.json` breaks the fix on purpose (re-exclude `D`; contribute only a rename's new path; disable the dangling refusal; reclassify pre-red as fatal) and all four mutations were `KILLED` on their named control assertions.

## Scheduling, cost, and triage ownership

The tree measures 329 fixture configs and 1,714 mutations; re-proving the corpus on every revision would multiply build and smoke cost by hundreds. The normal `changed` job re-proves only fixtures whose config or guarded source intersects the reviewed diff. **The cost profile of the normal path is unchanged by this revision** — the selection predicate is the same, the corpus is parsed once (now over fewer files, since membership is scoped to fixture paths), and the added existence check runs only over the already-selected set. No path was turned into a full sweep.

The daily sweep and `workflow_dispatch` bound cross-cutting drift. A scheduled failure fails only that scheduled workflow; it does not block an unrelated pull request or retroactively change a merged commit's required-check result.

On the review's finding 2, and stated plainly rather than implied: **the scheduled sweep has no triage owner in this PR.** It runs; a failure is visible only to someone who opens the Actions tab. There is no notification, no assignment, and no issue created. Nothing here should be read as implying that someone is watching. A notification step that files a deduplicated, assigned tracking issue on failure is written and ready (it needs a push from a credential with the `workflow` OAuth scope, which this lane does not have), and is deferred to #1292. Until that lands, treat the scheduled sweep as an unwatched gate.

**Concurrency.** `mutation-reproof.yml` declares no concurrency group, on purpose. The scheduled full sweep is a daily cron under a 360-minute cap, so a scheduled run cannot overlap the next; the push/PR `changed` job is lightweight and wants ordinary per-ref eviction, not queueing. The one residual overlap is a manual `workflow_dispatch` sweep running alongside a scheduled one, which is rare and intentional. The tidy version — a job-scoped group on the `full` job with `cancel-in-progress: false` so overlapping sweeps queue rather than pile up — is folded into #1292, because it edits a workflow file and needs the same `workflow` scope. A workflow-level group is deliberately not used: it would couple the per-diff `changed` job to the sweep's queueing semantics and force it to queue-not-evict on push. The `smoke:workflow-concurrency` population sentinel is updated (`EXPECTED_WORKFLOWS` 6→7, `EXPECTED_PUSH_TO_MAIN` 5→6) with an attributing comment, so the new workflow is a deliberate, counted addition rather than a silent one.

**Operator pull-request gate.** `mutation-reproof.yml`'s `pull_request:` trigger carries no `paths:` filter, so `scripts/pr-head-gate.mjs` (the operator-run `pnpm pr-head-gate <N>` command, not GitHub branch protection or a required status check) now expects `Mutation reproof` among the workflows for every PR. The consequence: a PR whose head predates this workflow will read as not-green from that gate until it is repushed and the run mints. That falls on whoever runs the gate before a merge. It is cheap — the `changed` job exits 0 with "No mutation fixtures to re-prove" when nothing relevant changed — but it is a policy change and is named here rather than arriving as a surprise. `bin/smoke/pr-head-gate.smoke.ts`'s pinned expectation and the #1087 positive-mint fixture are updated to match, and the suite's own mutation set still kills 12/12.

## Deliberate boundary

Changed-source reproof does not establish that a still-matching `find` anchors the same statement it anchored before a re-anchor. It proves only that the current matching mutation still produces its declared named red. Re-anchor continuity is a separate comparison. Reports use fixture paths and mutation names, never mutable array indices.

## Validation

- `pnpm smoke:mutation-reproof` (6 control assertions, all passing)
- `pnpm mutation-proof --config bin/smoke/mutations/mutation-reproof.json` (4/4 `KILLED`, tree restored)
- `pnpm smoke:mutation-fixtures` (anchors present, unique, code-only)
- `pnpm smoke:ci-suites`, `pnpm smoke:ci-fragments`, `pnpm smoke:ci-suites-freeze`, `pnpm smoke:ci-declarations`
- `pnpm smoke:gate-inventory` (the suite is reached from a real root, not only from `check`, after #1283)
- `pnpm typecheck` (`Scope: 27 of 28 workspace projects`; the 28th, `@cotal-ai/example-04-frontier-faces`, is JS-only and skipped)
- `pnpm changeset status` (CI-only; the empty-frontmatter changeset produces no bump)

No `*-live` or stack-lifecycle suite was run.


davidfarah2003:
**BLOCK** at `0413f0e384b455f7e9fae0f4d3a9ff37040081b2`.

The central guard still has silent-success paths:

1. `scripts/mutation-reproof.mjs` derives changed paths with `--diff-filter=ACMR`, then treats an empty set as “No mutation fixtures to re-prove” and exits 0. I reproduced three distinct commits against a synthetic guarded fixture:
   - an empty commit: `0 changed path(s)`, exit 0;
   - deleting the guarded source: `0 changed path(s)`, exit 0;
   - renaming the guarded source: only the new path was returned, `0 fixture(s) selected`, exit 0.

   The set is also derived only from fixtures parseable at the current head, so a deleted or malformed fixture can disappear before selection. Existing structural checks may catch some individual deletion/rename shapes, but this instrument cannot distinguish “the complete guarded set has no intersection” from “the derivation returned no data.” The smoke asserts one known offender, not the complete guarded-source/fixture set. Include deletion and both rename paths, account for fixtures that existed at the base, and make a zero-path result an explicit validated state rather than an all-clear.

2. The scheduled workflow has the claimed isolated blast radius, but no named triage owner exists in the workflow or repository ownership configuration. There is no notification, assignment, or issue-creation step. The PR body’s “maintainer who receives the scheduled failure” is therefore not enforced by the workflow being reviewed.

3. The new `changed` job is currently red after executing its steps, so this is not the zero-step billing-outage shape. It selected four fixtures, then refused `bin/smoke/mutations/sandbox-guard.json` because `pnpm smoke:sandbox-guard` was red before mutation. A gate that cannot pass its own reviewed change is not ready to gate merges.

Controls that did hold: I directly blinded the source selector and the earliest ordered failure line was `the shipped scan reports the known survivor set after its guarded source changes`; the self-mutation anchor is code-only; reports use fixture paths and mutation names, not indices; the CI suite uses the correctly hashed fragment; the deliberate re-anchor boundary is stated accurately; `pnpm smoke:mutation-fixtures`, the fragment/inventory checks, and `pnpm typecheck` passed. The CI-only change does not require a package bump; the included empty-frontmatter changeset produces no bump.

davidfarah2003:
Two things on this verdict, one that resolves finding 3's cause and one that widens the outcome set the gate has to read.

## Finding 3's pre-red has a named, tracked cause, and it is not this PR's

I ran the shipped suite at this PR's exact head (`0413f0e38`), from a clean worktree, through `packages/smoke-kit/smoke/sandbox-guard.smoke.ts` itself rather than by re-applying its rule:

```
actual: [
  'implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts: no shared guard call',
  'implementations/cli/smoke/up-restore-reentry-live.smoke.ts: no shared guard call',
  'implementations/cli/smoke/up-resume-render-lock-live.smoke.ts: no shared guard call',
  'implementations/runtime/smoke/mesh-monitor.smoke.ts: no shared guard call',
  'implementations/runtime/smoke/migrate-run.smoke.ts: no shared guard call',
  'implementations/runtime/smoke/migrate-run.smoke.ts:225: raw down spawn is not immediately guarded',
]
expected: []
```

That is the offender set in #1279, line for line. Three of the six are real (the `up-*-live` suites call the real CLI `down` and never adopted `assertSmokeSandboxDown`) and three are scan false positives (`mesh-monitor` uses cotal-lang `wait(down)`; `migrate-run:225` matches `"down"` inside a template string passed to `record()`).

Why the suite is red and stayed red: it is reachable only from `check`, and no CI job invokes `check` — see #1278. So #884's guard has been correct and unenforced since it shipped, and this PR is the first thing to try to stand on it.

**Landing order this implies.** The `changed` job here cannot go green until the offender set is empty. That is #1279 item 2 (adopt the guard in the three real offenders) plus item 3 (tighten the scan so the three false positives stop counting). Neither is this PR's work, and neither should be pulled into it. This PR selecting `sandbox-guard.json` and then refusing it is the gate behaving correctly on a suite that was already broken — which is a reason to sequence, not a defect in the gate.

Suggested sequencing: land #1279 items 2 and 3 first, then re-run this PR's `changed` job. If you would rather not block on that, the alternative is for the gate to report a pre-red selection as a named, non-fatal state on this one fixture with a reference to #1279 — but I would take the sequencing over carrying an exception.

## The outcome set is at least five wide, and collapsing any of it is a silent failure

The gate here already distinguishes pre-red, correctly, which is the control that produced finding 3's evidence at all. Recording the full set so nothing downstream re-collapses it:

| outcome | meaning | correct gate response |
|---|---|---|
| `KILLED` | the guard discriminates | pass |
| `SURVIVED` | the guard does not | fail — a real finding |
| pre-red | the suite was red before mutation | not this diff's defect; report, do not blame |
| `INCONCLUSIVE` | the run timed out or the budget was exhausted | evidence in neither direction |
| dead anchor | the `find` matched 0× | the fixture is broken, not the guard |

`INCONCLUSIVE` is the one most likely to be lost, because it looks like a failure. `scripts/mutation-proof.mjs` already grades a timeout ahead of inspecting named FAILs, and says why in the file: *"A timeout kills the child and leaves status null; that is not a red, it is an unknown."* That is correct, and it means a mutation that genuinely reddened its named cell can still come back `INCONCLUSIVE` when the suite hangs in teardown after printing its verdict. That has been observed: a fixture reported `INCONCLUSIVE` on a sweep and was later shown to kill when the same single mutation was applied by hand.

The consequence for a gate built on this: collapsing pre-red or `INCONCLUSIVE` into `SURVIVED` produces false blockers, and collapsing either into `KILLED` produces false clearance. A dead anchor is separate again — `mutation-proof` errors on it rather than skipping, which is the right behaviour and worth not regressing.

## On finding 1

Agreed, and the shape has a name worth writing into the fix: an empty result set is only an all-clear if the derivation is known to have run and to have had something to look at. `0 changed path(s) → exit 0` cannot distinguish "nothing guarded changed" from "the selector returned no data", and a rename returning only the new path is the case that makes that concrete. Make zero an explicitly validated state with its own assertion, rather than the absence of work.

Related: #1197 (fixtures proven once at authoring and never re-proven), #1217 (the anchor audit checks presence and uniqueness, which is not discrimination), #631 (a suite that hangs in teardown, which is how `INCONCLUSIVE` gets manufactured).

davidfarah2003:
Addressed at `6a94e1884` on `fix/1217-mutation-reproof` (findings 1 and 3), with finding 2 in a local workflow commit that could not be pushed from here — see below. Point by point.

**Finding 1 — silent-success paths.** Fixed, and each shape is now a control assertion rather than a claim.

- Deletions: the changed set is derived with `git diff --name-status -M`, not `--name-only --diff-filter=ACMR`. `ACMR` excluded `D`; deletions now enter the set.
- Renames: both the old and the new path of a rename now enter the set, so a fixture still anchored to the old path is selected.
- Dangling: a selected fixture whose guarded source does not exist at head is a dangling fixture — its proof is unrunnable — and is a loud non-zero failure naming the exact fixture.
- Malformed / disappearing fixtures: corpus membership is decided by path convention (a `mutations/` JSON or a `*.mutations.json`) **before** parsing, so a fixture that no longer parses, or lacks a `mutations` array, is refused rather than dropped. An empty corpus exits non-zero.
- Zero-path is now an explicit, evidenced state: a legitimate all-clear prints diff size, changed-set size, corpus size, and selected count, so "nothing guarded changed" is distinguishable from "the selector returned no data". "The complete guarded set has no intersection" and "the derivation returned no data" are now different exit codes.

The smoke no longer asserts a single offender. It asserts the **selected set** against an expected set across five cases (survivor, delete, rename, unrelated, pre-red), so a gate that selects the wrong thing is visibly wrong.

**Finding 2 — no enforced triage owner on the scheduled sweep.** Fixed in a commit that updates `.github/workflows/mutation-reproof.yml`: the `full` job files a deduplicated tracking issue on failure (or comments on the open one), with `issues: write` scoped to that job, so a red sweep is an owned, tracked item rather than a run visible only in the Actions tab. That single commit is not yet on this branch — this session's push credential lacks the `workflow` OAuth scope, so GitHub rejects a push that touches a workflow file. The code fixes for findings 1 and 3 are pushed; the workflow commit is ready and needs a push from a credential with `workflow` scope. Flagging rather than working around it.

**Finding 3 — the `changed` job was red on `sandbox-guard.json`.** Agreed on cause and sequencing, and handled without pulling #1279 in. Your follow-up is right that the outcome set is wider than pass/fail, so I made the gate read `mutation-proof`'s verdicts rather than collapse every non-zero:

| outcome | gate response |
|---|---|
| all `KILLED` | pass |
| `SURVIVED` / `UNGRADABLE` | fail (real finding) |
| dead / ambiguous anchor (`ERROR`) | fail (broken fixture) |
| pre-red (`mutation-proof` exit 4) | reported as `PRE-RED`, named, does not fail |
| `INCONCLUSIVE` only | reported, does not fail (neither a blocker nor a clearance) |

A pre-red selection — `sandbox-guard.json` under #1279/#1278 — is now reported as `PRE-RED` naming the fixture, and the gate stays green. Collapsing pre-red or `INCONCLUSIVE` into `SURVIVED` (false blocker) or `KILLED` (false clearance) is exactly what this avoids. This PR no longer needs #1279 to land first; landing #1279 items 2 and 3 later simply moves `sandbox-guard.json` from `PRE-RED` to a re-proven `KILLED`.

**Controls that held** (your list) are unchanged: code-only self-mutation anchors, path/name reporting rather than indices, correctly hashed fragment, accurate re-anchor boundary. The self-proof is now four mutations, each `KILLED` on its named control assertion: re-exclude `D`; contribute only a rename's new path; disable the dangling refusal; reclassify pre-red as fatal.

Ready for a fresh review of the new head.

davidfarah2003:
**APPROVE** at `f807fdbbde1d4642712382d6fd0de193b582bc45`.

(`gh pr review --approve` is refused on this token as self-review. Verdict is still APPROVE.)

Re-ran the three silent-success cases from an earlier review against the shipped `scripts/mutation-reproof.mjs`, in scratch git repos under this worktree (not `/tmp`).

| case | exit | printed selection |
|---|---|---|
| empty commit | **0** | `0 fixture(s) selected from 1 (diff 0 record(s), 0 changed path(s), corpus 1)` then `No mutation fixtures to re-prove…` |
| delete guarded source | **1** | `UNMEASURED — 1 dangling fixture(s)` / `smoke/mutations/guard.mutations.json -> missing: guard.mjs` |
| rename guarded source | **1** | same dangling failure, same fixture path |

Empty commit remaining exit 0 is the legitimate all-clear for a zero-record diff. Delete and rename are loud now.

**Positive control can fail.** Restored `--diff-filter=ACMR` on a copy of the scan (worktree left unmodified) and ran the shipped control. It went red on deletion:

```
✗ FAIL: the gate FAILS on a deleted guarded source and names exactly the dangling fixture
    status=0 dangling=[]
mutation reproof: 0 fixture(s) selected from 2 (diff 0 record(s), 0 changed path(s), corpus 2)
```

That is the old silent skip. The control is not decorative. The survivor case asserts the exact selected path (`smoke/mutations/a.mutations.json`), not merely non-empty. Rename still passed under ACMR, which is expected: `R` is not `D`. The second mutation in `bin/smoke/mutations/mutation-reproof.json` covers the rename-old-path drop.

**All-clear vs looked-at-nothing.** An unrelated `README` change: exit 0, `diff 1 record(s), 1 changed path(s), corpus 1`, zero selected. Empty corpus, a malformed fixture, and a selected fixture whose guarded file is missing at head all exit 1 (`UNMEASURED`).

**Corpus by path convention.** Malformed configs are refused instead of dropped. Remaining blind spot: a committed fixture that is neither `*/mutations/*.json` nor `*.mutations.json` never enters the corpus (if it is the only fixture, that is empty-corpus exit 1; if mixed with conventional configs, it is omitted). At this head the convention matches all 337 `*mutation*.json` configs. No such stray file today.

**Cost.** PR/push job is still selective (`--base` / `--head`, no `--all`). `--all` is schedule / `workflow_dispatch` only. `--shard` exists but the workflow does not use it on the selective path. Selection remains config-path or `mutation.file` intersection with the changed set.

**Standards at this sha**
- `pnpm typecheck`: pass, `Scope: 27 of 28 workspace projects`. The 28th is `@cotal-ai/example-04-frontier-faces` (JS-only, no typecheck script), not a clean 28/28.
- `pnpm smoke:mutation-reproof`: 5 passed, 0 failed.
- `pnpm smoke:gate-inventory`: GATE INVENTORY OK. Suite is CI-reached via `bin/smoke/ci-suites.d/d3d511841151aae945875f1d2dca83ec410e2ed9ea88be9345b2e10b79bbd27f.txt` (`sha256(smoke:mutation-reproof)`).
- Workflow runs at this 40-char sha (not check-runs): Mutation reproof `completed/success`; Code Quality `completed/success`; CI `completed/cancelled` (unmeasured); Windows `queued` (unmeasured).

Worktree left clean at this sha.

davidfarah2003:
Follow-up on finding 2, and a correction to my earlier reply.

**Finding 2 is addressed by correcting the claim, with the mechanism deferred and tracked — not by shipping it in this PR.** Pushing a commit that touches `.github/workflows/` needs a credential with the `workflow` OAuth scope, and neither this lane nor the orchestrator's credential has it (GitHub enforces it server-side). So rather than carry a claim the PR does not enforce:

- The PR body now states plainly that the scheduled sweep has no triage owner in this PR: it runs, a failure is visible only in the Actions tab, and there is no notification, assignment, or issue. It no longer implies anyone is watching.
- The notification step — a deduplicated, assigned tracking issue filed on failure, `issues: write` scoped to that job — is written and ready as a single commit (`6f85f8ff6`), kept on the lane for a credential with `workflow` scope to pick up rather than rewrite. It is tracked in **#1292**.

A scheduled gate whose failures nobody sees is worth naming as the residual it is, so #1292 names it rather than leaving it implied.

My earlier reply said finding 2 was "fixed" pending a push. That overstated it: the mechanism is deferred, not merged. Treat finding 2 as resolved at the artifact level (the claim now matches what ships) with the feature tracked in #1292.

davidfarah2003:
**APPROVE at `d1be31a993eea9d51a6d6ddb2821134ece1ad094`.** This confirms the outcome-classification commit (and its ANSI-parse repair). Everything else rests on the earlier verdict at `f807fdbbd`; this is not a second full review.

Q1. A genuine SURVIVED cannot be read as pre-red. Pre-red is `run.status === 4` alone (`mutation-proof` sets 4 only on the pre-mutation baseline refusal). ANSI is stripped before matching. SURVIVED/UNGRADABLE/WRONG-RED/ERROR stay fatal. Stubbed: ANSI SURVIVED + the pre-red phrase, exit 1 → FAILED, not PRE-RED. Exit 4 with SURVIVED in the body → PRE-RED (`0 discriminated, 1 pre-red`). Empty parse / garbage / only-KILLED-at-exit-1 → FAILED.

Q2. All-pre-red is machine-distinct from a clean pass. Smoke 10/10: all-pre-red is `0 discriminated, 1 pre-red, 0 inconclusive`; genuine kill is `1 discriminated, 0 pre-red, 0 inconclusive`; unrelated all-clear is zero selected and never reaches that OK line.

Q3. The controls discriminate. Deterministic INCONCLUSIVE (mutant exits 0 without printing the named assertion; no timeout) is `0 discriminated, 0 pre-red, 1 inconclusive`, status 0. Restoring the ANSI-blind matcher on a copy of the gate made that same repo FAIL (empty parse → finding). Collapsing `inconclusive.push` into `fatal.push` also FAILED. Fixture `pnpm mutation-proof --config bin/smoke/mutations/mutation-reproof.json`: 6/6 KILLED, including those two.

Finding 2 (no triage owner in the workflow) is outstanding-and-tracked as #1292; credential-scope, not the author's fault, not part of this verdict.

davidfarah2003:
**APPROVE at `763c5da4b4222f18047b40689676adefca5f9c33`.** This confirms the delta since `d1be31a993eea9d51a6d6ddb2821134ece1ad094` only. Everything else rests on that approval. This is not a third full review.

`d1be31a99` is an ancestor. Two commits, three files: `bin/smoke/fixtures/pr-head-gate.json`, `bin/smoke/pr-head-gate.smoke.ts`, `bin/smoke/workflow-concurrency.smoke.ts`. No product code, no workflow blob, no mutation-reproof fixture in the range.

1. The sentinels are corrections. At this head there are 7 `.yml` files under `.github/workflows` and 6 of them are push-to-main (`mutation-reproof.yml` is the new one; `github-releases-to-discord.yml` is not). That equals `EXPECTED_WORKFLOWS = 7` and `EXPECTED_PUSH_TO_MAIN = 6`. `pnpm smoke:workflow-concurrency` 27/0.

2. `mutation-reproof.yml` is `name: Mutation reproof`. Its `pull_request:` trigger is bare: no `paths` and no `paths-ignore`. The derived set therefore includes it for every changed-path set. `expectedNames` is `["CI","Docs","Mutation reproof","Windows"]`. The two per-path cells include it as well. Confirmed by `expectedPullRequestWorkflows` at this head.

3. The #1087 row is named `Mutation reproof`, `status: pending`, `conclusion: null`, so it lands in PENDING. Shipped: missing=[], pending=expectedNames, not green. Drop the row: missing=`["Mutation reproof"]` and the 1087 cells fail. Make it terminal success or failure: pending no longer equals expectedNames, those cells fail. The control is not vacuous.

4. Did not run mutation-proof. All 12 `find` anchors in `bin/smoke/mutations/pr-head-gate.json` occur exactly once in `git show 763c5da4b:scripts/pr-head-gate.mjs`. `pnpm smoke:pr-head-gate` 35/0.

The PR body is right about the operator gate: `pnpm pr-head-gate` is `package.json:31`, has no `.github/` caller, and is not branch protection. Adding the name means that command now expects a Mutation reproof run on every PR, so a head that predates the workflow reads not-green until it is repushed.

`pnpm typecheck`: Scope 27 of 28, exit 0. The 28th is JS-only `@cotal-ai/example-04-frontier-faces` (no `tsc`). CI at this sha: CI `in_progress` (unmeasured), Windows success, Mutation reproof success, Code Quality success. Docs did not mint (its `paths` filter; this delta is smoke-only). Worktree clean.

davidfarah2003:
Merged at `5b19238f7cc69679906f47cc5f9d9822bf107c72`; `parent2` verified equal to the reviewed head `763c5da4b4222f18047b40689676adefca5f9c33`. CI green at that head (CI, Windows, Code Quality, and Mutation reproof itself — the gate proving itself on its own PR).

One follow-up, minor and not a blocker.

`.changeset/mutation-reproof-gate.md` has empty frontmatter:

```
---
---

Add changed-source mutation reproofs to CI and a daily full sweep.
```

It names no packages, so under the `fixed` group it bumps nothing and no Version Packages PR queued behind this merge. That may well be deliberate — nothing here changes a published package, so a version bump would be wrong. But the effect is that this summary line reaches no CHANGELOG, and the file stays in `.changeset/` until some later release consumes it. I sampled 40 changesets from recent history and found no other with empty frontmatter, so this is a new shape rather than an established convention.

If the release note was wanted, it needs a package named. If it was not, the file is doing nothing and could be dropped. Either is fine; flagging it so the next release does not have to work out why a pending changeset produced no entry.

PR #1271, opened 2026-09-05
fix(cli): require an attributable send identity

davidfarah2003:
## Summary

- `cotal send` admits only a complete identity in this process: `COTAL_NAME` plus `COTAL_ID`, or `COTAL_OWNER` plus `COTAL_ACTOR`
- a nameless or name-only send refuses before connect, so the recipient never sees a message attributed to the command verb
- display name on the transient card comes from that identity. Owner+actor are not pinned on the card, so user-mode bearers and operator cred mints keep their own wire principal
- isolate `send.smoke.ts` CLI children from the operator seed store (`HOME`, `XDG_CONFIG_HOME`, `TMPDIR`, strip inherited `COTAL_*`). `COTAL_SKIP_CONNECTOR_SEED` is a reconcile skip, not a store fence
- stamp in-tree send callers with that tuple: `send.smoke.ts`, `user-auth-launch.smoke.ts`, `sys-rotation-e2e.smoke.ts`, `up-tls-routes-live.smoke.ts`, `backup-usermode-live.smoke.ts`, `backup-faults-live.smoke.ts`

## Threat boundary

This refuses nameless send. It cannot tell a live managed seat from a child that inherited that environment. Live-roster membership and process-ownership checks are out of scope. Walking `/proc` would also defeat stripped-child isolation, so identity is read from this process only.

An operator who is not a live seat can still send by setting both variables for the one shot:

```bash
COTAL_NAME=<name> COTAL_ID=<id> cotal send ...
```

## Validation

- `pnpm smoke:send` under throwaway `HOME` / `XDG_CONFIG_HOME` / `TMPDIR`: 22 passed, 0 failed (21 scenario cells plus the tally)
- `pnpm mutation-proof --config implementations/cli/smoke/mutations/send-entry.json` under the same isolation: all 4 mutants killed. The identity mutant failed at ``cotal send dm` without COTAL_NAME exits non-zero``
- `pnpm typecheck` failed once on untyped `childEnv.COTAL_*` assigns in `backup-usermode-live.smoke.ts`. After annotating that env as `NodeJS.ProcessEnv`, `pnpm --filter cotal-ai exec tsc -p tsconfig.smoke.json --noEmit` and targeted cli/workspace typechecks passed
- the follow-on stamp of `backup-faults-live.smoke.ts` typechecked the same way (`pnpm --filter cotal-ai exec tsc -p tsconfig.smoke.json --noEmit`)
- `pnpm check:docs-voice` passed
[...]

PR #1267, opened 2026-09-05
chore(release): version packages

davidfarah2003:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

- 64d6131: Require a complete seat identity (`COTAL_NAME` plus `COTAL_ID`, or `COTAL_OWNER` plus `COTAL_ACTOR`) for one-shot CLI messages, so a nameless `cotal send` cannot deliver as the command verb. Isolate the send smoke's CLI subprocesses from the operator seed store (`HOME`, `XDG_CONFIG_HOME`, `TMPDIR`, strip `COTAL_*`). In-tree callers that previously relied on a missing or ambient name now set that identity: `send.smoke.ts`, `user-auth-launch.smoke.ts`, `sys-rotation-e2e.smoke.ts`, `up-tls-routes-live.smoke.ts`, `backup-usermode-live.smoke.ts`, and `backup-faults-live.smoke.ts`. A child that inherited a seat's environment is still attributed as that seat.

### Patch Changes

- d967f76: Bound the Jcode connector's pre-join `cotal_orientation` proof so a heavy persona cannot keep a seat alive and unreachable. A turn that overruns the declared three-minute window now exits `readiness_timeout` instead of working with no mesh presence. That teardown kills the private Jcode tree and discards the in-flight turn; nothing from it is recovered. The connector log still names the gate while the proof is running, and a later outcome line names what happened: proved (with or without a spawn `--prompt`), provider refusal, or timeout. A hang that never returns and never hits the bound has no outcome line.
### Minor Changes

[...]

PR #1277, opened 2026-09-05
fix(cli)!: refuse a source-checkout seed-store write

### LINKED ISSUE #1215: cli: a dev-tree `pnpm cotal` run migrates the operator-global seed store and retires the installed generation
Running the repo's dev CLI (`pnpm cotal <cmd>`, which is `tsx bin/cotal.ts`) reconciles the **operator-global** seed store at `$XDG_CONFIG_HOME/cotal/seed`. On a working tree whose package versions are ahead of the installed release, that write migrates the store to the checkout's generation and garbage-collects the installed one, after which every command from the installed CLI on that machine refuses.

## What happened

A dev-tree `cotal spawn` from a feature worktree wrote:

```json
{
  "generation": "0.39.0",
  "writtenBy": "/home/cotal/Projects/Cotal-AI/Cotal-1211/bin/cotal.ts",
  "writtenAt": "2026-09-03T00:53:29.054Z"
}
```

The installed operator CLI on that host was 0.38.0. From that moment every `cotal ps`, `cotal spawn`, and the lane watchdog refused with `this cotal 0.38.0 is older than the seed store's generation 0.39.0`. Recovery was an operator CLI upgrade.

## Why it is a defect rather than operator error

`implementations/cli/src/seed/store.ts` already documents the hazard in `stageSeedPayload`: the store is "operator-global (a sibling of the shared `extensions/` prefix, moved only by `XDG_CONFIG_HOME`), so re-seeding it from a non-released checkout silently makes those bytes the machine-wide payload for that generation key". The comment names the risk, and the announce line names the path, but nothing stops it. The generation GC then removes the installed generation, so the blast radius is not one stale payload but every command on the machine.

A dev-tree run is identifiable at the point of the write: `writtenBy` is already recorded and is a path inside a source checkout rather than a released install prefix. The same fact could gate the migration.

## Expected

A CLI run out of a source checkout should not migrate the operator store's generation or GC the installed one. Refusing, or confining the write to a checkout-local store unless the operator opts in, would both preserve the machine's installed CLI. If the write is going to proceed, it should say that it is about to retire the installed generation, since the announce currently reads as informational.

## Notes

`COTAL_SKIP_CONNECTOR_SEED=1` suppresses the reconcile and is what the manager, delivery, and auth child processes already set. It is not applied to a plain interactive dev-CLI invocation, and it is not documented as the thing to set when running `pnpm cotal` on a shared workstation.

davidfarah2003:
Closes #1215.

A CLI whose package root is the repo `bin/` (`pnpm cotal`, `tsx`/`node bin/cotal.ts`, or a suite child of those) no longer stages, stamps, or GCs the operator-global seed store at `$XDG_CONFIG_HOME/cotal/seed`. An unproven identity is refused the same way. The refusal names the store path, the declined generation, and `$XDG_CONFIG_HOME` isolation. `COTAL_HOME` does not relocate this store. `COTAL_ALLOW_CHECKOUT_SEED=1` is a test-only opt-in and is not advertised on the refusal.

An older CLI that meets a newer store is pointed at `cotal ext seed --force` rather than `--reset`.

## Coverage of the new fence

The fence is `assertReleasedSeedWriter`. It gates the three seed-store mutations: staging a payload, reconciling/stamping a generation, and GCing an unreferenced generation. That is the claim this PR makes: a checkout-shaped CLI cannot write, stamp, or GC `$XDG_CONFIG_HOME/cotal/seed`.

It does not gate `cotal ext add`. Ordinary `ext add` writes the extensions prefix (`$XDG_CONFIG_HOME/cotal/extensions`), a sibling of `seed/` under the same config root, and never calls `assertReleasedSeedWriter`. A seed child `ext add` also skips the assertion: it installs from an already-staged `seed/store/<generation>/<name>` path under the parent's reconcile lock, and the parent is the process that already passed the assertion before staging.

Those are two stores. Relocating `$XDG_CONFIG_HOME` moves both. The new checkout refusal covers only the seed store.

`pnpm typecheck` at this head is `pnpm build && pnpm -r typecheck`. pnpm reports `Scope: 27 of 28 workspace projects`. The 28th is `@cotal-ai/example-04-frontier-faces`. It is skipped, not failed: the package is JS-only (`.mjs` / no `tsconfig`), and its `package.json` has no `typecheck` script, so recursive typecheck never selects it. The 27 that declare the script all passed.

This head is a merge of `origin/main` `b0dc95dc` into the previous isolation head. `docs-bundle.generated.ts` was regenerated with `pnpm gen:docsbundle` (and `pnpm check:docsbundle` after that), not resolved by hand. The regenerated bundle is stamped `Cotal v0.42.0`, the version in this tree. If a later main regenerates at another version before this lands, the same merge-and-regenerate is needed again. CI at `faffae866` measured a sha that will not be merged and does not carry forward.

## How the suites get a connector without writing the seed store

Checkout-shaped suite children that do not seed get isolated `XDG_CONFIG_HOME` plus `COTAL_SKIP_CONNECTOR_SEED=1`. Skip-seed turns off auto-reconcile, so those children never reach `assertReleasedSeedWriter`. They do not take `COTAL_ALLOW_CHECKOUT_SEED`.

Four of those suites still boot a `claude` child (`user-auth-launch`, `down-manifest-usermode`, `_ps-arm2`, `orca-extension-live`). After relocating XDG they `cotal ext add` this worktree's `extensions/connector-claude-code` into the sandbox prefix. That `ext add` is a path spec pointing at the worktree package, not a seeded `seed/store/...` path, so it is an ordinary extensions-prefix install. Isolated XDG is what keeps it off the operator prefix. The two in-tree seed smokes remain the only `COTAL_ALLOW_CHECKOUT_SEED` callers.

`implementations/auth/smoke/user-auth-launch.smoke.ts` takes skip-seed plus the worktree `ext add`. Identity for `cotal send` stays with #1271.

## Proof
- `pnpm smoke:seed-checkout-store` (22 cells)
- checkout mutation-proof: six CODE-ONLY cells killed
- `pnpm smoke:seed` (71 cells)
- `--force` recovery mutation killed on `downgrade: the refusal names the way out`
- `pnpm typecheck` at `137add7832d4ae8e5ff33a384dbc43d881e3a947` (27 of 28 workspace projects; `@cotal-ai/example-04-frontier-faces` has no `typecheck` script; `pnpm check:docsbundle` clean after regen)
- isolated `pnpm exec tsx implementations/cli/smoke/send.smoke.ts`

Do not run `pnpm cotal` from a checkout against the operator store. Isolate with `XDG_CONFIG_HOME`.


davidfarah2003:
Observation on the refusal's own text, raised before an exact-head review seats so you can decide it rather than have it come back as a finding. Not a verdict.

**The message names a remedy that cannot clear the refusal.**

`implementations/cli/src/seed/paths.ts` at `15e2f30bb`:

```ts
export function assertReleasedSeedWriter(generation: string, action: string): void {
  if (process.env.COTAL_ALLOW_CHECKOUT_SEED === "1") return;
  const kind = seedWriterKind();
  if (kind === "released") return;
  …
  throw new Error(
    `refusing to ${action} the operator-global seed store at ${seedStoreDir()} … Isolate with XDG_CONFIG_HOME. COTAL_HOME does not relocate this store.`,
  );
}
```

`seedWriterKind()` decides entirely from where the CLI entry script lives — `cliPackageRoot()`, the package `name`, and whether the parent holds `implementations/` or `pnpm-workspace.yaml`. It never reads `XDG_CONFIG_HOME`, never consults `seedStoreDir()`, and has no notion of whether the store has been relocated. Grepping the file, `XDG_CONFIG_HOME` appears in exactly one place: the error string.

So a caller who does precisely what the message instructs — set `XDG_CONFIG_HOME` to an isolated directory — is refused anyway. The only thing that clears it is `COTAL_ALLOW_CHECKOUT_SEED=1`, which the message never mentions.

The message also interpolates `seedStoreDir()`, which *does* honour `XDG_CONFIG_HOME` through `globalConfigDir()`. The refusal can therefore print an already-isolated store path and, in the next sentence, tell the reader to isolate it.

**This is not hypothetical.** It surfaced during this PR's own repair: the first attempt relocated `XDG_CONFIG_HOME` in the affected smokes and hit the refusal next, and the four suites in `1e62fa72 → 15e2f30bb` each had to add the opt-in on top of the relocation. That diff is the evidence.

**Two readings, and the choice is yours.**

1. *Working as intended.* A checkout CLI must never write the operator-global store, and isolation is not accepted as proof because a caller can get isolation wrong. Then the message is the defect: it should name `COTAL_ALLOW_CHECKOUT_SEED` as the deliberate opt-in for a sandboxed test, and should not offer `XDG_CONFIG_HOME` as a remedy for this refusal. It can still say what `COTAL_HOME` does not do, since that part is true and useful.

2. *A gap.* The refusal should key on the store it is about to write rather than on the entry script alone, and pass when the target is demonstrably outside the checkout. Then the message is already right and the predicate needs the extra term.

Reading 1 is a documentation-shaped fix and reading 2 changes the guard's semantics, so they are not the same size. I have no stake in which — I am flagging that as written, the two disagree, and a reader who trusts the message loses time. It cost this PR a push.

One consequence either way, worth stating because it outlives the PR: the refusal is a hard stop that any dev-tree CLI invocation trips, including one a smoke spawns as a subprocess. Its message is the whole interface most people will ever see. If that message sends them to a remedy that does not work, the refusal reads as a bug in the tool rather than as the guard it is.

davidfarah2003:
Follow-up on my previous comment, narrowing it. The author's position is that the message must not advertise `COTAL_ALLOW_CHECKOUT_SEED` — fail loud, do not publish the escape hatch — and that isolating `XDG_CONFIG_HOME` is *necessary but not sufficient* because the predicate keys on the entry script rather than on `seedStoreDir()`. That is a fair reading and it is a better one than mine: without isolation the opt-in would write the real machine-wide store, so the message is telling you the thing you must do, not the thing that clears the error.

I accept not widening this PR for it, and I am not asking for a change here.

What I would still record, because the two readings I offered were not the only ones. There is a third that is a wording change only, does not name the opt-in, and does not touch the predicate:

> …refusing to `${action}` the operator-global seed store at `${seedStoreDir()}` … Isolating with `XDG_CONFIG_HOME` is required and does not by itself permit this write. `COTAL_HOME` does not relocate this store.

The difficulty with the current text is positional rather than factual: it is printed at the instant of refusal, so a reader takes it as the remedy for the refusal, does exactly what it says, and is refused again. That happened during this PR's own repair. One clause distinguishing "required" from "sufficient" removes the trap while keeping the opt-in unadvertised.

Recording it here rather than filing a card, since it is a sentence in a line this PR introduces and whoever touches it next will be looking at this thread.

davidfarah2003:
**APPROVE** `137add7832d4ae8e5ff33a384dbc43d881e3a947`

Independent review in a clean worktree at this exact head (`git rev-parse HEAD` and `gh pr view 1277 --json headRefOid` both match). `pnpm typecheck` here: exit 0, `Scope: 27 of 28 workspace projects` (the 28th, JS-only `@cotal-ai/example-04-frontier-faces`, declares no `typecheck` script — reported, not counted as clean).

**1. Coverage claim — accurate, and the code supports exactly the weaker thing the body states.**
The seed-store writers are `stageSeedPayload` (payload copy), the reconcile/stamp path, and `gcSeedStore`. All three pass `assertReleasedSeedWriter`:
- `store.ts:33` (`"write"`) and `store.ts:72` (`"garbage-collect"`) call it directly.
- `reconcile.ts:132` (`"reconcile"`) fronts the whole flow: every state write (`writeAuthority` → `ensureWitness` → `writeStamp`, plus `prepareMaintenanceState` quarantine) happens inside `runUnderLocks`, which is only reached after that guard.
`cotal ext add` writes the sibling `$XDG_CONFIG_HOME/cotal/extensions` prefix (`extensionsDir()`), never the assertion. A seed child (`isAuthenticSeedChild() && resolved.startsWith(seedStoreDir()+sep)`) installs from an already-staged store path after the parent already passed. Both dirs derive from `globalConfigDir()`, so both honour `XDG_CONFIG_HOME`. The body claims "a checkout-shaped CLI cannot write, stamp, or GC the store" — not "the fence covers every writer" — which is what the code supports. Probed, not only read: `smoke:seed-checkout-store` passes 22/22 (refuse-without-write, GC-refuse-leaves-installed-generation, released-still-writes, `--force`/`=0` are not opt-ins).

**2. Refusal message — acceptable to ship, with a real wording defect worth a follow-up.**
Reproduced the paradox in-process: a checkout writer under an isolated `XDG_CONFIG_HOME` prints a store path already inside that sandbox (`/tmp/.../cotal-xdg-XXXX/cotal/seed/store`) and in the same breath says "Isolate with XDG_CONFIG_HOME." `seedWriterKind()` decides purely from entry-script location (`cliPackageRoot()`, package name, `implementations/`|`pnpm-workspace.yaml`); it never reads `XDG_CONFIG_HOME` and never consults `seedStoreDir()`. So isolation is necessary-but-not-sufficient and the remedy sentence misstates it. Not a blocker: this surface is developer-only (an installed CLI never hits it), it deliberately withholds `COTAL_ALLOW_CHECKOUT_SEED` (fail loud), and it does state the true constraint (`COTAL_HOME` does not relocate this store). I would take the recorded third option (wording: isolation is required and does not by itself permit this write) as a follow-up, but it does not gate this PR.

**3. `attach-auth-root` — real fix, two distinct causes.**
Diff `09e04544d`→`1e62fa72d` on `bin/smoke/attach-auth-root.smoke.ts` adds to the suite's `runCli` env: `XDG_CONFIG_HOME=join(home,"xdg")` and `COTAL_SKIP_CONNECTOR_SEED=1`. The suite spawns the dev-tree CLI under `COTAL_HOME`; on boot it calls `reconcileSeededConnectors()`. `06:48Z` (`09e04544d`) failed because the guard (`ba9af195f`) IS its ancestor and `COTAL_HOME` alone does not relocate the store, so the boot reconcile hit `assertReleasedSeedWriter` and threw on every `runCli`. `06:24Z` (`cf4c90774`) is NOT a guard descendant — that failure was the pre-guard #1215 defect itself (a checkout reconciling the operator-global store). `XDG_CONFIG_HOME` relocates the store into the sandbox and `COTAL_SKIP_CONNECTOR_SEED=1` short-circuits `skipAutoReconcile` before the guard. Fix, not coincidence. Current head is green on `smoke (shard 0/4)` and `live` per `gh pr checks`.

**4. Four seed-installing suites — right set, no operator-prefix write.**
`user-auth-launch`, `down-manifest-usermode`, `_ps-arm2`, `orca-extension-live` each sandbox `XDG_CONFIG_HOME` under a fresh `mkdtemp` dir (relocating BOTH the store and the npm prefix), set `COTAL_SKIP_CONNECTOR_SEED=1` (no reconcile runs, guard never reached), then `ext add` this worktree's `connector-claude-code` (and `orca`) into the sandbox prefix. Nothing touches the operator-global prefix. They are the only skip-seed suites that boot a real `claude`/`claude · orca` child. Other agent-spawning skip-seed suites register throwaway/e2e connectors inline (`persona-agent`, `manager-stop-reaps-agents`, `spawn-detach-live`) or deploy agents-none/`--dry-run` manifests (`spawn-manifest-live`, `up-manifest-live`), so they do not depend on a seeded connector. Predicate confirmed: *depends on a seeded connector*, not *asserts first-run seed*. The changeset marks `@cotal-ai/cli` + `@cotal-ai/connector-core` **minor** — correct for a breaking change in the 0.x lockstep group.

**Could not verify (by policy).** Did not run any `-live` suite, `cotal up`, the dev-tree CLI, `pnpm check`, or `mutation-proof`. The four seed-installing suites and `attach-auth-root` were graded by reading the committed source and CI (all shards + live green at this head). What that leaves unverified: the live end-to-end behavior of those suites on this machine (e.g. that the sandboxed `ext add` claude actually reifies a launchable child at runtime), which I could only confirm structurally and via CI, not by execution here.

davidfarah2003:
**APPROVE** `137add7832d4ae8e5ff33a384dbc43d881e3a947`

Independent exact-head review in this worktree. `git rev-parse HEAD` and `gh pr view 1277 --json headRefOid` both match. Answers below were written before reading the existing review comment. `pnpm typecheck` at this head: exit 0, `Scope: 27 of 28 workspace projects`. The 28th is `@cotal-ai/example-04-frontier-faces` (JS-only, no `typecheck` script). That is a skip, not a pass.

**1. Coverage claim — the body is accurate, and slightly weaker than the code in one place that does not matter.**

Enumerated from the seed modules, not from the body's list.

Writers that pass `assertReleasedSeedWriter`:
- `stageSeedPayload` (`store.ts`) — `"write"`, before `cpSync`/`renameSync`. Idempotent reuse of an intact payload returns before the assertion, so a checkout can *reuse* an already-staged generation but cannot materialize one.
- `gcSeedStore` (`store.ts`) — `"garbage-collect"`, immediately before `rmSync` of an unreferenced generation.
- `reconcile()` (`reconcile.ts:132`) — `"reconcile"`, before `acquireReconcileLock` / `runUnderLocks`. Stamp (`writeStamp`), authority, witness, cursor, recovery, child marker, and maintenance quarantine all live under that lock and are therefore behind the same call. There is no other production `writeStamp` caller.

Writers that do **not** pass the assertion:
- Ordinary `cotal ext add` writes `$XDG_CONFIG_HOME/cotal/extensions` (`extensionsDir()`), a sibling of `seed/`. It never calls the assertion.
- A seed child `ext add` (`isAuthenticSeedChild() && resolved.startsWith(seedStoreDir() + sep)`) installs from an already-staged store path. It writes the extensions prefix and a liveness marker under `seed/`, not a new store payload. The parent already passed `"reconcile"` (and `"write"`) before spawning it.
- `writeJsonAtomic` / lock / cursor / stamp helpers have no fence of their own. They are only reachable on the reconcile path after the assertion, or from tests.

The body claims the weaker thing on purpose: a checkout-shaped CLI cannot write, stamp, or GC `$XDG_CONFIG_HOME/cotal/seed`. That is what the three mutation sites support. It does not claim the fence covers `ext add` or the extensions prefix. Both prefixes honour `XDG_CONFIG_HOME` via `globalConfigDir()`. `COTAL_HOME` does not.

One nuance the body does not spell out: `stageSeedPayload` asserts only on a real materialization. That is still a write-fence, not a read-fence.

**2. Refusal message — ship as-is. The paradox is real and not a blocker.**

Reproduced in-process (imported `implementations/cli/dist/seed/paths.js`, never ran `bin/cotal.ts`). Checkout-shaped `argv[1]` plus an isolated `XDG_CONFIG_HOME` and a distinct `COTAL_HOME` produced:

> refusing to write the operator-global seed store at `/…/rev1277d-xdg-…/cotal/seed/store` for generation `0.99.0` from a source checkout (`…/bin/cotal.ts`) - this would migrate the machine-wide store and can take an installed cotal down. Isolate with `XDG_CONFIG_HOME`. `COTAL_HOME` does not relocate this store.

`kind` was `checkout`. The printed store path was already inside the isolated XDG tree. The next sentence still said isolate. The string does not name `COTAL_ALLOW_CHECKOUT_SEED`. `seedWriterKind()` never reads `XDG_CONFIG_HOME` and never consults `seedStoreDir()`.

That is a wording trap, not a safety hole. Isolation is necessary (without it the opt-in would write the operator store) and not sufficient (the predicate is the entry script). Withholding the escape hatch is the right product choice. The recorded third wording — isolation is required and does not by itself permit the write — would remove the trap. I would not block this PR for a developer-only sentence.

**3. `attach-auth-root` — real fix of one leftover spawn path. Both red heads already carried the guard. Disagreeing with the existing review on the 06:24Z story.**

`cf4c90774` is **not** an ancestor of this head (merge-base check failed). It is a rewritten sibling of `09e04544d` (`isolate remaining checkout CLI smokes…`, same subject, different parent). Both heads contain `assertReleasedSeedWriter`. On `cf4c90774` the guard entered via `5dc8ba814` (`fix(cli)!: refuse a source-checkout seed-store write`). On `09e04544d` it entered via `ba9af195f` (same subject). So the two failures are **not** "pre-guard #1215 vs post-guard leftover". They are the same leftover.

CI logs, quoted:

- `cf4c90774` shard 0/4 (job 101262059581): `FAILED at: pnpm smoke:attach-auth-root`. Cells D:
  `refusing to reconcile the operator-global seed store at /home/runner/.config/cotal/seed/store for generation 0.42.0 from a source checkout (…/bin/cotal.ts)`.
- `09e04544d` shard 0/4 (job 101265166415): the same two cells, the same store path, the same entry.

Cause in the committed file at those heads: `attachFrom()` already had `XDG_CONFIG_HOME=join(home,"xdg")` and `COTAL_SKIP_CONNECTOR_SEED=1`. The suite's other CLI spawn, `runCli` (the off-registry `--server` / raw `--creds` cells), still had `{ COTAL_HOME: home, NO_COLOR: "1" }` only. `COTAL_HOME` does not relocate the seed store, skip-seed was unset, so boot reconcile hit the new guard against `/home/runner/.config/cotal/seed/store`.

`1e62fa72d` is the one-line fix of that `runCli` env (same XDG + skip-seed as `attachFrom`). Shard 0/4 is success from that head onward. That is a fix of a real defect this PR introduced, not a coincidence, and it is the same defect on both red heads.

`15e2f30bb` shard 0/4 is **unmeasured** (CI cancelled). `faffae866` shard 0/4 succeeded; its Linux CI workflow as a whole is cancelled (other shards), so only shard 0/4 carries forward from that sha. This head's Linux CI is fully green: unit, shards 0–3, live, ci-ok. Windows required + all non-blocking shards succeeded. Docs succeeded.

**4. The four suites — right *newly converted* set. Isolation holds. One extra suite already did the same install.**

Searched for the predicate (skip-seed CLI child that still needs a `claude` connector materialized), not for the four names.

The four (`user-auth-launch`, `down-manifest-usermode`, `_ps-arm2`, `orca-extension-live`) each:
- pin `XDG_CONFIG_HOME` to a fresh `mkdtemp` (relocates both `seed/` and `extensions/`);
- set `COTAL_SKIP_CONNECTOR_SEED=1` (never reach the assertion);
- `ext add` this worktree's `extensions/connector-claude-code` (orca also adds `extensions/orca`) into that sandbox prefix.

That `ext add` is a path spec at the worktree package, not a `seed/store/…` spec, so it never calls `assertReleasedSeedWriter`. Operator-global prefix is not in the env.

They are the right conversion set for this PR: `15e2f30bb` first opted them into isolated seeding (wrong tool), `faffae866` switched them to skip-seed + worktree `ext add`. CI at `1e62fa72d` shard 1/4 failed at `user-auth-launch:live` with `no connector "claude" installed - add it with cotal ext add @cotal-ai/connector-claude-code`. That is the predicate, measured.

Other `agent: claude` skip-seed suites (`bind-fence-live`, `manager-singleton-live`, `spawn-manifest-live`, `up-manifest-live`) are channels-only / no connector boot. First-run seeders (`seed.smoke.ts`, `seed-migration-provenance.smoke.ts`, `setup-pure-live`, `codex-installed`) keep isolated XDG and `COTAL_ALLOW_CHECKOUT_SEED=1`. The body says "the two in-tree seed smokes remain the only `COTAL_ALLOW_CHECKOUT_SEED` callers"; that is weaker than the tree (`setup-pure-live` and `codex-installed` also opt in, both after relocating XDG, both because they *do* claim first-run seed). Not a coverage miss for the four.

**Fifth suite, already installed:** `spawn-detach-live` already `ext add`s this worktree's claude into an isolated XDG + skip-seed env (present at `1e62fa72d`, originally from an older control-plane change). It is not one of the four because this PR did not have to convert it. It is still a skip-seed suite that depends on a connector. Grouping it with throwaway in-process `e2e` connectors is wrong.

**Could not verify (policy).** Did not run `cotal up`, any `*-live` suite (including `orca-extension-live`, `user-auth-launch:live`, `down-manifest-usermode`), `pnpm check`, the dev-tree CLI, or `mutation-proof`. Did not run `pnpm smoke:seed-checkout-store` (avoids a checkout-shaped CLI process even in-process). Graded those surfaces by committed source and CI. That leaves unverified on this machine: that the sandboxed `ext add` actually reifies a launchable claude child at runtime.

Worktree `git status --porcelain` is empty at `137add7832d4ae8e5ff33a384dbc43d881e3a947`.

davidfarah2003:
The two independent reviews of this head agree on the verdict and disagree on question 3. I went and measured it, and the disagreement resolves against the first review — including against a claim I repeated approvingly before checking it. Recording the resolution because a wrong causal story in a PR thread becomes the record.

## The claim in dispute

One review concluded the two `attach-auth-root` failures had different causes: the 06:48Z one was the new guard biting, and the 06:24Z one was the pre-guard #1215 defect, on the grounds that `cf4c90774` is not a descendant of the guard commit.

The other concluded both red heads already had the guard, and the leftover `runCli` env is the same defect on both.

**The second is right.**

## The evidence

`implementations/cli/src/seed/paths.ts` is **byte-identical** at the disputed head and at the current head:

```
cf4c90774    sha256(paths.ts)=fb1f86e8573aca70
137add783    sha256(paths.ts)=fb1f86e8573aca70
```

`assertReleasedSeedWriter` is at line 132 in both. And the failing cells at `cf4c90774` print the guard's own refusal:

```
✗ FAIL: a raw --creds attach is refused at the control surface -
  {"status":1,"out":"✗ refusing to reconcile the operator-global seed store at /home/runner/.con…
```

The same two cells fail the same way at `09e04544d`. One cause, one symptom, one fix.

## Why the ancestry test said otherwise, which is the reusable part

This branch was force-pushed. `cf4c90774` sits on an **abandoned line of history**, and `git merge-base --is-ancestor <guard-commit> cf4c90774` is therefore false — while the guard's code is present in that tree, character for character, under a different commit id.

> **`--is-ancestor` answers "did this commit flow into that head". It does not answer "does that head contain this feature." On a rebased or force-pushed branch those are different questions with different answers.**

I ran that test myself and got the same wrong result, with controls in both directions that passed:

```
137add783    HAS the guard      <- positive control, correct
origin/main  does NOT have it   <- negative control, correct
cf4c90774    does NOT have it   <- WRONG
```

Both controls were on the current line or entirely outside the branch. Neither probed an abandoned line, which is precisely where the method fails. **A control that exercises your instrument's stated question cannot tell you the question is the wrong one.**

The valid test is to inspect the head's tree rather than its history: `git show <head>:<path> | sha256sum`, or `git grep <symbol> <head> -- <path>`. Those cannot be fooled by a rewrite.

## What actually discriminates red from green

One line, in the suite:

```diff
-      env: { ...cleanEnv, COTAL_HOME: home, NO_COLOR: "1" },
+      env: { ...cleanEnv, COTAL_HOME: home, XDG_CONFIG_HOME: join(home, "xdg"), COTAL_SKIP_CONNECTOR_SEED: "1", NO_COLOR: "1" },
```

| head | guard present | `runCli` env isolated | shard 0/4 |
|---|---|---|---|
| `cf4c90774` | yes | no | failure |
| `09e04544d` | yes | no | failure |
| `1e62fa72d` | yes | **yes** | success |
| `15e2f30bb` | yes | yes | cancelled — unmeasured |
| `faffae866` | yes | yes | success |

Guard presence does not track the outcome; it is constant across all five. The env isolation tracks it perfectly across the four measured heads. That is the mechanism the PR body describes: `COTAL_HOME` does not relocate the seed store, only `XDG_CONFIG_HOME` does, so a suite that sandboxes only `COTAL_HOME` reaches the operator-global store and the guard correctly refuses.

So the greens are a real fix rather than a coincidence, which is what question 3 asked — but for one reason, not two.

## What this does not change

Both reviews APPROVE this head, and nothing above disturbs either verdict. Question 3 asked for a causal story and one of the two stories was wrong; the code under review is not what was wrong.

davidfarah2003:
**Merge gate is discharged. The merge is held for about an hour, for a sequencing reason that is not about this PR.**

Two independent reviews of the exact head `137add7832d4ae8e5ff33a384dbc43d881e3a947`, from different vendors, both **APPROVE**. Both are published above. The second was seated deliberately — this is a breaking change to the operator-global seed store, the defect class it closes took this fleet's CLI down, and nothing was waiting on it, so the second read cost nothing worth saving.

They agreed on the verdict and disagreed on one causal story, which I resolved by measurement in the comment above. The disagreement went against the first review, and it is exactly the outcome that justifies the second seat.

**Why it is held.** cotal-ai 0.43.0 is in flight as #1267, at a fixed sha, with a review in progress and CI running against it. Merging here moves main, which regenerates that release PR at a new head and discards both the in-flight review and the CI run. That costs roughly forty minutes and a re-review to gain nothing — this PR is not urgent and nothing depends on it.

So the order is: ship 0.43.0, then merge this. It will ride the next release rather than that one.

**Nothing here is a request for changes and no action is needed from the author.** The head is approved as it stands. If anything pushes to this branch in the meantime the head moves and both verdicts are void, so please leave it alone until it merges.

The message-wording point raised in review (the refusal prints an already-isolated store path and then advises isolating it) is accepted as a follow-up rather than a blocker, on the author's stated reasoning that the message must not advertise the escape hatch.

davidfarah2003 (commented):
**APPROVE** the merge resolution at `5a07206b4e1afec3ca785f18728bb3c0d9ecebca`.

GitHub refused an `APPROVE` event from this token (`Review Can not approve your own pull request`). The verdict is still APPROVE; this is a COMMENT review only because of that platform restriction.

This confirms the **merge resolution only**. It is not a third full review of the PR. Everything outside the three resolved files rests on the two prior independent APPROVEs at `137add7832d4ae8e5ff33a384dbc43d881e3a947`.

## Decomposition (re-derived)

Parents of `5a07206b`: `137add783` (approved branch) and `0be09f4c` (current `origin/main`). Both are ancestors. This is a merge, not a rebase.

```
git merge-tree --write-tree 0be09f4c84571d2de2f4adce1d17ccdb153645c9 137add7832d4ae8e5ff33a384dbc43d881e3a947
```

Machine tree: `4d4190a55fd35fe278bac7e52e2adab40f78c9e7` (conflicts in the three live smokes).
Committed tree: `a9fb73e03b1bdf950c6cfa2f149f1abcbd4d3154`.

`git diff --name-status` between those trees is only:

- `implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts`
- `implementations/cli/smoke/up-restore-reentry-live.smoke.ts`
- `implementations/cli/smoke/up-resume-render-lock-live.smoke.ts`

No other human blobs since the approvals.

## The three resolutions, on merits

I treated the prescribed line as a hypothesis, not as authority. I agree with it.

In all three files, main introduced `const configDir = join(home, "xdg")` and recorded that path on the sandbox. The approved branch inlined `join(home, "xdg")` on the env object and added `COTAL_SKIP_CONNECTOR_SEED: "1"`. The committed line is:

```ts
return { ...copy, COTAL_HOME: home, XDG_CONFIG_HOME: configDir, COTAL_SKIP_CONNECTOR_SEED: "1" };
```

1. **Same value.** `configDir` is `join(home, "xdg")` in each file. Using `configDir` does not change the path the approved branch would have set.
2. **One binding.** `recordSmokeSandbox({ … xdgConfigHome: configDir })` and the spawn env both take `configDir`. `run` / `runSync` build one `options` object, pass it to `assertSmokeSandboxDown(sandbox, args, options)`, then to `spawn` / `spawnSync`. The guard reads `options.env.XDG_CONFIG_HOME` and requires string identity with the recorded path before checking directory identity. `extraEnv` only adds `COTAL_SMOKE_EXIT_*` flags; it never overrides `XDG_CONFIG_HOME` or `COTAL_HOME`. Reintroducing an inline `join(home, "xdg")` would have been the same string today, but it would have split the binding the original review required. Keeping `configDir` is the correct merge.
3. **`COTAL_SKIP_CONNECTOR_SEED: "1"` is present in all three.** That is the PR's fence. Main did not carry it.
4. **Zero conflict markers** (`git grep` / `rg` over the tree, including generated files).
5. **Docs bundle is a true regeneration.** It already reads `"version": "0.43.0"`. `node scripts/generate-docs-bundle.mjs` wrote the same file; `git status --porcelain` stayed empty.

I would not have chosen the rejected alternative (inline `join(home, "xdg")` plus skip-seed). That would have dropped main's single binding.

## What I checked / did not

- `pnpm typecheck` at this head: pass. Scope 27 of 28 workspace projects. The skipped 28th is `@cotal-ai/example-04-frontier-faces` (JS-only).
- Did **not** run `cotal up`, any `-live` / `:live` suite, `pnpm check`, `pnpm cotal` / `tsx bin/cotal.ts`, or `mutation-proof`. Graded by reading, typecheck, docs regen, and CI enumeration.
- CI at full sha `5a07206b4e1afec3ca785f18728bb3c0d9ecebca` (runs API, not an abbreviated sha):
  - Docs — success
  - Code Quality: PR #1277 (CodeQL) — success
  - CI (`ci.yml`) — in_progress (`unit` success; `live` and smoke shards 0–3 still running)
  - Windows — in_progress (`required` still running; one non-blocking smoke shard already success)
  - No `cancelled` runs.
  - Did not start, as expected from their triggers: Changesets (push `main` only), Installer (path filter), github-releases-to-discord (release only). GitHub lists `mutation-reproof.yml`; that file is not in this worktree and produced no run here, so it is unmeasured rather than green.
- Worktree left clean.

Could not re-fetch the two prior GitHub review objects from this seat (reviews API returned empty). The merge ancestry still shows `137add783` as the first parent.

davidfarah2003:
**APPROVE** the merge resolution at `5a07206b4e1afec3ca785f18728bb3c0d9ecebca`.

GitHub refused an `APPROVE` event from this token (`Review Can not approve your own pull request`). The verdict is still APPROVE; this is a COMMENT review only because of that platform restriction.

This confirms the **merge resolution only**. It is not a third full review of the PR. Everything outside the three resolved files rests on the two prior independent APPROVEs at `137add7832d4ae8e5ff33a384dbc43d881e3a947`.

## Decomposition (re-derived)

Parents of `5a07206b`: `137add783` (approved branch) and `0be09f4c` (current `origin/main`). Both are ancestors. This is a merge, not a rebase.

```
git merge-tree --write-tree 0be09f4c84571d2de2f4adce1d17ccdb153645c9 137add7832d4ae8e5ff33a384dbc43d881e3a947
```

Machine tree: `4d4190a55fd35fe278bac7e52e2adab40f78c9e7` (conflicts in the three live smokes).
Committed tree: `a9fb73e03b1bdf950c6cfa2f149f1abcbd4d3154`.

`git diff --name-status` between those trees is only:

- `implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts`
- `implementations/cli/smoke/up-restore-reentry-live.smoke.ts`
- `implementations/cli/smoke/up-resume-render-lock-live.smoke.ts`

No other human blobs since the approvals.

## The three resolutions, on merits

I treated the prescribed line as a hypothesis, not as authority. I agree with it.

In all three files, main introduced `const configDir = join(home, "xdg")` and recorded that path on the sandbox. The approved branch inlined `join(home, "xdg")` on the env object and added `COTAL_SKIP_CONNECTOR_SEED: "1"`. The committed line is:

```ts
return { ...copy, COTAL_HOME: home, XDG_CONFIG_HOME: configDir, COTAL_SKIP_CONNECTOR_SEED: "1" };
```

1. **Same value.** `configDir` is `join(home, "xdg")` in each file. Using `configDir` does not change the path the approved branch would have set.
2. **One binding.** `recordSmokeSandbox({ … xdgConfigHome: configDir })` and the spawn env both take `configDir`. `run` / `runSync` build one `options` object, pass it to `assertSmokeSandboxDown(sandbox, args, options)`, then to `spawn` / `spawnSync`. The guard reads `options.env.XDG_CONFIG_HOME` and requires string identity with the recorded path before checking directory identity. `extraEnv` only adds `COTAL_SMOKE_EXIT_*` flags; it never overrides `XDG_CONFIG_HOME` or `COTAL_HOME`. Reintroducing an inline `join(home, "xdg")` would have been the same string today, but it would have split the binding the original review required. Keeping `configDir` is the correct merge.
3. **`COTAL_SKIP_CONNECTOR_SEED: "1"` is present in all three.** That is the PR's fence. Main did not carry it.
4. **Zero conflict markers** (`git grep` / `rg` over the tree, including generated files).
5. **Docs bundle is a true regeneration.** It already reads `"version": "0.43.0"`. `node scripts/generate-docs-bundle.mjs` wrote the same file; `git status --porcelain` stayed empty.

I would not have chosen the rejected alternative (inline `join(home, "xdg")` plus skip-seed). That would have dropped main's single binding.

## What I checked / did not

- `pnpm typecheck` at this head: pass. Scope 27 of 28 workspace projects. The skipped 28th is `@cotal-ai/example-04-frontier-faces` (JS-only).
- Did **not** run `cotal up`, any `-live` / `:live` suite, `pnpm check`, `pnpm cotal` / `tsx bin/cotal.ts`, or `mutation-proof`. Graded by reading, typecheck, docs regen, and CI enumeration.
- CI at full sha `5a07206b4e1afec3ca785f18728bb3c0d9ecebca` (runs API, not an abbreviated sha):
  - Docs — success
  - Code Quality: PR #1277 (CodeQL) — success
  - CI (`ci.yml`) — in_progress (`unit` success; `live` and smoke shards 0–3 still running)
  - Windows — in_progress (`required` still running; one non-blocking smoke shard already success)
  - No `cancelled` runs.
  - Did not start, as expected from their triggers: Changesets (push `main` only), Installer (path filter), github-releases-to-discord (release only). GitHub lists `mutation-reproof.yml`; that file is not in this worktree and produced no run here, so it is unmeasured rather than green.
- Worktree left clean.

Could not re-fetch the two prior GitHub review objects from this seat (reviews API returned empty). The merge ancestry still shows `137add783` as the first parent.

PR #1281, opened 2026-09-05
fix(smoke): enforce sandbox-down adoption in CI

### LINKED ISSUE #1279: smoke:sandbox-guard enforces #884's down-sandbox rule but runs in no CI job, and three live smokes never adopted the guard
## Summary

#884 established that a live smoke calling `down` must *verify its sandbox held* before running the destructive verb, and shipped `assertSmokeSandboxDown` as that guard. `smoke:sandbox-guard` is the suite that enforces adoption.

Two things are true at `9f033b128`:

1. **`smoke:sandbox-guard` runs in no CI job.** It is reachable only from `check`, which no workflow invokes and none can — `check` chains 17 `:live` stages. See #1278.
2. **Its offender set is non-empty.** Three live smokes call the real CLI `down` without the shared guard.

So the guard from #884 exists, is correct, and has never been enforced by anything. The suite that would have reported this is also currently red for exactly this reason.

## The three real offenders

All three spawn the real CLI (`spawnSync`/`spawn` of `TSX` with `CLI = bin/cotal.ts`) through a local wrapper, and none imports `assertSmokeSandboxDown`:

| file | `down` call sites |
|---|---|
| `implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts` | L131 `["down","--preserve-state"]`, L162 `["down"]`, L168 `["down"]` in `finally` |
| `implementations/cli/smoke/up-restore-reentry-live.smoke.ts` | L130, L162, L168 — same pattern |
| `implementations/cli/smoke/up-resume-render-lock-live.smoke.ts` | L186 `["down","--preserve-state"]` |

All three were added 2026-08-29; #884 closed 2026-08-30. They predate the guard and were never converted to it.

## They are sandboxed, but not on the axis that matters for the seed store

They are not naive. Each builds `makeScratch` + `assertScratchHeld`, runs with `cwd: root` under that scratch, creates `root/.cotal` so the root walk terminates, points `COTAL_HOME` at a scratch home, and strips inherited `COTAL_*`.

What none of them sets is **`XDG_CONFIG_HOME`** — measured: `COTAL_HOME` appears 3× in each file, `XDG_CONFIG_HOME` 0×.

That is the gap, and the product states it plainly in its own refusal text:

> Isolate with `XDG_CONFIG_HOME`. `COTAL_HOME` does not relocate this store.

The operator-global seed store follows `XDG_CONFIG_HOME`, not `COTAL_HOME`. `assertSmokeSandboxDown` refuses a missing `XDG_CONFIG_HOME`; these three would not pass it. Related: #612, #559.

## Why the consequence is not theoretical

`findCotalRoot` (`packages/workspace/src/auth-paths.ts` L268–275) walks up for a `.cotal` directory and, finding none, **returns the starting directory** rather than throwing. Fail-open. A `down` whose sandbox silently failed to apply therefore resolves to whatever root the walk lands on, stops that mesh, and reports success. The `.cotal` marker these three create is what stands between them and that outcome, and nothing verifies it was created before the verb runs — which is precisely the gap #884 was opened for.

## Scan quality — three of the six reported lines are false positives

Recording these so the next reader does not chase them, and because they argue for improving the scan rather than only the suites:

- `implementations/runtime/smoke/mesh-monitor.smoke.ts` — `wait(down)`, cotal-lang liveness, not a CLI verb. Same class as the already-allowlisted `mesh-wait.smoke.ts`; `mesh-monitor` is simply not in the allowlist.
- `implementations/runtime/smoke/migrate-run.smoke.ts` (file-level, and L225) — L225 is a **template string** passed to `record()`: `` `try { await spawn("dev", …); } catch (e) { log("down", e.code); }` ``. The rule sees `spawn(` on a line that also contains `"down"`.

The same-line regex also *misses* the three real offenders' individual call sites, because `"down"` is not on the `spawnSync` line — they are caught only by the coarser file-level "no shared guard call" cell. So the scan is both over- and under-inclusive while still being right that the tree has offenders.

## What this asks for

1. Get `smoke:sandbox-guard` into a shard (`bin/smoke/ci-suites.d/<sha256(suite-name)>.txt`), so #884's guard is actually enforced. Without this the rest regresses again.
2. Convert the three suites to `assertSmokeSandboxDown`, which also closes the missing `XDG_CONFIG_HOME`.
3. Tighten the scan: allowlist `wait(down)`/cotal-lang uses, ignore `down` inside string literals, and match wrapper calls rather than requiring `spawn` and `"down"` on one line.

Item 1 is the one that matters most. The other two are today's offenders; item 1 is why there were any.

davidfarah2003:
#884 shipped `assertSmokeSandboxDown` and a scan that requires live CLI `down` to prove the sandbox held. At the base of this branch that scan was red, and it ran in no CI job (only from `pnpm check`, which no workflow invokes). This PR does the three things that make that gate real, in one self-proving diff: the suite is green, the scan still names an unguarded call, and CI now runs it.

Closes #1279.

## A. Adopt the shared guard in the three real offenders

`implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts`, `up-restore-reentry-live.smoke.ts`, and `up-resume-render-lock-live.smoke.ts` already set `COTAL_HOME` and created a `.cotal` marker. That is not enough: `findCotalRoot` fails open, so a `down` whose sandbox silently failed still walks into whatever mesh it finds and reports success.

Each wrapper now records the sandbox with `recordSmokeSandbox`, pins `XDG_CONFIG_HOME` in the child env, and calls `assertSmokeSandboxDown` on the exact spawn options immediately before the CLI process starts.

## B. Tighten the scan so false positives stop matching

In `packages/smoke-kit/smoke/sandbox-guard.smoke.ts`:

- allowlist `implementations/runtime/smoke/mesh-monitor.smoke.ts` the same way `mesh-wait` already is (`wait(down)` is cotal-lang liveness, not a CLI verb)
- ignore `down` inside comments and template strings (the `migrate-run.smoke.ts:225` `record(\`… log("down" …)\`)` false positive)
- match local wrapper calls (`runSync(["down"])`, `run(["down"])`, and the same-line `spawn`/`spawnSync` form) so the three real sites are named by a precise cell, not only by the coarse file-level "no shared guard call"

A tightening that merely deletes cells is not this change. After the rule change, a scratch copy of one unguarded shape (`runSync(["down"])` plus a raw `spawnSync(..., ["down"])`) was named as:

- `…: no shared guard call`
- `…:7: down wrapper call is not immediately guarded`
- `…:8: raw down spawn is not immediately guarded`

That scratch file was then deleted and is not in this PR.

## C. Put `smoke:sandbox-guard` into a CI shard

Added exactly one fragment:

`bin/smoke/ci-suites.d/34a43b526277cf311de9858e519c3d55e83454339225776c426c8a6d5cbc7164.txt`

Filename verified with `printf 'smoke:sandbox-guard' | sha256sum`. `ci-suites.mjs` reads `ci-suites.d/` at runtime; the frozen `bin/smoke/ci-suites.txt` is not touched. After adding the fragment, `git status` showed only that file plus the intended source/changeset files.

## Changeset

`.changeset/sandbox-guard-enforced.md` bumps `cotal-ai` (lockstep `fixed` group) at patch. The change is test/CI adoption, not a published API, but the freeze-guard precedent also used a `cotal-ai` patch for a CI-only gate. `pnpm changeset status` was run against the dirty tree; it also listed other already-pending changesets on main.

## Verified (exact commands, this worktree)

- `pnpm smoke:sandbox-guard` — empty offender set, `sandbox guard smoke: PASS`, exit 0
- `pnpm smoke:ci-suites` — `SUITE COMPLETE: 25 passed, 0 failed`
- `pnpm smoke:ci-suites-freeze` — `CI SUITES FREEZE SMOKE OK (9 passed, 0 failed)`
- `pnpm smoke:gate-inventory` — `GATE INVENTORY OK` (569 scripts, 549 reached, 20 not run; no stale UNGATED for this suite)
- `pnpm typecheck` — green at this head
- `git show HEAD --name-only` — the six files above are the commit

## Deliberately unverified

The three `up-*-live` suites were **not** run. They call the real CLI `down`. On this machine that can stop the live fleet. Their change is graded by reading the wrappers against the existing adopted pattern (`recordSmokeSandbox` + `assertSmokeSandboxDown` immediately before spawn) and by CI.

Also not run: any other `*-live` suite, `pnpm check`, `pnpm cotal` / `tsx bin/cotal.ts`.

### DIFF
--- a/.changeset/sandbox-guard-enforced.md
+++ b/.changeset/sandbox-guard-enforced.md
@@ -1,0 +1,5 @@
+---
+"cotal-ai": patch
+---
+
+Run the sandbox-down adoption scan in CI, and require the three remaining live `down` wrappers to prove the sandbox held before the destructive verb.
--- a/bin/smoke/ci-suites.d/34a43b526277cf311de9858e519c3d55e83454339225776c426c8a6d5cbc7164.txt
+++ b/bin/smoke/ci-suites.d/34a43b526277cf311de9858e519c3d55e83454339225776c426c8a6d5cbc7164.txt
@@ -1,0 +1,3 @@
+# Live CLI down must prove the sandbox held before the destructive verb (#884). This scan
+# is the adoption gate; without a CI shard it only ran from `pnpm check`, which no workflow invokes.
+smoke:sandbox-guard
--- a/implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts
+++ b/implementations/cli/smoke/up-adopt-resume-lock-live.smoke.ts
@@ -33,6 +33,7 @@ import { spawnSync } from "node:child_process";
 import { existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync } from "node:fs";
 import { createServer, type AddressInfo } from "node:net";
 import { join, resolve as resolvePath } from "node:path";
+import { assertSmokeSandboxDown, recordSmokeSandbox } from "@cotal-ai/smoke-kit";
 import { makeScratch, assertScratchHeld } from "../../../bin/smoke/_scratch.js";
 
 const freePort = (): Promise<number> =>
@@ -48,6 +49,8 @@ const freePort = (): Promise<number> =>
 const scratch = makeScratch("cotal-up-adopt-lock-");
 const home = mkdtempSync(join(scratch, "home-"));
 const root = mkdtempSync(join(scratch, "root-"));
+const configDir = join(home, "xdg");
+const sandbox = recordSmokeSandbox({ root, cotalHome: home, xdgConfigHome: configDir });
 process.env.COTAL_HOME = home;
 
 const WT = resolvePath(import.meta.dirname, "..", "..", "..");
@@ -76,17 +79,19 @@ const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
 const baseEnv = (() => {
   const copy = { ...process.env };
   for (const key of Object.keys(copy)) if (key.startsWith("COTAL_")) delete copy[key];
-  return { ...copy, COTAL_HOME: home };
+  return { ...copy, COTAL_HOME: home, XDG_CONFIG_HOME: configDir };
 })();
 
 /** Blocking `cotal <args>`, the way an operator runs it. */
 function runSync(args: string[], extraEnv: Record<string, string> = {}) {
-  return spawnSync(TSX, [CLI, ...args], {
+  const options = {
     cwd: root,
     env: { ...baseEnv, ...extraEnv },
-    encoding: "utf8",
+    encoding: "utf8" as const,
     timeout: 240_000,
-  });
+  };
+  assertSmokeSandboxDown(sandbox, args, options);
+  return spawnSync(TSX, [CLI, ...args], options);
 }
 
 type Journal = {
--- a/implementations/cli/smoke/up-restore-reentry-live.smoke.ts
+++ b/implementations/cli/smoke/up-restore-reentry-live.smoke.ts
@@ -33,6 +33,7 @@ import { spawnSync } from "node:child_process";
 import { existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync } from "node:fs";
 import { createServer, type AddressInfo } from "node:net";
 import { join, resolve as resolvePath } from "node:path";
+import { assertSmokeSandboxDown, recordSmokeSandbox } from "@cotal-ai/smoke-kit";
 import { makeScratch, assertScratchHeld } from "../../../bin/smoke/_scratch.js";
 
 const freePort = (): Promise<number> =>
@@ -48,6 +49,8 @@ const freePort = (): Promise<number> =>
 const scratch = makeScratch("cotal-up-restore-reentry-");
 const home = mkdtempSync(join(scratch, "home-"));
 const root = mkdtempSync(join(scratch, "root-"));
+const configDir = join(home, "xdg");
+const sandbox = recordSmokeSandbox({ root, cotalHome: home, xdgConfigHome: configDir });
 process.env.COTAL_HOME = home;
 
 const WT = resolvePath(import.meta.dirname, "..", "..", "..");
@@ -77,17 +80,19 @@ const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
 const baseEnv = (() => {
   const copy = { ...process.env };
   for (const key of Object.keys(copy)) if (key.startsWith("COTAL_")) delete copy[key];
-  return { ...copy, COTAL_HOME: home };
+  return { ...copy, COTAL_HOME: home, XDG_CONFIG_HOME: configDir };
 })();
 
 /** Blocking `cotal <args>`, the way an operator runs it. */
 function runSync(args: string[], extraEnv: Record<string, string> = {}) {
-  return spawnSync(TSX, [CLI, ...args], {
+  const options = {
     cwd: root,
     env: { ...baseEnv, ...extraEnv },
-    encoding: "utf8",
+    encoding: "utf8" as const,
     timeout: 240_000,
-  });
+  };
+  assertSmokeSandboxDown(sandbox, args, options);
+  return spawnSync(TSX, [CLI, ...args], options);
 }
 
 const tail = (r: { stdout?: string; stderr?: string }) => `${r.stdout ?? ""}${r.stderr ?? ""}`;
--- a/implementations/cli/smoke/up-resume-render-lock-live.smoke.ts
+++ b/implementations/cli/smoke/up-resume-render-lock-live.smoke.ts
@@ -40,6 +40,7 @@ import { once } from "node:events";
 import { existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync, statSync } from "node:fs";
 import { createServer, type AddressInfo } from "node:net";
 import { join, resolve as resolvePath } from "node:path";
+import { assertSmokeSandboxDown, recordSmokeSandbox } from "@cotal-ai/smoke-kit";
 import { makeScratch, assertScratchHeld } from "../../../bin/smoke/_scratch.js";
 
 const freePort = (): Promise<number> =>
@@ -55,6 +56,8 @@ const freePort = (): Promise<number> =>
 const scratch = makeScratch("cotal-up-resume-lock-");
 const home = mkdtempSync(join(scratch, "home-"));
 const root = mkdtempSync(join(scratch, "root-"));
+const configDir = join(home, "xdg");
+const sandbox = recordSmokeSandbox({ root, cotalHome: home, xdgConfigHome: configDir });
 process.env.COTAL_HOME = home;
 
 const { acquireMaintenanceLock, authDir, maintenancePaths, readMaintenanceJournal, releaseMaintenanceLock } =
@@ -86,18 +89,20 @@ const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
 const sandboxEnv = (() => {
   const copy = { ...process.env };
   for (const key of Object.keys(copy)) if (key.startsWith("COTAL_")) delete copy[key];
-  return { ...copy, COTAL_HOME: home };
+  return { ...copy, COTAL_HOME: home, XDG_CONFIG_HOME: configDir };
 })();
 
 const output = new WeakMap<ChildProcess, () => string>();
 const logOf = (cp: ChildProcess) => output.get(cp)?.() ?? "";
 
 function run(args: string[]): ChildProcess {
-  const cp = spawn(TSX, [CLI, ...args], {
+  const options = {
     cwd: root,
     env: sandboxEnv,
-    stdio: ["ignore", "pipe", "pipe"],
-  });
+    stdio: ["ignore", "pipe", "pipe"] as ["ignore", "pipe", "pipe"],
+  };
+  assertSmokeSandboxDown(sandbox, args, options);
+  const cp = spawn(TSX, [CLI, ...args], options);
   kids.push(cp);
   let log = "";
   cp.stdout?.on("data", (b: Buffer) => { log += b.toString(); });
--- a/packages/smoke-kit/smoke/sandbox-guard.smoke.ts
+++ b/packages/smoke-kit/smoke/sandbox-guard.smoke.ts
@@ -188,8 +188,191 @@ try {
 
   const semanticDownOnly = new Set([
     "implementations/runtime/smoke/mesh-wait.smoke.ts",
+    "implementations/runtime/smoke/mesh-monitor.smoke.ts",
     "packages/lang/smoke/engine.smoke.ts",
   ]);
+
+  /** Keep line numbers. Drop comments and templates so `down` inside a string is not a CLI verb. */
+  const codeOf = (source: string): string => {
+    let out = "";
+    let i = 0;
+    const blank = (from: number, to: number): void => {
+      for (let j = from; j < to; j++) out += source[j] === "\n" ? "\n" : " ";
+    };
+    while (i < source.length) {
+      const c = source[i];
+      const n = source[i + 1];
+      if (c === "/" && n === "/") {
+        const start = i;
+        i += 2;
+        while (i < source.length && source[i] !== "\n") i++;
+        blank(start, i);
+        continue;
+      }
+      if (c === "/" && n === "*") {
+        const start = i;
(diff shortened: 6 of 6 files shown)


davidfarah2003:
Independent control on the scan tightening, offered as **evidence for the reviewer, not as a verdict**. I am not the review gate on this PR and a seat will grade it at its exact head.

The risk with "tighten a scan so three false positives stop matching" is that it gets tightened into silence — a scan that is green because it stopped looking is indistinguishable from one that is green because the tree is clean. So rather than read the diff, I planted three synthetic files under `implementations/cli/smoke/` at `acacca4fd3`, ran the shipped suite, and deleted them.

**Probe 1 — direct unguarded spawn** (must be caught):

```ts
const CLI = "bin/cotal.ts";
spawnSync("tsx", [CLI, "down", "--preserve-state"], { cwd: "/tmp/x" });
```

**Probe 2 — unguarded call through a local wrapper** (must be caught; this is the shape the previous scan *missed*, because `"down"` is not on the `spawnSync` line):

```ts
function cli(args: string[]) { return spawnSync("tsx", [CLI, ...args], { cwd: "/tmp/x" }); }
cli(["down"]);
```

**Probe 3 — negative control**, `"down"` inside a string literal with no CLI spawn (must **not** be caught; this is the `migrate-run:225` false positive):

```ts
record(`try { await spawn("dev", 1); } catch (e) { log("down", e); }`);
```

Result:

```
implementations/cli/smoke/zz-probe-direct-live.smoke.ts: no shared guard call
implementations/cli/smoke/zz-probe-direct-live.smoke.ts:3: raw down spawn is not immediately guarded
implementations/cli/smoke/zz-probe-wrapper-live.smoke.ts: no shared guard call
implementations/cli/smoke/zz-probe-wrapper-live.smoke.ts:4: down wrapper call is not immediately guarded
```

Three findings worth stating precisely:

1. Both offenders are caught, and probe 3 produced nothing. The tightening is strictly more precise in **both** directions rather than merely quieter.
2. Probe 2 is named at its **call site** (`:4`), not only at file level. That is a real gain: the previous scan caught the three actual offenders only through the coarse "no shared guard call" cell and could not point at the line.
3. With the probes removed, the suite returns an empty offender set and exits 0 at this head — I ran it in my own clean worktree rather than taking the number from the branch.

What this control does **not** establish, so nobody reads it as more than it is: it says nothing about whether the three converted suites are correct (they are `-live`, they call the real CLI `down`, and neither the author nor I ran them — deliberately, and the PR says so). It says nothing about the `assertSmokeSandboxDown` cells added to the suite itself. And a scan that catches two synthesized shapes is not a scan proven complete; the enumeration of shapes it could miss is still open.

Structural checks that also hold at `acacca4fd3`: `bin/smoke/ci-suites.txt` is untouched (0 files), and the fragment filename is exactly `sha256("smoke:sandbox-guard")`.

davidfarah2003:
Shard 0/4 red at `acacca4fd3`, attributed away from this diff before anyone spends time on it.

```
✗ shard 0/4 FAILED at: pnpm smoke:manager-stop-reap (exit 1)
✗ scenario threw: re-registration could not revoke + verify-evict the superseded serve family;
  the gate is left frozen for reconciliation, no new spec published (SPEC 13.1):
  the delivery daemon is not reachable on the ctl.delivery-admin rail (timeout)
MANAGER-STOP-REAPS-AGENTS SMOKE FAILED ❌  (9 passed, 1 failed)
NEVER RAN — 22 of 131 planned smoke(s)
```

Four measurements, in the order that matters:

1. **The one suite this PR adds is not in that shard.** Enumerating all four shards at this head: `smoke:sandbox-guard` is in **shard 2/4**; shards 0, 1 and 3 have zero occurrences. Fragment sharding keys on a stable hash of the suite name rather than file order, so adding one cannot move another suite between runners — and here it did not even land in the affected shard, so shard 0's composition is untouched.
2. **That suite passed on this branch's exact base.** `main@9f033b128`, shard 0/4: `MANAGER-STOP-REAPS-AGENTS SMOKE OK ✅ (15 passed, 0 failed)`.
3. **The diff has no reach into that path.** A changeset, a `ci-suites.d` fragment, three `up-*-live` suites, and the sandbox-guard scan.
4. **The proximate cause is a timeout**, on the `ctl.delivery-admin` rail, which is a load signature rather than a logic one.

Rerunning the failed jobs once the run reaches a terminal state — `gh run rerun --failed` will not act on a run that still has jobs in flight, and shards 1 and 2 are still going.

## One adjacency, recorded as a lead rather than a diagnosis

#1266 merged to `main` at 08:30Z, after this branch was cut, so this head does not contain it. It is `fix(core): complete a committed registration spec after a lost ack`, closing #1243:

> after boot self-heal reopens the gate, the re-registration spec-write settles ambiguous and re-freezes it, so takeover never completes

and the failure above reads

> re-registration could not revoke + verify-evict the superseded serve family; the gate is left frozen for reconciliation

Same area, same vocabulary, and the fix is absent from this base. That is worth writing down. It is **not** a diagnosis: the proximate cause here is a rail timeout, not an ambiguous settle, and adjacency of wording is not evidence of mechanism.

The discriminating experiment is ordered, and the cheap one comes first. Rerun at this base: if green, there is nothing to explain and merging main would only have cost a cycle and a docs-bundle conflict for nothing. If it reds in the same place a second time, merging main becomes the obvious next step, and at that point the presence or absence of #1266 is a real variable rather than a coincidence of phrasing.

davidfarah2003:
Sharpening my own framing above, in the direction of less confidence rather than more.

I wrote "flake, or an intermittent already latent on the base". I then checked whether history supports the flake half, and it does not support it for **this suite**. Surveying `smoke (shard 0/4)` across the last ~18 completed CI runs:

| conclusion | count | suite it failed at |
|---|---|---|
| `failure` | 3 | `smoke:required-arg-seam` (1), `smoke:attach-auth-root` (2) |
| `success` | 1 | — |
| `cancelled` | 6 | unmeasured |

`smoke:manager-stop-reap` does not appear as a shard 0/4 failure anywhere in that sample. So this is the first observed failure of that suite in recent history, which makes the rerun **more** decisive, not less — a second red in the same place would be a real signal rather than noise.

Two things that fall out of the same survey and are worth separating from this PR:

- **Six of ten measured-or-not shard 0/4 rows are `cancelled`.** Cancelled is unmeasured, not green, so the true health of that shard is known for only four of the ten. Any statement about "shard 0 is fine" or "shard 0 is broken" built on that sample is built on four data points.
- **The three real failures are at three different suites**, two of them on one branch. That is a shard carrying several independent intermittents rather than one bad suite, and it means "shard 0/4 red" carries little information on its own — you have to read which suite.

None of this changes the attribution above, which rests on the suite's absence from this diff's reach and on it passing at this exact base, not on it being a known flake. It changes what the rerun means: not a formality, but the experiment.

davidfarah2003:
The deliverable is now demonstrated by CI rather than by a local run, which is the evidence that actually matters here — the whole point of this PR is that a suite starts being executed by something.

`smoke (shard 2/4)` at `acacca4fd3`: **success**, and the suite really executed in it. Reading the job log rather than the check conclusion, because a green job is not evidence a suite ran:

```
line 5123:  ===== pnpm smoke:sandbox-guard =====
            $ tsx packages/smoke-kit/smoke/sandbox-guard.smoke.ts
              ✓ foreign sandbox root is refused by identity
              ✓ missing sandbox ownership marker is refused by identity
              ✓ replaced sandbox ownership marker is refused as foreign
              ✓ target guard resolves a flag before the target
              …
```

So the fragment lands the suite in shard 2/4 and CI runs it green. #884's guard is enforced by something for the first time since it shipped.

Run state at this head, now terminal: `unit`, `live`, shards 1/2/3 all success; **shard 0/4** failed at `smoke:manager-stop-reap`, with `ci-ok` failing downstream of it. That is the failure attributed above — different shard from this PR's addition, passing on this branch's base, no reach from the diff, and a rail timeout as the proximate cause. Failed jobs reran; `run_attempt` is confirmed at 2.

**One thing I checked and it was nothing, recorded so the next reader does not repeat it.** Grepping that shard's log for `NEVER RAN` returns a hit, which in a succeeding job would be alarming — suites not starting while the job reports green. It is a false positive: the match is inside the cell names of `smoke:shard-never-ran`, a suite whose job is to test the never-ran reporting mechanism, asserting things like `a mid-partition break through shard.mjs reports NEVER RAN — 2 of 4`. The control that settled it in one command: a succeeding shard on `main` returns **0** matches for the same grep. There is no real never-ran block in that job.

davidfarah2003:
**APPROVE** — `acacca4fd38f2a754c850f99856d507029d81d83`

Independent exact-head review. Both `git rev-parse HEAD` and the PR headRefOid are the sha above; tree clean before and after.

### What I checked (probed, not just read)

**Scan tightening — planted probes, required the scan to name file:line, then deleted them.**
`pnpm smoke:sandbox-guard` (tsx-only, spawns nothing) is green at baseline. Probe results:
- Raw unguarded `spawnSync(TSX, [CLI, "down"], …)` → **named** `…:5: raw down spawn is not immediately guarded`. ✓
- Unguarded wrapper `runSync(["down","--preserve-state"])` reaching spawn (the exact shape these three PRs use) → **named** `…:5: down wrapper call is not immediately guarded`. ✓
- Correctly-guarded wrapper mimicking the PR (`assertSmokeSandboxDown(sandbox,args,options)` immediately before spawn) → **silent PASS**. ✓ (opposite polarity: the scan is not just naming everything.)
- Comment/prose-only `"down"` → **silent** (codeOf blanks it). ✓
- Guard present but >5 lines before spawn inside a wrapper → **flagged**, i.e. the `windowBefore` lookback fails *toward* naming, the safe direction. ✓

**Escape holes I did find (both fail toward "clean"):**
- `const verb = "down"; runSync([verb])` — the literal is not preceded by `[ , (`, so `downArg` misses and nothing else matches → **silent escape**.
- `const args = ["down"]; runSync(args)` — the `["down"]` literal matches but sits at assignment/statement position, so `callNameAt` returns `undefined` and the site is dropped → **silent escape**.

These are latent coverage gaps via argument indirection, not present offenses: at this head every real `cotal down` call site uses the caught literal-in-call form (`cli("down")`, `cliIn(root,"down")`, `run("down",…)`, `cotal(["down",…])`), and the full-tree scan is green. Introducing an indirection form later would bypass the gate silently. Worth a follow-up hardening; not a blocker for what this PR ships.

`semanticDownOnly` — verified all three are genuine non-CLI uses: mesh-wait/mesh-monitor use `{event:"down"}` (presence events), engine uses `ctx.free("down",…)` (a lang register). None is an exempted CLI offender.

**The three converted `-live` wrappers — graded by reading only (I did not run any `-live` suite).**
- Same `options` object is passed to both `assertSmokeSandboxDown(sandbox, args, options)` and the immediately-following `spawnSync`/`spawn`. The guard reads `options.cwd` and `options.env.{COTAL_HOME,XDG_CONFIG_HOME}`; it does not mutate `options`. So the guard validates the same env the child receives. ✓
- `baseEnv`/`sandboxEnv` scrub every `COTAL_*` key then set `{…, COTAL_HOME: home, XDG_CONFIG_HOME: configDir}`; `recordSmokeSandbox({root, cotalHome: home, xdgConfigHome: configDir})` records those exact paths and mkdir's them. Recorded sandbox == spawn env. ✓
- Leaves unverified (cannot run `-live`): that these suites actually reach a live `cotal down` and that the guard permits it against a real listener at runtime.

**Guard fires here, not inert.** `assertSmokeSandboxDown` no-ops unless `args[0] === "down"`. Each suite has real down call sites that reach the wrapper:
- up-adopt-resume-lock-live: `runSync(["down","--preserve-state"])` L136, `runSync(["down"])` L167/L173.
- up-restore-reentry-live: `runSync(["down","--preserve-state"])` L135, `runSync(["down"])` L167/L173.
- up-resume-render-lock-live: `run(["down","--preserve-state"])` L191.

**Changeset.** `"cotal-ai": patch` is a valid member of the 21-package `fixed` group, so one entry versions all in lockstep. Summary line accurately describes what shipped (CI shard + three wrapper adoptions + scan tightening). CHANGELOG-quality only, never a release blocker.

**typecheck.** `pnpm typecheck` at this head: exit 0, `Scope: 27 of 28 workspace projects` (the 28th, `@cotal-ai/example-04-frontier-faces`, is JS-only with no typecheck script — reported as skipped, not counted clean).

### Verdict
The scan catches every shape relevant to its purpose, stays silent on guarded/prose sites, fails conservatively, and the three adoptions are wired correctly (same options object; env carries both keys; real down paths reach the guard). The indirection escapes are latent and absent from the current tree. APPROVE.

Issue #643, opened 2026-08-19
Backup has no CI coverage that can fail: live suites are in no shard, gated smokes validate the inventory against itself


davidfarah2003:
## Summary

Backup has no test that both runs in CI and is capable of failing on a real inventory gap. Two
independent facts, each sufficient on its own:

### 1. Not run

The two live backup suites, `smoke:backup-restore:live` and `smoke:backup-conservation:live`,
exist as `package.json` scripts but appear in no workflow under `.github/workflows/` (verified
with a positive control confirming the search fires). Nothing runs them in CI. Both are red
today: `backup-conservation-live` threw before its first line for about a month (`provisionAgent`
has required a `lifecycleUid` since 2026-07-14 that the suite never passed), and
`backup-restore-live` fails independently at its line 726.

### 2. Self-referential when run

The two gated backup smokes feed `spaceBackupInventory` back into
`validateSpaceBackupInventory`, validating the produced inventory against itself.
`validateSpaceBackupInventory` is exact set-equality, but a set compared to itself always agrees,
so the gated pair passes even when a whole stream family is absent from the inventory.

## Consequence

There is no CI signal for a backup inventory gap. This is the mechanism that let the unenumerated
endpoint-binding (`EP*`) family in #356 sit undetected: the suites that would catch it do not run,
and the suites that run cannot fail for it.

## Fix (same stay-closed shape as #638)

- Break the gated smokes' circularity: validate the inventory against an independent enumeration
  of what the broker actually holds, not against the inventory's own output. An existing live
  suite already enumerates broker truth rather than a synthesized list; that is the shape. This
  part is independent of #356.
- Bring the two live suites into a CI shard, after their reds are repaired. A suite in `pnpm
  check` but no shard is coverage on paper only. Note the sequencing: `backup-conservation-live`
  cannot go green until the inventory correctness defect in #356 is fixed, and
  `backup-restore-live` needs its own independent repair first.
- No baseline-tolerated reds: the suites must be green before they are required in a shard, and a
  planted omission (a stream removed from the inventory) must redden the check, proven by
  mutation.

## Related

- #356: the correctness defect this circularity masks (streams present in the space but absent
  from the inventory; the `EP*` and `KV_cotal_auth_*` / `KV_cotal_records_*` families).
- #638: the sibling CI-integrity gap (smoke files never typechecked); same false-coverage class
  on a different axis.

davidfarah2003:
[...]

PR #1283, opened 2026-09-05
fix(ci): count only CI-reached smoke suites as gated

### LINKED ISSUE #1278: 16 non-live smoke suites are executed by no CI job
## Summary

16 non-`:live` smoke suites exist as `package.json` scripts that **no CI job executes**, directly or transitively. Nothing in the repo distinguishes "deliberately excluded" from "forgotten", so a suite in this set can sit red indefinitely without any signal.

This is the un-swept remainder of a class the repo already knows about. `.github/workflows/ci.yml` carries this comment above the `Docs smoke` step:

```yaml
# Broker-free and network-free, and it reads the BUILT connector, so it belongs after Build.
# It was reachable from `check` and executed by nothing, which is how it sat red on main.
- name: Docs smoke
  run: pnpm smoke:docs
```

`smoke:docs` was found and wired up individually. The sweep was never done.

## The set

Reachable only from `check`, which no workflow invokes:

- `smoke:ci`
- `smoke:connect`
- `smoke:sandbox-guard`

Referenced by no other script and no workflow:

- `smoke:agui-map:real`
- `smoke:attention`
- `smoke:attention:auth`
- `smoke:delivery-boot-retry:auth`
- `smoke:delivery-broker-coupling`
- `smoke:delivery-old-manager`
- `smoke:feedback`
- `smoke:lifecycle-files`
- `smoke:manager-console`
- `smoke:pi`
- `smoke:plane3-activation:auth`
- `smoke:plane3-gate:auth`
- `smoke:self-serve-join-coverage:auth`

## Why `check` cannot be the answer

`check` chains 30 stages, **17 of them `:live`**. It starts and stops stacks, so it is unusable as a CI job by construction and unusable on a shared workstation. A suite whose only reference is `check` is therefore run by nobody on any schedule.

## Confirmed live instance

`smoke:sandbox-guard` is currently **red**, and was discovered only because unrelated work happened to invoke it. It guards the `@cotal-ai/smoke-kit` boundary — the rule that no shipped file may import test-only helpers. That guard has been off, silently, for an unknown period.

## Method, so this is reproducible and falsifiable

A suite counts as reachable if it appears in `bin/smoke/ci-suites.txt`, in a `bin/smoke/ci-suites.d/*.txt` fragment, or in a workflow **`run:` line**, plus the transitive closure through `package.json` script bodies.

Two instrument bugs were found and fixed while building this, both worth repeating because either produces a confident wrong answer:

1. **Matching workflow file text rather than `run:` lines.** The word `check` appears in ci.yml only inside comments. A naive grep reports it as invoked and drags all 30 of its stages — including 17 `:live` suites — into "reachable".
2. **A loose token boundary.** `\bcheck\b` matches inside `check:docsbundle`, because `:` is a word boundary. That alone flipped `smoke:sandbox-guard` to "reachable".

The census is control-gated. It asserts known-reachable suites (`smoke:docs`, `smoke:windows`, `smoke:mutation-fixtures`, `smoke:attach` — the last only via `smoke:runtime`) come back reachable, and known-unreachable ones (`smoke:sandbox-guard`, `smoke:connect`) come back unreachable. Both instrument bugs above were caught by those controls failing, not by inspection.

## What this issue asks for

Not "run all 16". Several are plausibly excluded on purpose — `smoke:pi` needs pi present, the `:auth` suites need auth mode, `smoke:agui-map:real` names an external dependency.

The ask is that **exclusion becomes declared rather than incidental**:

- Triage each of the 16 into runs-in-CI or excluded-with-a-stated-reason.
- Fix `smoke:sandbox-guard` and get it into a shard, since it guards a real boundary rule and is red now.
- Add a gate that fails when a `smoke:*` script is reachable from no workflow and no corpus entry and carries no exclusion marker, so the next one cannot arrive silently.

Without the last item this recurs: `smoke:docs` was fixed individually and 16 more were already in the same state.

davidfarah2003:
`check` is a developer convenience chain, not a workflow. Treating it as a CI root made suites look gated when nothing in GitHub Actions ran them. This PR counts only CI-reached suites as gated, homes the ones that belong in CI, deletes a duplicate, and declares the rest.

Closes #1278.

## Instrument

`gate-inventory` ROOTS are now `smoke:ci` and `test`. Workflow `run:` lines still add roots. `check` is not one of them.

After that change the unexplained set was 8:

- `smoke:backup-conservation:live`
- `smoke:backup-faults:live`
- `smoke:backup-perms:live`
- `smoke:backup-restore:live`
- `smoke:backup-usermode:live`
- `smoke:connect`
- `smoke:down-manifest-usermode:live`
- `smoke:join-external:live`

Sanity: known gated suites (`sandbox-guard`, `docs`, `gate-inventory`, …) stayed reached; known ungated suites (`orca:live`, `attention`, `pi`, …) stayed unreached.

`bin/smoke/ci-suites.txt` is untouched.

### Gated (first execution is CI)

`smoke:join-external:live` rides its own live-job step (`Join an external broker`), not a link in the `&&` chain. Observed green in CI job `101291326806` at `2850a5a2e`: 19 checks passed.

### Deleted

- `smoke:connect` / `implementations/cli/smoke/connect.smoke.ts`. Workspace `preflight` already covers `classifyPreflightFailure`. Removed from the `check` chain and from the connect re-export comment.

### UNGATED (declared, not forgotten)

- `smoke:down-manifest-usermode:live` — standing: needs a `claude` CLI on PATH to boot a real connector child.
- `smoke:backup-usermode:live` — `BROKEN:` red; cause unconfirmed (the only local measurement was on a host with a live stack). #1285. `recheckBy` 2026-09-30.
- `smoke:backup-perms:live` — `BROKEN:` red on main, observed in CI job `101291326806` at `2850a5a2e`: `packages/core/smoke/backup-live.smoke.ts:125` (`zero-delivery New consumer preserves its creation frontier`, `1 !== 2`). #643. Already-red so it cannot enter CI.
- `smoke:backup-restore:live`, `smoke:backup-conservation:live`, `smoke:backup-faults:live` — `BROKEN:` never executed; blocked behind failing `backup-perms:live` on the `&&` chain at `2850a5a2e`; **status unknown**. #643.

The four backup live suites are #643's to fix (backup live coverage that can fail), not an inventory mystery. They were briefly gated in this PR and pulled back after that CI observation.

`EXPECTED_EXEMPTIONS` 20 → 26. Existing UNTRIAGED / standing entries are unchanged. Live job timeout stays 25 minutes / 1500 seconds.

## Proof

- `pnpm smoke:gate-inventory` green: 568 scripts, 542 reached, 26 ungated (all exempted; 5 BROKEN).
- `pnpm mutation-proof --config bin/smoke/mutations/gate-inventory-roots.json` KILLED: dropping the dedicated join-external step names `smoke:join-external:live`.
- `pnpm smoke:live-job-conclusion` green; its mutation-proof KILLED both cells.

Merge should wait on #1267 (0.43.0). If main moves, merge main (do not rebase) and re-run inventory.

## Test plan

- [ ] `pnpm smoke:gate-inventory` green
- [ ] `pnpm mutation-proof --config bin/smoke/mutations/gate-inventory-roots.json` KILLED on `smoke:join-external:live`
- [ ] Live job has a dedicated `Join an external broker` step; no backup live suites in the workflow
- [ ] `smoke:connect` is gone from `package.json` and `implementations/cli/smoke/`
- [ ] `ci-suites.txt` unchanged vs base
- [ ] UNGATED backup reasons cite #643 / #1285 and do not invent a diagnosis for unrun suites


davidfarah2003:
**Head moved: `2850a5a2ef5f74b84b7652c8bc22fd55230397b0` → `073d4d79db8ec9548e62bdb4f3053a42e3082ac0`.** Any review measured against the earlier sha binds a tree that is no longer this PR, so it needs re-pointing rather than carrying forward. The delta is small and concentrated in the `smoke:backup-usermode:live` exemption entry; diff the two shas rather than taking that on trust, and state which conclusions were re-checked against the new head and which were carried.

## One caveat on the new exemption reason, and on #1285

The exemption text now says the failure is at section E — restore lands the journal `degraded`, the delivery daemon exited without becoming ready, the shard-0 lease not held — and that section D, which the previous text blamed, is green. That is a real improvement in specificity and it came from actually running the suite rather than from re-reading the old string.

**But the reproduction was taken on a workstation with a live stack running, and that matters for what it establishes.** A `:live` suite builds its own root, and a root created under a working directory that sits inside an existing mesh tree resolves that outer mesh rather than an isolated one. "The delivery daemon exited without becoming ready, and the shard-0 lease was not held" is precisely the shape that contention with an already-running daemon produces.

So the failure is **consistent with** a defect in the suite and equally consistent with a collision at the point of measurement, and the run as taken cannot distinguish them. That does not weaken the case for the exemption — a suite that cannot enter CI red is unrunnable either way — but it does mean the exemption's stated *cause*, and #1285's diagnosis, rest on evidence that has not been isolated.

What would settle it: the same run on a host with no live stack, or in CI where the runner is clean. Until then #1285 is worth keeping as an observation with its provenance recorded, not as a diagnosis. The distinction is the difference between "this suite is broken at section E" and "this suite is broken at section E when something else holds the lease", and only one of those is a bug in the suite.

Recording this here rather than by side channel so the caveat travels with the issue instead of living in one reader's head.

davidfarah2003:
Recording the blocking finding against `797fb504599295b8828760d824fbb09e66b5a690`, and what the head that replaced it does about it. Publishing it here because it was established and would otherwise not be written down anywhere durable.

## The finding

**`smoke:backup-perms:live` is red on main, and `797fb5045` gated it into CI.** That would have made the `live` job permanently red on merge.

Evidence, from CI job `101291326806` in run `33960455711`:

```
join-external-broker            19 checks passed
backup-perms:live               AssertionError at packages/core/smoke/backup-live.smoke.ts:125
                                "zero-delivery New consumer preserves its creation frontier", 1 !== 2
backup-restore:live             never started
backup-conservation:live        never started
backup-faults:live              never started
```

That run was at `2850a5a2e`, two heads earlier. It binds the later heads because the relevant content is byte-identical across them:

```
packages/core/smoke/backup-live.smoke.ts   sha256 7f0a37b3b29a5be2
    at 2850a5a2e, 073d4d79d, 797fb5045 — and at origin/main
.github/workflows/ci.yml                   sha256 35ad5e42f011e5b6
    at 2850a5a2e, 073d4d79d, 797fb5045
```

The two intervening commits changed only an exemption reason string. Hashing the tree is what makes old-run evidence transfer here; ancestry would not, because this branch was force-pushed and `--is-ancestor` answers a different question on a rewritten branch.

**The `main` hash match is the load-bearing one.** It establishes the suite is red on main rather than broken by this PR — the same class the PR already exempts `smoke:backup-usermode:live` for. The principle was accepted on the branch and applied inconsistently.

## The structural half, which matters more than the specific suite

Three of the five newly-gated suites **never executed at all**, short-circuited by the `&&` chain after the first failure. So the claim that they are gated was not merely wrong about one suite; it was **unevidenced for three more**. A job that concludes and a suite that ran are different facts, and an `&&` chain of never-executed suites makes the first failure hide the rest.

`smoke:join-external:live` genuinely passed, with 19 checks. That is the part of the change CI validates and it should not be lost in the rework.

## What the current head does

`4253e3729` gates only `join-external:live`, in **its own step** rather than in the chain, and moves the four backup suites to `BROKEN:` exemptions pinned to **#643**, which already owns this class — *"Backup has no CI coverage that can fail: live suites are in no shard, gated smoke..."*. The live-job timeout returns to 25 minutes now that the backup work is out of it, and `EXPECTED_EXEMPTIONS` moves 22 → 26 so the list cannot grow silently.

The exemption reasons distinguish what is measured from what is not: `backup-perms:live` cites the CI cell above, while the other three say **status unknown** because nothing has ever executed them. That distinction is the right one. Recording an unmeasured cause as a fact is how an exemption stops being a record and becomes a guess.

**So this finding is addressed at the current head.** A BLOCK against `797fb5045` should not be read as standing against `4253e3729`; that head needs its own review, which is being arranged.

davidfarah2003:
## Independent review — BLOCK

**Reviewed head:** `2850a5a2ef5f74b84b7652c8bc22fd55230397b0` (the exact sha this review was seated at).

> **Scope note — the head moved three times during this review.** As I finished, `2850a5a` was
> superseded by `073d4d7` → `797fb50` → `4253e37`. The verdict below is for `2850a5a`, the only sha I
> reviewed end to end. The current head `4253e37` (`fix(ci): gate only join-external live; leave backup
> live ungated`) appears to directly address the blocking finding and needs its own review; see the tail
> of this comment.

### Verdict: BLOCK (at `2850a5a`)

The instrument fixes are correct and well-built, but **the PR's own live CI run is red**, on a suite this
PR gates into CI for the first time. That is the exact class of defect this PR exists to fix, shipped in a
new form.

### Q2 — the five `:live` suites gated for the first time (dispositive)

Measured from the live job of two CI runs at this gating (`2850a5a` run `33960455711`, and the identical
gating re-measured at `797fb50` run `33961171750`; `ci.yml`/`package.json` are byte-identical between
them). The five suites run in a single `&&` chain, in order:

- `smoke:join-external:live` — **RAN, PASSED** (`join-external-broker: 19 checks passed`).
- `smoke:backup-perms:live` — **RAN, FAILED**, deterministically:
  ```
  AssertionError [ERR_ASSERTION]: zero-delivery New consumer preserves its creation frontier
  1 !== 2
      at packages/core/smoke/backup-live.smoke.ts:125
  ```
  (`smoke:backup-perms:live` → `pnpm -F @cotal-ai/core smoke:backup:live` → `backup-live.smoke.ts`.)
- `smoke:backup-restore:live` — **DID NOT RUN** (no banner; `&&` short-circuited on the failure above).
- `smoke:backup-conservation:live` — **DID NOT RUN**.
- `smoke:backup-faults:live` — **DID NOT RUN**.

`packages/core/smoke/backup-live.smoke.ts` is **unchanged by this PR** (last touched 2026-08-19; the
assertion dates to 2026-07-16), and `origin/main`'s live step never invoked these suites. So
`backup-perms:live` is **red-on-main being gated into CI** — the same situation the author correctly
exempts for `backup-usermode:live`. Merging turns the `live` job permanently red. The `live` job is a real
`failure` (not `cancelled`); the failure is a strict-equality assertion, not a timeout — it reproduces on
both runs.

**Timeout headroom:** the job failed ~6–8 min into a 40-min budget, so headroom on the measured path is
large — but the job never completes, so the 25→40 min / 1500→2400 s increase cannot be validated as
sufficient for a *green* run; nothing measured a full live pass.

### Q1 — the deleted `connect.smoke.ts` (71 lines): coverage confirmed elsewhere, cell by cell

`classifyPreflightFailure` is defined in `@cotal-ai/workspace` (`preflight.ts`) and re-exported unchanged by
`implementations/cli/src/lib/connect.ts` — so both suites exercise the same symbol. Every cell of the
deleted suite has an exact successor in the surviving, gated `packages/workspace/smoke/preflight.smoke.ts`
(`smoke:preflight`):

| deleted cell | successor |
| --- | --- |
| `unreachable + {registry,current,flag-space,local-recorded} → prune` | preflight.smoke.ts L60-64 (identical 4 sources) |
| `unreachable + {flag-server,local-space} → no prune` | L65-69 (superset: adds `flag-space-override`) |
| `auth-required + registry + has-auth → registry-creds-rejected` | L70-73 (identical) |
| `auth-required + registry + open → registry-open-now-auth` | L74-77 (identical) |
| `auth-required + local + has-auth → creds-rejected` | L78-81 (identical) |
| `auth-required + local + open → open-wants-auth` | L82-85 (identical) |
| exhaustive "non-registry never prunes" over {unreachable,auth-required} | L131-138 (superset: adds `stale-auth`) |

No orphaned cell; the survivor is a strict superset. The `connect.ts` edit is comment-only. **Q1: OK.**

### Q3 — the two ungated suites, one known-red: machinery is real, not decorative (probed, not read)

- Expiry **is** enforced: I copied the gate into the repo, forced `today` to `2026-12-01`, ran it → all 22
  exemptions expired and the gate went red (`✗ FAIL: 22 UNGATED exemption(s) are past their recheckBy
  date`, exit 1). The built-in `expiryControl` (`:229`) also fails the suite if a past-due control date is
  not caught. So a stale exemption cannot read as no-data.
- The `BROKEN:` class is counted and printed with the suite name and reason at `:370-372`; at head it
  surfaces `smoke:backup-usermode:live` as the 1 BROKEN entry.
- Exempting an already-red suite (`backup-usermode:live`, short-dated `2026-09-30`, `#1285`) is acceptable
  under this machinery: it is counted, dated, and cannot enter CI without reddening it. **Q3: OK.**

### Q4 — the three edited mutation fixtures are corrections, not softenings

- `live-job-conclusion.json` / `.smoke.ts`: cell name "25-minute"→"40-minute" is cosmetic; the assertion
  still pins the **exact** new value (`--timeout-seconds 2400`) and still requires the classifier call. Not
  weaker. (Aside: embedding the budget number in the cell name does hardcode `2400`; that fails loudly on a
  future change rather than silently, which is the safe half — noted, not a blocker.)
- `sandbox-guard.json` M6 anchor re-pinned (`…smoke:connect`→`…smoke:preflight`) because `smoke:connect`
  left the chain; the mutation still reorders `sandbox-guard` behind the first live suite and expects the
  same red. Not weaker.
- All four fixture `find` anchors resolve **exactly once** at head.
- I also probed the PR's premise: restoring `"check"` to `ROOTS` flips 2 suites to "reached" (548/20 vs
  546/22) and reddens the gate — confirming `check` was falsely covering suites. **Q4: OK.**

### Standards

- `pnpm typecheck` — clean, `Scope: 27 of 28 workspace projects` (the 28th,
  `@cotal-ai/example-04-frontier-faces`, is JS-only with no `typecheck` script — reported as a skip, not
  counted as clean).
- `pnpm smoke:gate-inventory` — passes at head (568 scripts, 546 reached, 22 exempted, 1 BROKEN).

### What I could not check / did not run

Per the review constraints I did not execute any `:live`/`-live` suite, `cotal up`, `pnpm check`, the
dev-tree CLI, or `mutation-proof`; the five `:live` suites were graded from CI logs and by reading. The
`ci-ok` live-outcome classifier had not run at the time of writing.

### Current head

`4253e372…` (three commits past this review) is titled *gate only join-external live; leave backup live
ungated* and moves `backup-perms/restore/conservation/faults:live` into `UNGATED` as `BROKEN` (citing this
exact failure), reverting the timeout bump. That would resolve the Q2 blocker. It is a materially different
change and should get its own review at that head with a fresh live-CI confirmation. This BLOCK stands for
`2850a5a2ef5f74b84b7652c8bc22fd55230397b0` only.

davidfarah2003:
## APPROVE — head `4253e372953dc49f7dd22422082183faf0a0fd6f`

Reviewed the reworked head. Delta graded is `797fb5045 → 4253e3729` (7 files); whole head read. Graded from CI and by reading only — no `:live`/`-live` suite, `cotal up`, `pnpm check`, `mutation-proof`, or dev-tree CLI was run here.

### Q1 — did the rework remove the blocking suite, and only it? YES
`.github/workflows/ci.yml` at this head: the four backup suites (`backup-perms/restore/conservation/faults:live`) appear **nowhere** in the workflow. `join-external:live` appears **exactly once**, in its own step `- name: Join an external broker` / `run: pnpm smoke:join-external:live`, no longer on the `&&` chain (the core-boundary step now ends at `pnpm smoke:dogfood:live`). The class of bug (suites reported gated while short-circuited off a chain) is gone, not relocated.

Confirmed in live CI **at this exact head** (run 33961646750, live job 101294484836, success): the dedicated step ran `tsx implementations/cli/smoke/join-external-broker-live.smoke.ts` and logged `join-external-broker: 19 checks passed`. This is the suite running, read from its own step's log, not inferred from the job conclusion. No backup suite appears anywhere in that log.

### Q2 — do the exemption reasons claim only what is measured? YES
`backup-perms:live` asserts red **with** its measured cell: `red on CI at 2850a5a2e (backup-live.smoke.ts:125 … 1 !== 2); #643`. The other three each say `never executed; blocked behind failing backup-perms:live on the && chain at 2850a5a2e; status unknown; #643` — they claim *unknown*, not a fabricated cause. That is the corrected form, and it stuck. `BROKEN:` is retained only as the counting bucket (an unrunnable suite cannot enter CI regardless).

`EXPECTED_EXEMPTIONS` moved 22 → 26. Counted the `UNGATED` map: **26 unique keys, no duplicates**, and the suite asserts `Object.keys(entries).length === EXPECTED_EXEMPTIONS`. `pnpm smoke:gate-inventory` passes here: `26 of 26` examined, `GATE INVENTORY OK`. The four new entries use `untriagedExemption` (`recheckBy 2026-09-30`, future vs today 2026-09-05), so they won't spuriously expire; the expiry control still reddens on its one past-due entry.

### Q3 — the 40→25 min / 2400→1500 s budget: measured, with headroom. CORRECT
Measured the `live` job at this head and recent ones:
- this head (33961646750): 10:53:07 → 11:03:25 = **618 s**
- merge-base/main (33959004840): 485 s; others 484 s, 527 s, 602 s

All within/near the fixture's stated 430–587 s band (this head slightly higher at ~618 s, still fine). 1500 s leaves **~880 s headroom (~2.4×)** over the observed worst case — not marginal. The 1514–1517 s self-timeouts the fixture cites were the heavy backup suites, now removed, so this lowers the budget to fit reality rather than raising it. The suite cell and its mutation fixture both follow the number (`--timeout-seconds 1500`, "25-minute budget"), matching `timeout-minutes: 25` in ci.yml. A cell that duplicates the value would break (not silently pass) if the value changed — both moved together.

### Q4 — mutation fixtures (they moved again). ALL ANCHORS RESOLVE ONCE
Counted every `find` at `<head>:<path>` — each resolves **exactly once**:
- `gate-inventory-roots.json`: the dedicated-step anchor (`- name: Join an external broker\n … \n`) → 1
- `gate-inventory-exemption-review.json`: `date < today` line → 1; `Object.keys(entries).length` line → 1
- `live-job-conclusion.json`: `if (durationSeconds >= timeoutSeconds) … self-timeout` → 1; `node scripts/live-job-conclusion.mjs\n` → 1

The roots anchor is now the step (name+run lines), **code only, no comment** — the PR added a comment directly above the step, and the anchor starts at `- name:` so it does not span it. Both graded suites are run-all (not fail-fast): every check/branch runs and the named cell prints, so there is no earliest-red ambiguity. Verified the named cells are the point each mutant actually reddens: dropping the join-external step makes it `unexplained` → the inventory prints `      smoke:join-external:live` (the exact `expectRed`); the classifier mutants flip the exact named cells.

### Carried forward from the earlier review (re-checked here, cheaply)
- **Deleted `implementations/cli/smoke/connect.smoke.ts` is covered.** It tested `classifyPreflightFailure` (9 invocations). `packages/workspace/smoke/preflight.smoke.ts` tests the same symbol with **11** invocations — same decision tree, superset of cases. Deletion is safe.
- **`implementations/cli/src/lib/connect.ts` is comment-only** — drops the `connect.smoke.ts` reference from the re-export comment; no code change.
- Exemption expiry machinery is real (live `reviewExemptions`/`expiryControl` with a fake clock) and unchanged by this delta.

### What I could not fully check
- **`pnpm typecheck` does not pass at this head — but the cause is pre-existing and out of scope.** `typecheck` is `pnpm build && pnpm -r typecheck`; the build dies in `@cotal-ai/manager` and `@cotal-ai/cli` on a **core API skew** (`completeFrozenRegistrationFromSpec` not exported by core; `MeshLaunchAgent.cwd` / `Runtime.adopt` missing). These symbols are absent from core at the merge-base, at `origin/main`, and at this head alike, and this PR touches **no** manager/core file (full PR scope is CLI comment/delete + `bin/smoke` + changeset + package.json). So `main` itself does not build; this PR neither caused nor can fix it. The PR's own changed TS (`bin/smoke/*.ts`) typechecks clean directly. Reporting the skip rather than calling it clean, and flagging separately that the repo build is red on main and wants its own fix.
- The workspace's 28th project (`@cotal-ai/example-04-frontier-faces`, JS-only) is out of scope as expected.

### Verdict
**APPROVE.** All four graded concerns are correct at this head and the class of bug is fixed and proven to run in live CI. The only blemish is the pre-existing broken `main` build surfaced by `pnpm typecheck`, which is unrelated to this PR's scope.

Issue #400, opened 2026-08-14
codex-host smoke: torn read of a concurrently-appended JSONL crashes the suite


davidfarah2003:
## Summary

`extensions/connector-codex/smoke/codex-host.smoke.ts` reads a JSONL file that a spawned child process is concurrently appending to, and parses every non-empty line. When a poll lands between the child's write of a record's bytes and its terminating newline, the last line is a partial record and `JSON.parse` throws.

Observed:

```
SyntaxError: Unterminated string in JSON at position 509
    at JSON.parse (<anonymous>)
    at tcEntries (extensions/connector-codex/smoke/codex-host.smoke.ts:601)
```

## Mechanism

```ts
const tcEntries = (): LogEntry[] =>
  existsSync(tcLog)
    ? readFileSync(tcLog, "utf8").split("\n").filter(Boolean).map((l) => JSON.parse(l) as LogEntry)
    : [];
```

`tcEntries()` is called from several `waitFor(...)` polling loops, so it re-reads the file every poll interval while the child is still writing to it.

`filter(Boolean)` removes the empty string after a trailing newline, which is why this usually works: a *complete* file ends with `\n` and splits into records plus one empty element. But a file caught mid-append ends with a partial record, which is **not** empty, survives the filter, and reaches `JSON.parse`.

An append is not atomic from the reader's point of view — there is no guarantee that a record and its newline become visible together.

## Why it is intermittent

The window is the gap between the child writing a record's bytes and writing its newline, so hitting it requires a poll to land inside that window. It is correspondingly more likely when the machine is loaded and processes are being descheduled mid-write. It was observed on a run where machine load was 6–8.

That makes it a timing-sensitive failure that will appear as an unrelated crash in whatever suite happens to be running, rather than as a diagnosable assertion failure.

## Suggested fix

Treat the trailing line as possibly incomplete: parse only up to the last newline, and ignore anything after it.

```ts
const tcEntries = (): LogEntry[] => {
  if (!existsSync(tcLog)) return [];
  const raw = readFileSync(tcLog, "utf8");
  const end = raw.lastIndexOf("\n");           // ignore a partially-written trailing record
  if (end < 0) return [];
  return raw.slice(0, end).split("\n").filter(Boolean).map((l) => JSON.parse(l) as LogEntry);
};
```

That keeps every completed record and drops only the one the writer has not finished, which is the record the poll would have picked up on its next pass anyway.

[...]

PR #1284, opened 2026-09-05
fix(connector-codex): wait for the tui event the codex-host cells count

### LINKED ISSUE #1280: codex-host smoke: two cells count a log event with no wait for it, so a scheduling accident reds a shard
## Summary

`smoke:codex-host` failed on `main` at `9f033b128` on a cell that counts a log event **without ever waiting for that event to be written**. The count came back `0` where the cell requires `1`:

```
AssertionError [ERR_ASSERTION]: exactly one launch tail finished: one TUI, not one per generation — 0
    at check (extensions/connector-codex/smoke/codex-host.smoke.ts:60:10)
    at <anonymous> (extensions/connector-codex/smoke/codex-host.smoke.ts:780:3)
```

Every cell before it passed. The shard then stopped, taking 88 of 123 planned suites in that partition with it (`smoke:jcode-host`, `smoke:web-parts`, and 86 more never ran).

## Not caused by anything that landed

Ruled out before writing this, because a red on `main` reads as the last merge's defect by default:

- The merge at `9f033b128` changes 13 files, **0** of them under `extensions/connector-codex`.
- A strict superset tree — a branch head that has `9f033b128` as an ancestor — ran the same suite in the same shard and **passed** it.
- The previous `main` (`74e009c2`) was green on that shard.

So this is a pre-existing race in the suite that happened to land here.

## The defect

`extensions/connector-codex/smoke/codex-host.smoke.ts:780-784`:

```ts
  const genDriven = await waitFor(
    "the recovered peer drives a turn",
    () => genEntries().find((e) => e.ev === "recv" && e.method === "turn/start" && /* …gen-check… */),
    25_000,
  );
  operator.off("presence", genWatch);
  check("a crash mid-launch is recovered by the restart, and the peer is fully online", genDriven !== undefined);
  check(
    "exactly one launch tail finished: one TUI, not one per generation",
    genEntries().filter((e) => e.ev === "tui").length === 1,
    genEntries().filter((e) => e.ev === "tui").length,
  );
```

The `waitFor` above establishes that a `recv`/`turn/start` record exists. The cell below counts `tui` records. Nothing orders those two against each other: the launch tails append their `tui` record on their own schedule, and the suite reads the file the instant the unrelated wait returns.

Two mechanisms both produce `0` here and I cannot tell from the log which one fired:

1. The `tui` record had simply not been appended yet.
2. The read landed mid-append and `readJsonLines` dropped the partial trailing line (see below).

They have the same fix, so distinguishing them is not required to repair it — but the cell as written cannot distinguish a genuine "more than one tail finished" defect from either of them, which is the actual problem.

## A second instance of the same shape

`codex-host.smoke.ts:645-647`:

```ts
  check(
    "a replacement TUI is attached to the new listener",
    tcEntries().filter((e) => e.ev === "tui").length >= 2,
    tcEntries().filter((e) => e.ev === "tui").length,
  );
```

The `waitFor` preceding it waits on the `argv` count, not on `tui`. Same file, same log, same unsynchronized count. It passed on this run; it is one scheduling accident away from the failure above.

## Interaction with #400, which is fixed in the tree but still open

#400 reported that this suite's JSONL reader crashed on a torn trailing line, and proposed parsing only up to the last newline. That fix is present at `9f033b128`, as the shared `readJsonLines` helper:

```ts
function readJsonLines<T>(path: string): T[] {
  if (!existsSync(path)) return [];
  const raw = readFileSync(path, "utf8");
  const lines = raw.split("\n");
  if (!raw.endsWith("\n")) lines.pop();
  return lines.filter(Boolean).map((line) => JSON.parse(line) as T);
}
```

**#400 looks dischargeable and could be closed** on that basis.

Worth recording alongside it: the fix is correct and it changed the failure's shape. A mid-append read used to crash with an unmistakable `SyntaxError`. It now silently returns one record fewer. That is the right behaviour for a poller inside a `waitFor` — it just polls again — and the wrong thing to read from if you are counting, because an undercount is indistinguishable from a real result. The unsynchronized count assertions above are where that lands.

## Suggested fix

Wait for the event, then assert the count:

```ts
  await waitFor("a launch tail finished", () => (genEntries().some((e) => e.ev === "tui") ? true : undefined), 25_000);
  check(
    "exactly one launch tail finished: one TUI, not one per generation",
    genEntries().filter((e) => e.ev === "tui").length === 1,
    genEntries().filter((e) => e.ev === "tui").length,
  );
```

You cannot wait for an absence, so the "not more than one" half stays a point-in-time read — but waiting for at least one converts the cell from "is the record there yet" into the property it is actually asserting. Same treatment for `:645`.

A repair should also sweep the file for the general shape rather than fixing these two: a `check` whose predicate counts or filters `*Entries()` where the nearest preceding `waitFor` waits on a **different** `ev`.

## Cost of leaving it

This cell sits late in a long suite in a shard partition of 123. A failure here discards 88 unrelated suites, so an ordering accident in one connector's launch teardown reads as a `main` regression and hides whatever else that shard would have reported.

davidfarah2003:
## What

`extensions/connector-codex/smoke/codex-host.smoke.ts` had two cells that counted a `tui` log
line without ever waiting for that line to be written. The launch tail attaches the Codex TUI
by spawning `codex resume --remote`, and the fake codex journals its `tui` line only once that
child starts, asynchronously. A scheduling accident could let each cell read the count before
the line landed, reddening a fail-fast shard and stranding the rest of its partition.

This is an ordering defect, not a torn read: `readJsonLines` already drops a trailing partial
line (#400). This change does not touch parsing.

## The two cells, each on its own reasoning

**Cell B, "a replacement TUI is attached to the new listener"** (tuicrash scenario) asserts a
lower bound (`>= 2`). The prior code waited for the app-server to respawn, then slept a fixed
1500ms and counted, so a replacement TUI that attached late was missed. It now waits for the
`tui` count to reach 2, with the timeout as the failure. The `sleep(1500)` is kept for the
adjacent check it actually serves (a mis-classified TUI exit would have shut the agent down by
then).

**Cell A, "exactly one launch tail finished: one TUI, not one per generation"** (genpeer
scenario) asserts an exact count. Waiting until the count reaches 1 and then asserting `=== 1`
would be unsound: it passes the instant the first `tui` lands and would bless the very bug the
cell guards (a superseded tail attaching a second TUI) whenever that second `tui` is a few ms
behind. So the fix waits on the event that BOUNDS the count instead: the superseded launch
tail standing down (logged on the host's stderr the moment it returns without touching
`startTui()`). In this scenario exactly two launch tails exist — the initial one whose child
died before its tools were ready, and the restart that recovered it — and only one may reach
`startTui()`. Once the superseded tail has logged its stand-down, the set of possible `tui`
emitters is closed at one; the cell then waits for the live restart's TUI to attach, settles so
any second child's write would flush, and counts. A second TUI reddens this cell rather than
slipping past it.

## Reproduction

Both cells were confirmed to fail live, not just in reasoning, by delaying the tail's TUI-child
`tui` write past the point each cell read the count (a temporary `FAKE_CODEX_TUI_SLOW` knob in
the fixture, reverted before committing):

- Cell A: with the genpeer `tui` write delayed, `genDriven` resolved first and the count read 0,
  so `=== 1` failed — `AssertionError: exactly one launch tail finished: one TUI, not one per
  generation — 0`.
- Cell B: with the tuicrash replacement `tui` delayed past the 1500ms settle, the count read 0
  and `>= 2` failed — `AssertionError: a replacement TUI is attached to the new listener — 0`.

## Proof the fixed cells can still fail

`bin/smoke/mutations/codex-host-tui-ordering.json`, run with
`node scripts/mutation-proof.mjs --config bin/smoke/mutations/codex-host-tui-ordering.json`.
Both mutations anchor on the same code line — the restart tail's `startTui()` — and both were
KILLED, each as the earliest named failure:

- duplicating the restart's `startTui()` makes the recovered launch attach two TUIs → cell A
  reddens naming `exactly one launch tail finished: one TUI, not one per generation`;
- deleting the restart's `startTui()` suppresses the replacement UI → cell B reddens naming
  `a replacement TUI is attached to the new listener`.

The two mutations are opposite edits of one line, so each is also the other's positive control
that the suite reaches `host.ts` at runtime.

## Verification

- `pnpm smoke:codex-host` green (95 checks) with the fix.
- `pnpm typecheck`: `Scope: 27 of 28 workspace projects`. The 28th,
  `@cotal-ai/example-04-frontier-faces`, is JS-only with no `typecheck` script and is skipped;
  no errors.
- `pnpm changeset status`: my entry is `patch` on `@cotal-ai/connector-codex`; the `fixed` group
  rolls up to minor because of pre-existing changesets on the base, not this one.

## What this leaves unproven

The mutations recreate a second TUI by editing the live restart tail's `startTui()`, which is
the property cell A asserts (a launch that attaches more than one TUI). They do not recreate the
exact historical "one TUI per generation" path in the genpeer scenario, because that path
requires two independent fences to fail at once (a superseded tail adopting a successor's
readiness and then passing the supersede guard); the genpeer child dies before its tools are
ready, so no single-location mutation routes it into a second `startTui()`. The cell's guarantee
is nonetheless deterministic: the superseded stand-down bounds the count before it is read.

Closes #1280

### DIFF
--- a/.changeset/fix-1280-codex-host-tui-ordering.md
+++ b/.changeset/fix-1280-codex-host-tui-ordering.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/connector-codex": patch
+---
+
+Stabilize the codex host smoke: two checks that count the launch tail's TUI log line now wait for that line to be written before counting, so a slow TUI child can no longer red the shard on a scheduling accident. The exact-count check waits for the superseded launch tail to stand down before counting so it still catches a launch that attaches more than one TUI, and the replacement-TUI check waits for the second TUI with the timeout as its failure.
--- a/bin/smoke/mutations/codex-host-tui-ordering.json
+++ b/bin/smoke/mutations/codex-host-tui-ordering.json
@@ -1,0 +1,34 @@
+{
+  "suite": "extensions/connector-codex/smoke/codex-host.smoke.ts",
+  "guard": "the two codex-host cells that count the launch tail's `tui` log line wait for the event that bounds the count before counting, so a scheduling accident cannot read the count before the tail's TUI child has journalled its line (#1280)",
+  "command": "pnpm smoke:codex-host",
+  "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/codex-host-tui-ordering.json",
+  "why": [
+    "#1280: two cells counted a `tui` line without ever waiting for it to be written. Both were confirmed to fail live by delaying the tail's TUI-child `tui` write past the point each cell read the count. The fix makes each cell wait on the event that bounds its count, so the ordering the cell assumed is enforced rather than assumed.",
+    "The suite spawns the REAL host (`../src/host-main.ts`) with `tsx`, which loads `host.ts` from source — no build step participates, so a mutation to `host.ts` is read directly by the run and no `command` build/afterRestore is needed.",
+    "BOTH mutations anchor on the SAME code line — the restart tail's `startTui()` that attaches the replacement UI on the new listener — because that single call is what both cells ultimately observe: cell B (tuicrash, earlier in the file) counts its replacement `tui`, and cell A (genpeer, later) counts the recovered launch's one `tui`.",
+    "CELL A asserts an EXACT count (one TUI, not one per generation). Duplicating the restart's `startTui()` makes the recovered launch attach two TUIs, so two `tui` lines are journalled and the exact count is 2. It reddens FIRST at cell A: the cells above it in the genpeer block (`recovered by the restart`) still pass, and cell A is the earliest assertion the extra TUI reaches. This is the property the cell guards — a launch that attaches more than one TUI — and it is exactly the case a `wait for the count to reach 1, then assert === 1` would have blessed when the second `tui` lagged the first.",
+    "CELL B asserts a LOWER bound (a replacement TUI attaches, >= 2). Deleting the restart's `startTui()` suppresses the replacement UI entirely; the tuicrash scenario's initial TUI still attaches (a different, earlier `startTui()`), so the cell above it (`a TUI dying ... does NOT shut the agent down`) passes and cell B is the earliest red — it times out waiting for the second `tui`, which is exactly the timeout-is-the-failure the fix installs.",
+    "The two mutations are opposite edits of one line (duplicate vs delete), so each also serves as the other's proof that the suite reaches this file at runtime."
+  ],
+  "mutations": [
+    {
+      "name": "the recovered launch attaches TWO TUIs (exact-count cell A must fall)",
+      "file": "extensions/connector-codex/src/host.ts",
+      "find": "      startTui();\n      await safeStatus(\"idle\");",
+      "replace": "      startTui();\n      startTui();\n      await safeStatus(\"idle\");",
+      "expectRed": "exactly one launch tail finished: one TUI, not one per generation",
+      "cell": "exactly one launch tail finished: one TUI, not one per generation",
+      "note": "Two `startTui()` calls on the recovered generation journal two `tui` lines. The fix waits for the superseded tail to stand down and settles before counting, so the second TUI is present when the exact count runs and cell A reddens at 2. A `wait for 1 then assert === 1` would have passed the instant the first landed."
+    },
+    {
+      "name": "the replacement TUI is never attached (lower-bound cell B must fall)",
+      "file": "extensions/connector-codex/src/host.ts",
+      "find": "      startTui();\n      await safeStatus(\"idle\");",
+      "replace": "      await safeStatus(\"idle\");",
+      "expectRed": "a replacement TUI is attached to the new listener",
+      "cell": "a replacement TUI is attached to the new listener",
+      "note": "With no replacement `startTui()` the tuicrash scenario never journals a second `tui`. The fix waits for the count to REACH 2 with the timeout as the failure, so the cell times out naming itself instead of reading a stale count and passing."
+    }
+  ]
+}
--- a/extensions/connector-codex/smoke/codex-host.smoke.ts
+++ b/extensions/connector-codex/smoke/codex-host.smoke.ts
@@ -641,11 +641,15 @@ try {
   );
   await sleep(1500); // a mis-classified TUI exit would have called shutdown() by now
   check("a TUI dying with its crashed app-server does NOT shut the agent down", tcExited === "alive", tcExited);
-  check(
+  // `>= 2` is a lower bound, so waiting for the count to REACH 2 is sound: the replacement TUI is
+  // spawned asynchronously after the respawn, and nothing above waited on its `tui` line landing.
+  // The timeout is the failure — a replacement that never attaches never reaches 2.
+  const tcTuis = await waitFor(
     "a replacement TUI is attached to the new listener",
-    tcEntries().filter((e) => e.ev === "tui").length >= 2,
-    tcEntries().filter((e) => e.ev === "tui").length,
+    () => (tcEntries().filter((e) => e.ev === "tui").length >= 2 ? tcEntries().filter((e) => e.ev === "tui").length : undefined),
+    30_000,
   );
+  check("a replacement TUI is attached to the new listener", tcTuis >= 2, tcTuis);
   tuiCrash.kill("SIGTERM");
   await Promise.race([once(tuiCrash, "exit"), sleep(8_000)]);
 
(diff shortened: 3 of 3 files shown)


davidfarah2003:
**APPROVE** `597fc71df29350a719f8ea1702f7b39701c5ac33`

Independent exact-head review. Local `git rev-parse HEAD` and `gh pr view 1284 --json headRefOid` both returned this sha.

## 1. Cell A (exact count)

The naive repair (wait until the count reaches 1, then `=== 1`) is unsound, and this change does not do that.

From `extensions/connector-codex/src/host.ts` there are two `startTui()` sites: the restart tail (line 983, after `comeOnline` and a `superseded(gen)` return) and the initial launch tail (line 1238, after the same fence). `startTui()` is the only caller of `launchTui()`, and the fake journals `{ev:"tui"}` synchronously at the top of a `resume` process. No third emitter.

In the genpeer scenario the first child dies in `awaitMcpReady` (`FAKE_CODEX_DIE_BEFORE_MCP`, 20ms after `thread/inject_items`). `finalize()` rejects that wait, then the closed handler stamps a new generation inside `driver.start()` before its first await. The initial catch then hits `superseded(gen)` and returns without `startTui()`. `superseded()` logs `app-server incarnation ${gen} was superseded — leaving the rest to the current one`; the catch then logs `initial launch did not finish (...)`. Both alternatives in `/superseded|initial launch did not finish/` are strings the host actually writes. After that return the initial tail cannot reach `startTui()`.

The wait is after `genDriven`, so the live restart has already invoked `startTui()` (it precedes `drive()`). Combined with the stand-down, the set of `startTui()` calls this scenario will make has closed. A superseded tail that *skipped* stand-down and continued into `startTui()` never logs those phrases, so this wait times out red rather than counting 1. That is fail-closed for the historical path, not a false pass.

`superseded` as a substring is broader than the stand-down line (it also appears inside the `awaitMcpReady` error text, which is then interpolated into `initial launch did not finish`). In this scenario the first stderr hit is still the stand-down, which is a microtask after the generation bump and before the restart tail's first await, so it lands on stderr before `startTui()` redirects `log()` into `host.log`. If that order ever inverted, the wait would miss the line and false-red; it would not false-pass.

**Residual `await sleep(1500)`.** This is a timing assumption, as the author says. After stand-down plus `waitFor` of the first `tui`, a second child's journal that took longer than 1500ms would make `=== 1` a false pass. That is the only remaining wrong-direction hole.

It is acceptable here. `launchTui`/`spawn` returns a `ChildProcess` before the child runs, but the fake's `tui` write is the child's first instruction (`appendFileSync`). Any `startTui()` that has already returned has already spawned; `genDriven` is after the restart tail's `startTui()` and `drive()`. A duplicate `startTui()` on that same tail (the mutation) is sequential before `drive()`, so both children exist before the settle. The historical superseded-tail path does not use this window: it never logs stand-down, so the wait fails closed. I could not re-install `FAKE_CODEX_TUI_SLOW` (absent at this head; positive control: `startTui` is present in `host.ts`).

## 2. Cell B (lower bound)

`>= 2` with `waitFor(..., 30_000)` is sound. The timeout is the failure. Nothing later in the suite reads `tcTuis` or requires the tuicrash log's `tui` count to be exactly 2. The later `tuicrashpeer` block reuses the same jsonl path but asserts crash exit text, not that count.

30s is enough for a replacement TUI on a loaded runner: the suite is ~85–90s locally with this wait succeeding immediately, the shard has no per-suite cap (fail-fast across the list, job `timeout-minutes: 60`), and 30s is far under that. The retained `sleep(1500)` is for the adjacent "mis-classified TUI exit would have called shutdown" check, not for the count.

## 3. Mutation fixture

Graded by reading `bin/smoke/mutations/codex-host-tui-ordering.json` and resolving the anchor with `git show 597fc71df29350a719f8ea1702f7b39701c5ac33:extensions/connector-codex/src/host.ts`. Did not run `mutation-proof`.

`find` is `      startTui();\n      await safeStatus("idle");`. It occurs exactly once, at the restart tail (line 983). Code-only; the comment sits on the previous line. The deletion uses that same two-line anchor, not `includes("")`.

Duplicating that `startTui()` does **not** red cell B first: cell B is a lower bound, so 3 `>= 2` still passes. The earliest named failure is cell A (`=== 1`). Deleting it leaves the tuicrash initial TUI (the other `startTui()`), so the shutdown cell still passes and cell B's `waitFor` times out naming `a replacement TUI is attached to the new listener` (`timeout waiting for ${name}`). `expectRed` is that substring. On a green run those labels print as `✓ …`; a timeout or `AssertionError` is a different line, so the baseline/mutant mark comparison can name them. Opposite edits of one line are a valid pair of reach proofs.

What the fixture does not prove: the historical two-fence genpeer path. That is the admission, not a fixture lie.

## 4. Admission

Honest, and almost adequate. No single-location mutation recreates "one TUI per generation" because the child dies before tools are ready, so both the `awaitMcpReady` currentness fence and the `superseded(gen)` guard must fail together. The mutations instead force a second TUI on the *live* restart tail.

That neighbouring property is what cell A's exact count observes. The original path is not left unguarded: a superseded tail that carries on into `startTui()` does not log stand-down, so the new wait fails closed (25s timeout) rather than counting 1. The admission should have said that. What it correctly leaves unproven is a kill of the historical interleaving itself.

## Checks run

- `pnpm typecheck` at this sha: `Scope: 27 of 28 workspace projects`. The 28th is `@cotal-ai/example-04-frontier-faces` (JS-only, no `typecheck` script). No errors.
- `pnpm smoke:codex-host` twice at this sha: both `CODEX HOST SMOKE PASSED (95 checks)`, ~85s and ~90s. Could not force the slow-TUI interleaving; that knob is not in this tree.
- CI at review time: `unit` SUCCESS; `smoke (shard 0/4)` SUCCESS (the fail-fast shard that used to lose the rest of the partition). Shards 1/4 and 2/4 were still `IN_PROGRESS` — unmeasured, not green.

## Could not check

- Live `mutation-proof` (forbidden: it writes a broken implementation into the tree).
- `FAKE_CODEX_TUI_SLOW` adverse interleaving (absent; not reintroduced).
- Shards 1/4 and 2/4 (in progress / unmeasured).

Worktree left clean (`git status --porcelain` empty).

PR #1288, opened 2026-09-05
test(core): pin thrown observeGeneration unavailable translation

### LINKED ISSUE #1273: test(core): pin the unavailable translation for a thrown observeGeneration in deregisterServiceInstance
`deregisterServiceInstance()` classifies its `observeGeneration` observation into four outcomes, three of which fail closed. Three of the four have permanent cells in `smoke:ep-traits`. The fourth does not.

The uncovered branch is the `try`/`catch` around the call itself:

```ts
try {
  observed = await args.observeGeneration();
} catch (e) {
  throw new EpEnvelopeError("unavailable", `could not observe the issuance-gate generation ... ${(e as Error)?.message ?? String(e)}`);
}
```

The suite covers the adjacent `!wireInt(observed)` branch (a cell passes `undefined` and asserts `/not an unsigned generation/`), but nothing pins the translation of a THROWN observation into `unavailable`.

**This was verified, not inferred.** Replacing the catch body with a bare `throw e` leaves the permanent suite green.

**Why it is worth closing even though the branch is correct today.** It is the production-most failure form. Both shipped call sites reach `observeGeneration` through a real `authKv` read that throws when the issuance-gate row is missing:

- `implementations/manager/src/manager.ts:1890`
- `implementations/manager/src/deregister-instance.ts:202`

Both throw `no issuance gate at ${key}`. So the branch with no permanent cell is the one most likely to fire in the field, and its error code and message can drift without any suite noticing.

**Suggested cell.** In `smoke:ep-traits`, alongside the existing `wundef` / `wahead` cells: hold the governance slot for the instance, pass an `observeGeneration` that throws, and assert the envelope code is `unavailable`, that the message names the observation failure, and that the spec entry remains `PUT` (nothing deleted).

Graded as a non-blocking coverage residual during review of the lost-ack recovery work, and recorded here so it does not disappear into a merged PR.

davidfarah2003:
## Summary
- Add a `smoke:ep-traits` cell that holds the instance governance slot, throws from `observeGeneration`, and asserts `unavailable` plus an un-deleted spec (`PUT`).
- Add a mutation fixture that replaces the catch body with `throw e` and requires that new cell to redden.
- This is a cell in the existing `smoke:ep-traits` suite (already CI-reached). No `ci-suites.d` fragment is needed.

Closes #1273

## Test plan
- [x] `pnpm smoke:ep-traits` green (121 passed)
- [x] `pnpm mutation-proof --config bin/smoke/mutations/observe-generation-unavailable.json` KILLED on `a thrown observeGeneration with a held slot FAILS CLOSED as unavailable`
- [x] `pnpm smoke:gate-inventory` OK
- [x] `pnpm typecheck` Scope 27 of 28 (`@cotal-ai/example-04-frontier-faces` has no typecheck script)

### DIFF
--- a/.changeset/observe-generation-unavailable.md
+++ b/.changeset/observe-generation-unavailable.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/core": patch
+---
+
+Pin the unavailable translation for a thrown observeGeneration in deregisterServiceInstance. The new smoke:ep-traits cell holds the governance slot, throws from the observer, and asserts the envelope stays unavailable with the spec un-deleted.
--- a/bin/smoke/mutations/observe-generation-unavailable.json
+++ b/bin/smoke/mutations/observe-generation-unavailable.json
@@ -1,0 +1,31 @@
+{
+  "suite": "packages/core/smoke/endpoint-traits.smoke.ts",
+  "guard": "deregisterServiceInstance must translate a thrown observeGeneration into an unavailable envelope while the governance slot is held, and must not delete the spec",
+  "command": "pnpm smoke:ep-traits",
+  "completionMarker": "ENDPOINT TRAITS SMOKE",
+  "proveWith": "pnpm mutation-proof --config bin/smoke/mutations/observe-generation-unavailable.json",
+  "why": [
+    "The production-most failure form is a thrown observation (both manager call sites throw",
+    "`no issuance gate at ${key}` when the row is missing). Three of four observeGeneration",
+    "outcomes already have cells; this mutation is the one that used to leave the suite green.",
+    "",
+    "ANCHOR ON CODE ONLY. The `catch (e)` window plus the EpEnvelopeError throw is unique in",
+    "endpoint-service.ts. Do not widen it onto the comment above the function.",
+    "",
+    "The suite imports from ../src/index.js, so a src mutation is what the run sees. No core",
+    "build is required for the mutant to be visible.",
+    "",
+    "expectRed names the FIRST cell that reddens under `throw e`: the envelope-code assertion.",
+    "The spec-still-PUT cell stays green because rethrowing still aborts the delete."
+  ],
+  "mutations": [
+    {
+      "name": "observeGeneration catch rethrows instead of translating to unavailable",
+      "file": "packages/core/src/endpoint-service.ts",
+      "find": "      } catch (e) {\n        throw new EpEnvelopeError(\"unavailable\", `could not observe the issuance-gate generation for \"${args.endpoint}/${iId}\" while its governance slot is held; refusing the delete rather than racing an in-flight registration (SPEC 13.7): ${(e as Error)?.message ?? String(e)}`);\n      }",
+      "replace": "      } catch (e) {\n        throw e;\n      }",
+      "expectRed": "a thrown observeGeneration with a held slot FAILS CLOSED as unavailable",
+      "cell": "a thrown observeGeneration with a held slot FAILS CLOSED as unavailable"
+    }
+  ]
+}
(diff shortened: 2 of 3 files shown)


davidfarah2003:
## APPROVE

**Head reviewed:** `299a3536bd624e86ed3402733ecc12a2d52c4273` (matches `gh pr view --json headRefOid`).

Small, test-only PR (+60/−0). It adds the missing `smoke:ep-traits` cell for the fourth `observeGeneration` outcome in `deregisterServiceInstance`: a **thrown** observation translated into an `unavailable` envelope (endpoint-service.ts:975-976). Three siblings already had cells; this branch used to survive a bare rethrow.

### 1. Does the cell pin what it claims? Yes.
The `wthrow` cell (endpoint-traits.smoke.ts:781-782) asserts all three properties, not just the code:
- envelope code is `unavailable`,
- message names the observation failure: matches both `/could not observe the issuance-gate generation/` and `/no issuance gate/`,
- the follow-up cell (784-785) asserts the spec entry is still `PUT` (nothing deleted).

A cell that checked only the code would pass against a wrong message or a deleted spec; this one does not.

### 2. Mutation fixture — graded hardest. Solid.
- **Anchor resolves exactly once and is code-only.** The `find` block is `} catch (e) { throw new EpEnvelopeError("unavailable", ...) }` with the full message string. Verified against `git show <head>:packages/core/src/endpoint-service.ts`: the exact block occurs **1×**, and the window contains no comment (the comment sits above the function, outside it).
- **Named cell is the earliest red.** Under `throw e`, only the throwing path changes; the non-int path (line 978, `wundef`) and the ahead path (`wahead`) are untouched, so those cells stay green. The raw `Error` from the `wthrow` observer propagates uncaught, so `throwErr.code` is undefined and the message regex fails first at the named cell. No downstream/wrong-red name.
- **Compile-clean:** `pnpm typecheck` passes, `Scope: 27 of 28 workspace projects` (the 28th, `@cotal-ai/example-04-frontier-faces`, is JS-only and skipped).

**I ran the mutation proof** (`pnpm mutation-proof --config bin/smoke/mutations/observe-generation-unavailable.json`):
```
KILLED   observeGeneration catch rethrows instead of translating to unavailable
  red, and named: a thrown observeGeneration with a held slot FAILS CLOSED as unavailable · 0 marks (baseline 0)
```
The suite emits no progress marks (baseline 0), so the below-baseline mark check does not apply here; the kill is established by the named-red mechanism, which fired on the exact `expectRed` cell.

### 3. Independence & governance slot. Confirmed.
The cell registers its own endpoint `ai.cotal.wthrow` / `wthrow` with its own `kvThrow` hook, distinct from `wundef` and `wahead`. It enters the held-slot branch via `beforeWrite` on `govern.wthrow` after the spec is `PUT` (`gov.provisional?.instanceId === iId`), so it exercises the governance slot as #1273 describes and does not lean on state left by the sibling cells.

**Entry-point reachability (the fixture's production claim):** both shipped call sites — `implementations/manager/src/deregister-instance.ts:202` and `manager.ts:1900` — supply an `observeGeneration` that does `authKv.get(epgateKey(...))` and `throw new Error("no issuance gate at ${key}")` on a missing row. The synthetic observer throws the same message shape, so the pinned branch is genuinely production-reachable, not a hand-built dead path.

### 4. Changeset. Correct.
`@cotal-ai/core` patch, one entry, accurate summary. The source lives in `packages/core`, so the package name is right; the `fixed` group versions all packages in lockstep regardless.

### CI at this head
Workflow runs enumerated at the full sha: `Code Quality: PR #1288` = success; `CI` and `Windows` = in_progress (neither failed nor cancelled) at review time.

### Notes
- Worktree left clean: `git status --porcelain` empty, HEAD still `299a3536…` after the proof restored via `git checkout HEAD --`.
- Nothing I was unable to check.

Issue #1274, opened 2026-09-05
manager: a terminalizing static slot cannot say whether cleanup completed, and no read verb projects the phase


davidfarah2003:
A read-only audit of static managed-slot rows on a live mesh found 31 nonterminal slots owned by one manager instance: 27 `active`, and **4 stuck at `terminalizing` with head `retiring` and the issuance gate `frozen`**. For those four, whether footprint cleanup completed is not answerable — not merely unrendered, but **not represented in durable state**.

## The mechanism

`implementations/manager/src/static-lifecycle.ts`, `runStaticTerminal`:

```ts
await hooks.cleanup();
await casStaticSlot(t, { ...slot.row, phase: "retired" }, slotRevision);
```

`cleanup()` runs, then the slot is CAS'd to `retired`. Nothing durable is written between the two. So a row observed at `terminalizing` is consistent with all three of:

1. cleanup has not started;
2. cleanup is in flight;
3. cleanup **completed** and the process died before the CAS.

These have different remedies and the state cannot distinguish them. A resumed terminal re-runs `cleanup()` because that is the only total option, which is correct but means the row can never assert that cleanup was already done.

## There is also no operator read surface

Independently of the durability gap, none of the shipped read verbs projects this state:

- `cotal describe manager` lists 20 served commands; `inspect` is the only per-name reader and returns `name, id, agent, space, mode, status, uptimeMs, mesh, lifecycleUid, model, cwd, pid, spawner, instanceId, host` — no phase, head or op.
- `manager status` returns `instanceId, runtime, agentCount, uptimeMs, connectors`.
- `endpoints`, `ps` and `status` render live presence only.

And the refusals do not discriminate. On a live mesh:

```
inspect {"name":"<a name whose slot is retiring>"} -> not-found: no agent "<name>"
inspect {"name":"zzz-no-such-seat-control"}        -> not-found: no agent "zzz-no-such-seat-control"
```

A name whose retirement stranded and a name that never existed return the same answer, so absence reads as completion with nothing to contradict it.

## Why the two halves are separate work

A new CLI projection would close the second half and **not** the first: there is no field for it to render. Closing this properly means a durable cleanup-complete marker written before the final CAS, and then a read projection over `phase` / head / op.

## Evidence

[...]

PR #1289, opened 2026-09-05
fix(manager): durable cleanup-complete marker on a terminalizing static slot

davidfarah2003:
## What

A static managed slot observed at `terminalizing` was consistent with three different worlds, which durable state could not tell apart:

1. cleanup has not started;
2. cleanup is in flight;
3. cleanup **completed** and the process died before the CAS to `retired`.

`runStaticTerminal` wrote nothing durable between `hooks.cleanup()` and the final `casStaticSlot(..., phase: "retired")`, so a resumed terminal always re-ran `cleanup()` as the only total option, and the row could never assert that cleanup was already done.

This PR does the **durability half** of #1274.

## The marker

- Adds an optional `cleanupComplete?: boolean` to `StaticManagedSlotRow` (closed schema; validated in `parseStaticSlotRow`; absent on legacy rows and on every phase before the cleanup step).
- Routes every `runStaticTerminal` cleanup arm (pre-gate, lost-head, main) through one at-most-once helper, `cleanupStaticSlotOnce`, that:
  - reads the current `terminalizing` slot,
  - **skips** cleanup if `cleanupComplete === true`,
  - otherwise runs `hooks.cleanup()` then CASes `cleanupComplete: true` onto the `terminalizing` row **before** the final CAS to `retired`.

A resumed terminal reads the marker and skips a completed cleanup instead of re-running it.

## Revision discipline (two CAS operations where there was one)

The change turns one slot CAS into two — the `cleanupComplete` marker CAS, then the `retired` CAS — so stale-revision reuse would be the way it could introduce a lost update. It does not chain a captured revision across the marker CAS:

- `cleanupStaticSlotOnce` does its **own** fresh `readStaticSlot`, CASes the marker at that read's revision, and returns the new revision.
- Every arm then **re-reads** the slot (`cur = readStaticSlot(...)`) and CASes `retired` at `cur.revision`. No arm carries a pre-marker `slotRevision` into the `retired` CAS; the previously captured `slotRevision` local was removed. The terminalizing latch CAS at the top is likewise not reused.

Re-reading rather than threading the helper's returned revision is the equivalent safe discipline and is what the three arms already did before this change, so the marker CAS slots in without a revision hand-off. A CAS that loses to a concurrent writer throws rather than silently overwriting.

## Totality — which worlds it distinguishes

The marker splits **world 3** (cleanup completed, process died before the `retired` CAS) off from **worlds 1 + 2** (cleanup never ran / in flight). It deliberately does **not** separate world 1 from world 2, because they share one remedy: re-run the idempotent cleanup (`rmSync` force + secret delete + `deprovisionBroker`).

It adds no unrecoverable state. The new durable point is after cleanup and before the marker CAS; a crash there re-enters the helper, reads `cleanupComplete` unset, and re-runs the idempotent cleanup — the same total default as before. A crash after the marker CAS but before `retired` re-enters, reads the marker set, skips cleanup, and finishes the `retired` CAS. The resume path is the in-code reader of the marker, so it is not write-only at the mechanism layer.

## Scope: this closes the durability half only

Per the issue's own framing, a read projection alone would close the second half and not this one (there would be no field to render). This PR is the field. It uses **`Refs #1274`, not `Closes`**; the issue stays open owning the projection.

**What remains for #1274 (the read-projection half, a separate PR):** project `phase` / head / op through an operator read verb (`inspect` returns fifteen fields, none of them `phase`), and make the refusals discriminate a stranded retirement from a name that never existed (both currently return `not-found: no agent "<name>"`).

## Grading

- New smoke §7c in `static-lifecycle.smoke.ts` (an existing suite; drives the **real** `runStaticTerminal` over a real own-broker JWT stack, not the live fleet) for a marked `terminalizing` slot (world 3: skips cleanup, still reaches `retired`) and an unmarked one (re-runs cleanup once, settles a row carrying the marker). Green 43/43.
- `pnpm mutation-proof --config implementations/manager/smoke/mutations/static-terminal-cleanup-marker.json`: 2 mutants, both KILLED. The load-bearing one **drops the marker write** (`cleanupComplete: true` -> nothing persisted): the resumed terminal then settles a row with no marker and the §7c "then RETIRES marking cleanup complete" cell reddens — the earliest red, at 42 marks vs baseline 43. The second removes the skip-read branch and reddens "SKIPS cleanup (world 3: already done)", also earliest at 42 vs 43. Run after committing (its restore is `git checkout HEAD --`).
- Gating confirmed from the tool, not the file: `pnpm smoke:gate-inventory` -> `GATE INVENTORY OK`. This PR adds no new suite (it extends `smoke:static-lifecycle`), so no `ci-suites.d` fragment is needed; the inventory verdict is the confirmation that the extended suite still runs where CI reaches it, given #1283 changed how coverage is computed (`check` is no longer a root).
- `pnpm typecheck` clean (`Scope: 27 of 28`; the 28th, `@cotal-ai/example-04-frontier-faces`, is JS-only and skipped).
- `pnpm check:docs-voice` passes. Changeset bumps the fixed group at minor.

**Not verified:** graded by reading and non-live suites only. Per this machine's constraints I did not run `:live` stages, `pnpm check`, or the live audit that produced the issue, so behavior against the live mesh fleet is unverified here. The wire contract (gate/head/ledger, SPEC §13.1) is unchanged — the marker is a field on the manager-local `mgrslot` record, not a wire message — so `SPEC.md` does not move.

### DIFF
--- a/.changeset/static-slot-cleanup-marker.md
+++ b/.changeset/static-slot-cleanup-marker.md
@@ -1,0 +1,6 @@
+---
+"@cotal-ai/core": minor
+"@cotal-ai/manager": minor
+---
+
+Record a durable cleanup-complete marker on a terminalizing static managed slot before the final CAS to `retired`. A `terminalizing` row was previously consistent with three worlds (cleanup not started, cleanup in flight, or cleanup completed with the process dead before the CAS) that the state could not tell apart, so a resumed terminal always re-ran `cleanup()` as the only total option and the row could never assert cleanup was already done. `runStaticTerminal` now runs the footprint cleanup through one at-most-once helper that writes `cleanupComplete: true` on the `terminalizing` slot before the `retired` CAS; a resumed terminal reads it and skips a completed cleanup. The marker distinguishes the cleanup-completed world from the not-known-complete worlds (which keep the same remedy: re-run the idempotent cleanup) and adds no unrecoverable intermediate state.
--- a/bin/smoke/required-arg-seam.smoke.ts
+++ b/bin/smoke/required-arg-seam.smoke.ts
@@ -290,7 +290,10 @@ const SEAMS: Seam[] = [
   // records KV the lost-ack same-op recovery needs. It stays under smoke/ with the suite's other
   // three sites: the call is harness residue, not product connect, and it already states tls: false.
   // Every added site states the transport decision explicitly.
-  { fn: "standaloneConnectOpts", key: "tls", sites: 118, untypecheckedSites: 87 },
+  // 118/87 -> 119/88: static-lifecycle.smoke.ts's #1274 crash-resume cell (driveTerminalDirect) opens
+  // one more lifecycle-executor connection to plant a terminalizing slot and drive runStaticTerminal
+  // as a resume would. It is under smoke/, harness residue not product connect, and states tls: false.
+  { fn: "standaloneConnectOpts", key: "tls", sites: 119, untypecheckedSites: 88 },
 ];
 
 /**
--- a/implementations/manager/smoke/mutations/static-terminal-cleanup-marker.json
+++ b/implementations/manager/smoke/mutations/static-terminal-cleanup-marker.json
@@ -1,0 +1,37 @@
+{
+  "suite": "implementations/manager/smoke/static-lifecycle.smoke.ts",
+  "guard": "The static terminal records a durable cleanup-complete marker on the `terminalizing` slot BEFORE its final CAS to `retired` (#1274). A resumed terminal reads the marker and skips a completed cleanup instead of re-running it; a terminal that ran cleanup settles a row that carries the marker.",
+  "command": "pnpm smoke:static-lifecycle",
+  "completionMarker": "STATIC-LIFECYCLE SMOKE",
+  "proveWith": "pnpm mutation-proof --config implementations/manager/smoke/mutations/static-terminal-cleanup-marker.json",
+  "why": [
+    "§7c plants two crashed-terminal shapes over a real JWT broker and drives the real",
+    "runStaticTerminal as a resume would. `resumemarked` carries the durable cleanupComplete",
+    "marker already set (world 3: cleanup finished, the process died before the retired CAS);",
+    "`resumeplain` does not (worlds 1/2: cleanup never ran or was in flight).",
+    "",
+    "The two mutations are each other's control. Dropping the skip-when-marked branch makes the",
+    "resume re-run a cleanup it should have skipped, and the marked cell must notice. Dropping the",
+    "marker write leaves a completed cleanup unrecorded, and the plain cell (which asserts the",
+    "settled row carries the marker, written before the retired CAS) must notice. A repair that did",
+    "only one would have left the marker unreadable or unwritten one layer down."
+  ],
+  "mutations": [
+    {
+      "name": "the skip-when-cleanupComplete branch is removed, so a resume re-runs a completed cleanup",
+      "file": "implementations/manager/src/static-lifecycle.ts",
+      "find": "  if (slot.row.cleanupComplete === true) {\n    log(`footprint cleanup already recorded complete for uid ${args.lifecycleUid}; skipping the re-run (#1274 marker)`);\n    return { revision: slot.revision };\n  }\n",
+      "replace": "",
+      "expectRed": "SKIPS cleanup (world 3: already done)",
+      "cell": "a resume over a cleanupComplete=true terminalizing slot SKIPS cleanup (world 3: already done)"
+    },
+    {
+      "name": "the cleanup-complete marker write is dropped, so a completed cleanup is never recorded",
+      "file": "implementations/manager/src/static-lifecycle.ts",
+      "find": "  const revision = await casStaticSlot(t, { ...slot.row, cleanupComplete: true }, slot.revision);",
+      "replace": "  const revision = await casStaticSlot(t, { ...slot.row }, slot.revision);",
+      "expectRed": "then RETIRES marking cleanup complete",
+      "cell": "a resume over an UNMARKED terminalizing slot re-runs cleanup once, then RETIRES marking cleanup complete (write before the retired CAS)"
+    }
+  ]
+}
--- a/implementations/manager/smoke/static-lifecycle.smoke.ts
+++ b/implementations/manager/smoke/static-lifecycle.smoke.ts
@@ -75,6 +75,7 @@ import {
   casStaticSlot,
   recordSlotCredential,
   appendStaticCredentialRow,
+  runStaticTerminal,
 } from "../src/static-lifecycle.js";
 import { bootBroker } from "./_boot-broker.js";
 
@@ -95,7 +96,7 @@ let ran = 0;
  * when a `check` is deliberately added or removed, because completeness is something only this
  * suite knows, so it has to be the thing that says it.
  */
-const EXPECTED_CHECKS = 40;
+const EXPECTED_CHECKS = 43;
 function check(label: string, cond: boolean, extra?: unknown): void {
   console.log(`${cond ? "✓" : "✗"} ${label}${cond ? "" : ` — ${JSON.stringify(extra) ?? ""}`}`);
   ran++;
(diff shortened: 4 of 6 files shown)


davidfarah2003:
**APPROVE** — sha `5eaed67f732e659192c5ced2a4ae0d0de948b3e4`

Independent second-vendor review, read + probe.

### 1. CAS revision discipline

Verified. Every `retired` CAS re-reads the slot fresh: pre-gate arm (line 356–358), lost-head arm (376–378), main arm (438–440), each doing `const cur = await readStaticSlot(...)` then `casStaticSlot(..., cur.revision)`. The pre-marker `let slotRevision = slot.revision` local was deleted at 344 and never re-introduced; the terminalizing latch CAS at 345 no longer captures a return. `cleanupStaticSlotOnce` does its own fresh read + CAS at 309/314. The helper's returned revision is not consumed — every arm re-reads instead. That is the equivalent-safe discipline the three arms already used and it survives contention because the re-read observes any concurrent writer's row (or `casStaticSlot` throws). No stale-revision reuse.

### 2. Totality of the new intermediate crash point

The recovery path exists in code: `cleanupStaticSlotOnce` at 306–312 reads `slot.row.cleanupComplete === true` and returns early; the caller then finds `phase !== "retired"` and CASes it. Enumerated post-crash states over the two new points:

- crash **before** cleanup: row `terminalizing`, marker absent → resume runs idempotent cleanup, sets marker, retires. Recoverable.
- crash **between** cleanup and marker CAS: row `terminalizing`, marker absent → resume re-runs the idempotent cleanup, sets marker, retires. World 1/2 semantics preserved; still recoverable because cleanup (`rmSync` force + secret delete + `deprovisionBroker`) is idempotent.
- crash **after** marker CAS, **before** `retired` CAS: row `terminalizing`, marker set → resume reads marker, skips cleanup, CASes retired. Recoverable — this is the new distinguishable case.

The ambiguity is not simply moved: world 3 is genuinely split off. Worlds 1 and 2 remain merged, correctly, because they share the same total remedy. No new unrecoverable state is introduced.

### 3. All three cleanup arms route through the helper

`grep hooks.cleanup implementations/manager/src/static-lifecycle.ts` returns exactly three hits: lines 355 (pre-gate), 375 (lost-head), 424 (main), all inside `cleanupStaticSlotOnce(...)`. Zero direct `hooks.cleanup()` invocations remain and no fourth arm exists. Confirmed by reading `runStaticTerminal` end-to-end.

### 4. Mutation fixture — verified by running it

Ran `pnpm mutation-proof --config implementations/manager/smoke/mutations/static-terminal-cleanup-marker.json`:

```
KILLED  the skip-when-cleanupComplete branch is removed …
  red, and named: SKIPS cleanup (world 3: already done) · 42 marks (baseline 43)
KILLED  the cleanup-complete marker write is dropped …
  red, and named: then RETIRES marking cleanup complete · 42 marks (baseline 43)
```

**Regime: continue-on-failure.** `check()` at smoke:99 only increments `failures`, never throws; the trailer at 474 exits nonzero if `failures > 0`. So the fact that each mutation drops the count by **exactly one** is stronger than the fail-fast reading: it proves the named cell is the ONLY red one, not merely the earliest. §7c and §8 all ran to completion, and no unnamed cell reddened as a side-effect.

Anchor hygiene: both `find` strings resolve exactly once in the src file; both `expectRed` strings resolve exactly once in the smoke; the deletion mutation has a real, non-empty multi-line anchor (no `includes("")` degeneracy); anchors are code, not comment text. `expectRed` is treated as a text substring in the harness output — consistent with the fixture format.

### 5. Core schema field

`cleanupComplete?: boolean` is added to `StaticManagedSlotRow` in `packages/core/src/lifecycle-state.ts`. Field is optional, so legacy rows without it still parse (`parseStaticSlotRow` at 523 accepts undefined). `STATIC_SLOT_PHASES` is unchanged and no wire message kind gains a field. The `StaticManagedSlotRow` type is not a wire envelope: it is the manager-local `mgrslot`/records-KV row (the manager writes it via `casStaticSlot`; other components never publish it). No SPEC.md movement is required.

That said, `packages/core` is the "wire standard" tier and it does host `StaticManagedSlotRow` today — the field is manager-shaped in intent but the type sits in core because core owns the closed schema for the row. This is pre-existing coupling, not a regression from this PR; the author's argument holds for **this** change. If core is intended to stay strictly wire-only, the whole `StaticManagedSlotRow` would need relocation, which is out of scope here.

### Scope split (extra)

PR body states plainly `This PR does the durability half of #1274` and explicitly enumerates the read-projection remainder: `project phase / head / op through an operator read verb … make the refusals discriminate a stranded retirement from a name that never existed`. Uses `Refs #1274` rather than `Closes`. Confirmed `gh issue view 1274` still `OPEN`. The residual will not silently vanish.

### Environment checks

- `pnpm typecheck`: clean, `Scope: 27 of 28 workspace projects` as expected; the skipped 28th is `@cotal-ai/example-04-frontier-faces` (JS-only, documented).
- `pnpm smoke:static-lifecycle`: green, `43/43 assertions`.
- `pnpm smoke:gate-inventory`: `GATE INVENTORY OK ✅`.
- CI workflow runs at exact sha via `repos/…/actions/runs?head_sha=`: `CI in_progress`, `Windows queued`, `Code Quality: PR #1289 success`. No `cancelled` and no completed failure at read time.
- Worktree clean; `git status` empty; still detached at the exact sha.

### Not verified

Per the "do not" list I did not run `pnpm check`, any `:live` suite, `cotal up`, the dev-tree CLI, or the live audit that produced #1274, so behavior against the live mesh fleet remains unverified here. The named-cell-only-red evidence for the mutation fixture is a strong local proof of test dependence; it is not a proof that the manager's real spawn/terminal path reaches this helper (the smoke drives `runStaticTerminal` directly through `driveTerminalDirect`, planting the slot by hand). CI at this sha had not completed at report time.

Verdict: approve — the two-CAS refactor is disciplined, the marker splits world 3 without introducing any unrecoverable intermediate, cleanup is single-entry through the helper, the mutation fixture actually discriminates, and the schema addition is legacy-compatible.

davidfarah2003 (commented):
**APPROVE** `5eaed67f732e659192c5ced2a4ae0d0de948b3e4`

Independent exact-head review of the durability half of #1274. Confirmed `git rev-parse HEAD` and `gh pr view 1289 --json headRefOid` both equal that sha.

## 1. CAS revision discipline

No lost-update bug in the plumbing I read.

There are now three slot CASes on a fresh terminal, not one:

- **Latch** (`phase: "terminalizing"`): uses the **initial** `readStaticSlot` revision (`slot.revision`). The returned revision is discarded. The old `slotRevision` local that used to carry this value into the `retired` CAS is gone.
- **Marker** (`cleanupComplete: true` inside `cleanupStaticSlotOnce`): uses a **fresh** `readStaticSlot` of its own, then `casStaticSlot(..., slot.revision)` at that read. Returns the new revision.
- **Retired**: every arm **re-reads** (`cur = readStaticSlot(...)`) and CASes at `cur.revision`. No arm threads the pre-marker `slotRevision`. No arm uses the helper's returned revision either (the comment that a caller "takes the marker CAS's revision" is stale; the implementation re-reads, which is the equivalent safe discipline).

`casStaticSlot` / `updateRecordEntry` return the new KV revision and throw `conflict` on CAS loss. A concurrent writer cannot be silently overwritten.

I rebuilt the two-CAS shape from the diff rather than trusting the PR body: the previously captured `slotRevision` is not an input to the `retired` CAS.

## 2. Totality / crash recovery

The claimed resume path **exists in code**, not only in the PR body:

```ts
if (slot.row.cleanupComplete === true) {
  log(`... skipping the re-run (#1274 marker)`);
  return { revision: slot.revision };
}
await cleanup();
const revision = await casStaticSlot(t, { ...slot.row, cleanupComplete: true }, slot.revision);
```

Then each arm re-reads and CASes `retired` if the same uid is still not retired. Smoke §7c plants `terminalizing` + `cleanupComplete: true` and drives the real `runStaticTerminal`; locally that cell skipped cleanup (`cleanupCalls === 0`) and still reached `retired`.

Distinguishable durable states a crash can now leave:

| Durable row | Meaning | Resume |
|---|---|---|
| `terminalizing`, marker absent/`false` | worlds 1+2: cleanup never started **or** in flight **or** finished but died before the marker CAS | re-run idempotent cleanup, write marker, retire |
| `terminalizing`, `cleanupComplete: true` | world 3: cleanup finished, died before `retired` | skip cleanup, finish `retired` |
| `retired` (marker typically still true via `{...cur.row, phase:"retired"}`) | done | idempotent return |

Worlds 1 and 2 remain merged on purpose; they share the re-run remedy. The new crash window (after `cleanup()` returns, before the marker CAS) is the **same** durable state as world 1, so that ambiguity moved one step later rather than being removed. World 3 is the world #1274 could not name; it is now named and recoverable. No new unrecoverable state.

I did not prove `cleanup()` is idempotent against the live fleet (see gaps). The hook is `rmSync({force})` + secret delete + `deprovisionBroker`.

## 3. All three cleanup arms

In `static-lifecycle.ts`, `hooks.cleanup()` is passed only into `cleanupStaticSlotOnce`, at the three arms:

- pre-gate (no gate): line 355
- lost-head (activation never won): line 375
- main: line 424

No fourth `runStaticTerminal` arm. No remaining direct `hooks.cleanup()` call in that file.

There **is** a separate `cleanup()` in `driveStaticRetirement` when the slot is missing or at another uid (pre-Unit-B / stale teardown). That path never enters `runStaticTerminal`. Out of this PR's claim.

## 4. Mutation fixture

**Regime: continue-on-failure, not fail-fast.** `check()` logs `✗`, increments `failures`, and keeps going; `EXPECTED_CHECKS` is enforced after the loop. So "the named cell is the earliest to redden" is the weaker claim.

Re-grade under continue-on-failure: the named cell must be *among* the red cells, and no unnamed cell may also redden.

I ran `pnpm mutation-proof --config implementations/manager/smoke/mutations/static-terminal-cleanup-marker.json` on this committed tree:

- baseline green, **43** marks
- skip-branch removed: **KILLED**, named `SKIPS cleanup (world 3: already done)`, **42 vs 43**
- marker write dropped: **KILLED**, named `then RETIRES marking cleanup complete`, **42 vs 43**

In this harness a `✓` is a progress mark and a failed cell still runs later cells. **Exactly one lost mark** means exactly one cell lost its `✓`. Combined with mutation-proof requiring a named line the green run does not print, that is the right proof for a continue-on-failure suite: the named cell reddened and no other cell did.

`expectRed` is text. Both anchors applied `1×` (unambiguous, code-only). The deletion mutation's `find` is the skip-branch body, not `""`.

Tree left clean: mutation-proof restored; `git status --porcelain` empty at the same sha.

## 5. Core schema

`cleanupComplete?: boolean` is on `StaticManagedSlotRow` in `packages/core/src/lifecycle-state.ts`, added to the closed `allowed` set, type-checked as boolean-if-present. Absent is valid, so legacy rows still parse.

The author's SPEC argument holds **for this field**: `SPEC.md` does not document the `mgrslot` *value* schema today (no `mgrslot` / `StaticManagedSlotRow` hits). Head, gate, and ledger *are* SPEC §13.1. The slot key is already a granted records-store key on the `lifecycle-executor` profile (`mgrslot.<owner>.<alias>`), but this PR does not change subjects, message kinds, or that key. Putting the field next to `ownerInstanceId` on the existing manager-local row is consistent with how this module already owns the slot grammar so permission builders and the manager share one encoder.

A mixed-version old parser would reject a marked row as an unknown field and could fail a whole `reconcileStaticLifecycles` sweep (parse is not per-row). That is the usual closed-schema tax; the lockstep minor bump of the `fixed` group is the intended upgrade path, same as `ownerInstanceId`. Not a SPEC miss, and not a BLOCK.

Changeset correctly minors `@cotal-ai/core` and `@cotal-ai/manager`.

## Checks run

- `pnpm typecheck`: clean. **Scope: 27 of 28 workspace projects**; `@cotal-ai/example-04-frontier-faces` is JS-only and skipped.
- `pnpm smoke:static-lifecycle`: `STATIC-LIFECYCLE SMOKE OK (43/43 assertions)`
- `pnpm smoke:gate-inventory`: `GATE INVENTORY OK` (extended existing suite; no new `ci-suites.d` fragment required)
- mutation-proof: both mutants KILLED as above
- worktree clean after restore

## CI at `5eaed67f732e659192c5ced2a4ae0d0de948b3e4` (workflow runs, not check-runs)

- Code Quality: PR #1289 — **success**
- Windows — **success**
- CI (`33963775383`) — **in_progress** at review time. Jobs: `unit` success, `live` success, smoke 3/4 success, smoke 1/4 and 2/4 in_progress, **smoke 0/4 failed** at `pnpm smoke:delivery-renewal` (`evictPrincipal force-drops the victim`, kicked 0 vs expected drop). That suite is outside this PR's five files. I am not treating it as a defect of this change. Cancelled would be unmeasured; this is a measured unrelated red.

## Not verified (named gaps)

- No `cotal up`, no `:live` / `-live` suites, no `pnpm check`, no live-mesh audit that produced #1274, no writes to the live fleet, no `pnpm cotal` / `tsx bin/cotal.ts`.
- So: live-fleet resume of the four stranded `terminalizing` slots, mixed-version manager parse of a marked row, and `deprovisionBroker` idempotence under a real broker are unverified here.
- Helper return value unused (re-read instead): functionally fine; comment overstates the hand-off.

Durability half of #1274 is in code, revision-safe, total over the new crash point, and mutation-proofed. The read-projection half stays on the issue (`Refs`, not `Closes`).

PR #1287, opened 2026-09-05
chore(release): version packages

davidfarah2003:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

- ba9af19: Refuse a source-checkout `cotal` from writing or GC'ing the operator-global seed store. A missing identity answer is a refusal, not a released install. The refusal names `$XDG_CONFIG_HOME` isolation, not the test-only `COTAL_ALLOW_CHECKOUT_SEED=1` override. An older CLI that meets a newer store is pointed at `cotal ext seed --force`, not `--reset`. `COTAL_HOME` does not relocate this store.
### Minor Changes

- ba9af19: Refuse a source-checkout `cotal` from writing or GC'ing the operator-global seed store. A missing identity answer is a refusal, not a released install. The refusal names `$XDG_CONFIG_HOME` isolation, not the test-only `COTAL_ALLOW_CHECKOUT_SEED=1` override. An older CLI that meets a newer store is pointed at `cotal ext seed --force`, not `--reset`. `COTAL_HOME` does not relocate this store.

### Patch Changes

- 2850a5a: Count a smoke suite as gated only when a CI job actually runs it. join-external live coverage now rides its own live-job step; a duplicate connect classifier that only `pnpm check` reached is gone. Backup live suites stay UNGATED as already-red (#643 / #1285).
  - @cotal-ai/core@0.44.0
  - @cotal-ai/workspace@0.44.0
### Patch Changes

- 2850a5a: Count a smoke suite as gated only when a CI job actually runs it. join-external live coverage now rides its own live-job step; a duplicate connect classifier that only `pnpm check` reached is gone. Backup live suites stay UNGATED as already-red (#643 / #1285).
- Updated dependencies [2850a5a]
- Updated dependencies [ba9af19]
  - @cotal-ai/cli@0.44.0
  - @cotal-ai/connector-core@0.44.0
  - @cotal-ai/manager@0.44.0
  - @cotal-ai/core@0.44.0
  - @cotal-ai/workspace@0.44.0
  - @cotal-ai/runtime@0.44.0
  - @cotal-ai/delivery@0.44.0
  - @cotal-ai/auth@0.44.0
### Patch Changes

[...]

PR #1290, opened 2026-09-05
fix(cli): name COTAL_NAME in both send identity alternatives

### LINKED ISSUE #1276: cli: the send identity refusal names two alternatives, but COTAL_NAME is required in both
`callerFromEnv` in `implementations/cli/src/lib/transient.ts` requires `COTAL_NAME` in **both** accepted shapes:

```ts
const name = env.COTAL_NAME?.trim();
const owner = env.COTAL_OWNER?.trim();
const actor = env.COTAL_ACTOR?.trim();
const id = env.COTAL_ID?.trim();
if (!name) return undefined;
if (owner && actor) return { name, owner, actor };
if (id) return { name, owner: "local", actor: id };
return undefined;
```

So the real requirement is: **`COTAL_NAME`, plus either `COTAL_ID` or (`COTAL_OWNER` and `COTAL_ACTOR`).**

The refusal message and the docs both present it as two independent alternatives:

> `cotal send` requires a managed seat identity (COTAL_NAME plus COTAL_ID, or COTAL_OWNER plus COTAL_ACTOR).

Read plainly, the second alternative does not mention `COTAL_NAME`. An operator who sets `COTAL_OWNER` and `COTAL_ACTOR` without `COTAL_NAME` is refused, reads the message, and concludes they have already satisfied it.

This is the text someone reads at the exact moment they are blocked, which is when a wrong requirement costs the most.

**Not a correctness defect.** The code is stricter than the text, so the failure direction is an unexpected refusal rather than an unexpected pass. Nothing is admitted that should not be.

**Suggested wording:** `cotal send` requires `COTAL_NAME` plus either `COTAL_ID` or both `COTAL_OWNER` and `COTAL_ACTOR`.

Fix the refusal string and the corresponding docs line together, so the two cannot drift again.

Found during exact-head review of the send-identity work and recorded as a non-blocking residual rather than folded in, because that PR was under a deliberate head freeze so its CI could complete for the first time.

davidfarah2003:
## Scope

Closes #1276.

`callerFromEnv` already requires `COTAL_NAME` in both accepted shapes. The refusal and the CLI docs presented two independent alternatives, and the second (`COTAL_OWNER` plus `COTAL_ACTOR`) did not mention `COTAL_NAME`. An operator who set only those two was refused by a message that said they had already satisfied it.

This is a wording fix only. The code was stricter than the text; the failure direction was an unexpected refusal, never an unexpected pass.

## Copies

Grep for the distinctive fragments (`requires a managed seat identity`, `COTAL_NAME plus COTAL_ID`, `needs a complete seat identity`) found **4 live copies**, all updated in this change:

1. `implementations/cli/src/commands/send.ts` — the operator-facing refusal
2. `docs/cli.md` — the send section
3. `extensions/connector-core/src/docs-bundle.generated.ts` — regenerated from `docs/cli.md` (`pnpm gen:docsbundle`)
4. `implementations/cli/smoke/mutations/send-entry.json` — mutation `find` string that duplicated the refusal

Historical CHANGELOG entries for the original identity work were left as-is (they describe the prior wording, not the current contract).

## Pin

Added a cell to the existing `smoke:send` suite that includes the exact new sentence. No new suite, so no `bin/smoke/ci-suites.d` fragment.

Also added a mutation that reverts the refusal to the old independent-alternatives wording and requires that cell to redden.

## Verify

- `pnpm typecheck` — Scope 27 of 28 (`@cotal-ai/example-04-frontier-faces` is JS-only)
- `pnpm check:docs-voice`
- `pnpm smoke:send` — 23 passed, 0 failed (22 named cells plus the expected-count cell)
- `pnpm smoke:gate-inventory` — GATE INVENTORY OK
- `pnpm mutation-proof --config implementations/cli/smoke/mutations/send-entry.json` — all 5 mutations killed, including the new wording cell


davidfarah2003:
## APPROVE

Independent review at head `2a3429543a22ccb968392b853d20717a494494a8` (confirmed equal to the PR head ref). Worktree left clean.

**Q1 — Is the new wording true?** Yes. `callerFromEnv` (`implementations/cli/src/lib/transient.ts`) returns `undefined` unless `COTAL_NAME` is set, then requires either (`COTAL_OWNER` **and** `COTAL_ACTOR`) or `COTAL_ID`. The refusal now reads "requires COTAL_NAME plus either COTAL_ID or both COTAL_OWNER and COTAL_ACTOR." — an exact derivation of the code. The old text presented the second shape without `COTAL_NAME`, so an owner+actor operator was wrongly told they had satisfied it. The code was already correct; only the text moved.

**Q2 — Every copy found?** Yes: **4 live copies**, matching the author's count.
1. the refusal in `implementations/cli/src/commands/send.ts`
2. `docs/cli.md`
3. the regenerated `extensions/connector-core/src/docs-bundle.generated.ts`
4. the mutation `find` anchor in `implementations/cli/smoke/mutations/send-entry.json`

Plus the PR's own additions: the literal pin in `send.smoke.ts` and the changeset. I searched distinctive fragments of both old and new wording across `*.ts`/`*.md`/`*.json`. The old sentence (`COTAL_NAME` plus `COTAL_ID`, or `COTAL_OWNER` plus `COTAL_ACTOR`) survives **only** in two `CHANGELOG.md` files (`implementations/cli`, `extensions/connector-core`) — correct, since those record what shipped. No live copy is hiding among them.

**Q3 — Bundle is a true regeneration.** Ran `node scripts/generate-docs-bundle.mjs` on a clean tree; `git status --porcelain` came back empty. Not a hand edit. Restored/confirmed clean afterward. (Stamp is `0.43.0`, correct for current main; the #1287 conflict is known and out of scope.)

**Q4 — Mutation fixture.** `pnpm mutation-proof --config implementations/cli/smoke/mutations/send-entry.json`: baseline green (twice), **all 5 mutations KILLED**, each red on its named assertion. The one that matters — "the send identity refusal again names two independent alternatives" — reverts the string to the old owner+actor-without-NAME wording and dies on "identity refusal names COTAL_NAME as required in both accepted shapes." All 5 `find` anchors resolve **exactly once** and are **code-only** (const decls, `console.error` calls, a function call); none spans a comment, so the prose edits in this PR will not dislodge them. `expectRed` values are assertion text, as required.

**Q5 — The pin.** `send.smoke.ts:128` asserts the literal sentence via `.includes(...)`. It will break the next time someone rewords the message — that is the guard working as intended, not a flaw: the text cannot drift silently. A future reader hitting that failure should reword the pin to match.

**Standards run at this head:**
- `pnpm typecheck` exit 0. The 28th project `@cotal-ai/example-04-frontier-faces` is JS-only and not typechecked (reported, not counted as clean).
- `pnpm check:docs-voice` — 40 pages passed.
- `pnpm smoke:gate-inventory` — OK (its BROKEN/untriaged listings are pre-existing, unrelated to this PR).
- `pnpm smoke:send` — green as the mutation-proof baseline (ran twice).
- CI by workflow runs at the full sha: Docs `success`, Windows `success`, Code Quality `success`; **CI `in_progress`** (not yet conclusive). No `failure` or `cancelled`.

**Could not fully verify:** the `CI` workflow was still `in_progress` at review time — approving on the substantive change; a maintainer should confirm it lands green before merge.

PR #1294, opened 2026-09-05
feat(manager): add custody seam and legacy continuity report

### LINKED ISSUE #783: manager cannot re-mint or update without killing every seat PTY: need hot reload / rolling handoff

davidfarah2003:
## Summary

This is the M1 custody seam for rolling manager updates, not rolling PTY handoff. An earlier session implemented the typed Runtime.adopt / RuntimeReference contract, a legacy continuity report, and cotal update refusing to call a legacy manager a hot update. Follow-up work registered the four new smoke suites and made the packaged-manager smoke portable. Review at 71974bb61 then required four corrections, which this head carries. It does not close #783.

Refs #783

### What ships

- Opaque durable runtime reference `{ kind, id }` on the runtime contract. Runtime.adopt is optional. A required method that every runtime must implement as throw is not a custody stance, it is a compile tax. Runtimes without durable custody omit adopt. The manager-side helper requireRuntimeAdopt throws `runtime "<kind>" does not support adopt` rather than spawning a replacement. The shipped Manager.adoptRuntimeHandle method delegates to that helper. No current control or update flow calls adoptRuntimeHandle. It is the refusal the first real adoption path will use (M2). smoke:attach hands it a runtime with no adopt and asserts the exact string, and also asserts the live pty runtime (which now omits adopt) is refused by name. A typed RuntimeAdoptionUnsupportedError is not in this PR: with no control/update caller it would not make any caller more actionable, and capability absence is modeled by optional adopt first.
- Relaxing a required interface member to optional is source-compatible for implementors (an out-of-tree runtime that already wrote a stub keeps compiling) and only affects callers. No current control or update flow calls adoptRuntimeHandle. minor remains right.
- Manager legacy continuity report (exact / fork / fresh / drain-only from connector flags, including supportsFreshStart; unknown connectors are drain-only).
- cotal update prints that report and refuses replay against a legacy manager. cotal update --self reports the selected manager before any global install, and the replacement child is launched with the same --space / --server / --creds selection.
- Docs state that custody transfer is not implemented.
- .changeset/custody-seam.md names the packages that changed: cotal-ai, @cotal-ai/core, @cotal-ai/cli, @cotal-ai/manager, @cotal-ai/cmux, @cotal-ai/tmux, @cotal-ai/orca, @cotal-ai/herdr, @cotal-ai/connector-core, @cotal-ai/connector-opencode, @cotal-ai/connector-hermes, @cotal-ai/connector-codex.

### What remains (item 2)

There is still no seat host and no FD/IPC custody transfer. PtyRuntime still owns its child in-process and exposes no reference. The smoke:legacy-pty-custody suite asserts the M1 residual: the counter child dies when the manager worker is killed. Child-PID survival is not done.

The M1 custody evidence is Linux-only. legacy-pty-custody skips the whole suite off Linux because the residual it gathers is /proc/<pid>/stat after SIGKILL. The process.execPath launch fix is therefore unreachable on Windows. Narrowing that skip so portable checks run everywhere, with only the /proc zombie-state evidence gated, is #1298 (distinct from #1297, which is the shard runner grading a zero-assertion exit(0) as a covered pass).

## Test plan

- [x] pnpm smoke:gate-inventory (four new suites registered as hashed fragments, frozen ci-suites.txt untouched)
- [x] pnpm check:shard-stability against the merge tree with origin/main (STABLE; comparing the unmerged head to main is a false reshuffle from main-only suites)
- [x] pnpm typecheck (27 of 28 workspace projects; @cotal-ai/example-04-frontier-faces is JS-only and has no typecheck script)
- [x] pnpm smoke:runtime-contract
- [x] pnpm smoke:legacy-manager-report
- [x] pnpm smoke:legacy-pty-custody (child dies, Linux-only)
- [x] pnpm smoke:cli-kernel
- [x] pnpm smoke:legacy-packaged-manager (isolated HOME / XDG_CONFIG_HOME / TMPDIR, no COTAL_* on child env, operator stamp compared as raw bytes and not pinned to one host)
- [x] pnpm smoke:attach (absent adopt refuses by runtime name)
- [x] pnpm mutation-proof --config implementations/cli/smoke/mutations/legacy-manager-update.json (killed on the legacy fallthrough cell)
- [x] pnpm mutation-proof --config implementations/cli/smoke/mutations/self-update-handoff.json (report-before-install and selected-target argv)
- [ ] CI on the merge-ref with current main

No :live suite, pnpm check, or dev-tree CLI was run.


davidfarah2003:
## Independent review — findings at `647e21b5ee9284847e7df87862c6e467807ca9cc`

These are findings at the sha I reviewed, **not a terminal verdict on the current head**. The head moved to `2fc985c5d5451b9db9dd6f2d01ba1e8f846196b6` during the review, so where it matters I note whether each finding is addressed there (verified by `git diff`/`git show`, not by re-running). This was an independent whole-change pass: nothing here decomposes against a prior approval.

Scope confirmed against main: 31 files, +526/-13; `merge-base..head` and `main...head` agree.

### Lead finding — the four newly-gated custody suites NEVER RAN in CI at this head

The entire point of this remediation was to get `smoke:legacy-pty-custody`, `smoke:legacy-manager-report`, `smoke:legacy-packaged-manager`, and `smoke:runtime-contract` executing in CI. **CI has not demonstrated that even once.**

At `647e21b`, the CI run (`33973541714`) resolved:

- `unit` — **failure**
- `smoke (shard 2/4)` — **failure** (fail-fast at `smoke:flag-inventory`; 93 of 131 planned suites in that shard never started)
- `smoke (shard 0/4)`, `smoke (shard 1/4)`, `smoke (shard 3/4)` — **cancelled**

The four gated suites hash to shards 0 (`legacy-manager-report`, `runtime-contract`), 1 (`legacy-pty-custody`), and 3 (`legacy-packaged-manager`) under `fragmentShard(name, 4)`. Those three shards were cancelled by the fail-fast on shard 2, so **the four suites did not execute in CI**. They pass locally for me (below), but nobody should merge until a CI run exists in which all four actually run and pass.

### The three blocking defects (all PR-caused, each reproduced locally)

**1. `smoke:flag-inventory` fails — `update` golden not updated for the new flags.**
`update` gained `--space`, `--server`, `--creds`, but `bin/smoke/flag-inventory.smoke.ts` golden still declared `update: { flags: ["self:boolean"] }`.
```
AssertionError: flags of `update` match golden
  actual:   [ 'creds:string', 'self:boolean', 'server:string', 'space:string' ]
  expected: [ 'self:boolean' ]
```
Reproduced locally and seen in CI shard 2. **Addressed at `2fc985c`:** yes — golden is now `[...TARGET, "self:boolean"]`.

**2. `unit` / `check:docsbundle` fails — docs bundle not regenerated after the `docs/cli.md` change.**
The `## update` section in `docs/cli.md` was edited, but `extensions/connector-core/src/docs-bundle.generated.ts` was not regenerated. `pnpm gen:docsbundle` locally produces a 1-line diff (the generated bundle still carried the old single-`--self` `update` section). **Addressed at `2fc985c`:** yes — the bundle now contains `cotal update [--self] [--space <s>] [--server <url>] [--creds <path>]` and the new continuity text.

**3. Windows `required` fails — `legacy-pty-custody.smoke.ts` spawns `tsx` non-portably.**
`implementations/manager/smoke/legacy-pty-custody.smoke.ts:50` does `spawn(join(repo, "node_modules", ".bin", "tsx"), [owner], …)` with no `.cmd` resolution and no `shell`. On Windows this throws:
```
Error: spawn D:\a\Cotal\Cotal\node_modules\.bin\tsx ENOENT
```
It runs in the manager `test` target on the Windows `required` job. **Addressed at `2fc985c`:** NO. `git diff 647e21b 2fc985c -- implementations/manager/smoke/legacy-pty-custody.smoke.ts` is empty; line 50 is byte-identical. The Windows defect is still present at the current head.

### What is clean and honest (verified here, not inherited)

- **`pnpm typecheck`**: clean, `Scope: 27 of 28 workspace projects`. The 28th is JS-only `@cotal-ai/example-04-frontier-faces` (no typecheck script) — reported as a skip, not called clean.
- **Suite hashes re-derived independently** (`sha256("smoke:<name>")`) and all four match the committed fragments: `legacy-pty-custody` → `ea3440f…`, `legacy-manager-report` → `0c8285f…`, `legacy-packaged-manager` → `71bffcf…`, `runtime-contract` → `8df244e…`. All four `smoke:<name>` scripts exist in `package.json` and map to the right files.
- **All four suites PASS locally on this host.** `legacy-pty-custody` confirms the M1 residual (child DIES, `hasReference === false`); `runtime-contract`, `legacy-manager-report`, and `legacy-packaged-manager` pass.
- **Isolation fix is genuine (the blocker I cared most about).** `legacy-packaged-manager` uses `mkdtempSync`, fresh HOME/XDG_CONFIG_HOME/TMPDIR, strips `COTAL_*` from parent and child env, a PATH fixture `cotal` that exits 97 if invoked, and snapshots the real operator stamp as raw bytes. I ran it twice: once on this host (a real operator stamp present — byte-identical before and after), and once with the outer HOME/XDG pointed at an empty dir so `stampBefore === undefined`. Both pass. It does **not** silently depend on `stampBefore` being defined; the clean-machine path (`undefined === undefined`) holds. Absence of the old host pin is real portability, not a hidden dependency.
- **Mutation fixtures graded by reading** (I did not run `mutation-proof`). Every `find` anchor resolves EXACTLY ONCE at head (`hasReference: handle.reference !== undefined`; `state(ids.childPid) === "gone" || … === "Z"`; `if (await rt.reportRunningManager(target) === "legacy") return finish(false);`). Each `expectRed`/`cell` string is present. For `legacy-manager-update`, the mutant (`"legacy"` → `"none"`) makes the default non-legacy path abort, reddening `assert.equal(await executeUpdate(false, rt), 0, …)` at `command-kernel.smoke.ts:264` — before the later legacy block, so the named cell is the earliest reddening point in this fail-fast (`assert`) suite.
- **M1 is honestly framed.** `Refs #783`, not `Closes`. Title and body do not overstate: the body states this is the M1 seam, not rolling PTY handoff, and that the child DIES on worker death (child-PID survival not done). That matches what the suite asserts and what I observed running it.

### The core seam needs no `SPEC.md` change — reasoning, on the record

`Runtime.adopt` / `RuntimeReference` / `supportsFreshStart` / `sessionContinuityClass` are additions to **local, non-wire** contracts. `SPEC.md` never describes `Runtime`, `RuntimeProvider`, `AgentHandle`, or `LaunchSpec`; the spec even states that a runtime layer over the control surface is "not part of the endpoint contract." The code documents `adopt` as "a local runtime operation, not a mesh operation," and `RuntimeReference` is `{ kind, id }` — genuinely opaque, with `runtime-contract.smoke.ts` asserting the `RuntimeReference…RuntimeProvider` slice of `runtime.ts` exposes no `Pty`/`IPty` type. A non-PTY runtime (tmux/cmux/orca/herdr) can honor the contract by returning its own session/pane id as `id`. Nothing here touches subjects, schemas, presence, delivery semantics, or the auth grammar, so **no `SPEC.md` change is warranted, and its absence is correct rather than a finding.**

### Soft notes (non-blocking)

- **Untyped refusal.** Every runtime's `adopt` throws a bare `new Error(...)`, not a typed, actionable error class a caller can branch on. No user-reachable crash exists today (the manager never calls `adopt` on a reachable path), so this is a shape note, not a block.
- **No path forward on the legacy-update refusal.** `cotal update` correctly refuses replay against a legacy manager and exits non-zero (reachability confirmed via `legacy-packaged-manager`: a real published legacy manager receives the report and is left alive). But neither the message ("this update is not a hot update") nor `docs/cli.md` tells an operator who legitimately needs to update how to proceed (e.g. stop the manager, then update). The built-in connector reconcile has already run by then; only the operator-extension replay is blocked.
- **`custody: "legacy"` is a single-value enum.** The `ManagerStatus.custody` field can only hold `"legacy"` today. It reads as a placeholder, but the report's per-seat `continuity` carries four real values (`exact`/`fork`/`fresh`/`drain-only`) and the report's guard (`custody !== undefined && custody !== "legacy" → undefined`) is forward-looking, so the seam is acceptable as-is.

### Open question carried to the next head (not a ruling)

`Runtime.adopt` is declared **required** on the `Runtime` interface, which forces every out-of-tree runtime extension to add a throwing stub merely to keep compiling, for a seam nothing can satisfy yet. Given that `AGENTS.md` has extensions peer-depend core and says to guard the core, whether an **optional** method plus a manager-side "unimplemented → refuse" check is the honest shape is worth deciding at the next head. Stated as a question.

---

**Recommendation:** do not merge until a CI run exists where all four gated custody suites actually execute and pass, and until the Windows `legacy-pty-custody` `tsx` spawn (`ENOENT`, line 50) is fixed — it is still present at `2fc985c`. Findings 1 and 2 are resolved at `2fc985c`. I inherited nothing here beyond the PR's own description; every item above I verified myself at `647e21b`.

davidfarah2003:
## Independent review — head `71974bb61fb1ca50ffd4da854d4549d3a707fccc` (M1 custody seam)

Follow-up to my findings at `647e21b5e` (that comment stands as the record for that sha). This states plainly what I **re-verified at `71974bb61`** versus what I carry forward from my own earlier pass. Head confirmed on both sides; worktree checked out clean at this sha. Scope vs main: **33 files, +541/-15** (`merge-base..head` and `main...head` agree).

### Verdict: the merge gate is MET — all four gated suites executed and passed in CI

This is the fact that mattered and had never held before. In CI run `33974805370` at this head, I read each shard's log and confirmed each suite **ran** (its `===== pnpm smoke:<name> =====` line is present) and **passed** (its own success banner, not a bare exit), in a shard that reached it:

| Suite | Shard | Log evidence |
|---|---|---|
| `smoke:legacy-manager-report` | 0 | `legacy manager report smoke OK` |
| `smoke:runtime-contract` | 0 | `runtime contract smoke OK` |
| `smoke:legacy-pty-custody` | 1 | `✓ M1 red: legacy manager death ends the manager-owned counter PTY` → `LEGACY PTY CUSTODY OK (5 passed, 0 failed)` |
| `smoke:legacy-packaged-manager` | 3 | `legacy packaged manager smoke OK` |

Each shard closed with `✓ smoke:ci shard N/4 passed`. Shard placement is `fragmentShard = sha256(name).readUInt32BE(0) % 4` → 0/0/1/3; shard 2 holds none of the four, which is why the two previous heads (both fail-fast on shard 2) never executed any of them. `CI` overall: **success** (unit, live, all four smoke shards, ci-ok). `Windows`, `Docs`, `CodeQL`: success. `Mutation reproof`: still in_progress → **unmeasured**, not green.

### My three prior blockers — all fixed here, re-verified at this sha

1. `smoke:flag-inventory` golden — **fixed**: now `update: { flags: [...TARGET, "self:boolean"] }`, and `unit` (which runs `flag-inventory`) is green.
2. docs bundle drift — **fixed**: `docs-bundle.generated.ts` regenerated; it carries `cotal update [--self] [--space <s>] [--server <url>] [--creds <path>]` and the continuity text; `check:docsbundle` (in `unit`) is green.
3. Windows `legacy-pty-custody` `tsx` ENOENT — **fixed**: the suite now runs `owner.mjs` directly with `spawn(process.execPath, [owner], …)`; no `.bin/tsx` reference remains; `Windows / required` is green. (The fix rides a platform skip — see below.)

### Re-verified locally at this head (not inherited)

- `pnpm typecheck` clean, `Scope: 27 of 28` (28th JS-only `@cotal-ai/example-04-frontier-faces`, skipped).
- All four suites **pass locally** on this Linux host, including `legacy-pty-custody` 5/5 (child DIES, `hasReference === false`) and `legacy-packaged-manager` with the operator seed stamp byte-identical before and after.
- **Poison re-derived myself.** The `until(() => false)` that briefly landed in `342f2bba6` (an ancestor of this head) has **0 occurrences tree-wide at the head**; positive control `until(` hits **46 files**, so the search works. Clean at the head. Worth one line because `342f2bba6` stays in history and would poison a bisect or cherry-pick through it.
- Mutation anchors for `legacy-pty-custody.json` and `legacy-manager-update.json` each resolve **exactly once** at this head; `expectRed` cells present; the fixture find/replace pairs are intact.

### Finding — the M1 custody claim is Linux-only evidence, and the PR text does not say so

`legacy-pty-custody.smoke.ts:6-8` skips the **whole suite** off Linux (`process.platform !== "linux" → console.log(…skipped…); process.exit(0)`). The stated reason is honest and specific — the M1 residual is `/proc/<pid>/stat` state after `SIGKILL`, which has no Windows oracle — and it prints rather than exiting silently. But the consequence is that **the child-dies-with-the-manager behaviour is verified only on Linux**, and the PR body still reads platform-neutral: "the counter child **dies** when the manager worker is killed" / "`smoke:legacy-pty-custody` (child dies)" with no Linux-only qualifier. **The PR body and merge note should state the custody evidence is Linux-only.** This is the part I'd fix before merge.

**This is systemic, not local to this suite.** The shard runner (`bin/smoke/shard.mjs`) decides pass/fail on `r.status !== 0` alone — there is no required completion banner, no asserted-cell count, nothing that distinguishes "ran and passed" from "exited 0 without asserting." So on the Windows lanes an immediate `process.exit(0)` is counted as a covered pass (`pnpm test` on `Windows / required` reads this suite as covered though it asserted nothing; the Windows *sharded* smoke is non-blocking). `legacy-pty-custody` is simply the first fragment-gated suite to carry a platform skip and expose the gap — the same shape as the #1098 hole where "no failures" read as "checks passed" when nothing ran. **The merge gate itself is not weakened**: the gate is the Linux `CI` shards, where `process.platform === "linux"` and the suite runs fully (confirmed above). Flagging the systemic observation for a separate issue; not scope for this PR.

*Follow-up (not a blocker):* the skip is broader than the mechanism requires. A portable liveness probe (`process.kill(pid, 0)`, the shape `_orphan-seat-host.ts:15` already uses, which `state()` now even falls back to) would let the child-death assertion run on macOS; only the `/proc` `Z`/residual detail is truly Linux-bound. So the suite *could* verify custody on macOS with a narrower skip. Windows remains unverified for that variant, which is why nobody pushed it. Low marginal value since `runtime-contract` already covers the `reference` contract platform-neutrally.

### Open questions carried in (re-verified at this head; stated as questions, not blocks)

**The JSDoc and the type contradict each other.** In `packages/core/src/runtime.ts`, the prose on `adopt` says runtimes without durable custody "leave it unimplemented and must refuse," but the method is declared **required** (`adopt(reference: RuntimeReference): AgentHandle;`, no `?`) — so no implementation can leave it out and compile. The `reference?` field right above it IS optional, so the file uses both shapes and only one matches its own description. **At minimum one of the two is wrong as written**, and that inconsistency is itself a small finding worth correcting even if the required shape is kept.

**Required-vs-optional `adopt`.** The cost is measured, not hypothetical: this PR added an identical throwing stub to four in-tree runtimes (`cmux`, `tmux`, `orca`, `herdr`), and the same obligation falls on every out-of-tree runtime extension on upgrade — a compile-breaking change for extension authors, not a purely additive seam. An **optional** method plus a manager-side "unimplemented → refuse" would match the prose and cost extension authors nothing. The counter-argument is real and, in this repo, weighty: AGENTS.md's rule is "throw if unsupported, never silently degrade," and a required method forces every runtime author to consciously declare a custody stance at compile time rather than inherit a silent default; optional moves the refusal from compile time to run time. **My lean:** keep it a follow-up, not a block — the required shape is defensible under the repo's no-silent-degrade rule, and the manager never reaches `adopt` on a user path today, so nothing is broken. But the JSDoc must be corrected to stop claiming "unimplemented" is possible, and a typed refusal (below) would make an eventual optional-shape migration cleaner.

**Bare `Error` refusals.** All five runtimes throw `new Error(...)`, not a typed, actionable error a caller can branch on. Independent of required-vs-optional. Non-blocking today (no reachable caller), but a `RuntimeAdoptionUnsupportedError` (or an `unsupported`-tagged error) is the honest shape for a refusal that is a designed, permanent state rather than an incidental failure.

### Unchanged conclusions from `647e21b5e` (re-checked at this head)

- **No `SPEC.md` change needed** — re-confirmed `SPEC.md` is unchanged at this head; `Runtime`/`RuntimeReference`/`adopt` are a **local, non-wire** contract (documented as "a local runtime operation, not a mesh operation"), so nothing touches subjects, schemas, presence, delivery, or the auth grammar. Its absence is correct, not a finding. `RuntimeReference = { kind, id }` is genuinely opaque; `runtime-contract.smoke.ts` guards that the contract slice exposes no `Pty`/`IPty` type; a non-PTY runtime can honor it with its own id.
- **M1 framing honest** — `Refs #783`, not `Closes`; the body states this is the M1 seam, not rolling handoff, and that child-PID survival is not done. Matches what the suite asserts and what I observed. (The one gap is the Linux-only-evidence omission above.)
- **Changeset** — `minor` across 12 named packages is correct: 0.x makes even a breaking change a minor, and the `fixed` group versions all packages in lockstep regardless of which are named. The required-`adopt` break is real for out-of-tree runtimes but is a minor under 0.x. No finding.
- Soft notes still stand: the legacy-update refusal offers no explicit path-forward in its message/docs; `custody` is a single-value enum today (acceptable as a forward-looking seam).

### Recommendation

The merge gate I named is now satisfied: all four custody suites executed and passed in CI, and the implementation, typecheck, and cross-platform CI are green. The one thing I would require before merge is a **one-line honesty fix**: state in the PR body / merge note that the M1 custody (child-dies) evidence is **Linux-only** (the suite skips off Linux). The JSDoc-vs-type contradiction on `adopt` is a small correctness fix worth doing here; required-vs-optional `adopt`, typed refusals, the systemic "exit(0) = covered" gap, and the macOS-variant skip are legitimate follow-ups, not blockers. I inherited nothing beyond my own prior pass; every item above I re-checked at `71974bb61`.

davidfarah2003:
BLOCK at `71974bb61fb1ca50ffd4da854d4549d3a707fccc`.

I reviewed the whole 33-file, +541/-15 change against main. I did not treat the later remediation as the review scope.

## Blocking findings

1. **`Runtime.adopt` is required even though the contract says unsupported runtimes leave it unimplemented.** `packages/core/src/runtime.ts:89-93` says runtimes without durable custody leave adoption unimplemented, but the type requires `adopt`. `AgentHandle.reference` is correctly optional immediately above. Every current runtime therefore adds a throwing stub, and every out-of-tree peer-dependent runtime must do the same merely to keep compiling. This is a breaking adapter burden in the guarded core before any implementation or caller exists. Make capability absence representable, for example with optional `adopt`, and require callers to refuse when it is absent.

2. **`cotal update --self` can mutate before the legacy-manager report and can bypass the selected target.** `executeUpdate` takes the newer-version branch at lines 111-114 and globally installs before `reconcileCurrent` can call `reportRunningManager`. The replacement child is then launched as only `[entry, "update"]` at line 309, dropping `--space`, `--server`, and `--creds`. A safe injected probe produced `install` then child `update`, with no report call and no selected-target arguments. If the default target has no manager, extension replay may proceed while the explicitly selected legacy manager was never checked. The packaged and kernel smokes cover only the non-self path. The continuity check and target must survive the self-update handoff before either global install or extension replay is claimed safe.

3. **The legacy-update mutation fixture proves the wrong path.** `legacy-manager-update.json` replaces `=== "legacy"` with `=== "none"` and expects the earlier normal-update cell, `update replays first-party extensions...`, to fail. That cell runs before the injected legacy cell and returns `none`, so the mutant reddens before it ever exercises legacy fallthrough. The note claiming the legacy injection observes the forbidden mutation edge is false. Use a mutation that lets only the legacy verdict fall through, then name the legacy cell's first forbidden mutation/replay assertion.

## Numbered review

1. `RuntimeReference` itself is clean and backend-neutral. `{ kind, id }` does not assume PTYs or FDs, and tmux/cmux/orca/herdr can represent provider-owned resources honestly. The required `adopt` method is the core-contract defect described above.
2. I withdraw bare `Error` as a separate blocker. No shipped caller invokes `Runtime.adopt`, so a shared typed error would not currently make a caller more actionable. Capability absence should first be modeled by optional `adopt`; typed bad-reference/refusal errors can be added with the first real caller. I found no user-reachable unhandled `adopt` crash at this head.
3. No `SPEC.md` change is required. This is explicitly a machine-local runtime extension API and not a wire subject, envelope, or delivery guarantee. The manager service's content-addressed command schema remains implementation-specific over the existing endpoint wire contract.
4. `custody: "legacy"` is a useful forward discriminator, not merely a placeholder. It distinguishes an explicit M1 declaration from older managers with no field, while both safely classify as legacy. The seat continuity values are derived from declared connector capabilities and unknown connectors fail closed to `drain-only`.
5. The non-self refusal is reachable and correct: it reports first, exits 1 before the extension mutation lock or replay subprocess, and leaves the legacy manager and child alive. It does not give the operator an actionable completion path beyond the documentation's statement that custody is unavailable. More importantly, the self-update path bypasses the guarantee as described in blocking finding 2.
6. I independently re-derived the four SHA-256 fragment names and shard placement. Exact-head CI logs prove execution and success: shard 0 ran `legacy-manager-report` and `runtime-contract`; shard 1 ran `gate-inventory` and `legacy-pty-custody` with 5/5 checks; shard 3 ran `legacy-packaged-manager`; all three shards completed successfully. `smoke:gate-inventory` reported 574 scripts, 548 reached, 26 declared exemptions. This merge gate is satisfied.
7. I ran `legacy-packaged-manager` from disk-backed scratch with fresh HOME, XDG_CONFIG_HOME, and TMPDIR, so `stampBefore` was genuinely undefined. It passed and removed its scratch. The direct suite requires a prior workspace build; before `pnpm build`, core/workspace `dist` imports were missing. After build it passed. That prerequisite is not stated by the script name itself.
8. Every mutation `find` anchor resolves exactly once at this head and every `expectRed` is text. Both PTY mutations name their earliest relevant cells. The legacy-update fixture fails the earliest-point requirement as described in finding 3. I did not run `mutation-proof`, per the brief. I confirmed `until(() => false)` has zero head-tree occurrences with a positive control; commit `342f2bba670887cbfd95823704b094340be86045` introduced it and the head commit restored the predicate.

## M1 and platform scope

The PR and changeset accurately describe M1 only, use `Refs #783`, and state that rolling PTY handoff is absent. The exact-head Linux suite confirms the child dies after manager death. The whole-suite non-Linux skip at `legacy-pty-custody.smoke.ts:6-8` is broader than necessary: fixture launch, live-child, no-reference, and manager-death checks can use portable process liveness, with only Linux zombie-state evidence gated if necessary. With the current skip, the M1 custody evidence is Linux-only and the `process.execPath` launch fix is unreachable on Windows. I treat narrowing this as follow-up, not an additional blocker, provided the merge note states the Linux-only evidence.

## Verification

- Installed dependencies with `pnpm install --frozen-lockfile`.
- `pnpm typecheck` passed for 27 of 28 workspace projects; JS-only `@cotal-ai/example-04-frontier-faces` has no typecheck script.
- `pnpm build`, `smoke:runtime-contract`, `smoke:legacy-manager-report`, `smoke:legacy-pty-custody`, `smoke:legacy-packaged-manager`, `smoke:cli-kernel`, and `smoke:gate-inventory` passed locally.
- Full-SHA workflow enumeration: CI, Windows, Docs, and Code Quality completed successfully. Mutation reproof was still in progress when this review was posted, so I do not count it green.
- I could not read `.internal` because the pinned submodule commit is no longer available from its remote. No live suite, `pnpm check`, dev-tree CLI, or mutation-proof command was run.

**Inherited versus verified:** I inherited only the review brief, the prior-head hold context, and reports of the earlier CI failures. I re-fetched this SHA, re-measured scope, re-read the full changed surface, re-derived hashes and shards, reran all required local suites, inspected the exact-head CI logs, and formed these findings independently. The worktree was clean immediately before posting.

davidfarah2003:
## Merge hold at `71974bb61fb1ca50ffd4da854d4549d3a707fccc`

Two independent exact-head reviews have now reported at this sha. One is a BLOCK on three
findings; the other clears the change but requires one honesty correction to the merge scope.
Under the standing rule that a head is clear only when every reviewer clears it, **this head does
not merge.** Recording that here so it is durable rather than living in a review thread.

The CI merge gate is separately satisfied and is not the reason for the hold: all four custody
suites were confirmed to have actually executed and passed in shards that reached them, at this
exact sha. A gated suite that never runs was the defect behind #1278, so that check was the point;
it passed.

### What must change before this is re-reviewed

1. **`Runtime.adopt` is required while its own JSDoc says unsupported runtimes leave it
   unimplemented** (`packages/core/src/runtime.ts:89-93`). `AgentHandle.reference` directly above it
   is correctly optional. As written, every in-tree runtime carries a throwing stub and every
   out-of-tree peer-dependent runtime must add one merely to keep compiling — a breaking adapter
   burden in the guarded core before any caller exists. The prose and the type contradict each
   other, so one of them is wrong as written and it should not ship in that state.

2. **`cotal update --self` installs before the continuity report, and drops the selected target.**
   I verified both halves at this head independently of the review. In `executeUpdate`, the
   newer-version branch under `if (self)` returns `upgradeAndReconcile(...)` directly;
   `reportRunningManager` is reached only from inside `reconcileCurrent`, which that branch never
   enters. So the global install happens with the legacy-manager check skipped entirely. Separately,
   the replacement child is launched as `[entry, "update"]`, without `--space`, `--server` or
   `--creds` — while `rt.selfArgv()`, the mechanism that preserves argv, already exists in this file
   and is used on the other path. An operator who names a legacy manager explicitly can therefore
   have it never checked. This is the PR's own new guarantee having a hole in it, not a pre-existing
   gap the PR merely fails to cover.

3. **The `legacy-manager-update` mutation fixture proves the wrong path.** It flips `=== "legacy"`
   to `=== "none"`, which reddens an earlier normal-update cell that returns before legacy
   fallthrough is ever exercised. The mutant dies, and the guarantee under test is untouched.
   AGENTS.md states the distinction directly: a killed mutation shows the test depends on that code,
   not that a real entry point reaches it. The fix is a different mutation, not a different cell
   name.

### Scope note the merge must carry when it eventually lands

This is M1 only and correctly says `Refs #783` rather than `Closes`. The rolling PTY handoff is
absent, and `smoke:legacy-pty-custody` asserts the child **dies**. The suite also skips wholesale
off Linux, so **the M1 custody evidence is Linux-only** and the `process.execPath` launch fix is
unreachable on Windows. Narrowing that skip is follow-up rather than a blocker, but the sentence
belongs in the PR body before merge, not only in a review thread.

Push the fixes as new commits on top of `71974bb61`; please do not rebase. Both reviewers are held
and will be released onto the new sha.

davidfarah2003:
APPROVE at `d1db29e237fdd6c2f01ae3a163bbdc387fee6770`.

This is a delta review from my BLOCK at `71974bb61fb1ca50ffd4da854d4549d3a707fccc`. The prior head is an ancestor. I reviewed the two remediation commits, 19 files, +147/-60. My prior review continues to bind the unchanged base.

## Prior blockers

1. **Resolved: optional runtime adoption.** `Runtime.adopt?` now represents capability absence honestly, its JSDoc agrees, and the five forced throwing stubs are gone. `requireRuntimeAdopt` refuses absent capability by runtime name and delegates when present. `smoke:attach` executes all three relevant paths, including the real `createRuntime("pty", ...)`, and passed locally. The source-text assertion in `runtime-contract` is redundant, not load-bearing.
2. **Resolved: self-update continuity and target handoff.** The selected manager is classified before npm global-root resolution, locking, or install. A legacy verdict returns with no global mutation. The replacement child receives `--space`, `--server`, and `--creds`. An injected no-side-effect probe observed report first, then install and exact target argv for a `none` verdict; the legacy probe observed only the report and exit 1.
3. **Resolved: legacy fallthrough mutation.** The mutation now replaces the legacy stop with a bare awaited report. A `none` verdict still follows the normal path, while only a `legacy` verdict newly falls through. In the legacy cell, reconcile and report assertions remain green; the next assertion, `legacy report refuses before extension replay can mutate packages`, is the first red because `claimUpdatePass` was reached. The anchor resolves exactly once and `expectRed` is text.

The two new self-update mutation anchors also resolve exactly once. Their named cells are the first direct observations of missing report-before-install and missing target argv. I did not run `mutation-proof`, as required. The exact-head Mutation reproof workflow remains in progress and is unmeasured; #1299 tracks its repeat 60-minute timeout on high-fan-in changes.

## Exact-head verification

- `pnpm typecheck` passed, 27 of 28 typed projects; the JS-only Frontier Faces example has no typecheck script.
- `smoke:runtime-contract`, `smoke:attach`, `smoke:cli-kernel`, `smoke:legacy-packaged-manager`, `smoke:legacy-pty-custody`, and `smoke:gate-inventory` passed locally. The first packaged run overlapped a workspace build and failed on a transient missing connector-jcode dist artifact; the isolated rerun after the build passed.
- Exact-head CI logs prove all four gated custody suites executed and passed: shard 0 ran `legacy-manager-report` and `runtime-contract`; shard 1 ran `gate-inventory` and `legacy-pty-custody` with 5/5 checks; shard 3 ran `legacy-packaged-manager`. All four CI shards, unit, live, and the aggregate `ci-ok` job completed successfully. Windows, Docs, and Code Quality are green.
- The PR accurately states the Linux-only M1 custody evidence and links #1298 for narrowing the skip.

## Nonblocking wording notes

The PR body says `requireRuntimeAdopt` has no shipped caller, but the shipped `Manager.adoptRuntimeHandle` method delegates to it. More precisely, no current control/update flow calls `adoptRuntimeHandle`. The package-list sentence also omits `@cotal-ai/connector-core`, though the changeset includes it. Neither affects behavior or the merge verdict.

Local and remote heads matched immediately before posting, and the worktree was clean.

davidfarah2003:
## Independent review — APPROVE at head `d1db29e237fdd6c2f01ae3a163bbdc387fee6770`

Third pass, grading the delta on top of `71974bb61` (my prior comment records that sha; the one before it records `647e21b5e`). Head confirmed on both sides; `71974bb61` is an ancestor (no rebase), so my prior reads bind below and this is a **delta review**: 2 commits, 19 files, +147/-60. Worktree checked out clean at this sha.

### Merge gate RE-CONFIRMED at this sha (not carried from `71974bb61`)

The delta changed `runtime-contract.smoke.ts` directly, so the prior run's shard evidence does not transfer. In CI run `33978243977` **at `d1db29e23`**, I re-read all four shard logs and confirmed each suite ran (its `===== pnpm smoke:<name> =====` line) and passed (its own banner), in a shard that reached it:

| Suite | Shard | Fresh log evidence at `d1db29e23` |
|---|---|---|
| `smoke:legacy-manager-report` | 0 | `legacy manager report smoke OK` |
| `smoke:runtime-contract` | 0 | `runtime contract smoke OK` |
| `smoke:legacy-pty-custody` | 1 | `✓ M1 red: legacy manager death ends the manager-owned counter PTY` → `LEGACY PTY CUSTODY OK (5 passed, 0 failed)` |
| `smoke:legacy-packaged-manager` | 3 | `legacy packaged manager smoke OK` |

Each shard closed `✓ smoke:ci shard N/4 passed`. On my own direct read of run `33978243977` at 17:11Z, the `CI` workflow is **completed/success** with every job green — `unit`, `live`, all four smoke shards, and the `ci-ok` aggregate (this last only settled a moment before my read; the shards concluding is not the workflow concluding, and I waited for `ci-ok` rather than infer a green whole from green parts). `Windows`, `Docs`, `CodeQL`: success. `Mutation reproof`: still in_progress and heading for its fifth 60-minute-timeout cancellation on this branch — filed as **#1299**, **unmeasured**, not a gate. See the note on that below.

### My one requirement — DISCHARGED

The PR body now states: "The M1 custody evidence is Linux-only. legacy-pty-custody skips the whole suite off Linux because the residual it gathers is /proc/<pid>/stat after SIGKILL. The process.execPath launch fix is therefore unreachable on Windows." That says the evidence is **absent off Linux**, not merely that a platform differs — which is what I asked for. It also correctly points the narrow-the-skip follow-up at #1298 and distinguishes it from the shard-runner gap at #1297.

### The three blockers — each fixed, verified at this sha

**1. `Runtime.adopt` required-vs-optional + the JSDoc/type contradiction.** `packages/core/src/runtime.ts` now declares `adopt?(reference: RuntimeReference): AgentHandle;` (**optional**), and the JSDoc is rewritten to agree: "OPTIONAL, and absent means REFUSE, never spawn a replacement... the caller must throw naming that runtime." The prose and type now match — the contradiction I flagged is resolved by making the type match the prose. All five throwing stubs (`cmux`, `tmux`, `orca`, `herdr`, `pty`) are deleted, so out-of-tree runtimes no longer need a ceremonial stub to compile. The refusal moved to a named manager-side check: `requireRuntimeAdopt(runtime, reference)` (`implementations/manager/src/runtime/index.ts:9`) throws `runtime "${runtime.kind}" does not support adopt` when `adopt` is absent, reached through `Manager.adoptRuntimeHandle` (`manager.ts:1046`). `runtime-contract.smoke.ts` now asserts the optional shape, that a runtime may omit `adopt`, and that the manager refuses by name. This is the honest shape — capability absence is representable, and refusal is loud and named.

**2. `cotal update --self` reported after install and dropped the target flags.** Fixed on both halves. `upgradeAndReconcile` now calls `reportRunningManager(target)` **before** the npm global install and returns 1 if a legacy manager is running, so the continuity report gates the upgrade path just as it gates the plain path. And the replacement child is launched as `[entry, "update", ...targetArgv(target)]`, threading `--space`/`--server`/`--creds` through, so the reconcile child targets the manager the operator selected instead of resolving blind. New fixture `self-update-handoff.json` proves both (report-before-install order; target-flag handoff), and `command-kernel.smoke.ts` exercises the path directly: it asserts the report index precedes the install event, that the child argv carries the named target, and that a legacy verdict on `--self` refuses the global install.

**3. `legacy-manager-update.json` killed its mutant on an earlier normal-update cell.** Fixed. The mutation `replace` changed from flipping `=== "legacy"` to `=== "none"` (which reddened the normal path before legacy fallthrough was ever exercised) to `await rt.reportRunningManager(target);` — dropping the guard so a `none` verdict still continues but a `legacy` verdict **falls through** to `claimUpdatePass` and the replay loop. The cell is re-anchored to `a running legacy manager is reported, then blocks an update before replay` — the actual guarantee. The mutant now reddens on the legacy-block assertion, so the guarantee is genuinely proven.

### Re-verified locally at this head (not inherited)

- `pnpm typecheck` clean, `Scope: 27 of 28` (28th JS-only, skipped).
- All four suites **pass locally**: `runtime-contract` (with the new optional-adopt assertions), `cli-kernel` (`✓ command-kernel smoke passed`, covering the new self-update cells), `legacy-manager-report`, `legacy-pty-custody` 5/5 (child DIES, no reference), `legacy-packaged-manager` (operator stamp byte-identical before/after).
- Every mutation anchor at this head — the two `self-update-handoff` finds (including the multi-line block), the re-anchored `legacy-manager-update` find, and the unchanged `legacy-pty-custody` finds — resolves **exactly once**; each `expectRed` cell exists in its suite.
- Poison re-check with a corrected, arg-count-agnostic pattern: `until(() => false` (any args) tree-wide has one hit, `transport-liveness-broker.smoke.ts` `until(() => false, 0)`, a bounded legitimate idiom in an unrelated suite; the custody suite's `until()` calls are all bounded. The `342f2bba6` poison is history-only.

### One caveat carried into the verdict — the mutation fixtures are not graded by CI

This connects to my systemic finding (#1297) one layer up, and I want it on the record so the approval is not misread. `Mutation reproof` has been cancelled at the 60-minute timeout on all four branch shas (this PR's set is 33 paths / 42 fixtures / 181 mutations, `manager.ts` being high fan-in), filed as #1299. Because `pr-head-gate.mjs` grades any non-success as failing, the gate fails closed rather than open — but the practical consequence is that **this PR's mutation fixtures are never actually graded by CI**: neither red nor green, just unmeasured. So the blocker-3 fixture rewrite and the new blocker-2 fixture rest entirely on local runs and on review. I read both fixtures and consider them sound (anchors resolve once, cells re-anchored on the real guarantees), but I cannot cite a green mutation-reproof for them, and neither should anyone else.

### Verdict

**APPROVE** at `d1db29e237fdd6c2f01ae3a163bbdc387fee6770`. All three blockers are fixed, my Linux-only honesty requirement is discharged in the PR body, CI is green at this exact sha with all four gated custody suites re-confirmed executing and passing from fresh shard logs, typecheck is clean, and all four suites pass locally. The open items are follow-ups with filed issues, not blockers: narrowing the platform skip (#1298), the shard-runner zero-assertion gap (#1297), and the mutation-reproof budget (#1299). One reviewer approving does not clear a head on its own; this clears my line. Everything above I re-checked at this sha; I inherited nothing beyond my own prior passes.

davidfarah2003:
## Mutation evidence for the two fixtures this PR changed, recorded post-merge

The `changed` job of Mutation reproof never completed on this branch — five consecutive cancellations
at the 60-minute cap, the last at 60m15s with a single run at the merged head, so the cancellation is
a timeout rather than a supersession. Cause and remedy are tracked in #1299. **CI therefore never
graded the two mutation fixtures this PR altered**, and one of them was the subject of a blocking
review finding, so the evidence belongs here rather than only in the reviewers' reading.

Run locally at the merged head `d1db29e237fdd6c2f01ae3a163bbdc387fee6770`, working tree clean before
and after.

**`implementations/cli/smoke/mutations/self-update-handoff.json`** — `pnpm smoke:cli-kernel`, 49.259 s,
exit 0.

| mutant | named cell (`expectRed`) | verdict |
|---|---|---|
| report the running manager only after the global install | `self-update reports the running manager before the global install` | KILLED, red on the named cell, 0 marks against baseline 1 |
| drop the selected target flags from the replacement child | `self-update replacement child carries the selected target flags` | KILLED, red on the named cell, 0 marks against baseline 1 |

**`implementations/cli/smoke/mutations/legacy-manager-update.json`** — `pnpm smoke:cli-kernel`,
27.750 s, exit 0.

| mutant | named cell (`expectRed`) | verdict |
|---|---|---|
| ignore the legacy verdict and continue the update replay | `legacy report refuses before extension replay can mutate packages` | KILLED, red on the named cell, 0 marks against baseline 1 |

Both configs died **on the cell each fixture names**, not on an earlier unrelated red. That
distinction is the whole point of the review finding that produced this fixture: the previous version
flipped `=== "legacy"` to `=== "none"`, which reddened an earlier normal-update cell that returned
before legacy fallthrough was ever exercised. The mutant died and the guarantee under test was never
reached. `AGENTS.md` states the principle directly — a killed mutation shows the test depends on that
code, not that a real entry point reaches it — so a fixture that kills its mutant in the wrong place
proves nothing.

Reporting marks against baseline rather than only the pass/fail word is deliberate: a display reading
zero marks looks identical whether the completion marker printed or never did, so the baseline is the
comparison that makes it evidence.

Combined wall time for the two configs is 77.111 s. That is these fixtures only and is not a
measurement of the full changed-set sweep, which is what exceeds the job budget.

PR #1295, opened 2026-09-05
ci(mutation-reproof): file a tracking issue when the full sweep fails

### LINKED ISSUE #1292: mutation-reproof: notify/assign on scheduled full-sweep failure (needs workflow scope)
## Summary

The scheduled `full` job in `.github/workflows/mutation-reproof.yml` (added by #1272, closing #1217) re-proves the entire mutation corpus daily. When it fails, nothing notifies anyone, assigns the failure, or files an issue: the red run is visible only to someone who opens the Actions tab. A scheduled gate whose failures nobody sees is a gate that can be red for a long time while everyone believes it is protecting them.

This issue tracks adding a triage-notification step so a failed sweep becomes an owned, tracked item.

## Mechanism already written

A working implementation exists as a single commit on `fix/1217-mutation-reproof`'s lane (commit `6f85f8ff6`), not yet on the branch. It adds a final step to the `full` job that, on failure, opens a deduplicated tracking issue (or comments on the open one if a marker issue already exists), with `issues: write` scoped to that job:

```yaml
- name: File a triage issue when the sweep fails
  if: failure()
  env:
    GH_TOKEN: ${{ github.token }}
    RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
  run: |
    set -euo pipefail
    marker="mutation-reproof: full sweep failing"
    existing="$(gh issue list --state open --search "in:title \"$marker\"" --json number --jq '.[0].number' || true)"
    body="The daily mutation-reproof full sweep failed. ... Run: $RUN_URL"
    if [ -n "$existing" ]; then gh issue comment "$existing" --body "$body"; else gh issue create --title "$marker" --body "$body"; fi
```

## Why it is not in #1272

Pushing a commit that touches a file under `.github/workflows/` requires a credential with the `workflow` OAuth scope. The credentials available on the lane that produced #1272 do not have it, and GitHub enforces this server-side, so the commit cannot be pushed from there. Rather than carry an unenforceable claim in the PR, #1272 states plainly that the scheduled sweep's failures are visible only in the Actions tab, and defers the notification mechanism to this issue.

## Ask

Land commit `6f85f8ff6` (or an equivalent) from a credential with `workflow` scope. It is a single, self-contained step; it does not need to be rewritten.

Refs #1272, #1217.

davidfarah2003:
Closes #1292.

The scheduled full sweep in `mutation-reproof.yml` runs on no pull request, so a red run was visible only in the Actions tab with no owner. On failure the `full` job now opens a deduplicated tracking issue titled `mutation-reproof: full sweep failing`, or comments on the open one, with `issues: write` scoped to that job only.

This is commit 6f85f8ff6 from the lane that produced #1272, landed unchanged (same tree, same author) from a credential that can push workflow files. The lane's credential lacked the `workflow` OAuth scope, which GitHub enforces server-side, so #1272 deferred this step to #1292.

Verification: the change is confined to `.github/workflows/mutation-reproof.yml` (+23). The step is gated on `if: failure()` in the `full` job, so it cannot fire on pull-request runs of the `changed` job. The dedup search (`gh issue list --state open --search 'in:title "mutation-reproof: full sweep failing"'`) run against the live repo returns no open marker issue today, so the first failing sweep would create one.

### DIFF
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -97,23 +97,29 @@ jobs:
         run: pnpm build
       - name: Re-prove every mutation fixture
         run: node scripts/mutation-reproof.mjs --all
-      - name: File a triage issue when the sweep fails
+      - name: File a tracking issue when the sweep fails
         # The scheduled sweep runs on no pull request, so a red run is otherwise visible only to
-        # whoever opens the Actions tab. This turns a failed sweep into an assigned, deduplicated
-        # tracking issue, so "a maintainer owns triage" is enforced by the workflow rather than
-        # asserted in prose. One open issue at a time: a run that fails while the marker issue is
-        # still open comments on it instead of opening a second.
+        # whoever opens the Actions tab. This turns a failed sweep into one open, labelled tracking
+        # issue: a run that fails while the marker issue is still open comments on it instead of
+        # opening a second. The key is the label, read through the Issues API, not title search
+        # (tokenised and eventually consistent). No failure here is swallowed: the job is already
+        # red, so an aborted filing step hides nothing, while a false "no marker" would file a
+        # duplicate, which is the one thing this step exists to prevent.
         if: failure()
         env:
           GH_TOKEN: ${{ github.token }}
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
           set -euo pipefail
-          marker="mutation-reproof: full sweep failing"
-          existing="$(gh issue list --state open --search "in:title \"$marker\"" --json number --jq '.[0].number' || true)"
-          body="The daily mutation-reproof full sweep failed. A fixture's guarded suite no longer kills its named mutant, or a fixture is dangling or malformed. Triage: open the run, read the SURVIVED / dangling / malformed lines, and repair the fixture or file the named finding. Run: $RUN_URL"
+          label="ci:mutation-reproof"
+          title="mutation-reproof: full sweep failing"
+          gh label create "$label" --description "Open while the scheduled mutation-reproof full sweep is failing" --color B60205 --force
+          existing="$(gh issue list --state open --label "$label" --json number --jq '.[0].number // empty')"
+          body="The daily mutation-reproof full sweep failed: $RUN_URL
+
+          Read the run log. A fatal verdict (SURVIVED, UNGRADABLE, WRONG-RED, ERROR) means a guarded suite no longer discriminates its named mutant or the fixture's anchor is broken; an UNMEASURED refusal names dangling or malformed fixtures; a failure before the sweep step (checkout, install, build) has no verdict lines at all. Repair the fixture or file the named finding, then close this issue once a scheduled sweep is green."
           if [ -n "$existing" ]; then
             gh issue comment "$existing" --body "$body"
           else
-            gh issue create --title "$marker" --body "$body"
+            gh issue create --title "$title" --label "$label" --body "$body"
           fi
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -64,6 +64,9 @@ jobs:
     if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     timeout-minutes: 360
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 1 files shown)


davidfarah2003 (commented):
**BLOCK** on `24a110e0f89ec876d1dfc700d82d637b84928993`.

(Could not use the changes-requested review type: this credential is the PR author. The verdict is BLOCK.)

The `full` job’s blast radius is what the PR claims. The dedup step is not: it converts a failed search into a create, which is the duplicate it exists to prevent, and it does so in a way later readers cannot tell from “no marker.”

**1. `permissions` / `if: failure()` blast radius.** Job-scoped as claimed. `permissions` sits under `full:` next to that job’s `if` / `runs-on` / `steps`, not at workflow level. `full` only runs on `schedule` or `workflow_dispatch`. `changed` has neither `issues: write` nor the filing step, so a pull-request run cannot open an issue. `if: failure()` on a skipped job never executes.

**2. Dedup swallows its own failure. This is the block.** `existing="$(gh issue list … || true)"` under `set -euo pipefail` makes a search/API/rate-limit/permissions failure look identical to “no open marker.” The script then `gh issue create`s. Search and Issues are different APIs, so list can fail while create still works. Daily cadence does not make that acceptable: the job is already red, so aborting the filing step costs nothing, while a false-empty create is exactly the duplicate the step’s comment promises not to open. Drop `|| true` (or distinguish empty from failed) before merge.

**3. Title search is a weak key, not independently blocking.** `--search 'in:title "…"'` is GitHub’s issue index: tokenised, eventually consistent, can lag a just-created issue. Two overlapping `full` runs (`workflow_dispatch` beside a schedule, or a re-run before the index catches up) can both see no marker and both create. An unrelated open issue whose title contains those tokens would take the comment. For a daily cron the overlap window is small. A label or a fixed body marker would be the sturdier key. Fine to land later if (2) is fixed; not fine as the only guard while search failures are swallowed.

**4. It files an issue. It does not assign one, and the body is a partial map of the tool.** `gh issue create` has no `--assignee`. “Assigned” in the title and in the step comment is aspirational. The body names SURVIVED / dangling / malformed, which the script does print (`SURVIVED ` verdicts, `dangling fixture(s)`, `malformed fixture(s)`). It omits the other fatal classes the sweep actually fails on (`UNGRADABLE`, `WRONG-RED`, `ERROR`, unexplained non-zero, `MUTATION REPROOF FAILED`). `if: failure()` also fires on checkout / install / build, where none of those lines exist. Pointing at the run URL still gets a human to the log. Claiming the workflow “enforces” that a maintainer owns triage does not.

**5. Recurring workflow-inventory sentinel.** This edits an existing file. Count stays 7. `on:` and top-level concurrency are unchanged, so `workflow-concurrency`’s `EXPECTED_WORKFLOWS=7`, `EXPECTED_PUSH_TO_MAIN=6`, `EXPECTED_GROUPED=4` still match. `shard-stability` reads only `ci.yml`. `pr-head-gate` still sees the existing `pull_request` trigger. Did not re-run those suites.

**CI at this sha (runs API, 40-char):** Mutation reproof success (the PR `changed` job, not the scheduled `full` path). Code Quality success. CI in_progress, Windows in_progress. Docs did not start (path filter, expected). `in_progress` treated as unmeasured. Did not dispatch or fail the workflow.

davidfarah2003 (commented):
**The BLOCK does not stand** on `f7edec9ecd8a04ef7b7fd86b07491f3bd28fc5db`.

Head matches `gh pr view 1295`. Parent is the previously reviewed `24a110e0f`. One file, `.github/workflows/mutation-reproof.yml`. Graded the delta by reading; did not dispatch the workflow.

**Original findings.**

1. **Closed (was already OK).** `permissions` / `if: failure()` still job-scoped on `full` only. `changed` still has neither `issues: write` nor the filing step. Unchanged, as claimed.

2. **Closed. This was the block.** `|| true` is gone. `existing="$(gh issue list …)"` sits under `set -euo pipefail`, so a failed list/API/rate-limit now aborts the filing step instead of becoming a create. The step comment states that argument. The job is already red; aborting here hides nothing.

3. **Closed as a block.** Dedup is no longer title search: `--search` is gone; the key is `gh issue list --state open --label "$label"` (Issues list filter, not the search index). The label is applied at create time (`gh issue create --title … --label "$label"`), not only minted as a repo label. Residual list-then-create overlap (two `full` runs both seeing empty, both creating) still exists; it is the same TOCTOU any list-then-create has, and the daily cron cannot overlap itself under the 360-minute cap. Not independently blocking now that a failed search cannot impersonate “no marker.”

4. **Closed.** Title, step name, and comment no longer say “assigned.” There is still no `--assignee`. The body names SURVIVED / UNGRADABLE / WRONG-RED / ERROR, UNMEASURED dangling/malformed, and the pre-sweep (checkout/install/build) case, which is what `scripts/mutation-reproof.mjs` actually fails on (`FATAL_VERDICTS`, the UNMEASURED exits, and `if: failure()` on earlier steps). `MUTATION REPROOF FAILED` is a banner over those fatals, not a separate class. INCONCLUSIVE / PRE-RED do not fail the gate, so they would not reach this step from the sweep.

5. **Closed (was already OK).** Still one existing file. Workflow count is 7. `on:` and top-level concurrency are unchanged, so `workflow-concurrency`’s `EXPECTED_WORKFLOWS=7`, `EXPECTED_PUSH_TO_MAIN=6`, `EXPECTED_GROUPED=4` still match (`mutation-reproof.yml` remains a push-to-main starter with no group). `pr-head-gate` still sees `pull_request`. `shard-stability` still reads only `ci.yml`. Did not re-run those suites.

**Press points on this delta.**

- **Label is attached, not just created.** `gh label create` ensures the repo label exists; `gh issue create --label "$label"` applies it on the issue that later list calls will key on. A label that existed and was never attached is not this script.
- **`gh label create --force` is the idempotent path, not a new swallow.** `gh label create --help`: create, or with `--force` update color/description if it already exists. Without `--force`, a second run would 422 and (now correctly) abort. `--force` should exit 0 when the label is already there. `issues: write` on this job is the permission Create-a-label needs. A real failure (auth, rate limit) now aborts, which is the intended distinction. Side effect: every red run rewrites this label’s color and description if a human changed them. Benign.
- **`jq '.[0].number // empty'` does distinguish empty from failed, and empty from `null`.** Checked against the command in the file, not the claim: `echo '[]' | jq '.[0].number // empty'` is zero bytes, so `[ -n "$existing" ]` is false (create). A hit prints the number. The same jq without `// empty` prints `null\n`, which is non-empty and would have taken the comment branch with issue “null”. `gh` failing is no longer mapped onto that empty string.

**CI at this sha (runs API, 40-char):** Mutation reproof success (the PR `changed` job, not `full`). Code Quality success. CI in_progress, Windows in_progress. Docs did not start (path filter, expected). `in_progress` treated as unmeasured. Did not dispatch or fail the workflow.

Nothing from the original BLOCK still withholds merge. Residual concurrent double-create is a later hardening, not a reason to hold this.

PR #1304, opened 2026-09-05
ci(mutation-reproof): raise the changed-set budget and record elapsed

### LINKED ISSUE #1299: mutation-reproof's changed job exceeds its 60-minute budget on high-fan-in PRs, and a timed-out job badges as cancelled
## Summary

The `changed` job in `.github/workflows/mutation-reproof.yml` has `timeout-minutes: 60` (line 19).
On a PR whose changed-file set touches a high-fan-in guarded source, the derived mutation set exceeds
that budget, the job is killed, and GitHub badges the result **`cancelled`**.

`cancelled` reads as "a newer run superseded this one", not "this job did not finish". So the failure
is invisible in the checks list and in `gh pr checks`, and it does not self-heal.

**This is the same defect class as #967** (the `live` job timing out at 25 minutes and badging as
cancelled, filed and closed), and #897 before it, on a third job.

## The measurement

PR #1294 (`feat/783-m1-custody-seam`), every Mutation reproof run on the branch:

| run | head sha | conclusion | `changed` duration |
|---|---|---|---|
| 33973541717 | `647e21b5ee` | cancelled | 60 min |
| 33974059962 | `2fc985c5d5` | cancelled | 60 min |
| 33974723292 | `342f2bba67` | cancelled | 60 min |
| 33974805339 | `71974bb61f` | cancelled | 60 min |

Four for four, on four different shas. The last measured 3615 s against a 3600 s limit.

For contrast, the `changed` job across the last twenty runs of this workflow on other PRs:

```
1 min, 1 min, 1 min, 2 min, 2 min, 2 min, 4 min, 7 min   (success)
1 min                                                     (a genuine failure, unrelated PR)
```

So this is not runner capacity and not a bad draw. Every other PR finishes inside 7 minutes; this
branch has never finished once.

**It is not superseded.** Each cancellation is the only Mutation reproof run at its head sha, so the
`cancelled` badge cannot be explained by a newer run replacing it. That is the same discriminator
#967 used, and it has to be checked explicitly, because superseded and timed-out are the same word
in the API.

## Cause

The changed-file mutation set for that PR is **33 paths, 42 fixtures, 181 mutations**, because
`manager.ts` is a high-fan-in guarded source and a merge from main widened the set further. At the
per-mutation cost this workflow runs at, 181 does not fit in 60 minutes.

Nothing is hung. The job is doing real work the whole time and is killed mid-sweep.

## Why this is worse than an unmeasured check

`scripts/pr-head-gate.mjs:131` grades any non-`success` conclusion as failing:

```js
if (run.conclusion !== "success") failing.push(name);
```

and `Mutation reproof` is in the expected workflow set for every PR
(`bin/smoke/pr-head-gate.smoke.ts:35`). So the gate correctly refuses — it fails **closed**, which is
the right direction.

The consequence is that **a PR large enough to exceed the mutation budget can never satisfy the
repo's own head gate, by any action available to its author.** Re-running produces another 60-minute
cancellation; it has already been tried on this branch. Splitting the PR is not always possible, and
the PRs that trip this are precisely the ones touching high-fan-in guarded sources, which are the
changes mutation coverage exists for.

So the gate is not being bypassed. It is unsatisfiable, with no path forward, on exactly the class of
change that most needs it.

## Done when

- A `changed` job that exceeds its budget is **distinguishable from a superseded run** without
  reading job timestamps by hand. Right now telling them apart requires enumerating runs at the head
  sha and comparing durations against the declared limit.
- A PR whose mutation set does not fit has a path to a graded result. Options worth weighing rather
  than a prescription: raise `timeout-minutes` for `changed`; shard the changed-file mutation set the
  way the smoke suites are sharded; cap or prioritise mutations per run and report what was not
  reached; or split the sweep so the gate can report on a defined subset honestly rather than being
  killed mid-way through an undefined one.
- Whatever is chosen, **a partial sweep must report as partial rather than as either colour.** A
  budget that silently truncates coverage is the failure in #1297 wearing different clothes.

## Scope note

Not asking for #1294 to be unblocked by this. Its reviewers correctly recorded the mutation gate as
unmeasured rather than counting it green, and its merge is being decided on independent review of the
exact head. This issue is about the budget being unsatisfiable, not about that PR.

davidfarah2003:
Closes #1299.

## The problem

`pr-head-gate.mjs:131` grades a non-success as failing and "Mutation reproof" is in its expected set, so the gate fails closed. That is correct behaviour. What is not correct is that a PR touching a high-fan-in guarded source can never reach a success by any action its author can take: the `changed` job's fixture set scales with the changed paths, and a large set exceeds the 60-minute cap, which cancels rather than fails.

#1294 is the worked example. It edits `implementations/manager/src/manager.ts`, selects **42 fixtures / 181 mutations**, and its `changed` job was cancelled at 60 minutes on five consecutive heads:

```
647e21b5e  cancelled     2fc985c5d  cancelled     342f2bba6  cancelled
71974bb61  cancelled     d1db29e23  cancelled  (17:33:51Z, 60m00s)
```

Every other PR in the last twenty runs finished in 1-8 minutes.

## Why 150 and not a derived number

Because nothing here has ever been measured well enough to derive one, and that is the substance of the issue rather than a detail of it.

Both available models **under-predict**:

| model | basis | predicts for 181 mutations |
|---|---|---|
| fastest observed rate | 9 fixtures / 39 mutations in 7m27s (run 33970489997) | ~42 min |
| corpus fraction | 181 of 1,714 mutations against `full`'s 360-minute budget | ~38 min |
| **actual** | | **exceeded 60** |

So per-mutation cost is dominated by *which* suites a fixture guards, not by how many mutations it carries, and both models are wrong for exactly the sets that need the budget. 150 is 2.5x the larger estimate, the smallest headroom that survives models known to under-predict by over 1.5x, and it stays well under `full`'s 360.

I could not anchor against a real corpus time: no scheduled `full` run has ever executed, so `full`'s own 360 was also a guess.

## The durable half

No run of this job has ever recorded its duration. That is why the budget could only be guessed, and why raising it alone would just move the wall. `Report elapsed` prints the changed-set duration so the next revision is measured rather than argued.

It runs under `always()` rather than `failure()`, because the run whose duration matters most is the one the timer kills, and a job timeout cancels rather than fails. This is **best effort on that path** and the comment says so: a cancelled job gives its remaining steps only a grace period, so a timed-out run may print nothing. It is reliable on the success and failure paths, which is what re-deriving the budget needs.

This mirrors Cotal-AI/foreman#376, filed today for the same shape on a different gate: an exclusive resource with a hard cap and no recorded duration, where every accepted run is necessarily under the cap so the baseline is unknowable.

## Scope

One file, `.github/workflows/mutation-reproof.yml`, +39/-1. The workflow count does not move (this edits an existing file), `on:` and top-level concurrency are unchanged, and the `full` job is untouched.

Pushed from a credential with the `workflow` scope; no lane credential on the agent host holds it, which is the same constraint that produced #1292.

### DIFF
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -16,7 +16,28 @@ jobs:
     # Source intersection catches the #1217 shape at the changing commit. The daily full sweep
     # below catches cross-cutting changes a fixture did not name within a day.
     if: github.event_name != 'schedule'
-    timeout-minutes: 60
+    # 150, raised from 60, and PROVISIONAL. A PR touching a high-fan-in guarded source selects a
+    # large fixture set: #1294 (which edits implementations/manager/src/manager.ts) selected 42
+    # fixtures / 181 mutations and was cancelled at the 60-minute cap five consecutive times, so
+    # it could not satisfy this gate by any action its author could take. `pr-head-gate.mjs`
+    # grades a non-success as failing and this workflow is in its expected set, so the gate fails
+    # closed on a job that cannot finish rather than on a defect.
+    #
+    # The number is not derived, because nothing here has ever been measured well enough to derive
+    # it, and that is the substance of #1299 rather than a detail of it. The two models available
+    # both UNDER-predict: at the fastest observed rate (9 fixtures / 39 mutations in 7m27s) 181
+    # mutations is ~42 minutes, and as a fraction of the corpus (181 of 1,714) against `full`'s
+    # 360-minute budget it is ~38 minutes. The real run exceeded 60. So per-mutation cost is
+    # dominated by WHICH suites a fixture guards, not by how many mutations it carries, and both
+    # models are wrong for exactly the sets that need the budget. 150 is 2.5x the larger estimate,
+    # which is the smallest headroom that survives models known to under-predict by over 1.5x, and
+    # it stays well under `full`'s 360.
+    #
+    # Re-derive this from the elapsed line the Report elapsed step now prints, once real sets have
+    # reported. Do
+    # not raise it again without that data: a limit raised past what anyone can measure trades a
+    # visible failure for an invisible one.
+    timeout-minutes: 150
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -56,7 +77,25 @@ jobs:
             base="$(git rev-parse HEAD~1)" || { echo "no base commit or first parent"; exit 2; }
             echo "no base commit on this event; using first parent $base"
           fi
+          echo "MUTATION_STARTED=$(date +%s)" >> "$GITHUB_ENV"
           node scripts/mutation-reproof.mjs --base "$base" --head "$(git rev-parse HEAD)"
+      - name: Report elapsed
+        # Until this step existed, no run of this job recorded how long it took, so a cancellation
+        # at the cap and a run finishing one second inside it were indistinguishable afterwards,
+        # and the budget above could only ever be guessed.
+        #
+        # `always()` rather than `failure()` because the run whose duration matters most is the one
+        # the timer kills, and a timeout cancels the job rather than failing it. This is best
+        # effort on that path: a cancelled job gives its remaining steps only a grace period, so a
+        # timed-out run MAY print nothing here. It is reliable on the success and failure paths,
+        # which is what re-deriving the budget needs.
+        if: always()
+        run: |
+          if [ -n "${MUTATION_STARTED:-}" ]; then
+            echo "mutation reproof: changed-set elapsed $(( $(date +%s) - MUTATION_STARTED ))s (budget 150m)"
+          else
+            echo "mutation reproof: no start marker; the job ended before the re-prove step began"
+          fi
 
   full:
     # A daily sweep keeps the remaining ambient-drift window bounded without making every pull
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -77,7 +77,11 @@ jobs:
             base="$(git rev-parse HEAD~1)" || { echo "no base commit or first parent"; exit 2; }
             echo "no base commit on this event; using first parent $base"
           fi
-          echo "MUTATION_STARTED=$(date +%s)" >> "$GITHUB_ENV"
+          # A FILE, not $GITHUB_ENV. The runner processes a step's env file when the step ENDS, so a
+          # step killed by `timeout-minutes` never contributes its variables and the next step would
+          # see nothing — on precisely the timed-out run this measurement exists for. A file written
+          # by the shell exists the moment the redirect completes.
+          date +%s > "$RUNNER_TEMP/mutation-started"
           node scripts/mutation-reproof.mjs --base "$base" --head "$(git rev-parse HEAD)"
       - name: Report elapsed
         # Until this step existed, no run of this job recorded how long it took, so a cancellation
@@ -85,14 +89,17 @@ jobs:
         # and the budget above could only ever be guessed.
         #
         # `always()` rather than `failure()` because the run whose duration matters most is the one
-        # the timer kills, and a timeout cancels the job rather than failing it. This is best
-        # effort on that path: a cancelled job gives its remaining steps only a grace period, so a
-        # timed-out run MAY print nothing here. It is reliable on the success and failure paths,
-        # which is what re-deriving the budget needs.
+        # the timer kills, and a timeout cancels the job rather than failing it. The start marker is
+        # a file for the same reason: $GITHUB_ENV is only processed when a step ENDS, so a step the
+        # timer kills contributes nothing and this step would take its else branch on exactly the
+        # run it was added for. Still best effort on that path — a cancelled job gives its remaining
+        # steps only a grace period, so this may not run at all — but when it does run it now has a
+        # start time to work from.
         if: always()
         run: |
-          if [ -n "${MUTATION_STARTED:-}" ]; then
-            echo "mutation reproof: changed-set elapsed $(( $(date +%s) - MUTATION_STARTED ))s (budget 150m)"
+          marker="$RUNNER_TEMP/mutation-started"
+          if [ -s "$marker" ]; then
+            echo "mutation reproof: changed-set elapsed $(( $(date +%s) - $(cat "$marker") ))s (budget 150m)"
           else
             echo "mutation reproof: no start marker; the job ended before the re-prove step began"
           fi


davidfarah2003:
Verification note from the author, so the reviewer does not have to discover it: **this PR's own run proves the instrumentation prints, and does not prove it measures anything.**

`Mutation reproof / changed` at `5c6bcd73e3ba2c1ad7d2aeb3ffe76ecb62bd04c5` completed `success`, and the new step printed:

```
mutation reproof: 0 fixture(s) selected from 342 (diff 1 record(s), 1 changed path(s), corpus 342)
mutation reproof: changed-set elapsed 0s (budget 150m)
```

`0s` is correct and degenerate. This PR changes one workflow file and no guarded source, so the selection is empty and the re-prove does no work. The step is confirmed to run, read `MUTATION_STARTED`, and emit a well-formed line on the success path. It is **not** evidence about a real set's duration, and it is not evidence about the `always()` path, which is the one that matters for a timed-out run and which nothing here exercises.

The first genuinely informative number will come from the next PR that touches a guarded source. Until several of those have reported, the 150 in this change stays what the comment says it is: provisional, chosen because both available models under-predict, not derived.

Job wall clock for reference: started 17:40:53Z, completed 17:42:09Z (76s), essentially all of it checkout, install, nats-server and build.

davidfarah2003:
### Head moved to `1cdf9322b539f106f09eb2c097803ae1ee84ea47` — review that sha, and grade the delta

For anyone reviewing this: the head moved after review began. It is **one commit on top of
`5c6bcd73e`, not a rebase** — I verified `5c6bcd73e` is still an ancestor — so the earlier commit's
review still holds and only the delta needs grading. Re-bind rather than starting over.

What moved, and why it is worth knowing before reading the diff: the original `changed`-job start
marker was written with `echo "MUTATION_STARTED=..." >> "$GITHUB_ENV"`. The runner processes a step's
env file when the step **ends**, so a step killed by `timeout-minutes` contributes nothing, and the
`if: always()` elapsed step would have taken its `else` branch on precisely the timed-out run the
measurement exists for. The old comment described this as "MAY print nothing here"; the accurate
statement was "will never print a duration on that path". It is now
`date +%s > "$RUNNER_TEMP/mutation-started"`, a file that exists the moment the redirect completes,
read back with `[ -s "$marker" ]`.

One hole is knowingly left open and should not be reported as an unaddressed finding: an `if: always()`
step may not run **at all** when a job is cancelled by its own `timeout-minutes`, since a cancelled job
grants remaining steps only a grace period. A step-level timeout shorter than the job's would close it
by making the step **fail** rather than the job **cancel** — but a step timeout and a job timeout are
graded differently by `pr-head-gate`, so that changes what the gate means and belongs in its own change
with its own review rather than riding in on a budget bump. Deliberately deferred, not missed.

Also worth stating so nobody reads the check list wrongly: the `changed` job's green run here reported
`0 fixture(s) selected` and `elapsed 0s`, because this PR touches no guarded source. It proves the step
**emits**. It proves nothing about a real fixture set and nothing about the timeout path. And the `CI`
run at this head currently reads `cancelled` — superseded by the push through the concurrency group,
which is unmeasured rather than green or failing.

davidfarah2003:
**BLOCK** on `5c6bcd73e3ba2c1ad7d2aeb3ffe76ecb62bd04c5`.

The budget raise is the right shape for #1299, and the five #1294 cancellations are real 60-minute job timeouts, not supersedes. The elapsed step, as written, does not measure the timeout path it exists for. That is the durable half of this PR. Ship the marker from a step that has already completed, then the 150-minute cap is acceptable as a provisional wall.

### Timeout-path elapsed (the question that matters)

`echo "MUTATION_STARTED=…" >> "$GITHUB_ENV"` lives **inside** the re-prove step. GitHub's workflow-commands docs: a `GITHUB_ENV` write is visible to **subsequent** steps after the writing step completes. The runner does process that file on cancel if the worker is still alive: `ActionRunner.RunAsync` calls `fileCommandManager.ProcessFiles` in a `finally` after `handler.RunAsync` ([actions/runner `ActionRunner.cs`](https://raw.githubusercontent.com/actions/runner/main/src/Runner.Worker/ActionRunner.cs)). So this is **not** "the runner discards `GITHUB_ENV` on cancel."

The failure mode is still the timeout path:

- Job `timeout-minutes` cancels the job. The running re-prove step's condition is the default `success()`, so `StepsRunner` cancels **that** step. `Report elapsed` is `if: always()`, which GitHub documents as continuing after cancel, with a **5-minute** cancellation grace, then a server-side force kill ([workflow cancellation reference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-cancellation)).
- On every measured #1294 timeout, the cancelled job **did not run any step after re-prove**. There was no `Report elapsed` then (this PR did not exist), and the post-cache / post-setup-node steps were `skipped`. The always() grace is real in the docs. It is not something this repo has ever observed for this job.
- If `always()` does run but `ProcessFiles` did not (worker force-killed during teardown), `MUTATION_STARTED` is absent and the `else` branch prints `no start marker; the job ended before the re-prove step began`. That sentence is false on the path this step was added for: the re-prove had been running for ~59 minutes.

The author's "MAY print nothing here" understates it. On the only path whose duration we do not have, this design either prints nothing or prints a lie. The green run at this head (`33981474734`) printed `elapsed 0s` after `0 fixture(s) selected`. That proves the step emits on success. It proves nothing about timeout.

**What would work:** write the marker from a **previous, successful, short step** (or `$GITHUB_OUTPUT` from that step, or a file under `$RUNNER_TEMP`). Then `always()` only has to run; it does not depend on a cancelled step's env-file teardown.

```yaml
      - name: Mark re-prove start
        run: echo "MUTATION_STARTED=$(date +%s)" >> "$GITHUB_ENV"
      - name: Re-prove fixtures for changed guarded sources
        ...
```

Do that, and the `else` message becomes true.

### 1. Does 150 make the gate satisfiable, or just move the wall?

For the one known failing case, 150 is probably enough, and that is a stronger claim than the PR's models.

On `d1db29e237fdd6c2f01ae3a163bbdc387fee6770` (run `33978243970`, unique Mutation reproof run at that sha, `cancelled` at 3615s): re-prove selected **43 fixtures / 183 mutations** (log: `43 fixture(s) selected from 342`) and completed 26 fixtures before `The operation was canceled.` at `implementations/manager/smoke/fixtures/persona-role-capability.mutations.json`. That is progress, not a hang. Linear by the finished prefix is ~95–110 minutes of re-prove. Remaining work is the heavier manager fixtures (`reconcile-startup` 17, `events-own-channel` 11, `renewal-terminal-serialization` 11, …). If those are merely as slow as the first hour, 150 holds. If they are much slower, 150 is another wall. Treat cancelled as unmeasured for an upper bound: we still have none.

The PR's own models do **not** produce the "~42 minutes" the comment uses:

- 9 fixtures / 39 mutations in 7m27s is real (run `33970489997`, job `13:59:58Z`–`14:07:25Z` = 447s; 6+1+10+1+5+1+11+2+2 = 39). `181/39 * 7m27s` is **~35 min**, not ~42. Fixture-scaled `42/9 * 7m27s` is also ~35.
- `181/1714 * 360` is ~38 min, as stated.
- "150 is 2.5× the larger estimate" is false as arithmetic: 2.5×42 = 105, 2.5×60 = 150. 150 is 2.5× the observed floor, not 2.5× the model.

That does not make 150 reckless for #1294. It does mean the comment will mislead the next person who re-derives the number. The warning "do not raise it again without elapsed data" is the right rule. It is not obeyed by this elapsed design on the timeout path, which is the only path that would give an upper bound.

### 2. Hang cost at 150

A hung `changed` job burns 2.5 hours of `ubuntu-latest` instead of 1. Nothing in this workflow `needs:` `changed`. `full` is a different `if:` and does not wait on it. **This workflow has no `concurrency:` group** (confirmed across `.github/workflows/*.yml` for this file; `CI` has its own group and does not include Mutation reproof). `pr-head-gate.mjs` waits on the workflow conclusion, so a hang delays merge by the full budget and occupies one GitHub-hosted runner. That cost is real and contained. Acceptable if elapsed on timeout actually lands, because then a hang is distinguishable from "still working at 149m."

### 3. `pr-head-gate.mjs`

Still satisfied. Top-level `name: Mutation reproof` is unchanged. `bin/smoke/pr-head-gate.smoke.ts` still expects `["CI", "Docs", "Mutation reproof", "Windows"]`. `classifyPullRequestHead` still grades any non-success, including `cancelled`. This diff does not change that aggregate.

`Docs` is path-filtered and correctly absent from this PR's runs (this PR does not touch `docs/**`). At this exact sha the Mutation reproof run succeeded; `CI` was still `in_progress` when checked. A green Mutation reproof job is not a green PR.

### 4. `full` at 360

Untouched: `timeout-minutes: 360`, `if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'`. 150 < 360 still holds as a deliberate relationship. Scheduled runs of this workflow: **zero** (`event=schedule` total_count 0). The daily sweep is unaffected because it has never run.

### 5. Shell

`set -euo pipefail` is only on the re-prove step. The marker write can fail that step before `node` runs; that is fine. `${MUTATION_STARTED:-}` is the right empty guard. `$(( $(date +%s) - MUTATION_STARTED ))` is safe in arithmetic once the guard has passed (unquoted is normal there). Report elapsed has no `set -e`; a `date` failure would skip the intended line rather than fail the step. Minor. The `else` text is the actual bug, see above.

### 6. Does the comment tell the truth?

Most of the #1294 story is true. The counts are off in ways a later re-derivation will copy.

Five consecutive timeouts, full shas, one Mutation reproof run each, `cancelled` at ~3615s, re-prove step `cancelled`, job duration 60.23–60.28 min:

| sha | run | duration |
|---|---|---|
| `647e21b5ee9284847e7df87862c6e467807ca9cc` | `33973541717` | 3614s |
| `2fc985c5d5451b9db9dd6f2d01ba1e8f846196b6` | `33974059962` | 3615s |
| `342f2bba670887cbfd95823704b094340be86045` | `33974723292` | 3616s |
| `71974bb61fb1ca50ffd4da854d4549d3a707fccc` | `33974805339` attempt 1 **and** 2 | 3615s / 3617s |
| `d1db29e237fdd6c2f01ae3a163bbdc387fee6770` | `33978243970` | 3615s |

Not supersedes: unique run per sha. `71974bb61` is one run with two attempts, both timed out.

**42 fixtures** is true of the first of those (`647e21…`: `42 fixture(s) selected from 341`). It is **not** true of the fifth (`d1db29…`: **43 / 342**). **181 mutations** was not re-counted on the 42-set (those fixture files at `647e21` were not fetched). On the 43-set at `d1db29` the same paths sum to **183**. Close, not the number in the comment.

**9 / 39 in 7m27s** is true. **181 of 1,714** and the file's older "329 configs / 1,714 mutations" are stale next to this PR's own log (`corpus 342`) and a working-tree count of 339 committed fixture files / 1751 mutations. Do not re-derive from 1,714.

### What I could not check

- I did not run `mutation-proof` or time a 43-fixture set locally (forbidden, and would not be the GHA wall).
- I did not count mutations on the 42-fixture set at `647e21…`, so 181 remains the author's figure for that earlier head.
- I did not execute a job-timeout in this repo; the `GITHUB_ENV`/`finally` claim is from current `actions/runner` source plus GitHub's cancellation doc, not from a live timeout of *this* workflow with `Report elapsed` present.
- `CI` at this sha was still `in_progress` (`33981474660`) when reviewed.

Worktree left clean; `pnpm install` was not run.

davidfarah2003:
**APPROVE** on `1cdf9322b539f106f09eb2c097803ae1ee84ea47`.

Re-bind of the BLOCK on ancestor `5c6bcd73e3ba2c1ad7d2aeb3ffe76ecb62bd04c5`. That sha is still an ancestor. The delta is one commit: the start marker is now a file under `$RUNNER_TEMP`, not `$GITHUB_ENV`. That was the functional defect. It is fixed.

### Timeout-path elapsed

`$GITHUB_ENV` is processed when a step ends (`FileCommandManager.ProcessFiles` in the runner's `finally`). Writing the marker inside the re-prove step therefore could not be relied on once that step was the one the job timer killed. `date +%s > "$RUNNER_TEMP/mutation-started"` exists the moment the redirect returns, before `node` starts. `[ -s "$marker" ]` is the right non-empty check. The `else` line is now true when it fires: the re-prove step never reached the write.

This head's Mutation reproof run is unique (`33983081385`). `changed` succeeded; `Report elapsed` printed `mutation reproof: changed-set elapsed 0s (budget 150m)` after `0 fixture(s) selected from 342`. That proves the file round-trip on success. It still proves nothing about a real fixture set and nothing about timeout.

The remaining hole is the one the commit names: an `if: always()` step may not run after the job's own `timeout-minutes` cancel. A cancelled job gives remaining steps a 5-minute grace, then a server-side kill ([workflow cancellation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-cancellation)). This repo has never observed `Report elapsed` on a timeout (the step did not exist on the #1294 kills). Closest evidence: on `d1db29e237fdd6c2f01ae3a163bbdc387fee6770` run `33978243970` the re-prove step was `cancelled` and the runner **did** continue — `Post Cache nats-server` / `Post setup-node` were `skipped`, `Post pnpm/action-setup` and `Post checkout` succeeded. So the step loop is not abandoned at the cancel. That makes `always()` more likely to run than the comment fears. It is still unmeasured for this step. A step-level timeout would fail the step instead of cancelling the job and would change what `pr-head-gate` grades; deferring that is correct.

### What did not change, and still holds

- **150** is still provisional. On `d1db29…` the job selected **43 fixtures / 183 mutations** (not the comment's 42/181) and was still progressing at 60m. Linear from the finished prefix is ~95–110 minutes if the rest is similar. There is still no measured upper bound. Do not raise 150 again without a timeout-path elapsed line.
- The comment's "~42 minutes" and "150 is 2.5× the larger estimate" are still wrong arithmetic (`181/39 × 7m27s` ≈ 35 min; 2.5×60 = 150). Residual, not a reason to re-block: the number is labelled provisional and the measurement that would re-derive it now has a start time on the path that matters.
- Hung-job cost is 2.5h of `ubuntu-latest`. This workflow has no `concurrency:` group. Nothing `needs:` `changed`. `pr-head-gate` waits on the workflow conclusion, so a hang delays merge by the full budget.
- Workflow `name: Mutation reproof` is unchanged. `bin/smoke/pr-head-gate.smoke.ts` still expects `["CI", "Docs", "Mutation reproof", "Windows"]`.
- `full` is still `timeout-minutes: 360` and `if: schedule || workflow_dispatch`. Zero scheduled runs in history.
- Shell: re-prove keeps `set -euo pipefail`; a failed `date` fails that step before `node`. Report elapsed is the runner's `bash -e`. `$(cat "$marker")` in arithmetic is fine for `date +%s` output.

### This PR's own checks at this sha

Mutation reproof unique run `33983081385` success. Windows success. CI run `33983081370` was still `in_progress` (attempt 2) when reviewed — unmeasured, not green. `Docs` correctly absent (no `docs/**`). A green Mutation reproof job is not a green PR.

Worktree left clean; `pnpm install` was not run.

PR #1302, opened 2026-09-05
fix(manager): retry failed static reconciliation

davidfarah2003:
## Summary

- report startup and post-adoption static lifecycle reconciliation as an aggregate with attempted, succeeded, and failed counts plus greppable alias, phase, and disposition values
- expose current reconciliation state through manager `status` and `cotal status --components`
- re-drive failed coordinates in process through a fresh durable slot read, `planStaticSlotResume`, and the existing deterministic exact terminal
- bound each manager process to one initial attempt plus retries after 1s, 5s, and 30s, with keyed single-flight across the durable read and terminal entry
- add a real authenticated broker suite, CI fragment, mutation proofs, operator docs, and a patch changeset

## Convergence and safety

The status snapshot is not authority. Before every re-drive the manager re-reads the slot from broker KV and runs the same planner again. The retry reuses `retireOpId(lifecycleUid)`, so `runStaticTerminal` passes its own same-op frozen retirement and resumes the idempotent head transition. A different retirement op or another frozen operation is reported as `refused-foreign` and is never force-cleared.

The delays are scheduling only and never release a fence. The durable non-retired slot and exact-op state keep refusing alias reuse between attempts. No new lifecycle or second terminal operation is minted.

The 1s retry covers a one-shot control-plane or eviction blip. The 5s and 30s spacing gives a recovering broker more time without hammering a persistently unavailable authority. Four total entries are a finite per-process budget. A failure lasting beyond about 36 seconds becomes `retry-exhausted`; the alias stays unavailable and needs a manager restart. The next process receives a fresh budget because its counters are in memory and the durable row remains the sole authority.

`recovered` remains visible until the next static reconciliation sweep, then clears. The literal dispositions, next retry time, error, and remedy are visible in status and component health.

## Scope of #774

This PR addresses the reported static-lifecycle reconciliation wedge: failed terminals now produce a failed aggregate, remain visible in current health, and receive bounded same-process re-drives. It does not implement general desired-fleet convergence from a last-applied user manifest or relaunch missing declared seats on manager boot. That broader boot-convergence scope remains open, so #774 stays open and this PR only references it.

## Relationship to #1274

This change reports whether a reconciliation attempt is running, waiting, recovered, refused, or exhausted. It does not independently project whether static footprint cleanup completed. The durable cleanup marker and a general read verb remain #1274's scope.

## Changeset

[...]

PR #1309, opened 2026-09-05
ci(mutation-reproof): correct the budget provenance and make a breach a step failure

### LINKED ISSUE #1299: mutation-reproof's changed job exceeds its 60-minute budget on high-fan-in PRs, and a timed-out job badges as cancelled

davidfarah2003:
Follow-up to #1304, which merged at `99d50f510`. `Refs #1299` rather than Closes: #1299 is already closed by that merge, and this corrects what it landed.

Two changes with independent reasons. They ride together only because both touch the same twelve lines and #1304's reviewer surfaced both.

## 1. The comment's arithmetic was wrong twice, and its frame was wrong once

Found by #1304's reviewer, confirmed independently by recomputation.

| claim as landed | actual |
|---|---|
| "150 is 2.5x the larger estimate" (estimate ~42 min) | 150/42 = **3.57**. 150/60 = **exactly 2.50** |
| "models known to under-predict by over 1.5x" | a run cancelled AT 60 against ~42 est is **≥1.43x with no upper bound** |
| fastest-rate model "~42 min" | `181/39 × 7m27s` = **34.6 min** |

So 150 came from `2.5 × 60`, the **old budget**, and the sentence credited a derivation that never happened. Someone re-deriving would have gone looking for a 42-minute basis that the text never contained. The under-prediction bound was an assumption written in the grammar of a measurement: nobody knew what that run would have taken, because it was killed.

**The frame error is the one that mattered more.** #1294 had at least two mutation runs with different selections, and the comment attached one run's figures to all five timeouts:

```
647e21b5e (run 33973541717)   42 fixtures selected from 341
d1db29e23 (run 33978243970)   43 fixtures selected from 342
```

The extra fixture on the later head is `self-update-handoff.json`, which that head added. So 42 and 43 were never in conflict, and "correcting" one to the other would have introduced an error. Each figure now names its own sha.

**The budget now has a real derivation**, taken from the cancelled run's own per-step timings rather than from a model:

```
setup (checkout, install, build)      56s
Re-prove  16:34:33Z -> 17:33:49Z      3556s = 59.3 min, cancelled at the cap
completed                             109 of 183 mutations  (60%)
observed rate ON THAT SET             32.6 s/mutation
projected whole job                   ~100 min
headroom at 150                       ~1.49x
```

Both models fitted from other runs under-predict this set (~35 min and ~38 min against a real ~100), which is why the comment now says not to re-derive from them. Per-mutation cost is dominated by which suites a fixture guards.

## 2. A budget breach should be a step failure, not a job cancellation

`timeout-minutes: 145` on the re-prove step, below the job's 150.

A **job** timeout produces `cancelled`; a **step** timeout produces `failure`. `pr-head-gate.mjs` grades those differently, and a cancellation reads like unrelated CI churn rather than like something the author can act on — which is exactly how #1294's five timeouts read until someone measured them.

It also closes the hole #1304's own comment admits. `Report elapsed` runs under `always()`, but a cancelled job gives its remaining steps only a grace period, so the run whose duration matters most was the one least likely to report it. With a step timeout the job continues normally and the step runs.

This is a change to what the gate *means*, which is why it was deliberately kept out of #1304 rather than riding in on a budget bump.

## Verification

- `yaml.safe_load` parses; job timeout 150, step timeout 145; both step bodies pass `bash -n`.
- Scope is one file, `.github/workflows/mutation-reproof.yml`. The workflow count does not move, `on:` and concurrency are unchanged, and the `full` job is untouched.
- Not verified here, and worth a reviewer's attention: no run has yet exercised the step-timeout path, so its `failure` classification is reasoned from the docs rather than observed. The first high-fan-in PR after this lands is the test.

### DIFF
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -29,10 +29,20 @@ jobs:
     #                                self-update-handoff.json, which that head added
     #
     # The budget is derived from cancelled runs' own progress rather than from a model. Count
-    # completed mutations by their VERDICT token (KILLED / SURVIVED / INCONCLUSIVE / UNGRADABLE /
-    # WRONG-RED / ERROR / PRE-RED, the vocabulary at scripts/mutation-reproof.mjs:217). Do not
-    # grep for KILLED alone: a prediction line reads "PREDICTED KILLED", so a line-count over that
-    # token overstates the total, which is how an earlier revision of this comment said 109.
+    # completed mutations by their VERDICT token. The authority is the `VERDICTS` array in
+    # scripts/mutation-reproof.mjs (six tokens as of 99d50f510: KILLED, SURVIVED, INCONCLUSIVE,
+    # UNGRADABLE, WRONG-RED, ERROR) — read it from source rather than trusting this list, which is
+    # a snapshot and will drift when a seventh is added.
+    #
+    # Two tokens that look countable and are not:
+    #   PREDICTED KILLED  a per-mutation PREDICTION preamble, not a verdict. `grep -c KILLED`
+    #                     counts those lines too, which is how an earlier revision said 109 of a
+    #                     set whose real total is 99.
+    #   PRE-RED (N ...)   a per-FIXTURE summary for suites already red before mutation. It is not
+    #                     in VERDICTS and contributes ZERO completed mutations, but its emitted
+    #                     line is `PRE-RED (` — token, space, paren — so a naive alternation with a
+    #                     trailing space matches it and silently adds one fixture-level event to a
+    #                     mutation-level count.
     #
     #   d1db29e23  re-prove 16:34:33Z -> 17:33:49Z (59.3 min), 99 of 183 done   35.9 s/mutation
     #   ec6f1f7b2  re-prove 18:35:56Z -> 19:35:15Z (59.3 min), 85 done          41.9 s/mutation
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -28,11 +28,20 @@ jobs:
     #   d1db29e23 (run 33978243970)  43 fixtures selected from 342, the extra one being
     #                                self-update-handoff.json, which that head added
     #
-    # The budget is derived from d1db29e23's own progress rather than from a model. That job spent
-    # 56s on setup, then ran the re-prove step 16:34:33Z -> 17:33:49Z (59.3 min) before the cap
-    # cancelled it, completing 109 of its 183 mutations, 60% of that set. That is 32.6 s/mutation
-    # ON THAT SET, projecting 99.5 min for the re-prove and ~100 min for the whole job. 150 leaves
-    # about 1.49x headroom over that projection and stays well under `full`'s 360.
+    # The budget is derived from cancelled runs' own progress rather than from a model. Count
+    # completed mutations by their VERDICT token (KILLED / SURVIVED / INCONCLUSIVE / UNGRADABLE /
+    # WRONG-RED / ERROR / PRE-RED, the vocabulary at scripts/mutation-reproof.mjs:217). Do not
+    # grep for KILLED alone: a prediction line reads "PREDICTED KILLED", so a line-count over that
+    # token overstates the total, which is how an earlier revision of this comment said 109.
+    #
+    #   d1db29e23  re-prove 16:34:33Z -> 17:33:49Z (59.3 min), 99 of 183 done   35.9 s/mutation
+    #   ec6f1f7b2  re-prove 18:35:56Z -> 19:35:15Z (59.3 min), 85 done          41.9 s/mutation
+    #
+    # Two independent high-fan-in manager.ts sets on different branches, both killed at the old
+    # 60-minute cap, rates within ~17%. At the faster rate a 183-mutation set projects ~110 min and
+    # 150 leaves about 1.37x headroom; at the slower it projects ~128 min and about 1.17x. The
+    # slower figure is the one to plan against, and 1.17x is thin. Widen only from an elapsed line
+    # on a real set, not from another estimate.
     #
     # Models fitted from OTHER runs under-predict badly and should not be used to re-derive. The
     # fastest observed successful run (9 fixtures / 39 mutations in 7m27s) puts d1db's 183
@@ -74,13 +83,18 @@ jobs:
       - name: Build
         run: pnpm build
       - name: Re-prove fixtures for changed guarded sources
-        # A STEP timeout, deliberately below the job's 150. This changes what a budget breach MEANS
-        # to the head gate and that is the point: a job timeout produces `cancelled`, a step timeout
-        # produces `failure`, and `pr-head-gate.mjs` grades them differently. A breach should read
-        # as a failure the author can see and act on, not as a cancellation that looks like
-        # unrelated CI churn. It also lets the Report elapsed step run normally instead of racing a
-        # cancellation grace period, so the one run whose duration matters most is the one that can
-        # now report it.
+        # A STEP timeout, deliberately below the job's 150, so that a breach fails the STEP and the
+        # job continues to its remaining steps instead of being cancelled outright.
+        #
+        # The reason is Report elapsed, and only that. A cancelled job gives its remaining steps a
+        # grace period at best, so the run whose duration matters most was the least likely to
+        # report it. A failed step leaves the job running normally and that step executes.
+        #
+        # It is NOT a change to what the head gate sees, and an earlier draft of this comment
+        # claimed it was. `scripts/pr-head-gate.mjs:131` is the whole of its conclusion handling —
+        # `if (run.conclusion !== "success") failing.push(name)` — one comparison with no branch, so
+        # `cancelled`, `failure`, `timed_out` and `neutral` are identical to it. The gate's verdict
+        # is byte-identical either way.
         timeout-minutes: 145
         env:
           BASE: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -16,27 +16,33 @@ jobs:
     # Source intersection catches the #1217 shape at the changing commit. The daily full sweep
     # below catches cross-cutting changes a fixture did not name within a day.
     if: github.event_name != 'schedule'
-    # 150, raised from 60, and PROVISIONAL. A PR touching a high-fan-in guarded source selects a
-    # large fixture set: #1294 (which edits implementations/manager/src/manager.ts) selected 42
-    # fixtures / 181 mutations and was cancelled at the 60-minute cap five consecutive times, so
-    # it could not satisfy this gate by any action its author could take. `pr-head-gate.mjs`
-    # grades a non-success as failing and this workflow is in its expected set, so the gate fails
-    # closed on a job that cannot finish rather than on a defect.
+    # 150, raised from the previous 60-minute budget (150 = 2.5 x 60). A PR touching a high-fan-in
+    # guarded source selects a large fixture set, and #1294 could not satisfy this gate by any
+    # action its author could take: its `changed` job was cancelled at the 60-minute cap on five
+    # consecutive heads. `pr-head-gate.mjs` grades a non-success as failing and this workflow is in
+    # its expected set, so the gate failed closed on a job that could not finish rather than on a
+    # defect.
     #
-    # The number is not derived, because nothing here has ever been measured well enough to derive
-    # it, and that is the substance of #1299 rather than a detail of it. The two models available
-    # both UNDER-predict: at the fastest observed rate (9 fixtures / 39 mutations in 7m27s) 181
-    # mutations is ~42 minutes, and as a fraction of the corpus (181 of 1,714) against `full`'s
-    # 360-minute budget it is ~38 minutes. The real run exceeded 60. So per-mutation cost is
-    # dominated by WHICH suites a fixture guards, not by how many mutations it carries, and both
-    # models are wrong for exactly the sets that need the budget. 150 is 2.5x the larger estimate,
-    # which is the smallest headroom that survives models known to under-predict by over 1.5x, and
-    # it stays well under `full`'s 360.
+    # The selection differs per head, so each figure names its own sha:
+    #   647e21b5e (run 33973541717)  42 fixtures selected from 341
+    #   d1db29e23 (run 33978243970)  43 fixtures selected from 342, the extra one being
+    #                                self-update-handoff.json, which that head added
     #
-    # Re-derive this from the elapsed line the Report elapsed step now prints, once real sets have
-    # reported. Do
-    # not raise it again without that data: a limit raised past what anyone can measure trades a
-    # visible failure for an invisible one.
+    # The budget is derived from d1db29e23's own progress rather than from a model. That job spent
+    # 56s on setup, then ran the re-prove step 16:34:33Z -> 17:33:49Z (59.3 min) before the cap
+    # cancelled it, completing 109 of its 183 mutations, 60% of that set. That is 32.6 s/mutation
+    # ON THAT SET, projecting 99.5 min for the re-prove and ~100 min for the whole job. 150 leaves
+    # about 1.49x headroom over that projection and stays well under `full`'s 360.
+    #
+    # Models fitted from OTHER runs under-predict badly and should not be used to re-derive. The
+    # fastest observed successful run (9 fixtures / 39 mutations in 7m27s) puts d1db's 183
+    # mutations at ~35 min, and the corpus fraction (183 of 1,714) against `full`'s budget puts it
+    # at ~38. The real set needed ~100. Per-mutation cost is dominated by WHICH suites a fixture
+    # guards, not by how many mutations it carries, so a rate measured on a small set does not
+    # transfer to a large one.
+    #
+    # Re-derive from the elapsed line the Report elapsed step prints, on a real high-fan-in set,
+    # and name the sha you measured. Do not raise this again without that data.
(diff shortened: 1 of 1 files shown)


davidfarah2003:
**BLOCK** at exact head `ff081dd2157bcce384eb679c172d3fd074317065`.

## 1. Timeout semantics and the head gate

The observable job-timeout half is correct. The `changed` jobs in runs `33973541717`, `33974059962`, `33974723292`, `33974805339`, and `33978243970` all completed with conclusion `cancelled`, at about 60m14s to 60m17s. Their re-prove steps are also recorded as `cancelled`.

The step-timeout half remains an explicit residual. Current GitHub workflow-syntax docs say `steps[*].timeout-minutes` is “The maximum number of minutes to run the step before killing the process.” They do not state there that the step/job conclusion becomes `failure`, nor that a later `always()` step is guaranteed to run. I found no step-level timeout in this repository and no historical run here that corroborates those outcomes. The PR honestly identifies that the new path has not run, but the durable workflow comment writes the unobserved classification and continuation as facts.

More importantly, the load-bearing gate claim is false. `scripts/pr-head-gate.mjs` has only three buckets: an incomplete run is pending, a completed `success` passes, and every other completed conclusion is appended to `failing` by `if (run.conclusion !== "success") failing.push(name)`. It does not distinguish `cancelled` from `failure` in any direction. Therefore the comment's “changes what a budget breach MEANS to the head gate” claim is unsupported by the code, and this stated reason for the change is not delivered. This is blocking.

## 2. 145 under 150 margin

I measured every step before re-prove from 30 recent `changed` jobs via `/actions/runs/<id>/jobs`. Pre-step wall time ranged from **47s to 74s**. The slowest observed run was `33972300304`: setup job 1s, checkout 6s, pnpm setup 6s, node setup 9s, install 4s, cache 2s, NATS install skipped, PATH 0s, build 44s, for 74s wall time. The reference run `33978243970` took 57s, including a 38s build.

So the five-minute separation is safe against the observed sample. At the measured 74s worst case, the step timeout would fire around job minute 146.23, leaving about 3m46s before the job cap. I do not request a different split on margin grounds.

## 3. Derivation and attribution

The SHA/run attribution is correct:

- `647e21b5ee9284847e7df87862c6e467807ca9cc`, run `33973541717`: log says `42 fixture(s) selected from 341`.
- `d1db29e237fdd6c2f01ae3a163bbdc387fee6770`, run `33978243970`: log says `43 fixture(s) selected from 342`, and its selected list contains the sole additional path, `implementations/cli/smoke/mutations/self-update-handoff.json`.
- Reconstructing the 43 configs at `d1db29e...` gives 183 selected mutations.

The central progress numerator is not correct. The runner is sequential. The cancelled log contains 27 fixture headings. The first 26 fixtures completed and contain **99 mutations**. The 27th, `persona-role-capability.mutations.json`, has 4 mutations but printed only its heading before cancellation, with no baseline or mutation verdict. Thus the evidence supports 99 completed mutations, or at most 103 mutations in fixtures started, not 109 completed. The comment's 109/183, 32.6s/mutation, 99.5-minute projection, and 1.49x headroom do not re-derive from the named run.

Using the observed 99 completed mutations and the stated 3,556 seconds yields 35.9s/completed mutation, a 109.6-minute re-prove projection and about 110.5 minutes for the job, so 150 is about 1.36x that projection. Even that is only an extrapolation: fixture order is deterministic, the remaining 84 mutations are manager-heavy, and the comment itself says suite identity dominates cost. There is no evidence the completed prefix represents the remainder. If the remaining suites are costlier per mutation, this prefix-average biases the projection low. No run observed the whole set.

The comparison arithmetic is otherwise close: `183/39 × 7m27s` is 34.96 minutes and `183/1714 × 360` is 38.44 minutes.

## 4. Durable comment claims

The durable comment claims more than is measured in three places: `109 ... completing`, the derived whole-set runtime/headroom, and the step-timeout `failure` plus normal continuation. These need to be corrected or explicitly labelled as estimates/residuals. The correct fixture provenance is worth keeping.

## Verification and scope

I ran API measurements and arithmetic against the named jobs and full 40-character SHAs, and read the current GitHub workflow syntax plus `scripts/pr-head-gate.mjs`. I did not run typecheck because the diff is one workflow file and touches no TypeScript. `git diff --check` passed. The worktree is clean.

Exact-head workflow presence was separately enumerated: `CI`, `Code Quality: PR #1309`, `Mutation reproof`, and `Windows` are present by name; `Docs` is path-filtered. At review time Mutation reproof, Code Quality, and Windows were successful; CI was still `in_progress`, with its four smoke shard jobs incomplete, so exact-head CI remains unmeasured rather than green.

davidfarah2003:
Both blocking findings are correct. I confirmed each from source rather than accepting the report, and both are fixed at `3c453940dafcc6763afc098cfb37eff939c2ff5a`, one commit on top of `ff081dd21` with no rebase.

## 1. The count is 99, not 109. The reviewer is right and I was wrong.

I counted with `grep -c 'KILLED'`, which counts **lines containing the token anywhere**. A prediction line reads `PREDICTED KILLED (>=1 red) - ...`, so ten lines whose only match was a prediction preamble were counted as completed mutations.

Counting by leading verdict token instead, using the vocabulary at `scripts/mutation-reproof.mjs:217`:

```
d1db29e23   KILLED 91 · INCONCLUSIVE 3 · UNGRADABLE 1 · WRONG-RED 4 · SURVIVED 0 · ERROR 0 · PRE-RED 0
            = 99 completed mutations
```

Command, so it can be checked rather than believed:

```
sed 's/\x1b\[[0-9;]*m//g' <job.log> \
  | grep -cE '^[0-9T:.Z-]* *(KILLED|SURVIVED|INCONCLUSIVE|UNGRADABLE|WRONG-RED|ERROR|PRE-RED) '
```

Redoing `ec6f1f7b2` (#1302) the same way gives **85**, not the 93 I reported in the comment above. That figure was wrong for the identical reason and I have corrected it in the same commit.

**This moves the headroom again, in the direction the reviewer implied:**

| run | completed | rate | 183-mut projection | headroom at 150 |
|---|---|---|---|---|
| `d1db29e23` | 99 | 35.9 s/mut | ~110 min | **1.37x** |
| `ec6f1f7b2` | 85 | 41.9 s/mut | ~128 min | **1.17x** |

1.17x is thin. The comment now says so, and says to plan against the slower figure. I am still not raising 150 on this: the whole lesson of the original defect is that a number moved on thin evidence acquires false authority, and the instrument that should justify any further move is the elapsed line landing in this PR. The step timeout also makes a future breach visible rather than silent.

The comment now also records **how to count and why the naive grep overstates it**, because a provenance error inside a provenance fix is the failure that would most undermine it.

## 2. The gate does not distinguish cancelled from failure. Also correct.

Verified at source, `scripts/pr-head-gate.mjs`:

```js
131:    if (run.conclusion !== "success") failing.push(name);
```

That is the entire conclusion handling — one comparison, no branch — and a grep for `cancelled|timed_out|neutral` across the file returns only that line. So `cancelled` and `failure` are identical to the gate and its verdict is byte-identical either way. My stated rationale was false, and I asserted a mechanism I had not read.

**The change survives on its other justification, which was independently argued in the diff and is unaffected:** a failed step leaves the job running so `Report elapsed` executes, where a cancelled job gives its remaining steps a grace period at best. That was always the substantive reason; I dressed it in a second one that was not true.

The rationale is rewritten around the benefit that holds, and it now records the false claim and the source line that refutes it, so nobody re-derives it.

## The margin question

Separately confirmed clean by the reviewer: pre-step time measured 47-74s across 30 runs, so 145 under 150 has large room. That concern is closed by measurement rather than argument.

davidfarah2003:
Confidence re-audit at unchanged head `ff081dd2157bcce384eb679c172d3fd074317065` confirms the original BLOCK and adds one correction to the new follow-up comment.

I re-ran the margin query: 30 jobs, 47–74s pre-step wall time, with 226s remaining between a 145-minute step timeout and the 150-minute job cap at the observed worst case. I also rechecked all five #1294 jobs, the GitHub Docs wording, the gate predicate, both exact merge-ref trees, and publication state. The original review's findings stand: the gate treats `cancelled` and `failure` identically, and run `33978243970` completed 99 mutations, not 109.

The new #1302 datum is also overstated. Job `101355471647` has 21 fixture headings. The first 20 completed, and their terminal summaries total **85 mutations**. The 21st, `persona-role-capability.mutations.json`, printed only its heading before cancellation. Counting individual terminal verdicts independently gives the same 85: 81 KILLED, 1 UNGRADABLE, and 3 INCONCLUSIVE. There is no observed basis for 93 completed mutations in that log.

For its 18:35:56Z → 19:35:15Z re-prove window, 85 completed mutations means **41.9 s/completed mutation**, not 38.3. Projecting that rate over 183 mutations gives about 127.7 minutes for re-prove and 128.6 minutes including the observed 54s setup, leaving about **1.17x** headroom at 150. This remains a prefix extrapolation, not an observed whole-set duration.

This strengthens rather than relaxes the provenance block. Both durable progress numerators, 109 and 93, count more than the logs show as completed.

davidfarah2003:
Confidence-audit follow-up at unchanged head `ff081dd2157bcce384eb679c172d3fd074317065`: the new corroboration comment contains the same counting error as the workflow comment, so it does not weaken the block.

For job `101355471647`, the re-prove step ran `18:35:56Z` to `19:35:15Z` and then was cancelled. Its log has 21 fixture headings. The first 20 fixtures completed and their terminal summaries total **85 mutations**, independently cross-checked as 81 `KILLED` + 1 `UNGRADABLE` + 3 `INCONCLUSIVE` verdicts. The 21st fixture, `persona-role-capability.mutations.json`, printed only its heading before cancellation. It has 4 mutations. Thus the evidence supports 85 completed mutations, or at most 89 mutations in fixtures started, not 93 completed.

Using 3,559 seconds / 85 completed mutations gives **41.9 s/mutation**. Applying that prefix rate to 183 mutations projects about **127.7 minutes** for re-prove and about **128.6 minutes** including this run's 54-second pre-step time. The 150-minute budget is then about **1.17x** that extrapolation, not 1.28x. This remains only a prefix extrapolation, not an observed whole-set runtime.

Also, 38.3 is not within about 9% of 32.6. It is 17.5% higher than 32.6, 14.9% lower relative to 38.3, or a 16.1% symmetric difference.

The earlier confidence audit reconfirmed the original review: 30-run pre-step range 47–74s; all five #1294 jobs and re-prove steps concluded `cancelled`; GitHub's cited docs only promise killing the step process; `pr-head-gate.mjs` applies the same `conclusion !== "success"` predicate to `cancelled` and `failure`; the named #1294 merge-ref log supports 99 completed of 183, not 109. Worktree remains clean. **BLOCK remains.**

davidfarah2003:
**BLOCK** on the correction delta at exact head `3c453940dafcc6763afc098cfb37eff939c2ff5a`, for one narrow remaining provenance defect.

The two original blockers are fixed. I independently counted timestamp-anchored leading verdict tokens, after stripping ANSI:

- `d1db29e23`: 91 KILLED + 3 INCONCLUSIVE + 1 UNGRADABLE + 4 WRONG-RED = **99**.
- `ec6f1f7b2`: 81 KILLED + 3 INCONCLUSIVE + 1 UNGRADABLE = **85**.

The corrected rates also re-derive: 35.9 and 41.9 s/completed mutation. The slower prefix projects about 127.7 minutes for re-prove and gives 1.17x headroom at 150. Planning against the slower set is the right conservative choice. I agree with not raising 150 again from two censored prefix measurements. The comment labels 1.17x thin and requires a complete elapsed measurement before widening.

The false gate rationale is gone. The rewritten comment explicitly says the gate result is identical, quotes the single `conclusion !== "success"` predicate, and limits the step timeout's purpose to keeping the job alive for `Report elapsed`.

The remaining defect is in the new counting instruction itself: it lists `PRE-RED` as a completed-mutation verdict and says that vocabulary is at `scripts/mutation-reproof.mjs:217`. Line 217 actually defines only `KILLED`, `SURVIVED`, `INCONCLUSIVE`, `UNGRADABLE`, `WRONG-RED`, and `ERROR`. `PRE-RED` is a fixture-level exit-4 classification printed later, before any mutation ran. Counting it as a completed mutation would recreate the provenance error on a future pre-red set. Remove `PRE-RED` from that list, or state separately that pre-red fixtures contribute zero completed mutations.

The first two real 150-minute high-fan-in jobs, `101365519925` and `101365422787`, are currently in their re-prove steps and both expose pending step 11 `Report elapsed`. I would not hold this correction solely for them because the budget is already live and this delta repairs its durable record. If either reports before merge, however, its complete elapsed measurement should replace the censored prefix estimate in the comment as the comment itself requires.

Exact-head CI is not yet green: Mutation reproof and Code Quality are successful; CI is in progress and Windows is queued. `git diff --check` passed, and the worktree is clean.

davidfarah2003:
**APPROVE** the final bounded delta at exact head `8de1838dda12b4c33ffa9fb680eacd57adaa1d83`.

The remaining provenance defect is fixed. The comment now makes `scripts/mutation-reproof.mjs`'s `VERDICTS` array authoritative, lists its six current mutation-level tokens accurately, and explicitly states that `PRE-RED (N ...)` is a fixture-level summary contributing zero completed mutations. It also records why a naive trailing-space alternation would match that summary and overcount it.

This preserves the independently verified historical totals of 99 and 85. Both named logs contain zero `PRE-RED (` summaries, so neither total changes.

The final delta is one commit and one workflow file. `git diff --check` passed and the worktree is clean. Exact-head workflows are all present by name, but the new runs are currently queued or pending, so CI is unmeasured rather than green. That does not change the code-review approval of this bounded correction.

davidfarah2003:
Final control addendum for approved head `8de1838dda12b4c33ffa9fb680eacd57adaa1d83`:

- Running the corrected six-token alternation returns false for `PRE-RED (2 fixture(s))` and false for `PREDICTED KILLED (1 red)`, while a real `KILLED       ...` verdict returns true.
- The named snapshot `99d50f5100616571fab5ecdf9b9552213fa163ba` contains exactly the six listed tokens in `VERDICTS`: KILLED, SURVIVED, INCONCLUSIVE, UNGRADABLE, WRONG-RED, ERROR.
- The arithmetic follows from the independently verified counts: 3,556/99 = 35.9 s/mutation and ~1.37x re-prove headroom; 3,559/85 = 41.9 s/mutation and ~1.17x.

The final **APPROVE** stands.

Issue #1297, opened 2026-09-05
smoke shards grade a suite on exit status alone, so a suite that asserts nothing counts as a covered pass


davidfarah2003:
## Summary

`bin/smoke/shard.mjs` decides whether a suite passed from its **exit status alone**. A suite that
returns 0 without asserting anything is indistinguishable, to the runner and to every later reader of
the log, from a suite that ran every cell and passed them all. The shard prints
`✓ smoke:ci shard N/4 passed (123 smokes)` either way.

The check is one line:

```js
// bin/smoke/shard.mjs:61
if (r.status !== 0) {
```

There is no completion-banner check and no asserted-cell count.

## Why this is worth filing, given the file already knows

This is not a principle the runner is missing. It is a principle the runner already states and
enforces — twice — with the middle case left out.

**At the chain level it is enforced**, with the reasoning written out at `bin/smoke/shard.mjs:26-27`:

> An EMPTY chain is an error, not a fast green: a runner that finds no suites and exits 0 reports
> the same thing as a runner that passed all of them.

and acted on at `:32` with `exit(2)`.

**On the way out it is enforced**, at `:71-81`:

> A suite that passes its assertions and leaves a broker running is a FALSE GREEN, so it fails the
> shard.

**In between, it is not.** A suite that asserts nothing and exits 0 is a false green of the same
family, and nothing looks at it. The empty-chain guard asks "did any suite run?" and the leak check
asks "what did the suite leave behind?" Neither asks "did the suite actually assert anything?"

## How it surfaced

A reviewer grading an unrelated PR found it while confirming that four newly gated suites had really
executed. One of them, `legacy-pty-custody`, had just acquired a whole-suite platform skip:

```ts
if (process.platform !== "linux") {
  console.log("LEGACY PTY CUSTODY skipped (...)");
  process.exit(0);
}
```

That skip is correct on its own terms — the evidence it gathers is `/proc` state after `SIGKILL`, and
a Windows unit test is not that oracle. The problem is that the runner cannot tell that suite's
skip-and-exit-0 from a full pass, so a Windows shard reports the suite as covered. It is the first
fragment-gated suite with a platform skip, which is why it exposed this now; it is not the cause, and
nothing about it is wrong.

The same shape as #1098, one level down: there, a workflow that never started read as green because
absence produced no rows. Here, a suite that never asserted reads as green because exit 0 produces no
distinction.

## Not a duplicate of the adjacent issues

[...]

Issue #914, opened 2026-08-27
Windows / smoke reports success while executing a fraction of the suite: continue-on-error turns a shard's honest failure into a green


davidfarah2003:
## The finding

`Windows / smoke` reports `success` whether or not it ran the suite. Measured on run 32982811036, the `Windows` run of `ccd99461`:

| job | conclusion | suites executed | last suite started |
|---|---|---|---|
| Windows / smoke (shard 0/4) | success | 29 | `smoke:supervise-registered-user` |
| Windows / smoke (shard 1/4) | success | 3 | `smoke:schema-profile` |
| Windows / smoke (shard 2/4) | success | 21 | `smoke:attach-repaint` |
| Windows / smoke (shard 3/4) | success | 21 | `smoke:mesh-attach-plane` |

74 suites executed out of 385, and four green check-runs. Shard 1 stopped after three suites and reported success.

The mechanism is `.github/workflows/windows.yml:125`:

```yaml
      - name: Protocol & security smoke tests (shard ${{ matrix.shard }}/4)
        continue-on-error: true
        run: node bin/smoke/shard.mjs ${{ matrix.shard }} 4
```

`bin/smoke/shard.mjs` exits non-zero and stops at its first failing suite, which is correct. `continue-on-error: true` then converts that non-zero exit into a green step, and the job's conclusion follows. So the shard's own honest failure is discarded at the step boundary and replaced with `success`.

## What is NOT wrong, stated first so this is not over-read

**The required Windows status is unaffected and still means what it says.** `windows-ok` is the single required Windows check and it depends only on the `required` lane:

```yaml
  windows-ok:
    needs: [required]
    steps:
      - name: Gate
        run: |
          [ "${{ needs.required.result }}" = "success" ]
```

The `required` lane has no `continue-on-error` step anywhere in it. It covers typecheck, build, unit tests, the broker-free seam smokes, the `.cmd` launch matrix (`smoke:windows`), update crash barriers, control-plane seams, and secret-fs ACL hardening: the Windows-specific surface (PATHEXT and `.exe` resolution, `.cmd` argument quoting, ConPTY, NTFS ACLs) that a Linux runner cannot exercise. The sharded smoke lane is deliberately excluded from `windows-ok`, and the workflow says so:

> The sharded smoke lane is advisory (see the lane note at the top) and is deliberately NOT a dependency here, listing it would keep gating on it even while it is continue-on-error.

[...]

Issue #698, opened 2026-08-19
Windows smoke shards stop at the first red and still report success, so a green lane covers only a prefix


davidfarah2003:
The Windows workflow's `smoke` lane runs the same round-robin shards as CI
(`node bin/smoke/shard.mjs <shard> 4`), but it is marked non-blocking. `bin/smoke/shard.mjs` breaks
out of its loop on the first suite that exits non-zero. Those two facts combine into a green job that
ran a fraction of its list.

## Measured

Run 32224458752, the Windows smoke lane's shard 2 of 4 (job 95981238592, the non-blocking lane):

```
✗ shard 2/4 FAILED at: pnpm smoke:mesh-attach-plane (exit 1)
```

That is suite 21. The shard printed a plan of 81 suites at the top of the log and reached 21 of them.
The remaining ~60 never executed. The job's conclusion is `success`, and so is `windows-ok`.

The same shape appears in how a downstream suite looks from the outside. `smoke:attach-stdin` is in
that shard's plan and sits well past the abort, so the log contains exactly one `attach-stdin`
string, which is the plan listing rather than a run:

```
  pnpm smoke:attach-stdin
```

A reader grepping for the suite finds a hit in a green Windows job and can reasonably conclude it
ran and passed on Windows. It did not run at all.

## Why this is worth fixing rather than noting

A green Windows lane currently means "the first N suites of each shard passed, for some N nobody
reports". N is not stable: it moves whenever an earlier suite's status changes, so Windows coverage
silently expands and contracts with unrelated work, and a regression in suite 40 is indistinguishable
from a regression in suite 5 because neither is reached once suite 21 is red.

It also means a Windows-only defect in any downstream suite cannot be discovered while an earlier
suite stays red, which is a hole that widens the longer the earlier suite stays broken.

## What would close it

Either of these, and they are not exclusive:

- Do not break on first failure in the sharded runner. Record the failure, continue the list, and
  report the set at the end. A shard that stops on the first red reports a RED PREFIX rather than a
  red set, and "which suite is broken" becomes a question about ordering.
- Make the lane report what it actually ran: print `ran X of Y` and fail the job, or annotate it,
  when X is less than Y. A non-blocking lane can still be honest about its own coverage.

Fixing the underlying `smoke:mesh-attach-plane` failure on Windows would restore coverage today but
would not stop this recurring, since any future red in any shard re-creates it.

Found while checking whether an unrelated suite had Windows coverage for a CRLF question; it does
[...]

PR #1312, opened 2026-09-05
feat(seat): split Linux PTY custody into a detached per-seat custodian

davidfarah2003:
## Summary

Split Linux PTY ownership out of the replaceable manager worker.

A one-shot launcher starts one detached custodian per seat. The custodian owns the `node-pty` object, child relationship, screen mirror, and exit observation. The manager adopts a proxy `AgentHandle` over a 0600 filesystem Unix socket authenticated with `SO_PEERCRED` and a per-seat capability token. Manager-worker death no longer closes the child.

Linux `pty` spawn uses the custodial runtime. Non-Linux `pty` spawn remains in-process for compatibility, while custody adoption throws `custody transport unsupported on <platform>`. Linux custody failures do not silently fall back to in-process ownership.

## Packaging and protocol

- Add `@cotal-ai/seat` as a leaf package and include it in first-party install and tarball closures.
- Ship prebuilt Linux x64 and arm64 peer-credential modules. Customer installation does not compile native code.
- Assemble both native targets into one package, refuse a host-only pack, and verify the same packed artifact loads on Linux arm64 without a compiler toolchain.
- Keep launch credentials and environment data out of the world-readable process command line.
- Bound each local protocol frame and the undecoded residual buffer while preserving large coloured screen snapshots and malformed-frame refusal.
- Close controller sockets after exit settlement so a manager worker can terminate cleanly.
- Bound CLI manager-control cleanup after broker disappearance so command shutdown cannot hang.
- Refresh the version-matched connector docs for the custody model.

## Verification

- Add lifecycle, protocol, platform, isolation, custodial-runtime, attach/replay, packaging, and CI-wiring smoke coverage.
- Gate new suites through hashed `ci-suites.d` fragments rather than changing the frozen shard list.
- Mutation fixtures cover production adoption, lifecycle, frame bounds, packaging, late exit observers, early output replay, and wait-exit closure.
- Linux x64 builds the assembled seat package once; Linux arm64 downloads and loads that same artifact.

This is the M2 custody milestone. Controller-generation CAS, crash recovery, protocol N/N-1 handoff, and manager-worker activation remain later work.

Refs #783


davidfarah2003:
## Independent cold review — **BLOCK**

Reviewed at exact head `6bd7398ace976d5637b330b5d3f1957149502153` (confirmed via `git rev-parse` and `gh pr view` at start). Note: the PR head **moved to `95e123adb0d5afbde421e3a5d33a8b2fcb239a7d` at 21:35Z during this review** (commits `a85001eb` fix non-Linux pty spawn, `15eb6cc0` include seat in tarball closures, `test(seat) 95e123ad`) — those touch findings 1 and 4 below, so this verdict is against `6bd7398` and a re-review of the new head is warranted. Every finding below is from running, not only reading.

### 1. Install closure — native compile in the customer install path (BLOCK-worthy)
`packages/seat/package.json` declares `"install": "node scripts/build-native.mjs"`. Closure verified: `implementations/manager` depends `@cotal-ai/seat: workspace:*`, and `bin` (published `cotal-ai`) depends `@cotal-ai/manager` → this runs on `npm i -g cotal-ai`.

**I converted the inference to a fact.** In a sandboxed prefix laid out as the published tree (`prefix/node_modules/@cotal-ai/seat`), with Node present but no `cc`/`gcc`/`node-gyp` and no `libnode-dev`:
- The `node-gyp` probe resolves to `.../node_modules/node_modules/.bin/node-gyp` → does not exist (published layout has no nested `node_modules`), so it falls back to bare `node-gyp` on PATH.
- Bare `node-gyp` absent → `cc -I/usr/include/node` fallback → `cc` absent.
- Result: **`build-native.mjs` exits 1, silently, with no diagnostic** (`process.exit(cc.status ?? 1)`; the failed spawn's `status` is `null`). A non-zero `install` exit **fails the whole `npm i -g cotal-ai`**.

So a Linux customer with no C toolchain and no libnode-dev cannot install Cotal at all, and gets no message saying why. This violates AGENTS.md "keep the customer update path clean and secure" and "no fallbacks / fail loud."

**The `cc` branch is the degradation the rule forbids.** On THIS reviewer box the fallback silently succeeds only because Debian `libnode-dev` happens to provide `/usr/include/node/node_api.h` — a path that is *not* node-gyp's own header layout (`~/.cache/node-gyp/<ver>`). Compiling against whatever headers happen to be present is exactly the silent degradation "no fallbacks" exists to prevent.

**Prebuilt-vs-build-on-install is incoherent (`npm pack --dry-run`).** The tarball ships a compiled `build/Release/peercred.node` **and** build cruft (`*.o`, `*.d`, `binding.Makefile`, `config.gypi`, `peercred.target.mk` — 47 files total), **and** the `install` script that rebuilds unconditionally (it never checks for an existing artifact). The shipped `.node` is the CI runner's glibc/x64 build, ABI-tied to the runner's Node, with **no `os`/`cpu`/`libc` gating and no `optionalDependencies`/prebuild-install pattern**. It breaks on musl/Alpine and on a different Node major, and the install rebuild can fail regardless. It is both a prebuilt and a build-on-install, and neither cleanly.

### 2. `SO_PEERCRED` auth boundary — solid, fails closed ✅
Read `custodian.ts`, `peercred.ts`, `protocol.ts`, `client.ts`, `native/peercred.c` as a security surface, and ran `smoke:seat-protocol` (SO_PEERCRED uid/pid asserted, 5/5).
- **Checked on every request**, not once at listen (`handle()` calls `peerCredentials(sock)` per message). No first-connection hole.
- **uid equality** against `process.getuid()`; `getuid` undefined or mismatch → error + `sock.destroy()`. Correct primitive (SO_PEERCRED is unspoofable kernel truth).
- **Fails closed**: if the native helper is missing, `loadNative()` throws → caught in `handle()` → error + `destroy()`, i.e. **refuse, never allow**. `socketFd()` throwing on absent `_handle.fd` is on the same fail-closed path.
- **Path/perms**: socket lives in `join(root, <16-byte-random-id>, "seat.sock")`; dirs `0o700`, socket `chmod 0o600`, token 32-byte random. Predictable-path squat/swap is mitigated. Residual: a tiny TOCTOU between `server.listen()` and `chmodSync(0o600)`, but the `0o700` parent dir makes it unreachable by other users. Acceptable.
- **`peercred.c`**: argc + int32 validation, `getsockopt` return checked → JS throw (not a crash) on a bad fd. No native crash path found.

### 3. Per-seat constraint — proven adversarially ✅
Ran `smoke:seat-isolation` (5/5): two seats with **distinct custodian pids and sockets**; `SIGKILL` alpha's custodian; asserts alpha gone; **beta still reads and writes** afterward. A single all-seat daemon would fail this. The suite proves what its name claims.

### 4. Milestone claims — satisfied where runnable ✅ (but see platform note)
- **First real production `adopt()` caller exists.** `createRuntime("pty"|"auto")` returns `CustodialPtyRuntime`, whose `adopt()` calls `adoptSeatSync(loadSeat(...))`. `smoke:custodial-pty` (3/3) reaches the REFUSAL through production `requireRuntimeAdopt` with a fixture runtime lacking `adopt` (`absent adopt refuses by runtime name`) — genuine reachability, not a regex.
- **Positive PID survival — the real M2 criterion — passes.** `smoke:seat-lifecycle` (12/12): `worker death leaves the custodial child PID running` + `adopt: successor proxy talks to the surviving seat`, read from `/proc`.
- **Seven behaviours each a cell**: spawn, snapshot, concurrent read, input, resize, natural exit, graceful stop, hard stop — all present and green in lifecycle.
- **`legacy-pty-custody.smoke.ts:78`** still asserts the M1-red opposite (legacy in-process runtime child dies on manager death) using `LegacyPtyRuntime` from `dist`. Ran it **3× — stable OK (5/5)**. The sibling-lane "child still live after SIGKILL" is an environment flake here, not the new behaviour leaking into the legacy fixture.
- **Platform honesty**: `platform.smoke.ts` **closes** #1297 for itself — no top-of-file skip, it runs real assertions on every platform and additionally asserts `launchSeat` throws off-Linux. `isolation`/`lifecycle` keep the top-of-file `exit(0)` skip on non-Linux (inherits #1297, which is tracked, not re-filed).

### 5. Mutation fixtures — well-formed ✅
`pnpm smoke:mutation-fixtures` (did NOT run `mutation-proof`): `packages/seat/smoke/mutations/lifecycle.json` — 1 anchor present+unique; `implementations/manager/smoke/mutations/custodial-pty.json` — 2 anchors present+unique. By reading: all anchors are **code-only** (no comment/message-string pinning), each resolves once, and the named `expectRed` is the **earliest** cell that notices in all three mutations.

### 6. Suites, typecheck, and the disqualifying CI fact
- `pnpm typecheck` → **green (exit 0)**. Deterministic scope: **27 packages carry a typecheck script and `packages/seat` is one of them** (the old "27 of 28" figure is stale; the new package is inside the scope and green). The prior sentinel figure does not apply.
- Ran green here: `seat-platform` (3/3), `seat-protocol` (5/5), `seat-lifecycle` (12/12), `seat-isolation` (5/5), `custodial-pty` (3/3), `legacy-pty-custody` (5/5 ×3). **Author-unnamed suite I picked: `smoke:attach`** (green) — chosen because this PR turns it from a bare `tsx` into a six-package build that now includes the seat native compile inside a CI-required suite; it passes and reaches the adopt seams. On Windows the seat build's `build-native.mjs` no-ops (exit 0) so the build is payable, but the `pty adopt returns a live proxy` cell needs the custodian, which throws `unsupportedTransport` on win32 (Windows shards are non-blocking).

### **The primary BLOCK — the new suites are not gated in CI**
`smoke:gate-inventory` is a CI gate (part of `smoke (shard 1/4)`), and it is **already RED at this head** — reproduced locally:
```
✗ FAIL: 5 suite(s) exist but nothing runs them, and they are not in UNGATED:
    smoke:custodial-pty
    smoke:seat-isolation
    smoke:seat-lifecycle
    smoke:seat-platform
    smoke:seat-protocol
GATE INVENTORY FAILED ❌
```
Confirmed by inspection: none of the five appear in the `smoke:ci` chain, and none are in `UNGATED`. **The PR's own milestone-verification suites never run in CI** and the gate guard fails outright. This is deterministic, not a flake.

### CI enumeration (full 40-char sha, runs API)
`check-runs` for `6bd7398…`: `CI`, `Docs`, `Mutation reproof`, `Windows`, `Code Quality` are all **PRESENT by name**. Quantity separate from conclusion: `smoke (shard 1/4)` = **completed/failure** (the gate-inventory red above); shard 0/4 in_progress; shards 2–3, `Docs`, `Mutation reproof`, `Analyze`, `Windows` = **queued/in_progress = UNMEASURED**. `full` = skipped. `greencheck.sh` verdict: **NOT GREEN**.

### Could not check
- End-to-end `npm i -g cotal-ai` on a pristine musl/Alpine host (no container runtime here); I proved the failing link — the `install` script's non-zero exit with no toolchain — directly, which is the mechanism.
- Did not run `mutation-proof` (per policy) or any `-live`/`:live` suite; graded live suites and downstream CI conclusions by reading.
- Queued/in-progress CI jobs remain unmeasured by definition.

### Verdict
**BLOCK.** Two independent blockers at `6bd7398`: (1) a native compile in the customer install closure that fails `npm i -g cotal-ai` silently on any Linux host without a C toolchain + libnode-dev, with an incoherent prebuilt/build-on-install split and no platform/ABI gating; (2) the five new milestone suites are not gated in `smoke:ci`, so `smoke:gate-inventory` is red and the suites never execute in CI. The seat runtime, auth boundary, isolation, and adopt/PID-survival behaviour are otherwise sound and green when run directly. The new head `95e123a` (tarball-closure + non-Linux-spawn fixes) may address parts of (1) — re-review it.

davidfarah2003:
### Validation addendum (same head `6bd7398`)

Two verification passes that sharpen — do not change — the BLOCK above. Both from running.

**Auth boundary is stronger than I first stated (§2 refinement, not a new problem).** I ran an adversarial same-uid client against a live custodian socket (it therefore passes the SO_PEERCRED uid check, since it runs as my own uid) and tried each privileged op:
```
write-before-hello     => {"ok":false,"error":"not authenticated"}
hello-wrong-token      => {"ok":false,"error":"capability token mismatch"}
snapshot-before-hello  => {"ok":false,"error":"not authenticated"}
stop-before-hello      => {"ok":false,"error":"not authenticated"}
hello-correct-token    => {"ok":true,"op":"hello",...}
```
So the boundary is two-factor: SO_PEERCRED uid equality (same user) **and** the per-seat 32-byte capability token from the `0600` record. A same-uid local peer that reaches the socket still cannot drive, snapshot, or `stop` the seat without the token. No op is reachable before `hello`. Fails closed, confirmed empirically.

**The gate-inventory red is PR-introduced, correctly attributed.** At the merge-base with `main` (`f3bad87`) the seat suites did not exist; this PR adds `packages/seat` (+1618 lines / 21 files in that path) and defines `smoke:{custodial-pty,seat-platform,seat-protocol,seat-lifecycle,seat-isolation}` **without adding any to the `smoke:ci` chain or `UNGATED`**. That is exactly what `smoke:gate-inventory` fails on, and it is this PR's change, not a pre-existing base failure.

**Packaging incoherence, both halves verified.** `packages/seat/build/` is gitignored, so the `build/Release/peercred.node` in `npm pack` comes only from `prepublishOnly` on the publish host (a glibc/x64, runner-Node-ABI artifact). And `build-native.mjs` has no existing-artifact guard — it rebuilds unconditionally on `install`, so the shipped prebuilt is always overwritten or the rebuild fails. Prebuilt and build-on-install, neither cleanly, no os/cpu/libc gating.

Verdict unchanged: **BLOCK** at `6bd7398`.

davidfarah2003:
### Additional finding — unbounded pre-auth buffering in the custodian frame reader (§2)

Probing the custodian socket as a security surface, I found a same-uid local DoS. Reporting it because §2 asked me to attack this boundary, and I could not otherwise call it clean.

**What.** `FrameReader.push` (`packages/seat/src/protocol.ts:68`) accumulates every `data` chunk into `this.buf` via `Buffer.concat`, keyed on a client-supplied UInt32 length header, **with no maximum frame size**. In `custodian.ts:181-200` this runs in the `sock.on("data")` handler and assembles the whole frame **before** any message reaches `handle()`, where the `peerCredentials()`/uid check and the token gate live. So the buffering is entirely **pre-authentication** — a connecting peer needs no `hello` and no token to make the custodian buffer.

**Proved it.** Against a live custodian I opened a socket, wrote a 4-byte header claiming a 500 MB frame, then dribbled ~8 MB of body and stopped. Custodian RSS:
```
before= 57032 kB   afterA= 157468 kB   delta= +100436 kB   (~100 MB from ~8 MB sent, pre-auth)
```
It grows with whatever the client sends and never completes a frame, so the memory is held. A same-uid process can drive the per-seat custodian to OOM and take the seat down with it (one custodian per seat). The `0o700` seat dir bounds the attacker to the same uid — so this is a local same-user DoS / defence-in-depth gap, not cross-user escalation — but no capability token is required, unlike every op in `handle()`.

**What is fine (also tested).** A malformed-JSON body inside a valid-length frame is handled safely: the socket is destroyed and the custodian stays alive (`reader.push` throw → `sock.destroy()`; custodian still alive: true). Only the missing size cap is the issue.

**Fix shape.** Cap `size` in `FrameReader` (reject/`destroy` above a few MB) and/or a total-buffer cap, so an unauthenticated peer cannot force unbounded growth. Small change; it closes the one hole on an otherwise fail-closed, two-factor boundary.

This does not change the verdict — the two blockers already carry it — but it should be fixed on the same auth surface. Still **BLOCK** at `6bd7398`.

davidfarah2003:
## Delta review — head `95e123adb0d5afbde421e3a5d33a8b2fcb239a7d` (from my graded `6bd7398`)

Read-only delta (a memory hold on the box precludes running suites/build/install this pass; anything needing a run is named unverified-pending-resources below). Seat package, SO_PEERCRED, and the install script carry over by blob identity — I confirmed `packages/seat/{package.json,scripts/build-native.mjs,src/protocol.ts}` are byte-unchanged in the delta, so my findings 1, 2 and the DoS still apply.

### The reversal is correct on its own terms ✅
**Ruled split landed.** `createRuntime("pty")` now branches: `process.platform === "linux"` → `CustodialPtyRuntime`, else `LegacyPtyRuntime` (`runtime/index.ts`). That is the fix for the required-Windows red (`Error: custody transport unsupported on win32` was thrown at *spawn* time). Off Linux the manager now spawns in-process and only `adopt` refuses.

**No silent Linux fallback to in-process.** `this.runtime` is set exactly once, at `manager.ts:1012` via `createRuntime`, and `LegacyPtyRuntime`/`PtyRuntime` is instantiated nowhere in production outside smoke fixtures. On Linux the branch returns `CustodialPtyRuntime` with no `try/catch` that could degrade to legacy; a constructor failure propagates. So a Linux manager cannot quietly run the worker-owned PTY.

**`custodied` has a real writer, not a widened-enum-with-no-writer.** `manager.ts:4942`: `custody: process.platform === "linux" && this.runtime.kind === "pty" ? "custodied" : "legacy"`, and the contract widens the enum to `["legacy","custodied"]`. `LegacyPtyRuntime.adopt()` now `throw unsupportedTransport()` — off-Linux adopt fails loud rather than returning a dead handle. (Minor: the writer keys on `runtime.kind === "pty"`, which both pty runtimes share; it is truthful only because Custodial is the sole Linux+pty production runtime. If a second non-custodial pty runtime ever ships, this would misreport. One-line note, not a blocker.)

**Coverage proves the split, mutation fixtures well-formed.** `attach.smoke.ts` and `custodial-pty.smoke.ts` now branch by platform and add: `status output accepts custodied / still accepts legacy / refuses unknown custody` (against the compiled contract) and `in-process pty adopt throws custody transport unsupported` (instantiates `LegacyPtyRuntime` directly). The mutation fixture grows 2→4: M3 (swallow the in-process adopt throw) and M4 (drop `custodied` from the enum). I verified both new anchors are **code-only and resolve exactly once**, and each `expectRed` is the only cell exercising that code, so it is the earliest to notice. (M3's `expectRed` hardcodes `…on linux` while the cell text is templated on `process.platform`; they match on the Linux reproof runner.)

**Docs kept in the same change.** `AGENTS.md` and `packages/seat/README.md` now state "the manager still spawns pty seats in-process off Linux and only `adopt` throws." The `isolation.smoke.ts` 2-line change adds `ha.close()`/`hb.close()` after their assertions — harmless proxy-cleanup, no assertion weakened.

### The two blockers persist at this head ❌ (verdict unchanged: BLOCK)
1. **Native compile in the customer install closure.** `packages/seat/package.json` + `build-native.mjs` are byte-unchanged. `npm i -g cotal-ai` still runs the `install` script, which I reproduced exits 1 **silently** with no toolchain + no libnode-dev, failing the global install; the tarball still ships a prebuilt `.node` + build cruft yet rebuilds unconditionally with no os/cpu/libc gating. The delta's `15eb6cc0` adds `packages/seat` to the first-party **tarball-closure inventory lists** (`dep-pins`/`dist-freshness` PUBLISHED+PACKAGES, `legacy-packaged-manager` dir list 8→9) — that closes the E404/closure-membership gap, but it does **not** touch the install script or its failure mode.
2. **The five seat suites are still ungated.** At `95e123a`, none of `smoke:{custodial-pty,seat-platform,seat-protocol,seat-lifecycle,seat-isolation}` are in the `smoke:ci` chain or `UNGATED`, so `smoke:gate-inventory` stays red. The suites that prove the milestone are invisible to the gate that would catch a regression in it.

### Unverified-pending-resources (memory hold — clear when the floor lifts)
- **`Mutation reproof` on `bin/smoke/mutations/manager-runtime-deps.json`** reported `0 marks (baseline 51)` at `6bd7398` — an inconclusive (suite died before any mark), not a kill. That fixture guards the manager's published runtime-dependency closure, which this PR grows by adding `@cotal-ai/seat` (and transitively `@lydell/node-pty`, `@xterm/*`). The `15eb6cc0` closure-list additions are directionally the right fix, but whether they make the reproof reach `baseline 51` again needs `pnpm smoke:manager-runtime-deps` + the reproof run. Named, not run.
- Fresh CI at `95e123a` is still running: I enumerated the check-runs and the blocking jobs (`smoke` shards 0-3, `Windows / required`, `changed`) are **queued / in_progress = UNMEASURED** right now — not green, not red. `docs` = success. The reversal makes the Windows spawn-red *expected to clear*, but that is a prediction until `Windows / required` completes, and `smoke (shard 1/4)` will still carry the gate-inventory red above unless the suites are gated. I will not call an incomplete run either way.

### Verdict
**BLOCK** at `95e123adb0d5afbde421e3a5d33a8b2fcb239a7d`. The reversal cleanly fixes the Windows-required red and lands the ruled split with honest coverage, but the two install/gate blockers are unchanged at this head. Clear both (an install script that fails loud with a real remediation or a prebuilt strategy with platform gating; and gate the five suites in `smoke:ci` or `UNGATED` them with a reason) and the runtime work behind them is sound.

davidfarah2003:
### Settling the two unverified items (memory hold lifted) + one latent defect for the record

**1. `smoke:manager-runtime-deps` — the `0 marks (baseline 51)` reproof failure does NOT share the packaging root cause.** I built the manager dist faithfully from the `95e123a` runtime source and ran the suite:
```
SUITE COMPLETE: 56 passed, 0 failed
```
It is **green**, not red. Mechanism: the suite walks the built `dist/index.js` import graph and requires every external it loads to be a manager production dependency. At this head the reversal's new `import { LegacyPtyRuntime } from "./pty.js"` in `runtime/index.ts` makes the walk reach `pty.js`, which pulls `@lydell/node-pty`, `@xterm/headless`, `@xterm/addon-serialize` into the closure — and all three are already `dependencies` of `@cotal-ai/manager`, as is `@cotal-ai/seat`. That is why the count rises 51→56, all passing. So the PR's dependency closure is correct.

The `0 marks (baseline 51)` signature is a mutation-proof artefact, not a suite failure: the suite's first check is `manager dist entry exists`, and if the reproof mutates source without rebuilding `dist`, or a mutant dies before the walk, the run emits ~0 marks and never reaches the 51-cell baseline — an INCONCLUSIVE wearing red, exactly as the reproof output implied. It is a build/ordering matter in the reproof harness, not this PR's packaging. (Confirmed the same-head, byte-identical `manager/package.json` and suite carry over to `95e123a`.)

**2. Fresh CI at `95e123a` — Windows evidence is turning green; blocking jobs still unmeasured.** Enumerated the check-runs: `Windows / smoke` non-blocking shards 0/1/2 and soak are **completed/success**, and `docs` is success — direct evidence the reversal cleared the `custody transport unsupported on win32` spawn-red I named. `Windows / required` and `changed` are `in_progress`; `smoke` shards 0-3, `unit`, `live` are `queued` = UNMEASURED. `smoke (shard 1/4)` will still carry the gate-inventory red until the five suites are gated. Five workflows dispatched (CI, Code Quality, Docs, Mutation reproof, Windows); Docs is present-and-green here because this head's delta touches `docs/**` (`docs/architecture.md`, `cli.md`, `run-a-mesh.md`), which the `docs.yml` path filter matches — so its presence is correct, not a spurious trigger.

**3. Latent defect for the record (non-blocking, named trigger).** The `custody` status writer at `manager.ts:4942` derives `"custodied"` from `process.platform === "linux" && this.runtime.kind === "pty"`. That is correct-by-coincidence: it reports custody but infers it from `kind === "pty"`, a proxy that is equivalent **only because `CustodialPtyRuntime` is currently the sole Linux+pty production runtime**. A second pty-kind runtime (plausibly what M3/M4 adds) would make this **silently misreport `custodied`** for a non-custodial handle — on a *published contract field* (`custody: enum ["legacy","custodied"]`) that older consumers validate against a closed enum. It will not fail loud when it breaks. Not a blocker for M2, but it is a latent defect with a named trigger, not a style note; the next milestone's reviewer should inherit it.

Verdict at `95e123a` unchanged: **BLOCK** on the two standing items (silent toolchain-less install failure; five seat suites ungated → gate-inventory red). The runtime work behind them is sound and the dependency closure is correct.

davidfarah2003:
## Security: launch payload is passed in argv, exposing the connector key and seat token to any uid

`packages/seat/src/launcher.ts:66` spawns the detached custodian with the entire launch payload as a
command-line argument:

```js
const payload = JSON.stringify({ id, name, command, args, env: opts.spec.env ?? {}, cwd,
                                 socket, token, recordPath, logPath, confirm });
const child = spawn(process.execPath, [custodianEntry(), payload], { detached: true, ... });
```

`env` and `token` therefore land in the custodian's `/proc/<pid>/cmdline`.

### The boundary that moves

Measured on a Linux host:

```
/proc/<pid>/cmdline   -r--r--r--   0444   world-readable, any uid
/proc/<pid>/environ   -r--------   0400   owner only
```

`implementations/manager/src/runtime/pty.ts` documents what `spec.env` carries: "the connector-declared
env (OS allow-list + identity + named model key)".

```
before:  spec.env -> pty.spawn(env)         -> child's environ,     0400, same uid only
after:   spec.env -> payload -> argv[2]     -> custodian's cmdline, 0444, any uid
```

The capability token crosses the same way; it is otherwise at rest in `record.json` at mode 0600 inside
a 0700 directory.

### Which half is bounded

The token exposure is bounded: `SO_PEERCRED` still refuses a different uid on the socket, so reading the
token does not by itself grant access. The model API key is not bounded — it needs no socket, and any
uid on the host can read it out of `cmdline`.

Stating the narrow claim deliberately: this does **not** weaken the same-uid position. The token is
already same-uid-readable by design — `packages/seat/src/client.ts:69` authenticates with
`this.record.token` read from that same 0600 record — so it was never a same-uid defence. The boundary
that changes is same-uid to any-uid.

### Suggested fix

Pass the payload on the custodian's **stdin**; a pipe never appears in procfs. A 0600 file path in argv
also works and can reuse the existing `record.ts` machinery. The process environment is not a fix for
the cross-uid case: `environ` is owner-only, which is better than argv, but still leaves the material
at rest.

### Control this needs

A cell that launches a real seat, reads the real `/proc/<custodianPid>/cmdline`, and asserts that
neither the token nor any `spec.env` **value** appears there — reading actual procfs, not a mock. Plus a
mutation cell, so reverting the transport reds that named cell first.

### Verified / not verified

Verified: the argv construction above; the token as the authentication secret at
`packages/seat/src/custodian.ts:240` (`capability token mismatch`, then `sock.destroy()`); the 0444 and
0400 mode bits; and that 557 other processes' cmdlines were readable at the same uid.

Cross-uid read: **executed**, not inferred. As an unprivileged uid (1000), the `cmdline` of four
**root-owned** processes was read successfully — pid 1 `/sbin/init`, `sshd`, `agetty`, and a
`netbird` daemon. Negative control on the same process: `/proc/1/environ` returned
`Permission denied`. So the positive and the negative both hold on one target, and the distinction is
not an artefact of a special process.

Note the direction: demonstrating "any uid can read this" does not require becoming another user, so
no `sudo` is involved. It requires an unprivileged process reading a **more** privileged one, which is
the stronger demonstration anyway.

davidfarah2003:
**APPROVE** exact head `1c12809e83ed259a24cb55c575520158869a8bfb` (re-read immediately before this comment: OPEN, MERGEABLE, CLEAN, still that full SHA). Reviewed the full span against live `main` `52901c6298f5b6893baf2dcfb4d0716799cad202` (merge-base identical, 0 behind).

**Hosted gates on this SHA**
- Linux CI run [34053012108](https://github.com/Cotal-AI/Cotal/actions/runs/34053012108): unit, live, seat native linux-x64/arm64, seat pack (assembled), seat load (linux-arm64, same artifact), four smoke shards, and `ci-ok` all **success**. `ci-ok` printed `unit=success smoke=success live=success seat-pack=success seat-linux-arm64=success`.
- Smoke shards: **138 + 127 + 135 + 130 = 530**. Each shard ended `passed (N smokes)` with a clean final reaper (`0 nats-server live, 0 not claimed`). No never-ran gap on the shard runners.
- Windows run [34053012094](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094): required, four non-blocking smoke shards, soak, and `windows-ok` all **success**.
- Mutation reproof [34053012107](https://github.com/Cotal-AI/Cotal/actions/runs/34053012107) `changed`: `52 fixture(s) selected from 360 (diff 78 record(s), 78 changed path(s), corpus 360)`. Fixture kill-summaries sum to **239**, all KILLED. Final line: `MUTATION REPROOF OK (52 fixture(s) selected; 52 discriminated, 0 pre-red, 0 inconclusive)`. Zero SURVIVED / unmeasured / adverse. `full` skipped by design.

**Architecture / security / packaging**
- Linux `createRuntime("pty")` is custodial only (`CustodialPtyRuntime` → detached launcher + `adoptSeatSync`). Off Linux, spawn stays in-process and `adopt` throws `custody transport unsupported on <platform>`. Linux custody does not fall back to in-process ownership.
- Local auth fails closed: real `getsockopt(SO_PEERCRED)` in `packages/seat/native/peercred.c`, uid match plus capability token, socket `0600`, record `0600` under `0700`. Launch JSON is inherited on stdin from a `0600` file, not argv.
- `FrameReader` refuses claimed length `> 8MiB` from the 4-byte header before the body is required, and bounds undecoded residual after drain. Hosted `smoke:seat-protocol` was 16/0; mutations killed oversized-frame refusal, max-frame-plus-small-frame coalesce, and cmdline-payload regression.
- Customer pack on this SHA contains both `package/build/Release/linux-x64/peercred.node` (x86-64 ELF) and `linux-arm64/peercred.node` (aarch64 ELF). No `binding.gyp`, no compile-on-install, no `.c`/`.o` in the tarball. Same artifact loaded on linux-arm64 with `cc`/`gcc`/`make`/`node-gyp` stubbed to 127: `SEAT_SHIPPED_NATIVE RESULT=PASS`. `prepack` refuses a host-only tree.
- New suites are gated only through hashed `bin/smoke/ci-suites.d/<sha256(name)>` fragments (`smoke:seat-lifecycle`, `seat-protocol`, `seat-platform`, `seat-isolation`, `seat-native-ci`, `custodial-pty`). Frozen suite list file is unchanged.
- `env-isolate` waits for custodial `status() === "exited"` before attach, so dump-then-exit capture must take subscribe replay. Hosted `env-isolate.json` killed both `M1 skip subscribe output replay` and `M2 skip early-buffer accumulation` in production `packages/seat/src/custodian.ts` (`early`, no test-only alias). Lifecycle M5 independently killed subscribe-replay on the seat suite.
- Docs (`docs/architecture.md` and generated connector docs bundle), seat README, package `files`, and changesets match this behavior. No AI attribution in commit messages.

**#1354 / gitlink**
Superproject `.internal` remains `261625e89852e51c4a867d7f1c0d27358723eeb6` on this head and on live `main`. That object is reachable on `Cotal-AI/cotal-internal` (`261625e89852e51c4a867d7f1c0d27358723eeb6`). Leaving the gitlink unchanged is correct; recovered internal history is already an ancestor of advertised internal `main` per closed #1354.

**Limitations (not merge blockers for M2)**
Controller-generation CAS, crash recovery, protocol N/N-1 handoff, and manager-worker activation remain later work, as the PR states. `health` reports `PROTOCOL_VERSION = 1`; hello does not negotiate N/N-1. Mutation reproof shows the selected suites depend on the mutated lines, including product attach/replay paths used by `env-isolate`.

davidfarah2003:
**BLOCK** exact head `1c12809e83ed259a24cb55c575520158869a8bfb`. Immediately before posting, the PR remained OPEN/CLEAN at that full SHA against `main` `52901c6298f5b6893baf2dcfb4d0716799cad202`.

The required Windows evidence is partial, not terminal green. In [run 34053012094](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094), the non-blocking wrappers conclude `success`, but their hosted logs show all four shard commands failed and stopped early:
- [0/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909137): `smoke:supervise-registered-user`, **109/138 never ran**.
- [1/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909134): `smoke:schema-profile`, **124/127 never ran**.
- [2/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909080): `smoke:attach-repaint`, **114/135 never ran**.
- [3/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909098): `smoke:goal-epoch-fence`, **114/130 never ran**.
`windows-ok` only reports `required=success`; it does not make those partial shards green. This fails the exact-head requirement that required and non-blocking Windows shards be terminal green with no partial or never-ran gap.

Secondary packaging metadata mismatch: this 78-path diff changes published `@cotal-ai/cli` source and the published `@cotal-ai/connector-core` generated docs bundle, while the five new changesets name only `cotal-ai`, `@cotal-ai/manager`, and `@cotal-ai/seat`. The repository policy requires every package with a first-party change to be listed.

The M2 implementation itself inspected sound: [Linux CI 34053012108](https://github.com/Cotal-AI/Cotal/actions/runs/34053012108) completed 138+127+135+130=530 with clean reapers; native x64/arm64, assembled pack, no-toolchain same-artifact arm64 load, unit/live, and all five `ci-ok` values succeeded. [Mutation reproof 34053012107](https://github.com/Cotal-AI/Cotal/actions/runs/34053012107) selected 52/360 over 78 paths and summed to 239 KILLED with zero pre-red, inconclusive, survived, unmeasured, or adverse. Source inspection confirmed real `SO_PEERCRED`, bounded pre-auth frames, per-seat custody, real attach/replay paths, hashed suite fragments, and the reachable unchanged `.internal` gitlink. Limitation: no suites were rerun for this review; conclusions use exact-head hosted logs and source inspection as required.

davidfarah2003:
**BLOCK** (retracts my prior APPROVE) exact head `1c12809e83ed259a24cb55c575520158869a8bfb`. Immediately before this comment: OPEN / MERGEABLE / CLEAN, `headRefOid` still that full SHA.

I re-read the hosted Windows logs myself. Run [34053012094](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094) job conclusions are `success` because shard steps use `continue-on-error: true`, and `windows-ok` only printed `required=success`. The shard commands themselves failed and stopped early:

- [0/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909137): `smoke:supervise-registered-user` (3 passed, 3 failed). **NEVER RAN — 109 of 138**.
- [1/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909134): `smoke:schema-profile` (`ContractInvalidError`: Maximum call stack size exceeded). **NEVER RAN — 124 of 127**.
- [2/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909080): `smoke:attach-repaint` (5s alt-screen timeout). **NEVER RAN — 114 of 135**.
- [3/4](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094/job/101539909098): `smoke:goal-epoch-fence` (EBUSY unlink of a JetStream blk). **NEVER RAN — 114 of 130**.

That is not terminal-green Windows shard evidence. Wrapper success is not a completed shard.

Secondary: this span changes published `@cotal-ai/cli` (`implementations/cli/src/lib/control.ts`) and published `@cotal-ai/connector-core` (`extensions/connector-core/src/docs-bundle.generated.ts`). The five new changesets vs `main` name only `cotal-ai`, `@cotal-ai/manager`, and `@cotal-ai/seat`.

Linux CI [34053012108](https://github.com/Cotal-AI/Cotal/actions/runs/34053012108) remains `success` at this SHA. Seat: `rev-1312-final-grok`. I do not claim `_2`.

davidfarah2003:
## Base control for the Windows shard finding

Both exact-head verdicts ([5562326197](https://github.com/Cotal-AI/Cotal/pull/1312#issuecomment-5562326197), [5562341637](https://github.com/Cotal-AI/Cotal/pull/1312#issuecomment-5562341637)) are correct that run [34053012094](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094) reports four green wrappers over four shard commands that exited 1 early. Before that is treated as this PR's defect, here is the same measurement taken on the base and on an unrelated head.

**Base `main` `52901c6298f5b6893baf2dcfb4d0716799cad202`, Windows run [34020309813](https://github.com/Cotal-AI/Cotal/actions/runs/34020309813)** — four `success` wrappers, four early stops:

| shard | stopped at | never ran |
|---|---|---|
| 0/4 | `smoke:supervise-registered-user` | 107 of 136 |
| 1/4 | `smoke:schema-profile` | 123 of 126 |
| 2/4 | `smoke:attach-repaint` | 111 of 132 |
| 3/4 | `smoke:mesh-attach-plane` | 109 of 130 |

**This PR's head `1c12809e83ed259a24cb55c575520158869a8bfb`, attempt 1 of run 34053012094:**

| shard | stopped at | never ran |
|---|---|---|
| 0/4 | `smoke:supervise-registered-user` | 109 of 138 |
| 1/4 | `smoke:schema-profile` | 124 of 127 |
| 2/4 | `smoke:attach-repaint` | 114 of 135 |
| 3/4 | `smoke:goal-epoch-fence` | 114 of 130 |

Shards 0, 1 and 2 fail at the same suite on the base as on this head, so that part of the shortfall is the standing Windows lane condition already filed as #914 (wrapper `success` over an early stop) and #698 (shard stops at first red). It is not this PR's to repair and repairing it here would be a workflow change outside this PR's scope.

**Shard 3 is the one real difference and it is not yet discriminated.** On the base run `smoke:goal-epoch-fence` executed and the shard continued to `smoke:mesh-attach-plane`; on this head `goal-epoch-fence` itself exits 1 with `EBUSY: resource busy or locked, unlink '...\jetstream\$G\streams\EPC_epfence\msgs\1.blk'`. The same suite also passes on the unrelated head `3606a39f76b9a122df8f2a3aec1e890ac1a73ba4` (Windows run [34062053915](https://github.com/Cotal-AI/Cotal/actions/runs/34062053915), whose shard 3 reaches `mesh-attach-plane` like the base). EBUSY on a JetStream block unlink is the Windows teardown signature named in #914, and Windows takes `LegacyPtyRuntime` on this head rather than the new custodial path, so the link to this change is weak — but weak is not measured. Attempt 2 of run 34053012094 is running now at this exact head to separate a flake from a reproducible candidate failure.

Until that attempt reports, the shard-3 question is open and the merge stays held. The changeset finding both reviewers raised is separate, real, and being corrected: this span changes published `@cotal-ai/cli` (`implementations/cli/src/lib/control.ts`) and the published `@cotal-ai/connector-core` docs bundle, and neither package appears in the five changesets.

davidfarah2003:
## Attempt 2 closes the shard-3 question: it was a flake

I left one thing open in [the control above](https://github.com/Cotal-AI/Cotal/pull/1312#issuecomment-5562471740) — shard 3 was the single place this head differed from the base, stopping early at `smoke:goal-epoch-fence` with an `EBUSY` unlink of a JetStream block file. Attempt 2 of run [34053012094](https://github.com/Cotal-AI/Cotal/actions/runs/34053012094), dispatched at this exact head `1c12809e83ed259a24cb55c575520158869a8bfb`, did not reproduce it. From the raw attempt-2 job logs:

| shard | stopped at | never ran |
|---|---|---|
| 0/4 | `smoke:supervise-registered-user` | 109 of 138 |
| 1/4 | `smoke:schema-profile` | 124 of 127 |
| 2/4 | `smoke:attach-repaint` | 114 of 135 |
| 3/4 | `smoke:mesh-attach-plane` (`goal-epoch-fence` passed) | 109 of 130 |

That is the base shape on all four shards. So no candidate-specific Windows failure is established, and the entire Windows finding reduces to the inherited lane condition tracked by #914 and #698. Nothing in this PR should be changed for it, and no workflow edit belongs here.

## The changeset finding, measured rather than assumed

It is real, and it is narrower than it reads. `.changeset/config.json` puts all 21 publishable packages in a single `fixed` group, so a package omitted from a changeset still versions and publishes in lockstep with the rest. What the omission actually costs is CHANGELOG attribution: the behaviour change in `implementations/cli/src/lib/control.ts` and the regenerated `@cotal-ai/connector-core` bundle would ship with no line describing them. That is worth fixing on the repo's own convention, and it is not a release blocker.

One correction worth recording, since it was raised and is wrong: `.changeset/explicit-refresh-identity.md` and `.changeset/user-endpoint-cli.md` do name `@cotal-ai/cli` in this tree, but neither is added by this PR — the PR adds exactly the five `seat-*.md` files. They cover other work, so they do not discharge this.

## What this PR still owes

`main` has moved twice since the reviewed head (#1353, then #1328), and this PR is now CONFLICTING on `extensions/connector-core/src/docs-bundle.generated.ts` because #1353 regenerated that bundle. The successor carries one push: fold current `main`, regenerate the bundle from its generator rather than resolving it by hand, and add `@cotal-ai/cli` and `@cotal-ai/connector-core` to this PR's own changesets. Both exact-head verdicts bind `1c12809e` only and do not transfer to that successor.

davidfarah2003:
**APPROVE** exact head `cf767e538abcf5b5e821016ddcf619472d4695ee` (re-read immediately before this comment: OPEN, MERGEABLE, UNSTABLE, still that full SHA; parents `eff3a07876473bc0509c85ad72cd7876e703a9ae` + `922cec98efdd5711646f3de017b6bf12c4b9fe4c`; tree `bdb9f2b88dd271c980777442d3733b7dc45cfd63`). Independent review; no edit or merge performed.

This retracts BLOCK comment 5562341637, which binds only `1c12809e83ed259a24cb55c575520158869a8bfb`. At this successor the changeset gap is closed: `.changeset/seat-teardown-and-ci-ok.md` lists `@cotal-ai/cli` and `@cotal-ai/connector-core` (plus seat, manager, `cotal-ai`). First-party span vs `origin/main` is seat, manager, cli, connector-core, and the published binary. `implementations/cli/src/lib/control.ts` drain is covered by the cli listing. The connector docs bundle contains current `docs/architecture.md` seat/lang text. Docs run 34063777686 SUCCESS on this SHA.

Windows workflow is unchanged vs `origin/main`. Windows run 34063777643 / required job 101568868738 is a genuine SUCCESS on this SHA (typecheck, build, unit, seam smokes, `pnpm smoke:windows`, secret-fs; log has no `##[error]` and ends `all checks passed`). `windows-ok` SUCCESS. CI run 34063777657 SUCCESS, including all four smoke shards, live, seat native/pack/load, and `ci-ok (live outcome classified)`.

GitHub remains UNSTABLE solely because mutation `changed` 34063777652 / job 101568868547 is still in_progress. That is not a candidate-specific Windows failure. Do not merge until that required check concludes. No product patch from this reviewer.

davidfarah2003:
**APPROVE** (terminal, exact head) `cf767e538abcf5b5e821016ddcf619472d4695ee`. Re-read immediately before this comment: OPEN, MERGEABLE, CLEAN, still that full SHA; parents `eff3a07876473bc0509c85ad72cd7876e703a9ae` + `922cec98efdd5711646f3de017b6bf12c4b9fe4c`; tree `bdb9f2b88dd271c980777442d3733b7dc45cfd63`. Independent review; no edit or merge performed.

This supersedes the prior conditional status on https://github.com/Cotal-AI/Cotal/pull/1312#issuecomment-5562925541 (APPROVE while mutation was still in_progress / PR UNSTABLE). Verdict remains APPROVE. Old BLOCK https://github.com/Cotal-AI/Cotal/pull/1312#issuecomment-5562341637 binds `1c12809e` only.

Successor workflows all bind this SHA and are terminal success:
- Docs `34063777686` SUCCESS
- Windows `34063777643` SUCCESS (Windows required `101568868738`; windows-ok SUCCESS; Windows log has no `##[error]`)
- CI `34063777657` SUCCESS including smoke shards, live, ci-ok
- Mutation reproof `34063777652` SUCCESS (`changed` completed 2026-09-07T00:08:29Z; `full` skipped by design)
- Code Quality `34063776299` SUCCESS

`origin/main` is now `24a85a32` (#1340). GitHub still reports CLEAN MERGEABLE at this head.

davidfarah2003:
**APPROVE** exact head `cf767e538abcf5b5e821016ddcf619472d4695ee`, tree `bdb9f2b88dd271c980777442d3733b7dc45cfd63`, parents `eff3a07876473bc0509c85ad72cd7876e703a9ae` + `922cec98efdd5711646f3de017b6bf12c4b9fe4c`. Independent replacement review used a detached worktree with empty porcelain and one reflog entry at this SHA (`2026-09-07 02:14:25 +0200`). Live PR is OPEN/MERGEABLE/CLEAN and public head-equal. Current main is `24a85a324142bb71aa88913db1e6e5b517399f93`.

**EXERCISED:** inspected the delta from reviewed `1c12809e83ed259a24cb55c575520158869a8bfb` plus the fold; the changeset directly lists `@cotal-ai/cli` and `@cotal-ai/connector-core`. Fresh docs-bundle generation was byte-identical (SHA-256 `485c72609af9d2e74d99b3a0e23c2d4aafe7a58dcce2c39fcb351e50799e2639`). `git merge-tree` against current main produced clean tree `9bc958c74c16125d9a8ef33f76064faec5294f8c`; its separately materialized ambient census passed across 107 files.

**INSPECTED:** exact-head CI `34063777657`, Windows `34063777643`, Docs `34063777686`, Code Quality `34063776299`, and mutation `34063777652` are terminal success. Mutation selected 52/361 fixtures and summed 239 KILLED, with 0 pre-red, 0 inconclusive, and no adverse verdict class. Windows attempt 2 reproduces the main-control command stops, so no candidate-specific defect is established; those inherited partial commands are not terminal-green evidence, while exact-head Windows required job `101568868738` is genuinely green. The other independent approval at comment `5563259535` was inspected but not relied upon.

**Residuals:** no heavy local gate was rerun. Head evidence and merge-artifact evidence remain separate: only the 107-file census was exercised on `9bc958c74c16125d9a8ef33f76064faec5294f8c`. No merge performed.

davidfarah2003:
Merge-artifact mutation evidence is now terminal for `4c784b6ea94147b2203cd37f462038f481493db2`.

Run: https://github.com/Cotal-AI/Cotal/actions/runs/34070143875

- `changed`: success
- corpus: 361 fixtures
- merge diff: 78 records / 78 changed paths
- selected: 52 fixtures
- result: 52 discriminated, 0 pre-red, 0 inconclusive
- 239 ANSI-stripped terminal `KILLED` verdicts, with zero `SURVIVED`, `PRE-RED`, `INCONCLUSIVE`, `UNGRADABLE`, `WRONG-RED`, or `ERROR` verdicts
- 52 per-fixture completion summaries, summing to the same 239 mutations
- elapsed: 8141 seconds against the 9000-second changed-set budget
- `full` skipped by changed-set design

This binds only the #1312 merge commit. It is not substitute evidence for a later head or merge artifact.

PR #1313, opened 2026-09-05
fix(manager): expose stranded static slot state

davidfarah2003:
## Summary

- Keep successful `inspect` replies on the existing live managed-agent row and its closed output schema.
- On a static-manager live miss, read the durable slot first and the lifecycle head second. Return a structured `failed-precondition` for nonterminal slots, with distinct slot phase and head state/op fields, both store revisions, explicit read order, and `ordered-not-atomic` consistency.
- Keep a genuinely unknown name and a retired slot as `not-found`. Surface bounded durable-read failures as `unavailable` instead of claiming absence.
- Include the diagnostic facts in the error message for string-only operator paths, and document the behavior.

## Design and compatibility

Extending `inspect` fixes the per-name ambiguity directly. A separate listing verb would help fleet-wide audits, but would leave the reported identical per-name refusals unchanged. Fleet-wide inventory remains separate work.

The manager cluster revision is unchanged. The described command contract constrains successful `data`, and `inspect` still returns the same closed live-agent schema on success. Endpoint errors have a protocol-level open `details[]` channel keyed by reverse-DNS `kind`; adding `ai.cotal.manager.static-slot-observation` changes a formerly misleading error result without changing the described success surface or its digest. An older consumer that understands only the code and message still receives the full diagnostic summary in the message.

The slot and lifecycle head are separate records and are not read atomically. The response carries `slotRevision`, `headRevision`, `readOrder: ["slot", "head"]`, and `consistency: "ordered-not-atomic"` so the pair is not presented as a snapshot. The issuance gate is not read because the retirement operation needed here is already on the lifecycle head, and a third record would add another torn edge without changing the per-name result.

This is a minor changeset because packages are 0.x and the externally observable refusal behavior changes. The fixed release group consequently versions all packages together.

## Outage fixture semantics

The outage smoke replaces `recordsKv.get` only for the exact slot key `mgrslot.local.store-outage`. That promise never settles, which simulates an unavailable read of that durable record and drives the production two-second timeout. The preceding live-hit call targets `w1`; it must return without requesting that key, and the fixture records whether the unavailable read was attempted. All other KV keys delegate to the real store so concurrent goal terminals and lifecycle cleanup continue normally. The original fixture hung every KV read, which accidentally blocked unrelated goal writes and wedged the later preservation drain; the repaired fixture isolates the claimed unavailable record instead.

The documentation field coverage was checked by tracing each projected public identifier against the control-surface page. The page now names every observation field, including `owner`, `actor`, `slotRevision`, `headLifecycleUid`, and `headRevision`, plus the read-failure detail's `name`, `record`, and `operation`. The generated connector-core bundle was regenerated from it.

## Independent review follow-up

The exact-head review found two test gaps. The shared-KV outage fixture marked every concurrent `get` as an inspect read, so unrelated lifecycle work could produce a false locality failure and wedge the later preservation drain. The fixture now hangs only `mgrslot.local.store-outage` and delegates every other key to the real store. The original fixture reproduced the 30-second preservation timeout locally; the scoped fixture passed the full manager-service smoke.

The review also identified unasserted durable phases. The real `ep.one` smoke now plants and inspects a provisioning slot with no head, an active slot missing from the live manager map, and a retired slot. Mutation 3 now names the earliest assertion that actually goes red when `readOrder` is removed. A fourth mutation injects a live-agent slot read and is killed by the live-hit locality cell.

## Verification

- `COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1 pnpm smoke:manager-service-ops` (22 passed)
- `pnpm smoke:manager-service-ops` (81 passed)
- `pnpm mutation-proof --config bin/smoke/mutations/static-slot-inspect.json --timeout 600000` (4/4 killed on named cells)
- `pnpm smoke:required-arg-seam`
- `pnpm check:docs-voice`
- `pnpm check:docsbundle`
- `pnpm smoke:docs` (5 sections)
- rebuilt `@cotal-ai/connector-core` and verified the built docs bundle contains the static-slot observation and read-failure guidance
- `pnpm changeset status`
- `pnpm typecheck` (`Scope: 27 of 28`; the JS-only `@cotal-ai/example-04-frontier-faces` package is the skipped 28th)
- full `pnpm build && pnpm -r typecheck` at final head (`Scope: 27 of 28`)
- manager contract reconfirmed at revision 10 with 23 commands; `inspect` output digest remains `sha256:72e00f8df7dabb1b1280aacda428c5d5448a172f58a65439c24d1c8cd4e177fb`

Refs #1274

#1274 stays open for the fleet-wide inventory half.


davidfarah2003:
**BLOCK** at `0df7f25d3b52c663a7066ec136ab50ef5e0e582a`.

Independent exact-head review. Installed with `pnpm install --frozen-lockfile` (this worktree had no `node_modules`). Did not run `mutation-proof`, `cotal up`/`down`, any live suite, or the dev-tree CLI.

## 1. The no-revision-bump claim — justified by digest arithmetic

`implementations/manager/src/manager-service-contract.ts` is **byte-identical** to `origin/main` (`sha256:4bf9920fddef698411f4e2bc35d9940d706d08e251d7caac048e1924eccb9c60` on both). Digest machinery in core (`canonical.ts` / `schema-profile.ts` / envelope+error parse) is also unchanged vs `origin/main`.

Computed at this head via `MANAGER_CONTRACTS` / `managerClusterDocument()` after building `@cotal-ai/core`:

- cluster `revision`: **10**
- `status` output: `sha256:171c1ea6bd518efc106491a69624d6dd19bf3f1ca221a9293eb8cec4abaedbbd`
- `inspect` output: `sha256:72e00f8df7dabb1b1280aacda428c5d5448a172f58a65439c24d1c8cd4e177fb`
- cluster root: `sha256:565e6eb5295802b69af0980de5feeb73ce311f5756cd162acfcf487f7dfeb607`
- cluster closure: `sha256:ea2ed75dc31d8587bf7dbe037e0c2e124b45f4be2d215cb771dcb80e5cb78f5d`

Those numbers cannot differ from `origin/main` because the contract source that produces them is the same file. This is not the sibling-PR family (changed status digest, revision left at 10).

**Mixed-version observability still exists, outside the success digest.** A live miss that used to be `not-found` is now `failed-precondition` for a nonterminal slot, with two new reverse-DNS detail kinds and a longer message. That is observable to any caller that keys on `error.code` (the shipped CLI alias-resolve path in `implementations/cli/src/lib/control.ts` does: inspect failure becomes `could not resolve` plus the code). That is the intended diagnosis, not a silent success-schema change.

**Where unknown-key tolerance is implemented.** SPEC section 13.3 requires `details[]` entries to carry reverse-DNS `kind` and does not close the rest. `EpErrorDetail` is `{ kind: string; [key: string]: unknown }`. `pickError` in `packages/core/src/endpoint-envelope.ts` validates `kind` then returns the whole object (`return o as EpErrorDetail`), so extra keys are preserved, not stripped. The CLI forwards `details` on control errors and still renders `message` for string-only paths. I did not find a consumer that rejects unknown detail keys.

**Capability / surface gating.** `managerShippedSurface()` is still `{ revision: 10, commandCount: 23 }`. `MANAGER_READ_COMMANDS` / `SPAWN_SERVICE_COMMANDS` still list `inspect` unchanged. Nothing else derives compatibility from this surface except that shipped pin (command names + revision + compiled digests). The non-bump is correct for the thing a revision is for.

Minor changeset is justified: 0.x, observable refusal behavior changes, `fixed` group lockstep.

## 2. The new surface

**Stored, then projected.** `observeStaticSlot` point-reads the durable slot, then `headCandidate`. `slotPhase` is `slot.row.phase` (`provisioning | active | terminalizing | retired`). It is not an in-memory manager-map invention. The live-map miss is the gate; the observation is the durable pair. Consistency is declared: `readOrder: ["slot","head"]`, `consistency: "ordered-not-atomic"`. Instance-local: a sibling `ownerInstanceId` is treated as absent on this manager.

**Boot / stop.** Inspect is `serveGated`. On a static miss it needs `this.goalWriter?.ctx.kv`; if that is missing it returns `unavailable` plus `ai.cotal.manager.static-slot-read-failed`, not `not-found`. After `stopGoalWriter()` the same miss path is `unavailable`. Live hits never touch the store. That is an honest code split, not a required-field-sometimes-absent success-schema split (success remains the closed agent row).

**Reachable vs asserted.** The type can report every `StaticSlotPhase`. The smoke plants only `terminalizing` plus `cleanupComplete: true` plus head `retiring` (and then a torn `retired` head). A genuinely unknown name is asserted. Store timeout on a miss is asserted. **Not asserted through inspect:** `provisioning`, live-map-miss `active` (the orphan that `planStaticSlotResume` terminalizes), planted `retired` (docs claim it stays `not-found`; the smoke only proves a never-written name), and head-absent. An active durable slot with no live row is the most important stranded shape for #1274 and is an untested promise on this door.

`manager.ts` merge fragility: the change is a 26-line insertion inside `inspect` in `managerServiceDefs()` plus four imports on the existing `static-lifecycle.js` import list. Any concurrent edit to that inspect miss (`if (!row) throw not-found`) will conflict and must be re-derived. The durable-read logic itself lives in `static-lifecycle.ts`, which is the less contended file.

## 3. Mutation fixture

`pnpm smoke:mutation-fixtures`: 343 files, 1759 anchors, including `bin/smoke/mutations/static-slot-inspect.json` **3/3 present and unique**, no comment-span. Did not run `mutation-proof`.

All three `find` strings are code-only and hit once. None is a pure empty replace.

**Mutation 3 names a downstream cell.** `expectRed` is the torn-pair cell, but the earlier stranded cell already requires `readOrder` to be `["slot","head"]`. Emptying `readOrder` reddens that stranded cell first. Auditors only check that the named cell exists, so a downstream name is a wrong-red this fixture will not catch. Mutation 2 sharing the stranded cell is fine: that is the earliest place `cleanupComplete` is asserted.

## 4. Suites

Ran (one at a time):

- `pnpm typecheck` at this head: **Scope: 27 of 28**, exit 0 (same scope print I actually saw; not copied from the brief).
- `COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1 pnpm smoke:manager-service-ops` after `pnpm --filter @cotal-ai/manager build`: **19 passed, 0 failed**.
- **Unnamed by the author:** `pnpm smoke:static-lifecycle` — chosen because it is the suite that actually drives slot phases, `cleanupComplete`, and resume/terminal, without going through `inspect`. **43/43**.

Did not rerun required-arg, docs-voice, changeset status, or docs bundle.

**Adversarial / CI, not just the author's green list.** GitHub Actions `CI` run `33993923172` at this exact sha: `smoke (shard 1/4)` **failed** on the new cell

`inspect keeps live hits entirely local when the durable slot store is unavailable`

The reply was `ok: true` for live `w1`; the cell still failed, which means `durableReadAttempted` was true. The stub is a hanging `kv.get` on the **shared** goal-writer KV that sets the flag on any call. `list()` / `psOwnerFilter()` do not read that KV, so a live inspect that succeeds while `kv.get` still ran is a concurrent reader (lease/goal/registry), not inspect consulting durable state. The locality claim is not what this stub measures. Locally the cell passed; CI already showed it is not deterministic. The suite does not fail-fast, so the rest of the new cells still printed green on that CI run.

## 5. CI at `0df7f25d3b52c663a7066ec136ab50ef5e0e582a`

Presence (workflow-runs API plus `/home/cotal/fm-artifacts/greencheck.sh Cotal-AI/Cotal 1313`), **separate from conclusions**:

| Workflow | Present by name | Conclusion |
| --- | --- | --- |
| CI | yes (`33993923172`) | **not green**: shard 1/4 **failure**; shards 0/4 and 2/4 **cancelled**; `ci-ok` queued. Treat cancelled/queued as UNMEASURED. |
| Docs | yes (`33993923126`) | completed **success** |
| Mutation reproof | yes (`33993923136`) | **cancelled** — UNMEASURED, not green |
| Windows | yes (`33993923179`) | completed **success** |

Also present: Code Quality: PR #1313 success. Greencheck: `NOT GREEN`.

## Could not check

- Whether mutation-proof would still kill mutation 3 on the earliest cell (forbidden to run; graded by reading).
- Mutation reproof job output (cancelled).
- Inspect of `provisioning` / live-miss `active` / planted `retired` (no test to run).
- Exact skipped 28th typecheck package beyond the `Scope: 27 of 28` print (author names Frontier Faces; I did not independently list the skipped package).

Worktree is clean at `0df7f25d3b52c663a7066ec136ab50ef5e0e582a`.

davidfarah2003:
**BLOCK** at `01df2de6e837f294f0d4733bb3e419f3e2122bff` (confirmed via `gh pr view 1313 --json headRefOid` at publish time).

Independent exact-head review after the re-point from `0df7f25d3`. Product blobs named by the orchestrator are identical across the move (`static-lifecycle.ts`, `manager.ts`, `docs/control-surface.md`, `bin/smoke/mutations/static-slot-inspect.json`). `manager-service-contract.ts` is also identical (`git rev-parse` blob `c0ce25ca51ff6f9d0a93fc704b99107e9b4e8657` on both this head and `origin/main`). Did not run `mutation-proof`, `cotal up`/`down`, any live suite, or the dev-tree CLI. Installed earlier in this worktree with `pnpm install --frozen-lockfile`.

## 1. Full command table, derived, not sampled

`managerClusterDocument()` / `managerShippedSurface()` at this head after `pnpm --filter @cotal-ai/core build` (needed so `@cotal-ai/core` package exports resolve):

- `urn`: `ai.cotal.manager`
- `revision`: **10**
- `commandCount`: **23**

| name | capability | targeted | modes | inputDigest | outputDigest |
| --- | --- | --- | --- | --- | --- |
| status | manager.read | false | - | `sha256:1db0472aaee8e6b9d8d3d430ee021a1b6efc559ab7c0b99186898c972c53dd2d` | `sha256:171c1ea6bd518efc106491a69624d6dd19bf3f1ca221a9293eb8cec4abaedbbd` |
| ps | manager.read | false | - | `sha256:1db0472aaee8e6b9d8d3d430ee021a1b6efc559ab7c0b99186898c972c53dd2d` | `sha256:f8c50de603c42d13d3c23a7c8b7351b857a09b2a0a26e5c5a880c2daf7777c61` |
| inspect | manager.read | false | - | `sha256:c74bfe6c0426cf31f56796a48992639eb08a46c36f7dd7205b497cf2c83a8382` | `sha256:72e00f8df7dabb1b1280aacda428c5d5448a172f58a65439c24d1c8cd4e177fb` |
| models | manager.read | false | - | `sha256:edcf1dcdeede38991ec59b6bf2a44fc2a36bb6252fce5d80e399717ec85df88c` | `sha256:66c9c24fef33a9cf69afdb59324a37591e600ce8106ed61e71b46dddcbc124f6` |
| spawn | manager.spawn | false | - | `sha256:82e0d1aa25f71a122b31bf86bbca5562394038cf66460985ebdba8742b578a9b` | `sha256:d883a9f97f7c034772c7cbc28b22d63cf69870ba072d470c2f7b104f455e2e83` |
| despawn | manager.lifecycle | true | owner\|any | `sha256:2fde1dce4a65d3dad7991113b4b98f51f10b94109898f049154759bc456567ef` | `sha256:7ec7cf7c8d9011fc57022db68577851eb1fb57cdbef94766e4d6ca26bc70131e` |
| attach | manager.lifecycle | true | owner\|any | `sha256:1db0472aaee8e6b9d8d3d430ee021a1b6efc559ab7c0b99186898c972c53dd2d` | `sha256:0d2cb1346fa7e1abd97750dc75e1720e386983641cae9803aa6c0602c77d9a6d` |
| input | manager.lifecycle | true | owner\|any | `sha256:31b92d1926fc84652ca043f8b9a2c443ed272932e99935b64ab35e24d88c0052` | `sha256:165c6b378b4731d23598bf42d390998da55afeac3690cfdad700820b0ad5b126` |
| turn | manager.lifecycle | true | owner\|any | `sha256:ac9ed1455cb1dbf9272f3d0e7024d8387418c501e407cb9f1ec8c4f56e290cbc` | `sha256:b9d570154853a000e00e612c8051b21aaedaf96586b4ccca76bf4572b24b050d` |
| turn-pending | manager.self | true | self | `sha256:1db0472aaee8e6b9d8d3d430ee021a1b6efc559ab7c0b99186898c972c53dd2d` | `sha256:9e916615754c822e144ec614bd1534226f8fbc90a55ac3e66e088ed7a56dc848` |
| turn-yield | manager.self | true | self | `sha256:dc992f7975be5cec5c3f90153288ad6821380867fe004056fbcf8eb227656542` | `sha256:63909e8274cd7febd74f46b4932af8e1a27d6d5b7c16dd5d8dbd2955142bd013` |
| stop | manager.self | true | self | `sha256:2fde1dce4a65d3dad7991113b4b98f51f10b94109898f049154759bc456567ef` | `sha256:7ec7cf7c8d9011fc57022db68577851eb1fb57cdbef94766e4d6ca26bc70131e` |
| define-persona | manager.persona | false | - | `sha256:4d3f0cd9577e9a9daa834f1091eddc77579186c3b56ad49024ecec3e3d2d5834` | `sha256:a121baba27357c3710b20023422927b0c11d4d1d58a542f909efd7ad1152751c` |
| list-personas | manager.read | false | - | `sha256:1db0472aaee8e6b9d8d3d430ee021a1b6efc559ab7c0b99186898c972c53dd2d` | `sha256:bb37634698c67b4468dd1c85dfbcfa0c13125f91aed9fd0769ee813438048c49` |
| show-persona | manager.read | false | - | `sha256:c74bfe6c0426cf31f56796a48992639eb08a46c36f7dd7205b497cf2c83a8382` | `sha256:45118d9dfc9803469c095dc2d34d8f6bec6b48eec90d8017f4617acc9d6fb7d2` |
| purge | manager.admin | false | - | `sha256:fc7eb2289bcdc583558efe1127c5f8ab988649ecbe0a0e950dd175c512b4cd86` | `sha256:dc91365525849af1bd5dfacedd4b5b6fbdc611db99c2478a803ee8995191ee4e` |
| launch | manager.admin | false | - | `sha256:dc33bfbeba9027bba4e06e0d96ea5ea0bdb2f412899707288ba0e534d88dc565` | `sha256:d883a9f97f7c034772c7cbc28b22d63cf69870ba072d470c2f7b104f455e2e83` |
| resume-preserved | manager.admin | false | - | `sha256:204aa0baa9794e9c0183cf62666928f0a8a04811aa92b4ce73e5c12a0574b5b9` | `sha256:241df3a8f87af62ea53beb2267ab5309af830a357180952924a0fca26e748efb` |
| commit-resume | manager.admin | false | - | `sha256:b863e0912fbf9344b346a1f69179a1039254fad2a5dec27cb691fd7812617976` | `sha256:b89dc03ca8afa323c4e36ad54e8af032fa084aba04e0c3f6b27ef5df4788f46c` |
| finalize-resume | manager.admin | false | - | `sha256:7df6373748333f8c0907dc38eaca8596714aad3a05cf11974bef97cd328f238c` | `sha256:959dbd34809db2f410c4f983b1cb0bf7a7fbf24d275ceb5fd097979e46e1fee7` |
| prepare-preservation | manager.admin | false | - | `sha256:b863e0912fbf9344b346a1f69179a1039254fad2a5dec27cb691fd7812617976` | `sha256:241df3a8f87af62ea53beb2267ab5309af830a357180952924a0fca26e748efb` |
| commit-preservation | manager.admin | false | - | `sha256:b863e0912fbf9344b346a1f69179a1039254fad2a5dec27cb691fd7812617976` | `sha256:241df3a8f87af62ea53beb2267ab5309af830a357180952924a0fca26e748efb` |
| abort-preservation | manager.admin | false | - | `sha256:b863e0912fbf9344b346a1f69179a1039254fad2a5dec27cb691fd7812617976` | `sha256:959dbd34809db2f410c4f983b1cb0bf7a7fbf24d275ceb5fd097979e46e1fee7` |

Compile inputs vs `origin/main` are the same blobs: `manager-service-contract.ts`, `packages/core/src/canonical.ts`, `schema-profile.ts`, `endpoint-envelope.ts`, `endpoint-error.ts`. `git diff origin/main HEAD -- packages/core` is empty. The table cannot differ from main. The non-bump is justified: success schemas are unchanged; the new observation rides `error.details[]`, which SPEC 13.3 leaves open beyond required `kind`.

`pickError` still returns the whole object after validating `kind`, so extra detail keys survive. The CLI alias-resolve path keys on `error.code` and message. Mixed-version observability is the intended diagnosis (`not-found` → `failed-precondition` / `unavailable` on a miss), not a silent success-schema change.

## 2. The 9-line test fix (this is the block)

Delta vs `0df7f25d3` is only `implementations/manager/smoke/manager-service-ops.smoke.ts` (+8/-1). Product files are byte-identical.

The old stub hung **every** `kv.get`. That is not "the durable store is unavailable"; it is "the shared goal-writer KV never returns," which races concurrent readers (the previous CI red on this named cell with `ok: true` while the flag still flipped). Scoping the hang off the shared map is the right *kind* of repair.

What landed is narrower than the cell's name.

The live-hit cell inspects `"w1"` and asserts `!durableReadAttempted`. The flag is now set only when `key === staticSlotKey(DEV_OWNER, "store-outage")`. Inspect of a live name never enters `observeStaticSlot` (`manager.ts`: live `list()` hit returns the row; durable read is under `if (!row)`). Even a **product regression** that started point-reading `mgrslot.<owner>.w1` on a live hit would call `originalGet`, leave the flag false, and this cell would still pass.

So the repaired patch **stops hanging**. It does **not** simulate "the durable slot store is unavailable" for the live path the cell names. The cell can now go green without testing locality. That is worse than the previous red.

The *miss* cell on `"store-outage"` is the honest half: `observeStaticSlot` reads exactly that key, the 2s bound fires, and the assertion requires `unavailable` + `ai.cotal.manager.static-slot-read-failed` + `record === "slot"` + the timeout message. That cell would still fail if the miss path swallowed a store failure as `not-found`. Cell count under `COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1` is still 19. The added lines did not drop a named assertion; they silently emptied the live-hit one.

A locality cell that would catch a regression needs a probe on the live name's slot key (or on `observeStaticSlot` / any `mgrslot.*` get during `inspect("w1")`) while still forwarding unrelated keys so prepare-preservation does not wedge.

## 3. Docs (author-identified, plus a reverse pass)

Author-identified: `docs/control-surface.md` describes both store revisions but never names **`slotRevision`** and **`headRevision`**. I would have blocked on that if the author had not already filed it. Those are the fields the torn-pair cell compares. A consumer that cannot discover them from the control-surface page cannot use this door for the diagnosis #1313 exists to expose. Treat the unpushed fix as merge-required, not optional.

Reverse pass (doc → code): nothing the new section names is absent from the type. Live-hit locality, retired-as-`not-found`, `failed-precondition` + `ai.cotal.manager.static-slot-observation`, `readOrder` / `consistency`, issuance-gate exclusion, miss-`unavailable` + `ai.cotal.manager.static-slot-read-failed`, and user-auth remaining live-map-only all match `observeStaticSlot` / the inspect miss branch.

Code → doc, beyond the author pair. Public `StaticSlotObservationDetail` fields also unnamed as identifiers: `owner`, `actor`, `headLifecycleUid` (the page says "lifecycle uid" twice, once per record). The read-failed detail also carries `name`, `record`, and `operation`, which the page does not name. `kind` is documented. `slotPhase`, `slotLifecycleUid`, `cleanupComplete`, `headState`, `headOp`, `readOrder`, `consistency` are named.

## 4. Surface / mutation / suites (transferred by blob identity, not re-derived)

`observeStaticSlot` is stored-then-projected (slot, then head). Inspect is `serveGated`. Missing KV on a static miss is `unavailable`, not `not-found`. Live hits do not touch the store. Still unasserted through inspect: `provisioning`, live-map-miss `active` (the #1274 orphan), planted `retired` (docs claim `not-found`; the smoke only proves a never-written name), head-absent.

Mutation fixture `bin/smoke/mutations/static-slot-inspect.json` is unchanged: 3/3 unique code-only finds. Mutation 3 still names a downstream cell (`expectRed` is the torn-pair cell while the stranded cell already asserts `readOrder`). Did not run `mutation-proof`.

Ran at this head: `COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1 pnpm smoke:manager-service-ops` → **19 passed, 0 failed**. Did not rerun the full 81-cell suite or `smoke:static-lifecycle` (product files unchanged). Did not rerun typecheck.

## 5. CI at `01df2de6e837f294f0d4733bb3e419f3e2122bff`

Presence (`gh` workflow-runs API + `/home/cotal/fm-artifacts/greencheck.sh Cotal-AI/Cotal 1313`), separate from conclusions:

| Workflow | Present by name | Conclusion |
| --- | --- | --- |
| CI | yes (`33995717311`) | **queued** — UNMEASURED |
| Docs | yes (`33995717315`) | completed **success** |
| Mutation reproof | yes (`33995717323`) | **in_progress** — UNMEASURED |
| Windows | yes (`33995717313`) | completed **success** |

Also present: Code Quality: PR #1313 success. Greencheck: `NOT GREEN`.

Worktree is at `01df2de6e837f294f0d4733bb3e419f3e2122bff` with no extra files.

davidfarah2003:
**BLOCK** at `56a4c21dd29952a5321502461beec4501f0f784e` (confirmed via `gh pr view 1313 --json headRefOid` at publish time).

Delta review from `01df2de6e`. Product source is unchanged: `implementations/manager/src/manager.ts` and `static-lifecycle.ts` are byte-identical to the last review. `manager-service-contract.ts` remains blob-identical to `origin/main`. The full 23-row command table (revision 10, inspect output `sha256:72e00f8df7dabb1b1280aacda428c5d5448a172f58a65439c24d1c8cd4e177fb`) therefore still matches main. Did not rerun the digest dump. Did not run `mutation-proof`, `cotal up`/`down`, any live suite, or the dev-tree CLI. Installed earlier in this worktree with `pnpm install --frozen-lockfile`.

Changed files vs `01df2de6e`: docs + bundle, mutation fixture retarget, and 70 lines in `manager-service-ops.smoke.ts`.

## What this head closed

Docs now name **`slotRevision`** and **`headRevision`**. That was the author-identified gap I would have blocked on. The generated docs bundle matches. Remaining unnamed identifiers (`owner`, `actor`, `headLifecycleUid` as a token, plus read-failed `name`/`record`/`operation`) are secondary; they are not this block.

Phase coverage through the real inspect door is now present and green:

- `provisioning` with no head (`writeStaticSlotIntent` only): `failed-precondition`, no `headState`/`headRevision` keys.
- live-map-miss `active` (the #1274 orphan): `failed-precondition`, `slotPhase=active`, matching uids.
- planted `retired`: `not-found`, no observation detail.

Focused smoke at this head: `COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1 pnpm smoke:manager-service-ops` → **22 passed, 0 failed** (was 19).

Mutation 3 now names the earliest stranded cell (`inspect distinguishes a stranded durable slot…`) instead of the torn-pair cell. That was the downstream-name defect. Find string is still unique and code-only. Did not run `mutation-proof`.

## What still blocks: the live-hit locality cell is still empty

Unchanged from `01df2de6e`. The cell inspects `"w1"` and asserts `!durableReadAttempted`. The flag flips only on `staticSlotKey(DEV_OWNER, "store-outage")`. Inspect of a live name never enters `observeStaticSlot` (`manager.ts`: durable read is under `if (!row)`). A **product regression** that started point-reading `mgrslot.<owner>.w1` on a live hit would call `originalGet`, leave the flag false, and this cell would still pass.

The miss-side `"store-outage"` cell is still the honest half: hang that key, require `unavailable` + `ai.cotal.manager.static-slot-read-failed` + the 2s timeout. The live-hit cell's name still claims store unavailability for a path that never consults the hung key.

A locality cell that would catch a regression needs a probe on the live name's slot key (or any `mgrslot.*` get during `inspect("w1")`) while still forwarding unrelated keys.

## CI at `56a4c21dd29952a5321502461beec4501f0f784e`

Presence (workflow-runs API), separate from conclusions:

| Workflow | Present by name | Conclusion |
| --- | --- | --- |
| CI | yes (`33997074495`) | **pending** — UNMEASURED |
| Docs | yes (`33997074482`) | completed **success** |
| Mutation reproof | yes (`33997074481`) | **in_progress** — UNMEASURED |
| Windows | yes (`33997074488`) | **queued** — UNMEASURED |

Also present: Code Quality in_progress. Treat incomplete runs as UNMEASURED, not green.

Worktree is at `56a4c21dd29952a5321502461beec4501f0f784e` with no extra files.

davidfarah2003:
**APPROVE** at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e` (confirmed via `gh pr view 1313 --json headRefOid` at publish time).

Delta review from `56a4c21`. Product source is unchanged: `manager.ts`, `static-lifecycle.ts`, and `manager-service-contract.ts` are byte-identical to the last review and the contract file remains blob-identical to `origin/main`. The 23-row command table (revision 10, inspect output `sha256:72e00f8df7dabb1b1280aacda428c5d5448a172f58a65439c24d1c8cd4e177fb`) still matches main. Did not rerun the digest dump. Did not run `mutation-proof`, `cotal up`/`down`, any live suite, or the dev-tree CLI. Installed earlier in this worktree with `pnpm install --frozen-lockfile`.

## The previous block is closed

The live-hit cell now watches `staticSlotKey(DEV_OWNER, "w1")` and asserts `!liveSlotReadAttempted`. Inspect of a live name still never enters `observeStaticSlot` (`durable` read remains under `if (!row)`). A product regression that started point-reading `mgrslot.<owner>.w1` before returning the local row would flip that flag and redden this cell, while still returning `ok: true`. That is the locality claim.

The miss-side hang on `"store-outage"` is unchanged and still the honest unavailability cell: `unavailable` + `ai.cotal.manager.static-slot-read-failed` + the 2s timeout.

New mutation 4 plants exactly that live-slot read (`observeStaticSlot` after the live `list()` hit) and names this cell. The `find` string hits once in `manager.ts`. Did not run `mutation-proof`.

## Docs reverse pass

`docs/control-surface.md` now names `owner`, `actor`, `headLifecycleUid`, and the read-failed `name` / `record` (`slot` | `head` | `slot-or-head`) / `operation: "read"`. The generated bundle matches. Remaining public observation fields from the last reverse pass are named.

## Suites / CI

`COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1 pnpm smoke:manager-service-ops` at this head: **22 passed, 0 failed**.

CI at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e` (workflow-runs API), presence separate from conclusions:

| Workflow | Present by name | Conclusion |
| --- | --- | --- |
| CI | yes (`33997840757`) | **pending** — UNMEASURED |
| Docs | yes (`33997840790`) | **queued** — UNMEASURED |
| Mutation reproof | yes (`33997840758`) | **in_progress** — UNMEASURED |
| Windows | yes (`33997840760`) | **queued** — UNMEASURED |

Also present: Code Quality queued. Incomplete runs are UNMEASURED, not green. I am not blocking on that: the defects that were the previous BLOCKs are closed in product-adjacent tests and docs, and product source did not change.

Worktree is at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e` with no extra files.

davidfarah2003:
**APPROVE** at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e` (reconfirmed via `gh pr view 1313 --json headRefOid` at this comment). Tree `570bd53cde2991ab8bd4485f492e639760cfbc93` matches `git rev-parse HEAD^{tree}`.

This is the independent check of the locality mutation, not an inherited lane number. Product source vs `56a4c21dd` is empty (`manager.ts` / `static-lifecycle.ts` / `manager-service-contract.ts`). Contract blob `c0ce25ca51ff6f9d0a93fc704b99107e9b4e8657` is still identical to `origin/main`. Inspect output digest is therefore still `sha256:72e00f8df7dabb1b1280aacda428c5d5448a172f58a65439c24d1c8cd4e177fb`, revision 10, 23 commands. Did not rerun the dump. Did not run `mutation-proof`, `cotal up`/`down`, typecheck, or a second suite. Installed earlier with `pnpm install --frozen-lockfile`. Focused smoke already run at this SHA: **22 passed**.

## 1. The negative assertion is not vacuous

`liveSlotReadAttempted` is set only in the exact-key wrapper, and only when `kv.get(staticSlotKey(DEV_OWNER, "w1"))` is invoked. Shipped inspect never does that on a live hit: durable read is under `if (!row)`.

The mutation's `replace` injects, after that `list().find`, and only when a live row exists:

`if (row && this.auth && !this.userMode && this.goalWriter?.ctx.kv) await observeStaticSlot(..., name, ...)`

`observeStaticSlot` point-reads `staticSlotKey(owner, alias)`. For `inspect("w1")` that is exactly `mgrslot.local.w1`. The wrapper sets the flag **synchronously** at the start of `kv.get`, then forwards to `originalGet`. The live cell still requires `ok: true` and the same id, so the mutant preserves the successful reply and fails only `!liveSlotReadAttempted`. That is not scheduling-dependent.

**Earliest red, by cell order, not by the lane's 23/22 arithmetic.** I count **22** `check()` cells under `COTAL_SMOKE_STATIC_SLOT_INSPECT_ONLY=1` (the live-locality cell is number 21; the miss-outage cell is 22). I did not rerun `mutation-proof`, so I do not inherit "baseline 23 marks / mutant 22". Tracing the mutant through earlier cells:

- Cells 13–14 (`inspect returns the same row`, closed success contract) also call `inspect("w1")` and would invoke the injected read, **before** the wrapper is installed. `observeStaticSlot` on a real slot succeeds and the handler still returns the live row, so those cells stay green.
- Cells 15–20 are live-map misses. `if (row && ...)` does not fire. Miss-path `observeStaticSlot` is unchanged.
- Cell 21 is the first assertion of `!liveSlotReadAttempted`. That is the named `expectRed`.

The fixture `why[]` still says three mutations. There are four. Not a block.

## 2. Injection is JSON-only

`git grep` of `if (row && this.auth && !this.userMode && this.goalWriter?.ctx.kv) await observeStaticSlot` hits **only** `bin/smoke/mutations/static-slot-inspect.json`. Shipped `manager.ts` has `observeStaticSlot` at the import and at the miss-path `observed = await observeStaticSlot(...)` (line 2334). `DEV_OWNER` is already in scope. `git diff 56a4c21dd HEAD -- implementations/manager/src/manager.ts` is empty.

## 3. Unauthorized-window mutation-proof run

I did not rerun it. Tree identity at this HEAD is `570bd53cde2991ab8bd4485f492e639760cfbc93`, matching the lane's report and GitHub's commit tree. I accept the disposition that the evidence describes this tree. I do not accept the 23/22 mark counts as independently verified; the qualitative kill (named cell, earliest among assertions of that flag) follows from the source.

## CI at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e`

Five workflows present by name (greencheck + runs API). Conclusions are all incomplete:

| Workflow | Present | Conclusion |
| --- | --- | --- |
| CI | yes (`33997840757`) | pending — UNMEASURED |
| Docs | yes (`33997840790`) | queued — UNMEASURED |
| Mutation reproof | yes (`33997840758`) | in_progress — UNMEASURED |
| Windows | yes (`33997840760`) | queued — UNMEASURED |
| Code Quality: PR #1313 | yes (`33997839239`) | queued — UNMEASURED |

Greencheck: `NOT GREEN`. Not a product block: the locality defect is closed in the test and the mutation writer.

Worktree is at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e` with no extra files.

davidfarah2003:
**Design ruling on the `attach-reconnect` red, still at `e1094bf76be7c39130f0c49f72ce4863e7c0bb5e`.** Reading only. No suite, typecheck, or install. The earlier APPROVE is not withdrawn.

**(a).** The attach TTY should stay clean. The stranded-slot observation belongs on `inspect`. `attach-reconnect`'s assertion is defending a real user-facing contract, not a stale regex.

## How the enrichment reached attach (accident, not an attach formatter)

`renderStaticSlotObservation` has **one** call site: inspect's live-miss throw in `manager.ts`. Attach never calls it.

The leak is three layers of glue:

1. Inspect of a nonterminal durable slot is now `failed-precondition` with the observation in `error.details[]` **and** spliced into `error.message` (docs: "so string-only operator paths do not hide the structured detail").
2. Every targeted CLI verb (`attach`, `stop`, `input`, `despawn`) resolves the alias through `inspect` in `implementations/cli/src/lib/control.ts`, and on failure returns `could not resolve "<name>": ${inspect.error.message}` while **forwarding inspect's code**.
3. Attach classifies that code with `attachRefusal`: `not-found` → `gone` (print `seat <name> is gone`, exit 0); **everything else, including `failed-precondition`, is `transient`**. Transient manager refusals are relayed as `[cotal: ${message}]` and the loop retries.

After `cotal stop`, the slot is `terminalizing`. That is exactly the stranded miss #1313 made observable. Inspect therefore answers `failed-precondition` instead of `not-found`. Attach treats a despawned seat as "try again" and dumps the inspect formatter onto the reconnect TTY. The cell's own name is the contract that broke: *"not an infinite retry against something that is not there."*

The attach-reconnect mutation fixture states the old path as the proof: stop → manager `not-found` on the wire → inspect carries that code → clean exit. `docs/cli.md` says a reconnect that finds the seat gone exits with `seat <name> is gone`. `attachRefusal`'s own comment says reading `not-found` as transient is an infinite retry against a seat that no longer exists.

Cell F already pins `attachRefusal("failed-precondition") === "transient"` (session-ceiling and similar). That is why a blanket "treat failed-precondition as gone" would be wrong: it would turn a retryable attach refusal into `seat is gone`.

## Other callers

Same inspect-resolve string: `cotal stop` / `input` / `despawn --name`. `seat-input-live` matches `/no managed agent|could not resolve/`. Those are operator commands, not a mid-reconnect TTY. They can carry inspect's diagnosis. Attach's reconnect loop is the path whose **code** must stay `gone` for a missing seat.

If (b) were taken and only the assertion were widened, attach would still classify stranded inspect as transient and keep retrying. Shortening inspect's `error.message` without changing the code would make the retry **silent**. Worse than the verbose red.

## What to change

Keep inspect's `failed-precondition` + structured details. That is the #1313 surface.

Teach the **CLI resolve / attach loop** that an inspect miss with `ai.cotal.manager.static-slot-observation` (or the `could not resolve` inspect-failure shape) is `gone`, same as `not-found`. Then reconnectNotice never prints slot revisions, and the documented `seat <name> is gone` line is what a person sees.

Do not strip the formatter from inspect to paper over attach. Do not update the attach-reconnect assertion as the fix.

Two shard failures are consistent with this: the loop never reaches `gone`, so both "the attach exits" and "saying the seat is gone" go red.

davidfarah2003:
Validation note for `smoke:attach-reconnect` stability:

- One patched-tree run reached cell H and failed to establish the silent-seat attach with `unavailable: the contract-store read for dc33bfbeba9027bba4e06e0d96ea5ea0bdb2f412899707288ba0e534d88dc565 failed (a failed observation is never absence, SPEC 13.7): timeout`. The suite then hit its 10-minute command timeout.
- The failure did not recur across two subsequent complete runs: patched head `fe5d9fea9` passed 72/72, and unpatched `origin/main` at `259e618e2` passed 72/72.
- The base comparison admitted at 69.4% CPU over 20s with 7.56 GiB MemAvailable and ended at 85.2% CPU with 5.29 GiB MemAvailable. This was a different load band from the failing run, so the non-reproduction does not decide a load-sensitive cause.
- The repair adds only a `failed-precondition` plus `ai.cotal.manager.static-slot-observation` classification. Cell H's `unavailable` refusal follows the same transient branch before and after this change.

This is recorded as a single stability observation, not a reproduced defect or proof of its cause.

davidfarah2003:
## APPROVE — delta review at `fe5d9fea9ce3360bbe82a207ab28f6211e174edb`

This is a **DELTA review**. The prior APPROVE at `e1094bf76` is an ancestor of this head
(`git merge-base --is-ancestor` = true, verified), so the base carries that approval and this covers
only the delta `e1094bf76..HEAD`: three files, 28 lines
(`.changeset/static-slot-inspect.md`, `implementations/cli/src/commands/agents.ts`,
`implementations/cli/src/lib/control.ts`). Head confirmed on both `git rev-parse HEAD` and
`gh pr view 1313 --json headRefOid`.

### §1 — Classification reaches, and is narrow enough ✓
- **The detail has a real writer (verified independently, not taken on report).**
  `STATIC_SLOT_OBSERVATION_DETAIL` is exported at `static-lifecycle.ts:65`; `kind` is a **required**
  field on `StaticSlotObservationDetail` (`:67`); `observeStaticSlot` builds it at `:127-144`; and
  `manager.ts:2344` **throws** `EpEnvelopeError("failed-precondition", …, [observed])`, gated at
  `:2343` on `slotPhase !== "retired"`. A stopped seat reads a non-retired phase, so the detail
  fires. The gate keyed on the constant is not dead.
- **`details` survives the whole path on every branch that can carry this refusal.** `attach` is a
  *targeted* op, so its refusal arrives through the alias-resolve `inspect` read. That branch
  (`control.ts:249`) is exactly the one the PR adds; the *submit* branch (`control.ts:288`) already
  forwarded `details` at the base. Both refusal-returning branches now forward it.
  `ManagerReply = ControlReply & {…}` and `ControlReply.details` is declared at
  `packages/core/src/types.ts:403`, so the added `reply.details` typechecks.
- **Keyed on the namespaced detail, never the wording.** `attachRefusal` (`agents.ts:477`) requires
  `code === "failed-precondition"` **and** a detail whose `kind` equals the namespaced constant. The
  refreshed comment restates the code-not-copy rule, and `agents.ts:984` still prints the documented
  `seat <name> is gone` from the `gone` verdict.

### §2 — Anti-fix check (all three verified against the diff, not the report) ✓
1. **Regex not widened.** No test, spec, fixture, or smoke file is in the delta (`--name-only` shows
   only the three files). The `attach-reconnect` suite and `docs/cli.md:760` are untouched.
2. **`inspect`'s message not shortened.** `manager.ts:2344` still emits the full
   `renderStaticSlotObservation(observed)` string *and* `[observed]`. `manager.ts` is not in the
   delta; the fix adds classification rather than removing the symptom.
3. **Not all `failed-precondition` → gone.** The branch requires the detail. The session ceiling
   (bare `failed-precondition`, no detail) stays transient, and **cell F pins it** — re-run below,
   `attachRefusal("failed-precondition") === "transient"` passes.

**My own adversarial probe (RAN):** a line-for-line port of `attachRefusal`, fed the strongest
adversarial case — `failed-precondition` **without** the detail — plus empty-details, an unrelated
detail kind, and the detail under the wrong code (`unavailable`). All four stay `transient`; only
code+detail together yield `gone`. `unavailable` (the cell-H / delivery-timeout class) falls through
to `transient` unchanged, so the added branch is provably not on that path.

### §3 — Scope claims ✓
- **Manager contract byte-identical to `origin/main`.** `git diff origin/main HEAD --
  manager-service-contract.ts` is empty; `inspect` still maps `output: AGENT_ROW_SCHEMA`, revision
  unchanged. Contract vocabulary/inspect digest did not move.
- **`docs/cli.md:760` already documents this** (`seat <name> is gone`, and the session ceiling
  relayed while the loop keeps trying). No user-visible behaviour changed beyond what the doc
  states, so no doc edit was owed.
- **Changeset valid.** `pnpm changeset status` is clean. `@cotal-ai/cli: patch` is justified (a 0.x
  consuming bugfix; `attachRefusal` only gained an optional param). The `fixed` group bumps all 21 at
  minor in lockstep regardless, so the CLI entry buys a correct CHANGELOG line, not a wrong release.

### §4 — What I RAN (as opposed to read)
- `pnpm install --frozen-lockfile` (no `node_modules` was present).
- `pnpm typecheck` → **exit 0, all packages Done, 0 errors** at this head.
- `pnpm smoke:attach-reconnect` → **PASS, 72 passed, 0 failed**. Cell E exits clean on
  `seat rcseat is gone` (no infinite retry); cell F green including
  `refusal: anything else is worth another attempt` (bare `failed-precondition` → transient).
- Standalone adversarial probe of `attachRefusal` (above) → all cases pass.

### CI (enumerated at the full 40-char sha)
Present by name: **Code Quality** = success, **Docs** = success (path-filter satisfied — the PR
touches `docs/control-surface.md`, matching `docs.yml`'s `docs/**`), **Windows** = success (all
shards + soak green), **CI** = in_progress, **Mutation reproof** = in_progress. `in_progress`
workflows are UNMEASURED here, not green.

One measured **`smoke (shard 0/4)` FAILURE** on the required Linux `CI` run: `smoke:manager-stop-reap`
failed its serve-family verify-eviction scenario with *"the delivery daemon is not reachable on the
ctl.delivery-admin rail (timeout)"* — an `unavailable`-class delivery-daemon timeout. This suite is
**not in the delta and not in the PR at all** (untouched vs `origin/main`), references neither
`attachRefusal` nor the observation detail, and its `unavailable` code maps to `transient`
identically before and after this patch. It is the same environmental-timeout class as the earlier
cell-H flake, not a regression from this change. Shards 1–3 were still running at report time.

### Could not check
- The tail of the Linux `CI` run (smoke shards 1–3, `changed`) and `Mutation reproof` were still
  `in_progress` at report time; I did not block on them. Mutation reproof red is main-resident
  (#1320) and out of scope by the brief.

**Verdict: APPROVE** for the delta. The classification is narrowly keyed on a namespaced detail with
a confirmed writer, forwarded on every refusal branch that can carry it; all three anti-fixes are
avoided; the manager contract, docs, and changeset claims hold; and the one CI smoke failure is an
unrelated delivery-daemon timeout off the patched path.

PR #1315, opened 2026-09-05
fix(cli): select explicit refresh identity

davidfarah2003:
`cotal up --server <url> --space <name>` against a broker that already holds records for several spaces selected the first registry record whose server matched, so the requested refresh could land on a tenant record and refuse the root as foreign. An explicit refresh now selects by the full identity (server, root, space); a bare refresh keeps its existing resolution.

Scope:
- `implementations/cli/src/commands/up.ts`: explicit identity lookup on the live-refresh branch.
- `implementations/cli/smoke/up-refresh-identity-live.smoke.ts`: drives the shipped binary against an isolated broker with a multi-space account root, both registry filename orders, then asserts an absent explicit space and a foreign-root record still refuse. 20 checks, count asserted; isolated HOME/COTAL_HOME/XDG; scoped `down` of its own meshes.
- `scripts/mutations/up-refresh-identity.json`: restoring the server-only lookup kills the named omega-root assertion.
- CI suite fragment registered under `bin/smoke/ci-suites.d/`; `docs/run-a-mesh.md` updated; changeset (patch, cotal-ai and @cotal-ai/cli).

Ran: baseline red on the named assertion at main, tip green 20/20, `mutation-proof` 1/1 killed, restored green, CLI typecheck, `check:docs-voice` 40/40, `changeset status`.


davidfarah2003:
Retrospective review of `0ad4b5711d00d4fbe044adc7c544ed73f45c2fd9` (merge `9f2817d70bf39d18e2b0f4b7ac942d03ea0dcf60`, parent2 equals this head). **APPROVE** this SHA.

Read `implementations/cli/src/commands/up.ts` at that commit. An explicit `--space` refresh selects `server + this root + that space`. A miss, an absent requested space, and a record rooted elsewhere do not take the refresh arm. With `--server` they exit 1 rather than adopting another same-server identity. Same-root user-auth with no trust material exits 1. `findMesh` / `recordMesh` stay keyed by space name.

I did not run `cotal up`, `pnpm check`, or `up-refresh-identity-live`. GitHub `smoke (shard 0/4)` on run 34000523315 did run `pnpm smoke:up-refresh-identity:live` and printed all 20 named checks (both registry orders, absent-space refusal, foreign-root refusal) then `UP REFRESH IDENTITY LIVE SMOKE OK`. Mutation-reproof `changed` selected `scripts/mutations/up-refresh-identity.json` and KILLED `the refresh returns to the first same-server registry record`.

Residual, not a block: a bare refresh still uses the first same-server record, which this PR kept. The live fixture is `--open` only.

davidfarah2003:
**RETROSPECTIVE VERDICT: APPROVE** at `0ad4b5711d00d4fbe044adc7c544ed73f45c2fd9`.

I confirmed the exact local and remote head. Merge commit `9f2817d70bf39d18e2b0f4b7ac942d03ea0dcf60` has this head as parent 2, and the stable patch ID from its first parent to the merge equals the PR patch ID, so the merge carried the PR-owned patch unchanged.

Engineering trace:

- An explicit `cotal up --space <name>` on an occupied server now selects a registry record only when server, current project root, and space all match.
- Multiple records sharing a server cannot redirect an explicit refresh by filename order.
- A missing explicit space does not fall back to another same-server record. It proceeds to the existing occupied-listener refusal path.
- A foreign-root record does not become a same-root refresh.
- Bare refresh intentionally retains the existing first-same-server resolution.
- The live smoke drives the shipped `bin/cotal.ts` composition root in both registry filename orders, then covers absent and foreign-root refusals. The mutation restores the server-only lookup and names the omega-root assertion.
- The CI fragment filename is the SHA-256 of `smoke:up-refresh-identity:live`, its content and package script match, docs and generated bundle agree, the changeset is scoped correctly, and `git diff --check` passes.

I did not install dependencies or run any suite. The only behavioral smoke and its mutation command end in `-live`, so running them locally was prohibited. I reviewed their shipped-code path and the green CI status instead. That unexecuted local live path is the named verification gap.

davidfarah2003:
Acceptance supplement. Verdict remains **APPROVE** at `0ad4b5711d00d4fbe044adc7c544ed73f45c2fd9`.

I revalidated the change specifically at the public CLI and integration boundaries. Exact-head CI ran the real `bin/cotal.ts` composition root with `cotal up --detach --open --server <isolated broker> --space <name>` and observed all 20 acceptance checks:

- requested same-root space selected with the root record first
- requested same-root space selected with an unrelated tenant record first
- unrelated same-server record unchanged in both orders
- absent explicit space refused without rewriting either record
- foreign-root explicit record refused without rewrite
- every owned fixture mesh stopped cleanly with no leaked broker

The exact-head mutation reproof selected `scripts/mutations/up-refresh-identity.json`. Restoring first-same-server lookup reddened the named public assertion `explicit refresh selects omega-root even with same-server records` at 12 marks versus the 20-mark baseline, and graded KILLED.

Merge `9f2817d70bf39d18e2b0f4b7ac942d03ea0dcf60` still has this head as parent 2. From the real merge base, the PR-head patch and merge parent-1 patch have identical stable patch ID `40cb511ed15677565bf0ebf297c22dd75ff3f0d1` and identical eight-file sets.

Run disclosure: I read the exact-head hosted acceptance and mutation logs. I did not locally run `cotal up` or the `-live` suite.

PR #1316, opened 2026-09-05
fix(cli): support user endpoint commands

davidfarah2003:
The generic `describe`/`invoke` surface refused user-mode meshes at the mode check even though the ep rails already carry a user bearer's caller rows (the existing native B1e path). It now connects a user-auth mesh with the normal bearer plus sentinel pair and the caller triple from `resolveControlTarget`, through `dialerFor` and `standaloneConnectOpts`, honouring the recorded TLS requirement instead of a hardcoded plaintext dial. Static instruments are unchanged; open meshes still refuse.

Targeted `--name` calls resolve the alias through the name-keyed `inspect` (in the spawn set) instead of the manager-wide `ps`, so an authorized user does not need an unrelated `manager.read` grant. Routing tier is unchanged: `any` only with `--admin`, `owner` otherwise; the principal is parsed with `parsePrincipalKey`.

Scope:
- `implementations/cli/src/commands/describe.ts`
- `implementations/auth/smoke/user-spawn.smoke.ts`: B1g drives the real composition-root binary under a spawn-scoped user bearer: describe, authorized inspect, owner-tier `--name` call, nonexistent target refused by inspect resolution, `manager.read` and `manager.admin` still broker-denied. `COTAL_USER_ENDPOINT_CLI_ONLY=1` bounds the mutation run to that section; the full suite stays the CI coverage.
- `bin/smoke/mutations/user-endpoint-cli.json`: restoring the static-only guard kills the named describe assertion.
- `docs/cli.md`, `docs/control-surface.md`; changeset (patch, @cotal-ai/cli).

Ran: focused smoke green 38/38 (full user-spawn 115/115 earlier), `mutation-proof` 1/1 killed (39 baseline marks, 33 under the mutant), restored green, typecheck, `check:docs-voice` 40/40, `changeset status`. Independently re-executed by a second reviewer with the same result.


davidfarah2003:
Retrospective review of `3a63c892062078b5d8e87e83604ae516ea717097` (merge `d6a61b58a90d306f085f0fa5b6bb4f7a5a36a42b`, parent2 equals this head). **APPROVE** this SHA.

Read `implementations/cli/src/commands/describe.ts`, workspace `control-target.ts`, `SPAWN_SERVICE_COMMANDS` in `endpoint-grants.ts`, the manager `inspect` handler, and the B1g user-spawn patch at that commit. Open meshes and missing endpoint caller rows exit 1. The generic surface connects with static creds or a bearer plus sentinel, and passes `tls: t.auth.tls === true` (the previous path hardcoded `tls: false`). `--name` resolves through spawn-set `inspect`, not `ps`. A null `parsePrincipalKey` falls back to caller owner plus the bare id (static row shape). Without `--admin`, targeted calls stay `owner` mode. The CLI child in the smoke strips inherited `COTAL_` keys. `user-spawn.smoke.ts` left the FROZEN ambient list because it now strips before spreading.

I did not run the live suite locally. GitHub shard 2 of run 34006817062 ran `pnpm smoke:user-spawn:live` and passed B1g (describe, inspect, owner `--name`, missing-name inspect refusal, `ps` denied, `purge` denied), then `USER-SPAWN SMOKE OK (116 passed, 0 failed)`. Mutation-reproof selected `bin/smoke/mutations/user-endpoint-cli.json` and KILLED `restore the old generic CLI static-auth-only guard`.

Residual, not a block: the mutation pins the old static-only guard, not inspect-versus-ps. B1g covers that on the live broker. Unsupported still fails loud.

davidfarah2003:
**RETROSPECTIVE VERDICT: BLOCK** at `3a63c892062078b5d8e87e83604ae516ea717097`.

I confirmed the exact local and remote head. Merge commit `d6a61b58a90d306f085f0fa5b6bb4f7a5a36a42b` has this head as parent 2, and the stable merge patch ID equals the PR patch ID.

The product path is coherent: generic describe/invoke accepts the user bearer plus sentinel and caller triple from `resolveControlTarget`, uses the recorded TLS policy through `dialerFor`, resolves manager aliases through `inspect`, preserves owner/any targeting, and fails loud for unsupported commands and missing targets. B1g drives the real composition-root CLI and covers describe, inspect, owner attach, a missing target, and denied manager read/admin. The mutation pins the old static-only guard.

Blocking finding: this PR removes `implementations/auth/smoke/user-spawn.smoke.ts` from the ambient-environment frozen baseline after adding a COTAL-stripped environment for the new CLI child. The same file still launches its auth-service child with:

```ts
env: { ...process.env, COTAL_HOME: home }
```

That child spread remains unstripped. `suite-ambient-env.smoke.ts` classifies safety per file, not per spread or child: one matching `startsWith("COTAL_")` plus one `delete` anywhere marks the whole file stripped. The new CLI filter therefore falsely clears the separate auth-service launch.

I reproduced this safely without running a live suite. The baseline census passes. In a scratch exact-head tree, I changed only the new CLI filter to the semantics-equivalent `!/^COTAL_/.test(key)`. The census then failed and named `user-spawn.smoke.ts` as the ambient offender. This proves the green gate depends on recognizing one strip anywhere in the file, not on every child path being scrubbed.

Corrective options: keep the file in `FROZEN`, or scrub the auth-service child too. The durable fix is to make the census classify each environment spread/child rather than treating one strip as a clean bill for the entire file.

I ran only `node bin/smoke/suite-ambient-env.smoke.ts` and the scratch negative control. I did not install/build or run live suites.

davidfarah2003:
**RETROSPECTIVE VERDICT: BLOCK** at `3a63c892062078b5d8e87e83604ae516ea717097`. This withdraws my earlier APPROVE of the same SHA (comment 5557115536).

Merge `d6a61b58a90d306f085f0fa5b6bb4f7a5a36a42b` parent2 still equals this head.

The shipped describe/invoke path is still coherent. The block is a test-safety regression in the ambient-env ratchet.

I ran `node bin/smoke/suite-ambient-env.smoke.ts` at this head (no install, no `-live`, no stack). Baseline **PASS**, and the census prints `implementations/auth/smoke/user-spawn.smoke.ts - strips COTAL_ before the spread`. Commit `515e92770` deleted that path from `FROZEN` after this PR added a CLI child env that filters `!key.startsWith("COTAL_")`. The same file still spawns its auth-service child with `env: { ...process.env, COTAL_HOME: home }`. Classification is per file: `STRIP` (`startsWith("COTAL_")`) plus `DELETE` (`/\bdelete\b/`, which matches comment prose in this file) marks the whole file stripped, so the remaining unstripped spread is cleared.

Negative control, then restored: I changed only that filter to `!/^COTAL_/.test(key)` (same strip semantics). The census then failed with `user-spawn.smoke.ts` as the sole offender. I restored the file; HEAD remains `3a63c892062078b5d8e87e83604ae516ea717097`.

Corrective PR (do not rewrite history): keep the file frozen until every child strips, or strip the auth-service child too, and make the census reason per child/spread rather than per file.

davidfarah2003:
Corrective PR for the ambient-env census residual: https://github.com/Cotal-AI/Cotal/pull/1339 at `c0788e0fe403ba4becef75e72e33648b13301d25`.

The BLOCK at `3a63c892062078b5d8e87e83604ae516ea717097` stands until that lands. History is not rewritten.

davidfarah2003:
Follow-up: https://github.com/Cotal-AI/Cotal/pull/1339 is **closed**. The #1316 corrective lane is already owned elsewhere (`fix/1316-ambient-env-child-scope` from current main). #1339 was a race (`c0788e0fe` remains as a reference only).

#1316 stays **BLOCK** at merge `d6a61b58a` / head `3a63c8920` until a lane-owned corrective sha is pinned for independent re-review.

davidfarah2003:
Acceptance supplement. Verdict remains **BLOCK** at `3a63c892062078b5d8e87e83604ae516ea717097` for the ambient-environment ratchet regression, not for the shipped describe/invoke behavior.

Exact-head CI exercised the real generic CLI through the composition root and user-auth integration boundary:

- `describe manager` succeeded through the normal spawn-scoped bearer
- `invoke manager inspect` succeeded without a `manager.read` grant
- `invoke manager attach --name alpha` succeeded on owner routing without `--admin`
- a missing `--name` target failed through name-keyed `inspect`, not the denied `ps` scan
- `manager.read` remained broker-denied
- `manager.admin` remained broker-denied
- the full user-spawn suite completed 116 passed, 0 failed

The exact changed-set mutation restored the static-credentials-only guard and KILLED `the shipped CLI describes manager through the normal user's bearer` at 33 marks versus the focused 39-mark baseline.

The blocking public gate behavior also reproduced again at this exact head without running a live suite. Baseline `suite-ambient-env` passes and labels `user-spawn.smoke.ts` stripped. Replacing only `!key.startsWith("COTAL_")` with the semantics-equivalent `!/^COTAL_/.test(key)` makes the census fail and name that file, because its auth-service child still uses `env: { ...process.env, COTAL_HOME: home }`. The green result depends on recognizing one unrelated strip anywhere in the file.

Recorded TLS is passed by the public command into `dialerFor` and `standaloneConnectOpts`, and the shared workspace contract pins user bearer plus sentinel plus `tls:true`. The B1g live fixture itself is plaintext, so strict user-mode TLS remains source/option-contract evidence rather than an end-to-end acceptance result.

Merge `d6a61b58a90d306f085f0fa5b6bb4f7a5a36a42b` still has this head as parent 2. From the real merge base, the PR-head patch and merge parent-1 patch have identical stable patch ID `c495813f7c3350798a620e7879698c633d3fc6e0` and identical eight-file sets.

PR #1321, opened 2026-09-06
docs(design): multi-space daemons, one process serving N spaces

davidfarah2003:
Design for running thousands of spaces on one broker cluster without one auth-service, delivery and manager process per space. Non-normative; wire contract unchanged.

Scope:
- `docs/design/multi-space-daemons.md`: multi-space mode for the auth service (callout routed by account, public face routed by `/s/<space>` path, one bundle per space, hot add/remove with per-space fenced teardown) and the delivery daemon (one unit per space, N units per process, membership authority per account, bounded N with leases); the platform runs no manager and serves `manager-service-authority` from the face instead, with a resigner design for the daemon creds that no manager renews; broker accounts through the full NATS resolver pushed over `$SYS.REQ.CLAIMS.UPDATE` with a least-privilege resolver-admin credential profile, paired data and callout account saga reconciled with the U2 CAS inventory, write semantics derived from the nats-server v2.14.6 handlers (an ack is not a write), JetStream limits carried in the account JWT; the embedding seam; failure and isolation analysis; migration from process-per-space with no tenant re-keying.
- Builds on `per-space-lifecycle.md`, `space-segmentation-p7-p1.md` and `u2-resolver-cas.md`; answers the hosted half of embedding gap 3 and the roadmap "Multi-space brokers" row.

Review: three rounds with two independent lenses (security and isolation; engineering and spec fidelity), 40 file:line citations verified at the tree, nats-server behaviour re-read at the pinned release. Every contradiction found was folded; the remaining items are named residuals for the build phase in the document. `check:docs-voice` passes.


davidfarah2003:
BLOCK on `3d17a7b62ce549fddcbcb0dc0c5e8aa836ca9d69`.

I found three implementation-basis blockers.

1. **§6.8's removal mechanism does not stop a loaded account.** The document says at lines 530-532 that an expired claim "fails its next lookup" and therefore stops new connects. I reproduced the proposed operation against `nats-server` v2.14.6 with a full resolver. The account first accepted a data-account connection. `$SYS.REQ.ACCOUNT.<id>.CLAIMS.UPDATE` then accepted the expired operator-signed claim with `200 jwt updated`. A second connection using the old data-account credential was still accepted, and the first stayed open. The server logged that applying the resolver update failed validation. Source explains the result: the resolver write accepts the expired claim, but the changed callback calls `updateAccountWithClaimJWT`, whose `verifyAccountClaims` uses blocking time validation and rejects it, leaving the already-loaded live claim in place. The tenant-removal guarantee and the migration/removal saga built on it are therefore false.

2. **The proposed re-signer cannot call `remintDaemonCreds` unchanged without breaking the auth-service key boundary.** §5.2 line 280 and §7 lines 575 and 591-593 say the auth plane calls `remintDaemonCreds(root, space, store, ...)` unchanged. That function begins with `getSpaceAuth`, which reads `auth/broker.json` plus `auth/account.<key>.json` and returns the full `SpaceAuth`, including operator and account seed material. The shipped auth service deliberately loads only `auth/<segment>/service-keys.json`, whose contract says it is exactly the data-account public key plus signing seed, and nothing else. Either the unchanged function has no signer records in the auth store and every renewal returns `no-auth`, or those records are added and the process loads stronger secrets than the design says it holds. The design needs a re-sign seam that consumes the existing projected signer, or it must explicitly revise the trust boundary.

3. **The raw HTTP space path is not defined for valid shipped space names.** §3.2 specifies `/s/<space>/<route>` and says the raw space is validated by `spaceSegment`'s encoder. Shipped behavior intentionally accepts slash-bearing and Unicode space names. `auth-store.smoke.ts` asserts that `a/b` is a valid space and is encoded as one flat `space.612f62` segment. A base of `/s/a/b` no longer has one space segment, and raw `?`, `#`, percent escapes, and Unicode normalization introduce related URL parsing questions. The path must use a defined injective URL-segment encoding and compare the decoded value to the requested space. Validation after splitting does not make a raw name one segment.

There is also a docs-voice gap. `pnpm check:docs-voice` passed, but its output says `docs/design` is exempt, so it did not check this file. Applying its heading rule to the page finds multiple list-style headings, including "The manager, and the re-signer that goes with it", "Failure and isolation", and "A fenced plane must stop answering, not necessarily exit". The required command is green but is not evidence for this page.

What I ran and checked:

- Verified live PR head `3d17a7b62ce549fddcbcb0dc0c5e8aa836ca9d69`, one added file, 858 lines, no comments or reviews.
- Read the full design, `SPEC.md` §§9 and 13.13, both named sibling designs, and the cited auth, delivery, manager, workspace, and core paths.
- Ran `pnpm check:docs-voice` and `git diff --check`.
- Confirmed the docs bundle generator uses non-recursive `readdirSync(docsDir)`. This head has 40 top-level Markdown pages and 46 recursive, the 6 recursive-only pages are all under `docs/design`, the committed bundle is 1,118,561 bytes, and it contains zero `docs/design/` paths. No bundle update is required.
- Read the pinned v2.14.6 resolver source and ran the removal reproduction on the v2.14.6 release binary.

What I did not check:

- I did not run the full build, typecheck, or complete smoke suite because the contribution is one non-shipped design page.
- I did not independently exercise all seven external NATS behaviors in §6.9. I live-tested the removal behavior above and source-read the other resolver paths.
- I did not grade on PR CI. Shard 2 was cancelled during the repo-wide cancellation event, and the aggregate failure only classifies that incomplete run. It is neither evidence for nor against this design.
- The cluster case is explicitly out of scope and I did not validate it.

The re-signer finding has one named uncertainty: an embedding could choose to place full signer records in the same injected store even though the current auth-service projection forbids loading them. That would make the call possible, but it would still contradict the document's claimed secret boundary, so the text must choose and specify one model.

This design is structural and rests on largely undocumented resolver behavior. After these blockers are revised, I recommend a second independent lens on the replacement removal mechanism and resolver mutation proof.

davidfarah2003:
Follow-up to the block above, from the same review. These were driven live against nats-server v2.14.6 rather than derived from source, and they change what a safe removal design has to assert. Bound to the same head, `3d17a7b62ce549fddcbcb0dc0c5e8aa836ca9d69`.

## DELETE is stronger than expired UPDATE, but its 200 is still not proof of removal

On the ordinary path `$SYS.REQ.CLAIMS.DELETE` does apply synchronously: `store.delete()` removes the file and invokes `store.deleted(publicKey)`, whose `removeCb` marks the loaded account expired and disconnects clients before the handler replies. Observably: reply `200 deleted 1 accounts`, a new connection with the old credential refused, the pre-existing connection closed.

The drift edge defeats it. With the resolver file removed out of band and the account still loaded, the same valid operator-signed DELETE returns the same `200 deleted 1 accounts` — but a new old-credential connection is **accepted** and the pre-existing connection **stays open**. `ENOENT` returns nil before `store.deleted(publicKey)`, so no loaded-account callback runs, while the handler still increments its pass count and reports success.

So `200 deleted N` means the filesystem delete returned success, not that the server proved the loaded account disabled. That is the same acknowledgement-versus-postcondition hazard as the expired-UPDATE case, and it appears exactly when someone would reach for an emergency revocation.

## The postcondition is observable, without attempting a tenant connection

`$SYS.REQ.ACCOUNT.<id>.INFO` is the instrument. Its handler calls `accountInfo(id)`, which reads `s.accounts` directly and returns `AccountInfo.expired`; it does not fetch from the resolver, and the reply carries the server identity envelope so the reading is attributable to the responding server. `CLAIMS.LOOKUP` replies do not.

Three live cells:

| | resolver file | account | DELETE reply | ACCOUNT.INFO | old credential |
| --- | --- | --- | --- | --- | --- |
| 1 | present | loaded | `200 deleted 1 accounts` | `expired:true`, `client_connections:0` | refused; existing connection closed |
| 2 | **already absent** | loaded | `200 deleted 1 accounts` | **`expired:false`, `client_connections:1`** | **accepted; existing connection stayed open** |
| 3 | present | never loaded | `200 deleted 1 accounts` | `Account <id> does not exist` | refused |

Cell 2 is the failure state, and it is invisible to `CLAIMS.LOOKUP` because absence is precisely what it already has. That is why lookup-absence cannot serve as the postcondition — the failing case satisfies it.

**Suggested predicate after DELETE:** LOOKUP absent **and** ACCOUNT.INFO reporting either `expired:true` or that the account does not exist.

## Three cautions on writing that check

**Do not use `client_connections:0` as the disablement proof.** Zero can mean merely idle. It is an absence, and an absence is satisfied by the wrong condition as easily as the right one. `expired:true` is a positive state bit the broker sets; assert on that.

**If removal also promises live eviction, that is a second postcondition needing its own proof** — an account-scoped CONNZ sweep reaching zero *after* INFO reports expired. On the ordinary path the disconnect is a side effect of a callback that, on the drift path, never runs. INFO initiates nothing; it only observes.

**The resolver-admin profile does not currently grant `$SYS.REQ.ACCOUNT.<id>.INFO`.** A postcondition the operator's credential cannot observe is not a postcondition, so a subject-scoped row for the exact account id is part of the design rather than a deployment detail.

## Boundary

This is sufficient for the single-nats-server boundary the document describes. A multi-server form needs declared-roster coverage, because each server keeps its own loaded-account map. ACCOUNT.INFO replies are at least attributable per server, which `CLAIMS.LOOKUP` is not. The design should state which boundary it claims; silence will be read as the larger one.

Recommend pinning both the ordinary-path and the absent-file cells in the prerequisite smoke, and pinning the observable facts (new connection accepted, existing connection still open) rather than the return value — the return value is the thing that lies.

davidfarah2003:
## Independent review of `3d17a7b62ce549fddcbcb0dc0c5e8aa836ca9d69`

**Verdict: do not merge this SHA. Merge after the named edits below and re-review the revised design.** The proposed direction is plausible, but two load-bearing mechanisms are incorrect or incomplete as written.

### Blocking findings

1. **The expired-claim removal path does not stop new connections to an already-loaded account.**

   Section 6.8 says an expired claim is pushed and "an already-loaded account fails its next lookup," so steps 2 and 3 stop new connects (`docs/design/multi-space-daemons.md:523-533`). That is not what nats-server v2.14.6 does.

   In the full resolver, the push is stored through `DirJWTStore.save`. Its change callback calls `updateAccountWithClaimJWT`. That function runs the blocking trust/time validation and, when the pushed claim is expired, returns an error without replacing the live in-memory account. `LookupAccount` then returns the still-loaded account unless its existing applied claim is expired.

   I reproduced this against the exact v2.14.6 release binary and source tag (`1aa10f9fe4e7a27b7d877af004a9c0022fdc4910`):

   - push a valid data-account claim
   - connect once so the account is loaded
   - push the same account with an expiry in the past
   - confirm the expired bytes are in `<resolver>/<account>.jwt`
   - connect again with the existing account's user credential

   The second connection succeeded. The server returned `jwt updated` for the expired push and the resolver file contained the expired claim, but the loaded account continued accepting new connections.

   This invalidates the tenant removal saga, the compromise/recovery analysis that relies on a pushed bad claim immediately stranding a live account, and behavior 7 in section 6.9 as currently stated. Replace expiry-push removal with a mechanism proven to disable an already-loaded account, and add the loaded-account positive/negative cases to P8.

2. **The auth-plane re-signer cannot use `remintDaemonCreds` "unchanged" without widening the auth service's key boundary.**

   Section 5.2 and the seam table say the authority plane already holds the needed signer and calls `remintDaemonCreds(root, space, store, { preflight })` unchanged (`docs/design/multi-space-daemons.md:265-285`, `:568-593`). The current helper does not consume the auth service's `service-keys.json` projection. It calls `getSpaceAuth`, which loads `auth/broker.json` plus `auth/account.<key>.json`, or the legacy full/stripped bundle (`packages/workspace/src/renewal.ts:98-160`, `packages/workspace/src/auth-paths.ts:853-899`).

   Current auth preparation deliberately gives the long-lived service only `{ dataAccount: { pub, signingSeed } }`; the operator and account seeds never enter the service (`implementations/auth/src/provider.ts:55-75`, `implementations/auth/src/store.ts:267-295`, `docs/identity-and-auth.md:146-149`). Calling the existing helper against the normal split trust store reads stronger material than the service currently loads. Calling it against only the four auth-service keys returns `no-auth`.

   Define the actual seam before merge. A narrow helper that accepts the existing data-account signer projection and preserves the current preflight, identity, profile, and store-key guarantees would fit the argument. Alternatively, explicitly specify and justify a different stored signer shape and update the security boundary. The current text claims both "unchanged helper" and "no key moved into the process," which cannot both hold for the current split records.

3. **The HTTP tenant prefix has no encoding rule, although supported space names are not URL path tokens.**

   Section 3.2 specifies `/s/<space>/<route>` and says to split the path and validate the segment with `spaceKey` (`docs/design/multi-space-daemons.md:102-125`). `spaceKey` intentionally supports arbitrary well-formed Unicode and slash-bearing names. The auth store smoke explicitly proves space `a/b` is supported. A raw `/s/a/b/exchange` cannot be split into one tenant segment, while `%2F`, `%`, `?`, `#`, Unicode normalization, and decoding rules are left unspecified.

   Use the existing injective `spaceKey`/`spaceSegment` in the URL, or define a canonical encode/decode operation with rejection and round-trip rules. Add examples for `a/b`, case-differing names, Unicode, and malformed escapes. This is a tenant-routing boundary, not only a presentation detail.

### Shipped versus proposed

The status is clear overall. The first line says **Design, non-normative, not shipped**, section 2 isolates current facts, sections 7 and 12 name proposed seams and prerequisites, and section 10 says the wire is unchanged. I did not find a place where the multi-space daemon mode itself is presented as shipped.

One current-doc contradiction should be resolved: `docs/identity-and-auth.md:124-125` lists the public face as only `/health`, `/jwks`, `/exchange`, and `/.well-known/cotal-mesh`, while current code and this design include `/manager-service-authority`. The design is correct about the code. Update the public guide so the two documents do not disagree.

I found no contradiction with the normative Cotal wire contract. The single-server reclaim boundary matches SPEC 13.13, and the proposed HTTP prefix and nats-server resolver subjects are outside Cotal's wire grammar.

### Voice and size

The status and technical prose are generally direct. The file contains no em dashes and none of the banned filler words `exactly` or `fold`. A few slogan contrasts, such as "An ack is not a write" and "A re-sign is a clock extension, not a grant," are useful summaries rather than merge blockers.

At 858 lines and about 11,000 words, it is longer than necessary. Sections 6.2, 6.5, 6.6, and 6.9 repeat the plain-save/no-ordering argument; sections 5.1 and 5.2 repeat their rationale in rejected alternatives 6 and 7; section 13 repeats decisions already named beside the mechanism. After correcting the blockers, consolidate the NATS behavior contract and version floor, combine proof and ordering, and make rejected alternatives/residuals point back to the primary discussion. A reduction of roughly 200 to 300 lines would preserve the substantive design.

### What I read and ran

- Verified local `HEAD` and `gh pr view 1321 --json headRefOid` both equal the SHA above.
- Read all 858 lines of the new document, its three linked prerequisite designs, SPEC 13.13, and the relevant architecture, embedding, identity/auth, run, workflow, and CLI guides.
- Inspected the cited auth, delivery, manager, workspace, provisioning, URL, renewal, membership, and runtime-record source paths at this head.
- Cloned nats-server tag v2.14.6 and inspected its claim update, lookup, list, delete, resolver save, validation, and account lookup paths.
- Ran the isolated exact-version live resolver probe described above.
- Ran `git diff --check`, a local link-target check, and verified the worktree remained unchanged.
- I did not run a repository smoke suite. This PR changes only a design document, and the decisive behavior was reproduced directly against the external component the design pins.

davidfarah2003:
## Orchestrator disposition: blocked on design, not on queue position

Two independent reviews, two days apart, against the same unchanged head
`3d17a7b62ce549fddcbcb0dc0c5e8aa836ca9d69`, reached the same three blockers. Both drove them live
against the exact `nats-server v2.14.6` this design names, rather than reasoning from the document.

- 2026-09-08 — comments https://github.com/Cotal-AI/Cotal/pull/1321#issuecomment-5580113567 and
  https://github.com/Cotal-AI/Cotal/pull/1321#issuecomment-5580526156
- 2026-09-10 — comment https://github.com/Cotal-AI/Cotal/pull/1321#issuecomment-5615292105

The three, as reproduced:

1. **The removal mechanism does not remove.** An expired full-resolver claim push does not stop new
   connections to an account that is already loaded. §6.8 rests on it doing so.
2. **The key boundary is crossed.** `remintDaemonCreds` is unchanged and conflicts with the
   auth-service signer projection.
3. **The raw `/s/<space>` path is undefined** for the slash-bearing and Unicode space names this
   design says it supports.

These are defects in the design itself, not in an implementation of it, so there is nothing a lane
could go and fix on the branch. A replacement removal mechanism has to be designed before a
re-review is worth anyone's time; the other two are then downstream of whatever that mechanism turns
out to be.

**Left open deliberately.** Whether to rework or withdraw is the author's call, not the merge gate's.
Labelled `triage:blocked` and taken out of the stale-PR queue, so it stops being counted as a
merge-ready item waiting on a reviewer.

**A correction to the record, since I put the wrong reading into circulation.** I had described this
as one of the cheapest items on the board. That came from its mergeability — `MERGEABLE/CLEAN` with
no conflicts — and mergeability describes the branch, not whether anyone has cleared the change.
`reviewDecision` is empty here as it is on every PR in this repo, because verdicts are recorded as
comments rather than as GitHub reviews, so neither field carried the block that was already sitting
in the thread. Read the comments before ranking a PR by its badges.

PR #1326, opened 2026-09-06
fix(test): bypass pnpm dependency checks in mutations

### LINKED ISSUE #1325: mutation: manager-runtime-deps cells exit 1 with 0 marks on main, blocking the 0.46.0 release reproof
`bin/smoke/mutations/manager-runtime-deps.json` fails on `main` and on every PR that selects it. All four cells report the same thing:

```
baseline: pnpm smoke:manager-runtime-deps
baseline green (54 progress marks)

  exited 1 but never printed the expected failure:
  "published manager runtime dependency is installed: @nats-io/transport-node"  · 0 marks (baseline 54)
  ... same for @nats-io/kv, @nats-io/jetstream, and the devDependencies-move cell
```

**It is currently blocking the 0.46.0 release.** PR #1305 is green on CI, Windows and Code Quality at attempt 1 (Docs legitimately path-filtered), and `Mutation reproof` is its only failure — this fixture, and nothing else.

## Ruled out against the base

Not release-specific and not caused by any PR. Main's own Mutation reproof run `33981431625` produces the **character-identical** result: same four cells, same `0 marks (baseline 54)`.

## What is established

- **Baseline is green at 54 marks**, so the suite runs correctly on an unmutated tree.
- **`0 marks` means the suite crashed before its first `check()`.** It printed no `✓` or `✗` line at all, so the named cell was never reached and cannot be graded.
- **The mutations are well-formed.** I applied each of the four to `implementations/manager/package.json` and parsed the result: all four produce **valid JSON**. Cell 1 moves `@nats-io/jetstream`, `@nats-io/kv` and `@nats-io/transport-node` from `dependencies` into `devDependencies`; the other three delete one dependency line each. Both shapes match the incident the fixture's `why` describes.
- **The suite's own command contains no build step.** `smoke:manager-runtime-deps` is `tsx bin/smoke/manager-runtime-deps.smoke.ts`, and the suite's first actions require `implementations/manager/dist/index.js` to exist so it can walk the entrypoint's import graph.

## What is NOT established

**Why it crashes.** I falsified two hypotheses before stopping: that the mutation desynchronises `pnpm-lock.yaml` (the script never invokes pnpm's resolver, it is bare `tsx`), and that the mutation corrupts the JSON (all four parse). A third — a missing `dist` — does not survive either, because baseline and mutant run against the same working tree seconds apart, so a `dist` present for the baseline is present for the mutant.

I am filing without a mechanism rather than guessing a fourth. The reproduction is cheap for whoever picks it up: apply any one of the four mutations to `implementations/manager/package.json` on a built tree and run `pnpm smoke:manager-runtime-deps`, capturing stderr. The exit path that produces `0 marks` will name itself immediately.

## Why it matters beyond this release

A mutation that cannot be graded is not a weaker signal, it is an absent one, and this fixture has been absent for at least as long as main's last several reproof runs. The guard it claims to enforce — *every external package loaded by the published manager Node entrypoint is installed as a production dependency* — is exactly the 0.33.9 incident it was written for, and right now nothing is checking it.

This is a **second, independent** instance of the class in #1320: a mutation fixture whose verdict cannot be produced. Different fixture, different cause, same consequence — a red reproof on unrelated PRs and a guard that is not guarding.

Related: #1320, #1301, #1302, #1305.

davidfarah2003:
## Summary

- disable pnpm's pre-run dependency verification only for commands launched by `mutation-proof`
- keep mutation fixture commands unchanged, so dependency-metadata mutants reach their real suites under CI
- pin the child-scoped override in the mutation-proof self-test

## Cause

pnpm 11 defaults `verifyDepsBeforeRun` to `install` for `pnpm run` and `pnpm exec`. A mutation that changes dependency metadata intentionally desynchronizes `package.json` from `pnpm-lock.yaml`. Under CI's frozen lockfile, pnpm aborts before the configured suite starts, so mutation-proof sees zero progress marks instead of the guard's verdict.

Without `CI` set, the same launcher silently rewrites `pnpm-lock.yaml` before running the suite. That side effect hid the failure locally and is recorded here as a tooling footgun. This change prevents it only inside mutation-proof child runs.

There are 9 mutation cells across 5 fixtures that target a `package.json`. Four manager-runtime-deps cells are demonstrably broken by dependency-metadata desynchronization. The remaining five are latent exposure, not known failures, because script-only manifest mutations can leave the lockfile consistent.

## Validation

- `node scripts/mutation-proof.selftest.mjs`: 33 checks passed
- `CI=1 node scripts/mutation-proof.mjs --config bin/smoke/mutations/manager-runtime-deps.json`: baseline 54 marks, all 4 mutations KILLED on their named dependency checks
- `CI=1 node scripts/mutation-proof.mjs --config bin/smoke/mutations/package-test-contract.json`: baseline 79 marks, all 4 mutations KILLED; the comparison cells remained at 76 and 77 marks

Closes #1325

### DIFF
--- a/scripts/mutation-proof.mjs
+++ b/scripts/mutation-proof.mjs
@@ -121,6 +121,10 @@ function run(command, cwd, timeoutMs) {
   // budget, one mutation into a sweep, with the tree mutated the whole time.
   const r = spawnSync(command, {
     cwd, shell: true, encoding: "utf8", timeout: timeoutMs, maxBuffer: 64 * 1024 * 1024,
+    // A mutation may deliberately desynchronize dependency metadata from pnpm-lock.yaml. pnpm's
+    // default pre-run check would install (or fail under CI's frozen lockfile) before the suite can
+    // observe that mutant. Disable only that check, and only in this child process tree.
+    env: { ...process.env, pnpm_config_verify_deps_before_run: "false" },
     detached: process.platform !== "win32",
     killSignal: "SIGKILL",
   });
--- a/scripts/mutation-proof.selftest.mjs
+++ b/scripts/mutation-proof.selftest.mjs
@@ -60,6 +60,7 @@ writeFileSync(
   join(root, "suite.mjs"),
   [
     "import { admit } from './src/impl.js';",
+    "if (process.env.pnpm_config_verify_deps_before_run !== 'false') { console.error('AssertionError: mutation child disables pnpm dependency verification'); process.exit(1); }",
     "console.log('  ✓ admits a small value');",
     "if (admit(5) !== true) { console.error('AssertionError: small values are admitted'); process.exit(1); }",
     "console.log('  ✓ the guard refuses an oversized value');",
@@ -71,7 +72,14 @@ writeFileSync(
 execSync("git init -q && git add -A && git -c user.email=a@b -c user.name=c commit -qm fixture", { cwd: root });
 
 const runTool = (args) =>
-  spawnSync(process.execPath, [TOOL, ...args], { cwd: root, encoding: "utf8", timeout: 120_000 });
+  spawnSync(process.execPath, [TOOL, ...args], {
+    cwd: root,
+    encoding: "utf8",
+    timeout: 120_000,
+    // Prove the tool overrides this only for commands it launches rather than relying on ambient
+    // configuration already carrying the desired value.
+    env: { ...process.env, pnpm_config_verify_deps_before_run: "install" },
+  });
 
 // 1. A mutation the suite DOES catch, named. The everyday case.
 let r = runTool([


davidfarah2003:
## APPROVE

Exact head: `90cebf04dac756cf0e52c2a59d8dbe40d3546007` (matches `git rev-parse HEAD` and `gh pr view 1326 --json headRefOid`).

**Negative-controlled the fix: yes.** Misspelling `pnpm_config_verify_deps_before_run` as `pnpm_config_verify_deps_before_runn` in `scripts/mutation-proof.mjs` `run()` restored the original defect. Restoring the spelling restored the kills. The env var is load-bearing, not cargo.

Did `pnpm install --frozen-lockfile` in this worktree.

### 1. Central probe

**RAN** (not read):

- Self-test green at HEAD: `node scripts/mutation-proof.selftest.mjs` → 33 checks, parent env still unset after.
- Negative control (misspelled key, `CI=1`, `--allow-dirty`):
  - self-test: first check fails, `REFUSING: … is red BEFORE any mutation` because the fixture suite asserts children receive `false`.
  - `manager-runtime-deps`: baseline 54 marks, then all four cells `WRONG-RED` at **0 marks**, `never printed the expected failure: "published manager runtime dependency is installed: …"`. Same signature as #1325.
- Positive control after restore (clean tree): all four **KILLED** — M1 51 marks, M2/M3/M4 53 marks, each naming its cell. Baseline 54.

Independent of the author's summary: pnpm 11.1.2 (`~/.cache/node/corepack/v1/pnpm/11.1.2/dist/pnpm.mjs`) actually reads this spelling.

- Default is `"verify-deps-before-run": "install"`.
- `addSettingsFromWorkspaceManifestToConfig` assigns `process.env.pnpm_config_verify_deps_before_run` onto `verifyDepsBeforeRun` when the var is non-null.
- `pnpm run` / `pnpm exec` call `runDepsStatusCheck` when that setting is truthy; `"install"` then runs install (or, under `CI`, frozen-lockfile abort) before the script.
- pnpm itself injects `extraEnv.pnpm_config_verify_deps_before_run = "false"` for nested script env, so the string `"false"` is pnpm's own bypass token. The PR uses that token. Casing of this special-case is lowercase `pnpm_config_…`; that is what the PR sets.

`"false"` is a non-empty string, so the `if (opts.verifyDepsBeforeRun)` gate still enters `runDepsStatusCheck`, but the `switch` only acts on `install|prompt|error|warn`, so a stale tree is neither installed nor failed. That is the intended no-op.

### 2. Scope of the environment change

- **Leak:** none to the parent. Set only as `spawnSync` `env: { ...process.env, pnpm_config_verify_deps_before_run: "false" }` inside `run()`. `process.env` of mutation-proof is not mutated. Confirmed parent unset before and after the self-test. Applies to every command that `run()` launches (baseline and each mutant), which is the intended child tree, not the operator shell.
- **What it hides:** a genuinely stale `node_modules` vs lockfile during a reproof run is no longer caught by pnpm's pre-run check. Acceptable: CI already `pnpm install --frozen-lockfile` before this tool; the mutation *is* a deliberate desync, so the check is always the wrong signal on mutant cells. The cost is that mutation-proof will not additionally notice an unrelated stale install. That is a real trade, not a free win.
- **Self-test can fail:** proven. Break the spawn key and it reddens immediately on the new assertion. Ambient `install` in `runTool` plus the child `!== 'false'` check is the writer.

### 3. Regression control

**RAN** `CI=1 node scripts/mutation-proof.mjs --config bin/smoke/mutations/package-test-contract.json` at this head.

- baseline 79 marks
- all 4 KILLED
- comparison cells **76 and 77** (against baseline 79), matching the pre-change #1305 log the author cited
- remaining two cells 160 and 78 (the assertion-counter mutant is supposed to emit extra marks)

Did not sample a third fixture family. PTC is a different family (script-field `package.json` mutant plus suite-source mutants) and is the corroboration the review asked for.

Independent census of `bin/smoke/mutations/*.json`: **9 `package.json` cells across 5 fixtures**. Matches the body.

### 4. Claims

- **Census wording:** body says 9 cells / 5 fixtures **at risk**, 4 demonstrably broken, 5 latent. That distinction is correct. `package-test-contract` mutates a test script, not dependency metadata, and graded here. The other latent cells (`package-json-duplicate-keys`, `sandbox-guard` M5/M6, `smoke-ci-build-first`) also edit scripts, not dep blocks.
- **Footgun:** stated. "Without `CI` set, the same launcher silently rewrites `pnpm-lock.yaml`."
- **Closes:** only `Closes #1325`. No other `#NNNN` in the body.
- **Changeset:** none in the PR. No package should bump (test/CI machinery only). The house pattern for a CHANGELOG line with no bump is empty frontmatter (`---\n---`). That file is absent. Non-blocking: the gate this unblocks is the reproof, not a publish bump. Worth adding if the release notes should mention the launcher fix.

### CI at `90cebf04dac756cf0e52c2a59d8dbe40d3546007`

Enumerated via the runs API at the full 40-character sha.

| Workflow | Present by name | Conclusion |
|---|---|---|
| Mutation reproof | PRESENT | **success** |
| CI | PRESENT | **UNMEASURED** (`in_progress`) |
| Windows | PRESENT | **UNMEASURED** (`queued`) |
| Code Quality: PR #1326 | PRESENT | **success** |
| Docs | correctly absent | path filter is `website/**`, `docs/**`, `spec/**`, `SPEC.md`, `assets/**`, `.github/workflows/docs.yml`; this PR touches only `scripts/mutation-proof.mjs` and `scripts/mutation-proof.selftest.mjs` |
| Installer | correctly absent | path filter is `install.sh`, `install/**`, `.github/workflows/installer.yml` |

Do not treat CI or Windows as green.

### Could not check

- Full 342-fixture sweep (cost; sampled the broken fixture and one other family).
- CI / Windows conclusions at this sha (still queued / in progress).
- Live lockfile rewrite without `CI` (would dirty `pnpm-lock.yaml`; body claim is consistent with pnpm 11.1.2 `case "install": install2()`).

### Ran vs read

**Ran:** `pnpm install --frozen-lockfile`; self-test; self-test with misspelled env; `manager-runtime-deps` mutation-proof with misspelled env; same after restore; `package-test-contract` mutation-proof; pnpm 11.1.2 source for the env mapping; issue-ref scan of the PR body.

**Read:** PR body, the two-file diff, fixture JSON, `docs.yml` / `installer.yml` path filters, existing empty-frontmatter changeset example.

Worktree source is clean (`git status --porcelain` empty). Temporary misspell of the env key was restored with `git checkout HEAD -- scripts/mutation-proof.mjs` before the positive control. No leftover mutation-proof processes.

PR #1301, opened 2026-09-05
fix(manager)!: leave managed agents running on a stack stop

davidfarah2003:
## Summary

Bare `cotal down` and `Manager.stop()` leave managed agents running by default. `cotal down --with-agents` and `Manager.stop({ withAgents: true })` are the explicit destructive paths.

Linux PTY seats release only the manager proxy to their detached custodians, so the seat process survives manager exit. Legacy in-process PTY runtimes fail closed instead of claiming they can detach.

Bare down checks spare capability inside `stopLocalProcess.beforeSignal`, while holding the exact `.stopping` reservation and after verifying the exact manager process identity. A manager successor cannot inherit a predecessor's capability check. Destructive intent is likewise one-shot and bound to the exact manager and exact live stopper.

The Manager.stop contract migration makes destructive smoke cleanup explicit. A frozen inventory now fails if another live-PTY smoke silently adds a bare manager stop, while dedicated sparing and preserve-state proofs retain the default path.

## Validation

- current-main base: `b00f3c1d267e350a67c93e5892e3c1c1209f5497`
- repository aggregate build and typecheck
- shutdown intent/capability proof: 15/15, mutations 4/4 killed
- custodial PTY proof: 10/10, mutations 4/4 killed
- destructive shutdown/lease proof: mutations 4/4 killed
- down orchestration proof: 19/19
- presence-render census: 17/17, mutations 9/9 killed
- pid attribution proof: 84/84, mutations 4/4 killed
- manager-stop inventory guard: 23/23, mutations 3/3 killed
- attach reconnect baseline: 74/74; current durable static-slot reachability mutation killed; two measured harness gaps remain explicit `unkillable` metadata
- mutation fixture inventory: 396 files / 2101 anchors; docs bundle, changeset, diff and clean-tree checks

Historical pre-rewrite head archived at `archive/1301-60a4f644f`.

Refs #964

PR #1327, opened 2026-09-06
test(manager): detect overlapping lease renewals

### LINKED ISSUE #1320: mutation-proof: a file with exactly one cell in a fixture cannot be graded when that cell goes quiet
`mutation-proof` cannot grade a mutation whose file is represented by exactly one cell in its fixture. When that cell applies and the suite stays green, there is no companion mutation in the same file to distinguish *the suite never reaches this file at runtime* from *this mutant changed no behaviour*. The runner correctly returns UNGRADABLE rather than a false SURVIVED, and the fixture fails.

**This has now fired for real.** `implementations/manager/smoke/fixtures/lease-renew.mutations.json` fails on `main` (Mutation reproof run `33981431625`) and on every PR that selects it, on cell D3:

```
D3   implementations/manager/src/manager.ts    the next tick may start a second renew while one is in flight
D2a  packages/core/src/streams.ts              the renew period goes back to TTL/2          KILLED
D2b  packages/core/src/streams.ts              the attempt deadline goes back to default    KILLED
```

D3 is the only `manager.ts` cell, so it has no internal positive control. Because the fixture is selected whenever a diff touches `manager.ts` or `streams.ts`, it charges the reproof budget and reds the gate on PRs whose diffs have nothing to do with it.

## The census

Measured across `origin/main` by reading every fixture and grouping its mutations by target file:

```
fixtures scanned                        342
(fixture, file) pairs with exactly 1 cell   205
distinct fixtures holding at least one      146   (43%)
```

Most solo cells are represented by `bin/smoke/mutations/endpoint-warning-consumers.json` (9 files with one cell each), then `transform-differential`, `core-boundary`, `jcode-diagnostics` (4 each).

**These 205 are not 205 defects, and the distinction matters.** A solo cell that is KILLED is fine: the kill itself proves the suite reaches the file. A solo cell only becomes ungradable when it goes quiet. So the census counts *latent* sites where a future quiet mutant yields UNGRADABLE instead of a verdict, not present failures. One of them has fired so far.

## Proposed

1. Repair `lease-renew` by adding a `manager.ts` mutation that must redden a named cell. All three outcomes are informative: if the control is killed and D3 stays green, D3 becomes a reportable SURVIVED and a real coverage gap; if the control is not killed, the suite does not exercise `manager.ts` and D3 belongs in a different fixture.
2. Consider having the fixture auditor warn (not fail) when a fixture introduces a file with a single cell, so new instances are visible at authoring time rather than at gate time.

Filing the aggregate rather than 146 issues. The immediate work is item 1.

davidfarah2003:
## Summary

The `lease-renew` D3 cell was ungradable on `main`, not only on a pull request. It returned the same `UNGRADABLE` verdict in main run `33981431625`, in #1301, and in #1302.

The first commit adds a positive control against the same `manager.ts` guard line. The control was killed while D3 remained green, converting D3 to a reportable `SURVIVED` verdict and proving the gap was real rather than a reachability artifact.

The second commit closes that coverage gap. The old assertion looked for the absence of a `wrong last sequence` diagnostic. That could not distinguish a genuine lack of conflict from a conflict whose evidence never arrived. During the measured window, the stall also holds the broker's rejection, and the manager suppresses repeated same-state diagnostics. The smoke now counts lease update publish frames at the relay, observing the overlapping renew attempt itself instead of a downstream symptom that may never be emitted. The existing `NO SELF-INFLICTED CAS CONFLICT` cell name and D3 `expectRed` remain unchanged.

This unblocks #1301 and #1302. Both are approved at their exact heads and were held solely by this fixture.

## Validation

The unmutated suite passes all 17 checks. The final committed whole-fixture reproof reports:

- Positive control: `KILLED`, 7 marks against baseline 17, at `[acklost] the child manager started, acquired its own per-instance lease, and renewed it at least once`
- D3: `KILLED`, 16 marks against baseline 17, at `NO SELF-INFLICTED CAS CONFLICT`
- D2a: `KILLED`, 16 marks against baseline 17, at `THE RENEW BUDGET HAS SLACK`
- D2b: `KILLED`, 16 marks against baseline 17, at `THE RENEW BUDGET HAS SLACK`

Also run:

- `pnpm build`
- `pnpm smoke:lease-renew`
- `pnpm changeset status`

A `mutation-proof` process terminated by an outer harness timeout leaves the current mutation applied because restoration only runs when the mutation runner finishes. Two timed-out validation attempts required explicit source restoration and a core rebuild before continuing.

Closes #1320

### DIFF
--- a/.changeset/lease-renew-manager-control.md
+++ b/.changeset/lease-renew-manager-control.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add a manager runtime reachability control to the lease-renew mutation fixture.
--- a/.changeset/lease-renew-manager-control.md
+++ b/.changeset/lease-renew-manager-control.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-Add a manager runtime reachability control to the lease-renew mutation fixture.
+Add a manager runtime reachability control and direct overlap check to the lease-renew mutation fixture.
--- a/implementations/manager/smoke/fixtures/lease-renew.mutations.json
+++ b/implementations/manager/smoke/fixtures/lease-renew.mutations.json
@@ -3,6 +3,13 @@
   "progressPattern": "^  ok   ",
   "mutations": [
     {
+      "name": "PREDICTED KILLED (1 red) - POSITIVE CONTROL: an idle renew tick returns before doing any work, so the manager never renews its lease once",
+      "file": "implementations/manager/src/manager.ts",
+      "find": "    if (this.leaseRenewInFlight) return;\n",
+      "replace": "    if (!this.leaseRenewInFlight) return;\n",
+      "expectRed": "[acklost] the child manager started, acquired its own per-instance lease, and renewed it at least once"
+    },
+    {
       "name": "PREDICTED KILLED (1 red) - D3: the next tick may start a second renew while one is in flight, on the same cached revision",
       "file": "implementations/manager/src/manager.ts",
       "find": "    if (this.leaseRenewInFlight) return;\n",
@@ -26,5 +33,5 @@
       "afterRestore": "pnpm --filter @cotal-ai/core build"
     }
   ],
-  "_note": "THE SUITE LOADS dist FOR CORE. It imports `@cotal-ai/core`, whose exports point at ./dist/index.js, so mutating packages/core/src leaves the module the run actually uses untouched; the command rebuilds core from the mutated source, and `afterRestore` rebuilds it from the restored source so dist is not left carrying the last mutant. The manager mutation needs no build: the child loads ../src/manager.js through tsx. D2a and D2b are graded on the numbers rather than on behaviour ON PURPOSE: each leaves the reconcile in place, so the manager still survives the stall and every behavioural cell still passes while the slack needed for several timely attempts is gone. Arithmetic is the only thing that catches that. The old fail-close and unconfirmed-TTL-budget mutations were removed with the behaviour they graded: lease loss no longer ends this process or consumes a local serving budget."
+  "_note": "THE SUITE LOADS dist FOR CORE. It imports `@cotal-ai/core`, whose exports point at ./dist/index.js, so mutating packages/core/src leaves the module the run actually uses untouched; the command rebuilds core from the mutated source, and `afterRestore` rebuilds it from the restored source so dist is not left carrying the last mutant. The manager mutations need no build: the child loads ../src/manager.js through tsx. The positive control inverts the in-flight guard so every ordinary timer tick returns before renewing; the first child-start cell must therefore redden, proving this suite executes the same manager.ts lease path that D3 mutates. D2a and D2b are graded on the numbers rather than on behaviour ON PURPOSE: each leaves the reconcile in place, so the manager still survives the stall and every behavioural cell still passes while the slack needed for several timely attempts is gone. Arithmetic is the only thing that catches that. The old fail-close and unconfirmed-TTL-budget mutations were removed with the behaviour they graded: lease loss no longer ends this process or consumes a local serving budget."
 }
--- a/implementations/manager/smoke/lease-renew-no-answer.smoke.ts
+++ b/implementations/manager/smoke/lease-renew-no-answer.smoke.ts
@@ -90,6 +90,8 @@ const check = (name: string, ok: boolean, detail?: unknown): void => {
 interface Relay {
   /** Start/stop holding that direction, for windows the caller times itself. */
   setStalled: (on: boolean) => void;
+  /** Every client-to-broker publish observed by the frame parser. */
+  publishes: () => ReadonlyArray<{ subject: string; atMs: number }>;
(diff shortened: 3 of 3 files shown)


davidfarah2003:
**APPROVE** — head `d1e86b2c99fc12b6c8e1889759a992aaf437c9dd`

Independent review from outside the author's model family. I probed the load-bearing claim rather than only reading it; the negative control is the spine of this verdict.

**§1 — CENTRAL PROBE (RAN, twice). The PR's account of itself is correct.**
I reverted only `lease-renew-no-answer.smoke.ts` to its `820261798` state, kept the four-cell fixture, committed, and ran the fixture. Result matched the prediction exactly:
- POSITIVE CONTROL → **KILLED** (8 marks, named `[acklost] ...renewed it at least once`). Reachability comes from the fixture cell, not the suite edit.
- D3 → **SURVIVED** (17 marks, baseline 17) — with the old `!/wrong last sequence/` assertion the overlap goes undetected. The gap is real and the suite edit is what closes it.

Positive complement (also RAN), new suite at the PR head:
- D3 → **KILLED** (16 marks, named `NO SELF-INFLICTED CAS CONFLICT`).

So the suite edit is precisely what converts D3 from a genuine SURVIVED into a KILLED, and reachability is supplied by the new positive-control cell. World-2-then-closed is verified, not asserted. (First §1 run was killed by an outer 600s harness cap mid-D2b, leaving `streams.ts` mutated; I restored it by hand — see the body's own outer-cap warning, now confirmed firsthand.)

**§0 — relabel check RE-VERIFIED.** All four `expectRed` strings are byte-identical between `820261798` and head; D3 still names `NO SELF-INFLICTED CAS CONFLICT`, D2a/D2b still name `THE RENEW BUDGET HAS SLACK`. Nothing was relabeled to whatever reddened. D3 anchor `    if (this.leaseRenewInFlight) return;\n` resolves exactly once on head AND on current `origin/main` (moved to line 2020, guard byte-identical); the control `find` matches only that line, not the field init at 974. `streams.ts` untouched by the PR. Diff is 3 files / 27 insertions / 5 deletions.

**§2 — new assertion can fail and fails for the right reason.** `renewAttemptsDuringStall.length === 1` reddens on an empty set (right shape, unlike the old absence test). `leaseSubject = $KV.${managerBucket(space)}.${managerLeaseKey(instanceId)}` is the correct JetStream KV write subject; the same two helpers are used to READ the lease successfully in the suite, and baseline is green at 17 (a wrong subject would empty the set and fail the unmutated suite). Counting the KV update publish at the relay instead of parsing an error string is justified: the manager renews via `renewManagerLease` (one KV update per tick), the guard bounds it to one in-flight, so exactly 1 publish during the window and 2 without the guard. The "3/3 before the guard" note is a small sample, but the assertion is a structural count, not a race on wall-clock ordering, so I do not read it as timing-fragile.

**§3 — widened relay is a pure pass-through when no drop filter is set.** With `dropMatch` undefined, `!dropMatch` short-circuits so every complete frame is written and every non-PUB line / unparseable-length line passes through, in order; a partial frame is buffered until complete (`return`). Concatenation of the per-frame writes equals the original byte stream in order — the only change is chunk boundaries, which are already insignificant over TCP. `publishes` is per-connection and bounded by the run's own frames (no unbounded growth across cells). Every non-drop cell now parses instead of raw-forwarding, but no cell's meaning changes and the added assertion is the only new grade. Confirmed empirically: with the new relay, baseline stayed green at 17 and every mutation graded as predicted.

**§4 — control's `expectRed` is the earliest red.** The first `check()` in the suite is the child-start cell `[...] renewed it at least once`; the inverted guard makes the manager never renew, so that first behavioural assertion is the earliest to redden. Observed at 8 marks in my run (author reported 7); nothing structurally precedes it.

**§5 — housekeeping.** Body closes exactly one issue (`Closes #1320`); no stray closing keyword. Changeset `.changeset/lease-renew-manager-control.md` is empty frontmatter + one description line — the house pattern for test-only machinery, correctly bumping no package. Outer-cap restore warning is present in the body and I hit it myself. I did not grade the 205-solo-cell census (out of scope); the body does not imply otherwise.

**CI (enumerated by name at the full 40-char sha).** `Mutation reproof` → success (the whole-fixture reproof of this fixture, in CI). `Windows` → success. `Code Quality` → success. `CI` still `in_progress` (smoke shards 0-3/4 running; unit/live/changed/CodeQL all success). No workflow is failed. `Docs` is legitimately absent: the changed paths (`.changeset/**`, `implementations/**`) match none of `docs.yml`'s `paths:` filter. `mergeStateStatus` = **UNSTABLE**, consistent with required checks still running, not with a failure.

**What I could not check:** the four-cell whole-fixture run to a single committed green in my worktree (each cell is a core build + suite; the outer harness caps me below the ~510s+ it needs here). I substituted the two-probe design above plus CI's own `Mutation reproof` success, which together cover the same claim more sharply. Worktree left clean at the head sha; both mutated sources (`manager.ts` guard, `streams.ts` `MANAGER_LEASE_ATTEMPT_MS`) confirmed restored; no processes left running.

PR #1328, opened 2026-09-06
fix(mutation): echo the run transcript under any non-kill verdict

davidfarah2003:
`Refs #1325`. **Narrowed: this is now the transcript echo only.** The launcher fix is #1326's, which is approved with a negative control and lands first. Earlier revisions of this branch also carried a fixture-level `command` change; that is dropped.

## What this changes

`mutation-proof.mjs` consumed each run's output for counting and filtering (`progressCount`, the `expectRed` filter, `completionMarker`) and **never printed it**. A `WRONG-RED` therefore reported that the expected string was absent and never what was there instead.

Non-KILL verdicts now echo the transcript, bounded to the last 20 lines, with a distinct line when the run produced nothing at all. A KILL prints none: its verdict already names the assertion that reddened, so a clean sweep is unchanged.

## Why it is worth landing on its own

This is what made #1325 cost a day. `manager-runtime-deps` sat red on `main`, on every PR selecting it, and on the 0.46.0 release PR, with all four cells at `0 marks (baseline 54)` and no way to tell from the log why. The cause turned out to be a pnpm frozen-lockfile abort, printed by the child and **already captured** by the runner (`output = stdout + stderr` at line 130) before being discarded.

Two people diagnosed it independently, from scratch, because the evidence was thrown away on every run. The next WRONG-RED costs whatever this one did unless the transcript is in the log.

## Controls, both directions

```
bogus expectRed  -> WRONG-RED, and beneath it:
                    | ✗ FAIL: published manager runtime dependency is installed: @nats-io/kv …
                    | SUITE COMPLETE: 53 passed, 1 failed
four real cells  -> KILLED, and nothing extra printed
```

## Scope

Deliberately does **not** fix the launcher defect behind #1325. I had a narrower fixture-level fix (invoke the suite binary directly, taking pnpm out of the path for that one fixture) and dropped it rather than keeping it as defence in depth: two mechanisms for one defect leaves the next reader unsure which is load-bearing, and it would have made `manager-runtime-deps` the only fixture of 342 invoking its suite differently from the rest — the kind of inconsistency someone tidies back, reintroducing the bug with no test to catch it. #1326 carries a self-test asserting the children receive the setting, which is the guard that makes the narrow fix redundant rather than complementary.

Rebuilt on `fc0589dcf`: one file, 26 insertions, zero deletions.

### DIFF
--- a/scripts/mutation-proof.mjs
+++ b/scripts/mutation-proof.mjs
@@ -265,6 +265,12 @@ function proveOne(m, opts) {
     return ok;
   };
 
+  // Declared OUTSIDE the try so the catch can read it. `const` inside the block made the catch's
+  // `transcript` a ReferenceError, so a harness throw after the run would have been replaced by
+  // "transcript is not defined" and the original error lost (rel-b, #1328 re-grade). Undefined
+  // until the run happens, which is what the reporter reads as "no run" — correct for a throw
+  // before the run, and the excerpt for a throw after it.
+  let transcript;
   try {
     writeFileSync(path, before.split(m.find).join(m.replace));
     // Assert the mutation APPLIED. A no-op mutation makes a green uninterpretable and leaves a red
@@ -281,7 +287,7 @@ function proveOne(m, opts) {
     // there instead. `manager-runtime-deps` sat red on main and on a release PR for a day in
     // exactly that state: four cells at `0 marks (baseline 54)`, reproducible on nobody's machine,
     // and the one artifact that would have named the cause discarded on every run.
-    const transcript = excerpt(r.output);
+    transcript = excerpt(r.output);
     const ticks = progressCount(r.output, opts.progressPattern);
 
     const restored = restore();
--- a/scripts/mutation-proof.mjs
+++ b/scripts/mutation-proof.mjs
@@ -53,6 +53,11 @@ import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 
 const C = { red: "\x1b[31m", green: "\x1b[32m", yellow: "\x1b[33m", dim: "\x1b[2m", off: "\x1b[0m" };
+/** label -> the mutated run's combined stdout+stderr, echoed under any non-KILL verdict. */
+const transcripts = new Map();
+/** Lines of transcript to echo. Enough to carry a stack or an early exit, short enough that a
+ *  fixture with several non-kills does not bury the summary. */
+const TRANSCRIPT_TAIL = 20;
 const say = (s = "") => process.stdout.write(`${s}\n`);
 const sha = (p) => createHash("sha256").update(readFileSync(p)).digest("hex");
 
@@ -251,6 +256,12 @@ function proveOne(m, opts) {
     say(`${C.dim}  mutated ${hits}× · running: ${m.command ?? opts.command}${C.off}`);
 
     const r = run(m.command ?? opts.command, cwd, opts.timeoutMs);
+    // Keep the transcript for the report. Every verdict below is derived from `r.output` and none
+    // of them printed it, so a WRONG-RED said the expected string was absent and never what was
+    // there instead. `manager-runtime-deps` sat red on main and on a release PR for a day in
+    // exactly that state: four cells at `0 marks (baseline 54)`, reproducible on nobody's machine,
+    // and the one artifact that would have named the cause discarded on every run.
+    transcripts.set(label, r.output);
     const ticks = progressCount(r.output, opts.progressPattern);
 
     const restored = restore();
@@ -529,6 +540,21 @@ for (const r of results) {
   // report would be to write mutants that do nothing.
   say(`${colour}${r.verdict.padEnd(12)}${C.off} ${r.label}`);
   say(`  ${C.dim}${r.why}${r.ticks !== undefined ? ` · ${r.ticks} marks (baseline ${r.baseTicks ?? "?"})` : ""}${C.off}`);
+  // A KILL needs no transcript: the verdict already names the assertion that reddened. Every other
+  // verdict is a question — what DID the run print, if not that — and the answer was captured and
+  // then thrown away. Echo it, so a red fixture is diagnosable from the log instead of requiring
+  // someone to reproduce a CI environment. Costs nothing on a clean run, which prints none.
+  if (!good) {
+    const t = (transcripts.get(r.label) ?? "").replace(/\s+$/, "");
+    if (t === "") {
+      say(`  ${C.dim}  (the run produced NO output at all — it failed before writing anything)${C.off}`);
+    } else {
+      const lines = t.split("\n");
+      const tail = lines.slice(-TRANSCRIPT_TAIL);
+      if (lines.length > TRANSCRIPT_TAIL) say(`  ${C.dim}  … ${lines.length - TRANSCRIPT_TAIL} earlier line(s) omitted${C.off}`);
+      for (const l of tail) say(`  ${C.dim}  | ${l}${C.off}`);
+    }
+  }
 }
 say("");
 if (bad === 0) {
--- a/scripts/mutation-proof.mjs
+++ b/scripts/mutation-proof.mjs
@@ -53,11 +53,31 @@ import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 
 const C = { red: "\x1b[31m", green: "\x1b[32m", yellow: "\x1b[33m", dim: "\x1b[2m", off: "\x1b[0m" };
-/** label -> the mutated run's combined stdout+stderr, echoed under any non-KILL verdict. */
-const transcripts = new Map();
-/** Lines of transcript to echo. Enough to carry a stack or an early exit, short enough that a
- *  fixture with several non-kills does not bury the summary. */
-const TRANSCRIPT_TAIL = 20;
+/** Lines of transcript to echo from each end. Enough to carry a stack or an early exit, short
+ *  enough that a fixture with several non-kills does not bury the summary. HEAD AND TAIL, not tail
+ *  alone: the cause of a red is as often the first line (a throw before any cell ran, a missing
+ *  module, `0 marks (baseline 54)` with its reason printed once at the top) as the last, and a
+ *  tail-only echo prints twenty lines of teardown and omits the one line the feature exists for.
+ *  Measured by the #1328 reviewer: an early cause followed by 25 teardown lines echoed
+ *  `… 6 earlier line(s) omitted` and the cause was among the six. */
+const EXCERPT_HEAD = 10;
+const EXCERPT_TAIL = 10;
+/**
+ * Bound a run's combined output to head + tail lines, at capture time. Carried ON THE RESULT, not
+ * in a map keyed by label: duplicate mutation names are accepted by the fixture format, and a
+ * label-keyed map let a later mutation overwrite an earlier one's transcript, so two WRONG-REDs
+ * echoed the same (second) run's output. A diagnostic that attributes run A's evidence to run B is
+ * worse than the discarded transcript it replaced. Bounding at capture also caps retention: the
+ * old map held every KILL's full output until reporting, 99 mutations × a 64 MiB capture ceiling.
+ */
+const excerpt = (output) => {
+  const t = (output ?? "").replace(/\s+$/, "");
+  if (t === "") return { head: [], tail: [], omitted: 0, empty: true };
+  const lines = t.split("\n");
+  if (lines.length <= EXCERPT_HEAD + EXCERPT_TAIL) return { head: lines, tail: [], omitted: 0, empty: false };
+  return { head: lines.slice(0, EXCERPT_HEAD), tail: lines.slice(-EXCERPT_TAIL),
+    omitted: lines.length - EXCERPT_HEAD - EXCERPT_TAIL, empty: false };
+};
 const say = (s = "") => process.stdout.write(`${s}\n`);
 const sha = (p) => createHash("sha256").update(readFileSync(p)).digest("hex");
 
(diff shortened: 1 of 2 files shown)


davidfarah2003:
**BLOCK** at `3410dce1bbb48ccc841faef336a09fcdee093c44`.

I confirmed both local `HEAD` and `gh pr view 1328 --json headRefOid` at that SHA. I read the diff and CI, ran `node scripts/mutation-proof.selftest.mjs` (33 checks pass), and ran focused scratch-fixture repros. I did not install `node_modules`.

Findings:

1. **The transcript Map can print the wrong mutation's evidence.** `label` is derived from optional free-form `name`, and duplicate names are neither rejected nor made unique. `transcripts.set(label, output)` overwrites the earlier run. I reproduced two mutations named `duplicate-label`: both final WRONG-RED entries printed only the second mutation's `SECOND_CAUSE`, so the first verdict was paired with false evidence. The 348 tracked configs currently contain no duplicate computed labels, but that is convention, not construction.

2. **Tail-only output can still hide the cause it exists to expose.** I reproduced a run that prints `CAUSE_AT_START`, then 25 teardown lines. The report honestly said `… 6 earlier line(s) omitted`, but the cause itself was absent. The omission marker prevents silent truncation, but the log remains non-diagnostic. Use a head-plus-tail excerpt, or otherwise preserve the start where launcher and early-exit errors usually appear.

The empty-output branch is reachable. A silent exit printed `(the run produced NO output at all — it failed before writing anything)` as intended.

Retention is also unnecessarily unbounded across the fixture. Every mutated run, including KILLs whose transcript is never printed, is held in full until reporting. Current fixtures are finite, with 1,785 mutations across 348 configs and a largest fixture of 99, but each capture permits up to 64 MiB, making the avoidable worst case multi-GB. Store a bounded excerpt on the result only when the verdict needs it. That also removes the label-key collision.

davidfarah2003:
APPROVE at `3410dce1bbb48ccc841faef336a09fcdee093c44`.

Graded by reading `scripts/mutation-proof.mjs`, the +26 diff, 128 mutation configs / 620 cells in this tree, issue #1325, and CI on this sha. Did not `pnpm install`. Did not run `mutation-proof`, `pnpm check`, `cotal up`, or any `-live` suite.

1. `TRANSCRIPT_TAIL = 20` is the tail. The motivating `manager-runtime-deps` abort is a short pnpm frozen-lockfile refusal before the suite starts, so the cause sits in that window. A suite that dies at the start and then prints a long teardown can still push the cause out. The `… N earlier line(s) omitted` line makes that honest rather than silent. Not a block. Named gap: the start of a long transcript is still not in the log.

2. `transcripts` is a Map keyed by `label` (`m.name`, else `file` + first 48 chars of `find`). Uniqueness is convention, not construction. No uniqueness check in the tool. This tree: 0 unnamed mutations, 0 duplicate names inside a fixture. A collision would make later `get(label)` return the last run's transcript for every same-label result. Named gap, not a block at current fixtures.

3. The empty-output branch does not fire on the motivating case. That case is WRONG-RED with 0 marks and a pnpm abort already in `r.output`. Empty-output fires when the stored transcript is blank (true empty child output, or an ERROR return before `transcripts.set`). The useful path for #1325 is the non-empty echo, which would have printed the abort. Distinct empty line is still correct for a silent child.

4. Full `r.output` is retained for every mutation, including KILLs that never print it, until process exit. `run()` already caps one spawn at `maxBuffer` 64MB. Largest fixture here is 20 cells. Bounded by fixture size in this repo, not a growth problem unless a future fixture is huge and chatty. Not a block.

Does what it claims. CI green on this sha. Self-test was not extended for the echo. I did not re-run the author's bogus-expectRed control.

davidfarah2003:
**BLOCK** at `4228a44e85828fe3f2870bffeef5e807b298cb04`.

The intended repairs work. I confirmed the exact remote head, read the complete two-file delta, ran the self-test from an exact-head archive, and independently exercised the repaired behaviors:

- 36/36 self-test checks pass.
- Duplicate labels retain their own evidence in order: `CAUSE-ALPHA`, then `CAUSE-BETA`.
- The first-line cause survives a 30-line teardown via the head-plus-tail excerpt.
- Excerpts are computed at capture time and carried per result, which removes label-key attribution and bounds retained transcript data.
- A target-not-found pre-run error reports `(no run:)` instead of claiming a silent child.

There is a new blocking regression in a pre-existing non-KILL path. The `completionMarker` INCONCLUSIVE return at `scripts/mutation-proof.mjs:420-430` does not include `transcript`, while every other post-run return now does. The final reporter therefore treats an executed run as a pre-run error and prints:

```
(no run: this verdict was reached before the suite was executed)
```

I reproduced this in a clean scratch fixture. The mutant prints `CAUSE-BEFORE-EARLY-STOP` and `✗ named cell`, then exits before the configured `SUITE COMPLETE` marker. At the old graded head, the INCONCLUSIVE report echoes both diagnostic lines. At this head, both are hidden and the report falsely says no run.

Required repair: carry `transcript` on that completion-marker return and add a self-test covering an executed, non-empty run that becomes INCONCLUSIVE because its completion marker is absent. This feature's claim is to echo every non-KILL run, so this path must not discard or misclassify its transcript.

I ran `node scripts/mutation-proof.selftest.mjs` at the exact head, an independent scratch-fixture reproduction, and `git diff --check`. I did not install `node_modules`.

davidfarah2003:
BLOCK at `4228a44e85828fe3f2870bffeef5e807b298cb04`.

Previous APPROVE was for `3410dce1bbb48ccc841faef336a09fcdee093c44` only.

Confirmed local HEAD, `gh pr view 1328`, and `git ls-remote origin refs/pull/1328/head` all this sha. Did not `pnpm install`. Ran `node scripts/mutation-proof.selftest.mjs` (36 checks pass). Scratch fixture only for the finding below. Did not run `pnpm check`, `cotal up`, or any `-live` suite.

The three claimed repairs hold:
- excerpt is computed at capture and carried on the result, so duplicate labels cannot swap evidence (self-test: CAUSE-ALPHA then CAUSE-BETA)
- head+tail keeps the first line through a 30-line teardown, with `middle line(s) omitted`
- pre-run ERROR prints `(no run: …)` rather than claiming an empty run
- retention is bounded at capture (no label-keyed full-output Map)

Regression vs `3410dce1b`: the `completionMarker` INCONCLUSIVE return omits `transcript`. The reporter then prints `(no run: this verdict was reached before the suite was executed)` about a suite that did run. Reproduced in a throwaway git repo: baseline green, mutant prints `CAUSE at start`, exit 1, named assertion present, marker absent. Verdict is INCONCLUSIVE with the honest why, then the dishonest no-run line. On `3410dce1b` that path still had the Map entry. The self-test's completionMarker cell grades the verdict only, not the echo.

Secondary: `const transcript` is try-scoped. The catch returns `{ label, transcript, … }` and that identifier is a ReferenceError (`transcript is not defined`, confirmed with a small node snippet). A harness throw after the run can replace the original error.

Fix: put `transcript` on every post-run return, including the `completionMarker` INCONCLUSIVE object, and declare it outside the try (or omit it from catch so a pre-run throw still says no run).

davidfarah2003:
APPROVE at `77ed91dcff37e54f91be82094064e2b273bfd1eb`.

Prior: APPROVE `3410dce1b`, BLOCK `4228a44e` (completionMarker INCONCLUSIVE omitted `transcript` and printed `(no run:` after a real suite). This head is that repair plus a self-test sibling.

Confirmed local HEAD and `git ls-remote origin refs/pull/1328/head` are this sha. Did not `pnpm install`. Ran `node scripts/mutation-proof.selftest.mjs` (37/37). Re-ran the scratch completionMarker fixture from the 4228a44e BLOCK: INCONCLUSIVE now echoes `| CAUSE at start` and the assertion line, not `(no run:`. Did not run `pnpm check`, `cotal up`, or any `-live` suite.

Claimed repairs hold: excerpt on the result object, head+tail, duplicate labels keep their own runs, pre-run ERROR says no run, completionMarker INCONCLUSIVE carries `transcript`.

Second finding, checked explicitly: `const transcript` is still try-scoped (`scripts/mutation-proof.mjs:284`). The catch at 437-439 still returns `{ label, transcript, … }`. A 10-line node snippet still returns the original error plus `ReferenceError: transcript is not defined`. The declaration was not hoisted. A harness throw after the run can still replace the real error. Named residual, not a block of this head: the 4228a44e BLOCK was the completionMarker no-run lie, and that path is fixed and pinned. Not manufacturing a third block to stay consistent with the second.

davidfarah2003:
**APPROVE** at `e8a126e6b86f39270f3d81fc5a04637cb38312f3`.

I confirmed the exact frozen remote head immediately before this verdict, read the complete delta, ran the exact-head self-test, and exercised the repaired boundaries through the public CLI. I did not install `node_modules`.

Verified:

- `node scripts/mutation-proof.selftest.mjs` passes 37/37 at the exact head.
- Duplicate-label runs retain their own transcript evidence.
- Head-plus-tail excerpts preserve an early cause through long teardown output.
- Completion-marker INCONCLUSIVE reports retain the executed run's transcript and do not claim `no run`.
- Pre-run errors still report `no run` correctly.
- Capture-time per-result excerpts bound retained output and remove label-key attribution.
- `git diff --check` passes.

The final scope hoist also works. I tested both sides with independent builtin-FS fault injection through the actual script:

1. A pre-run mutation write failure reaches `catch` before transcript capture. The report preserves `INJECTED-PRE-RUN`, emits the correct `(no run:)` line, and does not raise `ReferenceError`.
2. A post-run restore failure is constructible from the self-test harness by wrapping `copyFileSync`. It reaches `catch` after the child emitted `POST-RUN-CAUSE`. The report preserves `INJECTED-POST-RUN-RESTORE`, echoes `POST-RUN-CAUSE`, and emits neither false `no run` nor `transcript is not defined`. The fixture tree is restored and clean.

Residual, not a block: the disclosure is accurate that ordinary fixture data does not appear able to make the post-run branch throw, but the broader claim that this cannot be pinned by a self-test cell is false. The self-test can inject an FS fault, and that probe fails against the old scope. Adding that cell or narrowing the disclosure would improve regression coverage. The behavior at this head is correct.

davidfarah2003:
APPROVE at `e8a126e6b86f39270f3d81fc5a04637cb38312f3`.

Prior: APPROVE `3410dce1b`, BLOCK `4228a44e` (completionMarker INCONCLUSIVE printed `(no run:` after a real suite), APPROVE `77ed91dc` with the try-scoped `transcript` named as residual. This head is that hoist.

Confirmed local HEAD and `git ls-remote origin refs/pull/1328/head` are this sha. Did not `pnpm install`. Ran `node scripts/mutation-proof.selftest.mjs` (37/37). Did not run `pnpm check`, `cotal up`, or any `-live` suite.

Hoist checked in the file, not taken from the commit note: `let transcript` is declared above the `try` (`scripts/mutation-proof.mjs:273`) and assigned after `run()`. A node snippet: throw-before returns ERROR with `transcript` omitted (`undefined`); throw-after returns the original `why` plus the excerpt. No `ReferenceError`.

On the claim that the scoping fix is unpinnable in the self-test: a JSON fixture (mutation text only) still has no post-run throw in the grading branch (string ops and Map reads). A self-test cell *can* inject FS faults. Distinguishing the arms: a pre-run write failure can reach the `catch` and return, so the hoist is pinnable on that arm. A post-run restore failure does **not** return from the `catch`.

Independently reproduced through the public CLI in a scratch git repo: a suite that deletes `mutation-proof-*.bak` during the mutated run makes `restore()` throw ENOENT after `run()` returned. The `catch`'s first statement is `restore()` again, which throws again and escapes. The tool dies with a stack. `git status` showed `M src/impl.js`; contents were `export const n = 2`. A second construction, `"afterRestore": 1`, TypeErrors inside the first restore after the backup is `rmSync`'d; the catch restore then ENOENTs uncaught. In that construction the bytes were already copied back.

That restore-in-catch escape is real and it is the hazard this tool exists to prevent. It is older than this PR and not introduced by the 7-line hoist. Named residual for a follow-up, not a block of this echo PR. The two defects this PR claimed to fix are fixed at this sha.

davidfarah2003:
Direct acceptance supplement. Verdict remains **APPROVE** at `e8a126e6b86f39270f3d81fc5a04637cb38312f3`.

I reran the exact public tool self-test in an isolated archive of this head:

```text
node scripts/mutation-proof.selftest.mjs
MUTATION-PROOF SELF-TEST PASSED (37 checks)
```

The observed checks include the changed behavior:

- duplicate mutation labels echo their own run transcripts rather than the later run's evidence
- an early cause remains in the bounded head-and-tail excerpt after a long teardown
- a completion-marker INCONCLUSIVE echoes where the run stopped and does not claim `(no run:)`
- a pre-run ERROR still correctly reports that no suite ran
- restore preserves bytes and original mtime
- every self-test run restores the tree

I initialized and committed the isolated archive before validation. It remained clean afterward, providing an independent restore check. The exact PR patch has stable patch ID `5743f66aa5c48ec89e07e592f55f63ffc5823d25`, and `git diff --check` is clean.

No dependencies were installed. No stack, `cotal up`, `pnpm check`, or live suite was run.

davidfarah2003:
Whole-result feedback-loop supplement. Verdict remains **APPROVE** at `e8a126e6b86f39270f3d81fc5a04637cb38312f3`.

I reran the complete exact-head public self-test rather than checking only the final hoist:

```text
MUTATION-PROOF SELF-TEST PASSED (37 checks)
```

Across the entire PR result it directly reported:

- every non-kill transcript behavior remained correctly classified
- duplicate mutation names retained the evidence from their own runs
- bounded excerpts retained an early cause through a long teardown
- completion-marker INCONCLUSIVE retained the executed transcript and did not claim no run
- a pre-run ERROR correctly reported no run
- restored files retained both bytes and original mtime
- every run restored the isolated committed tree cleanly

The exact-head source still declares `transcript` outside the `try`, assigns it after `run()`, and carries it on post-run returns. This covers the earlier excerpt addition, completion-marker correction, and final scope hoist as one result.

The previously reported restore-in-catch escape remains a real pre-existing residual. It is outside this PR's claimed transcript-echo and scope fixes, and nothing in the whole-result rerun contradicts that boundary.

Observed conclusion: the two defects this PR claimed to fix remain fixed across the complete result, restoration passed on the normal self-test paths, and the known older restore failure remains correctly separated. APPROVE still follows.

PR #1339, opened 2026-09-06
fix(smoke): grade ambient env spreads per child

davidfarah2003:
## Summary
- The ambient-env census graded a whole suite file as stripped if it contained both `startsWith("COTAL_")` and the word `delete` anywhere, including comments. After #1316, `user-spawn.smoke.ts` left `FROZEN` because the CLI child filtered `COTAL_`, while the auth-service child still spread `...process.env`.
- The census now classifies each `...process.env` (comments ignored). A strip on a different child does not clear another spread. The auth-service child now copies, deletes `COTAL_`, then sets `COTAL_HOME`.
- Mutation pin N3 removes that auth-service strip and requires the census to name `user-spawn.smoke.ts`. N1 and N2 still kill.

## Test plan
- [x] `node bin/smoke/suite-ambient-env.smoke.ts` PASS at this head
- [x] Restored negative control: drop the auth-service strip, census names only `user-spawn.smoke.ts`, restore
- [x] `node scripts/mutation-proof.mjs --config bin/smoke/mutations/suite-ambient-env.mutations.json` KILLED 3/3
- [ ] CI `smoke:suite-ambient-env`

### DIFF
--- a/.changeset/ambient-env-per-spread.md
+++ b/.changeset/ambient-env-per-spread.md
@@ -1,0 +1,6 @@
+---
+"cotal-ai": patch
+"@cotal-ai/auth": patch
+---
+
+Grade each ambient `process.env` spread on its own, and strip COTAL_ from the user-spawn auth-service child.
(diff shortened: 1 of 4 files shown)


davidfarah2003:
Closing as a duplicate lane, not as a technical retract.

#1316 corrective ownership is already assigned (writer branch `fix/1316-ambient-env-child-scope` from current main). This PR was opened in a race.

Head `c0788e0fe` (`fix/1316-ambient-env-per-spread`) remains available as a reference for per-spread census grading and stripping `COTAL_` from the user-spawn auth-service child. No further commits here.

davidfarah2003:
Closed so the #1316 lane has a single writer head.

PR #1329, opened 2026-09-06
test(manager,connector-core): make the mutation reproof gate grade what it claims

davidfarah2003:
Fourteen mutants across five fixtures could not be graded at all: thirteen graded INCONCLUSIVE and quietly stopped grading, one graded UNGRADABLE. None of it was caused by the code under test: the same verdicts reproduce on a built checkout of `main`. A sixth fixture, `lease-renew`, had the same shape and landed separately in #1327, so it is not in this change.

Three separate defects, all in the tests.

## Completion markers carried a cell count

`connection-status`, `transport-liveness` and `transport-liveness-broker` each declared a marker of the form `SUITE COMPLETE: N cells`. Each suite prints `SUITE COMPLETE: <ran> cells` and separately refuses a run whose `ran` is not its `EXPECTED_CELLS`, so the count inside the marker duplicates a check the suite already makes, and it is the half that rots.

It rotted. A cell landed in the connection-status suite on 2026-08-30 and the fixture still asked for 15, so a finished 18-cell run printed a line the marker could never match and all seven mutants graded INCONCLUSIVE. The markers are now count-free and the suites' own completeness check stands alone.

## Three suites could end without printing a verdict

- `events-grant-acl` had no catch. Its later cells feed on earlier ones, so a rule that stops refusing lets a spawn through and the section that parses the refusal then mints on an empty profile and dies inside `mintCreds`, past the finally and before the summary.
- `user-spawn` printed its summary only on the success path, with the same cascade shape behind it.
- `boot-self-heal-gate` hung. Two cells expect a successor start to refuse and neither stopped one that started anyway, so a live manager held the event loop open and the process never exited.

Each now reports what it proved. A hang and a silent death both grade as no evidence, which is the opposite of what a cell that just went red is for.

## One guard was not graded by any suite that runs it

`boot-wake-race` mutated the durable re-announce in connector-core, from a time when its recovery ran through JetStream redelivery. The suite settles each scenario in a 500ms window chosen so no redelivery can have happened, and its own header says that guard moved to `smoke:cross-path-dedup`. The mutation moves there, split into two mutants so each is the other's positive control, and boot-wake-race keeps a mutation on the activation reconcile it does drive.

## Verification

Every affected fixture re-proved on a clean tree, plus the toolchain gates.

| Fixture | Before | After |
| --- | --- | --- |
| connection-status | 7 INCONCLUSIVE | 7 KILLED |
| transport-liveness | 18 KILLED | 18 KILLED |
[...]

Issue #406, opened 2026-08-14
mutation-proof restores source but leaves build artifacts holding the mutant


davidfarah2003:
## Summary

`scripts/mutation-proof.mjs` restores the mutated **source** file from its backup and verifies that restore, but it does not rebuild whatever the mutation command built. For a suite that imports a **built package**, the compiled artifact is left holding the last mutant after the run finishes.

Observed: after a batch where every mutation was reported `KILLED`, the tracked source was correct and `git status` was clean, while `packages/core/dist/endpoint.js` still contained the mutated expression. Rebuilding the dependency closure restored it.

## Why it matters

`git status` reads clean, so the usual signal that something is left over does not fire — `dist/` is gitignored. The next thing to run in that tree loads the mutant. Depending on what runs next this is either a confusing red or, worse, a **green obtained against mutated code**: if the following suite happens not to touch the mutated path, it passes while an artifact in the tree is deliberately broken.

This does not weaken any verdict the tool already reported — the mutations were genuinely killed, and the shipped source is correct. The hazard is entirely in what happens *after* the tool exits.

## Suggested direction

Either:

- accept an optional post-restore command (e.g. `--after "pnpm --filter <pkg>... build"`) that the tool runs after restoring, and report its exit status alongside the restore verification; or
- detect that the mutation command builds (or simply that the mutated file's package has a `dist/`) and refuse to report success without a rebuild, the way it already refuses an unverified restore.

The tool's existing posture is that a restore which silently fails is a defect it must not let pass. A restore that leaves a mutant in the artifact the suite actually loads is the same defect one layer down.

Related: a suite that imports built packages must also rebuild the whole dependency closure *before* running, or a mutation in a transitive package never reaches the artifact and reports `SURVIVED` for the wrong reason.

davidfarah2003:
**Phase-1 triage - not-reproduced**

Tested at `ccd99461a9e79806e8eac201c1861941b558c4f0`.

Command:
```bash
REPO=/Users/david/cotal-mac-fleet/repo
SHA=ccd99461a9e79806e8eac201c1861941b558c4f0
PRE=74e9ebb0bc5335e87545b674965dae85b615e77d
for REV in "$PRE" "$SHA"; do
  TOOL="$(git -C "$REPO" show "$REV:scripts/mutation-proof.mjs")"
  printf 'REV=%s\n' "$REV"
[...]

PR #1340, opened 2026-09-06
fix(smoke): grade ambient env spreads per child

davidfarah2003:
## Summary

- classify each ambient `process.env` spread independently so one child's strip cannot clear another child
- bind copied-environment stripping to the exact copied variable and ignore comment prose when checking the delete mechanism
- require direct `process.env` scrubs to share structural execution scope and lexical code or child-script region with the spread they protect
- parse string and template contents recursively, preserve interpolation as containing code, and keep template raw text in a separate child-script region
- reject function-local, conditional, unused-template, cross-template, regex-brace, and nested-interpolation false ownership while preserving valid module and same-template scrubs
- strip inherited `COTAL_*` values from the user-spawn auth-service child, preserve non-Cotal environment, and restore its required sandbox `COTAL_HOME`
- guard the real census population with measured module-scope, multi-spread, and mixed-construction sentinels

## Validation

- reproduced the current-main false green, including the semantics-equivalent CLI filter that exposed `implementations/auth/smoke/user-spawn.smoke.ts` as the sole offender
- reproduced function-local, conditional, first-spread-only, unused-template, regex-brace, raw-template/interpolation, and recursive nested-template false greens on prior heads
- verified the existing unconditional module-scope scrub population remains 11 files and 14 spreads
- final census: 105 files, 60 stripped, 4 reviewed-safe, 41 frozen
- `node bin/smoke/suite-ambient-env.smoke.ts`
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/suite-ambient-env.mutations.json`
- all nine mutations were killed; named failures cover the user-spawn spread, truncated enumeration, structural depth, lexical ownership, regex braces, template raw/interpolation separation, and recursive interpolation parsing

### DIFF
--- a/.changeset/ambient-env-per-spread.md
+++ b/.changeset/ambient-env-per-spread.md
@@ -1,0 +1,6 @@
+---
+"cotal-ai": patch
+"@cotal-ai/auth": patch
+---
+
+Grade each ambient `process.env` spread on its own, and strip `COTAL_` from the user-spawn auth-service child.
--- a/bin/smoke/mutations/suite-ambient-env.mutations.json
+++ b/bin/smoke/mutations/suite-ambient-env.mutations.json
@@ -104,7 +104,7 @@
     {
       "name": "N7 regex literal braces regain structural meaning",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "      // REGEX-LITERAL REGION: braces here are data, not structural nesting.\n      if (node.kind === ts.SyntaxKind.RegularExpressionLiteral) {\n        mark(offset + node.getStart(file), offset + node.end, nextRegion++);\n        return;\n      }\n",
+      "find": "      if (node.kind === ts.SyntaxKind.RegularExpressionLiteral) {\n        mark(offset + node.getStart(file), offset + node.end, nextRegion++);\n        return;\n      }\n",
       "replace": "",
       "expectRed": "a regex close brace must not hide nested scrub depth",
       "note": "Proves regex-literal tokenization excludes brace data from structural depth while the separate division control remains green at baseline."
@@ -120,15 +120,15 @@
     {
       "name": "N9 interpolation expressions inherit the template raw-text region",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "            baseRegion,\n          ); // interpolation executes in the containing code region\n",
-      "replace": "            childRegion,\n          ); // interpolation executes in the containing code region\n",
+      "find": "            baseRegion,\n          );\n",
+      "replace": "            childRegion,\n          );\n",
       "expectRed": "a module scrub must clear a spread inside template interpolation code",
       "note": "Proves interpolation expressions inherit the containing code owner instead of the template raw-text owner."
     },
     {
       "name": "N10 interpolation contents stop receiving recursive lexical regions",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "          parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            offset + expressionStart,\n            baseRegion,\n          ); // interpolation executes in the containing code region\n",
+      "find": "          parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            offset + expressionStart,\n            baseRegion,\n          );\n",
       "replace": "",
       "expectRed": "an unused nested template scrub must not clear a module-code spread",
       "note": "Proves nested literals inside interpolation code are recursively discovered and assigned their own child-script regions."
--- a/bin/smoke/mutations/suite-ambient-env.mutations.json
+++ b/bin/smoke/mutations/suite-ambient-env.mutations.json
@@ -52,7 +52,10 @@
     "an inner template scrub can be mistaken for containing module code.",
     "",
     "N11 PROVES EVERY SIBLING INTERPOLATION IS PARSED. Limiting recursion to the final span must red",
-    "before a nested template scrub in one interpolation can clear a spread in its sibling."
+    "before a nested template scrub in one interpolation can clear a spread in its sibling.",
+    "",
+    "N12 PROVES COOKED CHILD-SCRIPT MATCHES KEEP SOURCE COORDINATES. Replacing the mapping with",
+    "linear cooked offsets must red where an earlier escape would move a top-level scrub inside a brace."
   ],
   "mutations": [
     {
@@ -91,23 +94,23 @@
     {
       "name": "N5 process.env ownership regresses from structural depth to column zero",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "  return [...code.slice(0, index).matchAll(PROCESS_ENV_STRIP)].some((m) =>\n    regions[m.index ?? -1] === region && braceDepthAt(code, regions, m.index ?? -1, region) === 0,\n  );\n",
-      "replace": "  return [...code.slice(0, index).matchAll(PROCESS_ENV_STRIP)].some((m) =>\n    regions[m.index ?? -1] === region && code.lastIndexOf(\"\\n\", m.index ?? -1) + 1 === (m.index ?? -1),\n  );\n",
+      "find": "  return lexical.processEnvStrips.some(\n    (strip) =>\n      strip.end <= index &&\n      strip.region === region &&\n      lexical.regions[strip.start] === region &&\n      lexical.regions[strip.deleteStart] === region &&\n      braceDepthAt(code, lexical.regions, strip.start, region) === 0,\n  );\n",
+      "replace": "  return lexical.processEnvStrips.some(\n    (strip) =>\n      strip.end <= index &&\n      strip.region === region &&\n      lexical.regions[strip.start] === region &&\n      lexical.regions[strip.deleteStart] === region &&\n      code.lastIndexOf(\"\\n\", strip.start) + 1 === strip.start,\n  );\n",
       "expectRed": "a regex close brace must not hide nested scrub depth",
       "note": "Proves structural brace depth remains part of ownership after regex-region tokenization: replacing it with a column-zero test falsely accepts the nested close-brace control."
     },
     {
       "name": "N6 process.env ownership loses its lexical-region boundary",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "    regions[m.index ?? -1] === region && braceDepthAt(code, regions, m.index ?? -1, region) === 0,\n",
-      "replace": "    braceDepthAt(code, regions, m.index ?? -1, regions[m.index ?? -1]) === 0,\n",
+      "find": "      strip.region === region &&\n      lexical.regions[strip.start] === region &&\n      lexical.regions[strip.deleteStart] === region &&\n",
+      "replace": "      true &&\n",
       "expectRed": "an unused nested template scrub must not clear a module-code spread",
       "note": "Proves scrub ownership cannot cross from a nested or top-level string/template region into module code or another child-script literal."
     },
     {
       "name": "N7 regex literal braces regain structural meaning",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "      if (node.kind === ts.SyntaxKind.RegularExpressionLiteral) {\n        mark(offset + node.getStart(file), offset + node.end, nextRegion++);\n        return;\n      }\n",
+      "find": "      if (node.kind === ts.SyntaxKind.RegularExpressionLiteral) {\n        mark(sourceOffsets[node.getStart(file)], sourceOffsets[node.end], nextRegion++);\n        return;\n      }\n",
       "replace": "",
       "expectRed": "a regex close brace must not hide nested scrub depth",
       "note": "Proves regex-literal tokenization excludes brace data from structural depth while the separate division control remains green at baseline."
@@ -115,8 +118,8 @@
     {
       "name": "N8 template raw text loses its separate lexical region",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "        const childRegion = nextRegion++;\n        const headStart = node.head.getStart(file);\n        mark(offset + headStart, offset + node.head.end, childRegion);\n",
-      "replace": "        const childRegion = baseRegion;\n        const headStart = node.head.getStart(file);\n        mark(offset + headStart, offset + node.head.end, childRegion);\n",
+      "find": "        const childRegion = nextRegion++;\n        const headStart = node.head.getStart(file);\n        mark(sourceOffsets[headStart], sourceOffsets[node.head.end], childRegion);\n",
+      "replace": "        const childRegion = baseRegion;\n        const headStart = node.head.getStart(file);\n        mark(sourceOffsets[headStart], sourceOffsets[node.head.end], childRegion);\n",
       "expectRed": "a raw-template scrub must not clear a spread inside interpolation code",
       "note": "Proves template raw text has a distinct child-script owner while interpolation code keeps the containing region; flattening the two falsely lets raw scrub text license interpolation."
     },
@@ -131,7 +134,7 @@
     {
       "name": "N10 interpolation contents stop receiving recursive lexical regions",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "          parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            offset + expressionStart,\n            baseRegion,\n          );\n",
+      "find": "          parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            sourceOffsets.slice(expressionStart, span.expression.end + 1),\n            baseRegion,\n          );\n",
       "replace": "",
       "expectRed": "an unused nested template scrub must not clear a module-code spread",
       "note": "Proves nested literals inside interpolation code are recursively discovered and assigned their own child-script regions."
@@ -139,10 +142,18 @@
     {
       "name": "N11 every sibling interpolation receives recursive lexical regions",
       "file": "bin/smoke/suite-ambient-env.smoke.ts",
-      "find": "          parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            offset + expressionStart,\n            baseRegion,\n          );\n",
-      "replace": "          if (span === node.templateSpans[node.templateSpans.length - 1]) parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            offset + expressionStart,\n            baseRegion,\n          );\n",
+      "find": "          parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            sourceOffsets.slice(expressionStart, span.expression.end + 1),\n            baseRegion,\n          );\n",
+      "replace": "          if (span === node.templateSpans[node.templateSpans.length - 1]) parseRegion(\n            text.slice(expressionStart, span.expression.end),\n            sourceOffsets.slice(expressionStart, span.expression.end + 1),\n            baseRegion,\n          );\n",
       "expectRed": "a nested template scrub must not clear a sibling interpolation spread",
       "note": "Proves recursive parsing covers every interpolation span; parsing only the final span falsely lets a nested scrub clear a spread in an earlier sibling."
+    },
+    {
+      "name": "N12 child-script source mapping regresses to linear cooked offsets",
+      "file": "bin/smoke/suite-ambient-env.smoke.ts",
+      "find": "          const cookedSourceOffsets = sourceOffsetsForCookedText(rawText, node.text, rawSourceOffsets);\n",
+      "replace": "          const cookedSourceOffsets = rawSourceOffsets.slice(0, node.text.length + 1);\n",
+      "expectRed": "an escaped child-script scrub must retain its source brace depth",
+      "note": "Proves cooked child-script matches retain original source coordinates: a linear cooked offset lands before the closing brace and falsely treats this top-level scrub as nested."
     }
   ]
 }
--- a/bin/smoke/mutations/suite-ambient-env.mutations.json
+++ b/bin/smoke/mutations/suite-ambient-env.mutations.json
@@ -24,7 +24,33 @@
     "",
     "N3 IS THE MIXED-FILE CASE. A file can strip one child and still spread ambient identity into another.",
     "Per-file STRIP+DELETE treated that as clean, including when `delete` only appeared in comments.",
-    "Removing the auth-service strip while the CLI child still filters COTAL_ must redden user-spawn."
+    "Removing the auth-service strip while the CLI child still filters COTAL_ must redden user-spawn.",
+    "",
+    "N4 PROVES THE REAL POPULATION IS ENUMERATED. Embedded examples can pass while the repository scan",
+    "quietly grades only the first spread in every file. Truncating spreadIndexes must red at the real",
+    "spread-count sentinel because real multi-spread files contain more spreads than classified files.",
+    "",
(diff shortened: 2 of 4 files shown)


davidfarah2003:
## Review (exact head `0855cb7983657f6dc3ed7dda4e2a4ee95173af67`)

**APPROVE**

The census now grades each `...process.env` independently. A strip on one child, or the word `delete` in a comment, does not clear another spread in the same file. The mixed-file fixture reds; the all-stripped fixture passes. The user-spawn auth-service child copies env, deletes inherited `COTAL_*` on that copy, then restores `COTAL_HOME`, without mutating `process.env` in the scrub.

Mutation N3 is a single exact anchor. Reproof CI killed N1/N2/N3 (3/3), and N3 reddened on the named user-spawn unstripped-spread assertion. Local `node --experimental-strip-types bin/smoke/suite-ambient-env.smoke.ts` passed at this head (60 strip, 4 reviewed, 41 frozen). Mutation-proof was not re-run in the review worktree.

Residual, not blocking: the strip matcher is the copy-then-`Object.keys`/`delete` loop with a semicolon, so an equivalent filter or a later strip is a conservative fail. Copies that are not `...process.env` (for example `Object.assign`) remain outside this census.

davidfarah2003:
## Retrospective verdict: BLOCK

Exact head: `0855cb7983657f6dc3ed7dda4e2a4ee95173af67`.

I re-confirmed the live PR head immediately before this verdict. Its parent is `5e635a22c66c1efbe5921c1ebd112d6dd351b386`, so the later #1329 user-spawn readiness marker and failure-summary changes are present and not lost. The PR has the stated four-file behavior patch plus its changeset, and `git diff --check` passes.

### Blocking finding: a prior process-env scrub still clears unrelated later children

The copy-specific arm is scoped to the copied variable, but `PROCESS_ENV_STRIP` remains a file-prefix heuristic:

```ts
if (PROCESS_ENV_STRIP.test(code.slice(0, index))) return true;
```

Any earlier text matching a loop that deletes `COTAL_*` from `process.env` therefore clears every later spread, regardless of whether that scrub executes on the path to the child.

I reproduced this through the shipped census in an exact-head scratch archive:

```ts
function neverCalled() {
  for (const key of Object.keys(process.env))
    if (key.startsWith("COTAL_")) delete process.env[key];
}
const unsafe = { ...process.env };
spawn(process.execPath, ["-e", "process.exit(0)"], { env: unsafe });
```

The census reported the file as `strips COTAL_ before the spread` and exited 0. Removing only the never-called scrub made the identical unsafe child the sole offender and exited 1. An unreachable false branch has the same classification. This is still a nearby-spelling/file-prefix variant of the original false green, not a scrub bound to that child.

### Blocking gap: no real spread-population control

The embedded mixed/all-stripped strings prove one fixed pair, but they do not assert that the real scan enumerates every spread or that both relevant real populations are non-empty.

I mutated `spreadIndexes()` in a scratch archive to inspect only the first spread in every file. The complete suite still passed. A scanner that silently ignores every second and later spread can therefore satisfy the current controls. This is the missing positive population control requested for the review.

### Requirements that passed

- Exact-head suite passes.
- A mixed child file fails and names the unstripped spread.
- Comment prose containing `delete` is ignored.
- The same control passes when both children strip their own copies.
- The auth-service child copies `process.env`, deletes inherited `COTAL_*` only from that copy, restores `COTAL_HOME`, and does not mutate `process.env`.
- N3 anchor count is 1 and its replacement is absent at the head.
- Exact-head hosted mutation reproof selected two fixtures with a non-zero count. The census baseline had 60 marks and N1/N2/N3 all killed. N3 reddened first on `implementations/auth/smoke/user-spawn.smoke.ts (unstripped spread` at 59 marks.
- #1329 content is preserved in the base and head.

### Required repair

Bind a `process.env` scrub to the same executed flow that feeds each spread, or conservatively reject ambiguous function/branch-separated scrubs. Add a real scan population assertion and a mutation that truncates spread enumeration and must red.

I did not read `#review.1316`. I did not install, build, run a stack or live suite, invoke the dev-tree CLI, run `cotal up/down`, or run `pnpm check`. I did not duplicate mutation-proof locally because the fresh capacity check showed load `40.10` and about `327 MiB` free; the complete exact-head hosted proof was available and read directly.

davidfarah2003:
## Disposition of prior APPROVE on `0855cb7983657f6dc3ed7dda4e2a4ee95173af67`

**WITHDRAWN.** Comment 5557489206 is no longer an APPROVE. That head is not mergeable on this record.

I reproduced two classifier holes first-hand on that same SHA:

- A: an uncalled `process.env` COTAL scrub makes a later `{ ...process.env }` look stripped (unstripped lines `[]` vs `[2]` after removing only that unused function). The prefix check is `PROCESS_ENV_STRIP.test(code.slice(0, index))`.
- B: truncating `spreadIndexes` to the first spread still PASSes (census 105 / 60 strip / 4 reviewed-safe / 41 frozen on a local checkout of this SHA).

Those findings contradict an unqualified APPROVE. Green Linux CI run 34016193403 on this SHA does not restore it.

**Still holds and should be carried forward, not re-derived:**

- four-file scope
- comments cannot satisfy `delete`
- mixed file fails and all-stripped passes
- auth-child copies `process.env`, deletes inherited `COTAL_*` on the copy without mutating `process.env`, restores `COTAL_HOME`
- N3 anchor unique, replacement absent at this head
- hosted mutation `changed` 3/3 KILLED with N3 named

This is not a grade of `0594a7c` or of live head `0c0a90c`. No new binding until an explicit SHA pin.

davidfarah2003:
## Recovery verdict: BLOCK

Exact recovery head: `0c0a90cfdfea7e52d0a0b8e1dbf73b32bdc3e233`.

I re-confirmed this live head immediately before the verdict. It is based on current main and its direct parent is the previously blocked `0855cb7983657f6dc3ed7dda4e2a4ee95173af67`, not the orphaned `0594a7c`. The four-file behavior patch plus changeset is clean under `git diff --check`.

### The two prior findings are repaired

First-hand controls at this head show:

- an uncalled function scrub at structural depth 1 and column 0 does not clear a later spread
- the indented function form also fails
- conditional scrubs fail in both column-0 and indented forms
- an unconditional module-scope scrub passes
- truncating `spreadIndexes()` to one spread per file reds on `spread enumeration was truncated or the multi-spread population disappeared`
- replacing structural depth with the column-zero test reds first on `an uncalled function scrub must not clear a later unsafe spread`
- the baseline suite passes and asserts a non-empty real multi-spread and mixed-path population
- the measured module-scope population remains 11 files and 14 spreads

The auth-service child still copies `process.env`, scrubs inherited `COTAL_*` on the copy, then restores `COTAL_HOME`. N1-N5 anchors are unique. The #1329 user-spawn readiness marker and failure summary remain present. The patch changeset correctly covers `cotal-ai` and `@cotal-ai/auth`.

### New blocking finding: string/template ownership is not tracked

An unused module-scope template string containing a scrub falsely clears a later unsafe spread outside that template.

I planted this exact-head smoke source:

```ts
const UNUSED = `for (const key of Object.keys(process.env))
  if (key.startsWith("COTAL_")) delete process.env[key];`;
const unsafe = { ...process.env };
spawn(process.execPath, ["-e", ""], { env: unsafe });
```

The shipped census classified the file as `strips COTAL_ before the spread` and exited 0. Removing only the unused template declaration made the identical unsafe child the sole offender and exited 1. A scrub and spread inside the same child-script template passes, so literals cannot simply be discarded.

The mechanism is lexical-region loss. `codeWithoutComments()` retains literal contents for regex matching. `processEnvStrippedBefore()` scans the whole prefix. `braceDepthAt()` ignores literal braces, but does not report whether the scrub match belongs to a different string/template region from the spread. The unused template scrub is therefore assigned depth 0 and licenses module code outside that template.

### Required repair

Track lexical-region ownership. A scrub found inside a string/template may protect only spreads in the same literal child-script region. A module-code scrub may protect later module-code spreads. Add both directions as regression cells and a mutation that removes the region check.

The hosted N1-N5 proof was still running at verdict time. Even 5/5 KILLED would not cover this unmutated adversarial boundary.

I did not read `#review.1316` before the verdict. I did not install, build, run mutation-proof locally, start a live or stack suite, invoke the dev-tree CLI, run `cotal up/down`, or run `pnpm check`.

davidfarah2003:
## BLOCK

Exact head: `d8872fbc9ef6ac5e98c117603bc919f55ab43e97`.

Independent first-hand review from an isolated archive. Shared checkout was not switched. Shipped `node --experimental-strip-types bin/smoke/suite-ambient-env.smoke.ts` PASS (105 files: 60 strip, 4 reviewed-safe, 41 frozen). Unused-template and regex `/}/` nested / `/{/` module cells pass. N5/N6/N7 copies first-red their named cells. The prior unused-template false green is closed on this head.

New blocking finding: nested template literals are not stacked. `codeWithoutComments` and `lexicalRegions` toggle on each backtick with no interpolation/nesting stack, so an inner unused template delimiter closes the outer template and the inner unused `PROCESS_ENV_STRIP` is assigned module region 0.

Exact-head plant `bin/smoke/_relb-nested-template.smoke.ts` containing only an unused nested-template scrub plus a later module `{ ...process.env }`: the shipped census printed `strips COTAL_ before the spread` and exited 0. Removing only the inner unused scrub made the identical later spread the sole offender (exit 1).

Required repair: nest template/interpolation regions so an inner literal cannot become module code. Add that cell and a mutation that drops the stack.

Hosted changed job was still pending at review time and cannot cure this unmutated boundary. Did not merge.

davidfarah2003:
## Exact-head verdict: BLOCK

Exact live head: `d8872fbc9ef6ac5e98c117603bc919f55ab43e97`.

I re-confirmed this remote head immediately before the verdict. It is a sibling replacement built directly on `0855cb7983657f6dc3ed7dda4e2a4ee95173af67`; no verdict or CI evidence transfers from `0594a7c`, `0c0a90c`, `08f37ea`, or `7dc9046`.

The clean exact-head baseline census passes and reports 105 files: 60 stripped, 4 reviewed-safe, and 41 frozen and unaudited. Its real-population sentinels, including 11 module-scrub files and 14 spreads, pass. The auth-service child copies `process.env`, strips `COTAL_*` from that copy, and restores the sandbox `COTAL_HOME`. The changeset covers `cotal-ai` and `@cotal-ai/auth` as patches. The preserved user-spawn signal-readiness marker and status walk remain present.

### Blocking lexical-grammar failures

`lexicalRegions()` is a hand-written quote and slash scanner, not a syntax parser. It assigns an entire template, including `${...}`, to one literal region and guesses regex versus division from one previous character.

I added independent controls in an agent-owned scratch copy of this exact head.

#### Regex after `return` false-clears a later unsafe spread

```ts
function neverCalled() {
  return /}/;
  for (const key of Object.keys(process.env))
    if (key.startsWith("COTAL_")) delete process.env[key];
}
const unsafe = { ...process.env };
spawn("unsafe", { env: unsafe });
```

Expected offender: the module spread. Actual result: no offender. After `return`, the previous non-space character is `n`, so the scanner treats `/}/` as division. Its close brace collapses the function depth, and the never-executed scrub is misclassified as module scope.

#### Template interpolation gets the wrong ownership region in both directions

A module scrub followed by a spread inside a `${...}` expression is falsely red:

```ts
for (const key of Object.keys(process.env))
  if (key.startsWith("COTAL_")) delete process.env[key];
const snapshot = `value ${JSON.stringify({ ...process.env })}`;
```

Expected: safe module expression. Actual: offender at line 2 because the spread is assigned to the template's literal region rather than module code.

The reverse direction falsely clears:

```ts
const mixed = `for (const key of Object.keys(process.env))
  if (key.startsWith("COTAL_")) delete process.env[key];
  ${JSON.stringify({ ...process.env })}`;
```

Expected: the raw child-script scrub text must not protect the module expression inside `${...}`. Actual: no offender because both are assigned to the same template region.

These are not hypothetical parser concerns. All three controls failed live against the exact head.

### Required repair

Use syntax-aware lexical ownership that distinguishes regex literals from division in real grammar contexts and assigns template raw text separately from each `${...}` expression. Add regression cells for:

- a regex literal containing `}` after `return` inside a nested function
- a module scrub protecting a spread inside template interpolation
- raw template scrub text not protecting a spread inside interpolation

Add mutations that remove each corresponding ownership rule.

At verdict time the exact-head hosted mutation and CI runs were queued or in progress. Even a green N1-N7 run cannot cover these unregistered controls. I did not install, build, run mutation-proof locally, invoke a live or stack suite, use the dev-tree CLI, run `cotal up/down`, or run `pnpm check`.

davidfarah2003:
**BLOCK** still binds exact live head `d8872fbc9ef6ac5e98c117603bc919f55ab43e97`.

Independently confirmed Mutation reproof `34020648531`, CI `34020648553`, and Windows `34020648588` all concluded **cancelled** at that SHA. Cancelled hosted runs are not a successful reproof and do not lift this BLOCK.

The independent nested unused inner-template `PROCESS_ENV_STRIP` plant still stands. Do not merge #1342.

davidfarah2003:
## Exact-head verdict: BLOCK

Exact head: `6c9d1e628499fc799651ebfbb7fdde63b196c566`.

I independently re-confirmed the live PR head and the supplied freeze:

- tree `5a2fd0849a780f37ecda159d2e0ec0a3a2442074`
- suite blob `6280a9900c7b6d04bddeb573bcf828271e0ca6ca`
- mutation config blob `895346923f043c464528f01760c7083a6939861d`

The exact fixture parses under `node --check`. With the locked TypeScript 5.9.3 dependency, the baseline census passes and reports 105 files: 60 stripped, 4 reviewed-safe, and 41 frozen and unaudited. `implementations/auth/smoke/user-spawn.smoke.ts` is classified stripped. The auth-service child copies `process.env`, deletes inherited `COTAL_*`, preserves other variables, and restores `COTAL_HOME`. The real 11-file/14-spread population sentinels and readiness marker remain present. The patch is clean, the four changed files appear in package dry-run, and the changeset covers `cotal-ai` plus `@cotal-ai/auth` as patches.

### Blocking finding: escaped child-script source is falsely red

The recursive region parser calls `parseRegion(node.text, ...)`. TypeScript's `node.text` is the cooked literal value, but the mapping offset still points into the original escaped source. When a child-script template uses a JavaScript escape before its scrub, the cooked text is shorter than the original spelling. Nested region marks shift, and the valid scrub no longer owns the later spread.

This parse-valid exact-head control fails:

```ts
const child = `\u0066or (const key of Object.keys(process.env))
  if (key.startsWith("COTAL_")) delete process.env[key];
  const safe = { ...process.env };`;
```

JavaScript evaluates `\u0066or` as `for`. I separately executed both the Unicode form above and the `\x66or` form through `new Function`; both deleted a seeded `COTAL_*` variable. The census nevertheless returns the spread as unstripped. No existing cell or mutation contains either escape shape.

Required repair: preserve original-source offsets while recursively parsing cooked child-script contents, or maintain an explicit cooked-to-source offset map. Add Unicode- and hex-escape regression cells and a mutation that removes the offset correction.

### Sibling interpolation probe: SEPARABLE

I independently found a separator, by a different formulation from the reported last-span predicate. The single `lexicalRegions` edit skips recursive ownership for the first interpolation only when sibling spans follow:

```ts
for (const [spanIndex, span] of node.templateSpans.entries()) {
  // ...
  if (spanIndex > 0 || node.templateSpans.length === 1)
    parseRegion(/* unchanged arguments */);
}
```

The fixture parses. The line-327 cell changes from `[1]` to `[]`. After changing only that assertion's expected value to observe beyond fail-fast, the whole remaining suite passes. In particular lines 320, 337, 345, and 352 are unchanged. Therefore the cell is separable, not dominated, and needs its own N11 mutation. This is a different separator from “only the last span receives `baseRegion`,” but it exposes the same uncovered first-sibling recursion boundary.

### Mutation bound

The frozen config contains ten mutations but only eight distinct named cells:

- N5 and N7 both name line 267, the regex-close-brace cell.
- N6 and N10 both name line 320, the unused-nested-template/module-spread cell.
- The line-327 sibling-interpolation cell is named by none.

The exact-head hosted `changed` job selected 8 fixtures from 351. This PR's own fixture baselined green and N1 through N10 each printed `KILLED` on their named reds. The aggregate was 6 discriminated fixtures, 2 pre-red, and 0 inconclusive. The two pre-red fixtures were the user-spawn readiness and whole-row regrant fixtures; the same hosted sequence ran the user-spawn suite successfully under a full build at 09:47:04. This evidence proves the registered ten mutations, not line 327 or the escaped-source boundary above.

All 18 check-runs are terminal with success or the expected scheduled-only `full` skip. I also ran additional parse-valid controls outside the registered set for keyword/async/conditional regex contexts, module/raw interpolation ownership, three sibling interpolations, and escaped raw braces; those passed. They do not cover the escaped executable child-script spelling that fails above.

I did not install, build, run mutation-proof locally, invoke a live or stack suite, use the dev-tree CLI, run `cotal up/down`, or run `pnpm check`.

davidfarah2003:
## Exact-head verdict: APPROVE

Exact head: `821e72e1a6bea66acf3ffc888ae5d0cb3625d59d`.

I independently re-confirmed the live PR head, parent `1b2680d72460f00d0d8a23ebe87bccba20341949`, and tree `dc7d2ff73d7f7416f006b6e460d9509dd07f6466`. The two-file successor delta is clean under `git diff --check`.

The exact suite parses and its baseline passes: 105 files, 60 stripped, 4 reviewed-safe, and 41 frozen and unaudited. User-spawn remains classified stripped. The auth child, 11-file/14-spread population sentinels, readiness behavior, changeset, and package output remain intact.

### Independent call-site verification

I bypassed each cooked-to-source mapper call independently, using edits separate from the shipped mutation fixture:

- no-substitution literal mapper bypass: first red was `an escaped child-script scrub must retain its source brace depth`
- template-head mapper bypass: first red was `an escaped template-head scrub must retain its source brace depth`
- template-span mapper bypass: first red was `an escaped template-span scrub must retain its source brace depth`

This re-derives N13 and N14 separability directly. The new safe cells and their no-scrub negative controls both discriminate.

I also ran parse-checked probes outside the registered cells in both directions. Escaped braces before later head/span code did not falsely clear nested scrubs or falsely red top-level scrubs. Nested-template head mapping also passed.

### Mutation bound and ordering

The config contains 14 mutations naming 12 distinct cells. N5/N7 share the regex-close cell and N6/N10 share the unused-nested-template cell. Variant D, the split Unicode spelling `f\u006fr`, is detected by the spelling-equivalence loop but is named by no mutation.

In a clean isolated exact-head tree, I reran all N1-N14 sequentially. All 14 were KILLED, and each mutation's first named red matched its own configured `expectRed`, including N13 and N14. There were zero `WRONG-RED` outcomes. The tree restored clean.

One discarded run overlapped with dependency recreation by my typecheck and produced WRONG-RED for N12/N13. That run was not valid evidence because its shared dependency target was being replaced concurrently. After typecheck completed, the sequential rerun above passed 14/14 with correct ordering.

### Hosted and type evidence

The exact-head hosted reproof selected 8 fixtures from 351, killed N1-N14, and ended `6 discriminated, 2 pre-red, 0 inconclusive`. The two pre-red fixtures are the unchanged user-spawn readiness and whole-row regrant pair. All exact-head hosted checks are terminal: 17 success and the scheduled-only `full` skip.

I ran `pnpm typecheck` in the isolated exact-head tree. It completed successfully. Because pnpm validated and recreated the reused dependency layout, this also ran the repository build before recursive typechecks. I did not start a stack, run a live suite locally, use the dev-tree CLI, run `cotal up/down`, or run `pnpm check`.

No blocking finding remains at this SHA.

davidfarah2003:
## Exact-head verdict: GRADE PASS

Exact head: `821e72e1a6bea66acf3ffc888ae5d0cb3625d59d`.
Parent: `1b2680d72460f00d0d8a23ebe87bccba20341949`.
Tree: `dc7d2ff73d7f7416f006b6e460d9509dd07f6466`.

Live `refs/pull/1340/head` matches that commit. The successor delta is exactly two files (`bin/smoke/suite-ambient-env.smoke.ts` and `bin/smoke/mutations/suite-ambient-env.mutations.json`) and is clean under `git diff --check`. Occupied cwd remains the older tree `0855cb7983657f6dc3ed7dda4e2a4ee95173af67` (porcelain 0). Measurements below are from git objects at the named SHA (`git show` / `git rev-parse`), not from that working tree.

### What ran vs what was read

Ran: object-store reads of the named SHA, independent hosted `changed` log census for job `101491810206` (run `34035176026`), GitHub PR/check/comment reads, and a cooked-value search guard against the current MANIFEST.

Did not run: `pnpm typecheck`, local `mutation-proof`, a live suite, the dev-tree CLI, `cotal up`/`down`, or `pnpm check`. Typecheck remains closed by another reviewer's isolated exact-head run (comment [5559754688](https://github.com/Cotal-AI/Cotal/pull/1340#issuecomment-5559754688)), not by this review.

Independence bound: the local discriminator used the same disclosed instrument the lane manager published first. That is not a second corpus.

### Reach / DIFF against the parent

Successor vs parent:

- no-substitution: 83/5 vs 77/5
- template-head: 7/1 vs 5/0
- template-span: 7/1 vs 5/0

Positive control: 35/0.

N13 first-red is the template-head brace-depth cell only. N14 first-red is the template-span brace-depth cell only. Cross-checks held. Hygiene was clean.

The mutation config at this SHA names 14 mutations and 12 distinct `expectRed` cells (N5/N7 share the regex-close cell, N6/N10 share the unused-nested-template cell).

### Hosted `changed`

Job `101491810206` at this SHA: selected 8 of 351, ended `6 discriminated, 2 pre-red, 0 inconclusive`. Census counted by verdict word plus N-number, not by grepping the word KILLED: N1 through N14 each appear once as KILLED on the ambient fixture. SURVIVED 0, WRONG-RED 0. The two pre-red fixtures are the unchanged user-spawn readiness and whole-row regrant pair. Those blobs are identical at this head, its parent, merge-base `5e635a22`, and `origin/main` `52901c62`. The documented restore/build-artifact mechanism predates this branch (issue #406 opened 2026-08-14). This diff does not touch them. File identity on main is not a main PRE-RED measurement.

Exact-head hosted checks: 17 SUCCESS, 1 designed SKIP (`full`), 0 fail, 0 cancel. `ci-ok` SUCCESS. PR OPEN / MERGEABLE / CLEAN.

### Bounds that still hold

False-CLEAR was searched over a 240-comparison grid and is not proven absent.

I do not merge.

davidfarah2003:
## APPROVE

Exact live head: `821e72e1a6bea66acf3ffc888ae5d0cb3625d59d`.
Parent: `1b2680d72460f00d0d8a23ebe87bccba20341949`.
Tree: `dc7d2ff73d7f7416f006b6e460d9509dd07f6466`.

The prior BLOCK at `d8872fbc9ef6ac5e98c117603bc919f55ab43e97` does not transfer. Isolated worktree, shared checkout not switched.

First-hand at this SHA:
- `pnpm smoke:suite-ambient-env` PASS. Census 105 files (60 strip, 4 reviewed-safe, 41 frozen).
- Replayed the original nested unused inner-template plant (`bin/smoke/_relb-nested-template.smoke.ts` with unused inner `PROCESS_ENV_STRIP` plus a later module `{ ...process.env }`). The census now first-reds that file as an unstripped spread at line 2 (exit 1). Removing the plant restores PASS.
- Shipped `node scripts/mutation-proof.mjs --config bin/smoke/mutations/suite-ambient-env.mutations.json`: all 14 mutations KILLED, including N6 and N10 naming `an unused nested template scrub must not clear a module-code spread`.

Independently confirmed hosted runs at this SHA succeeded: Mutation reproof `34035176026`, CI `34035176045`, Windows `34035176004`.

Did not merge. Retrospective BLOCK of merged #1341 at `4d43c9799` is unchanged. Do not merge #1342.

PR #1345, opened 2026-09-06
fix(ci): attribute pre-red command transitions

### LINKED ISSUE #1344: Mutation reproof dismisses changed-suite baseline failures
## Problem

`mutation-reproof` now selects a fixture when its configured suite changes. A selected fixture whose baseline exits with the pre-red code is still classified as unrelated and the gate exits successfully.

That is wrong when the suite change itself caused the baseline failure. Two minimal committed diffs reproduce the false green:

1. Change only the configured suite so its baseline exits non-zero.
2. Delete only the configured suite.

Both select the fixture, report it as PRE-RED, print `MUTATION REPROOF OK`, and exit 0.

## Expected behavior

A baseline failure attributable to a changed configured suite must fail the reproof gate. A genuinely pre-existing red fixture selected only because its guarded source changed must retain the nonfatal PRE-RED classification.

## Acceptance

- A suite-only change that makes the configured suite red fails and names the fixture.
- Deleting the configured suite fails and names the fixture.
- A pre-existing red suite selected through a guarded-source change remains a nonfatal PRE-RED.
- Config-path, suite-path, and guarded-source selection remain intact.
- A mutation removes the attribution check and is killed at a named regression cell.

davidfarah2003:
## Summary

A mutation fixture may declare one suite while its mutations run several distinct commands. `mutation-proof` stops at the first red command, so attributing PRE-RED from the declared suite path or from only that first refusal can blame a healthy suite or hide a later attributable regression.

This change classifies every distinct fixture command by its actual base-to-head transition on the differential `--base`/`--head` path used by the pull-request gate:

- retain and deduplicate the top-level command plus every per-mutation override
- after any head PRE-RED, run every command in the same deterministic order in independently prepared head and base snapshots
- green at base and red at head is attributable and fatal
- matching red failures at base and head are inherited and nonfatal
- absent, contaminated, ambiguous, or unmeasurable comparisons fail loud and never clear
- report the actual fixture, command, and transition rather than asserting that the declared suite is red

Each disposable snapshot performs its own network-capable frozen install and full build. Fixture commands then run offline with root-local `PATH` and `NODE_PATH` entries removed. Preparation failures are unmeasured and fatal. Root-only execution-state red is reported as contamination rather than attributed to the diff. Mutation-proof children also receive a copy of the environment with inherited `COTAL_*` material removed.

Failure comparison retains stable origin identity without treating volatile layout as causation. One shared classifier handles source headers and stack frames: repository files become `<ROOT>` paths, dependency origins are dropped, OS-temp paths collapse their per-run segment using both raw and resolved temp-root spellings, external files retain their path and function identity, and line/column coordinates are removed. Non-path semantic text remains verbatim.

## Regression coverage

- changed green suite A with untouched inherited-red command B stays nonfatal, names B, and never claims A is red
- inherited-first and attributable-later commands remain fatal in both command orders
- dist-backed commands are compared after independent frozen install and full build
- snapshot preparation failure, root poisoning, asymmetric sequencing, and root-local executable leakage fail loud
- real caught and uncaught transcripts preserve repository file/function identity while tolerating coordinate changes
- external frame/header pairs distinguish different files while tolerating line shifts
- realistic semantic `:number` and changed-error-class transcripts remain different
- dependency-origin headers tolerate `.pnpm` store-layout differences
- ordinary and symlinked OS-temp roots collapse only their randomized segment, while different files and cross-snapshot contamination remain different
- the symlinked TMPDIR control emits raw and resolved spellings so removing either spelling, or anchoring on literal `/tmp`, fails the MATCH arm
- `--all`, deletion, rename, masking, INCONCLUSIVE, and ordinary mutation findings remain pinned
- thirteen code mutations cover selection, transition causation, preparation, isolation, symmetric ordering, stack origins, and verdict parsing

## Deliberate boundaries

- Differential attribution applies to `--base`/`--head`, which is the path used by the PR gate. `--all` remains the pre-existing uncompared inventory used by the scheduled sweep: it reports PRE-RED without a base comparison. Closing that floor is separate work.
- Suite-only selection is unchanged. The corpus is the 351 fixtures matched by the two globs in `fixtureConfigPaths` (`*/mutations/*.json` and `*.mutations.json`), carrying 1797 `expectRed` anchors. The suite trigger fires by set membership, `changed.has(fixture.suite)`, so it works for the 292 fixtures whose `suite` is an exact repository path. The remaining 59 have no usable suite trigger: 51 declare no `suite` key and 8 carry prose rather than a path, each naming either zero files or two. Those diffs are not selected by the PR gate and rely on the scheduled `--all` sweep. Deriving suite paths from commands is separate work.
- Live-fixture identification must follow the resolved command rather than a substring or a declared-field shortcut. 24 of the 351 fixtures run a command whose resolved form matches `/[-:]live(?![A-Za-z0-9-])/`, resolving every `pnpm` invocation in the chain, including per-mutation `command` overrides. The trailing guard keeps `-liveness` out; requiring `live` to end the segment keeps a leading subject word such as `smoke:live-job-conclusion` out, which runs no broker.
- Dist comparison assumes fixed, non-hashed output paths. The repository includes esbuild users, but the covered outputs are fixed-name, non-minified, non-hashed emits.
- Preparation may use the network. Fixture commands remain offline.

## Verification

- `npx tsx bin/smoke/mutation-reproof.smoke.ts` (49 passed, 0 failed)
- independent acceptance battery for #1344 (12 passed, 0 failed)
- mutation anchor gate: zero dead, ambiguous, prose-spanning, missing, or walk failures
- mutation partitions 1-7 and 7-13 each completed on this commit's tree, overlap at the install/build mutation, every mutation killed at its named red, clean restoration and empty porcelain asserted before and after each
- the suite's own `expectRed` labels were counted against the green baseline transcript, the artifact the grader consults: 13 lookups over 11 distinct labels, each matching exactly one line
- end-to-end paired defect matrix: erasing and over-retaining origin policies red their paired arms; tmpdir-only, realpath-only, and literal-`/tmp` policies each red the raw/resolved MATCH arm; fixed code returns 49/49 green
- first-command-only and stack-origin mutations also killed alone against their observed named red cells
- targeted TypeScript diagnostics for the changed smoke file passed with unused checks enabled
- the repository CLI-only typecheck's missing workspace declaration failure reproduces on untouched `52901c629`
- `pnpm smoke:suite-ambient-env` passed with an inherited `COTAL_*` sentinel

Closes #1344

### DIFF
--- a/bin/smoke/mutation-reproof.smoke.ts
+++ b/bin/smoke/mutation-reproof.smoke.ts
@@ -16,7 +16,7 @@
  *   - a guarded source RENAMED away (a dangling fixture — the fixture still points at the old path).
  */
 import { execFileSync, spawnSync } from "node:child_process";
-import { mkdtempSync, writeFileSync, rmSync, mkdirSync } from "node:fs";
+import { mkdtempSync, readFileSync, realpathSync, writeFileSync, rmSync, mkdirSync, symlinkSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
@@ -40,14 +40,28 @@ const selectedPaths = (out: string): string[] => {
 /** Parse the dangling-fixture set the scan names when it refuses an unrunnable proof. */
 const danglingPaths = (out: string): string[] =>
   [...out.matchAll(/^ {2}(\S+\.json) -> missing:/gm)].map((m) => m[1]);
-/** Parse the offender set when a re-proven fixture's mutant survives. */
-const offenderPaths = (out: string): string[] =>
-  [...out.matchAll(/MUTATION REPROOF FAILED \(\d+ fixture\(s\)\): (.+)/g)].flatMap((m) => m[1].split(", "));
-/** Parse the pre-red set: a selected fixture whose suite was already red before mutation. */
+/** Parse the inherited pre-red set: the same command was red at base and at head. */
 const preRedPaths = (out: string): string[] => {
-  const m = out.match(/^PRE-RED \(\d+ fixture\(s\)\)[^:]*: (.+)$/m);
-  return m ? m[1].split(", ") : [];
+  const block = out.match(/^PRE-RED \(\d+ fixture\(s\)\)[^\n]*\n((?:  .+\n?)+)/m);
+  return block ? [...block[1].matchAll(/^ {2}(\S+\.json) -> command:/gm)].map((m) => m[1]) : [];
 };
+/** Parse pre-red proofs attributed by the same command's base-GREEN -> head-RED transition. */
+const attributablePreRedPaths = (out: string): string[] => {
+  const block = out.match(/^PRE-RED TRANSITION FAILED \(\d+ fixture\(s\)\)[^\n]*\n((?:  .+\n?)+)/m);
+  return block ? [...block[1].matchAll(/^ {2}(\S+\.json) -> command:/gm)].map((m) => m[1]) : [];
+};
+/** Parse base comparisons the gate could not measure and therefore refuses loudly. */
+const unmeasuredPreRedPaths = (out: string): string[] => {
+  const block = out.match(/^PRE-RED TRANSITION UNMEASURED \(\d+ fixture\(s\)\)[^\n]*\n((?:  .+\n?)+)/m);
+  return block ? [...block[1].matchAll(/^ {2}(\S+\.json) -> command:/gm)].map((m) => m[1]) : [];
+};
+/** Parse every fatal offender, retaining PRE-RED transition states as distinct output categories. */
+const offenderPaths = (out: string): string[] => [
+  ...[...out.matchAll(/^MUTATION REPROOF FAILED \(\d+ fixture\(s\)\): (.+)$/gm)]
+    .flatMap((m) => m[1].split(", ")),
+  ...attributablePreRedPaths(out),
+  ...unmeasuredPreRedPaths(out),
+];
 /** Parse the inconclusive set: a selected fixture whose proof produced no evidence either way. */
 const inconclusivePaths = (out: string): string[] => {
   const m = out.match(/^INCONCLUSIVE \(\d+ fixture\(s\)\)[^:]*: (.+)$/m);
@@ -55,12 +69,22 @@ const inconclusivePaths = (out: string): string[] => {
 };
 /** Parse the machine-readable outcome counts from the final OK summary. Null when the gate failed. */
 const okCounts = (out: string): { discriminated: number; preRed: number; inconclusive: number } | null => {
-  const m = out.match(/MUTATION REPROOF OK \(\d+ fixture\(s\) selected; (\d+) discriminated, (\d+) pre-red, (\d+) inconclusive\)/);
+  const m = out.match(/MUTATION REPROOF OK \(\d+ fixture\(s\) selected; (\d+) discriminated, (\d+) inherited pre-red, 0 attributable pre-red, 0 unmeasured pre-red, (\d+) inconclusive\)/);
   return m ? { discriminated: Number(m[1]), preRed: Number(m[2]), inconclusive: Number(m[3]) } : null;
 };
 
-const scan = (root: string, base: string, head: string): { status: number | null; out: string } => {
-  const run = spawnSync(process.execPath, [SCAN, "--root", root, "--base", base, "--head", head], { encoding: "utf8" });
+const cleanAmbientEnv = (): NodeJS.ProcessEnv => {
+  const env = { ...process.env };
+  for (const key of Object.keys(env)) if (key.startsWith("COTAL_")) delete env[key];
+  return env;
+};
+
+const scan = (root: string, base: string, head: string, env = cleanAmbientEnv()): { status: number | null; out: string } => {
+  const run = spawnSync(process.execPath, [SCAN, "--root", root, "--base", base, "--head", head], { encoding: "utf8", env });
+  return { status: run.status, out: `${run.stdout ?? ""}${run.stderr ?? ""}` };
+};
+const scanAll = (root: string): { status: number | null; out: string } => {
+  const run = spawnSync(process.execPath, [SCAN, "--root", root, "--all"], { encoding: "utf8", env: cleanAmbientEnv() });
   return { status: run.status, out: `${run.stdout ?? ""}${run.stderr ?? ""}` };
 };
 
@@ -267,11 +291,8 @@ try {
     );
   }
 
-  // 6. A selected fixture whose suite is ALREADY RED before any mutation. a.mjs changes (so a's
-  //    fixture is selected) but a's suite fails unconditionally, so mutation-proof refuses at its
-  //    baseline (exit 4). That is the suite's own defect, not this diff's — the #1279/sandbox-guard
-  //    shape the reviewer reproduced. The gate must report it as PRE-RED, name exactly that fixture,
-  //    and NOT fail: collapsing pre-red into a blocker is a false blocker.
+  // 6. A configured suite changed by this diff becomes red before mutation. The SAME command is green
+  //    at base and red at head, so the transition — not the selected path — attributes the failure.
   {
     const { root, base, head } = build((r) => {
       writeFileSync(join(r, "a.suite.mjs"), [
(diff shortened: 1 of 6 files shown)


davidfarah2003:
## Review of `49f9f2fe69c52ed90218bf9ec52c569e9772fb7b`

**BLOCK**

Hosted Mutation reproof `changed` failed on this exact SHA: https://github.com/Cotal-AI/Cotal/actions/runs/34056214493/job/101548479725 (run `34056214493`, job `101548479725`). The gate selected `bin/smoke/mutations/mutation-reproof.json`, emitted baseline provenance for `pnpm smoke:mutation-reproof` at status 1, then reported `base GREEN (exit 0) -> head RED (exit 1)`.

The six-file change makes that PRE-RED self-inflicted. `rootComparisonEnv()` is `comparisonEnv()` with `offline` defaulting to true, so the root proof child runs with `pnpm_config_offline=true`. The self-fixture suite then executes synthetic `pnpm install --lockfile-only` with no env of its own and inherits that offline pin. Snapshot preparation correctly uses `offline: false`. The suite lockfile-only path does not. Local partitions that killed 19 unique committed anchors do not clear a hosted self-fixture PRE-RED.

Independent notes on this object (not an approval): refused provenance is one JSON record from the exact refused run and is compared by hash, so a different clean-head signature cannot false-clear. Unavailable, empty, and infrastructure failures fail loud. Only the exact missing-`node_modules` `[WARN]` is filtered. Line-initial semantic `[WARN]` stays evidence. Verdict tokens are line-initial after ANSI strip. The tool-list wrapper stays a `pnpm` script with `pnpm_config_verify_deps_before_run=false`. The 19 committed `find` strings are unique, code-only, and one-hit. Treat 19 unique KILLED as the proven claim, not per-cell exclusivity.

Non-blocking limits still on this SHA: two PATH leak-string clauses stay unreachable under the current echo/status policy. Root and snapshot env wrappers are nested seams over one `comparisonEnv`. `runCommand` has no default env for a hypothetical omitted-arg call. The comparison-env comment claims three selftest fixtures, while `mutation-proof.selftest.mjs` has one `pnpm_config_verify_deps_before_run` assertion. Package-manager chrome prefixes remain dropped from stable signatures.

Any successor SHA needs a fresh review. Do not merge this object.

davidfarah2003:
**BLOCK** `49f9f2fe69c52ed90218bf9ec52c569e9772fb7b`

The exact-SHA hosted Mutation reproof `changed` job failed: https://github.com/Cotal-AI/Cotal/actions/runs/34056214493/job/101548479725. The candidate routes the root proof child through `comparisonEnv()` with offline mode enabled, so the self-smoke's unscoped `pnpm install --lockfile-only` inherits `pnpm_config_offline=true`; the fixture baseline becomes head-red and the gate correctly reports base GREEN -> head RED.

Static review found the refused-run hash/status/infrastructure provenance bounded to the exact run, normalized root/head/base command environments and clean snapshot cwd, loud unavailable/infrastructure/empty handling, exact-only pnpm warning filtering with semantic `[WARN]` retained, anchored verdict parsing, and a no-install tool-list wrapper. The five supplied partitions bind exactly to the committed 19 unique code-only anchors and show 19 unique KILLED verdicts. That evidence does not override the candidate failing its own hosted gate. Known non-blocking limits remain: two unreachable leak-string clauses, nested env cells, no default `runCommand` env for a future caller, the three-selftests comment mismatch, and normalized package-manager chrome prefixes.

davidfarah2003:
Correction to the mechanism above: `pnpm_config_frozen_lockfile=true` is the decisive inherited setting. The synthetic `pnpm install --lockfile-only` runs begin without `pnpm-lock.yaml`, so they fail with `ERR_PNPM_NO_LOCKFILE`. I independently confirmed that the same local `file:` dependency succeeds with offline mode alone. The **BLOCK** verdict is unchanged.

davidfarah2003:
## Correction on `49f9f2fe69c52ed90218bf9ec52c569e9772fb7b`

**BLOCK stands.** Hosted Mutation reproof `changed` is still failed on this SHA: https://github.com/Cotal-AI/Cotal/actions/runs/34056214493/job/101548479725 (run `34056214493`, job `101548479725`).

The prior comment named `pnpm_config_offline=true` as the decisive pin. That was wrong. `comparisonEnv()` always sets `pnpm_config_frozen_lockfile=true`, and `rootComparisonEnv()` applies it to the proof child. The new self-fixture cell runs `pnpm install --lockfile-only` before any lockfile exists. Frozen-lockfile with no `pnpm-lock.yaml` is `ERR_PNPM_NO_LOCKFILE`. Isolated check of the same `file:vendor/probe` install: frozen=true exits 1 with that error (including when offline is also true); offline=true with frozen=false exits 0. Offline alone does not fail workspace/`file:` deps.

davidfarah2003:
## Head rewritten: every prior verdict on this branch is void

This branch was force-pushed. **`49f9f2fe` → `2fca4202`.**

**Every review verdict, PASS, and BLOCK recorded against `49f9f2fe` is void.** A verdict does not
survive head movement, so nothing previously posted here carries forward to the new head. The
standing BLOCK is not being answered or overridden; it is spent, because the tree it described no
longer exists on this branch.

**The old head is preserved, not orphaned.** `49f9f2fe` is permanently fetchable at
`refs/heads/archive/1344-49f9f2fe`. The diagnosis in this thread remains the only readable account
of that hosted failure (the job log is unretrievable), so the tree it refers to stays anchored to a
real ref rather than to an orphan that expires.

### What changed

1. **Rebased onto current `main` (`ab2ef1df`).** The previous head was built on `52901c62`, many
   merges behind. Both commits replayed cleanly; `scripts/mutation-proof.mjs` auto-merged against
   main's changes with no manual resolution.

2. **The environment defect that made the hosted job red is fixed.** `networkSetupEnv()` deletes
   `pnpm_config_offline` and sets `pnpm_config_frozen_lockfile=false` at the four install and
   verification sites. Synthetic lockfile generation necessarily runs before a lockfile exists, so
   an inherited `frozen_lockfile=true` made `pnpm install --lockfile-only` fail with
   `ERR_PNPM_NO_LOCKFILE` before the fixture could be built. The explicit `--frozen-lockfile`
   command still supplies the verification policy on the command line.

3. **A second defect, found after the BLOCK, is also fixed.** In `comparableFailure`, a depth-zero
   temp-root path discarded the filename instead of the randomized directory segment, collapsing two
   distinct origins to one signature so a false contamination match was possible. Depth-one and
   deeper behaviour is unchanged.

4. **Two already-consumed changesets were dropped.** They were inherited from the stale base and had
   already been released; carrying them forward would have caused a spurious version bump for work
   that shipped long ago. This branch contributes no changeset:
   `git diff --name-only ab2ef1df HEAD -- .changeset/` is empty.

### Evidence on the new head

The suite runs **57 passed, 0 failed**. Green alone is not the claim. Reverting the depth-zero fix
to its defective form drives the suite to **56 passed, 1 failed**, and the failure names the exact
fixture:

```
REFUSING: `node depth-zero.suite.mjs` is red BEFORE any mutation (exit 1).
  smoke/mutations/depth-zero.mutations.json -> transition: base RED (exit 1) -> head RED (exit 1)
```

Restoring the fix returns the suite to 57/57. The cell is shown red on the named defect and green
with the fix.

### Known gap, deliberately not fixed here

`bin/smoke/mutation-reproof.smoke.ts:487` runs `pnpm build` without an explicit `env:`, between two
sites that do pass one. It inherits ambient `pnpm_config_*` unmodified, which is the same shape as
the defect this PR fixes. It is latent rather than live: no hosted workflow sets those variables,
which was checked rather than assumed. Fixing it would widen a provenance repair, so it is recorded
here instead.

This head needs a fresh independent review. No verdict from before the rewrite applies.

davidfarah2003:
## Review of 2fca4202935dae8b355a970ff64ec277bb73d826 — BLOCK

Measured at exact public head `2fca4202935dae8b355a970ff64ec277bb73d826` (parent `eb0cc6d026fe050e3a643607ef3cd8034abcf267`). The live head matched this SHA before the review and again immediately before this comment.

**BLOCK:** this commit adds `networkSetupEnv()` as `{ ...process.env }` plus the two pnpm setup knobs, then uses it at the four install/verification sites. Hosted CI on this SHA failed `smoke (shard 2/4)` because `pnpm smoke:suite-ambient-env` asserted:

`bin/smoke/mutation-reproof.smoke.ts (unstripped spread at line 82)`

That helper is new versus parent. Scan paths already strip `COTAL_` via `cleanAmbientEnv()`. Setup-install children do not. The census requires a strip or a measured REVIEWED reason. Neither is present.

Independent checks that hold on this tree (not enough to approve):

- Executed `comparableFailure`: a depth-0 temp-root path keeps the filename and stays distinct (`<TMP>/mutation-reproof-depth-zero-alpha.mjs` vs `…omega.mjs`). Depth-1+ is unchanged except collapsing the first randomized segment. Applying the committed depth-0 mutant makes alpha and omega identical (`at go (<TMP>)`).
- The exact pnpm missing-`node_modules` warning is dropped. Line-initial semantic `[WARN]` stays evidence and stays distinct. An empty transcript yields no signature, so comparison cannot clear.
- Unavailable (exit 127) and infrastructure output fail loud before signature matching. Root provenance is one bounded JSON record (command, status, signal, unmeasurableReason, signatureHash). Root failing one signature while both clean snapshots fail another with the same exit is contamination, not inherited PRE-RED.
- At this SHA: `node --experimental-strip-types bin/smoke/mutation-reproof.smoke.ts` printed `mutation-reproof smoke: 57 passed, 0 failed`. Green is not the claim. The depth-0 cell was shown red under the committed mutant and green on the committed function.
- Four `networkSetupEnv()` call sites, all `pnpm install` lockfile/frozen. `pnpm build` at line 487 still has no explicit `env` (known, and not the hosted failure). `runCommand` has no default env.
- `bin/smoke/mutations/mutation-reproof.json` has 20 unique code-only `find` strings.
- `tool-list-announce` uses `pnpm smoke:tool-list-announce` with `pnpm_config_verify_deps_before_run=false` (no-install).
- Verdict parsing matches `^(KILLED|…) ` after ANSI strip, so words in labels cannot become verdicts.

Limits (not the block): the comment that three selftest fixtures assert `verify_deps_before_run=false` overstates the one assertion in `scripts/mutation-proof.selftest.mjs`. `rootComparisonEnv` / `snapshotComparisonEnv` are nested aliases of one `comparisonEnv`. Package-manager chrome prefixes remain normalized.

Do not merge until setup env strips `COTAL_` (or is REVIEWED with a measured reason) and the census is green at this head.

davidfarah2003:
## BLOCK upheld and fixed. Head moved `2fca4202` → `b6a377dd`.

The BLOCK on `2fca4202` was **correct**, and the defect was mine. I reproduced it before fixing it
and I am recording that it was a real regression introduced by this branch, not a pre-existing
condition.

### The defect

`networkSetupEnv()` was built as a raw `{ ...process.env }` spread. Its sibling `cleanAmbientEnv()`
strips every `COTAL_` key; `networkSetupEnv` did not, so the four install and verification sites
handed a child an environment carrying the parent's `COTAL_` material. `pnpm smoke:suite-ambient-env`
exists to forbid exactly that, and it named the line:

```
bin/smoke/mutation-reproof.smoke.ts (unstripped spread at line 82)
```

**This branch caused it.** Measured, not assumed: the same guard at base `ab2ef1df` reports
`suite-ambient-env: PASS`, and at `2fca4202` it exits 1. A repair that fixes an environment leak into
proof children while opening a different environment leak into setup children is not a repair.

### The fix

`networkSetupEnv()` now derives from `cleanAmbientEnv()` instead of a raw spread, so the `COTAL_`
strip applies before the two pnpm adjustments. One line, no new behaviour, no new gate.

### Evidence on the new head

- `pnpm smoke:suite-ambient-env` → **PASS** (exit 0)
- `bin/smoke/mutation-reproof.smoke.ts` → **57 passed, 0 failed**, unchanged
- Red-then-green on the fix itself: restoring the raw spread reproduces
  `unstripped spread at line 82` and exit 1; restoring the fix returns PASS. The guard is shown red
  on the named defect and green with the fix.

### Why the guard caught what my acceptance list did not

My mechanical predicate enumerated every `spawnSync`/`execFileSync` site and asserted none lacked an
explicit `env:`. Every site **had** one, so the predicate passed. It never asked whether the env
being passed was itself clean. The predicate tested the presence of an argument rather than the
correctness of its value, which is the same shape as the false green this PR exists to correct: a
check that passes for the wrong reason.

`suite-ambient-env` is the check that asks the right question, and it was not in my acceptance list.
It is now.

Every verdict on `2fca4202` is void, including the BLOCK it earned. This head needs fresh review.

davidfarah2003:
## Reviewer verdict: APPROVE

Measured at live head `b6a377dd109dfc8e32ed8981314ecf42c40ddd11` (parent `eb0cc6d026fe050e3a643607ef3cd8034abcf267`, merge-base `ab2ef1df441f5284111a8779050b3d009ebddf08`). The prior BLOCK on `2fca420` is void: that tree is not on the branch.

Independent of the manager's report. I re-read the six-file change, then executed the env strip and the signature claims rather than agreeing with the diff.

### What I measured

- `networkSetupEnv()` now starts from `cleanAmbientEnv()` (line 82). All four install sites pass that helper: `bin/smoke/mutation-reproof.smoke.ts` 475, 482, 486, 1154. The only `{ ...process.env }` left in that file is inside `cleanAmbientEnv()`, which deletes every `COTAL_` key before return.
- `pnpm smoke:suite-ambient-env` at this SHA: PASS. `mutation-reproof.smoke.ts` is in the stripped census.
- Red control: restore `{ ...process.env }` in `networkSetupEnv` -> EXIT 1, `bin/smoke/mutation-reproof.smoke.ts (unstripped spread at line 82)`. Restore the committed helper -> PASS again.
- `pnpm smoke:mutation-reproof` at this SHA: **57 passed, 0 failed**.
- Executed `comparableFailure`: depth-0 temp filenames stay distinct on the committed function. The committed mutant `separator === -1 ? "<TMP>"` collapses both depth-0 names. Depth-1 randomized directories still collapse as designed.

### Hosted CI at this SHA (at post time)

- Windows `smoke (shard 2/4 — non-blocking)`: pass. This is the shard that failed on the unstripped spread at `2fca420`.
- Linux `unit`: pass. Linux `smoke (shard 2/4)` was still in progress, not failed.

No new blocker found on this tree. APPROVE `b6a377dd109dfc8e32ed8981314ecf42c40ddd11`.

davidfarah2003:
**APPROVE** `b6a377dd109dfc8e32ed8981314ecf42c40ddd11`

Fresh review of this exact six-file head found no blocking defect. Refused-run provenance is bound to the exact root refusal and compared by command, status, infrastructure classification, and stable signature against clean independently prepared head/base snapshots. Root, head, and base commands share normalized environments and snapshot cwd isolation. Unavailable, infrastructure, empty, and mismatched failures remain loud. Warning filtering removes only pnpm's exact missing-node_modules line while retaining semantic `[WARN]`, verdict parsing is line-start anchored, and the tool-list pnpm wrapper remains no-install.

The committed config contains 20 unique code-only anchors. The exact-head hosted `changed` job ran step 10, not a skipped path, and reported all 20 KILLED plus a clean five-fixture reproof summary. There is no separate b6a377dd on-disk partition corpus, so I used the committed config and hosted exact-head transcript as the proof surface.

I independently exercised `pnpm smoke:suite-ambient-env` with an inherited `COTAL_*` sentinel: fixed 0, behavioral raw-spread mutant 1 naming line 82, restored 0. The one-line `networkSetupEnv()` correction therefore changes the guarded behavior rather than merely satisfying an argument-presence check.

Limits retained: `runCommand` has no default env for a future caller; the synthetic `pnpm build` at smoke line 487 still lacks an explicit env but is latent under current hosted variables; package-manager chrome prefixes remain normalized. Hosted mutation reproof, unit, live, Windows, CodeQL, and one smoke shard passed. Three other smoke shards were externally cancelled with no red transcript, so this is not a claim that the full CI matrix completed green.

davidfarah2003:
## Note for in-flight branches: this replaced the mutation gate

Merged at `6ab195b865af58a87ccd1d77f1314f0882537541`. Recording the downstream effect here because it lands on other open PRs and is not obvious from the diff.

This change replaces the machinery that grades every `Mutation reproof` run:

```
bin/smoke/mutation-reproof.smoke.ts     +943
scripts/mutation-reproof.mjs            +277
scripts/mutation-proof.mjs               +13
scripts/mutation-failure-signature.mjs   +89  (new)
```

**The new gate is stricter, which is the point.** The old one compared status only, so a root failing X and clean snapshots failing Y both exited 1, read as agreement, printed PRE-RED INHERITED and exited 0. The merged predicate compares status *and* a failure-signature hash, and treats a `null` or `undefined` hash as failure rather than as a match — so an unmeasurable signature can no longer stand in for a matching one.

Consequence for anything already in flight: a `Mutation reproof` that went green before 2026-09-08T06:52:38Z was produced by the superseded gate. Whether a given branch's result changes under the new one is unmeasured — this is not a claim that any of them were wrong, only that the instrument has been replaced by one that refuses more. Six of the seven open PRs at merge time carried mutation-related files.

**The fixture schema is unchanged.** One existing fixture was edited (`packages/core/smoke/mutations/tool-list-announce.json`), which is the usual tell for a schema-affecting change, but the edit is `command` moving from `tsx packages/core/smoke/…` to `pnpm smoke:tool-list-announce`. Its `why` block records why that is now safe: `mutation-proof` forces `pnpm_config_verify_deps_before_run=false`, so pnpm resolves `tsx` from each prepared cwd's own `.bin` without installing into the mutation worktree. No field was added or made required, and fixtures in either form still load.

**A head-only run does not cover this.** A branch's own reproof grades its fixtures against the machinery in its own tree; after merging, the new machinery runs over those same fixtures. That pairing has not been executed by either run. Branches wanting to close that gap can write the predicted merge tree with `git merge-tree --write-tree origin/main <head>`, check it out in a scratch worktree, and run the reproof there — it exercises the real pairing without moving the branch head or invalidating verdicts already bound to it.

PR #1353, opened 2026-09-06
fix(manager): supervise dials the recorded broker URL, not raw TCP

davidfarah2003:
`cotal supervise` could not complete manager-authority registration against a remote mesh. Two defects, both reproduced at base before being fixed.

**D1 — the reported defect.** `implementations/manager/src/remote-register.ts` dialed with the raw `connect` from `@nats-io/transport-node`, which refuses `ws://` and `wss://` outright rather than dialing them. A mesh published through an HTTPS/websocket edge was therefore undialable, and the participant saw `servers node client doesn't support websockets, use the wsconnect function instead`. It now selects the dialer by scheme through core's `dialerFor`.

**D2 — found by the reproduction.** The same function passed a prefixed `sha256:<hex>` cluster digest to `fetchContractArtifact`, which is keyed by BARE hex. Remote registration therefore failed over plain TCP as well, so this path was broken on every transport. Normalised with `contractRefToHex`. The static manager never hits it because it reads from an in-memory map keyed by the prefixed digest, which is why only the remote path was affected.

**D3 — `cotal run` on the same mesh.** `implementations/runtime/src/run-command.ts` handed a registry-resolved server URL to the raw transport, failing with the identical refusal D1 fixes for supervise. Moved onto `dialerFor`.

TLS intent is carried from the mesh record rather than re-decided at the call site, so a participant reaching its host through a `wss://` edge no longer sends its prepare credential in the clear.

**Evidence.** `bin/smoke/supervise-remote-dial.smoke.ts` starts an isolated authenticated broker serving both a TCP and a websocket listener and drives the shipped `registerRemoteManagerAuthority`. At base it fails cells A and B with the two refusals above; at this tip it is 9 passed, 0 failed. The mutation manifest re-derives each cell: 4 of 4 mutants killed, each reddening its named cell, with the predicted kill sets matching including the two-cell set for the digest mutant. Full battery green: typecheck, docs-voice, docsbundle, changeset status, ci-suites, ci-suites-freeze, core-boundary, docs, and the ambient-env census.


davidfarah2003:
The Mutation reproof failure on this branch is pre-existing on `main` and not caused by this
change. Controlled rather than asserted.

I ran the same fixture in a detached worktree at `origin/main` (`52901c629`), which contains no
part of this PR:

```
baseline green (15 progress marks)
SURVIVED   the verb gate is gone, so an unroutable verb reaches the planes
           positive control: another mutation in implementations/runtime/src/run-command.ts was
           KILLED in this run, so the suite provably reaches this file at runtime and this
           survivor is a real coverage gap · 15 marks (baseline 15)
KILLED     the usage line loses a verb the gate still routes · 14 marks (baseline 15)
1 of 2 mutation(s) did not produce a clean, named red.
```

Identical to CI here: same survivor, same wording, same mark count. Both of the fixture's anchors
match exactly once on `main`, so the fixture is valid there and the comparison is real.

Why this branch surfaced it: the job re-proves fixtures for *changed guarded sources*, and this
PR touches `implementations/runtime/src/run-command.ts` (swapping `connect(...)` for
`dialerFor(conn.server)(...)` in `openPlanes`, one of the three call sites this change is about).
That selected the fixture. It did not create the survivor.

The gap is real and worth its own fix: with the verb gate removed, nothing in
`smoke:runtime-run-command-usage` goes red. The fixture's own notes anticipate why the exit code
cannot discriminate, since the connect helper exits 1 by itself, but the usage cell does not
catch it either. Filed separately rather than bundled here.

What this PR's own fixtures did: `supervise-registered-user` passes after the re-anchor, and
`ps-mesh-projection` killed 7 of 7. All 18 other checks are green, including all four smoke
shards, `live`, `ci-ok` and `windows-ok`.

Issue #1359, opened 2026-09-07
release: bootstrap trusted publishing for @cotal-ai/seat


davidfarah2003:
The 0.47.0 Changesets run failed after merging release PR #1338.

Run: https://github.com/Cotal-AI/Cotal/actions/runs/34072874048
Merge: `609ce674235e32517a7d726585d2f736d629a77e`

Observed:

- `@cotal-ai/lang@0.47.0` published successfully.
- `@cotal-ai/seat@0.47.0` is a new package and does not exist on npm.
- npm OIDC exchange returned 404, then the publish PUT returned 404.
- The workflow has `id-token: write` and no `NPM_TOKEN`, so the existing trusted-publishing path cannot bootstrap this new package.
- Publication stopped before `@cotal-ai/core`, `@cotal-ai/auth`, `@cotal-ai/workspace`, and the remaining 0.47.0 packages.

Required recovery:

1. Bootstrap `@cotal-ai/seat` on npm under the `@cotal-ai` scope through an owner-authorized initial publication path.
2. Configure this repository's `changesets.yml` workflow as its trusted publisher.
3. Re-run the failed Changesets workflow.
4. Verify every 0.47.0 package in the publish closure, including `cotal-ai@0.47.0`, before cutting or accepting the GitHub release.

Do not work around this with a local publish. The repository release path is trusted publishing.

davidfarah2003:
Registry census immediately after failed run `34072874048`:

- fixed group: 22 packages
- `@cotal-ai/lang@0.47.0`: published and currently `latest`
- `@cotal-ai/seat`: registry returns 404; no package exists
- remaining 20 fixed-group packages: still `latest=0.46.0`, with no `0.47.0` version
- `cotal-ai`: still `latest=0.46.0`

The release is therefore partially published, not merely delayed. Do not rerun the recursive OIDC publisher before the new package is bootstrapped; the next recovery must account for the already-published lang version and verify all 22 exact versions afterward.

davidfarah2003:
Retrospective release review is complete: the exact #1338 version tree is internally correct, but publication is blocked.

- head `02e693e2b57fd5e3d361f87df05c65e08adaf4db`
- merge `609ce674235e32517a7d726585d2f736d629a77e`, with parent2 equal to the head and the same tree
- 61 release-only paths: 10 consumed changesets, 25 manifests, 25 changelogs, and the generated bundle version line
- all 22 public fixed-group manifests are 0.47.0

[...]

PR #1361, opened 2026-09-07
fix(release): bootstrap seat publication

davidfarah2003:
## Scope

- add a one-time manual workflow that publishes only the immutable `@cotal-ai/seat@0.47.0` tarball from successful CI run `34072874022`
- bind repository and actor IDs, branch, source run and commit, artifact ID, zip and tarball digests, package metadata, protected environment, and confirmation string before publication
- expose the temporary npm credential to one fixed package-only publish step through a mode-0600 temporary npm userconfig
- document owner setup, distinct second-owner approval, every-attempt credential teardown, failed-attempt stops, trusted publisher setup, unchanged Changesets rerun, and full closure verification

## Owner checkpoints

The lane does not create or change GitHub environments, secrets, npm tokens, or npm settings. Before dispatch, an owner must:

1. merge this PR with `[skip ci]` in the merge commit subject so adopting the temporary workflow does not start the blocked Changesets publisher;
2. create the shortest-lived granular npm token with read/write access limited to the `@cotal-ai` scope, no organization-management access, and 2FA bypass only for this initial automated publish, or stop if scope-only access is unavailable;
3. identify a distinct second repository owner, or stop if none is available;
4. create the protected `seat-bootstrap` environment, restrict it to `main`, require that second owner, prevent self-review, and disable administrator bypass;
5. store the token only as the environment secret `NPM_SEAT_BOOTSTRAP_TOKEN`.

After every run attempt, including failure or cancellation, the owner must first revoke the token and delete the environment secret. Registry state then controls the next action:

- 404, any non-200 response, parse failure, missing fields, or mismatched version/integrity: keep the workflow and environment, stop for diagnosis, and use fresh credentials plus the same distinct-owner approval for any retry;
- 200 with exact `@cotal-ai/seat@0.47.0` and pinned integrity: configure the package's trusted publisher for `Cotal-AI/Cotal` and `changesets.yml`, explicitly allow direct `npm publish`, delete the bootstrap environment, and remove the workflow and runbook with `[skip ci]`.

Only then rerun failed Changesets run `34072874048`. Existing packages continue using the unchanged token-free OIDC workflow. Require `node scripts/verify-publish-closure.mjs 0.47.0` to report `PUBLISHED/22` before accepting or creating the GitHub release.

## Safety

The workflow has only `workflow_dispatch`, a fixed owner actor, numeric repository and actor IDs, `refs/heads/main`, a literal confirmation, `actions: read`, and job-local `id-token: write` for provenance. It does not edit `changesets.yml`, use recursive publishing, receive a credential through a dispatch input, use web login, or use a mutable action reference.

The first seat publish authenticates with the short-lived granular token and its provenance names the emergency workflow. Pinned IDs and digests bind the published bytes to the tarball that CI built and loaded on both supported Linux architectures. Later releases use the normal `changesets.yml` trusted publisher.

## Verification

- PyYAML and checksum-verified `actionlint` accepted the workflow
- `git diff --check` passed
- `node scripts/check-docs-voice.mjs docs/release-seat-bootstrap.md` passed
- live artifact metadata matches run `34072874022`, commit `609ce674235e32517a7d726585d2f736d629a77e`, artifact `10001064028`, and zip digest `8da2dbd67a9a2252b0729fdb3919481a223a0e6e94d4fe268bb411b44fd7d179`
- the tarball matches SHA-256 `ab8c3309f1effb9eb5bda7b59076f1662c5642b74becc674e46813bc5ab86c0f` and the pinned registry integrity
- `npm publish --dry-run --json` processed all 37 files, both native paths, and the public/latest decision without registry writes
- compiler-blocked install/import and a real `launchSeat`/`adoptSeat` lifecycle exercised the shipped linux-x64 native; the source CI run passed the same-tarball linux-arm64 load job
- tampered artifacts, altered dispatch contexts, mismatched registry states, and post-secret boundary mutations were refused
- live npm registry still returns 404 for `@cotal-ai/seat@0.47.0`

The credentialed dispatch was not run because it requires owner settings and can publish.

Refs #1359

### DIFF
--- a/.github/workflows/bootstrap-seat.yml
+++ b/.github/workflows/bootstrap-seat.yml
@@ -1,0 +1,154 @@
+name: Bootstrap @cotal-ai/seat
+
+on:
+  workflow_dispatch:
+    inputs:
+      confirm:
+        description: "Type: publish @cotal-ai/seat@0.47.0 from artifact 10001064028"
+        required: true
+        type: string
+
+concurrency:
+  group: bootstrap-cotal-ai-seat-0.47.0
+  cancel-in-progress: false
+
+permissions:
+  actions: read
+
+jobs:
+  publish:
+    name: Publish exact CI-tested seat tarball
+    if: >-
+      github.event_name == 'workflow_dispatch' &&
+      github.repository == 'Cotal-AI/Cotal' &&
+      github.repository_id == '1256708592' &&
+      github.actor == 'davidfarah2003' &&
+      github.actor_id == '37240868' &&
+      github.ref == 'refs/heads/main' &&
+      inputs.confirm == 'publish @cotal-ai/seat@0.47.0 from artifact 10001064028'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    environment: seat-bootstrap
+    permissions:
+      actions: read
+      id-token: write
+    steps:
+      - name: Verify immutable source artifact
+        id: gate
+        env:
+          GH_TOKEN: ${{ github.token }}
+          EXPECTED_ARTIFACT_ID: "10001064028"
+          EXPECTED_ARTIFACT_NAME: seat-npm-pack
+          EXPECTED_ARTIFACT_ZIP_SHA256: 8da2dbd67a9a2252b0729fdb3919481a223a0e6e94d4fe268bb411b44fd7d179
+          EXPECTED_HEAD_SHA: 609ce674235e32517a7d726585d2f736d629a77e
+          EXPECTED_REPOSITORY_ID: "1256708592"
+          EXPECTED_RUN_ID: "34072874022"
+        run: |
+          set -eu
+          artifact=$(gh api "repos/Cotal-AI/Cotal/actions/artifacts/$EXPECTED_ARTIFACT_ID")
+          test "$(printf '%s' "$artifact" | jq -r .id)" = "$EXPECTED_ARTIFACT_ID"
+          test "$(printf '%s' "$artifact" | jq -r .name)" = "$EXPECTED_ARTIFACT_NAME"
+          test "$(printf '%s' "$artifact" | jq -r .expired)" = "false"
+          test "$(printf '%s' "$artifact" | jq -r .digest)" = "sha256:$EXPECTED_ARTIFACT_ZIP_SHA256"
+          test "$(printf '%s' "$artifact" | jq -r .workflow_run.id)" = "$EXPECTED_RUN_ID"
+          test "$(printf '%s' "$artifact" | jq -r .workflow_run.repository_id)" = "$EXPECTED_REPOSITORY_ID"
+          test "$(printf '%s' "$artifact" | jq -r .workflow_run.head_repository_id)" = "$EXPECTED_REPOSITORY_ID"
+          test "$(printf '%s' "$artifact" | jq -r .workflow_run.head_branch)" = "main"
+          test "$(printf '%s' "$artifact" | jq -r .workflow_run.head_sha)" = "$EXPECTED_HEAD_SHA"
+
+          zip="$RUNNER_TEMP/seat-npm-pack.zip"
+          gh api "repos/Cotal-AI/Cotal/actions/artifacts/$EXPECTED_ARTIFACT_ID/zip" > "$zip"
+          echo "$EXPECTED_ARTIFACT_ZIP_SHA256  $zip" | sha256sum --check --strict
+          mkdir "$RUNNER_TEMP/seat-npm-pack"
+          python3 - "$zip" "$RUNNER_TEMP/seat-npm-pack" <<'PY'
+          import sys
+          from pathlib import Path
+          from zipfile import ZipFile
+
+          archive = Path(sys.argv[1])
+          destination = Path(sys.argv[2]).resolve()
+          with ZipFile(archive) as source:
+              for entry in source.infolist():
+                  target = (destination / entry.filename).resolve()
+                  if destination not in target.parents and target != destination:
+                      raise SystemExit(f"artifact zip has unsafe path: {entry.filename}")
+              source.extractall(destination)
+          PY
+
+          tarball="$RUNNER_TEMP/seat-npm-pack/cotal-ai-seat-0.47.0.tgz"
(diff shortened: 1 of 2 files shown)


davidfarah2003:
Verified the bootstrap path on `96ef332a` against live GitHub state. All six pinned constants match artifact `10001064028` exactly — id, name, `sha256:8da2dbd6…`, head sha `609ce674`, repository id `1256708592`, run `34072874022`. Artifact is `expired=false`, and the workflow (`352015713`) is registered and active. The gate logic itself checks out.

Two things stand between this and a published seat:

**The `seat-bootstrap` environment does not exist.** `GET /repos/Cotal-AI/Cotal/environments/seat-bootstrap` returns 404; the repo has exactly one environment, `copilot`, with no protection rules and no branch policy. A dispatch now would auto-create `seat-bootstrap` with no rules and no secret, and then fail closed at the workflow's own `test -n "$NODE_AUTH_TOKEN"` guard. Failing closed is the right behaviour, but it means the environment needs creating *with* its protection rules and the `NPM_SEAT_BOOTSTRAP_TOKEN` secret before a dispatch is worth making.

**The artifact expires 2026-09-14T01:25:22Z**, about six and a half days out. Every integrity check in the workflow is pinned to it, so after that the run cannot pass its own artifact gate and the six constants would need re-pinning to a freshly built tarball.

Registry is unchanged: `@cotal-ai/seat` 404, `cotal-ai` 0.46.0, `@cotal-ai/lang` 0.47.0.

Separately: the mesh has been unable to complete a bind for several hours — reads and replay work, sends fail with "not connected", and the presence KV stream is write-frozen. That is why this is here rather than on the channel, and no review panel can be staffed until it is repaired.

davidfarah2003:
Correcting my earlier comment. I said the gate logic checked out and named only the environment and the artifact expiry. That was incomplete: an independent two-vendor review of this exact merge (`96ef332a`, tree `6c20ae01`) had already returned a unanimous block on three further grounds, and one of them is live on main right now.

**Main currently fails `check:docsbundle`.** Reproduced on a clean detached worktree at `96ef332a`: `node scripts/generate-docs-bundle.mjs` writes 40 pages and leaves `extensions/connector-core/src/docs-bundle.generated.ts` dirty by exactly 7 added lines. The committed bundle has zero `release-seat-bootstrap` entries even though `docs/release-seat-bootstrap.md` is present. That check ends in `git diff --exit-code`, so it fails today. `[skip ci]` is why nothing caught it at merge time — only Code Quality ran.

**The runbook's finish step is unsafe as written.** `docs/release-seat-bootstrap.md:59` says to rerun failed Changesets run `34072874048`. A rerun executes that run's immutable `609ce674` workflow, whose `ci:publish` is `seat-assemble-natives.mjs && pnpm publish -r --provenance` with no all-22 preflight. That is the command which already published `@cotal-ai/lang@0.47.0` and then stopped when seat 404'd — `pnpm publish -r` is sequential, so a rerun can publish further packages before a later trust refusal and deepen the split rather than close it. Triggering a fresh Changesets run from reviewed current main avoids this; rerunning that specific run does not.

**Bootstrap success is byte-idempotent but not provenance-idempotent.** Both the post-publish and the existing-version paths check only `.version` and `.dist.integrity`. Integrity binds the tarball bytes, but the HTTP 200 path performs no publish at all, so a matching tarball published outside this workflow is accepted as success. If the intent is that a successful run establishes this workflow's provenance, integrity alone does not show it.

The two items from my earlier comment are unaffected and still stand: `seat-bootstrap` is absent (404, only `copilot` exists), and artifact `10001064028` expires 2026-09-14T01:25:22Z.

davidfarah2003:
One practical note on who can land the rest of this.

A corrective for the first and third blockers above exists and is validated locally — it installs the all-22 trust census ahead of both recursive publish paths and replaces the "rerun run `34072874048`" step. It cannot be pushed from this machine: the credential here has `repo` but not `workflow` scope, so any push whose diff touches `.github/workflows/` is rejected outright.

    ! [remote rejected] (refusing to allow an OAuth App to create or update workflow
      `.github/workflows/bootstrap-seat.yml` without `workflow` scope)

That is scope-specific rather than a repo problem — #1363, pushed from the same credential minutes earlier, touches only a generated `.ts` file and went through. It also applies to the whole earlier body of work on this lane, which was workflow YAML throughout.

Worth knowing for two reasons: the work is not missing, it is undeliverable from here; and splitting the workflow half out to make the rest pushable would install half of a control, which is worse than none, so I have not done that.

Separately, no CI, Windows, Docs or Changesets run has been created on this repository since 01:23:44Z. Schedule and push events still work — a scheduled Mutation reproof started at 10:21Z — but no `pull_request` has produced a CI run since 01:05Z, including #1363, which carries no `[skip ci]`. GitHub Actions reports operational and the repo is public, so hosted minutes are not the cause. I could not read the org-level Actions policy (`admin:org` scope), so that remains unchecked.

davidfarah2003:
Update on the corrective described in my previous comment: it now carries two independent cross-vendor terminal approvals, so review is no longer what is holding it.

Both seats reproduced independently in their own clean detached worktrees at `e70bbb476a7370ace177aa0585f1733f75378633` / tree `b110676dd877fe2f98fa8762430f0adfd0dc41d8`, and each read the other's verdict and reconciled before terminalizing — `gpt-5.6-sol` and `grok-4.6`, both FINAL APPROVE. Both records note that the orchestrator authored the commit and that their verdict comes from independent reproduction rather than deference to it.

What the approvals explicitly do not cover, in the reviewers' own words: the workflow was never dispatched, live OIDC was never exercised, and registry bytes, latest tag and SLSA verification are unreachable while `@cotal-ai/seat@0.47.0` is unpublished. The expected `E404` is the correct public boundary today.

So the ask narrows to one thing: the branch cannot be pushed from here because the credential lacks `workflow` scope, and the commit touches `.github/workflows/`. Either a push by someone whose credential carries that scope, or granting it, is all that stands between a reviewed artifact and a PR.

Unrelated but still open: `seat-bootstrap` remains HTTP 404 with no `NPM_SEAT_BOOTSTRAP_TOKEN`, and no CI, Windows, Docs or Changesets run has been created on this repository since 01:23:44Z.

PR #1363, opened 2026-09-07
docs(bundle): regenerate for the seat bootstrap runbook

davidfarah2003:
`docs/release-seat-bootstrap.md` was added in #1361 without regenerating the docs bundle, so `extensions/connector-core/src/docs-bundle.generated.ts` is missing that page on main.

`check:docsbundle` (CI, `.github/workflows/ci.yml:79`) ends in `git diff --exit-code` over `docs/mcp-tools.md` and the generated bundle, so it fails on current main. Nothing caught it at merge time because #1361 carried `[skip ci]` — only CodeQL ran on `96ef332a`.

Reproduced on a clean detached worktree at `96ef332a`: `node scripts/generate-docs-bundle.mjs` reports 40 pages and leaves the bundle dirty by exactly 7 lines. The committed bundle has zero `release-seat-bootstrap` entries. `check:docs-voice` passes there (41 pages), so the bundle is the failing half.

This commit is generator output only — one file, the seven lines the generator produces. At this head the generator leaves the tree clean.


davidfarah2003:
Completing the measurement, since `check:docsbundle` is composite and I had only verified two thirds of it.

    check:docsbundle = check:docs-voice && gen:tooldocs && gen:docsbundle
                       && git diff --exit-code -- docs/mcp-tools.md
                                                  extensions/connector-core/src/docs-bundle.generated.ts

- `check:docs-voice` — passes, 41 pages.
- `gen:tooldocs` — runs clean and writes `docs/mcp-tools.md` (22 tools) with **zero drift**. Both of its inputs, `docs/mcp-tools.md` and `extensions/connector-core/src/tool-specs.ts`, are byte-identical between `96ef332a` and this branch, so that result holds for main as well.
- `gen:docsbundle` — the only failing half, and what this commit fixes.

So the generated bundle was the sole cause and this is a complete fix for the check rather than a partial one.

Worth flagging separately: no CI, Windows, Docs or Changesets run has been created on this repository since 01:23:44Z, so this PR carries no CI result. GitHub Actions reports operational and the repo is public, so hosted minutes are not the cause; I could not read the org-level Actions policy to rule that out.

davidfarah2003:
This PR carries no CI result and cannot get one right now. Its head `7bb1281f4a2592ea11d774cc4cd171183b0b7661` has a single check suite, GitHub's own `Code Quality` app, and the two CodeQL `Analyze` checks are the whole of `statusCheckRollup`. `GET /actions/runs?head_sha=7bb1281f...` returns `total_count: 1`, that same app run.

That is not specific to this branch. This repository has created no Actions run for a `push` or a `pull_request` event since `2026-09-07T01:23:44Z`, while `schedule` events on it still run and a sibling repo in the org still gets `pull_request` runs. Measured and filed as #1364.

So the `CI` workflow, which is where `pnpm check:docsbundle` runs at `.github/workflows/ci.yml:79`, has not been exercised on this fix, and would not be exercised on `main` after a merge either.

What the change does have behind it, from a clean detached worktree at `96ef332a`:

- the break reproduces there, as a 7-line diff in `extensions/connector-core/src/docs-bundle.generated.ts` and zero `release-seat-bootstrap` entries committed;
- `check:docs-voice` passes over 41 pages and `gen:tooldocs` produces no drift at that base, so the bundle is the only cause and this one file is the whole fix.

It still has no independent review. The mesh this fleet uses for review panels is not accepting sends (`state: connecting`, `connected: false`), so I cannot staff one, and I am not going to merge an unreviewed head to work around that.

davidfarah2003:
Merge constraint on this PR, measured after the comment above.

`main` has zero pending changesets, so `changesets/action` takes its publish branch on the next push, and `ci:publish` still has no gate that refuses before the first write. Merging this normally would therefore start a sequential 0.47.0 publish across 21 packages against a registry that is already split. Filed as #1365.

So this should either wait for the fence, or be merged with `[skip ci]` in the merge commit, accepting that the `check:docsbundle` gate it repairs will not have run on it. I would take the second only with that stated plainly, not silently.

davidfarah2003:
Heads-up: this PR now collides with #1368, whose head `04fc02cbc121` fixes the same defect in the opposite direction by deleting `docs/release-seat-bootstrap.md` rather than adding its bundle citation.

Both changes are individually correct. Together they are not:

```
main today       page=1  refs=0    INCONSISTENT   (the bug both address)
#1363 alone      page=1  refs=1    consistent
#1368 alone      page=0  refs=0    consistent
both merged      page=0  refs=1    INCONSISTENT   (bundle cites a deleted page)
```

`check:docsbundle` regenerates the bundle from `docs/` and runs `git diff --exit-code`, so a citation to a page that no longer exists fails it exactly as a missing citation does.

`git merge-tree --write-tree` returns rc 0 in **both** orders and produces the identical tree `4daa067ff62fee6e411c5ae75dce733fa5220e90`, so nothing will flag this at merge time. Only one of the two can land.

Not a request to close this — that depends on whether the runbook page is meant to survive, which is #1368's open question given that recovery is still at 17 of 22 packages. Flagging it so the second merge does not quietly re-break the bundle.

davidfarah2003:
This PR and #1368 are opposite resolutions of the same defect, and both are mergeable right now.
This one keeps `docs/release-seat-bootstrap.md` and adds its bundle entry (40 slugs, 1 ref). #1368
deletes the page and the `bootstrap-seat.yml` workflow, leaving 39 slugs and 0 refs.

Measured with `git merge-tree --write-tree` at head `7bb1281f4a25` against `main` `da4d3cb37c9a`:
both merge orders converge on the identical tree `70ebea21263448f2139db424368fcd7a6775f86c`, where
the bundle cites `release-seat-bootstrap` and the page is gone. No conflict is reported at any step,
and `check:docsbundle` diffs for exact equality, so that is red on `main`.

Each head is self-consistent; the breakage exists only in the merged result. So one of these should
land and the other should be closed rather than held. Full working and the merge-order table are on
#1368 (comment 5577021307).

davidfarah2003:
Correcting something I implied earlier: **this PR alone does not return `main` to green.** It fixes
`unit`, but `main` is red on a second required check for the same underlying cause.

From run `34152956010`'s job logs at the head merged as `da4d3cb3`:

- `unit` — `pnpm check:docsbundle`: generator writes 40 pages, committed bundle has 39. This PR
  fixes that.
- `smoke (shard 3/4)` — `pnpm smoke:workflow-concurrency`, cell A:
  `✗ FAIL: exactly 7 workflow files (update deliberately if you added one) { found: 8 }`. The 8th is
  `bootstrap-seat.yml`, added by #1361 without bumping `EXPECTED_WORKFLOWS`. This PR does not touch
  that file, so shard 3/4 stays red.

Both gaps come from #1361 adding the workflow and its runbook without updating the two counters that
track them. The unpushed `e70bbb47` closes both (bundle → 40 pages, `EXPECTED_WORKFLOWS` → 8) while
keeping the mechanism, so it subsumes this PR. #1368 closes both by deleting the workflow and the
page. Detail and the decision matrix are on #1365.

davidfarah2003:
Superseded by #1368. The completed bootstrap runbook was removed there, so regenerating the bundle to include that deleted page would make the docs bundle inconsistent.

davidfarah2003:
**This PR should now be closed rather than merged.** #1368 merged at 2026-09-08T00:34:41Z as
`d956e578265282927356dcdd4f576839a30f05fd`, and it resolved this defect in the opposite direction:
it deleted `docs/release-seat-bootstrap.md` and `.github/workflows/bootstrap-seat.yml` and left the
bundle at 39 slugs, which is self-consistent.

This PR adds a bundle entry for a page that no longer exists. Merging it now produces the tree I
measured earlier, `70ebea21263448f2139db424368fcd7a6775f86c` — bundle citing
`release-seat-bootstrap`, page absent — and `check:docsbundle` regenerates from `docs/` and diffs
for exact equality, so that is `unit` red on `main`. `git merge-tree` reports no conflict, so
nothing mechanical will stop it.

The defect this PR was opened for is genuinely fixed on `main`; it just is not fixed by this diff
any more.

Issue #1365, opened 2026-09-07
release: any push to main now starts an unfenced non-atomic publish of 0.47.0, and only the Actions outage is stopping it


davidfarah2003:
`main` is currently one non-`[skip ci]` merge away from an unfenced, non-atomic recursive publish of 0.47.0, and the only thing holding it is the unrelated Actions fault in #1364.

## The measurement

On `96ef332a`:

- `.github/workflows/changesets.yml` triggers `on: push: branches: [main]`, and its `version` job runs when `github.event_name == 'push'`.
- That job is `changesets/action@v1` with `version: pnpm ci:version` and `publish: pnpm ci:publish`. With no changesets pending it takes the publish branch.
- `.changeset/` on `96ef332a` holds **zero** changeset files, so that is the branch it takes.
- `ci:publish` is `node scripts/seat-assemble-natives.mjs && pnpm publish -r --provenance --access=public --no-git-checks`. Byte-identical at `609ce674` and at `96ef332a`. No existence preflight, no trust census, nothing that fails before the first write.
- This repo pins `pnpm@11.1.2`, where `publish -r` is per-package and sequential. `--batch` all-or-nothing exists only from 11.7.0.

Version skew right now: every workspace package is `0.47.0` in tree; npm has `@cotal-ai/lang@0.47.0`, `cotal-ai` / `@cotal-ai/core` / `@cotal-ai/cli` / `@cotal-ai/workspace` at `0.46.0`, and `@cotal-ai/seat` is HTTP 404.

So the next push to `main` without `[skip ci]` walks 21 packages one at a time against a registry that is already split, with no gate that can refuse before the first one lands. That is the shape that produced the split: run `34072874048` published `@cotal-ai/lang@0.47.0`, then `@cotal-ai/seat` failed OIDC 404 and PUT 404, and the packages after it were never attempted.

The `snapshot` job has the same exposure, with neither preflight nor census, though it is dispatch-only.

## Why it has not fired

Two accidents, in sequence, neither of them a control:

1. The merge of #1361 carried `[skip ci]`, so Changesets did not run on it.
2. Since `2026-09-07T01:23:44Z` this repo creates no Actions run for any `push` or `pull_request` event at all (#1364).

**Restoring Actions dispatch therefore re-arms this.** Whoever fixes #1364 should know that the first ordinary merge afterwards starts a recursive publish.

## The runbook points straight at it

[...]

Issue #1364, opened 2026-09-07
ci: no Actions run has been created for a push or pull_request event since 2026-09-07T01:23:44Z, while schedule still fires


davidfarah2003:
Since `2026-09-07T01:23:44Z` this repository has not created a single Actions run for a `push` or a `pull_request` event. `schedule` events on the same repository still run, so Actions is enabled and the workflow files load.

## What still runs, and what stopped

Last `push` runs, all on `609ce674` at `01:23:44Z`: CI, Windows, Changesets, Mutation reproof.
Last `pull_request` runs, all on `fix/mutation-reproof-shards` at `01:05:26Z`: CI, Windows, Mutation reproof.

Since then:

| Event | Where | Actions runs created |
| --- | --- | --- |
| push to `main` (merge of #1361, `96ef332a`, 05:37:42Z) | `main` | none |
| pull_request #1360 | `refs/pull/1360/head` | none |
| pull_request #1361 | `refs/pull/1361/head` | none |
| pull_request #1362 | `refs/pull/1362/head` | none |
| pull_request #1363 (`7bb1281f`) | `refs/pull/1363/head` | none |
| schedule (Mutation reproof) | `main`, 10:21:49Z | created, still in progress |
| schedule (Installer) | `main`, 12:48:12Z | created, success |

Each of those four PRs did get a run named `Code Quality: PR #N`, but its `event` is `dynamic` and its workflow path is `dynamic/github-code-quality/codeql`. That is GitHub's own app, not a file in `.github/workflows`.

Check suites tell the same story. `609ce674` has five `github-actions` check suites created at `01:23:44Z`. `96ef332a` has one at its push time (the Code Quality app, 2 runs) and then only the two schedule suites. `7bb1281f` has one, the Code Quality app.

## What this rules out

- `GET /repos/Cotal-AI/Cotal/actions/permissions` returns `{"enabled": true, "allowed_actions": "all", "sha_pinning_required": false}`.
- All eight in-tree workflows report `state: active`.
- The repository is not archived, not disabled, and public.
- `.github/workflows` is unchanged between `609ce674` and `96ef332a` apart from the added `bootstrap-seat.yml` (+154). `ci.yml` still declares `on: pull_request:` and `push: branches: [main]`.
- githubstatus.com reports Actions, Webhooks, Git Operations, API Requests and Pull Requests all operational.
[...]

PR #1357, opened 2026-09-06
ci: shard changed-fixture mutation reproof

davidfarah2003:
## Scope

Run changed-fixture mutation reproof on twelve isolated runners using the existing deterministic path hash. Keep the stable `changed` aggregate, which waits for all shards and accepts only success. Disable matrix fail-fast so a failed shard does not cancel the remaining proofs.

Keep the 145-minute proof-step limit, 150-minute job limit, and per-shard elapsed-time reporting. The scheduled/manual full-sweep job is unchanged. Report empty-shard assignments without claiming the whole diff selected nothing.

Extend the smoke to run the shipped selector over a 24-fixture control, unsharded and across all twelve shards. Require full coverage without duplicate assignments, exercise every shard, and reject invalid shard arguments. Parse the actual aggregate configuration and execute its shell against success, failure, cancellation, and skipping.

Sanitize the default environments passed to the smoke's Git calls, aggregate probes, and scanner. Observe the scanner's own environment at Node startup as well as the fixture child's environment. Retain main's deliberate environment overrides and normalized-baseline attribution checks. The current-main integration preserves the failure-signature module and its proof integration.

This addresses the serial reproof timeout encountered by #1350 and #1220.

## Verification

At `fcfa4d0a1`, the merged reproof smoke passes 75 checks. All 15 sharding mutations and 20 selector/attribution mutations are detected at their named assertions. Source restoration is clean. The ambient-environment census and mutation-anchor census pass, with no dead or ambiguous anchors.

The full-sweep job is byte-identical to the current main base. The failure-signature helper and mutation-proof module also match that base. Hosted CI, Windows, code quality, and mutation reproof passed at this head. All 12 shards and the aggregate succeeded, with 35 mutations detected. Independent review remains pending.

## Limits

The aggregate controls establish shell rejection and checked YAML wiring. They do not execute GitHub's live skipped-job scheduling semantics. The hosted run selected two fixtures, assigned to shards 0 and 2; ten shards had no assigned work. Its longest shard job took 623 seconds. This establishes completion for this diff, not for the heavier workloads in #1350 or #1220. Deterministic path hashing does not balance runtime, so a heavier shard can still hit the timeout and fail the gate.

Workflow concurrency is unchanged; cancellation of superseded runs is tracked separately in #1382.

PR #1366, opened 2026-09-07
npm: point package homepage at cotal.ai

Lanzelot1:
The npm pages for `cotal-ai` and `@cotal-ai/cli` link to the GitHub readme as their homepage. Point them at https://cotal.ai, which is what npm's `homepage` field is for (the repository stays in `repository`). Takes effect on the next publish.

Also how agent-readiness scanners verify the package belongs to the domain.

### DIFF
--- a/bin/package.json
+++ b/bin/package.json
@@ -19,7 +19,7 @@
     "mcp",
     "a2a"
   ],
-  "homepage": "https://github.com/Cotal-AI/Cotal#readme",
+  "homepage": "https://cotal.ai",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/Cotal-AI/Cotal.git",
--- a/implementations/cli/package.json
+++ b/implementations/cli/package.json
@@ -2,6 +2,7 @@
   "name": "@cotal-ai/cli",
   "description": "Cotal mesh CLI: up, join, watch, console, spawn, mint, channels, history, ext.",
   "version": "0.47.0",
+  "homepage": "https://cotal.ai",
   "license": "Apache-2.0",
   "repository": {
     "type": "git",


davidfarah2003:
Nothing wrong with this change. Flagging repository state that makes merging it to `main` risky right now, which is not visible from the PR.

`main` currently has zero changeset files, so `changesets/action` takes its publish branch rather than its version-PR branch on the next push. `ci:publish` on `main` is still:

```
node scripts/seat-assemble-natives.mjs && pnpm publish -r --provenance --access=public --no-git-checks
```

That is byte-identical to the command in failed run `34072874048`, which published `@cotal-ai/lang@0.47.0` and then died on `@cotal-ai/seat`. `pnpm@11.1.2` publishes `-r` sequentially, and nothing in that command refuses before the first write. The registry is currently split: `@cotal-ai/lang` is 0.47.0, sampled siblings are 0.46.0, `@cotal-ai/seat` is HTTP 404. Detail in #1365.

What changed in the last two minutes is the reason for writing this. Per #1364, no Actions run had been created for a `push` or `pull_request` event since 01:23:44Z. This PR's runs at 17:23:49Z show `pull_request` dispatch is working again. `push` dispatch is a separate question and is currently unknown:

- the last `push` run of any kind was 01:23:44Z
- the merge that moved `main` to `96ef332a` at 05:37:42Z produced only a `dynamic` Code Quality run and no `push` runs at all, which is the #1364 symptom
- `main` has not moved since, so `push` dispatch has not been exercised in either direction

There is also no safe way to test it. `CI`, `Windows` and `Changesets` are all `push: branches: [main]`, so pushing a feature branch produces no `push` run by design and tells you nothing. The only probe for `push` dispatch is a push to `main`, and that probe is the hazard.

So merging this either does nothing, or starts a non-atomic recursive publish of 0.47.0 across ~21 packages against an already-split registry, and which one is not knowable in advance.

Suggested handling, in preference order:

1. Hold this until the publish fence lands. The fence exists and is reviewed; it is blocked on delivery, not on authorship (see #1365).
2. If it needs to land sooner, merge with `[skip ci]` in the merge commit subject and say so explicitly, accepting that the merge then gets no CI record.

The same applies to #1363 and #1357, which are both currently mergeable. This is a property of the repository state, not of any of these three changes.

PR #1367, opened 2026-09-07
skills: cotal-mesh skill and agent-plugins manifest

Lanzelot1:
Two files so the Cotal skill can be discovered and installed from this repo:

- `skills/cotal-mesh/SKILL.md`: the Agent Skills (agentskills.io) skill for putting an agent on a Cotal mesh and coordinating with peers. Same content as https://cotal.ai/skills/cotal/SKILL.md, which is listed with its digest in https://cotal.ai/.well-known/agent-skills/index.json. Lets `npx skills add Cotal-AI/Cotal` work and the skill appear on skills.sh under this org.
- `plugin.json`: the agent-plugins.org v1.0.0 manifest (`$schema`, `name`, description, author, links). Skills are discovered from `skills/`; the existing Claude Code marketplace in `.claude-plugin/` is untouched.

No code paths change. Suggest merging with `[skip ci]` like #1366 while the publish fence (#1365) is pending.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "https://agent-plugins.org/schemas/1.0.0/plugin.schema.json",
+  "name": "cotal",
+  "version": "1.0.0",
+  "description": "COTAL: the open protocol that lets AI agents from any vendor work as one team. Skills for putting an agent on a Cotal mesh and coordinating with peers; the cotal.ai product MCP server lives at https://cotal.ai/mcp.",
+  "author": { "name": "COTAL", "email": "hello@cotal.ai", "url": "https://cotal.ai" },
+  "homepage": "https://cotal.ai",
+  "repository": "https://github.com/Cotal-AI/Cotal",
+  "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


davidfarah2003:
Short pointer, not a repeat: the same `main`-state hazard that applied to #1366 applies here, and nothing about it has changed. `main` still has zero changeset files and an unfenced `ci:publish`, so a merge without `[skip ci]` can start a sequential recursive publish of 0.47.0 against a registry that is already split (`@cotal-ai/lang` 0.47.0, siblings 0.46.0, `@cotal-ai/seat` 404). Background is #1365.

The `[skip ci]` approach used on #1366 works and is the low-friction option while the fence is undelivered. One correction to what I wrote on #1366 that runs in the unsafe direction: I later claimed on #1364 that `push` dispatch was confirmed broken, and that was wrong. Every `main` commit since 01:23:44Z carries `[skip ci]`, which explains the missing runs by itself, so `push` dispatch is untested rather than known-dead and cannot be relied on to absorb a merge.

Nothing here is a review of this change.

PR #1368, opened 2026-09-07
fix(release): build workspace dependencies before publishing

davidfarah2003:
## Scope

Build the workspace before assembling seat native artifacts and recursively publishing packages. A partial release retry skips already-published packages and their prepublish hooks; the remaining packages still need those local dependencies built.

Preserve trusted publishing, provenance and the two-architecture native assembly step. No package versions or runtime code change.

Retire the unused emergency seat bootstrap workflow and its temporary runbook. The real seat 0.47.0 package is now published through Changesets OIDC; no bootstrap workflow runs exist. Removing these files also resolves the inherited workflow-count and generated-docs failures without weakening either check.

## Evidence

Release run 34072874048 attempt 2 successfully published seat and other packages, then failed building runtime with TS2307 for @cotal-ai/lang. Lang had already published in attempt 1 and was skipped on retry.

The new contract assertion fails against the previous publish command. With the fix, a fresh workspace build passes, the release contract smoke passes 13/13, and all 10 mutations are killed. Fixture anchors and clean restoration pass. After bootstrap cleanup, the workflow-concurrency smoke passes 27/27 and the generated docs check passes.

This repairs the build-order blocker encountered while recovering #1359. Completion still requires the trusted workflow to publish the remaining 0.47.0 packages and a full registry closure check.


davidfarah2003:
Two measured findings on this PR, one blocking it and one about what happens after it lands.

**1. The red `unit` check is a pre-existing defect, and #1363 is its fix.**

The failing step is `Docs bundle in sync`. This branch carries `docs/release-seat-bootstrap.md` but its generated bundle does not cite it:

```
#1368 head   docs page present, bundle refs to release-seat-bootstrap: 0
#1363 head   bundle refs to release-seat-bootstrap: 1
```

The content the check demands in the job log is byte-for-byte the entry #1363 adds. So this is not caused by the build-order change; it is inherited from main, and #1363 (one file, +7/-0, MERGEABLE/CLEAN) is exactly the regeneration that clears it. This PR cannot go green until that lands or the bundle is regenerated here.

**2. Re-running `34072874048` cannot complete the release, and the runbook's step 4 says to do that.**

A re-run does not create a new run; it adds an attempt to the existing one and is bound to the workflow and tree of the sha the run was created for. Measured:

```
run 34072874048   event=push   run_attempt=2
  created_at      2026-09-07T01:23:44Z
  run_started_at  2026-09-07T21:53:09Z
  head_sha        609ce674235e32517a7d726585d2f736d629a77e
  failed job      version, step "Create versions and publish"
```

That tree is fixed at `609ce674`, so no attempt on this run can ever contain the change in this PR. Attempt 3 would re-run the same build order and fail the same way. It is also permanently ungated: at `609ce674`, `ci:publish` is bare `seat-assemble-natives.mjs && pnpm publish -r`, and `scripts/preflight-npm-existence.mjs` and `scripts/verify-npm-publisher-readiness.mjs` do not exist in that tree at all.

The runbook now on main (`docs/release-seat-bootstrap.md`, "Finish a successful recovery", step 4) says to rerun `34072874048`. That instruction predates this build-order bug and can no longer complete the release. Completion has to come from a **new** push run on a main that contains this fix.

**What that implies for landing order.** Main has zero changeset files, so `changesets/action` takes its publish branch on the next push run, and all five unpublished packages are already at `0.47.0` in-tree on main (`cotal-ai`, `runtime`, `cli`, `manager`, `web`). So the next merge to main *without* `[skip ci]` will attempt exactly the remaining five. That is the right thing to happen once this fix is in, and the wrong thing to happen before it:

- #1363 merged without `[skip ci]` would fire a publish attempt on a main that still has the build-order bug, reproducing the failure.
- This PR merged after #1363, without `[skip ci]`, is the completion path.

**Current registry state, for the record:** 17 of 22 at 0.47.0. Missing: `cotal-ai`, `@cotal-ai/runtime`, `@cotal-ai/cli`, `@cotal-ai/manager`, `@cotal-ai/web`. No published 0.47.0 package depends on a missing one (17 inspected, 0 unread), and `cotal-ai@latest` is still 0.46.0 with all eight first-party dependencies pinned at 0.46.0, so a fresh install is self-consistent today.

Unrelated and minor: the PR body carries a "Generated with Claude Code" footer, which AGENTS.md prohibits in PR text.

davidfarah2003:
Follow-up: the second red check has the same shape as the first. **Neither failure is caused by this PR.** Both are main's debt, surfacing here because this is the first pull request to run CI against a main that contains the last five commits.

This PR touches three files, none of them a workflow and none the docs bundle:

```
bin/smoke/mutations/seat-native-ci.json  +10/-2
bin/smoke/seat-native-ci.smoke.ts        +6/-0
package.json                             +1/-1
```

**`smoke (shard 3/4)` fails at `pnpm smoke:workflow-concurrency`, 28 passed 1 failed:**

```
✗ FAIL: exactly 7 workflow files (update deliberately if you added one)
  found: 8
  files: [bootstrap-seat.yml, changesets.yml, ci.yml, docs.yml,
          github-releases-to-discord.yml, installer.yml, mutation-reproof.yml, windows.yml]
```

`EXPECTED_WORKFLOWS = 7` in `bin/smoke/workflow-concurrency.smoke.ts` on main, and main has 8 workflow files. The eighth is `bootstrap-seat.yml`, added by `e6ad61cb7 fix(release): bootstrap seat publication [skip ci]`.

**Why it was never caught.** Measured push runs on each of the five commits above `609ce674`:

```
e6ad61cb 0    96cca932 0    96ef332a 0    9930a08c 0    da4d3cb3 0
```

All five carry `[skip ci]`, so no push CI has run on main since 01:23:44Z. A counted-inventory sentinel like this one can only fire when something runs it, and nothing has. The same is true of the stale docs bundle: the runbook page was added by one of those commits and the bundle was never regenerated. Both defects have been sitting on main, invisible, and they are landing on the PR that repairs the release.

**One more consequence worth stating, because it affects how this run should be read.** The shard is fail-fast, and it reported it plainly:

```
NEVER RAN — 25 of 130 planned smoke(s) in this shard's partition did not start,
because the shard stopped at the failure above
```

So shard 3 is not evidence about anything after `workflow-concurrency`. A green shard 3 on the next attempt would be the first information about those 25.

**What unblocks this PR**, neither of which is a change to its own logic:

1. Regenerate the docs bundle — that is exactly #1363, one file, +7/-0, currently MERGEABLE/CLEAN.
2. Set `EXPECTED_WORKFLOWS = 8` deliberately, which is what the assertion's own message asks for.

On (2), one thing to decide rather than assume: the runbook on main (`docs/release-seat-bootstrap.md`, "Finish a successful recovery", step 3) says `bootstrap-seat.yml` is removed once recovery completes. If that still holds, the constant goes to 8 now and back to 7 at cleanup, and it would be worth a comment in the smoke saying so, so the next person does not read the bump as drift.

davidfarah2003:
CI has finished on `bd6dc0e4`. Three red checks, **two root causes, neither attributable to this PR**, and the jobs that actually exercise its change are all green.

Full job list:

```
success  seat pack (assembled)              success  smoke (shard 0/4)
success  seat native (linux-x64)            success  smoke (shard 1/4)
success  seat native (linux-arm64)          success  smoke (shard 2/4)
success  seat load (linux-arm64, same artifact)
success  live
failure  unit                        <- Docs bundle in sync   (main's stale bundle, = #1363)
failure  smoke (shard 3/4)           <- workflow-concurrency  (main's 8th workflow file)
failure  ci-ok (live outcome classified)
```

The third red is not an independent defect. `ci-ok` is the aggregate gate and its log prints the roll-up it is failing on:

```
unit=failure smoke=failure live=success seat-pack=success seat-linux-arm64=success
##[error]Process completed with exit code 1.
```

So it goes green automatically once the other two do. Worth stating explicitly so nobody hunts for a third cause.

The distribution is informative: `seat pack`, both `seat native` builds, `seat load` and `live` all pass, and three of four smoke shards pass. Those are the jobs covering the build-order change this PR makes. The two failures are in a docs-bundle check and a workflow-file census, neither of which this PR touches — it changes `bin/smoke/mutations/seat-native-ci.json`, `bin/smoke/seat-native-ci.smoke.ts` and `package.json`.

One caveat on reading shard 3 as evidence, repeated because it still applies: it is fail-fast and stopped at `workflow-concurrency`, so 25 of its 130 planned smokes never started. Shard 3 currently says nothing about those, and the next attempt would be the first information about them.

Unchanged since my earlier comment: the two fixes are to regenerate the docs bundle (#1363) and to set `EXPECTED_WORKFLOWS = 8` deliberately, as that assertion's own message asks.

davidfarah2003:
New head `04fc02cbc121` resolves both inherited failures at the source rather than by adjusting the checks, and the build fix is intact and strengthened (`ci:publish` is now `pnpm build && node scripts/seat-assemble-natives.mjs && pnpm publish -r …`). Measured at that head: `bootstrap-seat.yml` removed so the tree has 7 workflow files against `EXPECTED_WORKFLOWS = 7`, and `docs/release-seat-bootstrap.md` removed so the bundle's 0 references are correct. Both are self-consistent.

**But this now collides with #1363, and git will not warn about it.**

The two PRs are opposite fixes for the same defect. `check:docsbundle` is `pnpm gen:docsbundle && git diff --exit-code -- … docs-bundle.generated.ts`, so it regenerates the bundle from `docs/` and requires exact equality. A citation to a page that does not exist fails it just as a missing citation does.

Consistency of each tree, by page count vs bundle reference count:

```
main today       page=1  refs=0    INCONSISTENT   (the live bug)
#1363 alone      page=1  refs=1    consistent     (adds the citation)
#1368 alone      page=0  refs=0    consistent     (deletes the page)
both merged      page=0  refs=1    INCONSISTENT   (citation to a deleted page)
```

Merge order does not save it. `git merge-tree --write-tree` in both directions returns rc 0 and the **identical** tree `4daa067ff62fee6e411c5ae75dce733fa5220e90`, with the page absent and the bundle still citing it. A clean merge-tree is the absence of a textual conflict, not agreement about the result.

So only one of these two can land. If this PR merges as it stands, #1363 must be closed or emptied rather than merged after it; if #1363 merges first, this PR needs to drop the bundle citation as well as the page.

**One thing worth a deliberate decision rather than my assumption.** Deleting the runbook now is step 3 of that runbook's own "Finish a successful recovery", but recovery is not finished: 17 of 22 packages are at 0.47.0 and `cotal-ai`, `runtime`, `cli`, `manager` and `web` are not. The page also carries the per-attempt cleanup instructions (revoke the granular token, delete the environment secret) and the completion steps. Removing it while five packages are outstanding removes the operator's written procedure mid-recovery. The bootstrap workflow itself has arguably done its job, since `@cotal-ai/seat` now exists, so this may well be intended; I am flagging it rather than arguing against it.

**And a note on scope, since this PR is the completion path.** `ci:publish` here is still `pnpm build && seat-assemble-natives && pnpm publish -r` with no existence preflight and no trust census. This fixes the build order, not the fail-before-first-publish gap tracked in #1365. Merging it without `[skip ci]` will create a push run that attempts exactly the five remaining packages, which is the intended completion, but it will do so ungated.

davidfarah2003:
Head `bc0f909834375ae63796bf2464f023345c6d9287`, measured against `main` `da4d3cb37c9a`. Two things
here need a decision before this merges, and neither is visible from the diff stat.

### 1. This deletes the seat bootstrap mechanism, and collides with #1363 in both orders

The change removes `.github/workflows/bootstrap-seat.yml` (-154) and `docs/release-seat-bootstrap.md`
(-62). #1363 does the opposite: it keeps the page and adds its entry to the generated bundle. They
are two opposite resolutions of the same defect, and both are `MERGEABLE`/`CLEAN` right now.

`git merge-tree --write-tree`, no conflict reported at any step:

```
main + #1363 -> 37b13d51   page PRESENT, 40 slugs, 1 bundle ref    consistent
      + #1368 -> 70ebea21   page DELETED, 40 slugs, 1 bundle ref    INCONSISTENT

main + #1368 -> 9aa52523   page DELETED, 39 slugs, 0 bundle refs   consistent
      + #1363 -> 70ebea21   page DELETED, 40 slugs, 1 bundle ref    INCONSISTENT
```

Identical result tree `70ebea21263448f2139db424368fcd7a6775f86c` either way: the generated bundle
cites `release-seat-bootstrap` and the page it cites is gone. `check:docsbundle` regenerates from
`docs/` and then diffs for exact equality, so that is red on `main`. Nothing stops it — a clean
merge-tree is the absence of a text conflict, not agreement about the change.

Each head is fine on its own; `docs` passes here at `bc0f9098`. The breakage only exists in the
merged result. So one of these should land and the other should be **closed**, not held.

### 2. This is the publish trigger, and it carries no fence

`main` has zero changeset files, so `changesets/action` takes its publish branch on the next push
run, and `cotal-ai`, `@cotal-ai/runtime`, `@cotal-ai/cli`, `@cotal-ai/manager`, `@cotal-ai/web` are
already 0.47.0 in-tree. A merge here without `[skip ci]` is therefore the only path that completes
those five — and it fires them through:

```
pnpm build && node scripts/seat-assemble-natives.mjs && pnpm publish -r --provenance --access=public --no-git-checks
```

The `pnpm build` is right and it fixes the real cause: attempt 2 of run `34072874048` died building
`runtime` with TS2307 for `@cotal-ai/lang`, because lang published in attempt 1 and was skipped on
retry. But there is no existence preflight and no trusted-publisher gate in front of the write, so
this gets **further** than attempt 2 did with nothing in front of it. That is #1365 firing, not #1365
fenced.

The only artifact carrying both gates is `e70bbb47` (unpushed, `fix/1361-bootstrap-followup`):

```
node scripts/preflight-npm-existence.mjs && node scripts/verify-npm-publisher-readiness.mjs --automatic && node scripts/seat-assemble-natives.mjs && pnpm publish -r ...
```

and it keeps the workflow and runbook this PR deletes. They are not sequenceable in either
direction, so this is a third position in the #1359 design question — remove the mechanism — and it
is the one currently wired to the trigger. Worth stating explicitly rather than settling by
merge order.

Registry is unchanged: 17 of 22 at 0.47.0, the five above still 0.46.0. `@cotal-ai/seat@0.47.0` is
timestamped 2026-09-07T21:53:36Z, inside attempt 2, so it was always one of the 17.

Do not re-run `34072874048` to finish this. A re-run is bound to the tree of the sha the run was
created for (`609ce674`), where the build fix in this PR does not exist.

<sub>Method note: the bundle is 1.1 MB and the contents API returns empty inline content above 1 MB,
so `.content | base64 -d | grep -c` reports `0` for every ref including the one that is `1`. Read it
with `Accept: application/vnd.github.raw` and check the byte count before trusting the count.</sub>

davidfarah2003:
CI update on `bc0f909834375ae63796bf2464f023345c6d9287`, since my earlier comments described red
checks on older heads and that is no longer the state.

`unit` and `smoke (shard 3/4)` — the two checks `main` is currently red on — both **pass** here, and
shard 3/4 passes with full coverage rather than by stopping early:

```
smoke:ci shard 3/4 — 130 of 531 smokes:
✓ exactly 7 workflow files (update deliberately if you added one)
WORKFLOW CONCURRENCY SMOKE OK ✅  (27 passed, 0 failed)
✓ smoke:ci shard 3/4 passed (130 smokes)
```

130 announced against 130 planned, no never-ran line. That matters because the same shard on `main`
(run `34152956010`) failed at cell A with `{ found: 8 }` and left **25 of 130 never started**,
including `smoke:presence-watch-stall`, `smoke:transport-liveness:broker` and
`smoke:failed-bind-cleanup`. Those ran and passed here. The cell count drops 29 → 27 exactly because
the two `bootstrap-seat.yml` cells in section C go away with the file.

Shards 1/4 and 2/4 are still in progress; everything else on the run is green.

One asymmetry worth stating plainly for whoever takes the design decision, because it cuts against
the rest of what I have posted: **#1368 has demonstrated this green; `e70bbb47` has not.** Both
repair the same two checks, and I verified by reading its tree that `e70bbb47` bumps
`EXPECTED_WORKFLOWS` 7 → 8 to match the 8 files it keeps, and regenerates the bundle to 40 pages. But
it is unpushed, so no CI has ever run against it. Reasoned-green and observed-green are not the same
evidence, and the gap is not `e70bbb47`'s fault — it cannot be tested until it is pushed.

davidfarah2003:
Correcting my own earlier framing of this PR. I described it as a competing bootstrap design that
"deletes the mechanism". That was wrong in an important way: **this is step 3 of the recovery
runbook's own sequence**, and it performs that step as written — remove `bootstrap-seat.yml` and the
runbook, regenerate the docs bundle in the same commit. The deletion is intended.

What still needs a decision is that step 3 has preconditions which are measurably unmet:

- Step 3 says **"preserve the guarded Changesets code"**. There is none on `main`: its `ci:publish`
  is `node scripts/seat-assemble-natives.mjs && pnpm publish -r …`, no gates. The guarded version
  is only in the unpushed `e70bbb47`. So this step assumes a fence that never landed.
- Step 3 says **`[skip ci]`** and "do not let adoption or removal trigger Changesets". Merging this
  without `[skip ci]` is that trigger — and it is simultaneously the only route that publishes the
  five packages still at 0.46.0. Worth deciding deliberately rather than settling it with the merge
  button.
- Step 4 requires `NPM_TRUSTED_PUBLISHERS_SHA256` to equal
  `368e712eb3c23c2b30b67a215b7873c1124ba1658542c80ad54d8d99fd77257d` and says not to dispatch
  otherwise. Repo Actions variables are `total_count=0`, so it is absent.

None of that is an argument against this change. It is an argument that the step it implements is
the third of seven, and steps 1, 2 and 4 have not been shown done. Detail on #1365 (comment
5577189759).

CI here remains green on both checks `main` is red on; shards 1/4 and 2/4 were still running at
last look.

PR #1390, opened 2026-09-08
fix(test): require mutation suite source metadata

davidfarah2003:
## Summary

- require every mutation config to declare a non-empty `suite` array of repository-relative source paths
- validate declared suite paths centrally and migrate all mutation configs to the canonical representation
- select changed-set reproof fixtures when any declared source changes, including both sides of renames and deleted sources
- preserve config-path and mutation-target selection across the fixture census, reproof, coverage, proof, and language smoke consumers
- add focused mutation controls for missing, malformed, multi-source, renamed, and deleted suite metadata

Closes #1346

## Validation

- `pnpm smoke:mutation-fixtures` (364 configs, 1891 anchors)
- `pnpm smoke:mutation-reproof` (81/81)
- `pnpm --dir packages/lang exec tsx smoke/transform.smoke.ts` (124 cells)
- `pnpm --dir packages/lang exec tsx smoke/differential.smoke.ts` (212/212)
- fixture metadata mutation proof (12/12 killed)
- reproof metadata mutation proof split into bounded subsets with independent OOM brackets
  - 0-4: 5/5 killed
  - 5-9: 5/5 killed
  - 10-14: 5/5 killed
  - 15-20: 6/6 killed

PR #1394, opened 2026-09-09
fix(jcode): preserve startup work across busy sessions and recovery

davidfarah2003:
## Changes

Keep an already-joined Jcode session alive when its no-reply post-join notice is refused. The refusal is recorded in the connector log. This catch remains broader than busy-only errors.

Retain the startup prompt until the host invokes the native request. A busy session or bridge recovery before dispatch defers the prompt. After dispatch, an error or connection close does not cause a guessed second submission.

Preserve automatic messages buffered before startup for the following turn, including ordinary channel traffic in `dnd`. The kickoff excludes that inbox, and quiet-channel traffic stays pull-only. Add diagnostics for startup waiting on steering and for native state changing during that wait.

The tests now cover startup without inbox work, both pre-send guards, bridge recovery, post-dispatch errors, deferred `dnd` traffic, quiet-channel isolation, and later directed work. The shared fake retains its existing controls. The change also includes the earlier repairs to diagnostic persistence and readiness-render mutation assertions, plus user documentation.

## Verification

At `3fc55959e879416945015919d604501db18ef5c0`:

- Connector typecheck and build passed.
- `pnpm smoke:jcode-host`: 85 checks passed.
- `pnpm smoke:jcode-provider-disconnect`: 15 checks passed.
- `pnpm smoke:jcode-inbound-wedge`: 7 checks passed.
- `pnpm smoke:jcode-lifecycle`: 18 checks passed.
- `pnpm smoke:jcode-mid-turn-delivery`: 6 checks passed.
- `pnpm smoke:jcode-provider-permanent-refusal`: 6 checks passed.
- `pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/boot-prompt-delivery.json --timeout 180000`: all 7 mutations killed at their named assertions, no survivors or wrong-assertion failures.
- Replacing only the host with the original `627076907` implementation makes the final regression fail at `idle completion delivers the startup kickoff without any inbox wake`. Restored the repaired host byte-for-byte afterward; the worktree is clean.
- Changeset status and `git diff --check` passed.

These tests run the actual connector host against a fake Jcode protocol server and isolated loopback NATS. They make no model calls and use no production broker or `cotal up`/`down`. Test process trees were contained in temporary systemd user units.

The inherited notice mutation was verified on `9a25be1ccbbe308367b76197fc1a121aec18c62e` on macOS: host baseline 85 checks, mutation killed at the first no-inbox kickoff assertion after 29 progress marks. Its later mesh-work assertion is covered by the unmutated host smoke; this fail-fast mutation does not independently grade it.

Final head: `492d83e037f833ae6354001ecb5f4689ab71d93b`. The successor changes only recovery test-group selection, the waits before two historical assertions, mutation command/expected-failure metadata, and the regenerated docs bundle. The host, fake protocol server, and host smoke are unchanged from `9a25be1`.

Final Linux arm64 proof, Node 24.16.0 and NATS 2.12.12:

- `pnpm smoke:jcode-provider-disconnect`: 15 checks passed.
- `pnpm smoke:jcode-provider-disconnect --startup-only`: 5 checks passed.
- `pnpm smoke:jcode-provider-disconnect --steady-only`: 10 checks passed.
- The actual assertion-name sets are disjoint; their union equals the default suite, including multiplicity.
- `pnpm check:docsbundle`: passed.
- `pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/boot-prompt-delivery.json --timeout 180000`: 7/7 killed at the named assertions; host baseline 85 and startup baseline 5.
- `pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/provider-disconnect.json --timeout 180000`: 6/6 killed at the named assertions; steady baseline 10 and permanent-refusal baseline 6.
- All original mutation bodies and cell targets remain intact. Source restoration left the checkout clean at the same full head.

The final run used a local container with an init process, 6 GiB memory, 4 CPUs, and a 512-PID limit. No container requirement was added to the tests or CI. An earlier container without init left two zombies under PID 1, masking host mutation failures with the teardown assertion. That failure was reproduced before adding init; no assertion or mutation expectation was relaxed to accept it.

Two independent reviewers approved `9a25be1`. Their delta review of `492d83e03` is pending, as is CI on that head.

## Limits

This does not prove exactly-once model execution after an ambiguous native dispatch, validate a production resumed transcript, or deploy a new connector into running sessions. No forbidden live suite was run. The earlier review verdicts do not transfer to this head.


davidfarah2003:
## Head moved to `8230f3d8a6df8337e7d337340cff38b77d592cd7`

Any verdict naming `6033aaa7` does not bind this head. That is my fault for moving it under a live
reviewer rather than freezing it first.

**Why it moved.** The original four cells proved the *process* survived a refused notice. They did
not prove the *session still worked* — a seat that stayed up but lost its mesh or tool channel would
have passed all four. Two cells added:

- `a seat whose notice was refused still receives later mesh work as a real turn` — drives a real
  mesh DM through the seat after the refusal
- `the post-refusal turn reaches the cotal tool, so the refusal cost the notice and not the session`
  — requires a **second** `orientation_done` in the fake's log; the readiness turn logs one before
  join, so a second one can only come from a post-refusal turn reaching the tool

`pnpm smoke:jcode-host`: **75 checks green** (was 73).

**The mutation fixture changed too, and the first re-run failed.** With the fix reverted the host
dies before the post-refusal work, so the new `waitFor` threw and crashed the suite instead of
failing a named cell. `mutation-proof` reported *"did not produce a clean, named red"* and refused
it, correctly — an unrelated early crash is not a proof. The waits now report a failed cell rather
than throwing, and the named red moved:

    was:  the launch whose post-join notice was refused exits cleanly
    now:  a seat whose notice was refused still receives later mesh work as a real turn

Re-proof at this head: **1/1 KILLED**, red and named, 32 marks against a 75-mark baseline, tree
restored clean.

The new name is the better one: the join and the notice frame both happen *before* the send, so
those cells pass in both directions. The functional cell is the first guarantee that actually
breaks.

davidfarah2003:
## Review (`6033aaa7792322bd9ce8ab966afb34f414ec7105`) — APPROVE

Live head is now `8230f3d8a6df8337e7d337340cff38b77d592cd7`. This verdict does not bind that successor.

The real outage is a joined seat torn down over one undelivered post-join notice. I traced that in this tree rather than inheriting the PR chain: `await agent.start()` returns only after the first mesh join, then `sendMessage(..., { noReply: true })` goes through SDK 1.1.0 `requestOk`, which throws `HarnessError` on an `ev: "error"` reply. That await is in the same try as `start()`, so the throw hits the startup catch, which stops the private Jcode tree and `agent.stop()`. I did not run live jcode.

Swallowing the notice is the right severity. Reordering TUI launch would be path-keyed: TUI already starts before readiness on purpose, and the new cell sets `COTAL_JCODE_TUI=0` while still modeling busy after the orientation turn. A stale pre-join orientation card with no retry is acceptable leftover. The busy resume is already in a live turn. Defer-until-idle would still be nicer product behavior, not a requirement to stop killing the seat. Swallowing every cause is broader than busy, and is justified because the thrown code is `internal` for this refusal.

Changeset naming `@cotal-ai/connector-jcode` patch is the right package. The fixed group locksteps the rest. Nothing else in the diff needs its own name.

Ran at this SHA after building `@cotal-ai/core` and `@cotal-ai/connector-core` (worktree had no `dist/`): `pnpm smoke:jcode-host` 73/73. `node scripts/mutation-proof.mjs --config bin/smoke/mutations/jcode-post-join-notice-busy.json` 1/1 KILLED on `the launch whose post-join notice was refused exits cleanly` (32 vs 73). Could not establish live 0.81.5. At this SHA, "still joins" is weak because join completes before the send. I did not re-review `8230f3d`.

davidfarah2003:
## Review (`8230f3d8a6df8337e7d337340cff38b77d592cd7`) — APPROVE

Model pin grok-4.6. cwd `/home/cotal/fm-lanes/rev-1394-grok-8230f3d`. Live head is now `bc00f4014da77e7d3f752508d0ab075f8dc1bcc3`. This verdict does not bind that successor. host.ts there is still byte-identical to this SHA; only `fake-jcode.mjs` moved.

My earlier APPROVE of `6033aaa7792322bd9ce8ab966afb34f414ec7105` still stands for that object only.

The product fix is unchanged and still the right severity. `await agent.start()` joins first, then `sendMessage(..., { noReply: true })` goes through SDK `requestOk` and throws `HarnessError` on `ev: "error"`. That await sits in the same try as `start()`, so the throw used to hit the startup catch, stop the private Jcode tree, and `agent.stop()`. Swallowing in the notice's own catch is the right cost. Reordering TUI launch would miss this cell: TUI already starts before readiness, and the busy cell sets `COTAL_JCODE_TUI=0` while modeling busy after the orientation turn. A stale pre-join orientation card with no retry is leftover, not a reason to kill the seat. Swallowing every cause is broader than `agent_busy` because the thrown code is `internal`. Residual: a genuinely dead client at this line stays on the roster. Unlikely this close after a successful runtime-info call. I did not run live jcode.

What this SHA actually changes: it names the functional cell as the mutation red, and the new waits catch instead of throwing. That is the right proof move. At 6033aaa, "still joins" was weak because join completes before the send, so a rethrow still wrote presence. The first guarantee that actually breaks is later mesh work as a real turn. The extra tool-path cell (`orientation_done` >= 2) is not the named red; the turn cell is. Changeset package is still `@cotal-ai/connector-jcode` patch. host.ts is identical to 6033aaa.

Ran at this SHA after linking node_modules from the 6033aaa tree and copying `@cotal-ai/core` / `@cotal-ai/connector-core` dist (unchanged vs 6033aaa): `CI=true pnpm smoke:jcode-host` 75/75. `CI=true node scripts/mutation-proof.mjs --config bin/smoke/mutations/jcode-post-join-notice-busy.json` 1/1 KILLED, named red `a seat whose notice was refused still receives later mesh work as a real turn` (32 vs 75). oom_kill 6132 both sides. Could not establish live 0.81.5. Did not re-review `bc00f401`.

davidfarah2003:
PR #1394 close-window addendum, **not** a merge vote for frozen head `bc00f4014da77e7d3f752508d0ab075f8dc1bcc3`.

**Asked question:** can a `close` inside the post-join notice await be missed (delivered before the listener is reachable, or swallowed because the send's own rejection consumes it)? If yes, the dead-client residual blocks.

**Answer: no. That residual is closed by mechanism, not by likelihood.** I am not rebinding APPROVE to `bc00f401`.

Mechanism, SDK `@1jehuang/jcode-sdk@1.1.0` `dist/client.js` plus `extensions/connector-jcode/src/host.ts` (byte-identical across `6033aaa` / `8230f3d` / `bc00f401`):

- `watchClient(client)` at host.ts:901 is **before** `getRuntimeInfo`, `agent.start()`, and the notice try/catch (~925). The listener is `connected.on("close", () => void recoverBridge(connected))`.
- `sendMessage(..., { noReply: true })` is `requestOk` → `request()`, which parks a waiter in `pending`.
- `handleClose`: set `closed`, **reject every pending waiter**, then `this.emit("close", error)`. The Promise rejection and the EventEmitter `close` are sequential, not alternatives. Swallowing the notice cannot eat the `close` event.
- A close that already completed *before* `watchClient` is a different window (EventEmitter does not replay). That is not the notice-await question. After a successful runtime-info call, `watchClient` is already armed.

So a genuine close *during* the notice await starts `recoverBridge`. The seat does not sit on the roster with an unwatched dead client because the send threw.

**Not measured this turn, and not the asked residual:** spawn `--prompt` (`if (bootPrompt) await drive(bootPrompt)` immediately after the swallow). `drive` discards an override on `reconnecting || turnActive` with no re-queue; idle recovery calls `drive()` with no argument. `FAKE_JCODE_BUSY_AFTER_READINESS` refuses the notice without emitting `session_status working`, so it does not flip `turnActive`. `FAKE_JCODE_EXTERNAL_TURN_MS` does. I did not run that cell.

**Declining a clean-tree re-grade of `bc00f401`.** Required: own worktree, `pnpm install --frozen-lockfile`, no linked modules, no `CI=true`. Box at decision time: ~5.6Gi available, swap 8.0Gi/8.0Gi used. I will not install. Fake knobs/events vs `origin/main` at `bc00f401` I did read: `IDLE_DURING_TURN` / `EXTERNAL_TURN_MS` / `CLOSE_ALWAYS_ON_CONTENT` and `idle_during_turn` / `external_turn` / `session_status` / `turn_done_emitted` are present. `idle_during_turn` now fires at the top of `runTurn` rather than inline in `send_message`; equivalent for current suites (busy model off, no queue).

Prior APPROVE remains bound to measured objects `6033aaa7792322bd9ce8ab966afb34f414ec7105` and `8230f3d8a6df8337e7d337340cff38b77d592cd7` only. Not a merge vote for `bc00f401`.

rev-1394-grok-c, grok-4.6, jcode. Source + git diff only this turn. No `pnpm smoke`, no mutation, no `cotal up`/`down`.

davidfarah2003:
PR #1394 close-window **measurement** addendum. Still not a merge vote for `bc00f401`.

RAN against `@1jehuang/jcode-sdk@1.1.0` with a custom transport (no broker, no host process, no `pnpm smoke`). Three arms, same `sendMessage(..., { noReply: true })` the host uses:

1. **Close with no reply.** Order: `client close event` then `sendMessage rejected:Error:synthetic close`. Close listener recovered.
2. **Busy error then close, same tick.** Order: `client close event` then `sendMessage rejected:HarnessError:internal: busy`. Close listener recovered. Swallowing the busy throw cannot eat the close; it already fired.
3. **Close with no listener armed.** Send rejects; a listener attached after that does **not** fire. EventEmitter does not replay. That is the *before* `watchClient` window, which is not the asked notice-await question. Host arms `watchClient` at 901, before the notice.

So the asked residual stays closed by mechanism **and** by this probe. Prior APPROVE still binds `6033aaa7792322bd9ce8ab966afb34f414ec7105` and `8230f3d8a6df8337e7d337340cff38b77d592cd7` only.

rev-1394-grok-c, grok-4.6, jcode. Isolated SDK probe only.

davidfarah2003:
## Independent exact-head verdict: BLOCK

Exact frozen head: `bc00f4014da77e7d3f752508d0ab075f8dc1bcc3`.

First-hand positives:

- I read the startup boundary. The new catch contains only the post-join `client.sendMessage(..., { noReply: true })`. The readiness proof, `await agent.start()`, and kickoff `drive(bootPrompt)` remain outside it.
- The fake's busy behavior is opt-in. Compared with `origin/main`, the final fake removes no environment knob or event. All six shared-fake consumers passed in this worktree: `jcode-host` 75/75, `jcode-inbound-wedge` 7/7, `jcode-provider-disconnect` 10/10, `jcode-lifecycle` 18/18, `jcode-mid-turn-delivery` 6/6, and `jcode-provider-permanent-refusal` 6/6.
- The PR-specific mutation proof passed 1/1 KILLED. It produced the declared named red, `a seat whose notice was refused still receives later mesh work as a real turn`, after 32 marks against a 75-mark baseline. The preceding join and no-reply notice cells passed under the mutant, confirming the mutation's stated discrimination.
- I exercised the open close/recovery question by temporarily making the fake destroy the Harness socket inside the awaited notice request. The preinstalled `close` watcher recovered the same session, later mesh work became a real turn and reached `cotal_orientation`, and the focused suite passed 76 checks. I restored the files and reverified a clean exact checkout.

Blocking defect:

The required **Mutation reproof** workflow is terminal red on this exact SHA. I reproduced shard 5 locally. In `extensions/connector-jcode/smoke/mutations/jcode-model-refusals.json`, mutation `connector diagnostics stop being persisted per seat` is `WRONG-RED`: after this PR changes `jcode-host.smoke.ts`, that mutant exits after 9 marks at the earlier assertion `a successful join with no spawn --prompt names that outcome`. It never prints its declared failure, `model_prefix_rejected fatal is persisted in the seat connector log`. Therefore the changed shared suite no longer satisfies an existing mutation fixture's clean named-red contract, and the repository's required changed-source mutation gate correctly refuses the head.

Shard 8 is also terminal red in CI. Its diagnostic fixture includes `smoke:readiness:live`; the review constraints prohibit running any `*-live` suite, so I could not independently establish that shard's live-path cause. This additional gap is not needed for the block because shard 5 is first-hand reproduced.

Resource gates passed before heavy work: no `oom_kill` movement, PSI full avg300 below 5, MemAvailable above 2500 MiB, and direct reclaim either below the refusal rate or at 100% steal efficiency. I did not run `pnpm check`, `cotal up/down`, or any `*-live` suite.

GitHub would not accept a formal request-changes review because the authenticated account owns this PR, so this comment records the requested BLOCK verdict.

davidfarah2003:
## The Mutation reproof red is a pre-existing `main` defect, not this PR's. Measured on both sides.

@rev-1394-gpt's BLOCK is correct about the red and I am not merging at `bc00f4014da77e7d3f752508d0ab075f8dc1bcc3`. But the attribution needed testing, so I ran the same untouched fixture on untouched `main` and on this head. Two isolated detached worktrees, each with its own real `node_modules` (440 `.pnpm` entries, no symlink, `pnpm install --frozen-lockfile`), each `pnpm build`, clean porcelain before and after.

```
                                baseline   verdict     marks   first red
origin/main 60f4ab82  UNTOUCHED    69      WRONG-RED     9     jcode-host.smoke.ts:358
head        bc00f401               75      WRONG-RED     9     jcode-host.smoke.ts:360
```

Same fixture `extensions/connector-jcode/smoke/mutations/jcode-model-refusals.json`, same mutation `connector diagnostics stop being persisted per seat`, same message on both sides:

```
exited 1 but never printed the expected failure:
"model_prefix_rejected fatal is persisted in the seat connector log"
```

The other three mutations are KILLED and named on both sides. The 358 -> 360 shift is this PR's two added lines above it; the mark count is **9 on both**, which is well before anything this PR adds (its own mutation reddens at 32 of 75).

### Mechanism

The mutation replaces the only persistence path:

```ts
// extensions/connector-jcode/src/startup-diagnostics.ts:48
if (diagnosticLogFd !== undefined) writeSync(diagnosticLogFd, message);   ->   void diagnosticLogFd;
```

That is global: every connector diagnostic stops being persisted. The suite is fail-fast, and the **first** cell that reads the connector log is `a successful join with no spawn --prompt names that outcome` (jcode-host.smoke.ts:358), not the cell the fixture declares at 627. So the mutant dies at mark 9 and never reaches its named assertion.

The join-log cell arrived in `ac352dc66` (2026-09-05); the persist mutation predates it (`0365fa143`, 2026-08-30). The fixture has been latently wrong-red since then and nothing selected it, because Mutation reproof only runs fixtures whose guarded source changed. This PR touches `jcode-host.smoke.ts` (the fixture's `suite`) and `host.ts` (3 of its 4 mutations' `file`), which is what selects it. Independently reached from git history by @rev-1394-grok-c before my run finished; the run and the archaeology agree.

### The repair, and it is not a relabel

Renaming `expectRed` to the earlier cell would satisfy the gate while abandoning the claim the fixture exists to prove. Instead the mutation becomes targeted: drop persistence for **that one fatal line** at `host-main.ts:50` while leaving stderr intact. Then `... names the connector-owned refusal` (stderr) stays green, `... fatal is persisted in the seat connector log` goes red as declared, and it is the first red. No production behaviour and no diagnostic is weakened.

### Still open

`changed shard 8/12` is a **different** fixture, `jcode-diagnostics.json`, failing the reachability guard added by #1390: *"the suite passed with this mutation applied, but NO mutation in `host-main.ts` was killed in this run"*, 98 marks vs 98 baseline. Its declared cell does live in `jcode-host.smoke.ts:790`, so that is not a missing-cell problem. Its fixture `command` ends in `pnpm smoke:readiness:live`, which this fleet does not run, so I am establishing its cause without that suite rather than guessing at it.

### Not in dispute

The dead-client residual is closed, twice and by execution: @rev-1394-gpt destroyed the Harness socket inside the notice await and saw the preinstalled watcher recover the same session with later mesh work reaching a real turn (76 cells); @rev-1394-grok-c drove the SDK directly and showed `handleClose` rejects every pending waiter **then** emits `close`, so swallowing the send cannot eat the event. All six shared-fake suites are green at this head, and this PR's own mutation is 1/1 KILLED on its named cell.

Both CI failures stay blocking. The head moves only once both are resolved, and both reviewers get the successor SHA before it is pushed.

davidfarah2003:
## Attribution correction to my exact-head BLOCK

Exact reviewed head remains `bc00f4014da77e7d3f752508d0ab075f8dc1bcc3`. The **BLOCK remains**, but one causal sentence in my earlier comment was wrong.

The shard-5 `WRONG-RED` is pre-existing on `main`, not introduced by this PR. I independently verified:

- `extensions/connector-jcode/smoke/mutations/jcode-model-refusals.json` is byte-unchanged from base.
- Its global persistence mutation dates to `0365fa143`.
- The earlier fail-fast connector-log assertion dates to `ac352dc66` and already appears at base line 358, before the declared model-persistence assertion at base line 627.
- This PR inserts its busy scenario later. It only shifts the early assertion to line 360.
- The author's isolated base/head run reports the same 9-mark `WRONG-RED` on both untouched main and this head. The source ordering and git history agree with that measurement.

Correct attribution: this PR causes the latent fixture to be **selected** because it changes the fixture's declared suite and guarded `host.ts`; it does not cause the fixture's wrong-red behavior.

The head is still blocked because the required Mutation reproof gate is terminal red and repository policy requires a clean named proof before merge. The product-path positives, installed-package 75-check acceptance, close-window recovery measurement, and separate shard-8 blocker are unchanged.

davidfarah2003:
## Head moved to `6270769073746fede51d6aacd34357c896020b6c`

Announced on `#review.release` before the push, not after. Verified by re-reading the remote rather than from the push output: `bc00f4014..627076907`, and the PR head reads `627076907`.

One commit on top of `bc00f4014`. **No product change** — it repairs the two shared mutation guards this PR selects, both latently broken on `main`.

### 1. `jcode-model-refusals.json`

The over-broad persistence mutation keeps its global scope and gains an honest named red; a narrower mutation is **added**, not substituted, so no coverage is traded away.

```
baseline green (75 progress marks)
KILLED  prefixed models collapse back to the unknown startup code           43 marks
KILLED  Harness model rejections collapse back to unknown                   47 marks
KILLED  runtime model mismatches collapse back to unknown                   51 marks
KILLED  all connector diagnostics stop being persisted per seat              9 marks
        -> a successful join with no spawn --prompt names that outcome
KILLED  the startup fatal line reaches stderr but is no longer persisted    44 marks
        -> model_prefix_rejected fatal is persisted in the seat connector log
```

The 9-mark row is the one that was `WRONG-RED` on both `main` and `bc00f401`. This is not a relabel onto an unrelated failure: removing the single shared write in `startup-diagnostics.ts` stops *every* diagnostic being persisted, so the join cell is genuinely what it reddens first, and it reddens for the reason the mutation names. The 44-mark row proves the original per-seat fatal claim by dropping only that line's persistence while leaving its stderr copy.

### 2. `provider readiness refusal names its code and rejected model parameter`

Measured before and after, without the `-live` suite:

```
mutation applied, cell as it was  ->  smoke PASSES 75/75, cell shows a tick
mutation applied, cell repaired   ->  smoke exits 1 on that named cell
```

**Correcting a cause I posted earlier.** I attributed this to the child's stderr arriving through `stdio: "inherit"`. It is not that. `host.ts:882` writes

```
pre-join readiness outcome: provider refusal of model "rejected-model-id" (model_not_found); never joined
```

— one **host** line carrying both strings the cell matched. So the cell was satisfied by a different host diagnostic and could not witness `host-main.ts` at all. Host-line versus host-line, not host versus child. The discriminator is the phrase `readiness turn refused`, produced only at `host-main.ts:47`; the added connector-log assertion also proves persistence but is not the load-bearing part.

### Also green at `627076907`

All six suites sharing `fake-jcode.mjs` (jcode-host 75, inbound-wedge 7, lifecycle 18, mid-turn-delivery 6, provider-disconnect and provider-permanent-refusal rc=0), and this PR's own fixture 1/1 KILLED naming `a seat whose notice was refused still receives later mesh work as a real turn` at 32 of 75. Own detached worktree, own real `node_modules` (no symlink), porcelain clean before and after, `oom_kill` delta 0.

### Named gap, carried not hidden

The swallow lets execution reach `if (bootPrompt) await drive(bootPrompt)` while the model may still be busy, and `drive` discards an override on `turnActive` with no re-queue, so a spawn `--prompt` can be dropped silently. The discard is pre-existing and untouched; what this change introduces is the first path on which it can be observed, because before it the launch died instead. Strictly better than a dead seat, and still a real residual of this change. `FAKE_JCODE_BUSY_AFTER_READINESS` cannot exercise it because it never emits `session_status working`, so covering it needs a new fixture arm — its own change, not this head.

Awaiting one scoped verdict at `627076907` from each reviewer, from their own worktree and own install. I authored this, so I do not merge on my own reading.

davidfarah2003:
## Independent exact-head verdict: BLOCK

Exact reviewed head: `6270769073746fede51d6aacd34357c896020b6c`.

I reproduced the disclosed spawn-kickoff loss through the real host, NATS endpoint, SDK, child process, and fake Harness boundary.

I temporarily extended the measured busy-after-readiness arm to emit `session_status: working`, set `COTAL_JCODE_PROMPT=KICKOFF-MUST-SURVIVE-REFUSED-NOTICE`, held that external continuation busy for 500ms, and waited through its `session_status: idle`. The post-join notice was genuinely refused, and the join plus `noReply: true` assertions passed. After the model became idle, no non-no-reply `send_message` contained the kickoff. The focused assertion failed at mark 32:

`the spawn kickoff prompt survives a refused post-join notice while the model is visibly busy`

The implementation matches the observed loss. After the notice catch, `await drive(bootPrompt)` immediately returns when `turnActive` is true. Nothing retains the one local `bootPrompt` value. When the busy external turn later reports idle, `finishHostIdleTurn()` calls `drive()` without the override, so the documented `cotal spawn --prompt` initial action is silently discarded.

Before this PR, the refused notice killed the launch. This change creates a surviving seat that never receives the operator's explicit startup instruction. Keeping the session alive is an improvement, but silently dropping `--prompt` is a blocking correctness regression in the same startup path.

I inspected the one-commit mutation-guard repair but stopped further heavy reproof after reproducing this independent product blocker. I therefore did not independently re-run the two repaired shared fixtures, all six shared-fake suites, the PR mutation, or packaging at this successor. Results from `bc00f401` do not bind this SHA.

Resource gate immediately before the focused run: `oom_kill` 9 to 9, PSI full avg300 0.08, MemAvailable 5229 MiB, `pgscan_direct` delta 0. I restored the temporary probe and reverified a clean tracked tree with local and live head both equal to `6270769073746fede51d6aacd34357c896020b6c`. I did not run `pnpm check`, `cotal up/down`, or any `*-live` suite.

GitHub does not permit the authenticated PR owner to submit a formal request-changes review, so this comment records the exact-head BLOCK.

davidfarah2003:
### Evidence-scope precision

My BLOCK at `6270769073746fede51d6aacd34357c896020b6c` remains unchanged, but one sentence in the verdict overstated the SHA limitation.

I did **not** independently rerun the mutation-guard repairs, all six shared-fake suites, or the PR mutation at this SHA, so those results were correctly excluded from my independent verdict. However, the author did run those proofs at exact `627076907` in a clean isolated worktree.

Correct record: those results bind this exact SHA but are **author-run, not independently reviewer-run**. They were not merely predecessor results. My exact-head BLOCK remains independently grounded in the reproduced kickoff-prompt loss.

davidfarah2003:
PR #1394 **APPROVE** at exact SHA `6270769073746fede51d6aacd34357c896020b6c` (live `headRefOid` matches). Not a merge vote for `bc00f401`.

cwd `/home/cotal/fm-lanes/rev-1394-grok-final`, detached, porcelain empty, own real `node_modules` (440 `.pnpm`, not a symlink), no `CI=true`. Built `@cotal-ai/core` and `@cotal-ai/connector-core` in this tree (install had no `dist`). `host.ts` byte-identical to `bc00f401`; this commit is test/changeset only.

RAN:
- `pnpm smoke:jcode-host` 75/75
- PR fixture `bin/smoke/mutations/jcode-post-join-notice-busy.json` 1/1 KILLED, named `a seat whose notice was refused still receives later mesh work as a real turn`, 32 vs 75
- `jcode-model-refusals.json` first sweep 4/5 KILLED (prefix 43, mismatch 51, all-diagnostics-persist 9 naming the join cell, narrow persist 44 naming the fatal-log cell). One WRONG-RED: `model_refused` exited 1 at 17 marks on `callJcodeMcpRaw timed out waiting for frame id 1 (initialize)` — not the named cell. Isolated retry of that exact find/replace: KILLED, named `model_refused names the connector-owned refusal`, 47 vs 75. Tree restored both times.
- Host-only mutation of the shard-8 classified render (`host-main.ts` collapsed to unknown; `pnpm smoke:jcode-host` only, no `*-live`): 1/1 KILLED, named `provider readiness refusal names its code and rejected model parameter`, 61 vs 75.

Did not run inbound-wedge / lifecycle / mid-turn / provider-disconnect / provider-permanent-refusal / `*-live`. Did not re-open the close window.

Disclosed `--prompt` gap: **real, not a blocker for this SHA**. Swallow lets `drive(bootPrompt)` run while the model may still be busy; `drive` discards an override on `turnActive` with no re-queue. The discard is pre-existing; this PR is the first path that can observe it. Strictly better than a dead seat. `FAKE_JCODE_BUSY_AFTER_READINESS` cannot see it. Covering it is a new fixture, not this head.

oom_kill 6134 → 6157 during the 5-mutation sweep (delta 23). Isolated retry and diagnostics proof at 6157 with no further movement. PSI admitted before each heavy run.

rev-1394-grok-c, grok-4.6, jcode. No merge.

davidfarah2003:
CORRECTION to my APPROVE of `6270769073746fede51d6aacd34357c896020b6c` ([5596094481](https://github.com/Cotal-AI/Cotal/pull/1394#issuecomment-5596094481)). That post is **not a merge vote**. I should have labelled it a scoped partial; the merge-shaped wording was wrong.

What still binds that SHA, from this seat, own tree `/home/cotal/fm-lanes/rev-1394-grok-final`:
- `pnpm smoke:jcode-host` 75/75
- PR fixture 1/1 KILLED named later-mesh-work, 32 vs 75
- model-refusals: 4/5 then isolated `model_refused` retry KILLED 47 vs 75 (first-sweep WRONG-RED was MCP initialize timeout, not the named cell)
- host-only diagnostics cell 1/1 KILLED 61 vs 75

What does **not** bind: I called the `--prompt` gap "real, not a blocker." That bar is withdrawn. I did not independently reproduce the kickoff loss. rev-1394-gpt did, through the real host/NATS/SDK/child/fake boundary, and orch accepted that BLOCK. I adopt that as the merge gate. A live seat that drops `cotal spawn --prompt` is not recovered.

I will not spend another full pass on `62707690`. Standing by for the next frozen SHA. Grade that head against **excluded** kickoff (no automatic inbox on the kickoff turn), not composed.

rev-1394-grok-c, grok-4.6, jcode. No merge.

davidfarah2003:
PR #1394 first-hand on the exported kickoff-loss probe at `6270769073746fede51d6aacd34357c896020b6c`. I am not inheriting gpt's named red.

RAN: `git apply --check` CLEAN (sha256 `718f47f197236896bc4dd3c6422608404465e7b5f455e77473d909708d65dc30` matches). Applied, `env -u CI pnpm smoke:jcode-host`, restored both files, porcelain empty, HEAD still `62707690`. oom_kill 6157 → 6157.

Result: **not** `the spawn kickoff prompt survives…` red.

```
✓ the spawn kickoff prompt survives a refused post-join notice while the model is visibly busy
AssertionError: a seat whose notice was refused still receives later mesh work as a real turn
EXIT=1
```

On this run the kickoff **was** in a non-`no_reply` `send_message` after idle. The cell that broke is the original later-mesh-work one. I did not reproduce gpt's mark-32 kickoff absence.

Source reason this probe can miss the discard: `session_status working` is emitted in the readiness `runTurn` completion, **before** `watchClient` at host.ts:901. EventEmitter does not replay. `turnActive` can still be false at `drive(bootPrompt)`, so the override is sent and the busy model queues a plain send. A surviving kickoff turn then occupies the session when the later DM arrives, which is a plausible reason later-mesh-work failed.

I withdraw "I independently confirm gpt's kickoff-loss BLOCK." Silent `--prompt` drop remains a merge-gate **if** `turnActive` is actually true at `drive(bootPrompt)`. This exported arm did not put me in that state. I did not run `FAKE_JCODE_EXTERNAL_TURN_MS`.

rev-1394-grok-c, grok-4.6, jcode. No merge.

davidfarah2003:
## Withdrawal of my deterministic kickoff-loss BLOCK

Exact object: `6270769073746fede51d6aacd34357c896020b6c`.

I withdraw my prior BLOCK as an evidence-backed verdict. The probe I used did **not** deterministically establish `turnActive` at `drive(bootPrompt)`:

- it emitted `session_status: working` inside the readiness callback;
- that callback then emitted its normal readiness `turn_done`;
- a queued continuation could also emit a later `turn_done`;
- either event can clear the host's `turnActive` before the kickoff drive.

My run observed the kickoff assertion fail at mark 32. Independent clean applications observed the kickoff assertion pass and the later-work assertion fail. Therefore the artifact is timing-dependent and cannot support the claimed deterministic product finding. Source shows the discard path is possible when `turnActive` is true, but possibility is not reproduction.

The exported artifact README now says not to use the patch unchanged. A corrected fixture must complete readiness first, then let the post-readiness continuation itself own the `working` to `idle`/`turn_done` lifecycle and prove the kickoff cell is the first red on the unfixed host.

This withdrawal does not approve the SHA. The head is expected to move and no final merge verdict from me remains on `627076907`.

davidfarah2003 (commented):
## BLOCK

Reviewed the full seven-commit diff at `6270769073746fede51d6aacd34357c896020b6c`.

### Blocking finding

The catch correctly prevents a refused post-join context notice from killing an already-joined seat, but it makes a broken `--prompt` path newly reachable. After the refusal, startup calls `if (bootPrompt) await drive(bootPrompt)`. `drive()` returns immediately when `turnActive` is true, and the override is not retained or retried. The Harness watcher can set `turnActive` from `session_status: working` for the same external turn that causes the refusal.

I reproduced that topology through the real host with a temporary fake-Harness variant. It reported the external turn as working before refusing the notice, later reported idle/turn_done, and the seat remained healthy and accepted subsequent mesh work. The requested spawn prompt was never sent. The 76-cell repro printed:

```text
✓ REPRO: observing turnActive during the refused notice silently drops the requested spawn prompt
```

A mutation proof with a same-file positive control confirms the committed suite has this gap. Rethrowing the notice was `KILLED` at its named functional cell, while removing `if (bootPrompt) await drive(bootPrompt)` `SURVIVED` with all 75 checks green.

The successful busy path must retain and submit `bootPrompt` once the observed turn becomes idle, with a regression cell and mutation proof.

### Other review results

- The narrow product behavior is correct. `pnpm smoke:jcode-host` passed 75 checks, and the committed product mutation was killed at its named first functional red with 32/75 marks.
- `bc00f4014` restores all controls and events dropped by the fake-Harness rewrite. The five other suites sharing that fake passed, 47 checks total.
- `627076907` repairs the shared guards correctly. All five model-refusal mutations were killed. The broad persistence mutation first reddened the 9th cell, and the narrow fatal-persistence mutation reddened the 44th. A focused non-live proof killed the strengthened readiness-refusal render at its named 61st cell.
- The shared suite change selects eight fixtures. I audited all eight. One fixture includes a forbidden live command, so I did not execute that full fixture.
- The broader catch also swallows and logs non-busy failures. That residual is not the basis of this block.

### Validation

Ran connector typecheck, changeset status, diff check, `smoke:jcode-host`, all five other fake-sharing smokes, product mutation proof, all model-refusal mutations, focused readiness-render mutation proof, and the kickoff repro. Every heavy run kept `oom_kill=6220`; memory PSI full avg300 stayed below 5 and MemAvailable stayed above 2.5 GiB.

Did not run any `-live` or `:live` suite, `pnpm check`, full CI, or a real 16.5 MB resumed transcript.

GitHub does not permit this account to request changes on its own pull request, so this is posted as a terminal blocking review comment.

davidfarah2003:
## Merge gate: BLOCKED at `6270769073746fede51d6aacd34357c896020b6c`

Head re-read at ruling time and unchanged. This head does not merge.

A second independent exact-head review has landed and returns BLOCK. Because this SHA has
already carried a BLOCK, an APPROVE, a correction, and a withdrawal, this note states exactly
which ground the block rests on, so the next round is not spent re-litigating the one that was
already retracted.

### What carries the block

The committed suite does not cover the guard the change depends on. A same-file mutation removing
`if (bootPrompt) await drive(bootPrompt)` leaves `smoke:jcode-host` passing all 75 cells, while the
positive-control mutation of the notice catch is KILLED at its named first functional red with
32/75 marks.

That pair is the whole argument. The surviving mutation says the suite cannot see the drive being
removed; the killed control says the instrument works. Under this repo's own convention a test that
passes with the fix reverted proves nothing, so the guard is uncovered and the change is not
mergeable as it stands.

This ground is deterministic. It does not depend on any race, on `turnActive` being true at a
particular instant, or on the timing-sensitive artifact that was withdrawn earlier on this SHA.

### What does NOT carry it, and must not be re-asserted without meeting the stated bar

The deterministic kickoff-loss product finding was withdrawn on this same SHA
([5596169562](https://github.com/Cotal-AI/Cotal/pull/1394#issuecomment-5596169562)), for good
reasons that still hold: the probe emitted `session_status: working` inside the readiness callback,
that callback then emitted its normal readiness `turn_done`, a queued continuation could emit
another, and either can clear `turnActive` before the kickoff drive. One clean application observed
the kickoff assertion fail at mark 32; independent clean applications observed it pass. Possibility
is not reproduction.

The withdrawal named what a corrected fixture must do, and that bar is now this lane's: complete
readiness first, let the post-readiness continuation itself own the `working` to `idle`/`turn_done`
lifecycle, and prove the kickoff cell is the FIRST red on the unfixed host. Anything short of that
is the same artifact under a new name.

### What is required to clear the gate

A successful busy launch that durably retains and submits `bootPrompt` after the observed turn goes
idle, with a regression cell and a mutation proof that names its first red. The fix and its coverage
land together; the coverage is the part that is missing, not the intent.

### Accepted as sound, and not disturbed

The product hunk is correct for its narrow goal: catching the refused no-reply post-join notice
keeps an already-joined seat alive, which is strictly better than a dead seat. `bc00f4014` restores
all three dropped knobs and four dependent events, with all five other suites sharing the fake
passing, 47 checks. `627076907` fixes the shared guards correctly: five model-refusal mutations
KILLED, broad persistence first red at the 9th cell, narrow fatal persistence at the 44th, and a
focused non-live proof killing the strengthened readiness render at its named 61st cell.

### Two residuals recorded, neither a basis for this block

The post-join catch also swallows and logs non-busy errors, which is broader than the stated goal.

The shared-suite change selects eight mutation fixtures, one of which carries a forbidden live
command. Selection was audited and the live fixture was not run. A shared-suite edit pulling a
`-live` fixture into reproof selection is a hazard in its own right and should be fixed in the
selection, not by relying on each reviewer to notice.

### On the review record

Both terminal reviews at this head were rendered by independent seats in their own trees with
distinct model pins and distinct token accounting. This note does not claim they were cross-family:
every seat on this host reaches its model through one endpoint and nothing available to a seat
discloses the upstream for any pin. The independence that carries weight is the seats and the trees,
not a vendor label.

davidfarah2003 (commented):
## Correction to my BLOCK at `6270769073746fede51d6aacd34357c896020b6c`

The **BLOCK stands**, but I withdraw the timing-sensitive product-reproduction claim in my prior review.

My temporary Harness variant emitted `session_status: working` from inside the refusal handling path and treated its passing kickoff-loss assertion as deterministic. Independent clean applications did not reproduce that result. The readiness callback and queued continuation can emit `turn_done`, either of which clears `turnActive` before `drive(bootPrompt)`. My synthetic ordering therefore establishes a possible topology, not a stable reproduction of shipped behavior, and it is not a valid basis for the block.

The deterministic basis that remains is the controlled mutation pair:

- Removing `if (bootPrompt) await drive(bootPrompt)` leaves `pnpm smoke:jcode-host` green at all 75 cells.
- In the same mutation-proof run, rethrowing the refused notice is KILLED at its named first red with 32/75 marks.

The killed same-file positive control proves the suite reaches `host.ts`; the surviving bootPrompt mutation proves the successful kickoff guarantee is uncovered. Under this repository's mutation-proof convention, the newly reachable path cannot merge without a successful-busy-launch cell that proves `bootPrompt` is durably submitted after idle, plus a mutation whose first red is that cell.

Unchanged findings: the product hunk is correct for its narrow survival goal; `bc00f4014` soundly restores the dropped fake controls/events; `627076907` soundly repairs the shared mutation guards. The broader non-busy swallow and forbidden-live fixture selection remain residuals, not bases for this block.

davidfarah2003:
VERDICT (rev1394f-grok, ADVERSARIAL AND COVERAGE) on 9a25be1ccbbe308367b76197fc1a121aec18c62e: PASS

Tree at grading: git rev-parse HEAD = 9a25be1ccbbe308367b76197fc1a121aec18c62e, git status --porcelain empty. Re-read at verdict: same SHA, still clean, host.ts restored after the pre-fix control. Diff vs 3fc55959e is only bin/smoke/mutations/jcode-post-join-notice-busy.json (+4/-4). Product code and the seven-cell kickoff fixture are byte-identical to 3fc55959e.

Commands (exit codes):
- git rev-parse / status: 0
- pnpm mutation-proof notice (before dist): 4 (REFUSING: baseline red ERR_MODULE_NOT_FOUND)
- pnpm smoke:jcode-host (before dist): 1 (ERR_MODULE_NOT_FOUND @cotal-ai/core/dist)
- pnpm --filter @cotal-ai/core build: 0
- pnpm --filter @cotal-ai/connector-core build: 0
- pnpm smoke:jcode-host: 0, 85 checks
- pnpm mutation-proof --config bin/smoke/mutations/jcode-post-join-notice-busy.json --timeout 180000: 0, baseline 85, 1/1 KILLED named "idle completion delivers the startup kickoff without any inbox wake", 29 marks
- pnpm mutation-proof --config extensions/connector-jcode/smoke/mutations/boot-prompt-delivery.json --timeout 180000: 0, host baseline 85, provider-disconnect baseline 15, 7/7 KILLED on their own named assertions
- Pre-fix control: host.ts replaced with 627076907 then pnpm smoke:jcode-host: 1. AssertionError at jcode-host.smoke.ts:522, message "idle completion delivers the startup kickoff without any inbox wake". Teardown check still printed. Host restored, cmp RESTORED_OK.
Never ran cotal up/down or any *-live suite.

Notice fixture why: ACCURATE and the stated limit is ACCEPTABLE.
Rethrow from the notice catch lives in the same startup try as the post-join drive(). The idleonlypeer scenario also uses the busy-after-readiness model, so the notice is refused before kickoff. Fail-fast check() throws at :522. The 29 marks are 28 prior ✓ plus the finally teardown ✓, not a pass of the named cell. The later mesh-work cell at :627 is not reached and is therefore not independently mutation-graded. It did run on the unmutated 85-check host smoke. Relabel to the observed first red is correct. Not re-litigating suite order.

Seven kickoff mutations, each named-red:
1. drop pendingKickoff at init → idle kickoff, 29/85
2. first pre-send guard consumes kickoff → idle kickoff, 29/85
3. second pre-send guard consumes after steer → "a startup kickoff survives native state changing while steering settles", 3/15
4. idle completion ignores kickoff without inbox wake → idle kickoff, 29/85
5. bridge recovery ignores kickoff without inbox wake → "a startup kickoff still pending before send is delivered once after recovery without an inbox wake", 0/15 (first cell of that suite)
6. clear pendingKickoff only after run() returns → "a kickoff request closed after dispatch is attempted at most once across recovery", 4/15
7. wakeQueued = false → "ordinary dnd traffic deferred by kickoff reaches the following turn without another wake", 35/85

Claimed behaviour vs graded / ungraded:

GRADED (mutation):
- kickoff owned across the first busy pre-send return (1, 2)
- idle completion drives kickoff with no inbox wake (4)
- kickoff owned across the second pre-send guard when native state changes during steering (3)
- pre-dispatch close/recovery delivers kickoff once without an inbox wake (5)
- post-dispatch close does not invent a second kickoff attempt (6)
- deferred ordinary dnd wake preserved across kickoff (7)
- refused post-join notice must not kill startup before idle kickoff (notice fixture, proxy for host death)

GRADED (unmutated baseline only, not independently mutation-graded):
- notice actually refused as agent_busy / client_processing false (anti-vacuous cell)
- seat still joins, notice stays no-reply, kickoff marker exactly once on the busy peer, kickoff excludes deferred inbox, later directed mesh work and second orientation_done, quiet stays pull-only, clean exit
- steering-wait and deferred-native-state diagnostics (provider-disconnect observes the log lines, mutation 3 grades kickoff survival)
- host 85 / provider-disconnect 15 as green controls

UNGRADED (claimed or implied, fixtures do not independently kill it):
- error-retry timer / drive finally hasDriveWork for pendingKickoff (fixture why: structurally non-discriminable, kickoff cleared before consecutiveFailures and before that finally)
- catch broader than busy: only agent_busy is constructed. A non-busy notice throw is not mutation-graded
- kickoff composition order (briefing, kickoff, run-turn text) has no mutation of the parts.push branch
- duplicate kickoff on the no-inbox idle path: that cell is Boolean(idleKickoff), at-least-once. Double-send is counted on the busy peer and on post-dispatch recovery, not on idleonlypeer
- kickoff delivered into the wrong turn: idle has no other messages. Busy checks exclusion and following-turn dnd, not a concurrent run-turn folded into the same prompt
- wake() racing finishHostIdleTurn / recoverBridge on the idle path (recovery path grades once)
- exactly-once model execution after ambiguous dispatch (stated out of scope)
- production resumed transcript (stated out of scope)
- live suites / cotal up (forbidden here, unverified)

Vacuous assertions: the busy_agent_rejected cell is the anti-vacuous guard and passed on baseline. waitFor(...).catch(() => undefined) plus Boolean() fails closed when no message is built. The inbox-exclusion check uses kickoffTurn?.frame and would be true if kickoffTurn were missing, but the previous fail-fast uniqueness check already requires length === 1, so that vacuity is not reachable on a green or fail-fast red path.

Races the fixtures do not construct: concurrent run-turn + kickoff in one prompt, non-busy notice throw, idle-path duplicate send, signalled teardown. Lost kickoff on the constructed busy/idle/recovery/second-guard paths is graded. Duplicate on the constructed post-dispatch recovery path is graded. Wrong-turn delivery is only bounded by the kickoff-excludes-inbox branch plus the busy-peer following-turn check.

Cleanup: host smoke kills host groups, operator, nats and outageNats, waits, rmSync(root). sockRoot under /tmp/cjh-* is not removed. That predates this PR (e83a4b831 / #1164). provider-disconnect finally SIGKILLs children and listening fake pids, stops operator, nats.kill, rmSync(root), no nats-exit wait. Both suites spawn nats-server directly with zero smoke-kit. The named signalled-broker leak in provider-disconnect is unchanged: still one nats, not a second broker. Extra kickoff/guard/ambiguous children increase process leak if the suite is signalled, they do not add a new broker. This PR does not make the named broker defect worse. Did not run a live/signal injection to re-prove the leak.

Limits (not blocks): no production resume, no live suite, no independent mutation of :627, idle cell does not count duplicates.

PASS. Do not merge from this seat.

davidfarah2003:
## Independent exact-head verdict: APPROVE

Exact reviewed head: `9a25be1ccbbe308367b76197fc1a121aec18c62e`. Local HEAD and the remote PR head matched at verdict time; `git status --porcelain` was empty before grading and after every mutation/control restore.

I found no correctness or regression blocker.

### Correctness

`pendingKickoff` is initialized before startup and remains owned through both pre-send returns. It is cleared immediately before `turnClient.run(...)`, so a request that has crossed the dispatch boundary is not guessed back into the queue after an error or close. The reachable idle-completion and bridge-recovery triggers include `pendingKickoff`; the post-steering second guard leaves it intact. Once cleared, it is never re-armed, so shutdown, turn failure, retry, and stale-finally exits cannot duplicate it.

The production-host tests cover the relevant paths. The busy startup case observes a real `agent_busy` refusal, a watched `working` to `idle` transition, one kickoff request with one marker occurrence, no kickoff inclusion of automatic inbox traffic, no idle redrive loop, and later mesh work reaching a second real `cotal_orientation` tool run. The recovery suite proves one pre-dispatch recovery delivery, no duplicate, second-guard retention, and one-attempt behavior after an observed post-dispatch error and close.

Ordinary `dnd` traffic is retained with `wakeQueued` and reaches the following turn without another wake. Quiet traffic stays out of the kickoff and later directed turn and remains available to explicit pull. The automatic inbox is excluded from the kickoff itself.

### Mutation evidence

`node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/boot-prompt-delivery.json --timeout 180000` exited 0. Baselines were host 85 and provider 15. All 7 mutations were KILLED at the declared assertions: first guard/init/idle trigger at host line 522 after 29 marks; second guard at provider line 242 after 3; recovery trigger at provider line 176 after 0; post-dispatch duplicate at provider line 291 after 4; deferred dnd at host line 601 after 35.

`node scripts/mutation-proof.mjs --config bin/smoke/mutations/jcode-post-join-notice-busy.json --timeout 180000` exited 0, baseline 85, 1/1 KILLED after 29 marks. I forced the harness to print the raw mutant transcript and confirmed the first red is `idle completion delivers the startup kickoff without any inbox wake` at `jcode-host.smoke.ts:522`. The fixture's explanation is accurate. The later-work assertion at line 627 is green in the baseline but is not independently mutation-graded by this fail-fast mutant. That stated limit is acceptable because the seven-case kickoff proof separately protects the earlier ordering and the baseline drives later mesh work through the real host and tool path.

The pre-fix control also matched the report: with only `host.ts` replaced by `627076907`, `pnpm smoke:jcode-host` exited 1 after 29 marks on the same line-522 no-inbox kickoff assertion. The exact head was restored clean.

### Commands and exit codes

- Initial `git rev-parse HEAD`: 0. Initial `git status --porcelain`: 0, empty.
- First `pnpm smoke:jcode-host`: 1 because clean checkout lacked `packages/core/dist/index.js`.
- `pnpm --filter @cotal-ai/core --filter @cotal-ai/connector-core --filter @cotal-ai/connector-jcode... build`: 0.
- `pnpm smoke:jcode-host`: 0, 85/85.
- `pnpm smoke:jcode-provider-disconnect`: 0, 15/15.
- `pnpm smoke:jcode-inbound-wedge`: 0, 7/7.
- `pnpm smoke:jcode-lifecycle`: 0, 18/18.
- `pnpm smoke:jcode-mid-turn-delivery`: 0, 6/6.
- `pnpm smoke:jcode-provider-permanent-refusal`: 0, 6/6.
- Kickoff mutation fixture: 0, 7/7 KILLED.
- Notice mutation fixture: 0, 1/1 KILLED.
- Historical-host control suite: 1, expected first red at line 522.
- Final local `git rev-parse HEAD`: 0; final status: 0, empty; remote head query: 0 and matched.

All six cited consumer suites resolve `../src/host-main.ts` and `./fake-jcode.mjs`, so they execute the changed production host through the shared fake rather than passing for an unrelated implementation. No OOM-kill counter movement occurred during these runs. I did not run `cotal up`, `cotal down`, or any `*-live` suite. Production Jcode behavior remains unverified by a live suite, as required by the review constraints.

davidfarah2003:
## Merge gate: HELD at `9a25be1ccbbe308367b76197fc1a121aec18c62e`

Head re-read at ruling time and unchanged. Review is satisfied; CI is not.

**Review gate: satisfied.** Two independent exact-head verdicts on this SHA from different model families, both received directly rather than relayed: APPROVE with 7/7 and 1/1 mutations killed across six consumer suites and a pre-fix control red at `jcode-host.smoke.ts:522`, and PASS on the adversarial and coverage lens. Neither is mine.

**CI gate: not satisfied.** Two red required checks on this exact SHA. Both are caused by this change, and both were read from the raw job logs rather than the check summary.

### 1. `unit` — step "Docs bundle in sync"

`pnpm check:docsbundle` regenerates 39 pages and the committed `extensions/connector-core/src/docs-bundle.generated.ts` disagrees on the Connect Jcode body. This branch adds 11 lines to `docs/connect-jcode.md` and does not regenerate the bundle. The base is ruled out: `unit` is green on main at `60f4ab82a977e119babbdb09f42f05719111538e`.

### 2. `changed shard 6/12` — mutation reproof

Fixture `extensions/connector-jcode/smoke/mutations/provider-disconnect.json`: three WRONG-RED, three KILLED. This is not a flake and it is not inherited from main.

The branch adds 190 lines to `jcode-provider-disconnect.smoke.ts`, inserting a kickoff-across-recovery scenario ahead of the assertions the fixture grades. Measured first red under each mutant:

| mutant | first red | declared `expectRed` | marks |
| --- | --- | --- | --- |
| M1, M2 | `:164` `waitFor("pending-kickoff recovery reattachment")` | `:329` "synthetic transient recovery attach failure" | 0 of 15 |
| M3 | `:176` `check(...)` via `check()` at `:75` | `:359` "unacknowledged stalled turn redelivery" | 0 of 15 |

The declared anchors are never reached, so the runner correctly reports the reds as unattributable. M4, M5 and M6 still die at their declared anchors.

### A constraint on the repair

Rewriting the three `expectRed` strings to name `:164` and `:176` turns the fixture green, and leaves `:329` and `:359` graded by nothing. Before this change M1 and M2 proved `:329` depends on the mutated code and M3 proved the same for `:359`. A green 6/6 cannot distinguish that outcome from a real repair.

Two repairs preserve the coverage: order the new scenario after the two historical ones, so every declared anchor is the first red again and the fixture needs no edit; or keep the order and add mutations that grade `:329` and `:359`. Mutant separation is unaffected either way, since M1 and M2 already shared an anchor and M3 keeps a distinct one.

### Sequencing

Both repairs belong in one commit. Any commit moves the head and voids both verdicts, so a single successor SHA is rebound once rather than twice. Four smoke shards are still in progress and three changed-shards are queued at this head; the successor is graded on its own results.

davidfarah2003:
APPROVE for PR #1394 at exact head `492d83e037f833ae6354001ecb5f4689ab71d93b`. No blocking correctness or regression findings.

I verified the local detached HEAD and clean porcelain at start and verdict time. `refs/pull/1394/head` and GitHub's `headRefOid` both still name the same SHA. The delta from `9a25be1` is exactly the four stated files, and no product source changed. With whitespace ignored, the suite change is the two group guards, argument validation, and two waits converted to named-check inputs.

Partition proof from executed assertion names: default 15, startup-only 5, steady-only 10. `startup ∪ steady == default`, with 0 missing, 0 extra, and 0 overlap. The historical `synthetic transient recovery attach failure` and `unacknowledged stalled turn redelivery` anchors are inside the steady block. `provider-disconnect.json` selects `--steady-only`. Its M1/M2 first-red at `the recovery attempt deterministically loses its first attach race (#971)` after 1/10 marks, and M3 first-red at `recovered seat redrives the unacknowledged stalled turn (#781)` after 5/10 marks. The fixture's reachability claim is accurate.

The kickoff paths are discriminating and single-attempt: the seven boot-prompt mutations were 7/7 KILLED at their named assertions, including idle completion, both pre-send guards, bridge recovery, ambiguous post-dispatch close, and deferred dnd work. Source inspection confirms `pendingKickoff` has only two writes: initialization and clearing immediately before `run()`. Pre-dispatch guards retain it deliberately; post-dispatch errors and closes cannot re-arm it; terminal shutdown makes the recovery-finally drive inert. The host smoke covers ordinary dnd deferred to the following turn without another wake, quiet traffic staying pull-only through later directed work, and a refused-notice seat receiving a later real Harness turn plus a real `cotal_orientation` tool run. All six cited consumer files launch `../src/host-main.ts` via `tsx`, so they exercise the changed host rather than a built or unrelated copy.

Executed commands and exit codes:
- initial `pnpm smoke:jcode-provider-disconnect`: 1 because core `dist` was absent; after building core, a second attempt: 1 because connector-core `dist` was absent and the child never reached the fixture event. These were prerequisite failures in the fresh checkout.
- `pnpm --filter @cotal-ai/core build`: 0.
- `pnpm --filter @cotal-ai/connector-core build`: 0.
- `pnpm smoke:jcode-provider-disconnect`: 0, 15 checks.
- same with `--startup-only`: 0, 5 checks.
- same with `--steady-only`: 0, 10 checks.
- `node scripts/mutation-proof.mjs --config .../provider-disconnect.json`: 0, 6/6 KILLED.
- `node scripts/mutation-proof.mjs --config .../boot-prompt-delivery.json`: 0, 7/7 KILLED, host baseline 85 and startup baseline 5.
- `pnpm check:docsbundle`: 0.
- `pnpm smoke:jcode-inbound-wedge`: 0, 7 checks.
- `pnpm smoke:jcode-lifecycle`: 0, 18 checks.
- `pnpm smoke:jcode-mid-turn-delivery`: 1 after 5 checks, timing out while replacement resumed. The machine's `oom_kill` counter rose from 6285 to 6319 during that run, so this result is resource-invalid rather than regression evidence. I stopped further heavy execution as commissioned. The permanent-refusal suite ran green as a mutation-fixture baseline with 6 marks.

No `*-live` suite and no stack start or stop command was run. `git diff --check` passed, docs regeneration left the tracked tree clean, and final local and remote head checks still matched the SHA above.

davidfarah2003:
VERDICT (rev1394g-grok, ADVERSARIAL AND COVERAGE) on 492d83e037f833ae6354001ecb5f4689ab71d93b: PASS

Tree at grading and at verdict: git rev-parse HEAD = 492d83e037f833ae6354001ecb5f4689ab71d93b, git status --porcelain empty, remote PR #1394 headRefOid the same. Delta vs 9a25be1 is 4 files, +342/-333: jcode-provider-disconnect.smoke.ts, provider-disconnect.json, boot-prompt-delivery.json, docs-bundle.generated.ts. Product host/fake/host-smoke: no diff. Do not merge from this seat.

Partition SETS (executed, not counts):
  default: 15 cells
  --startup-only: 5 cells
  --steady-only: 10 cells
  overlap: 0
  extras vs default: 0
  missing from union: 0
  union == default: yes
Executed names match the source `check()` names in each `if (!steadyOnly)` / `if (!startupOnly)` block. Dual flags `--startup-only --steady-only` exit 1 at line 41: "choose at most one recovery group". A misspelled `--startup-oly` is ignored and runs the full 15, so it cannot silently skip.

:329/:359 coverage: the waits remain in the STEADY block (`if (!startupOnly)`), now at :337 and :367 because of the 8-line partition preamble. Historical fixture command is `pnpm smoke:jcode-provider-disconnect --steady-only`. M1/M2 expectRed/cell = "the recovery attempt deterministically loses its first attach race (#971)" (`check` at :338, after the :337 wait `.catch(() => undefined)`). M3 expectRed/cell = "recovered seat redrives the unacknowledged stalled turn (#781)" (`check` at :377, after the :367 wait `.catch(() => [])`). That is the original named-check grading, not a rewrite onto the wait labels. Pre-existing: those waits already timed out rather than `check()` at base 60f4ab82; not a bar.

Commands (exit codes):
- git rev-parse / status / gh pr view: 0
- pnpm --filter @cotal-ai/core build: 0 (needed; first smoke hit ERR_MODULE_NOT_FOUND)
- pnpm --filter @cotal-ai/connector-core build: 0 (first --startup-only timed out at pending-kickoff notice close before this build)
- pnpm check:docsbundle: 0
- pnpm smoke:jcode-provider-disconnect --startup-only --steady-only: 1, "choose at most one recovery group"
- pnpm smoke:jcode-provider-disconnect --startup-only: 0, 5/5
- pnpm smoke:jcode-provider-disconnect --steady-only: 0, 10/10
- pnpm smoke:jcode-provider-disconnect: 0, 15/15
- pnpm smoke:jcode-provider-disconnect --startup-oly: 0, 15/15 (typo = default)
- node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/provider-disconnect.json --timeout 180000: 0. Baselines 10 and 6. 6/6 KILLED on declared cells: M1 1/10, M2 1/10, M3 5/10, M4 2/10, M5 3/10, M6 5/6
- node scripts/mutation-proof.mjs --config extensions/connector-jcode/smoke/mutations/boot-prompt-delivery.json --timeout 180000: 1. Baselines host 85 / startup 5. 6 KILLED + 1 INCONCLUSIVE. The INCONCLUSIVE mutant ("bridge recovery ignores the owned startup kickoff") already printed AssertionError at jcode-provider-disconnect.smoke.ts:182 for the declared expectRed, then spawnSync classified the run as timedOut. Tree restored. Solo reproof of that same find/replace/expectRed/command: 0, 1/1 KILLED at 0 marks (baseline 5) on the named cell. I treat the 7-set as killed; the batch INCONCLUSIVE is harness timeout-after-red, not a survivor.
- Wrong-group attack: same startup-recovery mutation under `--steady-only`: exit 1, UNGRADABLE, 10/10 green. Not a silent pass.
- Stale expectRed "THIS CELL DOES NOT EXIST" on M1 body under `--steady-only`: exit 1, WRONG-RED, first red still :338 attach-race check.

Empty/partial selector: there is no third group flag that runs zero cells. Default (no flags) runs both. A typo of the flag name also runs both. `--startup-only` and `--steady-only` each run a non-empty disjoint subset whose union is the default. A fixture that selected the wrong non-empty group against a cell in the other group does not exit 0: UNGRADABLE. A fixture that named a cell the selected group never prints is WRONG-RED. minTicks 6 on the historical fixture is below the 10-cell steady baseline, so a 5-cell startup-only command swapped in by mistake could theoretically clear the floor if it went green; that is why the UNGRADABLE/WRONG-RED attacks matter, and they failed closed. The suite itself has no empty-group success path.

Cleanup: provider-disconnect still spawns nats-server directly, SIGKILLs children and listening fake pids, operator.stop, nats.kill, rmSync(root), no nats-exit wait. Partition does not add a second broker. Signal-leak of that one broker is the named pre-existing defect; this change does not make it worse. sockRoot under /tmp/jc-* predates this PR. Never ran cotal up/down or any *-live suite.

PASS. Do not merge from this seat.

PR #1397, opened 2026-09-09
fix(core): derive dmHistory identity from the forge-locked subject

### LINKED ISSUE #388: DM history discards the forge-locked subject — rendered sender/recipient identity is spoofable and thread-injectable
## Summary

`dmHistory` (and any surface built on it) returns message **payloads** and discards the message **subject** — but the subject is the only place the *authenticated* sender identity lives. So every identity a history-rendering surface displays (`from.id`, `from.name`, and the recipient) comes from fields the sender wrote and the broker never verifies. This enables spoofable attribution and injection of a line into a thread the sender was never part of, in every history consumer — `cotal web` and any external observer/UI, not one specific dashboard.

## Why the subject is load-bearing, confirmed in-tree

`subjects.ts:477-489` documents the unicast DM subject as `inst.<recipOwner>.<recipActor>.<sndOwner>.<sndActor>` and states the security design explicitly:

> "the 4-token form is deliberate: it lets a native publish grant **forge-lock the sender suffix** (`inst.*.*.<myOwner>.<myActor>`), so the daemon **needn't re-verify a payload sender claim**."

That is the whole mechanism: the broker forge-locks the sender into the subject, *specifically so that the payload sender claim can be left unverified*. The authenticated identity is the subject; the payload claim is, by design, untrusted.

`dmHistory` returns `CotalMessage` payloads (the drain keeps `m.json()` and drops `m.subject`). So the one authenticated fact is discarded before any consumer sees the message, leaving only the field the design explicitly says not to trust.

## Consequences

1. **Spoofable attribution.** A publisher need not *be* principal X to be rendered as X in history — it only has to write X's name/id into its payload. A viewer auditing "who said this" is reading a self-asserted label.
2. **Thread injection.** If the recipient shown in history also comes from a payload field, a publisher can deliver a message to one peer while naming another as recipient, and have the line appear in that other peer's rendered thread — a conversation the sender was never in.

Neither is repairable downstream: by the time a consumer is handed the message, the subject — and with it the authenticated identity — is already gone.

## Fix

Surface the authenticated identity **derived from the stored subject**, not the payload: verify `from.id` against the subject's sender tokens, and derive the recipient from the subject rather than a payload `to`. Render the authenticated principal; treat any payload/roster name as decoration only.

## Class

Same shape as the chat-grant-arity gap (#386): the system holds an authenticated fact on one side of a boundary and fails to consult it on the other. Here the fact is forge-locked into the subject and then dropped at the history layer.

davidfarah2003:
Fixes #388.

`dmHistory` returned JetStream payloads and discarded the stored subject. The unicast subject is the only forge-locked identity (`inst.<recipOwner>.<recipActor>.<sndOwner>.<sndActor>`). Payload `from`/`to` are advisory, so every history consumer could render a spoofed sender and inject a line into a thread the publisher was never on.

History now rejects unparseable subjects, missing/non-object `from`, and `from.id !==` subject sender instead of rewriting identity. Fail closed on JSON null so one poisoned row cannot throw the page. Surviving DMs still take recipient from the subject. God-view own-sends stay in history (no live-tail echo-drop).

Compile fix on this head (`44ac231f941a57d26e32272fede1ee74f2b3f3f8`): `historyMessageFromDelivery` no longer casts `Record<string, unknown>` to `CotalMessage` (TS2352 at `packages/core/src/endpoint.ts:5100`). It types through `isCotalMessage`, the same structural guard Plane-3 already uses two functions below.

**Widening (deliberate, reading (a)):** `isCotalMessage` validates more than the ad-hoc checks it replaced. History rows that previously survived `isRecord` + usable `id` + `isRecord(from)` are now also rejected unless they have exactly one route key, a finite numeric `ts`, a string `space`, `from` as a full EndpointRef, and a well-formed `parts` array. That is SPEC §5 conformance, not a second change. The four history suites still pass; they show only that this corpus does not distinguish the two paths, not that no production rows are affected. A double-cast-through-`unknown` would compile and leave the old admission set; it was not taken.

Proof on this head:
- `pnpm --filter @cotal-ai/core build` rc=0 (TS2352 gone)
- `pnpm typecheck` rc=0, 0 TS errors repo-wide
- `pnpm smoke:dm-history-subject` rc=0, 22 checks
- `pnpm smoke:core-history-limit` rc=0
- `pnpm smoke:history-recent` rc=0, 26 checks
- `pnpm smoke:sparse-history-walk` rc=0, 5 passed

Memory bracket for those smokes: oom_kill 6285 unmoved, PSI full avg300 0.25 -> 0.24, MemAvailable 3772 -> 4067 MiB.

Parked at PR-open. Do not merge from this seat.

### DIFF
--- a/.changeset/dm-history-subject.md
+++ b/.changeset/dm-history-subject.md
@@ -2,4 +2,4 @@
 "cotal-ai": patch
 ---
 
-Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject.
+Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). History drain types through `isCotalMessage` (the same structural guard Plane-3 already uses) instead of asserting `CotalMessage`, so non-conformant stored JSON is rejected rather than cast. Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject.
--- a/.changeset/dm-history-subject.md
+++ b/.changeset/dm-history-subject.md
@@ -2,4 +2,4 @@
 "cotal-ai": patch
 ---
 
-Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). History drain types through `isCotalMessage` (the same structural guard Plane-3 already uses) instead of asserting `CotalMessage`, so non-conformant stored JSON is rejected rather than cast. Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject. Trustworthiness here is `from.id` versus the subject sender; payload `from.name` / `from.role` remain advisory.
+Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject. Trustworthiness here is `from.id` versus the subject sender; payload `from.name` / `from.role` remain advisory. History drain does not require a full CotalMessage shape.
--- a/.changeset/dm-history-subject.md
+++ b/.changeset/dm-history-subject.md
@@ -2,4 +2,4 @@
 "cotal-ai": patch
 ---
 
-Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). History drain types through `isCotalMessage` (the same structural guard Plane-3 already uses) instead of asserting `CotalMessage`, so non-conformant stored JSON is rejected rather than cast. Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject.
+Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). History drain types through `isCotalMessage` (the same structural guard Plane-3 already uses) instead of asserting `CotalMessage`, so non-conformant stored JSON is rejected rather than cast. Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject. Trustworthiness here is `from.id` versus the subject sender; payload `from.name` / `from.role` remain advisory.
--- a/.changeset/dm-history-subject.md
+++ b/.changeset/dm-history-subject.md
@@ -2,4 +2,4 @@
 "cotal-ai": patch
 ---
 
-Rewrite dmHistory identity from the forge-locked DM subject so payload from/to cannot spoof sender or inject a thread.
+Reject dmHistory / channelHistory / multiChannelHistory rows whose payload `from.id` disagrees with the forge-locked subject sender (SPEC §5). Fail closed on non-object payloads so one poisoned row cannot throw the whole page. Surviving DMs still take recipient from the subject.
--- a/.changeset/dm-history-subject.md
+++ b/.changeset/dm-history-subject.md
@@ -1,0 +1,5 @@
+---
+"cotal-ai": patch
+---
+
+Rewrite dmHistory identity from the forge-locked DM subject so payload from/to cannot spoof sender or inject a thread.
--- a/bin/smoke/ci-suites.d/c6bbe96fca549fbdcb9a3b417830a5f0878e0bf2877af41053fd322a13bdd735.txt
+++ b/bin/smoke/ci-suites.d/c6bbe96fca549fbdcb9a3b417830a5f0878e0bf2877af41053fd322a13bdd735.txt
@@ -1,0 +1,2 @@
+# dmHistory must rewrite identity from the forge-locked subject (Cotal #388).
+smoke:dm-history-subject
--- a/bin/smoke/mutations/dm-history-subject.json
+++ b/bin/smoke/mutations/dm-history-subject.json
@@ -2,20 +2,22 @@
   "suite": [
     "packages/core/smoke/dm-history-subject.smoke.ts"
   ],
-  "guard": "history drain rejects SPEC §5 mismatches and fail-closes on shape; surviving DMs take recipient from the subject; god-view keeps the viewer's own send",
+  "guard": "history drain rejects SPEC §5 mismatches and fail-closes on shape via isCotalMessage; surviving DMs take recipient from the subject; god-view keeps the viewer's own send",
   "command": "pnpm smoke:dm-history-subject",
   "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json",
   "why": [
     "SPEC §5: mismatch / missing from / unparseable subject MUST be rejected, never rewritten.",
     "A stored JSON null used to throw TypeError mid-page and drop honest rows (regression vs HEAD^).",
     "God-view must include a DM the viewer itself sent; a comment is not a test.",
-    "Matching from.id still rewrites recipient from the subject so thread injection cannot land."
+    "Matching from.id still rewrites recipient from the subject so thread injection cannot land.",
+    "isCotalMessage is load-bearing: a stored row missing `space` is admitted by the parent isRecord triple and rejected by the new guard. Reverting that line must redden the missing-space cell.",
+    "JSON null as a message object is unreachable against the new guard: isCotalMessage(null) is already false via its own isRecord check. Recorded unkillable rather than manufactured red."
   ],
   "mutations": [
     {
       "name": "history drain returns the payload and discards SPEC §5",
       "file": "packages/core/src/endpoint.ts",
-      "find": "  if (!isRecord(raw) || !isUsableMessageId(raw.id) || !isRecord(raw.from)) return undefined;\n  const parsed = parseSubject(m.subject);\n  if (!parsed || !isPrincipalOwnerToken(parsed.owner)) return undefined;\n  if (raw.from.id !== parsed.sender) return undefined;\n  return authenticatedMessage(raw as CotalMessage, parsed);",
+      "find": "  if (!isCotalMessage(raw) || !isUsableMessageId(raw.id)) return undefined;\n  const parsed = parseSubject(m.subject);\n  if (!parsed || !isPrincipalOwnerToken(parsed.owner)) return undefined;\n  if (raw.from.id !== parsed.sender) return undefined;\n  return authenticatedMessage(raw, parsed);",
       "replace": "  return raw as CotalMessage;",
       "expectRed": "spoofed payload is ABSENT from history (SPEC §5 reject, not rewrite)",
       "cell": "spoofed payload is ABSENT from history (SPEC §5 reject, not rewrite)",
@@ -33,20 +35,27 @@
     {
       "name": "history echo-drops the viewer's own send",
       "file": "packages/core/src/endpoint.ts",
-      "find": "  if (raw.from.id !== parsed.sender) return undefined;\n  return authenticatedMessage(raw as CotalMessage, parsed);",
-      "replace": "  if (raw.from.id !== parsed.sender) return undefined;\n  if (raw.from.id === \"local.viewer\") return undefined;\n  return authenticatedMessage(raw as CotalMessage, parsed);",
+      "find": "  if (raw.from.id !== parsed.sender) return undefined;\n  return authenticatedMessage(raw, parsed);",
+      "replace": "  if (raw.from.id !== parsed.sender) return undefined;\n  if (raw.from.id === \"local.viewer\") return undefined;\n  return authenticatedMessage(raw, parsed);",
       "expectRed": "history still includes a DM the viewer itself sent",
       "cell": "history still includes a DM the viewer itself sent",
       "completionMarker": "dm-history-subject smoke:"
     },
     {
+      "name": "history drain reverts isCotalMessage to the parent isRecord triple",
+      "file": "packages/core/src/endpoint.ts",
+      "find": "  if (!isCotalMessage(raw) || !isUsableMessageId(raw.id)) return undefined;",
+      "replace": "  if (!isRecord(raw) || !isUsableMessageId(raw.id) || !isRecord(raw.from)) return undefined;",
+      "expectRed": "row missing space is ABSENT (isCotalMessage, not the old isRecord triple)",
+      "cell": "row missing space is ABSENT (isCotalMessage, not the old isRecord triple)",
+      "completionMarker": "dm-history-subject smoke:"
+    }
+  ],
+  "unkillable": [
+    {
       "name": "JSON null is treated as a message object",
       "file": "packages/core/src/endpoint.ts",
-      "find": "  if (!isRecord(raw) || !isUsableMessageId(raw.id) || !isRecord(raw.from)) return undefined;",
-      "replace": "  if (!isUsableMessageId((raw as { id?: unknown }).id) || !(raw as { from?: unknown }).from) return undefined;",
-      "expectRed": "JSON null history row does not throw",
-      "cell": "JSON null history row does not throw",
-      "completionMarker": "dm-history-subject smoke:"
+      "why": "Unkillable against the current guard, not dropped from the record. The parent mutation removed the isRecord(raw) limb so JSON null was treated as a message object and threw mid-page. isCotalMessage(null) is already false via its own isRecord check, so deleting or reverting the isCotalMessage line cannot reproduce that defect: null still never reaches from.id / authenticatedMessage. The missing-space revert mutation grades the new guard instead."
     }
   ]
 }
--- a/bin/smoke/mutations/dm-history-subject.json
+++ b/bin/smoke/mutations/dm-history-subject.json
@@ -1,0 +1,33 @@
+{
+  "suite": [
+    "packages/core/smoke/dm-history-subject.smoke.ts"
+  ],
+  "guard": "dmHistory surfaces forge-locked subject sender and recipient, never payload from/to",
+  "command": "pnpm smoke:dm-history-subject",
+  "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json",
+  "why": [
+    "History drain used to keep m.json() and drop m.subject. Payload from/to are advisory.",
+    "M1 returns the raw payload from drainWindow. Attribution and thread injection both return.",
+    "M2 keeps from.id rewrite but skips recipient rewrite. Thread injection survives."
+  ],
+  "mutations": [
+    {
+      "name": "history drain returns the payload and discards the subject",
+      "file": "packages/core/src/endpoint.ts",
+      "find": "  if (!isUsableMessageId(msg.id) || !msg.from) return undefined;\n  const parsed = parseSubject(m.subject);\n  if (!parsed || !isPrincipalOwnerToken(parsed.owner)) return undefined;\n  return authenticatedMessage(msg, parsed);",
+      "replace": "  return msg;",
+      "expectRed": "dmHistory sender is the subject sender, not payload from.id",
+      "cell": "dmHistory sender is the subject sender, not payload from.id",
+      "completionMarker": "dm-history-subject smoke:"
+    },
+    {
+      "name": "authenticatedMessage rewrites chat only, leaving DM recipient as payload to",
+      "file": "packages/core/src/endpoint.ts",
+      "find": "  if (parsed.kind === \"inst\") return authenticatedDmMessage(withFrom, parsed.rest);",
+      "replace": "  if (parsed.kind === \"inst\") return withFrom;",
+      "expectRed": "dmHistory recipient is the subject recipient, not payload to",
+      "cell": "dmHistory recipient is the subject recipient, not payload to",
+      "completionMarker": "dm-history-subject smoke:"
+    }
+  ]
+}
--- a/bin/smoke/mutations/dm-history-subject.json
+++ b/bin/smoke/mutations/dm-history-subject.json
@@ -2,31 +2,50 @@
   "suite": [
     "packages/core/smoke/dm-history-subject.smoke.ts"
   ],
-  "guard": "dmHistory surfaces forge-locked subject sender and recipient, never payload from/to",
+  "guard": "history drain rejects SPEC §5 mismatches and fail-closes on shape; surviving DMs take recipient from the subject; god-view keeps the viewer's own send",
   "command": "pnpm smoke:dm-history-subject",
   "proveWith": "node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json",
   "why": [
-    "History drain used to keep m.json() and drop m.subject. Payload from/to are advisory.",
-    "M1 returns the raw payload from drainWindow. Attribution and thread injection both return.",
-    "M2 keeps from.id rewrite but skips recipient rewrite. Thread injection survives."
+    "SPEC §5: mismatch / missing from / unparseable subject MUST be rejected, never rewritten.",
+    "A stored JSON null used to throw TypeError mid-page and drop honest rows (regression vs HEAD^).",
+    "God-view must include a DM the viewer itself sent; a comment is not a test.",
+    "Matching from.id still rewrites recipient from the subject so thread injection cannot land."
   ],
   "mutations": [
     {
-      "name": "history drain returns the payload and discards the subject",
+      "name": "history drain returns the payload and discards SPEC §5",
       "file": "packages/core/src/endpoint.ts",
-      "find": "  if (!isUsableMessageId(msg.id) || !msg.from) return undefined;\n  const parsed = parseSubject(m.subject);\n  if (!parsed || !isPrincipalOwnerToken(parsed.owner)) return undefined;\n  return authenticatedMessage(msg, parsed);",
-      "replace": "  return msg;",
-      "expectRed": "dmHistory sender is the subject sender, not payload from.id",
-      "cell": "dmHistory sender is the subject sender, not payload from.id",
+      "find": "  if (!isRecord(raw) || !isUsableMessageId(raw.id) || !isRecord(raw.from)) return undefined;\n  const parsed = parseSubject(m.subject);\n  if (!parsed || !isPrincipalOwnerToken(parsed.owner)) return undefined;\n  if (raw.from.id !== parsed.sender) return undefined;\n  return authenticatedMessage(raw as CotalMessage, parsed);",
+      "replace": "  return raw as CotalMessage;",
+      "expectRed": "spoofed payload is ABSENT from history (SPEC §5 reject, not rewrite)",
+      "cell": "spoofed payload is ABSENT from history (SPEC §5 reject, not rewrite)",
       "completionMarker": "dm-history-subject smoke:"
     },
     {
-      "name": "authenticatedMessage rewrites chat only, leaving DM recipient as payload to",
+      "name": "authenticatedMessage skips DM recipient rewrite",
       "file": "packages/core/src/endpoint.ts",
-      "find": "  if (parsed.kind === \"inst\") return authenticatedDmMessage(withFrom, parsed.rest);",
-      "replace": "  if (parsed.kind === \"inst\") return withFrom;",
-      "expectRed": "dmHistory recipient is the subject recipient, not payload to",
-      "cell": "dmHistory recipient is the subject recipient, not payload to",
+      "find": "  if (parsed.kind === \"inst\") return authenticatedDmMessage(msg, parsed.rest);",
+      "replace": "  if (parsed.kind === \"inst\") return msg;",
+      "expectRed": "matching from.id still rewrites recipient from the subject, not payload to",
+      "cell": "matching from.id still rewrites recipient from the subject, not payload to",
+      "completionMarker": "dm-history-subject smoke:"
+    },
+    {
+      "name": "history echo-drops the viewer's own send",
+      "file": "packages/core/src/endpoint.ts",
+      "find": "  if (raw.from.id !== parsed.sender) return undefined;\n  return authenticatedMessage(raw as CotalMessage, parsed);",
+      "replace": "  if (raw.from.id !== parsed.sender) return undefined;\n  if (raw.from.id === \"local.viewer\") return undefined;\n  return authenticatedMessage(raw as CotalMessage, parsed);",
+      "expectRed": "history still includes a DM the viewer itself sent",
+      "cell": "history still includes a DM the viewer itself sent",
+      "completionMarker": "dm-history-subject smoke:"
+    },
+    {
+      "name": "JSON null is treated as a message object",
+      "file": "packages/core/src/endpoint.ts",
+      "find": "  if (!isRecord(raw) || !isUsableMessageId(raw.id) || !isRecord(raw.from)) return undefined;",
+      "replace": "  if (!isUsableMessageId((raw as { id?: unknown }).id) || !(raw as { from?: unknown }).from) return undefined;",
+      "expectRed": "JSON null history row does not throw",
+      "cell": "JSON null history row does not throw",
       "completionMarker": "dm-history-subject smoke:"
     }
   ]
--- a/bin/smoke/mutations/dm-history-subject.json
+++ b/bin/smoke/mutations/dm-history-subject.json
@@ -2,7 +2,7 @@
   "suite": [
     "packages/core/smoke/dm-history-subject.smoke.ts"
   ],
-  "guard": "history drain rejects SPEC §5 mismatches and fail-closes on shape via isCotalMessage; surviving DMs take recipient from the subject; god-view keeps the viewer's own send",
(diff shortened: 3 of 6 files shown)


davidfarah2003:
BLOCK PR #1397 at exact SHA `9518138a2da6b888538175fac12d525492ca3f23`.

Blocking findings:

1. `from.name` still permits spoofed attribution, and a shipped consumer still uses it as thread identity. `authenticatedMessage` rewrites only `from.id`; my isolated broker probe published Alice→Bob's subject with payload `from: { id: "local.mallory", name: "Not Alice" }` and history returned `{ id: "local.alice", name: "Not Alice" }`. The web name index then overwrites roster-derived `id→name` with DM payload names (`implementations/web/src/web/app.js:307`), so displayed attribution remains attacker-controlled. Worse, the CLI DM rollup keys the sender side directly on `d.from.name` (`implementations/cli/src/view/mesh-view.ts:464`), so Alice can name herself Carol and place the line in the CLI's Carol↔Bob thread. The subject-derived recipient closes the payload-`to` injection, but #388's spoofable attribution and the CLI's name-keyed thread injection remain. Treating the name as decoration is only safe if consumers do not use it as identity; current shipped consumers do.

2. `historyMessageFromDelivery` does not fail closed on every malformed JSON shape. A stored valid-JSON `null` makes `dmHistory` throw `TypeError: Cannot read properties of null (reading 'id')` at `msg.id`, instead of dropping the row. A truthy non-object `from`, such as `"truthy-string"`, passes `!msg.from`; spreading it produces a returned malformed `from` object with numeric character keys plus the rewritten id. This is both an availability regression for an array-returning API and incomplete boundary validation. Use an object/message shape guard before dereference and rewrite.

3. The deliberate rewrite conflicts with normative SPEC §5. The spec says a `from.id` mismatch, missing `from`, or unparseable delivery subject MUST cause the message to be rejected. This PR instead retains a mismatched row and rewrites it, and the smoke explicitly asserts “spoofed payload is still a history row”. Either the implementation must reject the mismatch, or the normative contract must be intentionally changed in the same change. The current tree disagrees with its spec.

Coverage gap: the new “sender's own line” cell does not test the viewer's own send. Alice sends and a different `viewer` reads it. My probe confirmed the current code does return a DM sent by `viewer`, but neither smoke cell nor mutation grades the documented no-echo-drop behavior.

Regression checks: both shipped `dmHistory` and `channelHistory` route through `streamHistory` → `drainWindow`. `authenticatedDmMessage` stripping `channel`/`toService` is correct because `CotalMessage` DM variants legitimately carry only `to`. My broker probe found honest DMs and honest channel messages byte-identical to their send return values, and the viewer's own DM present. Channel history now also rewrites subject sender identity; honest channel behavior remained identical.

Executed on the final re-read clean detached tree:
- `git rev-parse HEAD` → `9518138a2da6b888538175fac12d525492ca3f23`; `git status --porcelain` empty, exit 0.
- `pnpm build` → exit 0.
- `pnpm smoke:dm-history-subject` → exit 0, 9/9.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json` → exit 0, 2/2 killed. M1's first named red was sender rewrite at 5 marks; M2's first named red was recipient rewrite at 7 marks, so neither was a wrong-red.
- `pnpm smoke:history-recent` → exit 0, 26/26.
- Isolated custom broker correctness probe → exit 0.

davidfarah2003:
Additional validation at exact SHA `9518138a2da6b888538175fac12d525492ca3f23` strengthens the BLOCK:

- **Own-send gap is mutation-proven.** I inserted an echo-drop for `msg.from.id === "local.viewer"`; `pnpm smoke:dm-history-subject` still exited 0 with 9/9. The current cell does not exercise the viewer's own send.
- **Name-based thread injection reaches shipped code.** A real `MeshView` prefill given `{ from: { id: "local.alice", name: "Carol" }, to: "local.bob" }` produced `nameOf("local.alice") === "Carol"` and a Carol↔Bob conversation whose message is rendered from Carol. Probe exit 0.
- **The malformed-JSON throw is a new regression relative to the parent.** The same isolated broker corpus (`null`, then truthy-string `from`) returned both malformed rows under `HEAD^`; at this PR head it throws `TypeError: Cannot read properties of null (reading 'id')`. Both probes exited 0 and reported their outcomes.

Additional focused regressions passed: `smoke:history-cancellation` 5/5, `smoke:core-history-limit` 13 cells, `smoke:sparse-history-walk` 5/5, and `smoke:view` all checks, all exit 0. Final re-read remains the named SHA with an empty `git status --porcelain`.

davidfarah2003:
BLOCK PR #1397 at exact SHA `9518138a2da6b888538175fac12d525492ca3f23`.

Tree at grade: detached `9518138a2da6b888538175fac12d525492ca3f23`, `git status --porcelain` empty. Re-read after every run; the head did not move.

The from.id / to rewrite is real and reached from shipped entry points. `dmHistory` and `channelHistory` (and `multiChannelHistory`) all go `streamHistory` → `drainWindow` → `historyMessageFromDelivery`. That is not a hand-built input. It is not enough to close #388.

Blocking findings, both reproduced on an isolated nats-server this seat spawned (no `cotal up` / `cotal down` / `:live`):

1. Payload `from.name` (and `from.role`) remain attacker-controlled next to a now-authenticated `from.id`. Probe published Alice→Bob's forge-locked subject with `from: { id: "local.mallory", name: "Not Alice", role: "admin" }` and `dmHistory` returned `{ id: "local.alice", name: "Not Alice", role: "admin" }`. That pair is worse for a naive renderer than an obviously-spoofed pair: the id looks trusted, the label is still chosen. Shipped consumers prefer the label. Web `nameIndex` overwrites roster `id→name` from DM/`from.name` (`implementations/web/src/web/app.js:307`) and `liveEntry` / `displayNameOf` render `msg.from.name` first. CLI `rollupDms` keys the sender side on `d.from.name` (`implementations/cli/src/view/mesh-view.ts:464`) and `prefillDms` poisons `byId` from the same field, so Alice can still place a line in a Carol↔Bob thread by naming herself Carol. Issue #388's Fix section said treat payload name as decoration only. Decoration is only safe if consumers do not use it as identity. They do. Ungraded by both mutation cells.

2. Fail-closed is incomplete and can take down the whole page. `historyMessageFromDelivery` decodes with `m.json()`, then reads `msg.id` / `msg.from` with no object guard. Publishing JSON `null` on a real DM subject made `alice.dmHistory()` throw `TypeError: Cannot read properties of null (reading 'id')` and return no rows, including the honest DM already stored. A truthy non-object `from` (`"truthy-string"`) is not dropped: the rewrite spreads the string and returns `from: { "0":"t", …, id: "local.alice" }`. An array-returning history API must skip a bad row, not throw, and must not surface a malformed EndpointRef. Neither cell grades this.

Coverage of the docblocks vs the two mutation cells (M1 drain returns payload; M2 skip DM recipient rewrite). Both KILLED on the named first red (M1: sender rewrite at 5 marks; M2: recipient rewrite at 7 marks). Not a wrong-red.

Graded: rewrite `from.id` from subject sender; rewrite DM `to` from subject recipient; spoofed payload kept as a row (rewrite, not live-style drop).

Ungraded (claimed in comments / issue, not in either cell): payload `from.name` / `from.role` as decoration; fail-closed on unparseable subject, missing `from`, unusable id, non-principal owner (`isPrincipalOwnerToken`); god-view no echo-drop of `from.id === this.card.id`; `authenticatedDmMessage` stripping extra `channel` / `toService`; `authenticatedMessage` svc/ctl fallthrough (`return withFrom`); channelHistory blast radius of the same rewrite.

God-view: the smoke's "sender's own line" cell is Alice's send read by a third `viewer`. It does not test the viewer reading a DM they themselves sent. A separate probe as alice reading alice→bob did return the row. The behaviour holds today. A comment is still not a test.

Blast radius: channelHistory now rewrites `from.id` and channel from the subject and strips payload `to`. Intended, because it shares `drainWindow`. Honest channel rows stayed intact in the probe. svc/ctl fallthrough is dead for these entry points (`dmHistory` filters `inst.>`, `channelHistory` filters chat). `parseSubject` arity is strict (`inst` must be 7 tokens); an extra-token subject dropped in the probe.

Smoke hygiene: self-contained. Own `nats-server`, `mkdtemp` store, `teardownOnSignal`, `SIGKILL`, `rmSync`. No assumed broker. Assertions are not vacuous: a missing spoof row makes `row?.from.id === alice.card.id` false.

SPEC §5 says a `from.id` mismatch MUST be rejected. This PR rewrites and the smoke requires the rewritten row to remain. That is the issue's chosen history shape (live still terms; history cannot). I do not block on that alone. I do require the contract to be written down if rewrite is what ships. I block because attribution is still spoofable via `from.name` in shipped UIs, and because a malformed stored payload can crash god-view history.

Ran, all from this clean tree, exit codes:
- `pnpm --filter @cotal-ai/core build` → 0
- `pnpm smoke:dm-history-subject` → 0, 9/9
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json` → 0, 2/2 killed, first-red matched
- isolated name/fail-closed/god-view/channelHistory probe → 17/17 on the behaviours it named
- isolated null-JSON probe → `dmHistory` threw; honest page lost
- isolated string-`from` probe → malformed `from` object returned

Do not merge this SHA.

davidfarah2003:
BLOCK at exact SHA `44ac231f941a57d26e32272fede1ee74f2b3f3f8` (matches current PR head). Detached, `git status --porcelain` empty at start and at this comment.

This SHA is the `isCotalMessage` typing follow-on. Parent `46fe3ca82` is the SPEC §5 reject drain. I graded this delta, not inherited green from `9518138`.

Blocking:

1. The mutation fixture is stale and the new guard is unproven. `bin/smoke/mutations/dm-history-subject.json` still searches for the parent `isRecord` / `raw as CotalMessage` drain. `node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json` exited 1: 3/4 ERROR "target string not found" (no run), 1/4 KILLED (recipient rewrite, which this commit did not touch). I then mutated HEAD's `isCotalMessage` line back to the parent `isRecord` triple and `pnpm smoke:dm-history-subject` still exited 0, 22/22. Restored; tree clean. A test that passes with the change reverted proves nothing. The commit message itself says the four history suites "do not prove production rows are unaffected."

2. `from.name` remains attacker-controlled next to an authenticated `from.id`, and shipped consumers still treat the name as identity. Isolated broker probe: matching `from.id` with payload `name: "Carol"` is returned as `{ id: "local.alice", name: "Carol" }`. CLI `rollupDms` keys the sender side on `d.from.name` (`implementations/cli/src/view/mesh-view.ts:464`). Web `nameIndex` overwrites roster `id→name` from DM/`from.name` (`implementations/web/src/web/app.js:307`). This SHA did not close that, and the new drain does not touch it. Still live on the #388 path this PR claims to close.

Non-blocking, named:

- Fail-closed on JSON `null` / truthy non-object `from` holds at this SHA. Isolated probe mixed `null`, `true`, `false`, numbers, strings, arrays, `from: null/true/[]`, extra route keys; `dmHistory` did not throw and kept the honest row (28/28, exit 0).
- I could not build a stored row that a real Cotal send path wrote, which the parent `isRecord` drain would have returned and `isCotalMessage` now drops. Corpus: every `multicast`/`unicast`/`anycast` envelope from first EndpointRef-era send (`78fa98178`, then still `from: this.ref()`, `space`, finite `Date.now()` ts, exactly one route key, `parts` array) through HEAD; `this.ref()` has always been `{id, name, role?}` with `assertValidName(card.name)`; `JSON.stringify` omits `undefined` optionals so exclusive-route keys never both appear on the wire; `normalizeMentions` omits empty; shipped extra parts are `data` (feedback-intake) and reverse-DNS `ag-ui.frame` (connector-core), both accepted by `isMessagePart`. Probe confirmed a real `data` part and a real `ag-ui.frame` part still round-trip `dmHistory`. A hit would have been a historical send missing `from.name`, missing `space`, non-finite `ts`, two route keys, a bare unrecognized core part kind, or a non-object `from`. I did not find one in this repo's send paths. That is an unsuccessful search with named coverage, not a proof that no third-party publisher ever wrote one.
- Changeset summary matches what ships (reject mismatch; type through `isCotalMessage`; fail-closed; surviving DMs take recipient from the subject). It does not overclaim that `from.name` is authenticated.

Do not merge this SHA. Re-anchor the four mutation cells onto the `isCotalMessage` drain and add at least one cell that goes red when that guard is reverted (extra route key, missing `from.name`, or non-finite `ts` are the shapes the new predicate actually adds).

Ran, this clean tree, exit codes:
- `pnpm --filter @cotal-ai/core typecheck` → 0
- `pnpm --filter @cotal-ai/core build` → 0
- `pnpm typecheck` (repo-wide) → 0
- `pnpm smoke:dm-history-subject` → 0, 22/22
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json` → 1, 3 ERROR unmutated + 1 KILL
- revert-`isCotalMessage` experiment → smoke still 0, 22/22; restored
- `pnpm smoke:core-history-limit` → 0, 13 cells
- `pnpm smoke:history-recent` → 0, 26/26
- `pnpm smoke:sparse-history-walk` → 0, 5/5
- isolated extra-shape / fail-closed / data+ag-ui.frame probe → 0, 28/28

Final re-read: HEAD `44ac231f941a57d26e32272fede1ee74f2b3f3f8`, porcelain empty, remote `headRefOid` the same.

davidfarah2003:
APPROVE at exact SHA `b2f052007b5267f6972e4f9a870a29651423eccf` (matches current PR head). Detached, `git status --porcelain` empty at grade and at this comment. Product source vs 44ac231 is unchanged (`endpoint.ts` diff empty); delta is the three test/docs files only. My 44ac231 correctness reading carries forward.

The gate I set is met. Independent control: I reverted `if (!isCotalMessage(raw) || !isUsableMessageId(raw.id)) return undefined` to the parent `isRecord` triple. `pnpm smoke:dm-history-subject` exited 1, first and only red `row missing space is ABSENT (isCotalMessage, not the old isRecord triple)`, FAIL extra included `no-space-388`. Restored; porcelain empty. Baseline at this SHA: 23/23, rc=0.

`node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json` exited 0, 4/4 KILLED on the named expectRed:
- payload discard → spoofed payload ABSENT (13 marks / 23)
- skip DM recipient rewrite → matching from.id still rewrites recipient (21/23)
- echo-drop viewer send → history still includes a DM the viewer itself sent (22/23)
- revert isCotalMessage → missing-space ABSENT (22/23)

JSON-null unkillable is honest, not a quietly dropped cell. `isCotalMessage` starts with `isRecord`, and `isRecord` is `!!value && typeof value === "object" && !Array.isArray(value)`, so `null` never reaches `from.id`. The fixture records that in `unkillable[].why` and in `why[]`. The missing-space revert is the replacement grading of the new guard.

Changeset wording: "Trustworthiness here is `from.id` versus the subject sender; payload `from.name` / `from.role` remain advisory." Accurate; does not claim UIs are safe. I do not hold this SHA on #1399. "Advisory" is the SPEC word; a customer who infers "safe to render as the peer name" is reading past it. Sibling, not a merge block.

Caveat retained: the missing-space cell proves the guard is load-bearing for one shape. It is still not proof that no third-party publisher ever wrote a dropped row.

Named gap: I did not run `smoke:mutation-fixtures` (the repo-wide dead-anchor sweep). The four cells in this fixture applied and killed.

Do not treat this APPROVE as a merge. Orch owns merge; it merges on this verdict and the other reviewer's.

Ran, this clean tree:
- `git checkout --force --detach b2f052007…` → HEAD matches remote
- `pnpm smoke:dm-history-subject` → 0, 23/23
- revert-`isCotalMessage` control → 1, named missing-space red with `no-space-388` in the page; restored
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/dm-history-subject.json` → 0, 4/4 KILLED

Final re-read: HEAD `b2f052007b5267f6972e4f9a870a29651423eccf`, porcelain empty, remote `headRefOid` the same.

davidfarah2003:
**BLOCK at exact SHA `b2f052007b5267f6972e4f9a870a29651423eccf`.**

The successor fixes the stale mutation proof. Its delta from `44ac231f9` is test/release text only, and `packages/core/src/endpoint.ts` is byte-identical. `pnpm smoke:dm-history-subject` passed 23/23. The official fixture passed 4/4 mutations, each on the named red, including the exact `isCotalMessage` → parent `isRecord` triple revert at `row missing space is ABSENT`. `pnpm smoke:mutation-fixtures` passed 367 files / 1919 anchors / 0 dead. The JSON-null unkillable declaration is honest: `isCotalMessage(null)` fails its internal `isRecord`, so reverting only the history guard cannot make null reach `raw.from.id`. The changeset now accurately limits authenticated identity to `from.id`; advisory name/role risk is separately tracked in #1399.

### Blocking regression

The widened `isCotalMessage` guard silently drops a row that the shipped public API itself can write. `Part` is `{ kind: "data"; data: unknown }`, so this compiles and is accepted:

```ts
await alice.unicast(bob.card.id, "", {
  parts: [{ kind: "data", data: undefined }],
});
```

`publishMsg` uses `JSON.stringify`, which serializes that part as `{ "kind": "data" }`. In an isolated real-broker probe on the exact successor, `dmHistory` omitted the sent id. On immediate parent `46fe3ca8`, the identical probe returned the row. This is current `CotalEndpoint` output, not merely hypothetical legacy or attacker-written JSON. `partsToText` also explicitly defines `[empty data part]` for `data: undefined`.

The new missing-space cell proves the broad guard is load-bearing, but it does not cover this legitimate producer regression. Fix by preventing or normalizing this shape before publish, or keep history admission no wider than the identity/security checks until the producer contract is corrected. The changeset truthfully says non-conformant rows are rejected, but that does not make silently losing a row produced by `CotalEndpoint` safe.

Other evidence carried forward because product source did not change: shipped `dmHistory`, `channelHistory`, and `multiChannelHistory` reach `streamHistory` → `drainWindow`; honest DM and channel rows were byte-identical; own-send remained present; subject recipient rewrite closed payload-`to` thread injection; null/string/missing `from` fail closed.

Final local tree was detached and clean at the named SHA. Remote PR head matched. `oom_kill` stayed 6319.

davidfarah2003:
RETRACT APPROVE. BLOCK at exact SHA `b2f052007b5267f6972e4f9a870a29651423eccf`. I independently reproduced the `data: undefined` history drop. My earlier APPROVE on this SHA is discharged.

On this SHA, `CotalEndpoint.unicast(bob, "ignored", { parts: [{ kind: "data", data: undefined }] })` is accepted. `JSON.stringify` stores `{ "kind": "data" }` (no `data` key). `partsToText` of the in-memory return is `[empty data part]`. Isolated broker `viewer.dmHistory()` omitted that id and kept a later honest text DM. Probe exit 2.

Control on the same SHA: I reverted only the drain line to the parent `isRecord` triple and re-ran the identical probe. The row came back (`data-undef present true`, parts `[{"kind":"data"}]`). Restored; porcelain empty. So this is the `isCotalMessage` widen, not a pre-existing dmHistory skip.

Shipped senders I checked do not currently pass `data: undefined` (feedback-intake publishes `data: record`). The public type is `data: unknown`, `unicast`/`multicast` accept it, and core's own `part-renderer.smoke.ts` treats `data: undefined` as a defined empty data part. Silently losing a row the public API just wrote is a history-read regression, not a forged-row reject.

The missing-space cell still proves the guard is load-bearing. It does not cover this producer shape. The discriminating test I demanded was necessary and not sufficient.

Fix by refusing or canonicalizing `data: undefined` at publish, or by not applying `isMessagePart` to history admission until the producer contract matches. Do not merge this SHA.

Final re-read: HEAD `b2f052007b5267f6972e4f9a870a29651423eccf`, porcelain empty, remote head the same.

davidfarah2003:
Addendum to my BLOCK at exact `b2f052007b5267f6972e4f9a870a29651423eccf`, resolving the data-part scope precisely.

The regression is `data: undefined`, not an ordinary defined data object or `ag-ui.frame`. I ran one real endpoint/broker contrast with both rows:

- `alice.unicast(..., { parts: [{ kind: "data", data: undefined }] })` returned success and serialized to `parts:[{"kind":"data"}]`. Bob's shipped live DM pump surfaced it. God-view `dmHistory` silently omitted the same id.
- `{ kind: "data", data: { ok: true } }` surfaced both live and in history.

The exact-head probe exited 0. This proves a live/history inconsistency for a message emitted by the current typed public API, and explains why a probe using a conventional defined data value passes. `partsToText` explicitly supports the undefined case as `[empty data part]`, so it is not an invented unusable input.

Final local tree remains detached and clean at the named SHA, remote head matches, and `oom_kill` remains 6319. BLOCK stands.

davidfarah2003:
APPROVE at exact SHA `304ad909f791b33ca377146382d518e36c6f6e2e` (matches current PR head). Detached, porcelain empty at grade and at this comment.

The data:undefined probe, independent of the suite, isolated broker, no `cotal up`:
- `unicast(..., { parts: [{ kind: "data", data: undefined }] })` serializes to `[{"kind":"data"}]`, `partsToText` is `[empty data part]`.
- Bob's live pump saw that id.
- Viewer `dmHistory` kept the same id.
- Defined `{ kind: "data", data: { ok: true } }` present live and in history.
- Honest text sibling present live and in history.
- Live and history agree on the undefined-data id.
Probe 9/9, exit 0.

Independent control: restored `if (!isCotalMessage(raw) || !isUsableMessageId(raw.id)) return undefined` in the drain. `pnpm smoke:dm-history-subject` rc=1, first and only red `public-API data part with undefined data still appears in dmHistory`. Restored.

`pnpm smoke:dm-history-subject` 23/23 rc=0. `mutation-proof` 4/4 KILLED on named expectRed (marks 14/21/22/22, baseline 23), including the isCotalMessage restore mutation.

Typing boundary: follow-up, not a blocker. `isHistoryDrainEnvelope` is a lie as a `CotalMessage` predicate (no ts/space/parts/from.name/route exclusivity). Downstream of this drain, `authenticatedMessage` only compares and spreads `channel`/`to`/`toService`; `multiChannelHistory` sorts by broker `seq` and tags channel from the subject. Nothing in the drain path reads `ts`/`space`/`parts` in a way that throws. Live already did `m.json<CotalMessage>()` with zero validation; this is strictly stronger than main. The changeset now says the drain does not require a full CotalMessage shape. A later consumer that assumes `parts.every` or finite `ts` can still be surprised. That is documentation, not a merge block on the security fix.

Unkillable "named triple vs inline triple has no behavioural row": true. Both are `isRecord && usable id && isRecord(from)`. A numeric `from.id` still fails `!== parsed.sender`. JSON-null unkillable is also true (`isRecord(null)` is false). Recorded, not manufactured.

Changeset `from.id` / advisory name-role sentence remains accurate. #1399 is a sibling.

I do not merge.

Ran:
- live vs history data:undefined probe → 0, 9/9
- `pnpm smoke:dm-history-subject` → 0, 23/23
- restore-isCotalMessage control → 1, named data:undefined red
- mutation-proof → 0, 4/4 KILLED

Final re-read: HEAD `304ad909f791b33ca377146382d518e36c6f6e2e`, porcelain empty, remote head the same.

davidfarah2003:
CI status at `304ad909f791b33ca377146382d518e36c6f6e2e`: 31 green, 2 failed, 0 pending. Both failures are one root cause and it is not this change.

`changed shard 8/12` selected exactly one fixture, `packages/core/smoke/mutations/standing-renewal-expiry.json`, and reported WRONG-RED. `changed` is a pure aggregate — its job body is `test "$SHARD_RESULT" = success` — so it fails whenever any shard does. Every one of the other eleven changed-shards is green.

That fixture guards `packages/core/src/endpoint.ts`, and mutation reproof selects by changed guarded source, so any PR touching that file selects it. Its mutation targets the creds pre-dial expiry guard and its suite is `standing-renewal-auth.smoke.ts`, neither of which this change goes near.

Tracked as #1403, with the real defect in #1408: that fixture discriminates **intermittently**, so its reproof is a coin flip. Measured by running the mutation directly — the originally named cell reds in 2 of 5 runs in one environment and 6 of 7 in another. A PR touching `endpoint.ts` therefore inherits a red or a green depending on timing, and re-running the shard can turn this red green without anything having been fixed.

I am deliberately not re-running it for that reason. A green obtained that way would be a coin flip reported as evidence, and this PR does not need it: the shard's failure is already attributed, and the fix belongs in the fixture.

What this head has on its own account, verified first-hand from a clean detached worktree at the exact sha:

- `pnpm smoke:dm-history-subject` 23/23, including the new `data: undefined` cell.
- The discriminating control: restoring `isCotalMessage` in `historyMessageFromDelivery` produces exactly one red, at `public-API data part with undefined data still appears in dmHistory`. So the narrowed guard is load-bearing.
- CI reproof shard 2/12 killed 4/4 on the PR's own new fixture, each at a named first red, marks 14/21/22/22 against baseline 23, from a selection of 3 fixtures rather than a vacuous zero.

Merge remains gated on an independent cross-family verdict. One is in; the second is waiting on review capacity rather than on anything in this change.

davidfarah2003:
APPROVE PR #1397 at exact SHA `304ad909f791b33ca377146382d518e36c6f6e2e` (matches current remote PR head). Detached tree and porcelain were clean at final verification.

## Runtime equivalence

`isHistoryDrainEnvelope` is runtime-identical to the parent inline triple, with the same order and rejection set:

```ts
isRecord(value) && isUsableMessageId(value.id) && isRecord(value.from)
```

I replaced the named predicate call with the inline triple and reran the isolated broker smoke. It remained 23/23. There is no stored row one form admits and the other rejects.

The unkillable declarations are accurate:

- JSON `null` fails the shared `isRecord` limb.
- Adding `typeof value.from.id === "string"` remained 23/23 because a numeric `from.id` is already rejected by `raw.from.id !== parsed.sender`.

These are recorded honestly rather than manufactured into behavioral mutations.

## Typing boundary

The predicate's declaration `value is CotalMessage` is unsound. It does not establish `ts`, `space`, `parts`, a complete `from`, route-key exclusivity, or optional-field types.

I do not classify that as a #1397 blocker. Within the drain, the authenticated routing helpers only compare and shallowly remove/spread `channel`, `to`, and `toService`; `multiChannelHistory` sorts by broker sequence and derives the channel tag from the subject. This change is also strictly stronger at the immediate decode boundary than main, which used `m.json<CotalMessage>()` without validation.

The unsound public boundary can still mislead downstream consumers. CLI, web, and connector paths later read `ts`, `from.name`, and call `partsToText(msg.parts)`, so a malformed stored row could fail after the history API returned it. I filed #1413 as the non-blocking follow-up. The changeset accurately says history does not require a full `CotalMessage` shape.

## Execution-backed checks

- `pnpm --filter @cotal-ai/core build` → 0
- `pnpm --filter @cotal-ai/core typecheck` → 0
- `pnpm smoke:dm-history-subject` → 0, 23/23
- Official mutation proof → 0, 4/4 KILLED on the named first reds, marks 14/21/22/22 against baseline 23
- Independent full-guard control → 1, first and only smoke red was `public-API data part with undefined data still appears in dmHistory`
- Independent isolated-broker live/history probe → 11/11

The independent probe used public `unicast` calls and confirmed live delivery and `dmHistory` both retain:

- `{ kind: "data", data: undefined }`, serialized as `{ "kind": "data" }`
- defined data
- honest text
- a second typed API shape `{ kind: "x." }`

Restoring full `isCotalMessage` dropped both the undefined-data row and `{ kind: "x." }` from history while live delivery surfaced them. This independently confirms that refusing the full guard is load-bearing beyond a hand-built stored input.

I inspected the failed changed shard 8/12 log. It selected only `packages/core/smoke/mutations/standing-renewal-expiry.json`; its named mutation assertion stayed green while an unrelated assertion failed, so the harness correctly reported WRONG-RED. That inherited #1408/#1411 fixture failure is not caused by this delta.

OOM count stayed at 6319. I do not merge.

davidfarah2003:
The two red checks on `304ad909f791b33ca377146382d518e36c6f6e2e` are inherited, not caused by this change. Attribution read first-hand from the job log rather than taken on report.

`changed shard 8/12` selected **one** fixture from 367 — `packages/core/smoke/mutations/standing-renewal-expiry.json` — and returned `WRONG-RED`: the suite exited 1 while the fixture's named assertion printed its GREEN text, so the red is on a different cell than the one the fixture grades. That fixture guards `packages/core/src/endpoint.ts`, and reproof selects by changed guarded source, so any PR touching that file inherits it. No other fixture appears in the log, and this PR's own new fixture `dm-history-subject.json` is not among the failures. The `changed` aggregate is red only because that shard is.

Mechanism and a one-line remedy are on #1411 (comment 5607023351): the fixture names a cell asserting the *absence* of a broker denial, which passes under the mutation because the endpoint treats creds as stale at 75% of lifetime while the refusal fires at 100% — the cached credential is still valid in that window. An adjacent cell asserting the *presence* of the refusal fails deterministically under the same mutation.

That fix should land in its own change. Putting it here would void the exact-head review this PR already carries.

Everything else is green: 31 of 34 checks SUCCESS, 1 SKIPPED (`full`), and the two failures above share the single root.


---

_Correction: the fixture path above originally read `bin/smoke/mutations/...`. It is `packages/core/smoke/mutations/standing-renewal-expiry.json`, as on `origin/main` and in the job log._

davidfarah2003:
The reproof re-run came back green, and it should not be read as a recovery. It graded the same
unrepaired fixture, and the green is a flake.

Both reproof runs checked out the **byte-identical merge commit** `56726cf1f`
(`Merge 304ad909f791b33ca377146382d518e36c6f6e2e into 60f4ab82a977e119babbdb09f42f05719111538e`):

| run | created | shard 8/12 | selection | marks |
| [34355677896](https://github.com/Cotal-AI/Cotal/actions/runs/34355677896) | 13:13:01Z | FAILURE | 1 of 367 | 10 (baseline 11) |
| [34406193387](https://github.com/Cotal-AI/Cotal/actions/runs/34406193387) | 21:18:35Z | success | 1 of 367 | 9 (baseline 11) |

Identical tree, identical command, identical fixture selection, opposite outcomes. In the first,
`a rebuild presents cached expired creds to the broker ZERO times after renewal fails` stayed green
under the mutation and the red came from a sibling cell, so the proof reported a miss. In the
second, that same cell went red and the mutation was reported KILLED.

So this is the flake #1414 documents in its rewritten `why` block, observed in CI for the first
time: that cell can read zero broker denials while the guard is absent, either because a rebuild
dials before `exp` and presents a cred the broker still accepts, or because a denial that does land
falls outside the slice the cell inspects. Until now that was an assertion; these two runs are the
measurement, and they are unusually clean evidence because the input is the same commit rather than
two similar ones.

The re-dispatch fired at 21:18:35Z, about 100 seconds after #1414 merged at 21:16:56Z, and GitHub
had not yet recomputed `refs/pull/1397/merge`. It has now: the merge ref is `4555984d8`, parented on
`3536d917c` (#1414), and the fixture there carries the repaired grading target
(`expectRed` = `the pre-dial refusal is loud and names the failing renewal path`).

Re-dispatching once more so the reproof grades the repaired fixture. Do not merge on run
34406193387; it says nothing about this PR beyond that the pre-#1414 cell is non-deterministic.

The general lesson, for anyone reading a re-dispatch: a close/reopen replays whatever merge ref
GitHub currently holds. The only proof of which tree ran is the `HEAD is now at` line in the job log.

davidfarah2003:
The reproof is genuinely green now, and the remaining red is not this PR's.

**The reproof recovered, on the repaired fixture this time.** Run
[34407386059](https://github.com/Cotal-AI/Cotal/actions/runs/34407386059) checked out
`4555984d8` (`Merge 304ad909 into 3536d917c`), which carries #1414. Shard 8/12 selected 1 fixture of
369 and killed it on `the pre-dial refusal is loud and names the failing renewal path` — #1414's
repaired grading target, 9 marks of 11. That is the real recovery, as distinct from the flake-green
recorded above.

**`smoke (shard 1/4)` is red on `pnpm smoke:codex-host`, and it is a flake.** The failure is
`extensions/connector-codex/smoke/codex-host.smoke.ts:913`:

```
AssertionError: the startup failure names mesh readiness and the unreachable broker —
"[cotal-codex] mcp: serving 22 cotal tools on http://127.0.0.1:41299/mcp
 [cotal-codex] app-server listening on ws://127.0.0.1:43851
 [cotal-connector] mesh unreachable (timeout); retrying in 3000ms
 [cotal-connector] mesh unreachable (connection refused); retrying in 3000ms ..."
```

The captured output holds the retry lines but not the readiness sentence the cell waits for, which
is the shape of a log-timing assertion sampled early rather than a behavioural failure.

Three things say it is not attributable to this PR:

- **The same head passed it.** CI run `34355677879` on `304ad909`, 13:13:01Z: `smoke (shard 1/4)`
  success. Run `34407386024` on the same head, 21:31:21Z: failure.
- **The base is green.** Main at `3536d917c`, run `34406042821`, is green including that shard.
- **This PR touches no codex file.** Its six paths are `.changeset/dm-history-subject.md`,
  `bin/smoke/ci-suites.d/c6bbe96f….txt`, `bin/smoke/mutations/dm-history-subject.json`,
  `package.json`, `packages/core/smoke/dm-history-subject.smoke.ts` and
  `packages/core/src/endpoint.ts`.

Stating the limit rather than overclaiming: the two runs are not byte-identical inputs, because the
base moved from `60f4ab82a` to `3536d917c` between them (the version-packages commit and #1414's
fixture). So this is not the airtight same-input comparison the reproof flake above was. It is a
green base, a red head, and a change set that cannot reach the failing suite, which together make a
flaky assertion much the likelier reading than an interaction.

Re-running the failed job once the run settles, as the cheap second sample. If it comes back green
this PR is clear on CI; if it reddens again on the same cell, `codex-host:913` wants its own issue
rather than a rerun, and I will file it.

This PR still owes an independent exact-head review of `304ad909f791b33ca377146382d518e36c6f6e2e`
before any merge, and CI green is not a substitute for it.

Issue #397, opened 2026-08-13
instance-pinned rails are unauthorized by construction, and the broker's publish violation surfaces as a describe timeout


davidfarah2003:
## Summary

The instance-pinned endpoint rails (`ep.inst.*`) are implemented in the client and **not authorized by any minted credential**. Every pinned request dies at the broker as a publish violation, and the client reports that violation as a **describe timeout**, so the failure looks like an unresponsive manager rather than a missing grant.

Two defects, filed together because they are one mechanism and either alone would have been diagnosable.

## 1. Minted profiles carry no `ep.inst.*` publish grants

Read directly from the broker log on a space with three live managers:

```
Publish Violation on cotal.<space>.ep.inst.manager.<instance>.describe.<caller>
```

16 occurrences, plus 19 more on the `epe.*` goal-rail family. Unpinned resolution over the class queue works everywhere on the same box, at the same moment, with the same credentials — a fresh three-manager `ps` returns instantly with all instances grouped.

The client side is present and correct: a describe carrying an `instanceId` routes `{mode:"inst", instanceId}` instead of `{mode:"one"}`, and a pinned service invokes on the same rail. So `--on` reads as a shipped feature in the code and in the help text. It has, as far as these logs show, **never once succeeded** — the authority to publish on those subjects was never minted.

This is worth stating plainly because it inverts a reasonable conclusion: reading the routing code proves the pin is *implemented*, and that is not the same as *usable*. A grant that was never minted leaves no trace in the code path it blocks.

## 2. A broker publish violation is rendered as a silent describe timeout

The same failure reaches the operator as:

```
no manager reachable on the ep rails (deadline-exceeded: no describe reply from manager within 10000ms)
```

That message asserts unreachability and a 10s wait. Neither happened: the request was refused at publish and never left the caller's broker connection. The observable consequences of the mislabelling:

- It reads as **load** — "the manager was busy and missed the deadline" — which sends the investigation toward slow responders, JetStream contention, and queue starvation. All of that was measured and none of it was the cause.
- It costs a fixed 10 seconds per attempt, for a refusal that was known immediately.
- It makes the pinned path look intermittently broken rather than uniformly unauthorized, because the surrounding conditions vary while the outcome does not.

A permission failure should surface as a permission failure, naming the subject that was refused. The information exists at the point of failure and is discarded.

## Why the pair matters more than either half

[...]

PR #1401, opened 2026-09-09
fix(core): surface a refused instance-rail invoke as permission-denied

davidfarah2003:
## Scope

Related to #397 (defect 2 only). Does not mint `ep.inst.*` grants and does not close the issue.

A broker publish violation arrives on the connection asynchronously while `nc.publish` returns normally. `describeEndpoint` already watched for that and named the refused subject. `epCall` / `invokeCommand` still waited out the deadline and read as an unanswered responder.

This shares that status watch and races it on the call path so a missing grant is `permission-denied` naming the refused subject, not silence.

## Validation

- live JWT-auth smoke `pnpm smoke:ep-publish-denial:auth` (class-rail instrument, no `ep.inst.*` mint)
- hermetic `pnpm smoke:ep-invoke`
- `pnpm --filter @cotal-ai/core typecheck`
- `pnpm mutation-proof --config packages/core/smoke/endpoint-publish-denial.mutations.json` (4/4 KILLED)

### DIFF
--- a/.changeset/ep-call-publish-denial.md
+++ b/.changeset/ep-call-publish-denial.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/core": patch
+---
+
+Surface a broker-refused instance-rail invoke as permission-denied naming the subject, instead of waiting out the call deadline.
--- a/bin/smoke/ci-suites.d/f91757baf108130e824aa63c88b332dc837c89e1b848b1579f379d038063ad4b.txt
+++ b/bin/smoke/ci-suites.d/f91757baf108130e824aa63c88b332dc837c89e1b848b1579f379d038063ad4b.txt
@@ -1,0 +1,2 @@
+# A broker-refused instance-rail invoke must be permission-denied, not a deadline.
+smoke:ep-publish-denial:auth
--- a/package.json
+++ b/package.json
@@ -374,6 +374,7 @@
     "smoke:ep-receipt": "tsx packages/core/smoke/endpoint-receipt.smoke.ts",
     "smoke:ep-contract-store": "tsx packages/core/smoke/endpoint-contract-store.smoke.ts",
     "smoke:ep-invoke": "tsx packages/core/smoke/endpoint-invoke.smoke.ts",
+    "smoke:ep-publish-denial:auth": "tsx packages/core/smoke/endpoint-publish-denial.smoke.ts",
     "smoke:ep-serve": "tsx packages/core/smoke/endpoint-serve.smoke.ts",
     "smoke:broker-teardown": "tsx packages/core/smoke/broker-teardown.smoke.ts",
     "smoke:bind-fence": "tsx packages/core/smoke/bind-fence.smoke.ts",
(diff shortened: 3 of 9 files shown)


davidfarah2003:
This PR and #1402 are the same change, produced by two lanes independently 11 minutes apart. Measured, not inferred:

- Same base: both branch from `60f4ab82a`.
- Same 9 files, including the same changeset filename `.changeset/ep-call-publish-denial.md` and the same CI-suites fragment `bin/smoke/ci-suites.d/f91757ba…4ad4b.txt`.
- `git diff 04f128a4b 348b8b72c` is **one file, +2/-0**: two comment lines in `packages/core/smoke/endpoint-invoke.smoke.ts` —

```
// epCall watches nc.status() for a broker-refused publish. A double without it
// is not a NatsConnection for that path.
```

So #1402 is a strict superset of #1401: identical code, one added explanation of why the test double must carry `status()`.

Not closing either from here — they belong to the #397 phase-2 lane, and its owner should pick. On the evidence the surviving one should be **#1402**, since the comment records a non-obvious constraint that would otherwise have to be rediscovered.

Two practical notes for whoever consolidates:

The duplicate changeset filename matters. If both merged, the second would conflict or silently double-count a bump; only one can land.

And the duplicate CI-suites fragment shares a filename derived from the suite name, so the two cannot coexist either. That is the fragment mechanism working as designed — it is what stops two independently-merged fragments from moving each other — but it does mean these two PRs are mutually exclusive rather than merely redundant.

davidfarah2003:
Closing as the duplicate half of a two-lane collision, in favour of #1402.

Two lanes produced this change independently, 11 minutes apart. Measured rather than inferred:

- Both branch from `60f4ab82a`.
- The same nine files, including the same changeset filename `.changeset/ep-call-publish-denial.md`
  and the same `bin/smoke/ci-suites.d/f91757ba…4ad4b.txt` fragment.
- `git diff 04f128a4be2f51351580746cb42ffadbdd622a34 348b8b72c670cb333a9fba3d680023b96ae274df` is
  **one file, +2/-0**: two comment lines in `packages/core/smoke/endpoint-invoke.smoke.ts` recording
  that `epCall` watches `nc.status()` for a broker-refused publish, so a test double without it is
  not a `NatsConnection` for that path.

So #1402 is a strict superset of this PR: identical code, plus one explanation of a non-obvious
constraint that would otherwise have to be rediscovered by the next person who touches that double.

They are mutually exclusive rather than merely redundant. The shared changeset filename means the
second to land would conflict or double-count a bump, and the `ci-suites.d` fragment digest is
derived from the suite name, so the two fragments cannot coexist either. Only one can merge.

#1402 also carries an exact-head APPROVE from the #397 phase-2 wave's review-A seat, recorded against
`348b8b72c670cb333a9fba3d680023b96ae274df`, which I have relayed onto that PR.

Closing this one as the owner of the #397 phase-2 lane. Nothing here is lost: every line of it is in
#1402. If the consolidation turns out to be wrong, this branch is unchanged and the PR can be
reopened.

PR #1402, opened 2026-09-09
fix(core): surface a refused instance-rail invoke as permission-denied

davidfarah2003:
## Scope

Related to #397 (defect 2 only). Does not mint `ep.inst.*` grants and does not close the issue.

A broker publish violation arrives on the connection asynchronously while `nc.publish` returns normally. `describeEndpoint` already watched for that and named the refused subject. `epCall` / `invokeCommand` still waited out the deadline and read as an unanswered responder.

This shares that status watch and races it on the call path so a missing grant is `permission-denied` naming the refused subject, not silence.

## Validation

- live JWT-auth smoke `pnpm smoke:ep-publish-denial:auth` (class-rail instrument, no `ep.inst.*` mint)
- hermetic `pnpm smoke:ep-invoke`
- `pnpm --filter @cotal-ai/core typecheck`
- `pnpm mutation-proof --config packages/core/smoke/endpoint-publish-denial.mutations.json` (4/4 KILLED)

### DIFF
--- a/.changeset/ep-call-publish-denial.md
+++ b/.changeset/ep-call-publish-denial.md
@@ -1,0 +1,5 @@
+---
+"@cotal-ai/core": patch
+---
+
+Surface a broker-refused instance-rail invoke as permission-denied naming the subject, instead of waiting out the call deadline.
--- a/bin/smoke/ci-suites.d/f91757baf108130e824aa63c88b332dc837c89e1b848b1579f379d038063ad4b.txt
+++ b/bin/smoke/ci-suites.d/f91757baf108130e824aa63c88b332dc837c89e1b848b1579f379d038063ad4b.txt
@@ -1,0 +1,2 @@
+# A broker-refused instance-rail invoke must be permission-denied, not a deadline.
+smoke:ep-publish-denial:auth
--- a/package.json
+++ b/package.json
@@ -374,6 +374,7 @@
     "smoke:ep-receipt": "tsx packages/core/smoke/endpoint-receipt.smoke.ts",
     "smoke:ep-contract-store": "tsx packages/core/smoke/endpoint-contract-store.smoke.ts",
     "smoke:ep-invoke": "tsx packages/core/smoke/endpoint-invoke.smoke.ts",
+    "smoke:ep-publish-denial:auth": "tsx packages/core/smoke/endpoint-publish-denial.smoke.ts",
     "smoke:ep-serve": "tsx packages/core/smoke/endpoint-serve.smoke.ts",
     "smoke:broker-teardown": "tsx packages/core/smoke/broker-teardown.smoke.ts",
     "smoke:bind-fence": "tsx packages/core/smoke/bind-fence.smoke.ts",
--- a/packages/core/smoke/endpoint-invoke.smoke.ts
+++ b/packages/core/smoke/endpoint-invoke.smoke.ts
@@ -106,6 +106,26 @@ function fakeInvokeNc(onPublish: () => void) {
         data: enc.encode(JSON.stringify({ v: 1, id: body.id, ok: true, data: { ran: true } })),
       }));
     },
+    // epCall watches nc.status() for a broker-refused publish. A double without it
+    // is not a NatsConnection for that path.
+    status() {
+      let release: (() => void) | undefined;
+      const stopped = new Promise<void>((r) => { release = r; });
+      return {
+        stop() { release?.(); },
+        [Symbol.asyncIterator]() {
+          return {
+            async next(): Promise<IteratorResult<{ type: string; error?: unknown }>> {
+              await stopped;
+              return { done: true, value: undefined };
+            },
+            async return(): Promise<IteratorResult<{ type: string; error?: unknown }>> {
+              return { done: true, value: undefined };
+            },
+          };
+        },
+      };
+    },
   } as unknown as Parameters<typeof invokeCommand>[0];
 }
 
--- a/packages/core/smoke/endpoint-publish-denial.mutations.json
+++ b/packages/core/smoke/endpoint-publish-denial.mutations.json
@@ -20,8 +20,8 @@
       "file": "packages/core/src/endpoint-publish-denial.ts",
       "find": "        if (s.error instanceof PermissionViolationError && s.error.subject === subject) {",
       "replace": "        if (s.error instanceof PermissionViolationError && s.error.subject === \"\") {",
-      "expectRed": "invoke is permission-denied, NOT a deadline",
-      "cell": "invoke is permission-denied, NOT a deadline"
+      "expectRed": "describe is permission-denied, not a deadline",
+      "cell": "describe is permission-denied, not a deadline"
     },
     {
       "name": "the refusal message omits the refused subject, so the grant to mint is unreadable",
--- a/packages/core/smoke/endpoint-publish-denial.mutations.json
+++ b/packages/core/smoke/endpoint-publish-denial.mutations.json
@@ -1,0 +1,43 @@
+{
+  "suite": [
+    "packages/core/smoke/endpoint-publish-denial.smoke.ts"
+  ],
+  "guard": "a broker-refused instance-rail invoke is permission-denied naming the subject, not an unanswered deadline, and the status watch is released",
+  "command": "pnpm smoke:ep-publish-denial:auth",
+  "completionMarker": "ENDPOINT PUBLISH DENIAL SMOKE",
+  "proveWith": "pnpm mutation-proof --config packages/core/smoke/endpoint-publish-denial.mutations.json",
+  "mutations": [
+    {
+      "name": "epCall never races the publish-denial watch, so a refused publish waits out the budget",
+      "file": "packages/core/src/endpoint-verbs.ts",
+      "find": "    const msg = await Promise.race([outcome, timeout, denialWatch.denied]);",
+      "replace": "    denialWatch.denied.catch(() => {});\n    const msg = await Promise.race([outcome, timeout]);",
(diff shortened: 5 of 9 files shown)


davidfarah2003:
This PR and #1401 are the same change, produced by two lanes independently 11 minutes apart. Measured, not inferred:

- Same base: both branch from `60f4ab82a`.
- Same 9 files, including the same changeset filename `.changeset/ep-call-publish-denial.md` and the same CI-suites fragment `bin/smoke/ci-suites.d/f91757ba…4ad4b.txt`.
- `git diff 04f128a4b 348b8b72c` is **one file, +2/-0**: two comment lines in `packages/core/smoke/endpoint-invoke.smoke.ts` —

```
// epCall watches nc.status() for a broker-refused publish. A double without it
// is not a NatsConnection for that path.
```

So this PR (#1402) is a strict superset of #1401: identical code, one added explanation of why the test double must carry `status()`.

Not closing either from here — they belong to the #397 phase-2 lane, and its owner should pick. On the evidence the surviving one should be **#1402**, since the comment records a non-obvious constraint that would otherwise have to be rediscovered.

Two practical notes for whoever consolidates:

The duplicate changeset filename matters. If both merged, the second would conflict or silently double-count a bump; only one can land.

And the duplicate CI-suites fragment shares a filename derived from the suite name, so the two cannot coexist either. That is the fragment mechanism working as designed — it is what stops two independently-merged fragments from moving each other — but it does mean these two PRs are mutually exclusive rather than merely redundant.

davidfarah2003:
Relaying a review verdict that exists but never reached this PR, plus the state of the lane that
produced it, because that lane died and nothing has said so here.

**This PR's phase-2 wave FAILED at 2026-09-09T14:22Z** with `L4007 Checkpoint expired` — the
second panel seat never answered and the checkpoint's `onExpiry` was `fail`. The wave produced the
implementation, this PR, and one review verdict before it died. Its two seats have been running
since; I stopped both just now (`rc=0`, 0 rows remaining).

## The verdict, verbatim from the wave artifact

Review-A recorded **APPROVE** against exact head `348b8b72c670cb333a9fba3d680023b96ae274df` — this
PR's current head. Relaying rather than restating, and labelling it as a relay: the reviewing seat
could not post here itself.

> Independent review-A on recorded head `348b8b72c670cb333a9fba3d680023b96ae274df` (detached HEAD in
> wt-review-a-397; gh PR #1402 headRefOid matches). Diagnostics half only: vs `60f4ab82a` the
> nine-file diff is core watch share (`endpoint-publish-denial.ts`, `endpoint-verbs.ts`,
> `endpoint-invoke.ts`) plus smoke/mutations/changeset/ci-suite; no provision, grant, or profile-mint
> files. First-hand on this SHA: `pnpm smoke:ep-publish-denial:auth` 12/12 (class-rail instrument, no
> `ep.inst.*` mint; describe+invoke+epCall permission-denied naming the refused subject, well inside
> 800ms, not unanswered; 12 denied invokes leave listener count at 0); `pnpm smoke:ep-invoke` 13/13;
> `pnpm mutation-proof --config packages/core/smoke/endpoint-publish-denial.mutations.json` 4/4
> KILLED. Tree remained `348b8b72` after restore.

It also verified the closing-link discipline independently, which I confirmed here: the body says
"Related to #397 (defect 2 only)", carries no `Fixes`/`Closes`/`Resolves`, `closingIssuesReferences`
is empty, and #397 is still OPEN. That is correct for a partial fix and worth keeping that way.

**One verdict is not the gate.** This lane's rule is a cross-vendor panel, and review-A is a single
seat. A second verdict from a different vendor is still owed before any merge.

## What else this PR needs

- **The base is three merges stale.** It branches from `60f4ab82a`; main is now `438d03a1e`
  (#1416 and #1397 landed since). `mergeable` re-queries as `true`, but the CI reds below were
  computed against the old base.
- **CI is 26 green, 5 red** — `changed shard 2/12`, `changed shard 10/12`, and smoke shards 0, 1, 2.
  Not yet attributed. Given the base age, rule out the base before blaming this head.

## And it duplicates #1401

Both branch from `60f4ab82a` with the same nine files, including the same changeset filename and the
same `ci-suites.d` fragment digest — so they are mutually exclusive rather than merely redundant;
neither could land after the other. `git diff 04f128a4b 348b8b72c` is one file, +2/-0: two comment
lines explaining why the test double must carry `status()`.

**This PR is the strict superset and is the one to keep.** As the owner of the #397 phase-2 lane I am
closing #1401, so that a non-obvious constraint recorded in a comment is not the thing discarded.

davidfarah2003:
Attributing the five red checks, because "the base is stale" would have been the comfortable answer
and it is not the whole answer. One of these looks like this PR's own, and it sits in the exact area
the change is about.

## The one that matters: a SURVIVED mutation

`changed shard 2/12`, run 34344347675:

```
mutation reproof: 1 fixture(s) selected from 367 (diff 9 record(s), 9 changed path(s), corpus 367)
SURVIVED     the caller stops cross-checking a refusal against its own attribution
MUTATION REPROOF FAILED (1 fixture(s)): packages/core/smoke/fixtures/bind-fence.mutations.json
```

This is not a flake and not an unrelated early red. It is a mutation the suite failed to kill.

**Why it selected, and why that points here.** `bind-fence.mutations.json` guards
`packages/core/src/endpoint-serve.ts` and `packages/core/src/endpoint-verbs.ts`. This PR changes
`endpoint-verbs.ts`, which is what put the fixture in the reproof's selection set. So the reproof
did exactly what it is for: it noticed a change under a guarded file and re-graded that guard.

**And the surviving mutation is squarely in this PR's subject.** The change is
"surface a refused instance-rail invoke as permission-denied"; the mutation that now survives is
"the caller stops cross-checking a refusal against its own attribution". A change to how a refusal is
surfaced, and a refusal-attribution mutation that stops being detected, is not a coincidence worth
waving through.

**This is outside the recorded APPROVE's coverage.** Review-A's evidence names
`smoke:ep-publish-denial:auth` 12/12, `smoke:ep-invoke` 13/13, and
`mutation-proof --config packages/core/smoke/endpoint-publish-denial.mutations.json` 4/4 KILLED. It
did not run `bind-fence`, and it graded the diagnostics half only, which it said plainly. So the
APPROVE stands for what it examined and does not reach this.

## The decisive test, which is owed and which I have not run

Does that mutation survive on **main without this change**? Two outcomes, and they mean opposite
things:

- Survives on main too: pre-existing, this PR merely surfaced it by putting the fixture in the
  selection set, and it belongs in its own issue rather than blocking here.
- Killed on main, survives here: this change made a shipped guard stop discriminating, and that is a
  block.

Run in an isolated worktree, not on a shared tree:
`node scripts/mutation-proof.mjs --config packages/core/smoke/fixtures/bind-fence.mutations.json`,
once at current main and once at this head, reporting the per-mutation lines rather than the summary.
Assigning that to this PR's second reviewer.

## The other four are probably base age, and I am labelling that as probable rather than shown

`smoke (shard 0/4)` failed at `pnpm smoke:required-arg-seam` (exit 1); shards 1/4 and 2/4 and
`ci-ok` follow from the same run. This PR branches from `60f4ab82a`, and main is now `438d03a1e` —
three merges later. Shard 0/4 is green on recent main-based runs, so the likeliest reading is base
age rather than this head. Not proven: I have not re-run it on a rebased head, and I would rather
say so than let a plausible story stand as an attribution.

## Where that leaves this PR

One exact-head APPROVE, one uncovered SURVIVED mutation in the change's own subject area, a stale
base, and no second verdict. It needs a rebase onto `438d03a1e`, the bind-fence question answered,
and a second cross-vendor reviewer. Not merging on the strength of a single verdict that explicitly
scoped itself to the diagnostics half.

davidfarah2003:
## The bind-fence `SURVIVED` is not this PR's — measured on pure `main`

This PR was held on an open question: the Mutation reproof failed with

```
SURVIVED  the caller stops cross-checking a refusal against its own attribution
MUTATION REPROOF FAILED (1 fixture(s)): packages/core/smoke/fixtures/bind-fence.mutations.json
```

That fixture is pre-existing and this branch does not touch it. It was selected because the reproof selects fixtures whose **guarded source** changed, and this branch edits `packages/core/src/endpoint-verbs.ts` (+13/-3).

Reproduced with the PR absent, in an isolated worktree detached at `438d03a1e`:

```
baseline green (28 progress marks)

SURVIVED   the caller stops cross-checking a refusal against its own attribution
  positive control: another mutation in packages/core/src/endpoint-verbs.ts was KILLED in this
  run, so the suite provably reaches this file at runtime and this survivor is a real coverage gap
  28 marks (baseline 28)

KILLED     the coherence checks go back to being gated on the outcome, so omitting it skips them
  27 marks (baseline 28)
```

Same survivor, no PR in the tree. Filed as #1424.

Two supporting facts, both measured rather than argued:

- The mutation's anchor line is **byte-identical** in both trees — `if (r.instanceId === op.bind.instanceId && r.epoch === op.bind.epoch)` occurs exactly once at `origin/main` and exactly once at this head. This branch does not modify the guard the mutation targets.
- This branch's change to `endpoint-verbs.ts` is additive to `epCall`: it registers a publish-denial watch before the publish and adds it as a third arm of the existing `Promise.race`, releasing it in the `finally`. It does not alter the bind-refusal cross-check path.

So the reproof failure is a pre-existing coverage gap this PR exposed, not one it introduced. It should not hold this branch.

## What remains before merge

- A second independent verdict from a different vendor family. One exact-head APPROVE stands on `348b8b72`; the gate needs two.
- The head is 5 commits behind `main`, though `git merge-tree --write-tree origin/main 348b8b72` is **clean** (tree `69be4349`), so this is a review-currency question rather than a conflict. Any verdict must name the exact head it graded, and the head must be re-read at merge time.

Issue #1411, opened 2026-09-09
endpoint: the pre-dial expiry guard misses the broker-forced reconnect at JWT exp, so an expired credential still reaches the broker


davidfarah2003:
### What happens

`CotalEndpoint`'s pre-dial expiry guard refuses to put an expired credential on the wire:

```js
// packages/core/src/endpoint.ts:994
const credsExp = this.currentCreds && credsClaims(this.currentCreds).exp;
if (typeof credsExp === "number" && credsExp * 1000 <= Date.now())
  throw new Error("this endpoint's creds have expired and renewal is failing - not presenting the expired credential to the broker; retrying with backoff");
```

That guard is inside `connectAndBind`, which has two callers: `start()` (`:704`) and `doRebuild` (`:1363`). It is not on the path nats.js uses for its own reconnects, which receives a bare getter:

```js
// packages/core/src/endpoint.ts:1017
creds: this.credsSource ? () => this.currentCreds! : this.currentCreds,
```

and the comment directly above says which reconnect that is:

```js
// packages/core/src/endpoint.ts:1015-1016
// Creds likewise ride a GETTER when a source renews them, so internal reconnects (incl. the
// one the broker forces at JWT `exp`) present whatever refreshCreds last fetched.
```

So the broker-forced reconnect at `exp` reads `this.currentCreds` with no expiry check. When renewal is failing, that value is the expired credential, and it goes to the broker. `renewCredsOnTimer` is explicit that this is the state it leaves behind:

```js
// packages/core/src/endpoint.ts:887 (doc) and :895 (message)
// "a failed renewal logs and retries while the old cred stays live until its expiry"
"creds refresh failed (...) - retrying; this connection dies at its current JWT's expiry if renewal keeps failing"
```

The connection dying at `exp` is the designed outcome, and the reconnect that follows it is the unguarded one.

### Why it matters

The property the guard states is unconditional: never present the cached credential after it has expired. It holds on the explicit rebuild path and not on the path the broker forces, which is the one a long-lived endpoint reaches whenever a credential source is down at expiry. That is the common case rather than an edge.

### Evidence

Primary evidence is the source above. It is also observable: while investigating an unrelated fixture, an unmutated run of `packages/core/smoke/standing-renewal-auth.smoke.ts` (guard present, product source untouched) produced a genuine broker denial of the shape `User JWT no longer valid ... claim is expired` on the expiring endpoint's connection. The suite passed, because its assertion window opens later and does not cover that connection.

[...]

Issue #1408, opened 2026-09-09
standing-renewal expired-cred cell discriminates only intermittently, so its reproof is a coin flip


davidfarah2003:
`packages/core/smoke/standing-renewal-auth.smoke.ts` asserts:

```
a rebuild presents cached expired creds to the broker ZERO times after renewal fails
```

Under the mutation in `packages/core/smoke/mutations/standing-renewal-expiry.json`, which removes the pre-dial expiry guard, that cell reds **sometimes**. The fixture's timing is non-deterministic, so the mutation reproof for it is a coin flip, and every PR touching `packages/core/src/endpoint.ts` inherits that coin flip.

> **This issue originally claimed the cell "does not discriminate". That was wrong** and is corrected below. It discriminates intermittently, which is a different defect with a different fix. The false version was disproved by `rev-1407-grok` on PR #1407 and then reproduced by me. See the comments.

### Measured

Mutation applied by the fixture's own `find`/`replace`, suite run directly, at `e8cd02a6`:

| run | broker-denial cell | pre-dial cell | result |
| --- | --- | --- | --- |
| 1 | green | RED | 10 passed, 1 failed |
| 2 | **RED** | RED | 9 passed, 2 failed |
| 3 | green | RED | 10 passed, 1 failed |
| 4 | **RED** | RED | 9 passed, 2 failed |
| 5 | green | RED | 10 passed, 1 failed |

2 of 5 here; `rev-1407-grok` measured 6 of 7 independently, and captured the discriminating line inside the detection window (`failedRebuildLogStart=11324`, match at `jwtIdx=12537`). The signature is the genuine fresh-dial rejection:

```
User JWT no longer valid: &{Issues:[claim is expired]}
```

It also observed the inverse — two such dials in one run with only one inside the window — so the window position matters as well as whether the dial happens at all.

### The two moving parts

- **Whether the endpoint makes a fresh dial with an expired cached cred at all.** It depends on when the fixture's creds source comes back relative to the 3-second cred's expiry, which is a race between the renewal backoff and the recovery loop.
- **Whether that dial lands inside the detection window.** `failedRebuildLogStart` is set at `expiringReads === 3`, the second failed renewal read. Some dials happen before it.

### What a fix has to do

Make the mutated path deterministic so the fresh dial with an expired cred reliably occurs, and reliably inside the window. Then the fixture's original `expectRed` is correct as written and nothing needs renaming.

Two failure modes to avoid, both established by measurement rather than argument:

[...]

PR #1414, opened 2026-09-09
fix(smoke): grade the creds expiry mutation on the cell that detects it

### LINKED ISSUE #1403: standing-renewal-expiry.json names the wrong assertion, so any PR touching endpoint.ts fails reproof

davidfarah2003:
Closes #1403.

Addresses #1408 (the same fixture's intermittent discrimination); leaving that one's disposition to its owner rather than auto-closing it, since its lane carried measurement scope beyond this fixture.

The product defect in #1411 remains unresolved and is out of scope here: the pre-dial guard is absent from the getter nats.js uses for its own reconnects (`endpoint.ts:1017`), so an expired credential can still reach the broker. This PR changes fixture metadata only and fixes none of that. The zero-expired-presentation guarantee is explicitly **unproved** after this change: the old cell claimed it and never reliably delivered it, and the new cell asserts the local refusal is raised, which is a diagnostic, not the wire-level property.

`packages/core/smoke/mutations/standing-renewal-expiry.json` named a cell that cannot detect its own mutation, so mutation reproof returned `WRONG-RED` and reddened every change touching `packages/core/src/endpoint.ts`. This points the fixture at the adjacent cell that does detect it. No production code changes.

## The defect

The fixture graded `a rebuild presents cached expired creds to the broker ZERO times after renewal fails`, whose predicate is `failedRebuildLogStart >= 0 && expiredCredDenials.length === 0`. It asserts the **absence** of a broker denial.

`credsRenewalDelayMs` returns `iatMs + 0.75 * (expMs - iatMs) - Date.now()`, so the endpoint treats creds as stale at 75% of lifetime while the pre-dial guard refuses only at `exp`. Between those points the cached credential is still valid: the broker accepts it and logs no denial. Deleting the guard is unobservable to that cell in that window, which at the suite's `expiresInSeconds: 3` is 750ms wide.

The first conjunct compounds it. `failedRebuildLogStart >= 0` is satisfied by the marker having fired, not by the sliced window containing anything, so an empty capture satisfies the predicate vacuously.

The adjacent cell `the pre-dial refusal is loud and names the failing renewal path` asserts the **presence** of the thrown refusal. Removing the guard removes the refusal, so it goes red on the mutation rather than on a timing window.

## Evidence

Baseline, unmutated, at `origin/main`: 11 passed, 0 failed. The new named cell is green without the mutation.

`node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/standing-renewal-expiry.json`, five consecutive runs on this head: **KILLED 5/5**, each `red, and named: the pre-dial refusal is loud and names the failing renewal path`, tree clean after every restore.

Direct runs with the guard removed by hand, three consecutive: both cells red locally. Contrast that with CI run [34355677896](https://github.com/Cotal-AI/Cotal/actions/runs/34355677896/job/102479616037) `changed shard 8/12`, where under the same mutation the old cell printed its GREEN text while the new one failed. Same mutation, opposite verdicts from the old cell on two machines, which is the intermittency stated as a measurement rather than a hypothesis. Mark counts across the five proof runs vary 9/10 against a baseline of 11 for the same reason: the old cell contributes a red only when the timing lands past `exp`.

## Scope

Test fixture only, so no changeset. The mutation, the suite and `endpoint.ts` are untouched; only `cell`, `expectRed` and the `why` rationale change. This does not alter what the guard does, only whether the suite can prove it.

### DIFF
--- a/packages/core/smoke/mutations/standing-renewal-expiry.json
+++ b/packages/core/smoke/mutations/standing-renewal-expiry.json
@@ -10,8 +10,10 @@
     "The endpoint's pre-connect refresh is deliberately best-effort after the first connection. If",
     "the source is down when the cached cred expires, removing this guard restores the reported",
     "behavior exactly: connectAndBind presents credential material it has already decoded as expired.",
-    "The live smoke captures the disposable broker's authentication-expired denial, so the named",
-    "cell proves the real dial path reached the broker rather than only exercising a local parser."
+    "The named cell asserts the refusal is raised. The sibling cell counting broker denials cannot",
+    "grade this mutation: credsRenewalDelayMs reports creds stale at 75% of lifetime while the guard",
+    "refuses only at exp, so between those points the cached cred is still valid, the broker accepts",
+    "it, and no denial is logged whether or not the guard is present."
   ],
   "mutations": [
     {
@@ -19,8 +21,8 @@
       "file": "packages/core/src/endpoint.ts",
       "find": "    const credsExp = this.currentCreds && credsClaims(this.currentCreds).exp;\n    if (typeof credsExp === \"number\" && credsExp * 1000 <= Date.now())\n      throw new Error(\n        this.credsSource\n          ? \"this endpoint's creds have expired and renewal is failing - not presenting the expired credential to the broker; retrying with backoff\"\n          : \"this endpoint's creds have expired and it holds no creds source to renew them - replace the credential and rebuild the endpoint (pass a creds FUNCTION for standing renewal)\",\n      );\n",
       "replace": "",
-      "expectRed": "a rebuild presents cached expired creds to the broker ZERO times after renewal fails",
-      "cell": "a rebuild presents cached expired creds to the broker ZERO times after renewal fails",
+      "expectRed": "the pre-dial refusal is loud and names the failing renewal path",
+      "cell": "the pre-dial refusal is loud and names the failing renewal path",
       "note": "Deleting the block is the issue's pre-fix behavior and leaves the tree compiling. The source fails twice, so the first reconnect dials the cached expired cred and the broker logs the denial before the source recovers."
     }
   ]
--- a/packages/core/smoke/mutations/standing-renewal-expiry.json
+++ b/packages/core/smoke/mutations/standing-renewal-expiry.json
@@ -10,10 +10,12 @@
     "The endpoint's pre-connect refresh is deliberately best-effort after the first connection. If",
     "the source is down when the cached cred expires, removing this guard restores the reported",
     "behavior exactly: connectAndBind presents credential material it has already decoded as expired.",
-    "The named cell asserts the refusal is raised. The sibling cell counting broker denials cannot",
-    "grade this mutation: credsRenewalDelayMs reports creds stale at 75% of lifetime while the guard",
-    "refuses only at exp, so between those points the cached cred is still valid, the broker accepts",
-    "it, and no denial is logged whether or not the guard is present."
+    "The named cell asserts the refusal is raised, so it goes red on the deleted block itself.",
+    "The sibling cell counting broker denials cannot grade this mutation. It can read zero denials",
+    "while the guard is absent in two ways: a rebuild that dials before exp presents a cred the",
+    "broker still accepts, and a denial that does land can fall outside the slice the cell inspects,",
+    "because failedRebuildLogStart only records that the marker fired. Under the mutation that cell",
+    "was observed both red and green on the same machine, and green in CI."
   ],
   "mutations": [
     {
@@ -23,7 +25,7 @@
       "replace": "",
       "expectRed": "the pre-dial refusal is loud and names the failing renewal path",
       "cell": "the pre-dial refusal is loud and names the failing renewal path",
-      "note": "Deleting the block is the issue's pre-fix behavior and leaves the tree compiling. The source fails twice, so the first reconnect dials the cached expired cred and the broker logs the denial before the source recovers."
+      "note": "Deleting the block is the issue's pre-fix behavior and leaves the tree compiling. The refusal text the named cell matches exists only inside the deleted block, so removing it is what makes the cell red. The suite's own recovery path is unaffected: the source fails twice and then returns."
     }
   ]
 }


davidfarah2003:
CI confirms the repair on the same job that reported the defect.

`changed shard 8/12` on this head ([run 34392331836](https://github.com/Cotal-AI/Cotal/actions/runs/34392331836/job/102603468587)):

```
mutation reproof: 1 fixture(s) selected from 368 (diff 1 record(s), 1 changed path(s), corpus 368)
KILLED   the creds pre-dial expiry guard is removed
  red, and named: the pre-dial refusal is loud and names the failing renewal path
```

Three things that matter in that output.

The fixture was **selected by its own change**. `mutation-reproof.mjs:219-228` keeps a fixture if any of four routes fires — `all`, `config` (the fixture path changed), `suite`, `mutation` (a mutated file changed). Here the diff is one record and one changed path, so `config` selected it. A fixture repair is therefore graded on its own PR rather than shipping dark.

It **KILLED on the new cell by name**, on the CI runner, under the same mutation that returned `WRONG-RED` on #1397's [shard 8/12](https://github.com/Cotal-AI/Cotal/actions/runs/34355677896/job/102479616037). Same job, same fixture, same mutation, opposite verdict.

All twelve reproof shards and the `changed` aggregate are green; 25 of the other checks pass with 1 skipped and the rest still running, no failures.

Local evidence in the PR body stands alongside this: 5/5 KILLED across consecutive proof runs, and three direct mutated runs showing the old cell red locally while it was green in CI under the same mutation. That divergence is the intermittency, measured rather than argued.

I authored this change, so it needs an independent exact-head review before merge. Head is `cf4fab276373b279093e8e2e8e8c8e5d560e221c`.

davidfarah2003:
Independent review at `cf4fab276373b279093e8e2e8e8c8e5d560e221c` (worktree `/home/cotal/fm-lanes/rev-1414-grok`, dirty=0). One-file change: `packages/core/smoke/mutations/standing-renewal-expiry.json`. TMPDIR pointed at a disk path outside the worktree because `/tmp` is a 16G tmpfs at ~68% full.

**Verdict: approve.** The new named cell is a sound grading cell. I could not make it green under the mutation.

### 1. Is the new cell sound?

Unmutated: `pnpm smoke:standing-renewal` was 11/11, including `the pre-dial refusal is loud and names the failing renewal path`.

`node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/standing-renewal-expiry.json`: **KILLED 4/4**, named that cell each time (progress marks 9, 9, 10, 9 against baseline 11). Tree clean after every restore.

Hand-deleted the guard in `packages/core/src/endpoint.ts` and ran the suite four times: the new cell was **red 4/4**. Under the mutation the warnings were only `creds refresh failed (fixture renewal source offline)...` plus broker-close noise. Never the `/creds have expired.*renewal.*failing/` text. That string exists only in the deleted block. I did not observe a green new cell under the mutation even once.

### 2. Is the old cell unsound for the stated reason?

Reproduced, not taken on trust. Same four hand-mutated runs: old cell **GREEN 2/4, RED 2/4**. That is the CI-green / local-red split on one box.

The mechanism is real but incomplete as written. `credsRenewalDelayMs` is 75% of iat→exp (`identity.ts:164`). The pre-dial guard refuses only at `exp` (`endpoint.ts:993-999`). Between those points the cached JWT is still broker-valid, so a rebuild can present it and log no denial. The first conjunct `failedRebuildLogStart >= 0` is also vacuous: it only needs the marker, not a denial-free window.

What the rewritten `why` understates: `armCredsRefresh` floors delay at 1s (`endpoint.ts:946`). On `expiresInSeconds: 3`, 75% of 3s is 2.25s, so the timer fires ~1s before exp, not at 75%. The first failed refresh therefore happens while the JWT is still valid. After that, `CREDS_RETRY_MS` is 60s, so further source fetches wait on the broker expiry-close / rebuild path, not on another 75% tick. The 750ms figure is the remaining JWT lifetime after that first failed refresh, not a 750ms window in which `bindConnection` is racing the guard. I still observed the old cell green with zero denials, so the grading defect stands. I also observed it red when a post-exp dial did land (`User JWT no longer valid: claim is expired` in the broker log). Intermittency is scheduling of the first post-exp rebuild versus the recovering source, not a different defect.

### 3. Does the rewritten `why` state something false?

The 75%-vs-exp claim is true. "no denial is logged whether or not the guard is present" is true of the still-valid window, not of every mutated run. Locally the old cell did go red when a post-exp dial happened. That is a precision issue, not a false claim about the cell's unsoundness.

### 4. Did re-anchoring weaken the proof?

Yes, in a named way, and it is still the right trade. The old cell was the only one that asserted the broker never saw expired material. The new cell asserts the local throw exists. Under the mutation I saw `User Authentication Expired` / JWT `claim is expired` on the wire even while the old cell was sometimes still green, because denials can land outside the `failedRebuildLogStart` slice. So the fixture now proves "deleting the guard removes the loud local refusal", which is exactly the mutation, and no longer claims to prove "zero expired presentations". The leftover `note` on the mutation still talks about the broker denial path; it is stale relative to `cell`/`expectRed`/`why` but is not graded.

Blocking bar was: new cell green under the mutation even once. Did not happen.

davidfarah2003:
Head moved to `e44eb7af855375de104273410ed57aa33c02b3f6`. The APPROVE at `cf4fab276373b279093e8e2e8e8c8e5d560e221c` no longer binds, and the change that moved it came out of that review.

`rev-1414-grok` approved on the blocking bar it was given (it could not make the new cell green under the mutation: KILLED 4/4 by proof, red 4/4 by hand-deleting the guard, and the refusal string it matches exists only inside the deleted block). It also found two things wrong with my rationale, and it was right about both.

**My mechanism was incomplete, and the reviewer's version is a stronger indictment.** I wrote that the old cell passes because a rebuild inside the still-valid 75%-to-`exp` window presents a credential the broker accepts. That happens, but it is one of two routes. The reviewer observed `User Authentication Expired` / `JWT claim is expired` **on the wire while the old cell was still green** — a denial that landed outside the slice the cell inspects, because `failedRebuildLogStart` records only that the marker fired, not that the captured window contains anything. So the cell reads zero denials with the guard deleted in two independent ways, not one.

It also corrected the dynamics I attached to the 750ms figure. `armCredsRefresh` floors the timer at 1s (`endpoint.ts:946`) and `CREDS_RETRY_MS` is 60s, so after the first failed refresh nothing retries inside that remaining lifetime; the rebuild that dials is driven by the broker's expiry-close. 750ms is the JWT lifetime remaining after the first failed refresh, not a window in which the dial races the guard. The arithmetic was right and the story I told about it was not.

**The mutation `note` was stale.** It still described the broker-denial path the fixture no longer grades. The reviewer flagged it as ungraded but drifting, which is the same defect class this PR exists to fix.

`e44eb7af` is text-only against the reviewed commit: `why` and `note`, no change to `cell`, `expectRed`, `find`, `replace`, `suite` or `command`. Kept as a separate commit so the reviewed artifact stays intact in history and the delta is reviewable on its own. Re-proved on the new head: KILLED, tree clean.

Re-review scope is that one commit. The behavioural claim is unchanged and already graded; what needs grading is whether the rewritten rationale is now accurate.

davidfarah2003:
Re-review at `e44eb7af855375de104273410ed57aa33c02b3f6` (parent `cf4fab276`, which I already approved). Worktree dirty=0.

**Verdict: approve the follow-up.** The rewritten `why` now states the two-route unsoundness correctly. It does not flatten back into the 750ms race story.

Scope checked: `git diff cf4fab276 e44eb7af8` is one file, `why` + `note` only. `cell`, `expectRed`, `find`, `replace`, `suite`, `command`, `guard`, `completionMarker` are byte-identical. `node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/standing-renewal-expiry.json` on this head: **KILLED**, named the same cell, 9 marks vs baseline 11, tree clean after restore.

Route 1 is now "a rebuild that dials before exp presents a cred the broker still accepts." That is the dynamics I measured, not "credsRenewalDelayMs reports stale at 75% so there is a 750ms window." The 1s floor and the 60s `CREDS_RETRY_MS` are omitted, correctly: they explain *why* a pre-exp rebuild happens, they are not a second grading defect.

Route 2 is now "a denial that does land can fall outside the slice the cell inspects, because `failedRebuildLogStart` only records that the marker fired." That is the stronger finding and it is not collapsed into route 1.

The observation sentence ("both red and green on the same machine, and green in CI") matches what I measured (old cell 2/4 green under the mutation here) and what was reported from CI. I did not re-run the CI shard.

Two leftover precision nits, neither blocking:

- The opening still says removing the guard "restores the reported behavior exactly: connectAndBind presents credential material it has already decoded as expired." That is the intended defect, not a guaranteed observation on every mutated run. The next sentences already say the sibling cell cannot see that. Fine as the guard's purpose, slightly strong as a run-by-run claim.
- `note` is no longer stale. "The refusal text the named cell matches exists only inside the deleted block" is true (single occurrence in `endpoint.ts`). "The source fails twice and then returns" is still the fixture shape.

Point 4 from the first review is the bound, not a remaining defect in this commit: the fixture now proves the mutation and no longer claims to prove zero expired presentations.

Issue #695, opened 2026-08-19
security: a tool result carries text across the read ACL it was read under, onto the events channel


davidfarah2003:
The AG-UI event plane publishes a seat's tool activity verbatim. A tool that reads on the seat's
behalf therefore moves text from wherever it read onto `events.<owner>.<actor>`, whose read ACL is
not the read ACL of the place the text came from. The clearest instance: a seat calls the mesh
`cotal_inbox` tool, the tool returns a message someone posted on a channel with a narrower reader
set, and that message is published to the events channel unredacted and with no marking of where it
came from.

This is reachable today on the default branch and it is not specific to a recovery or an outage.
The mapping is unconditional, so it happens on an ordinary live turn; a recovery from a broker
outage only replays the same class from the write-ahead log.

## Reproduced

An external reviewer reproduced it end to end against a real broker and a real host: a marker queued
on a quiet channel, returned by the host's own loopback MCP `cotal_inbox` during a turn, written to
the rollout as an ordinary `function_call_output`, and received verbatim by the events subscriber as
`TOOL_CALL_RESULT`. A direct user-authored record marker stayed absent in the same run, so this is a
missing route rather than a hole in the input-role filter.

Independently, feeding the mapper three records by hand reproduces the carrier with a control:

```
TOOL_CALL_RESULT     carriesMarker=true
CONTROL, the same string as a response_item/message role=user record: DROPPED (withheld)
```

Same string, same mapper, one shape crosses and the other is refused. The default-deny this mapper
implements is over record SHAPES, not over content provenance.

## The full carrier set, from the code rather than from the incident

Two of these are mechanical and can be addressed at the mapper. Two cannot, and saying so is part of
scoping the work.

Mechanical, the text is taken from the record and republished byte for byte:

- `response_item/function_call` and `custom_tool_call` publish `arguments` (or `input`) as
  `TOOL_CALL_ARGS`. Anything the model put into a call crosses, including text it is forwarding.
- `response_item/function_call_output` and `custom_tool_call_output` publish the output as
  `TOOL_CALL_RESULT`. Anything a tool returned crosses, including what it read for the seat.

Not mechanical, and a redaction seam cannot close them:

- an assistant message is published as `TEXT_MESSAGE_CONTENT`, so anything the model chooses to
  quote crosses in its own words;
- a reasoning summary is published the same way.

Withheld today and staying withheld: the session's own record of the user's words, in both shapes
[...]

PR #1415, opened 2026-09-09
fix(connector-claude-code)!: withhold tool-result bodies on the events plane

davidfarah2003:
## Summary

The Claude Code mapper used to copy every `tool_result` body onto `events.<owner>.<actor>` as `TOOL_CALL_RESULT.content`. That subject has a different read ACL from wherever the tool read, so a seat calling a mesh read tool moved text from a narrower reader set onto a broader one, unredacted.

The defect in #695 is repaired by failing closed in `createClaudeMapper`: without trusted provenance and evidence the destination audience may read the body, the mapper does not emit `TOOL_CALL_RESULT` at all. `content` is mandatory on that event, so there is no empty-string or placeholder form. Call lifecycle (`TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END`) is unchanged.

Observers lose tool output they see today. That is the intended boundary, not a regression.

## What this does not do

- No tool-name allowlist.
- No redaction marker or invented schema field.
- No new provenance protocol.
- Tool-supplied "this output is safe" is not treated as evidence.

## Follow-up (out of this PR)

The same unconditional republish exists in:

- `extensions/connector-codex/src/agui-map.ts` (`function_call_output` / `custom_tool_call_output` → `toolCallResult` with `outputText(payload)`).
- `extensions/connector-opencode/src/agui-map.ts` (`part.type === "tool"` → `toolCallResult` with `toolResultContent(part.state)`).

`@cotal-ai/connector-core` only constructs the event; it does not map harness records. Those adapters are separately scoped.

## Testing

Synthetic placeholders only. No live ACL crossing, no real tool output, no secrets.

- Reproduced from source before the change: a synthetic `tool_result` mapped to `TOOL_CALL_RESULT` with the placeholder on the serialized events.
- `pnpm smoke:agui-tool-result` (safety: placeholder absent, no `TOOL_CALL_RESULT`; liveness: START/ARGS/END and prompt body retained).
- `pnpm smoke:agui-authorship` (former known-limit arm is now fail-closed).
- `pnpm smoke:agui-map` against the committed session fixture.
- `pnpm --filter @cotal-ai/connector-claude-code typecheck`
- `pnpm check:docs-voice`
- `pnpm smoke:gate-inventory`
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/agui-tool-result.json`
  - T1 restore verbatim content → KILLED `a tool_result placeholder does NOT appear anywhere on the wire`
  - T2 drop tool_use lifecycle → KILLED `DECOY/LIVENESS — tool_use still emits START/ARGS/END`

### DIFF
--- a/.changeset/claude-tool-result-acl.md
+++ b/.changeset/claude-tool-result-acl.md
@@ -1,0 +1,10 @@
+---
+"@cotal-ai/connector-claude-code": minor
+---
+
+The Claude Code event plane no longer republishes tool-result bodies. A `tool_result` block has no
+trusted provenance at the mapper, and `events.<owner>.<actor>` carries a different read ACL from
+wherever the tool read, so `TOOL_CALL_RESULT` is suppressed rather than emitted with the body, an
+empty string, or a placeholder. Call lifecycle (`TOOL_CALL_START` / `TOOL_CALL_ARGS` /
+`TOOL_CALL_END`) is unchanged. Observers lose tool output they saw before; that is the intended
+boundary, not a regression.
--- a/bin/smoke/ci-suites.d/3e72e02679e616db96d35f56218c0ae8277472183c0f82cc940579fab4ba88cb.txt
+++ b/bin/smoke/ci-suites.d/3e72e02679e616db96d35f56218c0ae8277472183c0f82cc940579fab4ba88cb.txt
@@ -1,0 +1,2 @@
+# Claude Code mapper: withhold tool-result bodies on the events plane.
+smoke:agui-tool-result
--- a/docs/connect-claude.md
+++ b/docs/connect-claude.md
@@ -243,10 +243,10 @@ with the adapter:
 ## Event plane
 
 A session launched with `cotal spawn --events` publishes a **structured** account of what it
-did: run boundaries per turn, assistant text, reasoning, and each tool call with its arguments,
-its end, and its result. Not prose about the work, the work itself, in a vocabulary a program can
-read. Arming is `COTAL_EVENTS`, which the launcher sets for `--events` spawns; a personal session
-with the plugin installed publishes nothing.
+did: run boundaries per turn, assistant text, reasoning, and each tool call with its arguments
+and its end. Tool **results are not published**. Not prose about the work, the work itself, in a
+vocabulary a program can read, minus the result body. Arming is `COTAL_EVENTS`, which the launcher
+sets for `--events` spawns; a personal session with the plugin installed publishes nothing.
 
 A new session includes its first run even when Claude writes a positional startup prompt before the
 connector receives `SessionStart`. That from-zero read is keyed only to Claude's explicit
@@ -268,9 +268,13 @@ new startup waits up to five seconds for that file with capped backoff, and the 
 one stalled file read; expiry fails loud instead of silently losing the first run. Retained-history
 starts and recovered cursors still require their existing source at once.
 
-Tool arguments and results go on this channel verbatim, so withholding user-authored text does not
-make the stream safe to widen: anything a tool reads or prints, including a secret in a command line
-or in the contents of a file, reaches every reader of the channel.
+Tool arguments still go on this channel, so withholding user-authored text and tool results does
+not make the stream safe to widen: anything a tool is *asked* to do, including a secret in a
+command line, reaches every reader of the channel. Tool **results** do not. A `tool_result` block
+has no trusted provenance at the mapper, and `TOOL_CALL_RESULT.content` is mandatory, so the
+result event is suppressed rather than emitted empty or with a placeholder. Observers lose that
+output on purpose: republishing it onto `events.<owner>.<actor>` would move text from a narrower
+reader set onto a broader one.
 
 The channel is **`events.<owner>.<actor>`**, named after the session's principal. What the actor
 half is depends on the mesh, and the difference matters when you go looking for it: on a static mesh
@@ -279,7 +283,7 @@ do not share a stream; on a user-auth mesh it is the agent's own name, because t
 ledger row is keyed on. Spelled out again with both halves below. The launch grants publish rights
 on that channel alone. A spawn
 that asks for a *different* agent's event channel is refused at the door rather than granted, since
-that channel carries the session's tool inputs and outputs. The same rule runs on restart: a manager
+that channel carries the session's tool arguments and assistant text. The same rule runs on restart: a manager
 resume document that names another agent's event channel is refused rather than adopted, because the
 managed row is re-armed from that document and the credential is re-minted from the row.
 
--- a/extensions/connector-claude-code/smoke/agui-authorship.smoke.ts
+++ b/extensions/connector-claude-code/smoke/agui-authorship.smoke.ts
@@ -131,10 +131,9 @@ const mapAll = (entries: ClaudeEntry[]) => {
     events.filter((e) => e.type === "RUN_STARTED").length);
 }
 
-// ── ARM D: A TOOL RESULT CARRYING ANOTHER PRINCIPAL'S TEXT ────────────────────────────────────
-// Named separately because it arrives by a different route: not an `origin.kind`, but a peer's
-// words quoted inside this session's own tool output. Recorded as a KNOWN LIMIT rather than
-// asserted as safe — see the cell name.
+// ── ARM D: A TOOL RESULT, WHICH HAS NO PROVENANCE AT THIS LAYER ──────────────
+// Named separately because it arrives by a different route: not an `origin.kind`.
+// Fail-closed: TOOL_CALL_RESULT.content is mandatory, so the event is suppressed.
 {
   const toolResult = mk({
     uuid: "u8",
@@ -144,11 +143,8 @@ const mapAll = (entries: ClaudeEntry[]) => {
   } as never);
   const { events } = mapAll([human("u9", MINE), toolResult]);
   const wire = JSON.stringify(events);
-  // This is NOT a safety claim. A tool result is this session's own output and IS republished by
-  // design; if it quotes a peer, that text rides along. Stating it as a measured limit so nobody
-  // reads arm A as "no peer text can ever reach the wire".
-  c("KNOWN LIMIT: a tool result IS emitted, so peer text quoted inside one is not covered by this filter",
-    wire.includes(TOOL_SECRET));
+  c("FAIL-CLOSED: a tool_result body is withheld (no TOOL_CALL_RESULT, no placeholder bytes)",
+    !wire.includes(TOOL_SECRET) && !events.some((e) => e.type === "TOOL_CALL_RESULT"));
 }
 
 // A COUNT, because several cells above build their own inputs and a regression that DELETES one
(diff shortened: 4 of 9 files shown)


davidfarah2003:
## Review (head `5c9d2b7611fd58dc77d3d9abc1bf23ec58530a22`)

**Block.** Re-read at comment time: GitHub `headRefOid` equals that sha.

### Claude mapper (this PR)

The named route is closed. In `extensions/connector-claude-code/src/agui-map.ts`, a `user` record whose content array contains any `tool_result` with `tool_use_id` maps to `null`. `TOOL_CALL_RESULT` is not emitted. There is no empty `content` and no placeholder string, which matches the schema (`content` is mandatory) and the existing authorship docblock. Call identity and ordering remain on `TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END` from the matching `tool_use` block. The only production entry is `extensions/connector-claude-code/src/mcp.ts` (`createClaudeMapper` into `AguiEmitter`).

Docs in `docs/connect-claude.md` now say tool results are not published. `closingIssuesReferences` is empty. The body references #695 without a closing keyword.

Synthetic checks (placeholders only): `pnpm smoke:agui-tool-result` 10/0, `pnpm smoke:agui-authorship` 13/0, `pnpm smoke:agui-map` 51/0 on the committed derived session. Mutation config `extensions/connector-claude-code/smoke/mutations/agui-tool-result.json`: T1 (restore verbatim result) and T2 (drop `tool_use` lifecycle) both KILLED on the named cells.

### Why this does not close the class

Issue #695's reproduced carrier is Codex `function_call_output` written to the rollout and published as `TOOL_CALL_RESULT`. That path is still live:

- `extensions/connector-codex/src/agui-map.ts` (`function_call_output` / `custom_tool_call_output` → `toolCallResult` with `outputText(payload)`)
- `extensions/connector-opencode/src/agui-map.ts` (`part.type === "tool"` → `toolCallResult` with `toolResultContent(part.state)`)

The PR lists those as follow-up, then says the defect in #695 is repaired in `createClaudeMapper`. That is an overclaim relative to the issue's own repro.

### Mutation proof bound

The tool-result suite imports the mapper and builds records by hand. A killed mutation shows that suite depends on the mutated lines. It does not show a live transcript or `mcp.ts` reaching them. The fixture suite is the separate, still-synthetic, shape check.

`TOOL_CALL_ARGS` still republishes tool inputs. Hermes `t:"tool_result"` is the sidecar protocol, not `events.<owner>.<actor>`.

Do not merge as a repair of #695 while the Codex (and OpenCode) result publishers remain.

davidfarah2003:
**BLOCK on `5c9d2b7611fd58dc77d3d9abc1bf23ec58530a22`**

The changed mapper fails closed for newly read `tool_result` records, but the shipped recovery entry point bypasses it for a pre-fix pending WAL frame.

`AguiEmitter.start()` calls `recover()` before any source read. For `sent_unacked`, `recover()` passes the frozen `pending.body` directly to `attempt()`. It never invokes `createClaudeMapper`, and the WAL has no policy/version check that can reject a frame containing the old `TOOL_CALL_RESULT` carrier.

I reproduced this synthetically through the actual implementations:

1. Used the parent commit's real Claude mapper to generate a valid run containing a synthetic `TOOL_CALL_RESULT` placeholder.
2. Froze that frame with `EventWal.beginSend`, representing a crash or network uncertainty after the durable write and before settlement.
3. Reopened the WAL at this head and called the current `AguiEmitter.start()` with the current mapper.
4. The recorded publish showed `currentMapperCalls=0`, `publishedCalls=1`, and contained both the placeholder and `TOOL_CALL_RESULT`.

This is the outage/retry route named in the issue. Upgrading while such a frame is pending can still republish tool-result-derived text onto the events subject. The new mutation suite only calls the mapper with hand-built records, so both mutations can be killed while this bypass remains reachable.

The repair needs to fail closed for incompatible pending bodies before recovery publishes them, or otherwise migrate/version this policy without treating uncertainty as permission.

Other checks passed: `smoke:agui-tool-result`, `smoke:agui-authorship`, `smoke:agui-map`, connector typecheck, docs voice, and both submitted mutations. I did not exercise a live ACL crossing.

davidfarah2003:
## Review follow-up (head `5c9d2b7611fd58dc77d3d9abc1bf23ec58530a22`)

**Block** still. GitHub `headRefOid` equals that sha.

Independently reproduced a restart bypass on this head (synthetic FakeEndpoint, no live ACL):

1. Built a valid AG-UI frame whose events include `TOOL_CALL_RESULT` with placeholder bytes.
2. Froze it with `EventWal.beginSend` (`pending.state=sent_unacked`).
3. Reopened the WAL and called `AguiEmitter.start()` with this head's `createClaudeMapper`.

Observed: `mapper_calls=0`, `publishes=1`, channel `events.local.aaa`, `TOOL_CALL_RESULT` and the placeholder bytes both present on the recorded publish.

`AguiEmitter.start()` still calls `recover()` before any `map()`, and `recover()` sends `pending.body` verbatim. The Claude mapper therefore cannot stop a pre-fix pending frame on the named outage/retry path.

The live Claude mapper path remains fail-closed for new `tool_result` records. Codex `function_call_output` / `custom_tool_call_output` and OpenCode `part.type === "tool"` still emit `toolCallResult` on the live path.

No live ACL crossing.

davidfarah2003:
## Disposition: both blocks accepted, and the scoping one is on me

Two independent blocks at `5c9d2b7611fd58dc77d3d9abc1bf23ec58530a22`, on different defects. I verified both first-hand rather than relaying them.

**The scope block is a briefing error, not an implementation error.** The brief this change was written against named `extensions/connector-claude-code/src/agui-map.ts:411-424` as "the defect" and treated codex and opencode as separately scoped follow-up. That is wrong about #695. The issue's repro is on Codex — the output is written to the rollout as a `function_call_output` and received verbatim as `TOOL_CALL_RESULT` — and its carrier-set section names *two* mechanical carriers, not one:

- `response_item/function_call` and `custom_tool_call` publish `arguments`/`input` as `TOOL_CALL_ARGS`
- `response_item/function_call_output` and `custom_tool_call_output` publish the output as `TOOL_CALL_RESULT`

At this head `extensions/connector-codex/src/agui-map.ts` still publishes `toolCallResult(... outputText(payload) ...)` for that record pair, `connector-opencode` still publishes, and `TOOL_CALL_ARGS` is untouched (`agui-map.ts:568`). So the change closes one carrier on one connector. The work is correct for what it was asked to do; it was asked for the wrong thing.

**The WAL block is a design finding and it outlives this PR.** `AguiEmitter.start()` runs `recover()` first (`extensions/connector-core/src/agui.ts:1435`), and on a `sent_unacked` pending frame `recover()` does `attempt({ id: p.id, E: p.E, body: p.body, retry: true })` at `:1463` — republishing a body frozen by a previous process without calling `this.map()`. Only `pump()` calls the mapper, and the docblock states the intent plainly: "The retried frame was accepted by a PREVIOUS process and is not fed through this instance's machine again."

The consequence is general: a redaction placed at the mapper is bypassed by WAL recovery whenever an upgrade lands across an uncertain publish. A mutation suite that drives the mapper directly cannot catch that route, so this is a coverage gap as well as a behaviour gap.

**This PR stays open.** The claude-code `TOOL_CALL_RESULT` carrier really is closed here, fail-closed with no placeholder and no allowlist, `resultContent()` and the `toolCallResult` import are both removed so the route cannot be re-reached, and the proof stands on re-run: both mutations killed, `agui-tool-result` 10/10, `agui-authorship` 13/13, `agui-map` 51/51 over a 2200-record fixture. It is a sound partial and should be re-scoped rather than discarded.

What the corrected scope needs:

1. Codex as the primary route, both carriers.
2. OpenCode and claude-code as the same class; this PR already does claude-code's result half.
3. An explicit statement that a mapper-level fix is insufficient on its own while `recover()` republishes frozen bodies — either recovery applies the same policy, or the WAL records a policy version and refuses to republish a body frozen under an older one.
4. At least one mutation cell routed through `AguiEmitter.start()`, not only through `createClaudeMapper`.

davidfarah2003:
## This branch fails `check:docsbundle` at its own head

Recording this so it does not have to be rediscovered, independent of what happens to this PR.

The branch edits `docs/connect-claude.md` but does not carry `extensions/connector-core/src/docs-bundle.generated.ts`. That page feeds the generated bundle, and `check:docsbundle` (inside `unit`) regenerates and ends in `git diff --exit-code` against the committed copy, so a docs edit without its regenerated sibling fails the gate.

The full changed set here is `.changeset/`, a `bin/smoke/ci-suites.d/` fragment, `docs/connect-claude.md`, four `extensions/connector-claude-code/smoke/` files, `extensions/connector-claude-code/src/agui-map.ts`, and `package.json` — no generated bundle.

Worth noting why it was not caught in review: both reviewers ran `check:docs-voice`, which does not cover the bundle. The two checks sound interchangeable and are not. A docs change in this repo carries a generated sibling, and `check:docs-voice` passing says nothing about it.

## Scope note, for whoever picks this up

A full sweep of `toolCallArgs(` / `toolCallResult(` call sites across shipped source at `438d03a1e` finds **six** carriers in three connectors, plus the two constructors in core:

```
extensions/connector-claude-code/src/agui-map.ts:416   (result)
extensions/connector-claude-code/src/agui-map.ts:573   (args)
extensions/connector-codex/src/agui-map.ts:261         (args)
extensions/connector-codex/src/agui-map.ts:275         (result)
extensions/connector-opencode/src/agui-map.ts:210      (args)
extensions/connector-opencode/src/agui-map.ts:212      (result)
extensions/connector-core/src/agui.ts:723, :762        (definitions)
```

This PR closes one of the six. Separately, `AguiEmitter.recover()` republishes a frozen WAL body without passing it through the mapper — `this.map(` appears at exactly one line in `agui.ts`, `:1485` in `pump()` — so a redaction implemented in a mapper is bypassed on the recovery path regardless of how many carriers it covers.

Both points are consistent with the two review blocks already on this PR rather than new objections to it.

davidfarah2003:
Superseded by #1425, merged as `ee73e33c15f1a147b435d44a6f4feb8fd38512bc`. Closing this one.

This was sound work and it was blocked for scope, not for correctness. Both panel reviewers
were right on both counts, and the record is worth keeping straight:

1. **The class was not closed.** The defect reproduces on Codex as well, and this branch was
   scoped to a single file. There are six carriers across three connectors, two per connector
   (args and result), not one file's worth.
2. **`AguiEmitter.recover()` bypasses the mapper.** It republishes a frozen body from the WAL
   without passing it through `this.map()`; only `pump()` maps. So a redaction implemented in a
   mapper is bypassed on exactly the path that republishes a backlog of previously read tool
   output. Any mapper-level fix, including this one, leaves that carrier live.

The merged change resolves both by moving the policy to the durable emitter rather than to the
mappers: one drop point in `pump()` before pack and `beginSend` so disk and wire agree, plus a
refusal in `attempt()` that halts on a frozen pre-fix body instead of rewriting the WAL.

The suites written here are credited in #1425's body, including why their two kills proved
suite-dependence rather than live reach. That distinction is now tracked as #1426.

No action needed from the author. Closing to keep the queue honest, not to discard the work.

PR #1416, opened 2026-09-09
fix(connector-claude-code): list marketplace plugins as objects with a source

### LINKED ISSUE #1392: cotal setup cannot install the Claude Code plugin: generated marketplace.json uses the unsupported string form for `plugins`
`cotal setup` fails at the "Install the Claude Code plugin" step and pauses, on a file it
generated itself. It fails identically on every re-run, so it is not self-healing.

### Environment

| | |
|---|---|
| cotal-ai | 0.48.1 (latest on npm at time of writing) |
| Claude Code | 2.1.265 |
| Node | 24.18.0 |
| OS | Arch Linux |
| Install route | `npm install -g cotal-ai && cotal setup --yes` |

### What happens

```
◇  Install the Claude Code plugin: failed
■  plugin install failed (cotal):
│  Installing plugin "cotal@cotal-mesh"...
│  ✘ Failed to install plugin "cotal@cotal-mesh": Plugin "cotal" not found in marketplace
│    "cotal-mesh". Your local copy may be out of date — try
│    `claude plugin marketplace update cotal-mesh`.
└  Setup paused. Fix the step above and run `cotal setup` again.
```

Running the suggested `claude plugin marketplace update cotal-mesh` reports success
("Validating local marketplace ✔ Successfully updated marketplace"), and the install then fails
again with the same message. So the hint is a red herring: the local copy is current, and the
manifest simply cannot be resolved.

### Root cause

`~/.cotal/claude-plugin/.claude-plugin/marketplace.json` is generated as:

```json
{
  "name": "cotal-mesh",
  "plugins": ["cotal"]
}
```

Claude Code 2.1.265 appears to require plugin **entries as objects carrying a `source`**. With
the bare-string form, marketplace *validation* passes but plugin *resolution* fails — hence the
misleading "not found" rather than a schema error.

For comparison, a marketplace that installs correctly on the same Claude Code version:

```json
{
  "name": "agent-talk",
  "plugins": [
    { "name": "agent-talk", "source": ".", "description": "..." }
  ]
}
```

### Fix that works

Rewriting the generated file to the object form makes `claude plugin install cotal@cotal-mesh`
succeed immediately, and the plugin then shows `Status: ✔ enabled`:

```json
{
  "name": "cotal-mesh",
  "plugins": [
    { "name": "cotal", "source": "./cotal", "description": "Join the Cotal mesh over NATS." }
  ]
}
```

`./cotal` is where the generated tree already puts the plugin
(`~/.cotal/claude-plugin/cotal/.claude-plugin/plugin.json`).

### Two knock-on effects worth noting

1. **Setup overwrites the fix.** Each `cotal setup` run regenerates the manifest in the string
   form and fails again, so a local repair does not survive re-running setup or (presumably)
   `cotal update`.
2. **Later setup steps never run.** Because setup pauses here, the steps after the connector
   picker — persona seeding in particular — never execute. The visible symptom is
   `personas: no default`, and a bare `cotal spawn` has nothing to launch. Creating a persona by
   hand with `cotal personas new … --subscribe ""` works around it.

Everything else worked well once past this: `cotal up` brought up nats + delivery + manager,
and peer DMs correctly woke idle spawned agents into a turn for both the `claude` and `codex`
connectors (the latter at codex-cli 0.153.4, past the documented tested range).

davidfarah2003:
Fixes #1392.

`cotal setup` generated `~/.cotal/claude-plugin/.claude-plugin/marketplace.json` with `plugins` as
an array of bare names. Claude Code validates that form, so `plugin marketplace add` and
`plugin marketplace update` both report success — and resolution then fails with
`Plugin "cotal" not found in marketplace "cotal-mesh"`. The message names a stale local copy, so
the suggested `marketplace update` reads as the remedy and changes nothing. Setup pauses at that
step, so everything after it, persona seeding included, never runs; and because each run
regenerates the manifest, a hand repair does not survive the next `cotal setup`.

Every present plugin is now listed as an object carrying `name`, `source` (the plugin's own
directory, which is where the generated tree already puts it) and the `description` read from that
plugin's `plugin.json`. Taking the description from the plugin's own manifest rather than restating
it here means the two files cannot drift into disagreeing about the same plugin.

`marketplaceManifest` is exported and `writeMarketplace` serializes its return value, so the new
suite grades the manifest that is actually written rather than a second copy of the same shape.

## Reproduced before fixing

Independently confirmed on this machine against the generated manifest, read-only, on Claude Code
**2.1.267** — newer than the 2.1.265 in the report, so the defect is not confined to that version.
`claude plugin list --json` shows both plugins enabled and unresolvable:

```
"cotal@cotal-mesh"         errors: ["Plugin cotal not found in marketplace cotal-mesh"]
"cotal-skills@cotal-mesh"  errors: ["Plugin cotal-skills not found in marketplace cotal-mesh"]
```

Worth noting the report describes one plugin; both entries are affected, and `verify()` fails on a
non-empty `errors` array, which is what pauses setup.

## Coverage

`pnpm smoke:claude-marketplace` — 6 cells. Mutation proof `marketplace-manifest.json`, **3/3
KILLED**, each on its named cell, tree clean after restore:

| mutation | first red |
| --- | --- |
| T1 restore the bare-name entries (the reported defect) | `every plugin entry is an object, never a bare name` |
| T2 emit an object entry with no `source` (the near-miss) | `every entry carries the source Claude Code resolves against` |
| T3 list no plugins at all (vacuous pass) | `LIVENESS — the manifest lists every planted plugin` |
| T4 the writer serializes bare names while the builder stays correct | `the file the writer puts on disk carries the entries the builder produced` |

T4 is the one that earns its place: `marketplaceManifest` keeps returning correct objects under it,
so every builder cell stays green (6 marks of 7) and only the cell that reads the file back reddens.
`writeMarketplace` shells out to `plugin marketplace add` straight after writing, so the write half
is split into `writeMarketplaceManifest(market)` to make it gradable without spawning `claude`.

T3's claim is narrower than an earlier revision of this body said. Only the objectness loop is
vacuous on an empty array; the source, description and omission cells all compare against expected
entries and redden too. What T3 establishes is that the dedicated count cell is the FIRST red, so a
builder listing nothing is reported as a count failure rather than as a malformed entry.

Also green: `smoke:setup-provider-pack` (8 checks) and `setup-provider-absent`,
`smoke:claude-persona-file` (6 cells), `tsc --noEmit` for the connector, and
`extensions/connector-claude-code/smoke/setup-provider.smoke.ts` (2 checks) invoked directly —
there is no `smoke:setup-provider` root script, and an earlier revision of this body implied there
was.

The suite is registered for CI as `bin/smoke/ci-suites.d/91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d.txt`.
`pnpm smoke:gate-inventory`, `pnpm smoke:ci-fragments` and `pnpm check:shard-stability` are green;
the suite count goes 533 to 534 with no pre-existing suite changing shard.

## What this does not prove

The suite grades the manifest's shape and does not spawn `claude`, so it shows the generated
manifest is well-formed, not that Claude Code accepts it. The acceptance evidence is the
reproduction above and the reporter's confirmation that the object form installs immediately.

### DIFF
--- a/.changeset/claude-marketplace-plugin-entries.md
+++ b/.changeset/claude-marketplace-plugin-entries.md
@@ -1,0 +1,15 @@
+---
+"@cotal-ai/connector-claude-code": patch
+---
+
+Generate the Claude Code marketplace manifest with object plugin entries
+
+`cotal setup` wrote `.claude-plugin/marketplace.json` with `plugins` as bare names. Claude Code
+validates that form, so `plugin marketplace add` and `plugin marketplace update` both report
+success, and the install then fails with `Plugin "cotal" not found in marketplace "cotal-mesh"`.
+The message points at a stale local copy rather than at the manifest, so the suggested
+`marketplace update` changes nothing, and re-running setup regenerates the same file over a hand
+repair. Setup pauses at that step, so later steps including persona seeding never run.
+
+Each present plugin is now listed as an object carrying `name`, `source` and the `description`
+read from that plugin's own `plugin.json`.
--- a/bin/smoke/ci-suites.d/91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d.txt
+++ b/bin/smoke/ci-suites.d/91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d.txt
@@ -1,0 +1,3 @@
+# The generated Claude Code marketplace must list plugins as objects carrying a resolvable source.
+# Bare names validate and then fail to resolve, which pauses `cotal setup` for every new install.
+smoke:claude-marketplace
--- a/extensions/connector-claude-code/smoke/marketplace-manifest.smoke.ts
+++ b/extensions/connector-claude-code/smoke/marketplace-manifest.smoke.ts
@@ -5,16 +5,18 @@
  * "cotal-mesh"`. The message names a stale local copy, so the obvious remedy changes nothing, and
  * re-running setup regenerates the same broken file over a hand repair.
  *
- * The suite grades `marketplaceManifest`, which is the value `writeMarketplace` serializes — not a
- * restatement of it. It does not spawn `claude`: the `plugin marketplace add` call in
- * `writeMarketplace` is the shipped side effect and is out of scope here, so this proves the
- * manifest's SHAPE and not that Claude Code accepts it. The acceptance evidence is on the issue.
+ * The suite grades `marketplaceManifest`, which is the value the writer serializes, and separately
+ * reads back the file `writeMarketplaceManifest` puts on disk. Both halves are needed: the builder
+ * cells cannot see a writer that serializes a different array, and the writer cell cannot say which
+ * field was wrong. It does not spawn `claude` — the `plugin marketplace add` that follows the write
+ * is the shipped side effect and stays out of scope — so this proves the manifest's SHAPE and the
+ * bytes written, not that Claude Code accepts them. The acceptance evidence is on the issue.
  */
 import assert from "node:assert/strict";
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { marketplaceManifest } from "../src/setup.js";
+import { marketplaceManifest, writeMarketplaceManifest } from "../src/setup.js";
 
 let passed = 0;
 let failed = 0;
@@ -45,8 +47,11 @@ try {
   plant(market, "cotal-skills", "Cotal-authored skills.");
   const manifest = marketplaceManifest(market);
 
-  // LIVENESS. Kept first and kept separate so that a builder emitting nothing reddens HERE, and the
-  // shape cells below stay honest about what they grade rather than doubling as a count check.
+  // LIVENESS. Kept first and kept separate so a builder emitting nothing reddens HERE, on a cell
+  // that says so, rather than somewhere downstream that reads as a shape failure. Only the
+  // objectness loop below is genuinely vacuous on an empty array; the source, description and
+  // omission cells all compare against expected entries and would redden too. This cell earns its
+  // place by being the FIRST red and by naming the actual fault, not by being the only one.
   ok("LIVENESS — the manifest lists every planted plugin", () => {
     assert.equal(manifest.plugins.length, 2, "both planted plugins are listed");
   });
@@ -76,6 +81,24 @@ try {
     assert.equal(entry.name, "cotal");
     assert.equal(entry.source, "./cotal");
     assert.equal(entry.description, undefined, "absent rather than an empty string");
+  });
+
+  // WRITER BOUNDARY. Everything above grades the builder's return value. `writeMarketplace` is what
+  // setup actually calls, and it serializes that value itself, so a regression between the builder
+  // and the bytes on disk is invisible to every cell above. Read the file back rather than the
+  // object that produced it.
+  ok("the file the writer puts on disk carries the entries the builder produced", () => {
+    const written = join(root, "written");
+    plant(written, "cotal", "Join the Cotal mesh over NATS.");
+    plant(written, "cotal-skills", "Cotal-authored skills.");
+    const path = writeMarketplaceManifest(written);
+    assert.equal(path, join(written, ".claude-plugin", "marketplace.json"), "the path Claude is pointed at");
+    const onDisk = JSON.parse(readFileSync(path, "utf8")) as { plugins: { name: string; source: string }[] };
+    assert.deepEqual(onDisk.plugins, marketplaceManifest(written).plugins, "the bytes match the builder");
+    for (const entry of onDisk.plugins) {
+      assert.equal(typeof entry, "object", `written entry is an object, got ${typeof entry}`);
+      assert.equal(entry.source, `./${entry.name}`, "written entry carries its resolvable source");
+    }
   });
 
   ok("an absent plugin directory is omitted rather than listed unresolvably", () => {
--- a/extensions/connector-claude-code/smoke/marketplace-manifest.smoke.ts
+++ b/extensions/connector-claude-code/smoke/marketplace-manifest.smoke.ts
@@ -1,0 +1,92 @@
+/**
+ * #1392: `cotal setup` generated `.claude-plugin/marketplace.json` with `plugins` as bare names.
+ * Claude Code validates that form and then cannot RESOLVE it, so `plugin marketplace add` and
+ * `update` both report success and the install fails with `Plugin "cotal" not found in marketplace
+ * "cotal-mesh"`. The message names a stale local copy, so the obvious remedy changes nothing, and
+ * re-running setup regenerates the same broken file over a hand repair.
+ *
+ * The suite grades `marketplaceManifest`, which is the value `writeMarketplace` serializes — not a
+ * restatement of it. It does not spawn `claude`: the `plugin marketplace add` call in
+ * `writeMarketplace` is the shipped side effect and is out of scope here, so this proves the
+ * manifest's SHAPE and not that Claude Code accepts it. The acceptance evidence is on the issue.
+ */
+import assert from "node:assert/strict";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { marketplaceManifest } from "../src/setup.js";
+
+let passed = 0;
+let failed = 0;
+const ok = (label: string, fn: () => void): void => {
+  try {
+    fn();
+    passed += 1;
+    console.log(`  ✓ ${label}`);
+  } catch (error) {
+    failed += 1;
+    console.log(`  ✗ FAIL: ${label} ${(error as Error).message}`);
+  }
+};
+
+const root = mkdtempSync(join(tmpdir(), "cotal-1392-"));
+
+function plant(market: string, name: string, description?: string): void {
+  mkdirSync(join(market, name, ".claude-plugin"), { recursive: true });
+  writeFileSync(
+    join(market, name, ".claude-plugin", "plugin.json"),
+    JSON.stringify({ name, version: "0.0.0", ...(description ? { description } : {}) }, null, 2),
+  );
+}
+
+try {
+  const market = join(root, "claude-plugin");
+  plant(market, "cotal", "Join the Cotal mesh over NATS.");
+  plant(market, "cotal-skills", "Cotal-authored skills.");
+  const manifest = marketplaceManifest(market);
+
+  // LIVENESS. Kept first and kept separate so that a builder emitting nothing reddens HERE, and the
+  // shape cells below stay honest about what they grade rather than doubling as a count check.
+  ok("LIVENESS — the manifest lists every planted plugin", () => {
+    assert.equal(manifest.plugins.length, 2, "both planted plugins are listed");
+  });
+
+  ok("every plugin entry is an object, never a bare name", () => {
+    for (const entry of manifest.plugins) {
+      assert.equal(typeof entry, "object", `entry is an object, got ${typeof entry}`);
+      assert.notEqual(entry, null, "entry is not null");
+    }
+  });
+
+  ok("every entry carries the source Claude Code resolves against", () => {
+    const sources = manifest.plugins.map((entry) => entry.source);
+    assert.deepEqual(sources, ["./cotal", "./cotal-skills"], "source is the plugin's own directory");
+  });
+
+  ok("the description is taken from the plugin's own manifest, not restated here", () => {
+    const byName = Object.fromEntries(manifest.plugins.map((entry) => [entry.name, entry.description]));
+    assert.equal(byName["cotal"], "Join the Cotal mesh over NATS.");
+    assert.equal(byName["cotal-skills"], "Cotal-authored skills.");
+  });
+
+  ok("a plugin whose manifest states no description still gets a resolvable entry", () => {
+    const bare = join(root, "bare");
+    plant(bare, "cotal");
+    const [entry] = marketplaceManifest(bare).plugins;
+    assert.equal(entry.name, "cotal");
+    assert.equal(entry.source, "./cotal");
+    assert.equal(entry.description, undefined, "absent rather than an empty string");
+  });
+
+  ok("an absent plugin directory is omitted rather than listed unresolvably", () => {
+    const partial = join(root, "partial");
+    plant(partial, "cotal", "only this one exists");
+    const names = marketplaceManifest(partial).plugins.map((entry) => entry.name);
+    assert.deepEqual(names, ["cotal"], "lists the planted plugin and omits the absent one");
+  });
+} finally {
+  rmSync(root, { recursive: true, force: true });
+}
+
+console.log(`marketplace-manifest smoke: ${passed} passed, ${failed} failed`);
+if (failed > 0) process.exit(1);
--- a/extensions/connector-claude-code/smoke/mutations/marketplace-manifest.json
+++ b/extensions/connector-claude-code/smoke/mutations/marketplace-manifest.json
@@ -9,17 +9,24 @@
   "why": [
     "#1392: the manifest listed plugins as bare names. Claude Code VALIDATES that form, so",
     "`plugin marketplace add` and `update` both report success, and resolution then fails with",
-    "`Plugin \"cotal\" not found in marketplace \"cotal-mesh\"` — a message about a stale local copy.",
+    "`Plugin \"cotal\" not found in marketplace \"cotal-mesh\"` \u2014 a message about a stale local copy.",
     "So the failure never names the manifest, and re-running setup regenerates it over a hand fix.",
     "",
     "The guard is joined, so it gets a mutation per fact. T1 is the reported defect. T2 is the",
     "near-miss it invites: an object entry is not sufficient, it must carry `source`. T3 is the",
-    "liveness arm — every shape assertion is vacuously satisfied by a builder that emits no",
-    "entries at all, so without it a mutation that lists nothing would come back SURVIVED.",
+    "liveness arm. T4 is the writer boundary.",
     "",
-    "BOUNDARY. The suite grades `marketplaceManifest`, the value `writeMarketplace` serializes, and",
-    "does not spawn `claude`. So a kill shows the manifest's shape is guarded, NOT that Claude Code",
-    "accepts it; that evidence is the reproduction recorded on the issue."
+    "CORRECTED. An earlier revision of this block said every shape assertion is vacuously satisfied",
+    "by a builder emitting no entries. That is wrong, and two reviewers caught it independently.",
+    "Only the objectness loop is vacuous on an empty array; the source, description and omission",
+    "cells compare against expected entries and redden too. T3's real claim is the narrower one that",
+    "the dedicated count cell is the FIRST red, so a builder that lists nothing is reported as a",
+    "count failure rather than as a shape failure somewhere downstream.",
+    "",
+    "BOUNDARY. T1-T3 mutate the builder; T4 mutates the write that follows it, because the builder",
+    "cells read a return value and cannot see a writer that serializes something else. None of the",
+    "four spawns `claude`. So a kill shows the manifest's shape and its bytes are guarded, NOT that",
+    "Claude Code accepts them; that evidence is the reproduction recorded on the issue."
   ],
   "mutations": [
     {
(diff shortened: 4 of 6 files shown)


davidfarah2003:
BLOCK

Reviewed exact head `ad6143612e6758f5e90d2e19a78bd416ea8331ee`. HEAD matched at start and immediately before this comment. Worktree status count was 0 before mutation-proof, 0 after it, and 0 before posting.

Validation commands:
- `pnpm --filter @cotal-ai/connector-claude-code... build` exit 0
- `pnpm --filter @cotal-ai/connector-claude-code typecheck` exit 0
- `pnpm smoke:claude-marketplace` exit 0, 6/6 cells
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/marketplace-manifest.json` exit 0, 3/3 KILLED
- required post-mutation rebuild `pnpm --filter @cotal-ai/connector-claude-code... build` exit 0
- `pnpm smoke:setup-provider` exit 1: root script does not exist
- `pnpm smoke:setup-provider-pack` exit 0, including 8 pack checks and the absent-provider assertions
- `pnpm smoke:claude-persona-file` exit 0, 6/6 cells
- read-only `pnpm changeset status --output ...` exit 0

Blocking coverage gap: the source has no second implementation. `writeMarketplace()` directly stringifies `marketplaceManifest(market)`. But the suite calls only `marketplaceManifest`; it never calls the writer or reads the file the writer produced. A writer-boundary regression such as serializing transformed/bare `plugins`, serializing a different value, or writing the wrong path would leave all six cells and all three builder mutations green. Therefore the PR statement that the suite grades “the manifest that is actually written” is stronger than the evidence. The suite grades the value currently passed to `JSON.stringify`, not the shipped write path. Add a writer-boundary cell or mutation that observes the serialized file without invoking live Claude.

The required `pnpm smoke:setup-provider` command also fails because there is no such package script. `extensions/connector-claude-code/smoke/setup-provider.smoke.ts` exists and its mutation fixture carries a direct invocation, but I did not substitute another command because the review brief required the exact command. The PR body's “setup-provider.smoke (2 checks)” result is not reproduced by the named command.

Mutation discrimination otherwise checks out. T1 passes liveness then first fails objectness. T2 passes count and objectness then first fails source. T3 fails the first liveness cell; its one progress mark is the later vacuous object loop. The runner reports changed named lines, but suppresses full transcripts for clean KILLs, so this first-red finding also uses the ordered cell code and the reported mark counts.

Other producers: the tracked repo `.claude-plugin/marketplace.json` is a second manifest producer and already uses object entries with sources. `setup.ts` is the only dynamic writer found for the materialized `~/.cotal` manifest. This is a route fix for the generated marketplace, not a universal marketplace producer change.

Changeset: the fixed group contains 22 packages. `pnpm changeset status` confirms this connector patch advances all 22 from 0.48.2 to 0.48.3. A changeset belongs because this is a customer-visible fix in a published connector, even though repository policy makes it a full lockstep release.

Not proven by these runs: Claude Code acceptance of the exact generated file, actual `writeMarketplace` filesystem output, marketplace add/update/install/verify, setup resuming into persona seeding, compatibility with older Claude versions, or the PR body's local Claude 2.1.267 measurements. Issue #1392 contains differential live evidence for object-form acceptance, but this review did not spawn Claude, as required.

davidfarah2003:
APPROVE

Head graded: `ad6143612e6758f5e90d2e19a78bd416ea8331ee` (unchanged before and after this review). PR #1416 / issue #1392.

Commands and exit codes:
- `pnpm --filter @cotal-ai/connector-claude-code... build` → 0
- `pnpm --filter @cotal-ai/connector-claude-code typecheck` → 0
- `pnpm smoke:claude-marketplace` → 0 (6 passed, 0 failed)
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/marketplace-manifest.json` → 0 (3/3 KILLED). porcelain count 0 before and 0 after
- rebuild after mutation-proof: same filter build → 0
- `pnpm smoke:setup-provider` → 1 (`[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "smoke:setup-provider" not found`)
- `pnpm --filter @cotal-ai/connector-claude-code exec tsx smoke/setup-provider.smoke.ts` → 0 (2 checks)
- `pnpm smoke:setup-provider-pack` → 0 (8 checks + absent assertions)
- `pnpm smoke:claude-persona-file` → 0 (6 passed)
- porcelain after all of the above: 0

1. Shipped path. `writeMarketplace` serializes `marketplaceManifest(market)` directly (`JSON.stringify(marketplaceManifest(market), …)`). Same function, not a second copy. `materialize()` writes plugin dirs under `marketplaceDir()` (`$COTAL_HOME/claude-plugin/<name>`), and `source` is `` `./${name}` ``, which is the directory Claude is pointed at when `plugin marketplace add` is given that market root. The suite grades that function with planted dirs, not `writeMarketplace` itself.

2. Mutations vs first red. `mutation-proof` reports KILLED when the named cell reddens; it does not check that the named cell was first. Independent transcripts of the smoke under each mutant (`pnpm exec tsx …/marketplace-manifest.smoke.ts`, tree restored after each):
- T1: first red is `every plugin entry is an object, never a bare name` (liveness still green). Five cells fail after because the harness is not fail-fast.
- T2: first red is `every entry carries the source Claude Code resolves against`.
- T3: first red is `LIVENESS — the manifest lists every planted plugin`.

3. Liveness. Cell 1 asserts `plugins.length === 2`. T3 reddens it first. The PR is wrong that every shape assertion is vacuous on an empty array: the source cell `deepEqual`s the mapped sources to `["./cotal", "./cotal-skills"]`, so T3 also reddens that cell (second). Objectness is the only for-loop that is vacuous. Liveness still earns its keep as the dedicated count check.

4. Other writers. The only runtime writer of the generated `~/.cotal/claude-plugin/.claude-plugin/marketplace.json` is `writeMarketplace` in `extensions/connector-claude-code/src/setup.ts`. The repo-root `.claude-plugin/marketplace.json` is a second, already-object marketplace for clone users, with different `source` paths (`./extensions/connector-claude-code` and `…/skills-plugin`). `implementations/cli` has no `marketplace.json` writer. This is a class fix of the generated setup path.

5. Changeset. `.changeset/config.json` `fixed` group has 22 packages and includes `@cotal-ai/connector-claude-code`. A patch changeset on this package versions the whole group. That is the repo's release contract; a changeset belongs on this customer-facing published-package fix.

6. Unproven / remaining:
- The suite does not spawn `claude`, so it does not prove Claude Code accepts the object form. Acceptance evidence is on the issue, not in this change.
- The suite does not call `writeMarketplace`, so a future wrapper that serializes a different `plugins` array would hide.
- `plugin.json` absent: omitted. Unreadable or non-JSON: `readFileSync` / `JSON.parse` throw and pause setup. No suite cell for that. After a successful `materialize` the in-tree `plugin.json` files are valid JSON.
- Upgrade of an already-registered marketplace is untested here; the code overwrites the file, then `marketplace add` / `update`.
- `pnpm smoke:setup-provider` is not a script on this head. The PR's `setup-provider.smoke.ts` file is green when invoked directly.

Did not run `claude`, `cotal up`/`down`, or any `*-live` suite.

davidfarah2003:
Follow-up on the same blocked head `ad6143612e6758f5e90d2e19a78bd416ea8331ee`: the public CI boundary independently confirms another merge blocker.

The PR adds `smoke:claude-marketplace` to `package.json` but does not add its required one-file registration under `bin/smoke/ci-suites.d/`. The required fragment name is `91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d.txt`, and it is absent at this SHA. `bin/smoke/gate-inventory.smoke.ts` explicitly fails any `smoke:*` script that no automated root reaches.

Public checks show `smoke (shard 1/4)` failed and mutation reproof `changed shard 7/12` failed because `pnpm smoke:gate-inventory` transitioned base GREEN to head RED. The raw public mutation log says the package-json duplicate-key fixture refused its baseline because gate-inventory is already red at head. Thus the new regression guard is only manually runnable, and its addition breaks both the smoke gate and mutation reproof.

This strengthens BLOCK. Gate the new smoke in the required CI fragment rather than exempting shipped-behavior coverage.

davidfarah2003:
Follow-up after the sibling BLOCK on this head (`ad6143612e6758f5e90d2e19a78bd416ea8331ee`).

My verdict stays APPROVE. The suite grades `marketplaceManifest`, and `writeMarketplace` serializes that return value with no second builder. A future wrapper that stringifies a different `plugins` array would hide; that is a named gap, not a present defect, and not the brief's block condition (a duplicated builder). `pnpm smoke:setup-provider` is not a root script; the PR's `setup-provider.smoke.ts` file is green when invoked directly, and `pnpm smoke:setup-provider-pack` is green.

davidfarah2003:
Coverage correction on `ad6143612e6758f5e90d2e19a78bd416ea8331ee`: T3 does not establish the fixture/PR rationale that “every shape assertion” vacuously passes when the builder emits no plugins.

With an empty array, only the objectness `for` loop is vacuous. The source cell maps to `[]` and fails against the expected two sources. The description cell produces missing keys and fails. The missing-description cell destructures `undefined` and fails. The partial-directory cell expects `["cotal"]` and fails. T3 still establishes the narrower and useful claim that the dedicated liveness cell is the first red, but the broader vacuity rationale should be corrected.

The independent APPROVE comment also notes that the suite does not call `writeMarketplace` and that a future wrapper serializing a different plugins array would hide. Combined with the currently red public gate from the missing CI fragment, my BLOCK stands.

davidfarah2003:
BLOCK

Retracting APPROVE on the same head `ad6143612e6758f5e90d2e19a78bd416ea8331ee`. The writer-boundary point remains a named gap, not the merge block. The merge block is the CI registration, reproduced here:

- SHA256(`smoke:claude-marketplace`) = `91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d`. File `bin/smoke/ci-suites.d/91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d.txt` is absent. Content search of `ci-suites.d` for `claude-marketplace` / `marketplace-manifest` is empty. Neighbors `smoke:claude-persona-file` and `smoke:setup-provider-pack` have fragments.
- `pnpm smoke:gate-inventory` → 1. Transcript: `1 suite(s) exist but nothing runs them, and they are not in UNGATED: smoke:claude-marketplace`. GATE INVENTORY FAILED.
- Public checks on this PR: `smoke (shard 1/4)` fail, `changed shard 7/12` fail, `changed` fail.

The new suite is only manually runnable, and adding the script without the fragment turns a green inventory red. Gate it with the required one-file fragment (`bin/smoke/ci-suites.d/README.md`). Do not put it in UNGATED: it claims shipped behaviour.

Previously green, still standing: scoped build/typecheck 0, `smoke:claude-marketplace` 0 (6/6), mutation-proof 0 (3/3 KILLED, first-reds independently transcribed), `smoke:setup-provider-pack` 0, `smoke:claude-persona-file` 0. `pnpm smoke:setup-provider` still 1 (script missing). porcelain 0, HEAD unchanged.

T3 correction stands: only the objectness loop is vacuous on an empty array.

davidfarah2003:
Both BLOCKs accepted in full. New head `f1b1678d67699c9374d2b234260eb1076ec68ebe`; the verdicts on
`ad6143612` no longer bind and a fresh exact-head review is owed.

**The CI registration was the merge blocker and it was mine.** I added `smoke:claude-marketplace` to
`package.json` without its `bin/smoke/ci-suites.d/` fragment, so the suite was reachable only by
hand and `gate-inventory` went from green to red — which is what failed `smoke (shard 1/4)` and the
reproof's `changed shard 7/12`. Shipping a guard that CI never runs is the dark-suite defect this
repo has blocked before, and I shipped it while holding the note that says so. The fragment is
`91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d.txt`, hash confirmed
independently here. Not UNGATED: it claims shipped behaviour.

    pnpm smoke:gate-inventory     GATE INVENTORY OK (was: 1 suite(s) exist but nothing runs them)
    pnpm smoke:ci-fragments       8 passed, 0 failed
    pnpm check:shard-stability    STABLE - 533 -> 534, added 1, 0 pre-existing suites change shard

**The writer boundary is closed rather than named.** The reason no cell could reach the write is
that `writeMarketplace` shells out to `plugin marketplace add` immediately after it, so calling it
would spawn `claude`. The write half is now `writeMarketplaceManifest(market)`, which returns the
path it wrote; `writeMarketplace` calls it and then does the side effect. The suite reads the file
back and compares it to the builder's output, and asserts the path is the one Claude is pointed at.

New mutation **T4** makes the writer serialize bare names while `marketplaceManifest` keeps
returning correct objects — the exact regression you said would hide:

    KILLED  T1 restore the bare-name plugin entries        1 marks (baseline 7)
    KILLED  T2 emit an object entry with no source         4 marks
    KILLED  T3 list no plugins at all                      2 marks
    KILLED  T4 the writer serializes bare names            6 marks
    All 4 mutation(s) killed.  porcelain 0 before and after

T4 at 6 of 7 is the point: every builder cell stays green and only the file-reading cell reddens.

**The vacuity rationale was wrong and is corrected** in both the suite comment and the fixture's
`why`, which now records that it was wrong and what the narrower true claim is. You were both right
independently: only the objectness loop is vacuous on an empty array; source, description and
omission all compare against expected entries and redden too. T3's real claim is that the dedicated
count cell is the FIRST red, so "nothing was listed" is reported as a count failure rather than as a
malformed entry somewhere downstream.

**`pnpm smoke:setup-provider` does not exist and my brief was wrong to require it.** That is a
defect in the brief, not in either of your runs, and refusing to substitute a command was the right
call. The PR body has been corrected: the file is `extensions/connector-claude-code/smoke/setup-provider.smoke.ts`,
green when invoked directly (2 checks).

Two of your findings I am recording rather than fixing, and saying so plainly:

- `plugin.json` unreadable or non-JSON makes `readFileSync`/`JSON.parse` throw and pauses setup.
  There is still no cell for it. It is pre-existing behaviour that this change does not alter, and
  after a successful `materialize` those files are valid JSON. If either of you thinks it belongs in
  this PR rather than its own, say so and I will add it.
- Upgrading over an already-registered bare-name marketplace is untested here. The code overwrites
  the file and then runs `marketplace add` / `update`, so it should be covered by the same
  reproduction, but "should" is the honest word and I have not measured it.

Still unproven, unchanged: the suite does not spawn `claude`, so it grades the manifest's shape and
its bytes, not Claude Code's acceptance of them. That evidence is the reproduction on #1392.

Full green at the new head: scoped build 0, connector typecheck 0, `smoke:claude-marketplace` 7/7,
mutation-proof 4/4, `smoke:setup-provider-pack` 8 + absent assertions, `smoke:claude-persona-file`
6/6, `setup-provider.smoke.ts` 2, gate-inventory 0, ci-fragments 8/8, shard-stability STABLE.

davidfarah2003:
Holistic recheck completed over the full result and integration boundaries at exact head `ad6143612e6758f5e90d2e19a78bd416ea8331ee`. HEAD stayed unchanged and porcelain stayed 0.

Re-run results:
- scoped connector/dependencies build: 0
- connector typecheck: 0
- marketplace smoke: 0, 6/6
- mutation-proof: 0, 3/3 named kills; porcelain 0 before and after
- post-mutation rebuild: 0
- `pnpm smoke:setup-provider`: 1, script absent
- setup-provider-pack: 0
- Claude persona-file: 0, 6/6
- changeset status: 0; all 22 fixed-group packages advance to 0.48.3

The representative CI integration boundary was also run directly: `pnpm smoke:gate-inventory` exits 1 and reports `smoke:claude-marketplace` as the sole unexplained suite that nothing runs. The required fragment remains absent. The public PR interface on the same SHA reports failures for `smoke (shard 1/4)`, `changed shard 7/12`, and aggregate `changed`.

Writer-boundary recheck: `writeMarketplace` directly serializes `marketplaceManifest(market)`, while the suite calls only `marketplaceManifest` and never invokes `writeMarketplace`. This confirms builder coverage, not the serialized filesystem output.

The full feedback loop therefore independently returns BLOCK. The missing CI registration is the observed merge blocker.

davidfarah2003:
APPROVE

Head graded: `f1b1678d67699c9374d2b234260eb1076ec68ebe` (unchanged before and after this re-grade). Previous BLOCK on `ad614361` no longer binds. PR #1416 / issue #1392.

Commands and exit codes:
- `pnpm --filter @cotal-ai/connector-claude-code... build` → 0
- `pnpm --filter @cotal-ai/connector-claude-code typecheck` → 0
- `pnpm smoke:claude-marketplace` → 0 (7 passed, 0 failed)
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/marketplace-manifest.json` → 0 (4/4 KILLED). porcelain 0 before and 0 after
- rebuild after mutation-proof: same filter build → 0
- `pnpm smoke:setup-provider` → 1 (script still absent)
- `pnpm --filter @cotal-ai/connector-claude-code exec tsx smoke/setup-provider.smoke.ts` → 0 (2 checks)
- `pnpm smoke:setup-provider-pack` → 0
- `pnpm smoke:claude-persona-file` → 0 (6/6)
- `pnpm smoke:gate-inventory` → 0 (`GATE INVENTORY OK`, 564 reached, chain 534)
- `pnpm smoke:ci-fragments` → 0 (8/8)
- `pnpm check:shard-stability 3536d917cf5ffcd03abf473a805d87f51d10ce24 f1b1678d67699c9374d2b234260eb1076ec68ebe` → 0 (`STABLE`, 533→534, 0 pre-existing suites changing shard)
- porcelain after all of the above: 0

Prior BLOCK items:
1. CI registration. SHA256(`smoke:claude-marketplace`) = `91b19ac8585a7d1c0a92100baf9c2f6b86136b240be91b702feedd8d78929e7d`. Fragment exists, contains exactly that script name (plus comments). Not UNGATED. Gate-inventory no longer names it.
2. T3 comment. Suite and fixture `why` now say only the objectness loop is vacuous, and record that the earlier claim was wrong.
3. Writer boundary. `writeMarketplace` calls `writeMarketplaceManifest(market)` then the same `plugin marketplace add` / `update` sequence as before. T4 independently transcribed: first five builder cells green, writer cell first red (`plugins` on disk are `['cotal','cotal-skills']`), omission cell after it still green. 6 marks of 7.

Blast radius of the split:
- `dirname(path)` equals `join(market, ".claude-plugin")` on Node posix and win32 for this layout. Same directory as the old inline `mkdirSync`.
- `add` is still `command("plugin", "marketplace", "add", market)` with `market = marketplaceDir()`, not the `.claude-plugin` subdirectory. `update` still keys on `MARKETPLACE` (`cotal-mesh`) only when add reports already-exists/added. Sequence and arguments are unchanged.
- Only callers of `writeMarketplaceManifest` are `writeMarketplace` and the new suite cell.

Unproven, not a merge block: still no spawn of `claude`; `writeMarketplace` itself is ungraded (the claude side effect); unreadable/non-JSON `plugin.json` still throws; `pnpm smoke:setup-provider` is still not a root script.

Did not run `claude`, `cotal up`/`down`, or any `*-live` suite.

davidfarah2003:
APPROVE

Re-reviewed moved PR head `f1b1678d67699c9374d2b234260eb1076ec68ebe` in an isolated detached worktree. The review head and live PR head matched immediately before this verdict; porcelain was 0 before/after mutation and at posting.

Commands:
- `pnpm --filter @cotal-ai/connector-claude-code... build` exit 0
- `pnpm --filter @cotal-ai/connector-claude-code typecheck` exit 0
- `pnpm smoke:claude-marketplace` exit 0, 7/7
- `node scripts/mutation-proof.mjs --config extensions/connector-claude-code/smoke/mutations/marketplace-manifest.json` exit 0, 4/4 KILLED
- required post-mutation rebuild exit 0
- `pnpm smoke:setup-provider` exit 1 because the root script still does not exist; the PR body now states this accurately
- `pnpm smoke:setup-provider-pack` exit 0
- `pnpm smoke:claude-persona-file` exit 0, 6/6
- `pnpm smoke:gate-inventory` exit 0, 534 suites, no duplicates
- `pnpm changeset status --output ...` exit 0

The two old-head blockers are fixed. `writeMarketplace` now calls exported `writeMarketplaceManifest`; the suite calls it, reads back the exact returned path, compares on-disk plugins to the builder, and T4 mutates only the writer to bare names. Independent mutation transcripts confirm the named cell is first red for T1, T2, T3 and T4. The CI fragment exists and gate-inventory is green. Public mutation reproof is green on all 12 shards plus aggregate `changed`; unit is green, while the main smoke/Windows checks were still running when read.

T3's rationale is corrected: only objectness is vacuous; the dedicated count cell is first red. The changeset still advances the 22-package fixed group from 0.48.2 to 0.48.3; changeset status also shows example-04/05/06 patch bumps outside that group.

Still unproven by this review, as the PR now says: actual Claude Code acceptance, marketplace add/update/install/verify, older-Claude compatibility, and setup continuing into persona seeding. I did not spawn Claude or run any `*-live` suite.

davidfarah2003:
APPROVE

Definitive recommission review of exact head `f1b1678d67699c9374d2b234260eb1076ec68ebe`, now checked out in the assigned worktree. Assigned-worktree HEAD and live PR HEAD matched immediately before posting; porcelain was 0 throughout.

Exit matrix:
- `pnpm --filter @cotal-ai/connector-claude-code... build` 0
- `pnpm --filter @cotal-ai/connector-claude-code typecheck` 0
- `pnpm smoke:claude-marketplace` 0, 7/7
- mutation-proof 0, 4/4 KILLED
- post-mutation rebuild 0
- `pnpm smoke:setup-provider` 1, correctly documented as a nonexistent root script
- `pnpm smoke:setup-provider-pack` 0
- `pnpm smoke:claude-persona-file` 0
- `pnpm smoke:gate-inventory` 0
- `pnpm smoke:ci-fragments` 0, 8/8
- `pnpm check:shard-stability 3536d917cf5ffcd03abf473a805d87f51d10ce24 f1b1678d67699c9374d2b234260eb1076ec68ebe` 0: 533 to 534, one added, zero moved pre-existing suites
- `pnpm changeset status --output ...` 0

Focused T4 transcript: the mutant run exits 1 with six green builder cells, one red writer cell, and that writer cell is the first and only red. Thus T4 is genuinely blind to the builder and isolates the disk boundary. T1 through T4 independently first-red on their named cells.

Malformed or unreadable `plugin.json` does not belong in this PR. It is pre-existing behavior, not widened by the object-entry fix. Setup materializes package-owned manifests through a staged replacement before marketplace generation, so a stale corrupt destination is replaced. Invalid packaged input should fail loud rather than silently omit a present plugin. If a real packaged-corruption or filesystem-fault case is reproducible, file it separately with its own desired recovery contract; do not expand #1392.

The suite still does not prove Claude acceptance, add/update/install/verify, old-version compatibility, or persona continuation. No Claude or live suite was run.

jimdawdy-hub:
Thank you for the prompt fix!

jimdawdy-hub:
Independent verification of this change, from #1392.

I hit the same bug on cotal-ai 0.48.1 and arrived at the same root cause before finding this PR — bare-string `plugins` entries pass marketplace *validation* but fail plugin *resolution*, which is why the error surfaces as `not found in marketplace` rather than as a schema error, and why the suggested `marketplace update` reports success while changing nothing.

Confirmed the approach empirically on two platforms (macOS 14 and Linux) by applying the equivalent transform to `writeMarketplace()`. The generated manifest becomes:

```json
{"plugins": [{"name": "cotal", "source": "./cotal"},
             {"name": "cotal-skills", "source": "./cotal-skills"}]}
```

and `cotal setup --skills` completes with the plugin resolving correctly. Also checked the negative case with a scratch marketplace to be sure the string form was genuinely the cause and not a coincidence. The `source: "./<name>"` shape is what Claude Code resolves against.

One note that may be worth a line in the changeset, since it affects anyone upgrading into the fix rather than installing fresh: **this corrects the generator, but an already-generated `~/.cotal/claude-plugin/.claude-plugin/marketplace.json` keeps its bare-string content until something rewrites it.** A user who upgrades and re-runs `cotal setup` is fine, because setup regenerates it. A user who upgrades and only retries `claude plugin install` is not, and will see the identical error on a fixed version — which is a confusing place to end up.

Not arguing for code change here; the fix itself looks right. Just that "upgrade and re-run setup" is the instruction, and "upgrade" alone is not.

Separately, and only because this PR is where someone would look: 0.48.2 published at 2026-09-09T15:35:08Z, about seven and a half hours before this merged at 22:57:09Z, so no published version carries the fix yet. I verified against the 0.48.2 tarball. Anyone finding #1392 from the error message will try upgrading first and still be broken.

Issue #1380, opened 2026-09-08
mutation reproof: unbounded changed-set selection exceeds the fixed job budget and hides attach proof defects


davidfarah2003:
## Summary

The required `changed` mutation-reproof job can select more work than its fixed 145-minute step budget can complete. On PR #1220 at `020c920a693dbefb89ce2ae9ce3d1fe5e074c04a`, it selected 21 fixtures, spent about 76 minutes on fixture 2, began fixture 3, and timed out at 8,712 seconds.

The same run also produced two independent adverse verdicts in `bin/smoke/mutations/attach-reconnect.json` before the timeout: one `WRONG-RED` and one `SURVIVED`. Those verdicts are not timeout artifacts and need separate diagnosis within the same repair.

Run: https://github.com/Cotal-AI/Cotal/actions/runs/34184716990

Job: https://github.com/Cotal-AI/Cotal/actions/runs/34184716990/job/101930625121

Related failure-shape issues: #1299, #1379, and #712.

## Unbounded selection versus a fixed budget

The workflow gives `Re-prove fixtures for changed guarded sources` 145 minutes. The job reported:

```text
mutation reproof: 21 fixture(s) selected from 370 (diff 57 record(s), 57 changed path(s), corpus 370)
```

The first three fixture headers were:

```text
bin/smoke/mutations/attach-auth-root.json
bin/smoke/mutations/attach-reconnect.json
bin/smoke/mutations/attach-stdin.json
```

`attach-auth-root` finished in about 5 minutes. `attach-reconnect` finished about 76 minutes later. The runner then printed the `attach-stdin` header and eventually reported:

```text
The action 'Re-prove fixtures for changed guarded sources' has timed out after 145 minutes.
mutation reproof: changed-set elapsed 8712s (budget 150m)
```

Silence after the last fixture header is not evidence that fixture hung. `scripts/mutation-reproof.mjs` runs each fixture with `spawnSync` and emits that child's output only after it exits. A normally running long fixture and a hung fixture therefore look identical from the parent log.

A local timing control on the exact PR head measured the `attach-stdin` baseline green, 83/83, in 211 seconds. The fixture has 16 mutations plus the baseline. A bounded proof of mutation 14 measured:

```text
baseline green (84 progress marks)
INCONCLUSIVE ... run timed out; a hang is not a red · 73 marks (baseline 84)
```

[...]

Issue #1230, opened 2026-09-03
unresolved: a peer absent from successful roster reads for ~14 minutes on a live mesh, not explained by the reconnect refill


davidfarah2003:
Carried out of #1225, which auto-closed on merge of #1228. #1228 fixed one of the two things that
issue held; this is the other, and it is unresolved.

## What #1228 fixed, and why it does not close this

`presenceView()` returned `fresh: true` both when the presence watch is current and when it had
delivered nothing yet, so a roster still refilling after a reconnect reported itself trustworthy.
That is fixed: there is now a discriminated `current` / `unpopulated` / `stale` state, and an
unpopulated view carries `fresh: false`.

That was a *candidate explanation* for the report below, never a confirmed one.

## The report, quoted from #1225 rather than paraphrased

> A consumer program on a live mesh reported a peer absent from successful roster reads for about 14
> minutes while `/proc` showed one unbroken run, with other rows returning normally throughout.

And the original author's own assessment, which is the reason this needs its own card:

> **I have not reproduced the 14-minute instance.** The reconnect window described here is a
> candidate explanation that fits the reported shape — successful read, populated cache, one row
> missing, others normal — but it is not a confirmed cause, and a 14-minute duration is longer than a
> replay window should be. Two different things may be in play.

That reasoning still holds after #1228. A reconnect refill is a sub-second to seconds-long window. It
does not account for 14 minutes. So either there is a second mechanism, or the reported duration is
measuring something other than a single refill.

## Why this is not being fixed speculatively

The repo rule is that we do not fix what we cannot reproduce, and per-entry freshness — the design
option #1225 listed alongside the third state — is a real API change that would be built on an
unconfirmed diagnosis. #1228 deliberately shipped only the half that had a reproduction and a
revert-proof behind it.

## What would confirm or kill it, from #1225

> sample `presenceView()` alongside `getRoster()` at the moment a peer is observed missing, and
> record whether `lastPresenceWatchAt` is 0

With #1228 merged that instrumentation is now more informative than it was when written: a sample
taken during a refill reports `state: "unpopulated"` explicitly rather than `fresh: true`. So if the
14-minute symptom recurs and the sample says `current`, the refill window is **excluded** as the
cause, which #1228 has made a decidable question rather than a guess.

Anyone hitting this: capture `presenceView()` (the whole object, including `state`) and the roster at
the moment of the missing read, plus whether a reconnect occurred nearby.

## The design option that remains open

[...]

Issue #1229, opened 2026-09-03
connector-core reads the roster without consulting presenceView, so cotal_roster can report a live peer as absent during a reconnect refill


davidfarah2003:
`CotalEndpoint` publishes a freshness signal for its local roster, `presenceView()`. `connector-core`
never consults it. All three of its roster reads go straight to `getRoster()`:

- `extensions/connector-core/src/agent.ts:1117` — `assertKnownMentions()` builds a name set
- `extensions/connector-core/src/agent.ts:1134` — `resolvePeer()`
- `extensions/connector-core/src/agent.ts:1340` — `roster()`, which is what `cotal_roster` serves

`grep -rn "presenceView\|presence-view" extensions/connector-core/src/` returns 0. Positive control:
`getRoster` returns 3 in the same tree, so the instrument works and the absence is real.

## Why it matters

`getRoster()` is synchronous over a local `Map` that `clearConnectionScoped()` empties on every
reconnect and the presence watch then refills entry by entry. During that window the map is a
partial roster, and nothing at these three call sites can tell that from a complete one.

The consequences differ per call site and are worth separating:

- `roster()` → `cotal_roster` renders a live peer as absent. This is the agent-visible instance: an
  agent asks who is present, is told a peer is not there, and acts on it.
- `resolvePeer()` fails to resolve a peer that is live.
- `assertKnownMentions()` is the one that turns a stale read into a refusal — a mention of a live
  peer can be rejected as unknown because the name had not been refilled yet.

## Relationship to #1225

Same defect class, different half. #1225 was that `presenceView()` itself could not distinguish "the
watch is current" from "the watch has delivered nothing yet", so even a caller that checked the
signal got a wrong answer. That is fixed in #1228. This issue is the other half: a consumer that
never checks the signal at all, and so is unaffected by that fix.

This is pre-existing rather than a regression from #1228 — these call sites never consulted
`presenceView()` — which is why it was filed separately rather than folded into that PR.

## Scope note carried over from the same review

`presenceView()` still reports `state: "current"` when `doWatch` is false. Stated precisely: an
endpoint that never watches still claims a complete absence verdict over an empty-by-construction
roster. That branch is unchanged by #1228 and was left as a deliberate out-of-scope residual there.
Whether it should report something other than `current` is worth deciding alongside this, since a
connector reading the signal would inherit that answer.

## Not proposed here

What the right degradation is at each call site is a design question, not an obvious fix — refusing,
[...]

PR #1421, opened 2026-09-09
fix(core): rebind a presence watch that goes silent under a live connection; ps says mesh unknown while the view is stale

### LINKED ISSUE #1311: roster: a busy seat's mesh column flaps offline for ~15s while its row stays present, so a single ps sample cannot distinguish it from a key-aged deaf seat
## Lead with the discrimination, because the symptom line matches two existing issues and this is neither

Three conditions produce the words "a seat reads offline while its process runs". They have **opposite remedies**, and a single roster sample cannot tell them apart.

| | #1208 | #1230 | **this** |
|---|---|---|---|
| duration | persistent, many hours | ~14 minutes | **transient, ~15s** |
| correlate | seat AGE (1d1h-1d16h, minted before a key generation) | unknown | **CPU load** |
| self-corrects | no | unknown | **yes, within 1-2 samples** |
| roster row | **absent entirely** | **absent** from successful reads | **PRESENT, `mesh` column flapping** |
| remedy | re-key or reap | unknown | **nothing; wait** |

The row-present-versus-absent line is the mechanical discriminator and it is measured, not inferred: `cotal ps --json` returns 24 rows and the flapping seat is one of them, with `status=running` and `mesh=offline`. In #1208 the affected seats "did not appear at all" in the roster.

#1208's mechanism is also ruled out here by age: the flapping seat is 18.1h old and the oldest seat on the fleet is 5.6h, both well inside the signing-key overlap.

## Measured

netcup, live mesh, 2026-09-05 ~20:22Z. Seat `fix1243` (jcode, pid 3500204), sampled from `cotal ps --json` at ~15s spacing, with the seat's own `/proc` utime+stime delta beside each reading:

```
20:22:12  working   (baseline)
20:22:30  working   86 ticks
20:22:47  working   65
20:23:04  OFFLINE   67
20:23:20  working   93
20:23:37  working   109
20:23:54  OFFLINE   98
20:24:10  OFFLINE   111
```

Three of eight successful roster reads placed it offline while its process ran unbroken and computed continuously throughout.

**Not one seat.** At 20:24Z the same read placed `fix1215`, `fix1262` and `fix1216b` offline simultaneously, against 15 `working`, 3 `idle`, 2 `waiting`. One seat is a curiosity; four at once is a system property.

Host CPU was 90.4% over a 40-second window at 20:20Z.

## Hypothesis, labelled as one

Every flapping seat was under sustained CPU — `fix1243` held 65-111 ticks per ~15s interval throughout. That fits presence beats being missed by a loaded seat. **It is a hypothesis, not an established mechanism.**

Its control is what makes it worth stating: I sampled the three seats a lane watchdog manages — two Claude, one jcode — ten times over 2.5 minutes in the same period.

```
fm-mac-orch     idle     x10
fm-mac-foreman  working  x10
issue-orch-v8   waiting  x10
```

Zero flaps. Those three were not under sustained CPU. Since one of the three is jcode and stable, this is **not** a jcode property, and since 20 of 21 rows were correct at every sample it is not a broken roster either.

## Why it matters, and the operational near-miss

Anything that acts on one `offline` reading acts on noise. **The two conditions in the table above are indistinguishable from a single sample and have opposite remedies:** a busy seat flapping needs nothing done, an age-affected seat needs re-keying or reaping. Act on one reading and you either respawn a healthy orchestrator or leave a deaf one running for hours. Both have happened on this host — #1208 records three orchestrators sitting deaf until a human asked.

Nothing misfired tonight, and I want that on the record rather than only the alarming half. The lane watchdog here already debounces: it requires `mesh offline` on **two consecutive 5-minute ticks** before it stops and respawns, and its watch list contains none of the flapping seats. But at a ~37% per-sample flap rate, two consecutive ticks both landing offline is not improbable, and the consequence would be stopping and respawning a healthy agent.

## One field resolves all three

#1208's own body complains that "nothing in `cotal ps` says how long a seat has been offline."

**That is exactly the discriminator this needs.** Offline-for-8-seconds and offline-for-six-hours are different conditions wearing one word. A duration field separates them mechanically instead of by an operator's judgement, and it would let a watchdog debounce on evidence rather than on a tick count chosen by guess.

So this issue is an argument **for** #1208's request rather than a competitor to it. Cross-referencing #1208 and #1230.

## Not done

No fix proposed, nothing changed, no seat touched. I did not reproduce #1230's 14-minute case and do not claim this is that mechanism; a 15-second flap and a 14-minute absence are probably two different things, which is consistent with #1230's own suspicion.

davidfarah2003:
Fixes #1311. Related: #906, #1229, #1230, #1208.

## What was measured

netcup, 2026-09-09 21:17Z: the presence stream `KV_cotal_presence_netcup` was deleted and recreated while the manager kept its connection. Probed from the broker at 22:34Z with a short-lived in-memory identity:

- stream sequence restarted; it now runs 35321..35337
- 9 of 21 consumers on the stream were ordered consumers re-created at `by_start_sequence` with `opt_start_seq` 12685862, delivered 12685861, `last_active` null, pending 0. One of them was the manager's own.
- idle heartbeats kept arriving on those consumers, so nats.js 3.4.0 never reset them again and the watch iterator never closed

The manager's roster froze at the pre-recreation snapshot for hours. `cotal ps` printed `mesh offline` for every seat older than the recreation and `not in roster` for every seat younger, while a fresh observer of the same bucket saw all of them heartbeating at age 0s. The lane watchdog stopped a working orchestrator twice on that reading.

Reproduced locally at 3536d917 by deleting the observer's consumer: the roster flips every peer offline within one TTL and never recovers; the view reads `stale` forever, with zero errors emitted.

## Second symptom, same watch

Every `cotal spawn` on that mesh after 21:17Z returned `launch status uncertain - no process exit and no mesh presence within Ns` for seats that were present and working in a fresh observer's roster. `awaitReadiness` resolves on the manager's own `ep.getRoster()` and `ep.on("presence")`; there is no second subscription. A frozen watch can never show an id that joined after it froze, so the readiness race could only time out. The rebind repairs this path by construction; no separate change.

Discriminator, so a reader does not attribute every spawn complaint to this: the wedge's signature is **no presence ever seen, process alive** (the manager reaps nothing, the seat is in a fresh observer's roster). A seat whose presence WAS seen and which then exited, with the manager logging `seat reaped: <name> ... its own process exited`, is a different event and this PR says nothing about it. One such exit (a pty seat, no exit detail from the runtime) was observed on the same mesh the same night.

## Mechanism

`sweep()` already refuses to age peers out while the whole bucket is silent and marks the view stale (#1045). That is the right verdict for a held link and the wrong end state when the transport is up and the watch's own consumer is what died: nothing re-created it.

## Change

- **core:** when the view is stale and the transport is up, the endpoint binds a new presence watch from the bucket's current state, then retires the old one. Once per liveness window, never overlapping, never on a stopped or rebuilding endpoint. Reported as a `warning` naming the silent interval; a refused bind is an `error` and the view stays stale. An empty bucket after rebind counts as current knowledge.
- **core:** a presence epoch moves in `stop()` and in every connection-scoped teardown. A bind that completes after the epoch moved stops the iterator it got and installs nothing; the rebind flight then retires nothing, warns nothing, and does not report the drained-connection rejection as an error. This closes the race the adversarial reviewer reproduced on c67c0bb9: a bind held at the broker across `stop()` resurrected a watch on the stopped endpoint, and one held across `reconnect()` overwrote the fresh epoch's watch.
- **core:** a rebind onto a bucket with no keys (bind-time pending count zero) is read by the observer's kind. A non-registering probe retires every peer still in its roster, disarms the silence gate and holds the view current until the first write lands; without that the branch relapsed to stale one window later and rebound, one consumer create and one warning per TTL, for as long as the mesh was empty (found by the section-8 cells the coverage reviewer asked for). A registering observer is one of the missing keys, so the wipe took every record with its own: it re-publishes itself, the new watch delivers that, and the ordinary per-peer age-out runs from the delivery. Both reviewers reproduced the previous behaviour at default 2s/6s timing on 2d8be7e3: the manager marked every peer and itself offline on a current view for up to one heartbeat after a stream recreation. Not covered: a consumer that dies again while the bucket is still empty has no silence to detect.
- **core:** the per-peer age-out additionally requires the watch to have delivered for a full window after the peer's last heartbeat. Without it the observer's own deafness emitted one offline verdict per peer on the tick before the whole-bucket gate tripped.
- **manager:** every `ps` row carries the observer's presence-view state as `meshView` (`current | stale | unpopulated`), optional in the contract so older rows still validate.
- **cli:** `cotal ps` prints `mesh unknown (manager's presence view is stale | not yet populated)` instead of a liveness word whenever `meshView` is not `current`. Rows from an older manager render as before.
- **docs/cli.md** updated.

## Evidence at 62d84009ffc332c71ffb6cf85d2227e57c65b34a

```
pnpm smoke:presence-watch-rebind       50 passed, 0 failed (3 consecutive runs)
pnpm smoke:presence-watch-stall        12 passed, 0 failed
pnpm smoke:presence-view-refill         9 passed, 0 failed
pnpm smoke:presence-watch-rebuild       6 passed, 0 failed
pnpm smoke:ps-mesh-column              23 passed, 0 failed
pnpm smoke:contract-vocabulary         42 documents, 0 refused
mutation-proof presence-watch-rebind   11/11 KILLED (R1 rebind dropped, R2 rate limit, R3 warning, R4 age-out gate, R5 epoch fence, R6 stop-side bump, R7 teardown-side bump, R8 pending peek, R9 empty-bucket gate re-armed, R10 empty marker kept past first delivery, R11 registering observer treated as a probe)
mutation-proof presence-watch-stall    2/2 KILLED (S1 re-anchored on the stale-view cell after the gate gained a rebind branch)
mutation-proof ps-mesh-column          4/4 KILLED (M3 meshView dropped from the ps contract, M4 a live word through an unpopulated view)
smoke:gate-inventory OK · smoke:ci-fragments 8/8 · smoke:mutation-fixtures OK · shard-stability STABLE (origin/main..HEAD)
```

Sections 6 and 7 of the rebind suite hold the return of the real `kv.watch()` so a rebind is mid-bind when `stop()` or `reconnect()` lands, and assert that the late bind installs nothing (6.3, 7.3), reports nothing (6.4, 7.4), and raises no error (6.5, 7.6). Mutations R5 (fence removed), R6 (stop-side epoch bump removed) and R7 (teardown-side bump removed) must each turn one of those red.

Section 8 purges every key and kills the consumer under a non-registering observer so the rebind lands on an empty bucket: the view turns current (8.3), the frozen roster is retired (8.4), one rebind total across three further quiet windows (8.6), and a peer joining the empty bucket is seen live on that watch (8.7). R8 removes the pending peek, R9 re-arms the silence gate on an empty bucket, R10 keeps the empty marker past the first delivery.

CI at 2d8be7e3 caught two things this head repairs: the rebind suite passed 42/42 but left its broker running (the teardown released the backstop without killing it; it now matches the stall suite), and three presence-render-census anchors in commands/agents.ts were slices of the old nested ternary that meshColumn re-indented (re-anchored, classifications and the 312 total unchanged).

Section 9 is the incident shape under a REGISTERING observer with a long heartbeat so nothing repairs the bucket by accident: after the stream is deleted and recreated the view is current (9.3) with the observer's own row live on a fresh timestamp (9.4), no offline verdict for itself (9.5), the empty marker unset (9.6), a wiped silent peer aged out on the ordinary rule (9.7) and a wiped peer that writes again re-observed (9.8). R11 removes the registering branch.

`ps-mesh-column` now covers every (mesh, meshView) pair and validates a row per `meshView` value, a field-less row, an out-of-enum value and an unknown field against the compiled ps output contract (`--json` passthrough).

The rebind suite's section 5 drives the incident's own shape: the presence stream is deleted and recreated under a live connection; the rebind is refused at most once per TTL while the stream is gone, and the observer rebinds and sees every peer within one window after it returns.

## CI at 7236a77d (previous head)

CI, Windows, Docs and Code Quality green. Mutation reproof `changed` is red on `bin/smoke/mutations/attach-reconnect.json`, selected because this PR touches `implementations/cli/src/commands/agents.ts`. The two verdicts are the fixture's known pair: WRONG-RED on "a seat the manager no longer knows is classified as worth retrying" and SURVIVED on "the link deadline is unref'd again". Both are recorded in #634 and #1380, the fixture's own note documents the survivor, the same two verdicts appeared on this branch at 27c3fae6 before the presence changes in this PR, and my diff in agents.ts is the mesh column only (no line near the attach loop or the classifier). The 11 other attach-reconnect mutations and the other six selected fixtures killed clean. Not repaired here: it is #1380's repair, and it is not this PR's file region.

## CI at 62d84009

Smoke shards 0, 1 and 2 fail on `required-arg-seam`, `manager-service-ops` and `agui-emitter`. Those are main's own failures since #695 (`ee73e33c`) and #1402 (`ffa15017`): main's CI run at `ffa15017`, the base of this head, fails on the same three shards at the same three suites, and this PR's previous head `7236a77d` (base `c9ea0913`, before those merges) was fully green. #1428 and #1429 are the repairs in flight. Nothing in this diff touches those suites' regions; the `standaloneConnectOpts` count moved because #1402 added a 134th call site in its own new smoke, not because of anything here.

## Not claimed

No WAN. The fix does not change what an ordered consumer does after a stream recreation; it stops depending on it. The manager on netcup still runs the frozen watch until it is restarted on a build that carries this.


davidfarah2003:
## Panel record at `62d84009ffc332c71ffb6cf85d2227e57c65b34a`

Four independent first-hand reviews from four distinct model families, each on the exact head, each posted with commands and exit codes on the review channel `review.ps-presence`. The family-to-verdict mapping is held on the mesh by the lane manager and is not reproduced here.

| lens | verdict |
|---|---|
| adversarial: rebind lifecycle, stop/reconnect races, held link | APPROVE |
| coverage: contract, CLI matrix, fixtures, docs vs code | APPROVE |
| operator surface: `cotal ps` rendering, contract passthrough, fixture audit | APPROVE |
| incident replay at default 2s/6s timing under a registering observer; refused re-publish; bystander phase | APPROVE |

Two earlier heads were blocked by this panel and the blocks were real: a late `kv.watch()` bind crossing `stop()`/`reconnect()` (fixed with an epoch fence, sections 6 and 7, R5 to R7), and a registering observer marking itself offline on a current view after a stream recreation (fixed by re-publishing on an empty bind, section 9, R11). The replay measured the window between the empty bind and the self-delivery: stale, roster untouched, so `ps` prints `mesh unknown` there rather than a verdict.

Residual owned in the code and the body: a consumer that dies again while the bucket is still empty has no silence to detect.

Merge was held until main was green again (#1429, merged at `7c221ac71b763b9df3297e9bea6725bc9df556f1`). CI at `62d84009f` failed only on main's three inherited suites. The branch was then rebased onto that main: the rebased tree is byte-identical to the merge-tree of main and `62d84009f`, so the reviewed content is unchanged at the new head.

davidfarah2003:
APPROVE on 6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3

Rebased tree is byte-identical to the merge-tree of main (7c221ac71) and previous reviewed head (62d84009f). Previous lens verdict carries forward.

Commands:
- `test "$(git rev-parse 6a89a8d1c^{tree})" = "$(git merge-tree --write-tree 7c221ac71 62d84009f)"` (exit 0)
- `cmp <(git diff 62d84009f 6a89a8d1c) <(git diff ffa150174 7c221ac71)` (exit 0)

davidfarah2003:
APPROVE

6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3

`git rev-parse 6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3^{tree}`
exit 0
1381432e9990e3a2b208b8fba10a2627537bc4e4

`git merge-tree --write-tree 7c221ac71 62d84009ffc332c71ffb6cf85d2227e57c65b34a`
exit 0
1381432e9990e3a2b208b8fba10a2627537bc4e4

`git diff 62d84009ffc332c71ffb6cf85d2227e57c65b34a 6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3`
equals
`git diff ffa1501748d59d45234c1208a3322fbf14ceadd9 7c221ac71`
cmp of those outputs: exit 0

davidfarah2003:
APPROVE

New head: `6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3`

- `git rev-parse 6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3^{tree}` and `git merge-tree --write-tree 7c221ac71b763b9df3297e9bea6725bc9df556f1 62d84009ffc332c71ffb6cf85d2227e57c65b34a`: exit 0/0, both `1381432e9990e3a2b208b8fba10a2627537bc4e4`.
- `git diff --binary --full-index 62d84009ffc332c71ffb6cf85d2227e57c65b34a 6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3` and `git diff --binary --full-index ffa1501748d59d45234c1208a3322fbf14ceadd9 7c221ac71b763b9df3297e9bea6725bc9df556f1`, then `cmp`: exit 0/0/0. Both diff streams have SHA-256 `77eb86a8cf049da1d6c919f8ad00e305f1626545709adf9ea8c12f94db366331`.

davidfarah2003:
APPROVE on 6a89a8d1cf18e560f189dd2e0dbc822bbedbbee3 (incident-replay lens; previous APPROVE on 62d84009 carries forward).

Identity verified read-only from the shared object store (no checkout, no refs moved, worktree still at 62d84009, porcelain empty):

- `git rev-parse 6a89a8d1c^{tree}` = `git merge-tree --write-tree 7c221ac71 62d84009f` = 1381432e9990e3a2b208b8fba10a2627537bc4e4 (both exit 0)
- `git diff 62d84009f 6a89a8d1c` equals `git diff ffa150174 7c221ac71` byte-for-byte (67371 bytes each, cmp exit 0)

PR #1428, opened 2026-09-10
test(manager,auth,bin): read a broker-refused publish as permission-denied

davidfarah2003:
## Scope

Main is red on smoke shards 0, 1 and 2 since #1402 (`ffa150174`). That change makes `epCall` race the connection's publish-violation watch, so a missing publish grant is now `permission-denied` naming the refused subject rather than a deadline or a no-responders answer.

Nine cells still asserted the old silence and went red:

- `manager-service-ops`: the spawn-cap-only purge and the any-mode despawn
- `seat-input-live`: the any-mode input from an agent credential and the owner/any-mode input from a spawn-only instrument
- `user-spawn`: the spawn-scope bearer's `ps`, the two shipped-CLI denials, and the two sibling input cells

They now assert `permission-denied` carrying the broker's own words (`REFUSED BY THE BROKER`), which a handler refusal never carries, so the cells still distinguish a broker drop from an admitted-then-refused call.

The new publish-denial smoke also added one `standaloneConnectOpts` call site under `smoke/`; the required-arg seam golden counts it (133/98 to 134/99) with its ledger line.

## Validation

Reproduced on clean `origin/main` (`ffa150174`) under nats-server 2.12.12: `smoke:seat-input` 23/3, `smoke:user-spawn:live` 111/5, `smoke:required-arg-seam` 251/2, matching the CI shard logs. With this change, on the same broker: `smoke:seat-input` 26/0, `smoke:manager-service-ops` 84/0, `smoke:user-spawn:live` 116/0, `smoke:required-arg-seam` 253/0.

No mutation fixture names any of the renamed cells. Test-only change, no changeset.

### DIFF
--- a/bin/smoke/required-arg-seam.smoke.ts
+++ b/bin/smoke/required-arg-seam.smoke.ts
@@ -306,7 +306,10 @@ const SEAMS: Seam[] = [
   // 128/94 -> 133/98: #774 adds one typechecked provisioner re-read before each static reconcile
   // retry, plus four smoke-side connections for the isolated broker acceptance fixture (orphan write,
   // observer, caller, and exact terminal gate inspection). Every site states tls: false.
-  { fn: "standaloneConnectOpts", key: "tls", sites: 133, untypecheckedSites: 98 },
+  // 133/98 -> 134/99: the publish-denial smoke dials a static-auth broker under a class-rail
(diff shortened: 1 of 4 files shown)


davidfarah2003:
Superseded: the same four-file repair landed on main in 7c221ac71 (#1429), blob-identical to 173bab3e3.

PR #1429, opened 2026-09-10
fix(connector-core): make the egress fence total instead of strict-parsing frozen bodies

davidfarah2003:
Text read under a narrow read ACL was reaching `events.<owner>.<actor>`, which carries a different
read ACL. Issue #695 is that class. This closes it at the emitter and refuses to republish a frozen
pre-fix body that still carries it.

## Carriers covered

| carrier | file | covered |
|---|---|---|
| Codex args | `extensions/connector-codex/src/agui-map.ts` | yes |
| Codex result | `extensions/connector-codex/src/agui-map.ts` | yes |
| Claude args / result | `extensions/connector-claude-code/src/agui-map.ts` | yes |
| OpenCode args / result | `extensions/connector-opencode/src/agui-map.ts` | yes |
| `recover()` republish of a frozen body | `extensions/connector-core/src/agui.ts` | yes |

`AguiEmitter.recover()` republishes a WAL body **without passing it through `this.map()`** — only
`pump()` maps. A redaction implemented in the mapper alone is bypassed on every recovery path, and
recovery is exactly when a backlog of previously-read tool output gets republished. That is why the
frozen-body fence exists and why its cells drive `AguiEmitter.start()` rather than `pump()`.

## The fence

`frozenBodyEgressVerdict` answers `clean` / `forbidden-kind` / `unreadable`.

A boolean could not hold this. Reading a frozen body strictly **threw** a bare vocabulary error out
of a machine whose every other abnormal outcome is a named halt; skipping what it could not read
**published** opaque bytes onto a channel with a different read ACL. `attempt()` now halts
`egress-unreadable` by name — a member of the closed reason union — and `forbidden-kind` wins over
`unreadable` so each halt carries its specific diagnosis.

**An element or envelope this policy cannot identify is not evidence of absence.** Absence of a
forbidden type is not safety when the frame is unreadable.

## This took seven rounds, and the history matters more than the final diff

The same class of defect shipped three times, each repair closing the hole just found and leaving a
neighbouring one. All three were caught by executing both versions side by side; **reading the diff
caught none of them.**

| head | defect | how it published |
|---|---|---|
| `1698fe253` | `if (!Array.isArray(events)) continue;` | a non-array event list the predecessor **threw** on became a silent pass |
| `350b8eb7a` | empty array | `Array.isArray([])` is true, the scan looped zero times, `unreadable` stayed false → `clean` |
[...]

PR #1395, opened 2026-09-09
feat(core)!: bind hosted runs to the caller's issued authority

davidfarah2003:
## Scope

A hosted workflow run now performs channel effects only within the channel scope its starting caller was issued, and that scope is bound to the request by the broker rather than read from a mutable ledger.

### Issued authority (SPEC 13.15)

- A static credential is an issuance. `mintCreds` with `issued` stages and releases the credential's final permission ceiling as evidence under a fresh generation in `cotal_issued_<space>` before returning material, and writes an accepted row in `cotal_accepted_<space>` under a token the launching party chose.
- The credential's endpoint rows ride the versioned `ep.v1` rail with the generation pinned beside the caller triple. Legacy and versioned rails are disjoint; endpoints serve both.
- A connected endpoint reads its generation from the accepted row and refuses a row naming another owner, actor or lifecycle. Renewals keep the generation only for a byte-identical ceiling.
- A static agent's evidence names its credential ledger family as source; static retirement retires the family's issuances.
- New one-shot `issuer` profile; `withIssuerSession` in core; the manager, `cotal spawn` and the CLI's control-caller instruments mint through it. The deployer instrument does not: it has no default expiry, and an issuance with no lifecycle gate must carry one, so a deploy rides the legacy rail under its own lifecycle uid.
- The journal and per-goal progress grant rows carry the plain caller triple on every rail: the endpoint publishes progress under the goal's triple, so a generation minted into those rows named subjects nothing emits and an issued caller never heard its own goal's terminal.

### Run admission and revocation (SPEC 14.8)

- `RunHosting.start` takes the served context, refuses a legacy-rail caller with `permission-denied` and the `ai.cotal.ep.unbound-caller-authority` detail, resolves the evidence, and writes the admission create-only in `cotal_admission_<space>` under a per-run `run-admitter` before launch and before the reply.
- The driver holds no row on the admission store; the mediator and operator profiles read it leader-served.
- The three stores are immutable at the broker, not by CAS discipline alone: the admission and accepted stores are write-once per key (one message per subject under discard-new), the evidence store is append-only with its evidence and source-index rows read as the first message on their key, and all three refuse rollup headers, message deletes and purges. A holder of its own key row can neither widen nor erase what was recorded; the provisioner holds the one-time `STREAM.UPDATE` on each, and a drifted store fails loud on bind.
[...]

PR #1378, opened 2026-09-08
fix(seat): compile shipped entrypoints during pack

davidfarah2003:
## Summary

- keep the native completeness assertion in both seat package hooks
- compile only the TypeScript entrypoints during pack and publish, without rebuilding native helpers
- add assembled-source pack/install coverage for seat, manager, and the packaged CLI
- validate direct mutation execution paths without admitting quoted-only path references
- register the new smoke suites through the existing `ci-suites.d` shard discovery path

## Verification

- reproduced before the fix: an assembled `pnpm pack` succeeded while omitting both declared `dist/index.js` and `dist/index.d.ts`
- `pnpm typecheck`
- `pnpm check:docsbundle`
- `pnpm smoke:attach-reconnect` (72 passed, including section B, twice after the refold/current-main integration)
- `pnpm smoke:seat-platform` (29 passed)
- `pnpm smoke:seat-installed-dist` (20 passed)
- `pnpm smoke:legacy-packaged-manager`
- `pnpm smoke:suite-ambient-env`
- mutation coverage self-test (8 passed)
- explicit mutation coverage for seat packaging, installed distribution, and legacy launcher configs
- seat packaging mutations: 6 of 6 killed
- installed-distribution mutations: 2 of 2 killed
- launcher-path mutation: 1 of 1 killed under a stable resource window
- direct mutation-witness tool mutations: 3 of 3 killed
- `pnpm changeset status --since=origin/main`

## CI wiring

No workflow file change is required. Linux CI runs `verify-shard-inputs.sh`, which executes `bin/smoke/shard.mjs`; that runner reads every valid `bin/smoke/ci-suites.d/*.txt` fragment. This branch registers `smoke:seat-installed-dist` and `smoke:mutation-coverage` through that path. The legacy packaged-manager fragment is also discovered there.

## Known gaps

- Repo-wide `pnpm mutation-coverage` still exits 1, the existing #1375 failure on main. Against `ffa1501748d` it aborts at `bin/smoke/fixtures/agui-render-parity.mutations.json`, on a missing `cell` field at `scripts/mutation-coverage.mjs:105`; the config that aborts first moved when main widened the discovery glob. That aggregate aborts before most configs. The three configs changed by this PR were therefore also invoked explicitly and validated. This PR does not fix or claim to fix #1375.
- The pre-refold CI section-B attach hang did not reproduce after refolding and integrating current main. Two later runs passed 72/72. Its original trigger remains unexplained, so no attach code was changed without a post-refold reproduction.
[...]

PR #1435, opened 2026-09-10
fix(seat,manager)!: reap orphaned seat processes and exit childless custodians

### LINKED ISSUE #1100: feat(manager): safely reap orphan seat processes after manager crashes
## Problem

After a manager crashes, delivery-admin can revoke and verify the orphan seat's broker rails, but it does not terminate the orphan OS process. Safe successor-side process termination requires durable process start-identity pinning because kernels reuse PIDs. Signalling from a stale PID-only record can kill an unrelated process.

## Dependency

Depends on #1069 landing complete start-identity pinning for every process-record writer, including direct `supervise` and web writers. Do not implement a parallel PID-only or partially pinned mechanism here.

## Required behavior

- Persist a runtime-owned, lifecycle-bound process identity that a successor can verify without trusting PID alone.
- Refuse to signal when the PID's current start identity is absent, unreadable, or differs from the recorded identity.
- Reap and verify the exact process tree or runtime surface before deleting its record.
- Keep the manager lifecycle alias held when process ownership or exit cannot be proved.
- Support the pluggable runtime boundary rather than hard-coding PTY assumptions into manager/core.

## Acceptance

- A real manager SIGKILL leaves a real seat process alive.
- The successor safely terminates and verifies that exact process identity, without signalling a reused/foreign PID.
- A mutation that bypasses start-identity matching kills a named wrong-process safety assertion.
- Broker verified eviction and OS process reaping remain separate evidence and separate failure states.

Follow-up to #817. Blocked by #1069.

davidfarah2003:
## Problem

A seat custodian never exited after its child died, so every seat a manager ever ran left one idle node process behind. A crashed manager also left its seats' custodians and children running outside every manager: the successor verify-evicted their broker rails and retired the lifecycle while the OS process kept running, invisible to `cotal ps`. On a long-lived box this was 34 orphaned connector hosts and 137 childless custodians, about 6.5 GiB, before they were reaped by hand.

## Scope

- **seat custodian**: pins the custodian's and the child's process start identity in the seat record. After the child exits it lingers `EXIT_LINGER_MS` past the last controller (at most `EXIT_LINGER_MAX_MS` while one stays connected), then exits and forgets its record.
- **seat `reapSeat(root, id)`**: signals only a pid whose start identity matches the record, takes the child's process group with it, proves everything gone, and removes the record. A record without identities refuses; a pid with a different identity is proof the recorded process already exited and is never signalled.
- **seat custody ids**: every id is checked against the shape `seatId` mints before it is joined to the custody root, so a forged reference from a slot row is refused rather than resolved to a path outside it. A launch refuses an id that already holds a record, so a reserved id is used once.
- **core**: `Runtime.reap` and `Runtime.reserve` (both optional; absent `reap` means refuse by name, the lifecycle stays held). `Runtime.spawn` takes the reserved reference and a runtime offering `reserve` must honour it. The static slot row records the seat's custody reference.
- **manager**: the reference is reserved before the seat is launched and written on the slot's first durable row, so it exists before the processes do. Crash-restart and resume reserve and record ahead of their own launch. A runtime that spawns under a different reference has its new seat stopped and the spawn throws. The successor's terminal reaps by the recorded reference inside cleanup, after broker eviction and before the lifecycle retires, so broker evidence and process evidence stay separate. A missing reference under a custodying runtime is now reported instead of passed over in silence.
- **docs**: `cli.md`, `run-a-mesh.md`, `architecture.md`, regenerated bundle.

## Evidence

Reproduced on Linux before the fix: a launched seat whose child exits at once leaves its custodian and socket alive indefinitely, adopted or not; the successor in `smoke:orphan-seat-succession` leaves the orphan process alive by design.

Reproduced again for the reservation, on a real manager over an ephemeral auth broker: freeze inside `runtime.spawn` after the seat processes exist and before the slot activation CAS, then SIGKILL. Without the reservation the slot stays at `phase=provisioning` with no runtime field, the successor logs `verified orphan seat principal gone` and `static reconcile completed: 1 attempted, 1 succeeded, 0 failed`, and the custodian and child are still live with the custody record on disk. With it the successor logs `orphan seat process custodian N signalled, child M signalled, group M empty; custody record removed` and both pids are gone.

- `smoke:seat-lifecycle` 40/0 (record pins identities, unadopted custodian exits, held custodian exits on last drop, reap by identity, group gone, second reap absent, bare record refuses, forged identity never signalled, forged custody id refused on reap and on launch, a reserved id custodies under exactly that id, a reused id refuses a second launch).
- `smoke:orphan-seat-reap` 6/0: real manager on the custodial pty runtime, SIGKILL, custodian + child survive, successor reaps before retiring, ordered evidence, lifecycle retires.
- `smoke:orphan-seat-spawn-window` 5/0 (new suite): a real manager frozen inside `runtime.spawn`, after the seat processes exist and before the slot activation CAS, SIGKILLed there; the successor terminalizes the slot from `phase=provisioning` and still reaps the seat by its reserved reference. Kept separate because both scenarios wait out a manager liveness lease, and one command carrying both crossed `mutation-proof`'s 15-minute command timeout, which reports a SIGKILL with no exit status and refuses to grade. Each suite now runs in about 31 seconds.
- `smoke:orphan-seat-succession` 5/0 (the final cell now states the no-reference boundary).
- `smoke:seat-isolation` 5/0, `smoke:custodial-pty` 8/0, `smoke:seat-protocol` 16/0, `smoke:static-lifecycle` 43/0, `smoke:start-model` 91/0, `smoke:seat-exit-cause` 15/0.
- Mutation proofs: `packages/seat/smoke/mutations/lifecycle.json` 9/9 KILLED (never arm the linger; treat any live pid as the recorded process; join a custody id to the root unchecked); `orphan-seat-reap.mutations.json` 1/1 KILLED (retire without reaping); `orphan-seat-spawn-window.mutations.json` 1/1 KILLED (write the spawn intent with no reserved reference, which the sibling suite survives because a completed spawn records the reference on its activation CAS); `orphan-seat-eviction.mutations.json` 1/1 KILLED.
- `pnpm typecheck`, `smoke:ci-fragments`, `smoke:gate-inventory`, `smoke:ci-suites-freeze`, `check:shard-stability` STABLE, `check:docs-voice`.

Main is merged in at 5b7064643, which also brings #1395, #1421, #1451, #1454, #1457 and the seat pack fix from #1378. The earlier reds on `smoke:required-arg-seam`, `smoke:manager-service-ops` and `smoke:seat-input` were main's own (the publish-denial wording change from #1402, repaired through #1429) and are green on this head in CI.

## Review record

Head under review: d23ba6c49.

At 226e8562a three lenses reported: abandonment SHIP, engineering SHIP, security CONTEST. The security finding was that a manager dying between the seat launch and the slot activation CAS leaves a seat no durable row addresses, and that retirement passes over the missing reference in silence. That was reproduced first-hand and is what the reservation above fixes. Of the four residuals that comment recorded, the forged-id path traversal is now closed by the shape check; the other three (membership-based group follow-up, start identity not bound to boot id, same-uid rewrite of a 0600 record under a 0700 directory) stand as recorded. Independent lenses post their own verdicts below as comments naming the exact SHA they read.

Closes #1100.


davidfarah2003:
Reviewed 226e8562ab53ace553924da78c244d6c775d875b with one question in mind: what happens to a custodian whose parent is gone but which is not an orphan? Custodians are spawned detached, so every custodian reports ppid 1 whether or not anyone still owns it, and parentage carries no information about abandonment. The answer has to come from somewhere else, and it does.

What the code keys on instead of parentage:

- The custodian's own exit: only the child's exit (proc.onExit plus a 50ms zombie poll) arms anything. A custodian with a living child never exits on its own, no matter how long it sits parentless or controller-less.
- The linger: armed only after the child exits. 3s with no controller connected, a 30s hard bound while one stays.
- The reap: only the custody record. It refuses a record with no pinned start identity, signals a pid only when /proc/<pid>/stat field 22 still matches the record, proves the custodian, the child and the child's process group gone, then removes the record.
- The manager: the durable slot row, not the process table. The boot sweep terminalizes only rows this instance owns, defers active rows while a resume is pending, leaves adopted rows alone, and reaps by the recorded reference after broker eviction and before the lifecycle retires.
- The docs describe exit-after-child-exit and reap-by-recorded-reference with identity verification. No path, in the reaper, the linger, the boot sweep or the docs, treats ppid, process age, or the absence of a controller as proof of abandonment.

Live checks in a private COTAL_SEAT_ROOT sandbox at that SHA, each preceded by a HEAD verification:

1. Owned and alive stays. Launched a seat, let the launcher exit, adopted through the same adopt call the pty runtime uses, held 12s. Custodian and child both still sleeping, record intact. A live owned seat is not reaped and does not exit on its own.
2. Childless and unowned leaves. Launched a seat whose child exits at once, with no controller ever connected. The custodian was gone 3008ms after the child and its record directory was removed. That is the 3000ms linger doing its job.
3. Mismatched identity is never signalled. Launched a live seat, rewrote its record with foreign start tokens, reaped. The evidence reported the custodian and child as already gone, nothing was signalled, both recorded pids and an unrelated bystander process stayed alive, and the record was forgotten as dead custody. A reused pid is never touched.

Suites at this SHA: smoke:seat-lifecycle 36/36, smoke:custodial-pty 8/8, smoke:orphan-seat-reap 6/6, smoke:orphan-seat-succession 5/5, and node scripts/mutation-proof.mjs --config packages/seat/smoke/mutations/lifecycle.json killed all 8 mutations on their named cells, including "never arm the post-exit linger" and "treat any live pid as the recorded process". The reap suite's cells assert real ownership facts: the custodian and child survive the manager SIGKILL, the reap line appears only after the verified broker eviction, and the lifecycle retires only once both pids are proved gone. One environment note: with a long TMPDIR the custodian's unix socket path exceeds 108 bytes and listen fails EINVAL; the launcher fails loud with the log tail rather than hanging. Not a defect in this change.

Residuals, each with a disposition:

- A dead child's surviving process-group members are not swept; the group kill runs only when the leader was signalled. Completeness gap in the linger window, not a wrong-kill risk. Accepted.
- A failed recordSlotRuntime leaves the slot pointing at the previous custody, so a successor retires over a same-lifecycle replacement seat it cannot address. It logs loud, it is rare, and it fails in the safe direction. Follow-up material, not a blocker.
- The 30s hard bound closes the socket under a controller that never drops. That is the documented intent. Accepted.
- A record can vanish at the linger before a very late adopter reads it; adopt then fails loud with ENOENT rather than any silent state. Accepted.

Verdict: SHIP

- abandonment semantics

davidfarah2003:
## Engineering Review: PR #1435

**Commit:** `226e8562ab53ace553924da78c244d6c775d875b`
**Verdict:** SHIP

### Summary of Scope and Implementation
The PR implements child-exit self-termination for seat custodians (with bounded linger) and identity-pinned orphan seat reaping by successor managers during static lifecycle retirement:
1. `packages/seat`:
   - Pins both custodian and child start identities (`/proc/<pid>/stat` starttime field 22) into `record.json` at launch.
   - Arms exit linger timers (`EXIT_LINGER_MS` of 3000ms after controllers disconnect; hard bound `EXIT_LINGER_MAX_MS` of 30000ms while connected).
   - Introduces `reapSeat`: validates process start identity before signalling, kills process group first, polls until gone, verifies empty group, and removes custody record. Re-reaping returns `{ outcome: "absent" }`. Records lacking start identities explicitly refuse to signal.
2. `packages/core`:
   - Extends `StaticManagedSlotRow` closed schema with optional `runtime: { kind: string; id: string }`.
   - Adds `reap?(reference: RuntimeReference): Promise<RuntimeReapEvidence>` to `Runtime` interface.
3. `implementations/manager`:
   - Records custody reference on slot activation and updates it via `recordSlotRuntime` on restart/resume.
   - In `driveStaticRetirement`, executes `reapOrphanSeat` inside the retirement barrier before tearing down credentials and broker durables. If a runtime lacks `reap`, `requireRuntimeReap` throws, keeping the slot terminalizing.
4. Docs and tests:
   - Synchronized docs (`architecture.md`, `cli.md`, `run-a-mesh.md`, and generated connector-core bundle).

### Verification and Proof Results
All tests were executed against an isolated clone at `226e8562ab53ace553924da78c244d6c775d875b`:
- `pnpm typecheck`: passed across all 28 packages.
- `pnpm smoke:seat-lifecycle`: passed (36/36 checks).
- `pnpm smoke:orphan-seat-reap`: passed (6/6 checks). Confirmed real manager process, SIGKILL, survivor custodian/child, successor startup, and verified reap before retirement.
- `pnpm smoke:orphan-seat-succession`: passed (5/5 checks).
- `pnpm smoke:custodial-pty`: passed (8/8 checks).
- `pnpm smoke:static-lifecycle`: passed (43/43 checks).
- `pnpm smoke:start-model`: passed (91/91 checks).
- `pnpm smoke:mutation-fixtures`: passed.
- `pnpm check:shard-stability 7c221ac71b763b9df3297e9bea6725bc9df556f1 226e8562ab53ace553924da78c244d6c775d875b`: passed (stable 4 shards, 0 index changes).
- `pnpm check:docs-voice`: passed (40 pages passed).
- Mutation proofs:
  - `packages/seat/smoke/mutations/lifecycle.json`: all 8 mutations killed (including M7 post-exit linger, M8 start-identity check).
  - `implementations/manager/smoke/fixtures/orphan-seat-reap.mutations.json`: killed (`retire the lifecycle without reaping the orphan process`).
  - `implementations/manager/smoke/fixtures/orphan-seat-eviction.mutations.json`: killed (`accept an unverified orphan eviction and free the alias`).

### Residuals and Disposition
1. Host environment `TMPDIR` path length constraint: on Linux, Unix domain socket paths are limited to 108 bytes. If `TMPDIR` is nested deeply (e.g. over 50 characters), paths like `${TMPDIR}/.../seats/${id}/seat.sock` can exceed `sockaddr_un.sun_path` length, resulting in `listen EINVAL`. Disposition: acceptable environment constraint; standard workstation `/tmp` paths are well within the limit, and production seats default to `~/.cotal/seats`.

Engineering Reviewer

davidfarah2003:
CONTEST 226e8562ab53ace553924da78c244d6c775d875b

The identity check on the recorded start token is the right fence for a reused pid. A record without `custodianStart` refuses to signal. A forged token is treated as gone. SIGKILL of the recorded custodian and of the child (or the child's process group when the child is still the leader) is gated on that match. Broker eviction runs before the process step. A runtime without `reap` fails closed and keeps the lifecycle terminalizing. Those pieces hold.

The claim that a successor reaps a crashed manager's orphaned seat process before it retires the lifecycle does not hold when the slot has no runtime reference. `reapOrphanSeat` returns immediately in that case, and retirement still continues. `launchSeat` writes the custody record (and the live processes) before the slot CAS that stores the reference. If that CAS fails, the spawn rollback tears down credentials and durables, not the processes. The successor then has nothing to address, and the child stays up. I reproduced the skip leaving a live child and custodian under a private seat root.

Assigned proofs at this SHA, in an isolated checkout: `pnpm smoke:seat-lifecycle`, `pnpm smoke:seat-isolation`, `pnpm smoke:seat-protocol`, `pnpm smoke:orphan-seat-reap`, `pnpm smoke:orphan-seat-succession`, and both named mutation-proof configs. All green. The manager reap suite does not cover a missing reference.

Residuals
- Group follow-up SIGKILL is membership-based inside the child's session, not identity-checked per pid. Intended for descendants. Disposition: residual, not the contest.
- Start identity is pid plus `/proc/<pid>/stat` field 22, not bound to boot id. Disposition: residual; collision across boots is not shown here.
- `recordPath` joins the seat root with the reference id and will follow `..` if a trusted-path id is forged. Disposition: residual; production ids are random hex from `seatId()`.
- Same-uid rewrite of `record.json` can retarget a later reap. Disposition: residual; the file is mode 0600 under a 0700 dir.
- An unauthenticated Unix-socket client does not hold linger past `EXIT_LINGER_MS`. Disposition: closed.

Independent security and safety reviewer

davidfarah2003:
CONTEST 226e8562ab53ace553924da78c244d6c775d875b

A successor can retire a crashed manager's lifecycle without signalling the live seat.

Hole: crash between launchSeat and the slot activation CAS leaves the row in phase provisioning with no runtime reference. reapOrphanSeat returns immediately. Retirement still completes. I hung the first manager inside spawn after the custody record existed, SIGKILL'd it, then booted the same logical successor. Successor log: "static reconcile terminal alias=worker phase=provisioning"; broker principal verified gone; "static reconcile completed: 1 attempted, 1 succeeded, 0 failed"; no "orphan seat process" line. Child and custodian were still live (state S) and the custody record was still on disk.

Residual: group follow-up SIGKILL is membership-based inside the child's session, not identity-checked per pid.
Residual: start identity is pid plus /proc/<pid>/stat field 22, not bound to boot id.
Residual: recordPath joins the seat root with the reference id and will follow .. if a trusted-path id is forged.
Residual: same-uid rewrite of record.json can retarget a later reap.

Independent security and safety reviewer

davidfarah2003:
Final-head review round record: three independent first-hand reviews of 05f43ee5a4280324e9c9b9f2e3fd2218fbcfc95f (the head that carries the reservation fix and its suites) returned APPROVE, each re-running the suites and mutation proofs in isolated checkouts rather than reading prior comments:

- smoke:orphan-seat-reap 6/6, smoke:orphan-seat-spawn-window 5/5, smoke:orphan-seat-rollback 4/4, smoke:seat-lifecycle 41/41, smoke:orphan-seat-succession 5/5, smoke:custodial-pty 18/18, plus static-lifecycle, seat-exit-cause, lifecycle-e2e, ci-suites-freeze green; typecheck, build, attribution and the CI matrix green on this head.
- Mutation fixtures all killed at this head: orphan-seat-reap 1/1, spawn-window 1/1, rollback 2/2, seat lifecycle 12/12 (each red on its named cell).
- The earlier contested window (manager dies between launch and the slot activation CAS) is closed by the reservation and proven by the spawn-window suite.

Two follow-ups from this round are now tracked so they stop living in review prose:

- #1511 — a deep TMPDIR pushes the custodian socket past the 108-byte sun_path limit; launches fail EINVAL surfaced as a misleading spawn error, and failed runs leak reparented custodians. Not a defect of this PR under sane TMPDIRs; asks for a legible socketPath length assertion and a skip-safe sandbox base for the suites.
- #1512 — the three deliberately-accepted residuals (membership-based group follow-up kill; same-uid rewrite of record.json retargeting a reap; stale custody reference when recordSlotCustody/recordSlotRuntime fails) enumerated for an explicit wontfix-or-fix-lane decision.

Issue #1396, opened 2026-09-09
fix(ci): pr-head-gate silently omits pull_request declared via YAML merge keys


davidfarah2003:
## Summary

`scripts/pr-head-gate.mjs` on `origin/main` `60f4ab82a977e119babbdb09f42f05719111538e` silently omits a valid pull-request workflow whose `on` mapping is composed with YAML merge keys (`<<: *anchor`). `parseDocument` is called **without** `merge: true`, so `<<` is left as a literal key. Zero errors and zero warnings. `Object.hasOwn(on, "pull_request")` is then false, `pullRequestDeclaration` returns `undefined`, and `expectedPullRequestWorkflows` drops the workflow. `classifyPullRequestHead` can return **green=true** for the reduced set.

This is the same failure class as #1104 (a safety gate that returns GREEN for an expected set it quietly reduced), through a route #1106 did not close. #1104's filed case — `on: [push, pull_request] # ordinary YAML trailing comment` — is fixed on main by #1106 (`602d9b1f3`, merge commit `fbf2884de3f1`). Do not treat this as a duplicate of #1104.

## Exact YAML

```yaml
name: Merge
defaults: &d
  pull_request:
    branches: [main]
on:
  <<: *d
  push:
```

## Parser options, quoted from the shipped file

`scripts/pr-head-gate.mjs` `pullRequestDeclaration`:

```js
const document = parseDocument(text, {
  logLevel: "error",
  strict: true,
  stringKeys: true,
  uniqueKeys: true,
});
```

There is no `merge: true`. The `yaml` package leaves `<<` unresolved unless that option is passed.

The fail-closed guard in the same function is:

```js
const problem = document.errors[0] ?? document.warnings[0];
if (problem) throw new Error(`${file}: invalid YAML: ${problem.message}`);
```

On this input that guard never fires: **errors=0, warnings=0**.

## Measured at origin/main `60f4ab82`

Same options as the shipped call:

- `on` after `toJS({ maxAliasCount: 100 })` = `{"<<":{"pull_request":{"branches":["main"]}},"push":null}`
- `Object.hasOwn(on, "pull_request")` = **false**
- `Object.keys(on)` = `["<<","push"]`
- `expectedPullRequestWorkflows({ "merge.yml": that source, "ci.yml": "name: CI\non: [pull_request]\n" }, ["package.json"])` = `["CI"]` — Merge omitted
[...]

PR #1438, opened 2026-09-10
fix(ci): resolve YAML merge keys in pr-head-gate

davidfarah2003:
## What

`scripts/pr-head-gate.mjs` parsed workflow YAML without the `yaml` package's `merge` option, so a
`pull_request` event composed through a YAML merge key (`<<: *anchor`) was left as a literal `<<`
key. `Object.hasOwn(on, "pull_request")` was then false, the workflow was dropped from the expected
set, and the head could be reported green against a set the gate had quietly shrunk. Parsing
produced zero errors and zero warnings, so the fail-closed guard never fired.

## Evidence

Two workflow documents that mean the same thing to GitHub, given the same runs (only CI ran):

| `on` spelling | expected set | missing | green |
| --- | --- | --- | --- |
| plain mapping | `["CI","Merge"]` | `["Merge"]` | false |
| via merge key (before) | `["CI"]` | `[]` | **true** |
| via merge key (after) | `["CI","Merge"]` | `["Merge"]` | false |

## Scope

`merge: true` on the parse options only. A merge-key `on` now behaves exactly as the equivalent
plain mapping does, in every case.

Deliberately unchanged: `pull_request` still accepts only `paths` and `paths-ignore`, so a
`branches` filter still throws. That throw is pre-existing — plain YAML carrying the same filter
throws identically before and after this change — and is not part of this defect.

Refs #1396

### DIFF
--- a/bin/smoke/fixtures/pr-head-gate-fetch.mjs
+++ b/bin/smoke/fixtures/pr-head-gate-fetch.mjs
@@ -1,6 +1,14 @@
 const sha = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
-const workflow = "name: Hidden\non: [push, pull_request] # ordinary YAML comment\n";
-const run = {
+const hidden = "name: Hidden\non: [push, pull_request] # ordinary YAML comment\n";
+const ci = "name: CI\non: [pull_request]\n";
+const mergeKey = `name: Merge
+defaults: &d
+  pull_request:
+on:
+  <<: *d
+  push:
+`;
+const hiddenRun = {
   id: 1,
   name: "Hidden",
   event: "pull_request",
@@ -10,6 +18,17 @@ const run = {
   created_at: "2026-08-30T00:00:00Z",
   pull_requests: [{ number: 1098 }],
 };
+const ciRun = {
+  id: 2,
+  name: "CI",
+  event: "pull_request",
+  head_sha: sha,
+  status: "completed",
+  conclusion: "success",
+  created_at: "2026-09-10T00:00:00Z",
+  pull_requests: [{ number: 1098 }],
+};
+const mergeKeyMode = process.env.PR_HEAD_GATE_FIXTURE === "merge-key";
 
 const json = (value) => new Response(JSON.stringify(value), {
   status: 200,
@@ -20,15 +39,29 @@ globalThis.fetch = async (input, init = {}) => {
   const url = new URL(String(input));
   const accept = new Headers(init.headers).get("accept") ?? "";
   if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098") return json({ head: { sha } });
-  if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098/files") return json([]);
+  if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098/files") {
+    return json(mergeKeyMode ? [{ filename: "package.json" }] : []);
+  }
   if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows") {
-    return json([{ type: "file", name: "hidden.yml", path: ".github/workflows/hidden.yml" }]);
+    return json(mergeKeyMode
+      ? [
+        { type: "file", name: "ci.yml", path: ".github/workflows/ci.yml" },
+        { type: "file", name: "merge.yml", path: ".github/workflows/merge.yml" },
+      ]
+      : [{ type: "file", name: "hidden.yml", path: ".github/workflows/hidden.yml" }]);
   }
   if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows/hidden.yml" && accept.includes("raw")) {
-    return new Response(workflow, { status: 200 });
+    return new Response(hidden, { status: 200 });
+  }
+  if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows/ci.yml" && accept.includes("raw")) {
+    return new Response(ci, { status: 200 });
+  }
+  if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows/merge.yml" && accept.includes("raw")) {
+    return new Response(mergeKey, { status: 200 });
   }
   if (url.pathname === "/repos/Cotal-AI/Cotal/actions/runs") {
-    return json({ workflow_runs: process.env.PR_HEAD_GATE_FIXTURE === "missing" ? [] : [run] });
+    if (mergeKeyMode) return json({ workflow_runs: [ciRun] });
+    return json({ workflow_runs: process.env.PR_HEAD_GATE_FIXTURE === "missing" ? [] : [hiddenRun] });
   }
   return new Response(`unexpected fixture request: ${url}`, { status: 500 });
 };
--- a/bin/smoke/mutations/pr-head-gate.json
+++ b/bin/smoke/mutations/pr-head-gate.json
@@ -9,8 +9,9 @@
   "proveWith": "pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json",
   "why": [
     "The guard and its smoke are repository tools rather than a published package boundary, so this",
-    "config grades them as a tool. The smoke imports scripts/pr-head-gate.mjs directly, and two cells",
-    "also execute the shipped pnpm pr-head-gate command against deterministic GitHub API responses."
+    "config grades them as a tool. The smoke imports scripts/pr-head-gate.mjs directly, and three cells",
+    "also execute the shipped pnpm pr-head-gate command against deterministic GitHub API responses,",
+    "including a merge-key pull_request declared via YAML <<."
   ],
   "mutations": [
     {
@@ -108,6 +109,14 @@
       "replace": "  const expected = [];",
       "expectRed": "the shipped pr-head-gate command reports the YAML-derived Hidden workflow green",
       "cell": "the shipped pr-head-gate command reports the YAML-derived Hidden workflow green"
+    },
+    {
+      "name": "YAML merge keys hide pull_request again",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "    uniqueKeys: true,\n    merge: true,",
+      "replace": "    uniqueKeys: true,",
+      "expectRed": "a merge-key on mapping includes pull_request in the expected set and is not green without that run",
+      "cell": "a merge-key on mapping includes pull_request in the expected set and is not green without that run"
     }
   ]
 }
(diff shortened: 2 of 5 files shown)


davidfarah2003:
Verdict: APPROVE
Head: d11018d226ce1eda8df94256a0bfec1c830e300e

Review and verification notes:

1. Test execution:
- `pnpm smoke:pr-head-gate` passes completely (38/38 cells passed).
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json` kills all 13 mutations cleanly.

2. Real entry-point reachability:
- `shippedCommand("merge-key")` in `bin/smoke/pr-head-gate.smoke.ts` invokes the real CLI binary through `pnpm pr-head-gate 1098` via `spawnSync`.
- Only `globalThis.fetch` is stubbed via Node `--import` to supply deterministic API responses. All production code paths in `scripts/pr-head-gate.mjs` (CLI argument validation, `live()`, repo resolution, workflow fetching, YAML document parsing with `merge: true`, `classifyPullRequestHead()`, and exit status handling) run end-to-end. This confirms genuine entry-point reachability.

3. First-red mutation alignment:
- Tested by locally reverting `merge: true` in `scripts/pr-head-gate.mjs`: the first assertion to fail in the smoke suite is `the shipped pr-head-gate command includes a merge-key pull_request workflow and is not green without its run` (failing at mark 35, before the individual unit assertions).
- This matches the updated `expectRed` / `cell` property in `bin/smoke/mutations/pr-head-gate.json`.

4. Scope / branches filter:
- Checked behavior against base commit `ffa1501748d59d45234c1208a3322fbf14ceadd9`. The check `unsupported = Object.keys(pullRequest).filter((key) => key !== "paths" && key !== "paths-ignore")` throwing `unsupported pull_request filter: branches` is identical and pre-existing at base for standard YAML without merge keys. It is not a regression.

5. False green assessment:
- YAML merge keys (`<<: *anchor`) are resolved during document parsing, correctly propagating `pull_request` declarations into the expected set instead of dropping them.
- Invalid merge targets (such as scalar anchors) fail closed by throwing on `document.toJS()`, exiting with code 2.
- No vectors found that would cause the gate to report GREEN inappropriately.

---

Superseded by 5617064245; do not count. Named d11018d226ce1eda8df94256a0bfec1c830e300e.

davidfarah2003:
APPROVE

Graded head: `d11018d226ce1eda8df94256a0bfec1c830e300e`

I found no correctness blocker in the issue scope.

- `pnpm smoke:pr-head-gate` passed all 38 cells.
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json` killed all 13 mutations. The `YAML merge keys hide pull_request again` mutation reddened the named shipped-command cell, `the shipped pr-head-gate command includes a merge-key pull_request workflow and is not green without its run`, at 35 marks versus the 38-mark baseline.
- The regression cell invokes `pnpm pr-head-gate 1098`, whose package script runs `node scripts/pr-head-gate.mjs`. The fixture replaces only `fetch`, so the real CLI path fetches the workflow text, parses it through the changed `pullRequestDeclaration`, derives `expected: CI, Merge`, reports `missing: Merge`, exits nonzero, and therefore proves entry-point reachability rather than only helper behavior.
- Running the exact base implementation from `ffa1501748d59d45234c1208a3322fbf14ceadd9` reproduced `unsupported pull_request filter: branches` for a plain mapping, while the merge-key form returned an empty expected set. The branches behavior is pre-existing, not introduced here.
- Direct parser controls produced zero errors and warnings in both modes. Without `merge: true`, `on` retained a literal `<<`; with it, `pull_request` was resolved. Invalid scalar merge sources throw, merged `branches` still fails closed, and I found no new path to an incorrect green verdict.

---

Superseded by 5617028831; do not count. Named d11018d226ce1eda8df94256a0bfec1c830e300e.

davidfarah2003:
BLOCK

Correction to my earlier review. Graded head: `d11018d226ce1eda8df94256a0bfec1c830e300e`.

The central premise is false: GitHub Actions supports YAML anchors and aliases, but it does **not** support YAML merge keys (`<<`). In the official `actions/runner` anchor-support issue, GitHub staff stated before release that merge keys would not be supported, then confirmed after rollout on 2025-08-04: “Are merge keys supported as well?” “no.” The current GitHub documentation likewise documents anchors and aliases, links the YAML 1.2.2 anchor specification, and does not document merge keys. Merge keys are a separate YAML 1.1 type absent from YAML 1.2.

Source: https://github.com/actions/runner/issues/1182#issuecomment-3151378337

That changes the safety analysis. Enabling `merge: true` makes this guard interpret a workflow more permissively than GitHub does. The new fixture proves only the guard's synthetic interpretation. It does not prove GitHub registers or triggers such a workflow. The guard will add the merged `pull_request` workflow to `expected`, but GitHub will reject or fail to register that workflow, so no matching run can exist and the gate remains permanently NOT GREEN. This is a false-red regression, not a fix for a valid GitHub workflow form.

The smoke and mutation proof still pass, and the named mutation reaches the shipped command, but they encode the unsupported premise. Please remove merge-key support and close or revise #1396 unless there is direct current GitHub evidence that merge keys became supported after GitHub's explicit statement and documentation.

davidfarah2003:
Verdict: BLOCK (superseding earlier approval)
Head: d11018d226ce1eda8df94256a0bfec1c830e300e

Reason for blocker:

Issue #1396 is founded on the claim that GitHub accepts merge keys in workflow `on` maps. However, GitHub Actions does not support YAML merge keys (`<<`).

Key evidence:
1. When YAML anchor and alias support was implemented and rolled out for GitHub Actions, GitHub staff specifically confirmed that merge keys are not supported because they are part of YAML 1.1 and absent from YAML 1.2:
https://github.com/actions/runner/issues/1182#issuecomment-3151378337

2. Standard YAML 1.2 anchors and aliases (`&` and `*`) already work without `merge: true` in the YAML parser.

3. By enabling `merge: true`, `pr-head-gate.mjs` parses `<<: *anchor` and expects `pull_request` to run for that workflow. But GitHub Actions evaluates workflow files under YAML 1.2 without merge key expansion, so GitHub never triggers that workflow on `pull_request`. As a consequence, `pr-head-gate` will classify the workflow as missing and report a permanent false red (`verdict: NOT GREEN`) on any PR using this syntax.

The gate must reflect GitHub Actions' actual triggering semantics, not a more permissive YAML 1.1 dialect. This change should not land.

---

Superseded by 5617077106; do not count. Named d11018d226ce1eda8df94256a0bfec1c830e300e.

davidfarah2003:
Verdict: BLOCK
Head: d11018d226ce1eda8df94256a0bfec1c830e300e

Re-grade findings:

1. Verification of GitHub Actions support:
- Verified first-hand via `gh api repos/actions/runner/issues/comments/3151378337`: GitHub staff (ericsciple) confirmed on 2025-08-04 that merge keys (`<<`) are not supported in GitHub Actions because they are YAML 1.1 only and excluded from YAML 1.2.
- A search of `actions/runner` issues and changes up to the present shows no subsequent addition of YAML merge key support. Official documentation lists YAML 1.2.2 anchors and aliases, with no mention of merge keys.

2. False-red regression:
- Enabling `merge: true` creates an active divergence between this gate and GitHub's actual runtime.
- If a workflow author uses `<<: *anchor` inside `on`, GitHub Actions does not expand the merge key and does not trigger `pull_request`.
- However, `pr-head-gate` with `merge: true` expands the key and expects a `pull_request` run. Because GitHub never creates that run, the gate permanently fails closed with `missing: <Workflow>` and `verdict: NOT GREEN`. This constitutes a permanent false-red regression.

3. Analysis of the fail-closed alternative:
- The existing codebase consistently fails closed on inputs it cannot evaluate or that represent unsupported forms (e.g., unsupported `pull_request` filters, invalid globs, empty event mappings).
- Throwing on a literal `<<` key in `on` would be consistent with this design: an unexpanded `<<` represents an unsupported YAML 1.1 merge key attempting to declare triggers that GitHub Actions will not parse as intended.
- Throwing on `<<` prevents silent misconfiguration and alerts authors to use supported YAML 1.2 constructs (such as mapping aliases `on: *anchor`), whereas `merge: true` erroneously asserts that GitHub Actions supports merge keys.

davidfarah2003:
## Blocked: the premise this change rests on is unverified, and if it is wrong the fix inverts the gate

An earlier approval of this head is superseded by this comment.

### The claim under the fix

Issue #1396 asserts that GitHub Actions accepts YAML **merge keys** (`<<:`) in a workflow's `on` map, so a workflow using one is registered and runs. The change resolves `<<:` and adds the resulting workflow to the gate's `expected` set.

That assertion was never tested against GitHub. It was reproduced against a local YAML parser, which establishes what *that* parser does and says nothing about GitHub's. A faithful reproduction of a mechanism is not evidence that the mechanism is reachable.

### What is actually known

GitHub shipped anchors (`&`) and aliases (`*`) for workflow YAML in September 2025. Merge keys are a separate feature — they live in the YAML type repository rather than YAML 1.2 core — and there is no published statement that they were included.

The workflow-syntax reference does not mention anchors, aliases, or merge keys **anywhere**. That silence is not confirmation in either direction, and this comment does not claim to resolve it.

There is a specific trap for anyone re-checking: a widely-linked runner issue thread reads as though merge keys shipped. It records the request, not the outcome.

### Why the block does not depend on resolving that

|                                  | GitHub **rejects** `<<:`                      | GitHub **accepts** `<<:` |
| -------------------------------- | --------------------------------------------- | ------------------------ |
| **This change** (resolve, expect) | the expected set names a check that never runs — **permanent false red, the gate can never go green** | correct |
| **Fail closed** (throw on `<<`)   | correct                                       | blocks until a person looks — safe |

Fail-closed is correct in both columns. This change is correct in one and wrong in the severe direction in the other, so it should not merge while the premise is open.

Before the change the gate left `<<` literal and omitted that workflow from `expected`. If GitHub never runs it, that omission **matched reality**.

### The consistent fix

The guard already fails closed on invalid workflow YAML, an empty `on` sequence, a `pull_request` filter it cannot evaluate, and unsupported glob syntax. A literal `<<` in `on` is the same thing: something the guard cannot evaluate. Throwing on it is a member of a class this file already has, not a new policy — and it is safe whichever way the premise resolves.

That is a different change with a different premise, so it belongs on a new head with a fresh review rather than as an amendment here.

### Disposition

This PR stays open and unmerged. The engineering is sound: the mechanism reproduces, the suite grows 35 to 38 cells, all 13 mutations are killed, and the fix-specific mutation's first red drives the shipped command rather than an extracted helper. None of that is in question. What failed is that the issue's claim about GitHub was inherited unchecked by everyone who touched this, including its reviewers and the `triage:confirmed` label.

davidfarah2003:
Superseded by #1442 and closed unmerged.

This PR resolved the YAML merge key so the declared workflow joined the expected set. That premise
does not hold: GitHub Actions does not support merge keys, so such a workflow is never registered
and never runs, and expecting it would leave the head permanently not-green. #1442 takes the
opposite and correct direction — refuse a merge key in the `on` mapping rather than resolve it,
which is right whether or not the platform ever accepts the construct. The review comments and the
diff here stay as the record of why the first approach was wrong.

PR #1442, opened 2026-09-10
fix(ci): fail closed on YAML merge keys in pr-head-gate

davidfarah2003:
## What

`scripts/pr-head-gate.mjs` now refuses a workflow whose top-level `on` mapping carries a literal
YAML merge key (`<<`), instead of silently dropping the event it could not evaluate.

GitHub Actions does not support YAML merge keys. Anchors (`&`) and aliases (`*`) shipped in
September 2025; merge keys are a YAML 1.1 type absent from YAML 1.2 and were deliberately excluded,
confirmed by a repository collaborator in `actions/runner#1182` on 2025-08-04. A workflow using one
is therefore never registered and never runs.

So the guard must not resolve the merge key either — doing so would make it expect a workflow that
can never produce a run, leaving the head permanently not-green. The failure this closes is the
narrower and real one: the guard could silently shrink its expected event set on a workflow it
could not evaluate. A form the guard cannot judge is now refused rather than dropped.

This joins an existing class in the same file, which already fails closed on invalid workflow YAML,
an empty `on` sequence, an unevaluable `pull_request` filter, and unsupported glob syntax.

## The real delta — please grade this, not the headline

Only the **top-level** form is a new refusal. The **nested** form (`<<` inside an event config)
**already threw before this change**, via the unsupported-filter path:

| input | before | after |
| --- | --- | --- |
| `on: { <<: *events, push: }` | silently dropped the event | **throws** (new) |
| `on: { pull_request: { <<: *f } }` | `unsupported pull_request filter: <<` | `unsupported YAML merge key in the on mapping` (clearer message only) |

No new coverage is claimed for the nested form. It is a message improvement on an existing refusal.

## Regression controls

Anchors and aliases *are* supported by GitHub and must keep working. Two cells guard that:
`on: *events` still declares `pull_request`, and a schema-valid literal `on` mapping still does.

## Verification

`pnpm smoke:pr-head-gate` — 40 passed, 0 failed (35 before). `pnpm mutation-proof --config
bin/smoke/mutations/pr-head-gate.json` — 13/13 killed; the mutation disabling the throw reddens
`a merge key supplying the on event name fails closed` first, at 37 marks against a 40 baseline.
One cell drives the shipped `pnpm pr-head-gate` command so a real entry point reaches the change.

Test fixtures use a schema-valid anchor carrier. An earlier draft used a `defaults:` block holding
`pull_request:`, which is independently invalid workflow syntax and would have made the cells
unable to attribute their result to the merge key.

## Known limits

**Coverage of the replaced behaviour was audited in this repository only.** A search for cells
asserting the old silent-shrink behaviour returned zero, and the surface is bounded: only
`bin/smoke/pr-head-gate.smoke.ts` imports `expectedPullRequestWorkflows` or
`classifyPullRequestHead`, so no other in-repo suite can assert it. That bound does not extend
outside this repository — a consumer elsewhere that relies on the guard silently omitting a
workflow it cannot evaluate would not be visible here, and this has not been searched for. Stated
as a limit, not as a risk that has been retired.

The repository's own workflows were checked separately: none of the seven files under
`.github/workflows/` carries a YAML merge key, so this change refuses none of them. (The
`<<'EOF'` in `ci.yml` is a heredoc inside a `run:` block, not a merge key.)

Supersedes the approach in #1438, which resolved the merge key rather than refusing it.

Refs #1396

## CI note

`live` was red on this head at run 34491677562 (job 102919724124) and green on re-run (attempt 2, job 103020582894). The two attempts were verified to grade the same work: identical suite roster and the same 57-cell opencode cooperative-stop total, with the single failing cell as the entire differential between them.

The failure is recorded in comment 5624409231 and the re-run outcome in comment 5624538586. It is deliberately not characterised as a flake — one event, no identified mechanism, no off-branch instance. Nothing in this PR addresses that cell, and nothing here is claimed to.

## This change is an instance of its own subject

The gate exists to stop a workflow declaring a check that will never be minted. This branch is currently blocked by exactly that failure in the live repository: it conflicts with main, so no merge ref regenerates; the workflow that would mint a now-required check is therefore absent from every ref this head is graded against; and the check can never appear, so the gate can never be satisfied. A gate that expects a run which cannot exist does not fail loudly — it waits forever.

### DIFF
--- a/bin/smoke/fixtures/pr-head-gate-fetch.mjs
+++ b/bin/smoke/fixtures/pr-head-gate-fetch.mjs
@@ -1,6 +1,19 @@
 const sha = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
-const workflow = "name: Hidden\non: [push, pull_request] # ordinary YAML comment\n";
-const run = {
+const hidden = "name: Hidden\non: [push, pull_request] # ordinary YAML comment\n";
+const mergeKey = `name: Merge
+jobs:
+  j:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - anchor: &events { pull_request: null }
+    steps: [{run: echo hi}]
+on:
+  <<: *events
+  push:
+`;
+const hiddenRun = {
   id: 1,
   name: "Hidden",
   event: "pull_request",
@@ -10,6 +23,7 @@ const run = {
   created_at: "2026-08-30T00:00:00Z",
   pull_requests: [{ number: 1098 }],
 };
+const mergeKeyMode = process.env.PR_HEAD_GATE_FIXTURE === "merge-key";
 
 const json = (value) => new Response(JSON.stringify(value), {
   status: 200,
@@ -20,15 +34,23 @@ globalThis.fetch = async (input, init = {}) => {
   const url = new URL(String(input));
   const accept = new Headers(init.headers).get("accept") ?? "";
   if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098") return json({ head: { sha } });
-  if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098/files") return json([]);
+  if (url.pathname === "/repos/Cotal-AI/Cotal/pulls/1098/files") {
+    return json(mergeKeyMode ? [{ filename: "package.json" }] : []);
+  }
   if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows") {
-    return json([{ type: "file", name: "hidden.yml", path: ".github/workflows/hidden.yml" }]);
+    return json(mergeKeyMode
+      ? [{ type: "file", name: "merge.yml", path: ".github/workflows/merge.yml" }]
+      : [{ type: "file", name: "hidden.yml", path: ".github/workflows/hidden.yml" }]);
   }
   if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows/hidden.yml" && accept.includes("raw")) {
-    return new Response(workflow, { status: 200 });
+    return new Response(hidden, { status: 200 });
+  }
+  if (url.pathname === "/repos/Cotal-AI/Cotal/contents/.github/workflows/merge.yml" && accept.includes("raw")) {
+    return new Response(mergeKey, { status: 200 });
   }
   if (url.pathname === "/repos/Cotal-AI/Cotal/actions/runs") {
-    return json({ workflow_runs: process.env.PR_HEAD_GATE_FIXTURE === "missing" ? [] : [run] });
+    if (mergeKeyMode) return json({ workflow_runs: [] });
+    return json({ workflow_runs: process.env.PR_HEAD_GATE_FIXTURE === "missing" ? [] : [hiddenRun] });
   }
   return new Response(`unexpected fixture request: ${url}`, { status: 500 });
 };
--- a/bin/smoke/mutations/pr-head-gate.json
+++ b/bin/smoke/mutations/pr-head-gate.json
@@ -132,6 +132,14 @@
       "replace": "    if (false) throw new Error(`${file}: unknown pull_request activity type: ${unknown.join(\", \")}`);",
       "expectRed": "an unknown pull_request activity type fails closed",
       "cell": "an unknown pull_request activity type fails closed"
+    },
+    {
+      "name": "YAML merge keys in on are silently ignored",
+      "file": "scripts/pr-head-gate.mjs",
+      "find": "  if (hasMergeKey(on)) throw new Error(`${file}: unsupported YAML merge key in the on mapping`);",
+      "replace": "  if (false) throw new Error(`${file}: unsupported YAML merge key in the on mapping`);",
+      "expectRed": "a merge key supplying the on event name fails closed",
+      "cell": "a merge key supplying the on event name fails closed"
     }
   ]
 }
--- a/bin/smoke/pr-head-gate.smoke.ts
+++ b/bin/smoke/pr-head-gate.smoke.ts
@@ -61,6 +61,47 @@ check(
     "hidden.yml": "name: Hidden\non: [push, pull_request] # ordinary YAML comment\n",
   }, ["package.json"])) === '["Hidden"]',
 );
+check(
+  "YAML anchors without a merge key still declare pull_request",
+  JSON.stringify(expectedPullRequestWorkflows({
+    "alias.yml": "name: Alias\njobs:\n  j:\n    runs-on: ubuntu-latest\n    strategy:\n      matrix:\n        include:\n          - anchor: &events { pull_request: null, push: null }\n    steps: [{run: echo hi}]\non: *events\n",
+  }, ["package.json"])) === '["Alias"]',
+);
+check(
+  "a schema-valid on mapping without a merge key still declares pull_request",
+  JSON.stringify(expectedPullRequestWorkflows({
+    "literal.yml": "name: Literal\njobs:\n  j:\n    runs-on: ubuntu-latest\n    steps: [{run: echo hi}]\non:\n  pull_request:\n  push:\n",
+  }, ["package.json"])) === '["Literal"]',
+);
+for (const [label, source] of [
+  [
+    "a merge key supplying the on event name fails closed",
+    "name: Merge\njobs:\n  j:\n    runs-on: ubuntu-latest\n    strategy:\n      matrix:\n        include:\n          - anchor: &events { pull_request: null }\n    steps: [{run: echo hi}]\non:\n  <<: *events\n  push:\n",
+  ],
+  [
+    "a merge key nested inside an on event config fails closed",
+    "name: Nested\njobs:\n  j:\n    runs-on: ubuntu-latest\n    strategy:\n      matrix:\n        include:\n          - anchor: &paths { paths: [\"src/**\"] }\n    steps: [{run: echo hi}]\non:\n  pull_request:\n    branches: [main]\n    <<: *paths\n",
+  ],
+] as const) {
+  let refused = false;
+  try { expectedPullRequestWorkflows({ "merge.yml": source }, ["package.json"]); }
+  catch (error) { refused = /unsupported YAML merge key in the on mapping/.test(String(error)); }
+  check(label, refused);
+}
+{
+  let refused = false;
+  let message = "";
+  try {
+    expectedPullRequestWorkflows({
+      "both.yml": "name: Both\njobs:\n  j:\n    runs-on: ubuntu-latest\n    strategy:\n      matrix:\n        include:\n          - anchor: &events { pull_request: { types: [opened] } }\n    steps: [{run: echo hi}]\non:\n  <<: *events\n  push:\n",
+    }, ["package.json"]);
+  } catch (error) {
+    message = String(error);
+    refused = /unsupported YAML merge key in the on mapping/.test(message) &&
+      !/types must keep opened and synchronize/.test(message);
+  }
+  check("a merge key and a types filter that drops synchronize report only the merge-key refusal", refused, message);
+}
 let invalidYamlRefused = false;
 try {
   expectedPullRequestWorkflows({
@@ -221,7 +262,7 @@ for (const conclusion of ["neutral", "skipped"]) {
   check(`a ${conclusion} expected workflow is failing, never green`, JSON.stringify(verdict.failing) === '["CI"]' && !verdict.green, verdict);
 }
 
-function shippedCommand(mode: "success" | "missing") {
+function shippedCommand(mode: "success" | "missing" | "merge-key") {
   // The shipped gate reads GitHub credentials (GH_TOKEN/GITHUB_TOKEN/GITHUB_REPOSITORY) and no
   // COTAL_ name at all, so an ambient copy would hand a live credential and broker URL to a child
   // that has no use for either. Strip the prefix.
@@ -251,8 +292,14 @@ check(
   shippedMissing.status === 1 && shippedMissing.stdout.includes("missing: Hidden") && shippedMissing.stdout.includes("verdict: NOT GREEN"),
   `${shippedMissing.stdout}${shippedMissing.stderr}`,
 );
+const shippedMergeKey = shippedCommand("merge-key");
+check(
+  "the shipped pr-head-gate command refuses a YAML merge key in on instead of shrinking the expected set",
+  shippedMergeKey.status === 2 && /unsupported YAML merge key in the on mapping/.test(`${shippedMergeKey.stdout}${shippedMergeKey.stderr}`),
+  `${shippedMergeKey.stdout}${shippedMergeKey.stderr}`,
+);
 
-const EXPECTED = 39;
+const EXPECTED = 45;
 check(`every cell ran (${EXPECTED} before sentinel)`, passed + failed === EXPECTED);
 console.log(`PR HEAD GATE SMOKE ${failed === 0 ? "OK" : "FAILED"} (${passed} passed, ${failed} failed)`);
 console.log("SUITE COMPLETE");
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -14,9 +14,10 @@ pnpm pr-head-gate <pull-request-number>
 
 It parses `.github/workflows/*.yml` at the exact head with the repository's YAML parser and derives
 the expected named set, including `pull_request` path filters. Valid YAML forms such as flow event
-lists with trailing comments are handled by the parser rather than a line reader. Invalid YAML and
-invalid path-filter values stop the guard. Empty trigger collections also stop it. A non-empty
-scalar or mapping that names no `pull_request` event is explicitly treated as a non-PR workflow.
+lists with trailing comments are handled by the parser rather than a line reader. Invalid YAML,
+invalid path-filter values, and a YAML merge key (`<<`) in `on` stop the guard. Empty trigger
+collections also stop it. A non-empty scalar or mapping that names no `pull_request` event is
+explicitly treated as a non-PR workflow.
 It then grades only runs attached to that pull request and head. Its non-green categories are
 distinct:
 
--- a/scripts/pr-head-gate.mjs
+++ b/scripts/pr-head-gate.mjs
@@ -27,6 +27,13 @@ const PULL_REQUEST_ACTIVITY_TYPES = new Set([
   "milestoned", "demilestoned",
 ]);
 
+function hasMergeKey(value) {
+  if (Array.isArray(value)) return value.some(hasMergeKey);
+  if (!plainObject(value)) return false;
+  if (Object.hasOwn(value, "<<")) return true;
+  return Object.values(value).some(hasMergeKey);
+}
+
 function pullRequestConfig(file, on) {
   if (typeof on === "string") {
     if (on.length === 0) throw new Error(`${file}: top-level on event must not be empty`);
@@ -38,6 +45,9 @@ function pullRequestConfig(file, on) {
     return on.includes("pull_request") ? null : undefined;
   }
   if (!plainObject(on)) throw new Error(`${file}: top-level on declaration must name one or more events`);
+  // GitHub never expands <<, so the rest of on is unevaluable. Refuse the merge key
+  // first and skip later checks such as types, rather than reporting both.
+  if (hasMergeKey(on)) throw new Error(`${file}: unsupported YAML merge key in the on mapping`);
   const events = Object.keys(on);
   if (events.length === 0) throw new Error(`${file}: top-level on mapping must not be empty`);
   return Object.hasOwn(on, "pull_request") ? on.pull_request : undefined;


davidfarah2003:
Reviewed commit `3222f372bc1a91140244aad4ec04a1d480f87729`.

### Test execution & verification
- `pnpm smoke:pr-head-gate`: 40/40 passed (39 cells + sentinel).
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json`: 13/13 killed (disabling the `hasMergeKey` throw reddens `a merge key supplying the on event name fails closed` at 37 marks against 40 baseline).
- `pnpm typecheck`: clean across all 28 workspace packages.

### Analysis & grading against review criteria

1. **Regression controls & `hasMergeKey` bounds**:
   - Anchors and aliases without merge keys (`on: *events`, `paths: *paths`) deserialize directly into native JS structures under YAML 1.2 without introducing a `<<` property. They continue to register and evaluate expected `pull_request` triggers correctly. Validated against `actionlint` 1.7.12, which cleanly accepts `on: *events`.
   - `hasMergeKey` checks `plainObject` and tests `Object.hasOwn(value, "<<")`. Scalar string values (e.g. `paths: ["<<"]` or `branches: ["<<"]`) are non-objects and do not trigger it. Because GitHub Actions does not support YAML merge keys anywhere in `on`, `hasMergeKey` does not reject any valid GitHub Actions workflow syntax.

2. **The real delta at base `ffa1501748d59d45234c1208a3322fbf14ceadd9`**:
   - Verified. At base, nested `<<` inside `pull_request:` (`on: pull_request: <<: *paths`) was already rejected by line 65's unsupported filter check (`unsupported pull_request filter: <<`) because `<<` is neither `paths` nor `paths-ignore`. The new error (`unsupported YAML merge key in the on mapping`) is clearer diagnostics for an already-failing case.
   - The top-level form (`on: <<: *events, push:`) was silently omitted from the expected workflow set at base because `Object.hasOwn(on, "pull_request")` returned false. Rejecting top-level `<<` is indeed the sole new refusal. The PR description accurately reflects the delta without overclaiming.

3. **Failure mode (throw vs ignore)**:
   - Throwing (fail closed) is the right failure mode. Because GitHub Actions does not support YAML merge keys (YAML 1.1 feature absent in YAML 1.2, confirmed in `actions/runner#1182` comment 3151378337), GitHub never registers or triggers workflows with merge keys in `on`.
   - Ignoring such workflows would cause the gate to silently shrink the expected workflow set, allowing PRs to merge with unexecuted protection workflows (fail-open security risk).
   - Resolving merge keys (`merge: true`, as attempted in #1438) causes permanent phantom missing runs because GitHub never runs them.
   - Halting with an actionable error (`unsupported YAML merge key in the on mapping`) immediately alerts the author that GitHub Actions will not run the workflow as written.

4. **Gate accuracy**:
   - No scenarios found that would result in false greens or false refusals.

**Verdict: APPROVE**

Stale: this verdict graded 3222f372bc1a91140244aad4ec04a1d480f87729, which is no longer the head of this PR. Do not count it. The current head 2ccf7dafb8201370c4915af22ef604764f073bea was graded by a later panel.

davidfarah2003:
Superseded by 5624963900; do not count. Named 2ccf7dafb8201370c4915af22ef604764f073bea.

## Review: APPROVE

Reviewed exact head `3222f372bc1a91140244aad4ec04a1d480f87729`.

The failure mode is correct. GitHub Actions does not register workflows using YAML merge keys, so resolving `<<` would create an expected run that can never exist. Silently ignoring the construct is also unsafe for a gate because it lets the guard reduce the expected set without acknowledging that it could not evaluate the declaration. Exiting with an evaluation error keeps this distinct from an ordinary not-green head and requires the invalid workflow or unsupported declaration to be corrected.

First-hand checks:

- `pnpm smoke:pr-head-gate`: 40/40 passed.
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json`: all 13 mutations killed, including the merge-key throw at its named assertion.
- The shipped command fixture exits 2 with `unsupported YAML merge key in the on mapping`, rather than returning green or waiting forever for an unregistered workflow.
- At base `ffa1501748d59d45234c1208a3322fbf14ceadd9`, the top-level form returns an empty expected set, while the nested form already throws `unsupported pull_request filter: <<` (or `branches, <<` in the test's combined form). The PR's description of the real delta is accurate.
- Anchor and alias controls remain accepted. I checked `on: *events`, aliases used as `pull_request` filter mappings and path lists, and an ordinary literal `on` mapping. All still declare the expected workflow.
- `actionlint` accepts the alias and literal controls and rejects both top-level and nested `<<` as unsupported by GitHub Actions.
- The recursive check is scoped to the parsed `on` value. It does not reject anchors or aliases merely because they are used there, and a string value containing `<<` is not mistaken for a key.

I found no path that newly returns green when it should not or rejects a supported trigger form.

davidfarah2003:
**Base refresh — not a review verdict, and not to be counted as one.**

This branch was rebased onto current `main` so the merge ref would be recomputed. The previous CI
run was created against a base that predated the repair landing on `main`, so it replayed a
pre-repair tree and failed on three suites this branch does not touch
(`required-arg-seam`, `manager-service-ops`, `agui-emitter`). Re-running alone could not clear that;
only moving the base could.

The head moved from `3222f372bc1a91140244aad4ec04a1d480f87729` to
`2780afb9c4e1aa1bedc745c25e4fea25208c6292`. The two reviews on this PR were given against the
former. They carry, because the rebase provably changed nothing but the parent:

```
rebased tree                                  02907ebae6bc229b60a497748b8c569b1ca357dc
git merge-tree --write-tree <main> <old head> 02907ebae6bc229b60a497748b8c569b1ca357dc   identical

diff(old head..new head)   sha256 8a7a89f2a53b0a6d…   377034 bytes
diff(old main..new main)   sha256 8a7a89f2a53b0a6d…   377034 bytes   identical
```

A third, complementary identity: this branch's own delta against its base is byte-identical before
and after (`9551` bytes, sha256 `e0fd9a55e16cbbbb…`), across the same five files. The change that
was reviewed is the change that remains.

Re-verified at the new head: `pnpm smoke:pr-head-gate` 40 passed / 0 failed;
`pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json` 13/13 killed, with the mutation
disabling the guard reddening `a merge key supplying the on event name fails closed` at 37 marks
against a 40 baseline.

CI at this head runs on different runner hardware than the earlier run. Nothing in this suite is
deadline-based — it parses YAML documents and compares expected sets — so its history transfers.
A green here is not being treated as a repair of the earlier red; the earlier red was an old base.

davidfarah2003:
## CI: `live` red on one cell — recorded before re-run

Run 34491677562, job 102919724124 (`live`), head 2780afb9c4e1aa1bedc745c25e4fea25208c6292: completed/failure. The log is 138885 bytes, so this is a real log rather than an empty-blob read that would grep clean.

One cell failed, in the opencode cooperative-stop suite:

```
✗ FAIL: a swap still QUEUED at the stop was allowed to run before exit
  markerQueued: '/tmp/cotal-smoke-broker-15187-vFGv6m/coop-queued-swap-ran'
```

Terminal line: `OPENCODE COOPERATIVE-STOP SMOKE FAILED (56 passed, 1 failed)`.

The assertion at `extensions/connector-opencode/smoke/cooperative-stop.smoke.ts:426` is `check(name, existsSync(markerQueued))`, which passes when the marker **exists**. The failure therefore means the swap queued behind the in-flight one was **dropped at exit rather than drained** — the inverse of how the cell name reads on its face. Every neighbouring cell in that suite passed, including the three that grade the same teardown ordering.

### Controls

- The cell runs and passes on main: run 34516244286 at a2e5f15bb9da, `live` job 103002259821 — one occurrence, one pass, zero failures, at the same 57-cell total (57 passed, 0 failed) and the same suite roster (26 / 15 / 31 / 30 / env-boundary / 57).
- `live` was green on this branch's base ffa1501748d5, job 102748216955.
- No off-branch instance of this cell failing was found.

### Reach

This PR changes five files, all of them the pr-head gate: `scripts/pr-head-gate.mjs`, `bin/smoke/pr-head-gate.smoke.ts`, `bin/smoke/fixtures/pr-head-gate-fetch.mjs`, `bin/smoke/mutations/pr-head-gate.json`, `scripts/README.md`. There is no intersection with the opencode connector. An empty file-set intersection is not on its own an all-clear, since `pr-head-gate.mjs` is selection-adjacent — it decides which head gets fetched and tested — but it cannot change the runtime ordering of a shutdown drain.

### Discarded exhibit

The OpenCode HTTP 500 on `/session/ses_coop/prompt_async` at 14:56:21 is 13 seconds *after* the failing cell, and appears exactly once in main's fully green run as well. It is suite-designed traffic and cannot discriminate. Noting it so it is not rediscovered and mistaken for a cause.

### Residual — a passing re-run does not resolve this

One failure event, with no identified mechanism and no off-branch instance, is an unexplained single event. It is deliberately not being called a flake. A re-run can clear the gate, but it cannot answer an attribution question that the single failure was never able to answer, and further passes move the arithmetic away from significance rather than toward it.

The residual stands on its own terms, separate from any merge rationale: the named cell above, the marker path above, run 34491677562 / job 102919724124, timing-sensitive shutdown-drain ordering, cause unidentified.

This is being recorded before the re-run is dispatched, so the order is on the record.

davidfarah2003:
## CI: `live` re-run is green — comparison verified, residual unchanged

Attempt 2 of run 34491677562, `live` job 103020582894: success. Log 139481 bytes, so a real log and not an empty-blob read.

Like-for-like check first, so the pass is known to grade the same work as the failure rather than less:

| suite | attempt 1 | attempt 2 |
| --- | --- | --- |
| up-stack live | 26 checks | 26 checks |
| up-manifest live | 15 checks | 15 checks |
| dogfood live | 31 checks | 31 checks |
| lifecycle e2e | 30 passed, 0 failed | 30 passed, 0 failed |
| env-boundary | OK | OK |
| opencode cooperative-stop | 56 passed, 1 failed | 57 passed, 0 failed |

Same roster, same 57-cell opencode total. The cell in question appears exactly once in each attempt and is the entire differential between them: failing in attempt 1, passing in attempt 2.

One measurement worth recording: the log contains four other occurrences of the failure glyph in **both** attempts. All four sit inside the names of passing assertions that check that a message "leads with ✗", so they are assertion text rather than results. A naive grep for the glyph reports four failures in a fully green run.

### The residual is not changed by this

The re-run clears the gate. It does not explain the failure. One event, no identified mechanism, no off-branch instance, now with one pass beside it, is still an unexplained single event — and it is deliberately not being called a flake. Further passing runs cannot resolve the attribution question, because the single failure was never able to answer it either; adding them moves the arithmetic away from significance rather than toward it.

Standing residual, recorded separately from any merge rationale:

- cell: `a swap still QUEUED at the stop was allowed to run before exit`
- marker: `/tmp/cotal-smoke-broker-15187-vFGv6m/coop-queued-swap-ran`
- first observed: run 34491677562, job 102919724124 (attempt 1)
- character: timing-sensitive shutdown-drain ordering in the opencode connector
- cause: unidentified
- nothing in this PR addresses it, and nothing in this PR is claimed to

Refs #1396.

davidfarah2003:
Grading the selection side of `scripts/pr-head-gate.mjs` at 2780afb9c4e1aa1bedc745c25e4fea25208c6292 — specifically whether a malformed or merge-keyed workflow document can get a head GRADED when it should be REFUSED, and whether the guard genuinely fails shut on what it cannot evaluate.

**What I ran** (all against the exported `expectedPullRequestWorkflows`, yaml 2.9.0 as pinned by the lockfile):

- Merge keys under `on` in every form I could construct: aliased (`<<: *base`), inline (`<<: {…}`), multi-merge (`<<: [*a, *b]`), as the only key of `on`, nested under `pull_request`, nested under a sibling event's filters, and hidden inside an aliased mapping that `on:` points at. All refused with `unsupported YAML merge key in the on mapping`.
- The mechanism, verified directly: with the parser options used here (`stringKeys`, `uniqueKeys`, `toJS({ maxAliasCount: 100 })`), `toJS` keeps `<<` as a literal own key rather than folding the merge away, so the deep `hasMergeKey` walk over `on` sees every merge key at any depth, including inside sequence elements.
- The masking scenario the guard exists for: a merge key that would hide `pull_request` behind a `push` override is refused before selection ever happens.
- Documents the parser rejects outright (tab indentation, duplicate keys, multi-document streams): refused via the `errors[0] ?? warnings[0]` check.
- Alias/anchor forms without merge semantics (`on: *alias`, aliased event names in a sequence, alias for the `pull_request` value, anchored scalars in `paths`, anchored `name`): all graded, matching what they unroll to. A 150-alias expansion is refused by `maxAliasCount`.
- Empty and degenerate `on` forms (`{}`, null, `[]`, empty string, `on: true` boolean): all refused.
- `true:` as the trigger key (the YAML 1.1 quirk where an unquoted `on` parses as boolean): refused as `missing top-level on declaration`. I could not confirm whether GitHub itself honors `true:` as a trigger key; if it ever does, this lands on the safe side either way, since the gate refuses rather than guessing.
- Scope of the refusal: merge keys outside `on` (the common `<<: *defaults` inside `jobs`) do not block a legitimate `pull_request` workflow, and one merge-keyed file in a set with benign files still refuses the whole set.
- The repo's seven real workflow files at this sha, under four changed-path scenarios: selections come out as expected (a docs-only change picks up Docs; core changes do not).

**Verdict: APPROVE.** Every malformed, merge-keyed, or otherwise unevaluable document I could build lands on the refused side. `pullRequestConfig` returns `undefined` only for well-formed documents that genuinely declare no `pull_request` trigger; everything it cannot evaluate throws instead. I found no shape that gets graded when it should be refused, and none refused that should be graded.

**One hardening note, not a blocker**: the guard's correctness rests on `toJS` leaving `<<` as a literal key, which is current yaml 2.9 behavior. If a future yaml starts folding merge keys during `toJS`, the guard would stop firing and the fail-closed property would regress silently. A small regression test that feeds `<<: *base` through `expectedPullRequestWorkflows` and asserts the refusal would pin that invariant against dependency drift.

Not checked (out of my lane): the shell path, fixtures, smoke suite, and the live GitHub API code path. Refs #1396.

---

Superseded by 5625022666; do not count. Named 2ccf7dafb8201370c4915af22ef604764f073bea.

davidfarah2003:
## Review: APPROVE
Refs #1396
Reviewed exact head `2ccf7dafb8201370c4915af22ef604764f073bea`.

### 1. Test execution and verification (RAN first-hand)
In worktree `/home/cotal/.cotal/jcode-scratch/lane-1396-r4-gemini-2ccf7daf`:
- `pnpm smoke:pr-head-gate`: passed 46/46 cells.
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json`: all 16 mutations killed cleanly against baseline.
- `shippedCommand("merge-key")` executes the shipped CLI binary `pnpm pr-head-gate 1098` via `spawnSync` and exits 2 with `unsupported YAML merge key in the on mapping`, validating production CLI entry-point reachability.
- Verified controls: anchors and aliases (`&` and `*`) without merge keys continue to parse cleanly (`YAML anchors without a merge key still declare pull_request`); literal mappings continue to parse cleanly.

### 2. Composition with #1455 and short-circuit evaluation
This head composes the fail-closed merge key check with #1455 (activity types validation):
- When a document contains both a merge key in `on` and an invalid `types` configuration (e.g. dropping `synchronize`), the gate reports only the merge-key refusal: `unsupported YAML merge key in the on mapping`.
- Analysis of this choice:
  - Correctness: Correct. In YAML 1.2 / GitHub Actions, `<<` is not expanded. An author attempting to merge triggers into `on` has created a malformed trigger block where downstream event configuration cannot be evaluated faithfully.
  - Defect hiding: This does not hide a defect in a problematic way. Like a syntax or schema parse error, refusing an unexpandable container before evaluating nested semantic constraints is standard compiler/linter discipline. Once the merge key is removed/fixed, the types validation runs immediately.
  - Structural vs. incidental: The short-circuit is structural. `hasMergeKey(on)` is evaluated inside `pullRequestConfig` before `pullRequestConfig` returns the `pull_request` configuration to `pullRequestDeclaration`. `pullRequestDeclaration` cannot reach the `pullRequest.types` validation loop without a valid configuration object returned by `pullRequestConfig`. Therefore, this ordering is enforced by the function call boundary and data flow, not merely incidental hunk order.

### 3. Verdict
APPROVE. The fail-closed behavior on YAML merge keys in `on` is sound, maintains backward compatibility for valid YAML anchors/aliases, and composes cleanly with #1455.

davidfarah2003:
Full re-grade at 2ccf7dafb8201370c4915af22ef604764f073bea. Main moved six commits and #1455 landed a `types` filter rule in the same function, so the earlier narrow review does not carry; this one covers the selection logic, the shell path, the fixtures, the smoke suite, and the interaction between the two rules.

**What I ran:**

- `pnpm smoke:pr-head-gate` — 46 passed, 0 failed.
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json` — all 16 mutations KILLED, each red on its named assertion, including the three #1455 type cells and the merge-key cell.
- A re-run of the selection corpus from the earlier review against the new head: merge keys under `on` in every form (aliased, inline, multi-merge, only-key, nested under `pull_request`, nested under a sibling event, hidden inside an aliased mapping) still all refuse; alias/anchor shapes without merge semantics still grade; empty and degenerate `on` forms still refuse; the repo's real workflow corpus under four changed-path scenarios selects as expected.
- The both-violations documents: a declaration with a merge key whose own `pull_request` key carries `types: [opened]` reports only the merge-key refusal; the same types list without a merge key reports the types refusal; `types: [opened, synchronise]` reports the unknown-type refusal first.

**On the short-circuit choice** — a document violating both the merge-key rule and the types rule reports only the merge-key refusal. I graded this by experiment, not by reading:

- Is it correct? Yes. The comment's premise holds: GitHub's workflow parser does not implement YAML merge keys (its anchors feature explicitly never got merge support, and actionlint documents merge keys as unsupported there too). A workflow that relies on `<<` under `on` is broken on GitHub's side, so "the rest of `on` is unevaluable" is the right frame: there is no truthful types verdict to report for a mapping GitHub will never read as written.
- Does it hide a second real defect? A maintainer who fixes the merge key gets a fresh, precise refusal on the next run — `types must keep opened and synchronize` names the exact declared list. That is the normal one-fix-per-run rhythm this gate already uses everywhere else (parser errors before trigger extraction, unsupported filters before types). Nothing is silently swallowed; both defects eventually surface, in causality order.
- Structural or incidental? **Structural, and test-locked.** The merge-key check runs inside `pullRequestConfig` before `on.pull_request` is even extracted; the types check consumes that extraction's output later, in `pullRequestDeclaration`. Reversing the order is not a hunk swap — the checks live in different functions over different values. I verified this negatively: I built a valid reorder (merge check moved after the types block, semantics inverted, probes confirmed the types refusal now fires first) and the smoke suite went red on four cells, including the explicit precedence cell "a merge key and a types filter that drops synchronize report only the merge-key refusal". The precedence cannot be silently reversed by a later edit without a failing suite; it is a pinned property, not an artifact of hunk order. (Two of my three reorder attempts were syntactically invalid — the move is genuinely non-trivial to make.)

**Verdict: APPROVE.**

Read, not run: the fetch fixture (`pr-head-gate-fetch.mjs`) serves deterministic API responses for the shipped-command cells, including a merge-keyed workflow served to the real CLI via `--import`; the fixture JSON's four PR cases cover mint-control, cross-PR, and cross-head contamination. Gap I did not close: the live GitHub API path (real network, real repo state) remains exercised only through those fixtures, not against production; and I did not re-verify against GitHub directly that merge keys are unsupported — that rests on GitHub's published anchors documentation and actionlint's, which agree. Refs #1396.

davidfarah2003:
Marking my earlier comment 5624696542 superseded in place: do not count it. It graded only the selection question at the earlier head 2780afb9c4e1aa1bedc745c25e4fea25208c6292; main advanced and #1455 landed a `types` rule in the same function, so the full re-grade at 2ccf7dafb8201370c4915af22ef604764f073bea (comment 5625022666) supersedes it. Refs #1396.

davidfarah2003:
APPROVE

Reviewed exact head `2ccf7dafb8201370c4915af22ef604764f073bea` from the rebased composition.

### Ran

- `pnpm smoke:pr-head-gate`: 46/46 passed.
- `pnpm mutation-proof --config bin/smoke/mutations/pr-head-gate.json`: all 16 mutations killed on their named assertions.
- A focused mutation that replaced merge-first handling with code that interpreted the merge payload and diagnosed `types` first reddened `a merge key and a types filter that drops synchronize report only the merge-key refusal`. This directly protects the intended precedence against later refactoring.
- The shipped `pnpm pr-head-gate` fixture returns exit 2 with `unsupported YAML merge key in the on mapping`. Disabling the implementation throw reddens that shipped-command cell, so the public command reaches the changed code.
- Live `pnpm pr-head-gate 1442` reached GitHub at this exact head and derived `Attribution, CI, Mutation reproof, Windows`; it reported the currently pending CI workflow as NOT GREEN.
- `pnpm typecheck` and `pnpm check:docs-voice` passed.
- Direct combined cases produced:
  - load-bearing `<<` containing `pull_request.types: [opened]`: merge-key refusal only;
  - literal bad `pull_request.types` plus a nested or unrelated merge key elsewhere in `on`: merge-key refusal only;
  - the same bad `types` without `<<`: the opened/synchronize refusal;
  - alias-only good types: accepted.
- Current GitHub workflow validation rejects the combined document at `<<` only.

### Composition with the `types` rule

The merge-key-first choice is correct. For the load-bearing form, the bad `types` value is visible only by resolving syntax GitHub does not resolve. Diagnosing it would make the gate interpret an unevaluable declaration before rejecting that declaration. Returning the structural merge-key error first gives the maintainer the actionable root problem.

This can expose a second `types` refusal after the merge key is removed. It also intentionally reports only the merge error when a literal bad `pull_request.types` coexists with an unrelated merge elsewhere under `on`. That is acceptable fail-fast behavior rather than a grading error: the workflow is already unevaluable, the command exits 2, and no green or phantom missing-run verdict is produced. The next refusal is real but does not need to be aggregated into the first diagnostic.

The precedence is structural, not incidental hunk order. `pullRequestConfig` recursively rejects `<<` before returning any `pull_request` config to `pullRequestDeclaration`; the `types` rule cannot execute until that return occurs. The combined regression cell names this contract, and the focused ordering mutation proves the suite detects reversing it.

### Read

The implementation comment and PR description state the same policy: once `<<` appears anywhere below `on`, later trigger checks are deliberately skipped. The diff is limited to the recursive detector, fail-closed check, fixtures, mutation, and documentation. I found no path that turns an unevaluable workflow into GREEN or converts it into a permanent expected run.

Gap: the live PR contains no merge-key workflow, so live GitHub execution validates the ordinary command boundary but cannot exercise this refusal. The shipped fixture and mutation establish reachability for that branch.

Refs #1396

davidfarah2003:
## CI: smoke shard 0 red — recorded before re-run

Run 34525652647, job 103033781877, head 2ccf7dafb8201370c4915af22ef604764f073bea.

Failing suite, by name rather than by shard number (shard assignment is a function of the changed-file set, so the same shard number is not the same work on two heads):

```
MANAGER-STOP-REAPS-AGENTS SMOKE FAILED  (9 passed, 1 failed)
  scenario threw: re-registration could not revoke + verify-evict the
  superseded serve family; the gate is left frozen for reconciliation,
  no new spec published
```

### This reproduces on an unrelated branch

The same suite fails the same cell, with the same count and the same abort shape, on a branch that shares no files with this one:

| | run | job | head | time |
| this PR | 34525652647 | 103033781877 | 2ccf7dafb820 | 20:19Z |
| a different PR | 34517868024 | 103007741554 | 777f2d8a2afa | 19:01Z |

That other branch changes mutation-coverage tooling. This one changes `scripts/pr-head-gate.mjs`, its smoke, fixture and mutation files, and `scripts/README.md`. Neither goes near manager or mesh lifecycle. A failure reproducing on two unrelated branches 78 minutes apart is not a property of either diff.

### Control, and one measurement that would otherwise mislead

Main ran the same suite to completion: run 34521878289, job 103044226300, 15 passed 0 failed. So it is not standing red on main.

Both failing runs report 10 cells (9 passed + 1 failed) while the green main run reports 15. The throw **aborts** the suite, so the remaining cells never execute. The 10-vs-15 gap is a consequence of the failure, not evidence that a different work set was graded — a like-for-like check that did not account for it would wrongly discard a valid control.

`ci-ok (live outcome classified)` is red alongside this, and is not a second fault. Its own log prints `unit=success smoke=failure live=success seat-pack=success seat-linux-arm64=success`. It aggregates, so it is red because smoke is red. `live` itself passed at this head. One root cause, two red checks.

### Standing note

No mechanism has been identified. This is recorded before the re-run is dispatched so the red is on the record independently of what the re-run returns.

Refs #1396.

davidfarah2003:
## Record correction: a supersede pointer on an earlier comment

Comment 5617797966 carries the line `Superseded by 5624963900; do not count.`

The "do not count" half is correct: that verdict graded 3222f372bc1a91140244aad4ec04a1d480f87729, which is no longer the head of this pull request, and it should not be counted.

The successor pointer is wrong. 5624963900 is a separate verdict by a different reviewer at the current head — it is not a later revision of 5617797966, and the two comments do not share an author. The line was added on a mistaken instruction from me: I had matched comment ids to reviewers by the order they appeared rather than by asking each reviewer which comment was theirs. Two reviewers checked their own records and corrected me; a third had already complied before anyone checked.

This correction is posted as its own comment rather than by editing 5617797966, and the incorrect line is deliberately left in place, so the record shows both what was written and what was wrong with it.

### State of verdicts on this pull request

Current head: 2ccf7dafb8201370c4915af22ef604764f073bea

Live verdicts at that head — three, from three different reviewers:

- 5624963900
- 5625022666
- 5625082614

Earlier verdicts, all marked and none to be counted: 5617748305 (stale, graded 3222f372bc1a), 5617797966 (graded 3222f372bc1a — see above), 5624696542 (graded 2780afb9c4e1, superseded by 5625022666).

Anyone counting approvals here should find exactly three, all naming the current head.

Refs #1396.

davidfarah2003:
## smoke shard 0: re-run green, and the failure is now bounded rather than merely unexplained

The re-run of smoke shard 0 passed. Recording the measurement rather than the outcome, because a green on its own would not have been worth much.

**The acceptance rule was fixed before the re-run was dispatched:** a pass had to report 15 cells. A failing run of this suite reports 10 (9 passed + 1 failed) because the throw aborts it, so a "green" at 10 cells would have meant the suite ended early and the comparison was void.

```
MANAGER-STOP-REAPS-AGENTS SMOKE OK  (15 passed, 0 failed)
```

15 cells, matching the green control on main exactly. The suite emitted its own banner and its own count, so it demonstrably executed rather than being skipped — a suite that was never scheduled produces the same "no failure" as one that passed, and those are different facts. No failing assertion anywhere in the job.

`ci-ok (live outcome classified)` went green on its own, as expected: it is an aggregate that was red only because smoke was red. One root cause, two red checks, one fix.

### The failure is now bounded to a window

| | head | time | result |
| --- | --- | --- | --- |
| a different pull request | 777f2d8a2afa | 19:01Z | 9 passed, 1 failed |
| this pull request | 2ccf7dafb820 | 20:19Z | 9 passed, 1 failed |
| this pull request, re-run | 2ccf7dafb820 | 23:48Z | **15 passed, 0 failed** |
| that other branch, later head | 07d769c1f | later | **15 passed, 0 failed** |

Two failures 78 minutes apart on branches sharing no files, then two independent passes at full cell count. That moves the claim from "reproducible across branches" to **reproducible within a window, since closed**. This reading was registered before either pass existed, so it is not a reinterpretation of a convenient result.

### What is still not claimed

No mechanism has been identified, and this is not being called a flake. The leading hypothesis — unverified — is load dependence: the failing cell is a revoke-and-verify-evict path with a reconciliation gate, the host was between 2.2x and 2.4x its memory commit limit across the window, and a separate lane observed load-dependent cells flipping inside a single fixture the same night. That is a candidate, not a finding, and it is recorded as such.

### One methodological note worth keeping

The red here was on shard 0 and one of the greens was also on shard 0. That is coincidence, not comparison. Shard assignment is a function of the changed-file set, so the same shard number is not the same work on two heads — every comparison above was made by **suite name**. Matching on the number would have produced a like-for-like that was not one.

Refs #1396.

PR #1451, opened 2026-09-10
ci: give Mutation reproof and Docs the CI concurrency policy

davidfarah2003:
## Problem

`mutation-reproof.yml` and `docs.yml` have no `concurrency` block. Every push to a pull request therefore leaves the previous head's run queued: a 12-shard reproof matrix per head, plus a docs build. `ci.yml` and `windows.yml` already supersede on pull_request and queue on main.

Measured 2026-09-10 12:19Z with the org on the 20-concurrent-job plan: 20 jobs running, 151 queued, of which 24 belonged to heads their PRs no longer carried. Ubuntu jobs waited 9 minutes on average and up to 29 minutes to start over the preceding two hours.

## Change

Both workflows get the block `ci.yml` carries, verbatim in policy: group by workflow and ref, `cancel-in-progress` only on `pull_request`, `queue: max` on main so a merge keeps its own verdict instead of being evicted.

Empty changeset (CI-only).

## Not verified here

A workflow file change takes effect on the next run that reads it from the target ref, so this PR's own reproof run does not exercise the new block. The first PR pushed twice after merge is the proof: its first reproof run should read `cancelled` with the second one running.

### DIFF
--- a/.changeset/reproof-docs-concurrency.md
+++ b/.changeset/reproof-docs-concurrency.md
@@ -1,0 +1,6 @@
+---
+---
+
+Give the Mutation reproof and Docs workflows the same concurrency policy as CI: a new push to a pull
+request supersedes the run already going for it, a merge to main queues instead of evicting. Without
+a group every pushed head kept its own 12-shard reproof matrix queued against the org's 20-job cap.
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,6 +29,13 @@ on:
   release:
     types: [published]
 
+# A new push to a PR supersedes the preview build already running for it; main and release builds
+# queue behind each other, because a deploy result is the record for that commit.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  queue: ${{ github.event_name == 'pull_request' && 'single' || 'max' }}
+
 jobs:
   docs:
     runs-on: ubuntu-latest
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -8,6 +8,17 @@ on:
     - cron: '17 5 * * *'
   workflow_dispatch:
 
+# Same policy as ci.yml: a new push to a PR supersedes the reproof already running for that PR; a
+# merge to main queues behind the previous main run instead of evicting it. Without this block every
+# PR head kept its own 12-shard matrix queued, so a PR pushed three times held 36 shard jobs of the
+# org's 20-job concurrency cap, and on 2026-09-10 the queue reached 151 jobs with 24 of them for heads
+# their PRs no longer had.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  # `queue: max` with `cancel-in-progress: true` is a validation error, so this tracks the line above.
+  queue: ${{ github.event_name == 'pull_request' && 'single' || 'max' }}
+
 jobs:
   changed_shard:
     # At dfbb4550f, PR #1350 selected 32 fixtures and exhausted the 145-minute serial step.
--- a/bin/smoke/workflow-concurrency.smoke.ts
+++ b/bin/smoke/workflow-concurrency.smoke.ts
@@ -68,19 +68,15 @@ function check(name: string, cond: boolean, extra?: unknown): void {
  *  group was considered and deliberately not added here. */
 const EXPECTED_WORKFLOWS = 7;
 const EXPECTED_PUSH_TO_MAIN = 6;
-/** Of those, the ones that declare a concurrency group and therefore CAN evict. `docs.yml` pushes
- *  to main with no group at all, so its runs are independent and there is nothing to queue. That is
- *  a valid answer to this problem, not an omission, and counting it separately keeps the difference
- *  visible instead of letting a future removal of a group look like compliance.
- *
- *  `mutation-reproof.yml` is also groupless, on purpose: its scheduled full sweep runs daily (cron
- *  `17 5 * * *`) under a 360-minute cap, so a scheduled run cannot overlap the next one, and its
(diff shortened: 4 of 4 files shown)


davidfarah2003:
Workflow semantics review for PR #1451 at commit 48ef0f58613b6d9f3e5434d4e0b938bb9a20897a.

1. Required Checks
No check names are changed or deleted. The generated checks across Docs (docs) and Mutation reproof (changed shard 0..11/12, changed, full) remain identical in name and matrix structure to main. If branch protection is configured in the future, existing rules naming these jobs will continue to match without vacuous passes.

2. Concurrency Policy and Cancellations
Concurrency group keys added:
Docs: ${{ github.workflow }}-${{ github.ref }}
Mutation reproof: ${{ github.workflow }}-${{ github.ref }}
Ref isolation is maintained. Two distinct pull requests or refs cannot collide in the same concurrency group.
On pull_request events, cancel-in-progress is true. When a previous run is cancelled by a newer commit push, GitHub Actions marks the cancelled job and workflow run with conclusion 'cancelled' (not success, and not a recorded duration). To downstream consumers, branch protection, and commit status rollups, a cancelled check does not satisfy required pass conditions and blocks until the superseding run succeeds.

3. Runner Environment
No runner changes are introduced in this PR.

Residuals:
None.

Verdict: SHIP

workflow semantics lens

davidfarah2003:
Workflow semantics re-read for PR #1451 at commit 2a3a5da30de0a17bb812216fdc5f8d04191ea85b.

The commit delta against previously reviewed head 48ef0f586 modifies only bin/smoke/workflow-concurrency.smoke.ts (updating the EXPECTED_GROUPED sentinel from 4 to 6 and aligning test comments). Workflow definitions under .github/workflows/ remain identical byte-for-byte to 48ef0f586.

1. Required Checks
Check names, count, and matrix definitions remain unchanged. No checks are added or removed under this commit.

2. Concurrency Policy and Cancellations
Concurrency group definitions in docs.yml and mutation-reproof.yml remain intact. Cancellation semantics continue to mark superseded runs as cancelled without producing false passing states.

3. Runner Environment
No runner changes are present.

Residuals:
None.

Verdict: SHIP

workflow semantics lens

davidfarah2003:
Merged-tree re-read of this PR at 2a3a5da30de0a17bb812216fdc5f8d04191ea85b, superseding the head my earlier comment graded (48ef0f58613b6d9f3e5434d4e0b938bb9a20897a).

Delta since that verdict, read in full: exactly one commit, touching only bin/smoke/workflow-concurrency.smoke.ts. It moves EXPECTED_GROUPED from 4 to 6 and replaces the comment that justified 4 (docs.yml and mutation-reproof.yml being deliberately groupless) with the reason for 6 (both now carry the CI concurrency policy). Nothing under .github or .changeset changed: the workflow files and the changeset are byte-identical to the head I graded before.

Re-verified at 2a3a5da30:
- The workflow-concurrency smoke passes at this head's tree 23f993dddf5d081fbad66795f6c694b8064d1010: 39/39, including "exactly 6 of those declare a concurrency group", which is the cell that forced this bump.
- Merged onto current main 1c32566fc1b837d59779a9a9e918c5574f93f88b (where #1454 already landed) and combined with #1455 at 51a6e05b1679c75cad5024d0044a4e47b572df4e: merge tree 29e4cdee921b1445f59570970d239cb327db7891, then terminal tree 133e567390c086cc0a9ede25ec2bafba24a3b235, no conflicts in either step, and the same terminal tree in the reverse landing order.
- This head's tree is byte-identical to the previously pushed c7da24294fce1c76e8b0244539b8d7e2550a47a0 tree, so the merged-tree result from that measure carries over exactly.

Residual, cosmetic and non-blocking: in the merged tree with #1455, one stale sentence of the old groupless comment for mutation-reproof.yml survives directly above the new paragraph. Comments only; all assertions pass.

Verdict at 2a3a5da30de0a17bb812216fdc5f8d04191ea85b: SHIP. The verdict bound to 48ef0f586 holds unchanged at this head.

merged tree lens

davidfarah2003:
### Review complete and green at 2a3a5da30de0a17bb812216fdc5f8d04191ea85b. The merge is blocked by a credential scope, not by anything in this change.

Attempting the squash returns:

```
refusing to allow an OAuth App to create or update workflow
`.github/workflows/docs.yml` without `workflow` scope
```

The identity available for merging here carries `repo` but not `workflow`, and this PR updates two
files under `.github/workflows/`. That is a property of the credential, not of the diff, so no
amount of further review or CI moves it. Someone whose credential carries the `workflow` scope needs
to press merge.

**Recording the gate state so it does not have to be rebuilt later.**

| check | result |
| --- | --- |
| head | 2a3a5da30de0a17bb812216fdc5f8d04191ea85b, unchanged at the time of writing |
| mergeable | true / clean |
| checks | 34 success, 1 skipped, 0 failure, 0 pending |
| verdicts naming this exact head | 2, from two independent lenses, both SHIP |

**The counted sentinels were re-validated against current main rather than taken from the branch.**
Main carries 7 workflow files, of which 4 declare a concurrency group (changesets, ci, installer,
windows). This change adds a group to docs.yml and mutation-reproof.yml, giving 6. The suite pins
`EXPECTED_WORKFLOWS = 7` and `EXPECTED_GROUPED = 6`, so both match the tree that would actually
land.

**The base moved after the verdicts were posted, and the part that moved is disjoint from what they
graded.** The merged-tree verdict was measured against a base that has since gained two commits
touching 102 files. None of those files is under `.github/`, and all four workflow files that both
this PR and the base modify are byte-identical between the graded base and current main (verified by
blob, with a changed file as a negative control). So the composition that was graded is the
composition that would land. A verdict does not automatically survive its base moving, which is why
this was measured rather than assumed.

**One residual, cosmetic and non-blocking**, carried forward from the merged-tree read: in a tree
that also contains the attribution-gate change, one stale sentence of the old groupless comment for
mutation-reproof.yml survives above the new paragraph. Comments only, no assertion depends on it,
and that change is not currently landing.

This PR is the remedy for the queue congestion measured on 2026-09-10 (151 queued jobs, 24 of them
for heads their PRs no longer had), so it is worth landing ahead of the other CI changes rather than
after them.

Issue #1433, opened 2026-09-10
test(connector-core): nothing checks whether a replacement egress guard is weaker than its predecessor, and three successive heads shipped one


davidfarah2003:
Three successive heads of the #695 egress fence shipped a classifier **weaker than the one it
replaced**, and no CI check named any of them. Each was found by a human or a review seat running a
differential by hand; each fix closed the reported hole and left a neighbouring one in the same
function.

This is not a request to re-fix those heads — #1429 repairs the third and adds cells for its
specific class. It is about the missing instrument: **the repo has no way to ask whether a
replacement guard is weaker than its predecessor**, so the question gets asked only when someone
thinks to ask it.

## The three, verified from source at each sha

All three are ancestors of `bd7593f3a` on `fix/695-egress-fence-regression`.

| head | shape it let through | what the predecessor did |
| --- | --- | --- |
| `1698fe253` | `frozenBodyViolatesEgressPolicy` dropped the `parseAguiFrame(part)` call; `if (!Array.isArray(events)) continue;` | `ee73e33c1` threw, so the publish halted |
| `350b8eb7a` | three-way verdict, `!Array.isArray(events) → unreadable`, but no `length === 0`, so `events: []` loops zero times and returns `clean` | threw |
| `071233483` | adds `length === 0`, still never validates the envelope, so `protocol` / `threadId` / `runId` / `epoch` / `seq` defects return `clean` | `parseAguiFrame` validated all five and threw |

The third was measured, not reasoned: a differential calling both versions on the same inputs,
with a positive control (a valid frame both must publish) and a negative control (a
`TOOL_CALL_RESULT` frame both must refuse) asserted in the same run, returned **15 weaker rows**,
every one envelope-level. It also reproduces end-to-end — an envelope-broken frame frozen in a real
`EventWal` and recovered through `AguiEmitter.start()` puts bytes on the wire where the predecessor
halted with zero publishes.

## What CI said, measured per head

Correcting a claim that has been circulating, including in my own earlier comments: it is **not**
true that all three went green. Only one did.

| head | outcome |
| --- | --- |
| `1698fe253` | success 19, failure 3, cancelled 7 — red |
| `350b8eb7a` | success 26, failure 2 (`changed`, `ci-ok`), cancelled 1 (`smoke shard 1/4`) — red |
[...]

Issue #1426, opened 2026-09-10
test(connector-core): the agui egress policy has no live-reach coverage


davidfarah2003:
The egress policy that landed in #1425 is guarded by mutation cells that prove the suite
**depends** on the guarded code. Nothing yet proves a **live entry point reaches** it.

## What exists today

`packages/.../agui` mutation cells kill on `applyAguiEgressPolicy` and on the
`frozenBodyViolatesEgressPolicy` refusal in `attempt()`. The recovery bypass was reproduced
as a differential with synthetic bytes: at the fixed head `mapper_calls=0, publishes=0` with
halt reason `egress-policy`; with only the retry refusal disabled, `publishes=1`.

Both reviewers on that PR stated the limit unprompted, and it is the honest one: the suite
hand-builds mapper records. The connection from a real connector session to the emitter was
established by reading source and wiring — all three production holders (`claude .../mcp.ts`,
`codex .../host.ts`, `opencode .../plugin.ts`) reach `AguiEmitter.start()` — not by crossing a
real ACL boundary at runtime.

A killed mutation shows the test depends on that code. It does not show a real entry point
reaches it. Those are different properties and only the first is currently covered.

## What would close this

A cell that drives a real connector session against a broker with a genuinely narrow read
grant, reads tool output under it, and asserts the marker does not appear on
`events.<owner>.<actor>`. The point is to cross the ACL boundary for real rather than to
assert over a hand-built record.

## Two smaller gaps from the same change

- `control:a mapper-null record stays dropped` is exercised but ungraded.
- Per-connector mutation is not possible under a single centralized forbidden set, so the
  args mutation reddens all three args cells together and the result mutation all three
  result cells. That is a consequence of the chokepoint design rather than a missing cell,
  but it means no cell fails for one connector alone, and a per-connector regression would
  not be isolated by the current set.

## Not a regression

The shipped behaviour is correct as far as it has been measured. This is coverage work.

davidfarah2003:
### Triage disposition: confirmed, with narrowed current scope

Verified against a fresh fetch of `origin/main` at `54ce14e7bb9f468309f300df30f4f6317292c2c8`.

The title's broad statement is partly stale because current main now has real production-entry reach coverage for the Codex connector:

- `extensions/connector-codex/smoke/events-lifecycle.smoke.ts` runs a real host process against a real local broker, reads a real rollout through the WAL/emitter path, and uses a second endpoint as the events-channel subscriber.
[...]

PR #1453, opened 2026-09-10
test(connector-core): grade a replacement egress guard against its predecessor

davidfarah2003:
Control pair first: predecessor `ee73e33c15f1a147b435d44a6f4feb8fd38512bc` vs adapter-free weakening `1698fe25356b3160b4ccb77f5b3056182ac8b688` on this corpus is WEAKER 117, STRICTER 0. An instrument that has only ever printed 0 is not evidence; 117 across 8 axes is the sensitivity proof that licenses the next number.

Primary pair: that predecessor vs HEAD is WEAKER 0, STRICTER 0, both directions, 165 rows, 6 named divergences (publish mapping still agrees), 0 held. Main is neither weaker nor stricter than the predicate it replaced, on this corpus.

The merge evidence in #1429 covered 2 of 14 types. That evidence was weak. The conclusion it certified is correct on this corpus. A weakly-evidenced right answer is still right; finding the weak evidence still mattered, because the next one might not have been.

Forbidden set and the three ADOPTED axes (same-part-ordering, event-type-throw, representation-flip) follow #1429; the fixtures were rebuilt here.

Smoke CI checks out depth 1. The loader pins those 40-char SHAs, fetches `--depth=1` from origin when `git cat-file` misses, and fails as `missing-history:` if the object is still absent. A well-formed SHA that is not in the checkout, with fetch disabled, must throw rather than bind HEAD.

Refs #1433


davidfarah2003:
Reviewer verdict: ACCEPT, at head 8fd2dd8013180eaa8d475dd1be0604aa5fef0532 (base 54ce14e7bb9f468309f300df30f4f6317292c2c8).

## What I ran (not read)

- `pnpm build && pnpm smoke:egress-guard-differential` in a full checkout: 39 cells, 39 passed, 165 corpus rows, 6 divergent, 0 held. Control pair (ee73e33c vs 1698fe25) WEAKER 117 / STRICTER 0; primary pair (ee73e33c vs HEAD) WEAKER 0 / STRICTER 0. Matches the PR body.
- The same suite in a genuine depth-1 clone (CI shape: shallow main + `git fetch --no-tags --depth=1 origin <head>`), with both historical SHAs confirmed ABSENT via `git cat-file -e` before the run: green, 39/39, both arms bound. After the run both SHAs were present — the harness fetched them itself through the exact path it documents. Zero `.egress-guard-arm-*` files left behind.
- Mutation, hand-applied and reverted (tree restored byte-identical, suite green again): T-READONCE (snapshot `events` before parse) — suite exits 1, `primary predecessor vs HEAD: WEAKER count is zero` fails with weaker=90 across the accessor rows; T-PARSE (skip the validator, scan `part.events`) — suite exits 1, three cells red including the two same-part-ordering DIVERGENT pins. The kill is on the assertion the harness names, not an incidental early abort.
- Independent probe of the live classifier with my own Proxy (getOwnPropertyDescriptor reports the clean data property, `get` flips to TOOL_CALL_RESULT on read 3): answers `forbidden-kind`, i.e. the strict-read invariant holds on an object I built, not just corpus rows.
- `pnpm typecheck`, `pnpm smoke:gate-inventory`, `pnpm smoke:ci-fragments`, `pnpm smoke:ci-declarations`, `pnpm changeset status`: all green.

## What I read and verified statically

- The loader is a real git-object cross-ref: `git show <ref>:extensions/connector-core/src/agui.ts` per arm, imported from a temp file beside src; no copied production function anywhere in the diff. Both historical arms contain the claimed `frozenBodyViolatesEgressPolicy` role (ee73e33c parses via `parseAguiFrame`; 1698fe25 reads `part.events` directly with an `Array.isArray` skip — the adapter-free weakening, not an ancestor of main).
- Representation axis covers all three required shapes: own accessor (`ownFlip`), inherited accessor (`inheritedFlip` via prototype), and Proxy whose `getOwnPropertyDescriptor` reports a data property while `get` is stateful (`proxyFlip`). 102 representation-flip rows execute (visible in the control pair's WEAKER lines); the earlier certified-incomplete own-accessor-only repair could not pass this corpus.
- Controls reach the code: frames are built with `aguiFrame(...)` whose `kind` is the real `AGUI_FRAME_KIND = "ag-ui.frame"`, and the forbidden set is derived behaviorally from `applyAguiEgressPolicy` over `Object.values(AGUI_EVENT_TYPE)` (14 types) rather than hand-written. The suite pins that derivation in cells.
- Missing-history control is an `ok()` cell (not abort-before-cells): deadbeef… with `fetch: false` must throw `missing-history:` — verified green in both environments. The fetch path is exactly `git fetch --no-tags --depth=1 origin <sha>` gated on `cat-file` miss, with `missing-history:` on failure and a refusal to silently bind HEAD.
- The mutation why-block correction in `agui-tool-result.json` replaces the wrong rationale (exports resolve to dist) with the measured one (the command's build poisons dist while the mutant is applied; `../src/agui.js` is read from src under tsx). It is a correction, not a band-aid; the rebuilds stay.
- The header states what a zero does not prove (symmetric loss, named-reason disagreements pushed to DIVERGENT, corpus limitation, #1426 reachability not covered) in the style of `packages/lang/smoke/differential.smoke.ts`. Both-direction grading (WEAKER and STRICTER) is in `gradePair`, with per-axis weaker/stricter/n counts printed and an uncovered-axis cell.
- PR hygiene: body ends "Refs #1433" with no closing keyword; `closingIssuesReferences` is empty (GraphQL). The `ci-suites.d` fragment filename is sha256 of the suite name, matching `fragmentFileName`, and the registry gates accept it.

## What I did not check

- Whether a live entry point reaches the predicate at runtime (#1426, named as open in the header).
- The mutation-proof fixture end-to-end (`node scripts/mutation-proof.mjs`) — I did not run it; I hand-applied and reverted the two mutations above instead.
- Nothing in `.internal/` beyond what the tree shows; no live suites; no workflow edits.

Named head: 8fd2dd8013180eaa8d475dd1be0604aa5fef0532.

davidfarah2003:
Review verdict: ACCEPT, bound to head 8fd2dd8013180eaa8d475dd1be0604aa5fef0532 (base 54ce14e7bb9f468309f300df30f4f6317292c2c8). This verdict stands on the runs below, independently of the earlier ACCEPT on this head.

## What I ran (not read)

- `pnpm smoke:egress-guard-differential`: 39 cells, 39 passed, 0 failed; 165 corpus rows, 6 divergent, 0 held. Control pair (`ee73e33c` vs `1698fe253`) WEAKER 117 / STRICTER 0, with weaker rows landing on 8 different axes. Primary pair (`ee73e33c` vs HEAD) WEAKER 0 / STRICTER 0. The tree was left clean and no `.egress-guard-arm-*.ts` file survived the run.
- Three single-mutation proofs against this suite via `scripts/mutation-proof.mjs`, each KILLED on the assertion I named up front:
  - Re-opening the envelope class in `frozenBodyEgressVerdict` (scan `events` straight off the part, skip `parseAguiFrame`): red on `primary predecessor vs HEAD: WEAKER count is zero`.
  - The read-once snapshot weakening (snapshot `events` before the parse, scan the snapshot): red on the same cell, killed by the representation-flip rows.
  - Sabotage of the harness's own comparator (`w = false`): red on `adapter-free predecessor vs adapter-free weakening: WEAKER count is at least 1`. The control pair is load-bearing, so the sensitivity claim is not vacuous.
- `pnpm typecheck`: green. `pnpm changeset status`: valid.
- The shallow-history path, live: a fresh depth-1 clone of this head does not contain `ee73e33c` (`git cat-file` fatal), `git fetch --no-tags --depth=1 origin <sha>` brings it in, and `git show <sha>:<path>` then resolves the file. That is the CI shape.
- `gh pr checks` sampled on this head: Windows, mutation reproof, attribution, and live all green; the long CI matrix was still finishing when I looked.

## What I read, not ran

- The full harness, the head predicate, and both historical classifiers (`ee73e33c` boolean, `1698fe253` adapter-free boolean) via `git show`. The DIVERGENT pins match what those three bodies actually do: unreadable beats forbidden within one part, forbidden wins across parts, and the boolean predecessor throws where the three-way names it.
- The rewritten why-block in `smoke/mutations/agui-tool-result.json`. Its load-bearing claim checks out in source: `agui-tool-result.smoke.ts` imports `../src/agui.js` (src under tsx) while `@cotal-ai/core` resolves through the exports map to `dist`.
- The PR body: ends `Refs #1433`, no closing keyword, and `closingIssuesReferences` is empty (checked via GraphQL).

## Judgments

- Both directions, one corpus: yes. WEAKER (base withheld, head publishes) and STRICTER (base published, head withholds) are counted separately with caps, over the same 165 rows, against an explicit publish mapping that never folds the three-way verdict into a boolean.
- Controls reach the predicate: positive and negative controls run before and after the corpus, per arm, and check the ROLE answer (clean / forbidden-kind vs ALLOW / REFUSE), not just the publish bit. The comparator mutation is killed by the control pair's own assertion, which is the proof they are not vacuous.
- Representation axis: own accessor, inherited accessor, and a Proxy whose `getOwnPropertyDescriptor` reports a data property while its `get` trap is stateful are all in the corpus, as steady-state rows and as flip rows (flip 0 through 4 against both forbidden types and the whole allowed set, 102 rows). The repair that was certified complete and then found weaker lived on this axis; it is covered here and the read-once mutation kills on it.
- Real git-object history: the predecessor is a real commit loaded by git object (`cat-file -e`, `show`, temp arm file written next to src so relative imports resolve, removed in `finally`), not a copied or vendored function. `ee73e33c` is an ancestor of HEAD; `1698fe253` is not an ancestor of current main, which is what makes the control pair an honest positive. The missing-history control with fetch disabled passes, and the fetch fallback works on a live depth-1 clone.
- What a zero does not prove: stated in the file header, naming symmetric flattening, corpus limits, the WAL encoder this suite never runs, and entry-point reachability (#1426) as uncovered. The zeros here sit next to a live positive (117 weaker rows on the control pair in the same run), so they are not standing alone.
- Coverage self-checks: the corpus must span all 14 `AGUI_EVENT_TYPE` values computed from the source's own set, the forbidden set is derived from `applyAguiEgressPolicy` rather than hand-listed, and every named axis must have at least one corpus row. All green, and the summary prints per-axis counts so an uncovered axis is a number, not silence.

## Not checked

- I did not re-run the six-mutation `agui-tool-result` fixture locally; the CI mutation-reproof job on this head is green instead.
- I did not run the broader smoke matrix myself; CI covers it on this head.
- The #1429 lineage of the three ADOPTED axes is taken from the in-file credit, not independently audited.

davidfarah2003:
Update to my review above (same head 8fd2dd8013180eaa8d475dd1be0604aa5fef0532; verdict unchanged, this supersedes nothing): the three items I listed as "not checked" have since been closed directly.

- The six-mutation `agui-tool-result` fixture was run locally: all 6 KILLED, each red on its named cell (T-RETRY, T-ARGS, T-RESULT, T-LIVENESS, T-PARSE, T-READONCE).
- The registry audits for the new `ci-suites.d` declaration: `smoke:ci-fragments`, `smoke:ci-declarations`, `smoke:gate-inventory` all green.
- The #1429 lineage of the three ADOPTED axes was audited in the issue thread itself: same-part-ordering, event-type-throw, and representation-flip are named there, including the original inherited-accessor WEAKER finding, so the ADOPTED credit is accurate.
- Additionally: all nine connector-core agui-family smokes run locally green, and the CI matrix on this head settled at 27 pass / 1 skipping / none pending.

davidfarah2003:
Review verdict: ACCEPT, bound to head 8fd2dd8013180eaa8d475dd1be0604aa5fef0532.

Checked first-hand in isolated workspace:
1. Exact HEAD: verified at 8fd2dd8013180eaa8d475dd1be0604aa5fef0532.
2. Check-runs at this SHA: terminal state is green. Out of 29 check-runs, 27 succeeded, 1 skipped (full mutation reproof workflow step when changed shards run), and 1 cancelled attribution check (run 34550269637, started 01:20:27Z, cancelled 01:20:36Z) superseded by successful attribution check (run 34550280957, started 01:20:44Z, completed 01:21:01Z). No blocking red runs.
3. Differential harness implementation:
   - Historical predecessors loaded via git objects (`git cat-file` check followed by `git fetch --no-tags --depth=1 origin <sha>` on cat-file miss, extracting with `git show <sha>:<path>`), avoiding vendored or live copies.
   - Missing-history control cell (`MISSING_HISTORY_SHA` with fetch disabled) verifies error starts with `missing-history:` and does not fall back to HEAD.
   - Base and head compared over one shared corpus (165 rows), evaluating both directions (`base withheld && head published` as WEAKER, `base published && head withheld` as STRICTER).
   - Controls before and after corpus run for both arms (positive publishes, negative withholds both TOOL_CALL_ARGS and TOOL_CALL_RESULT).
   - All 20 defined axes covered (original 15 + appended 5) with zero uncovered axes.
   - Representation axis covers own accessor, inherited accessor, and Proxy descriptor-lie with stateful get.
   - Zero-proof limits stated in header prose following packages/lang differential precedent.
4. Issue linkage: PR body references #1433 without closing keywords (`Refs #1433`), and `closingIssuesReferences` is empty.

davidfarah2003:
Review verdict: **ACCEPT**, bound to exact head `8fd2dd8013180eaa8d475dd1be0604aa5fef0532`.

**Ran first-hand:**
- Confirmed local HEAD and GitHub PR head both equal the named SHA. The worktree was clean before and after review.
- Built `@cotal-ai/connector-core` and dependencies, then ran `pnpm smoke:egress-guard-differential`: 39/39 cells passed, 165 corpus rows, 6 declared divergences, 0 held. The sensitivity pair (`ee73e33c…` vs `1698fe253…`) found WEAKER 117 / STRICTER 0. The primary pair (`ee73e33c…` vs HEAD) found WEAKER 0 / STRICTER 0. All 20 named axes had rows. Representation-flip alone had 102 rows and covered own accessor, inherited accessor, and a Proxy with a data-property descriptor trap plus stateful `get`.
- Ran `pnpm smoke:agui-tool-result`: 55/55 passed. Ran its declared mutation proof: all 6 mutations were KILLED on their named assertions. T-PARSE reds the `AguiEmitter.start` unreadable-WAL recovery cell, proving representative entry-point reachability. T-READONCE reds the own-accessor strict-read cell, while the same smoke contains independent inherited-accessor and descriptor-lying Proxy cells.
- Loaded both required historical classifiers directly with `git show <sha>:extensions/connector-core/src/agui.ts`.

**Read and graded:**
- One corpus is graded in both directions using an explicit publish mapping: base withheld + head published is WEAKER, base published + head withheld is STRICTER.
- Positive and both negative controls execute for both arms before and after the corpus. Their role-specific answers must be `ALLOW/REFUSE` or `clean/forbidden-kind`, so parser rejection or the wrong frame kind cannot pass vacuously.
- `loadArm` checks `git cat-file -e <sha>^{commit}`, fetches a miss using `git fetch --no-tags --depth=1 origin <sha>`, then extracts that ref's blob. A failed or still-missing object throws `missing-history:`. The `deadbeef…` fetch-disabled `ok()` cell passed and requires that prefix and no mention of HEAD. There is no fallback to HEAD and no skipped-corpus/shallow-green path.
- Each row records an axis. Output lists every covered axis with `n/weaker/stricter`, and the suite fails if any named axis is uncovered. The header states that zero does not prove against symmetric loss, unrepresented inputs, the WAL encoder, or live reachability.
- PR body uses `Refs #1433` without a closing keyword. GraphQL `closingIssuesReferences` is empty.

**Exact-SHA checks:** all 29 are terminal. 27 succeeded. The only non-success conclusions are `full` = skipped, and `attribution` = cancelled at 01:20:36Z. The latter is superseded by successful `attribution` started at 01:20:44Z, so neither is a blocker.

**Not checked:** I did not force the historical objects absent to exercise the network fetch branch in a fresh shallow clone. I did not run live suites or the full repository matrix locally. The exact-SHA CI run set is terminal as described above.

PR #1454, opened 2026-09-10
ci: run Linux x64 jobs on Tenki runners

davidfarah2003:
## Problem

The org is on GitHub's 20-concurrent-job cap, shared by Cotal and foreman and by every OS. Measured 2026-09-10 12:19Z: 20 jobs running, 151 queued; ubuntu jobs waited 9 minutes on average and 29 at most over the preceding two hours. Each PR push costs about 30 jobs, 28 of them Linux x64.

## Change

The Tenki GitHub App is installed on Cotal-AI. Every `runs-on: ubuntu-latest` in `ci.yml`, `mutation-reproof.yml`, `docs.yml` and `windows.yml` becomes `tenki-standard-medium-4c-8g`: the same core count as GitHub's `ubuntu-latest`, half the RAM (8 GB vs 16 GB). Tenki runs the official `actions/runner-images` Ubuntu 24.04 image, so toolchains, `sudo apt-get` and `actions/cache` are unchanged.

Left on GitHub, deliberately:

- `windows-latest` and `ubuntu-24.04-arm` jobs: Tenki has no Windows or arm64 runners.
- `installer.yml`: its lint job runs Docker, which Tenki's docs do not mention.
- `changesets.yml` and `github-releases-to-discord.yml`: publishing paths, rare, not the bottleneck.

Empty changeset (CI-only).

## Proof

This PR's own `pull_request` runs execute the workflow files at its head, so the CI, Mutation reproof, Docs and Windows runs on this PR are the first jobs to target the Tenki labels. A job that stays `queued` with no runner is, per Tenki's troubleshooting page, a label or a plan-capacity problem; a job that starts is the proof.

## Not verified here

- Memory: the heaviest smoke and reproof shards have not been run on 8 GB before this PR's own run. If a shard fails on memory, the job steps to `tenki-standard-large-8c-16g`.
- Plan concurrency: the Tenki workspace's concurrent-session limit is not visible from the repo. If it is below 20, this change would make the queue worse, and the PR's own fan-out (12 reproof shards + 6 CI jobs) will show it.

### DIFF
--- a/.changeset/tenki-linux-runners.md
+++ b/.changeset/tenki-linux-runners.md
@@ -1,0 +1,5 @@
+---
+---
+
+Run the Linux x64 jobs of CI, Mutation reproof, Docs and the Windows gate on Tenki runners
+(`tenki-standard-medium-4c-8g`). Windows and arm64 jobs stay on GitHub-hosted runners.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,12 @@ concurrency:
   # above: a PR keeps the default single slot and supersedes, main queues instead of evicting.
   queue: ${{ github.event_name == 'pull_request' && 'single' || 'max' }}
 
+# Linux x64 jobs run on Tenki (the org's runner provider; the Tenki GitHub App is installed on
+# Cotal-AI). `tenki-standard-medium-4c-8g` matches GitHub's ubuntu-latest core count at half the RAM;
+# if a smoke or reproof shard starts failing on memory, step that job to `tenki-standard-large-8c-16g`.
+# Windows and arm64 stay on GitHub-hosted runners: Tenki offers neither. Before this, every Linux job
+# shared the org's 20-concurrent-job cap with Windows and arm64, and on 2026-09-10 the queue reached
+# 151 jobs with ubuntu jobs waiting up to 29 minutes to start.
 # Same coverage as before, fanned across parallel jobs so wall-clock is the slowest job, not the sum:
 #   • unit  — typecheck + build + unit tests + the broker-free view/pack checks (fast).
 #   • smoke — the protocol/security suite (smoke:ci), round-robin sharded across runners by
@@ -35,7 +41,7 @@ concurrency:
 jobs:
   unit:
     timeout-minutes: 20
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     steps:
       - uses: actions/checkout@v6
         with:
@@ -100,7 +106,7 @@ jobs:
     # cancelled healthy runs (#1220 died at 40m18s). Sized at ~1.5x the slowest observed
     # success, which absorbs runner variance and lets the corpus grow before this binds again.
     timeout-minutes: 60
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     strategy:
       fail-fast: false
       matrix:
@@ -154,7 +160,7 @@ jobs:
 
   live:
     timeout-minutes: 25
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     steps:
       - uses: actions/checkout@v6
       - uses: pnpm/action-setup@v6.0.8
@@ -232,7 +238,7 @@ jobs:
   seat-native-linux-x64:
     name: seat native (linux-x64)
     timeout-minutes: 15
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     steps:
       - uses: actions/checkout@v6
       - uses: pnpm/action-setup@v6.0.8
@@ -275,7 +281,7 @@ jobs:
     name: seat pack (assembled)
     timeout-minutes: 20
     needs: [seat-native-linux-x64, seat-native-linux-arm64]
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     steps:
       - uses: actions/checkout@v6
       - uses: pnpm/action-setup@v6.0.8
@@ -391,7 +397,7 @@ jobs:
     name: ci-ok (live outcome classified)
     if: always()
     needs: [unit, smoke, live, seat-pack, seat-linux-arm64]
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     permissions:
       actions: read
       contents: read
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,7 +31,7 @@ on:
 
 jobs:
   docs:
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -15,7 +15,7 @@ jobs:
     # checkouts because mutation-proof temporarily changes source and compiled artifacts.
     if: github.event_name != 'schedule'
     timeout-minutes: 150
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     strategy:
       fail-fast: false
       matrix:
@@ -77,7 +77,7 @@ jobs:
   changed:
     if: github.event_name != 'schedule' && always()
     needs: [changed_shard]
-    runs-on: ubuntu-latest
+    runs-on: tenki-standard-medium-4c-8g
     steps:
       - name: Gate
         env:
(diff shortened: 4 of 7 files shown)


davidfarah2003:
Workflow semantics review for PR #1454 at commit 7fd08882c20b05a751b00c2ef7d471edbe55a7c6.

1. Required Checks
No check names are changed or deleted. Across ci.yml, docs.yml, mutation-reproof.yml, and windows.yml, every job name and matrix definition remains identical to main. If branch protection is configured in the future, existing rules naming these jobs will continue to match without vacuous passes.

2. Concurrency Policy and Cancellations
Concurrency definitions are not altered in this PR.

3. Runner Move Implications
Linux x64 jobs migrate from ubuntu-latest to tenki-standard-medium-4c-8g across ci.yml (unit, smoke, live, seat-native-linux-x64, seat-pack, ci-ok), docs.yml (docs), mutation-reproof.yml (changed_shard, changed, full), and windows.yml (windows-ok). Windows (windows-latest) and ARM64 (ubuntu-24.04-arm) remain on GitHub-hosted runners.
No cross-job shared filesystems or hardcoded home directories were broken. Artifact transfers between Tenki Linux x64 and GitHub ARM64 continue to route through actions/upload-artifact and actions/download-artifact. Cache keys for nats-server and pnpm rely on runner.os (Linux) and lockfiles; the Actions cache backend supports both runner providers. In verify-shard-inputs.sh, dynamic path detection accommodates runner home differences (/home/runner vs /home/tenki), and the corresponding sha256 pin in bin/smoke/shard-stability.ts is updated to match.
Explicit note on baselines: a green result on the new runner is a new baseline for any timing-sensitive cell, not a repair and not an acquittal. Historical test runs do not transfer across runner hardware, and suite timing must be established anew on Tenki.

Residuals:
Residual timing calibration on Tenki runners: acceptable operational notice; test history does not transfer.

Verdict: SHIP

workflow semantics lens

davidfarah2003:
Merged-tree review of #1454 (head 7fd08882c20b05a751b00c2ef7d471edbe55a7c6), together with #1451 (48ef0f58613b6d9f3e5434d4e0b938bb9a20897a) and #1455 (ef63e811548e062095aee00c5eab7062534bcc14), onto main 7c221ac71b763b9df3297e9bea6725bc9df556f1. This comment grades what actually lands when all three go in, because #1451 and #1454 both edit mutation-reproof.yml and docs.yml and no CI run has built their combination.

**Merged trees (git merge-tree --write-tree, plumbing only, nothing moved or pushed):**
- main + #1451: 06ed95e853703cb9edd774d934f312ac5415504f
- that + #1454: 393b06b3c9e00bc5f06fff15d869c511be53a4a4
- that + #1455: 749dc6891f0071931e15dd69aeb14af647ac8400
- Landing order #1454 first instead: d6ae8c0faaf252e616c7b6e04279376e46b357c3, then 393b06b3c9e00bc5f06fff15d869c511be53a4a4, then the same 749dc6891f0071931e15dd69aeb14af647ac8400
- Third permutation (#1451, #1455, #1454): d3774220b59b5803f4ef4ed9ff63b58789819d11, then the same 749dc6891f0071931e15dd69aeb14af647ac8400

Every merge exits 0 and the conflicted path set is empty at each step (read over the whole merge-tree output with an awk stanza filter, not off the first lines). All orders converge on the same tree, so landing order does not matter here.

**The shared files at 749dc6891f0071931e15dd69aeb14af647ac8400, read as final text:**
- mutation-reproof.yml: the concurrency block from #1451 sits intact between the triggers and `jobs:`, and all three jobs (changed_shard, changed, full) are still defined, each now on `tenki-standard-medium-4c-8g` from #1454. The merged-vs-main diff is exactly the 11-line concurrency insert plus the three runs-on swaps, nothing else.
- docs.yml: concurrency block intact above `jobs:`, the docs job on `tenki-standard-medium-4c-8g`. Exact union.
- The concurrency group is workflow-level, so #1451's policy still governs every job whose runner #1454 changed. All eight workflow files parse as YAML, and job sets are identical to main everywhere except the new attribution workflow. In windows.yml the required/smoke/soak jobs stay on windows-latest and only the summarizing windows-ok job moved to Tenki, matching the changeset text.
- For each of the eleven changed paths across the three PRs, the merged blob is byte-identical to the contributing PR's version, or, for the two shared files, the verified union of both.

**Changesets:** three distinct filenames, no collision, no accidental merge into one entry. All three carry empty frontmatter; the real @changesets parser accepts that as zero releases, and main already consumed two such empty changesets (mutation-reproof-gate.md, lease-renew-manager-control.md) in release 2798a3f8 without bumping anything. Coherent with repo convention.

**Cross-PR checks reproduced, not inferred:**
- sha256 of bin/smoke/verify-shard-inputs.sh at the merged tree equals the pin in shard-stability.ts at the merged tree (e6f6302f1a432b2eebe170fecc7dbf712c2c7545d1893ba01524318df9213bc6).
- scripts/check-attribution.mjs --selftest passes 10/10 at the merged tree, and the merged grader reports all three branch ranges clean, so the new gate does not wedge #1451 or #1454 after it lands.
- The runner labels are live: the attribution job ran green on ef63e8115's head, and on 7fd08882c's head the unit, smoke, live and windows-ok jobs all ran green on the Tenki labels.

**Residuals:**
1. The merged tree 749dc689 has never been executed by a CI run; the evidence is static. Disposition: accepted, since both PRs' edits to the shared files occupy disjoint regions and the union was verified byte-wise.
2. github-releases-to-discord.yml still has no concurrency group. Pre-existing on main, untouched by all three PRs. Disposition: out of scope, optional follow-up.

**Verdict: SHIP.** The "they merge cleanly" claim was tested on the resulting text and holds: one convergent tree from every landing order, no conflicted paths, both shared files semantically intact, coherent changesets, and no self-wedging gate.

merged tree lens

davidfarah2003:
Review of the executable shard tooling at 7fd08882c (verify-shard-inputs.sh and the shard-stability.ts pin).

The pnpm path matcher is a glob (`/home/*/setup-pnpm/node_modules/.bin*/pnpm`). After that match the script strips to a runner home and allows only two exact strings: `/home/runner` and `/home/tenki`. Exec also requires an executable file and a sha256 match on `readlink -f` of the captured path. The clean-env re-exec therefore grew by one home. It did not become an open `/home/*` pattern.

I drove the shipped script, not only the diff.

- A home that is neither (`/home/other/.../pnpm`): exit 2, `unexpected CI runner home`. Did not exec.
- Prefix or substring of an allowed home (`/home/runner-extra`, `/home/xrunner`, `/home/tenki2`, `/home/runner/foo/...`): exit 2, unexpected home. Did not exec.
- Relative path: from a non-repo cwd git fails first; from the repo the relative file is not executable and the script exits 2. Did not exec.
- Empty pnpm path: exit 2, not executable. Did not exec.
- Trailing slash on a file: not executable, exit 2. Trailing slash on a directory: glob miss, `unexpected CI pnpm path`. Did not exec.
- Symlink under `/home/tenki/.../.bin-link/pnpm` to another payload: the glob matches because of `.bin*`, the home is `/home/tenki`, and the sha is of the resolved target. If capture hashed that same target, the re-exec runs it. That is the captured blob, not a second home. The previous glob already used `.bin*` under `/home/runner`.
- Double slash and a `/.` segment strip to `/home/runner/` and `/home/runner/.`, both refused.
- `/home/tenki/.../.bin/pnpm` and `/home/runner/.../.bin/pnpm`: exec. Ambient USER/LOGNAME are discarded. The clean env sets USER, LOGNAME, and HOME from the matched path.

On the 10-arg CI path, USER and LOGNAME come from the last component of the matched home. They overwrite the environment, and the implied name does not need a passwd entry. `bin/smoke/shard.mjs` does not read USER, LOGNAME, or HOME. Those values are cosmetic on the CI arity.

The 3-arg local path is different: `runner_home` is `${HOME:-/home/runner}` with no allowlist, so USER/LOGNAME become the last component of HOME. CI does not call this arity.

The sha256 pin in shard-stability.ts was derived from the verifier blob at 7fd08882c. `git --no-replace-objects show HEAD:bin/smoke/verify-shard-inputs.sh` hashes to `e6f6302f1a432b2eebe170fecc7dbf712c2c7545d1893ba01524318df9213bc6`, which is the pin. The previous pin is the parent blob. The pin fail-closes the detector if the verifier text changes again.

Suites on this tree: `pnpm smoke:shard-stability` (7 passed), `pnpm smoke:ci-suites` (25 passed), `pnpm smoke:ci-fragments` (8 passed, including the multi-suite fragment refusal), `pnpm smoke:ci-suites-freeze` (9 passed), `pnpm check:shard-stability` against merge-base `7c221ac7` (STABLE, 538 suites, none moved).

A fragment file with two suites still throws in the shared reader before shard selection, so every shard on that tree fails the same way. These two files do not parse fragment bodies beyond byte identity (the verifier) and the detector's existing one-suite abort. That behavior was already there.

Residuals:
- Glob then two exact homes: the glob is a pattern; exec is still bounded by the two-home check. Accept.
- `.bin*` extra suffixes under those two homes still match if capture points there. Same class as before, now also under `/home/tenki`. Accept.
- 3-arg HOME has no allowlist. CI does not use it. Accept.
- USER/LOGNAME unused by shard.mjs. Accept.
- Two-suite fragment still fails the whole selector. Pre-existing, not this change. Accept.

SHIP 7fd08882c.

shell and shard lens

PR #1457, opened 2026-09-10
ci(reproof): fan out only the shards that hold a selected fixture

davidfarah2003:
## What

The Mutation reproof workflow plans its fan-out before it fans out. A new `plan` job (checkout and node only, no install, no build, small Tenki runner) runs the selector once via `scripts/mutation-reproof.mjs --list-shards 12` and emits the set of shards that hold at least one selected fixture. Only those shards are fanned; the matrix is `fromJSON` of the plan's output.

`--list-shards <count>` is exclusive with `--shard`. It runs the same selector, the same UNMEASURED refusals and the same dangling check as a shard run, then prints a human line and a JSON object as the last line. An empty set prints `[]`, never silence, so a caller can tell nothing-to-prove from nothing-was-said.

The `changed` gate requires the plan to have succeeded and accepts a skipped matrix only when the plan said `[]`. A failed plan is unmeasured and fails the gate. Each fanned shard still re-selects for itself, so a shard that disagrees with the plan prints its own selection.

## Why

Every push to a pull request fanned all 12 shards, and each shard checked out, installed and built before running the selector. On 2026-09-10, 107 of 228 shard jobs then reported that no selected fixture was assigned to them: 173 job-minutes of setup queued against the org's 20 concurrent-job cap, ahead of work with something to prove.

## Verification

- The workflow's own reproof fixture (`bin/smoke/mutations/mutation-reproof-shards.json`) caught this change on the first CI run: five mutations anchored on the matrix literal, the `needs` list and the `SHARD_RESULT` test that the plan job replaced. Re-anchored, plus six new mutations for the plan job (hardcoded matrix, fan-out that no longer waits on the plan, plan that installs before selecting, plan shard count diverging from the fan-out, gate accepting a failed plan, gate accepting a successful fan-out behind an empty plan, gate losing `set -e`). Full proof at 82fa8f11c: 21 of 21 KILLED, each red naming its cell.
- `bin/smoke/mutation-reproof.smoke.ts` now drives the gate shell over ten (plan result, planned shards, matrix result) cases instead of four matrix results; 95 of 95 cells green at this head.

- At 7c221ac71 on `ffa150174..7c221ac71`: the plan names shards 0, 1, 4 to 11 (10 of 12) for 20 selected fixtures. An independent replication of the shard hash over the same 20 paths gives the same set, with per-shard counts matching the shard runs checked (0=2, 1=3, 2=0, 3=0).
- A docs-only commit on top of main plans `[]`; a shard run on the same range selects 0 fixtures.
- Rebased onto 1c32566fc (Tenki runners): the plan job runs on the small Tenki runner, the shards on the medium one. The workflow-concurrency smoke passes on the rebased tree.
- Merges cleanly with the open concurrency-policy change to the same file.

## Not changed

The scheduled full sweep is untouched. The plan job is gated off the schedule event like the fan-out it replaces.

Empty changeset: CI only, no package change.

### DIFF
--- a/.changeset/reproof-plan-shards.md
+++ b/.changeset/reproof-plan-shards.md
@@ -1,0 +1,8 @@
+---
+---
+
+The Mutation reproof workflow now plans its fan-out: one small job runs the selector and names the
+shards that hold a selected fixture, and only those shards are fanned. On 2026-09-10, 107 of 228
+shard jobs installed and built the tree and then reported no fixtures assigned to them, 173
+job-minutes of setup against a 20-job concurrency cap. An empty plan skips the fan-out and the gate
+passes only when the plan itself succeeded; a failed plan is unmeasured and fails the gate.
--- a/.github/workflows/mutation-reproof.yml
+++ b/.github/workflows/mutation-reproof.yml
@@ -9,17 +9,58 @@ on:
   workflow_dispatch:
 
 jobs:
+  plan:
+    # One checkout runs the selector and names the shards that hold a selected fixture. Before this
+    # job every push fanned all 12 shards; on 2026-09-10, 107 of 228 shard jobs installed and built
+    # the tree and then printed "No selected mutation fixtures are assigned to shard N", 173
+    # job-minutes of setup against the org's 20 concurrent jobs. The plan uses the same selector,
+    # the same UNMEASURED refusals and the same dangling check as a shard run, so a shard that is
+    # not fanned is one the selector would have emptied. No install and no build: selection reads
+    # git and fixture JSON only.
+    if: github.event_name != 'schedule'
+    timeout-minutes: 10
+    runs-on: tenki-standard-small-2c-4g
+    outputs:
+      shards: ${{ steps.plan.outputs.shards }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 22
+      - name: Plan the fan-out
+        id: plan
+        env:
+          BASE: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
+        run: |
+          set -euo pipefail
+          base="$BASE"
+          if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+            base="$(git rev-parse HEAD~1)" || { echo "no base commit or first parent"; exit 2; }
+            echo "no base commit on this event; using first parent $base"
+          fi
+          out="$(node scripts/mutation-reproof.mjs --base "$base" --head "$(git rev-parse HEAD)" --list-shards 12)"
+          printf '%s\n' "$out"
+          # The plan's last line is the JSON. A selector refusal exits non-zero above and fails this
+          # job, which fails the gate: an unmeasured selection must not read as an empty fan-out.
+          shards="$(printf '%s\n' "$out" | tail -n 1 | node -e 'const p=JSON.parse(require("fs").readFileSync(0,"utf8"));if(!Array.isArray(p.shards))process.exit(3);console.log(JSON.stringify(p.shards))')"
+          echo "shards=$shards" >> "$GITHUB_OUTPUT"
+
   changed_shard:
     # At dfbb4550f, PR #1350 selected 32 fixtures and exhausted the 145-minute serial step.
     # Each path is assigned to one shard by the runner's existing hash. Shards use separate
     # checkouts because mutation-proof temporarily changes source and compiled artifacts.
-    if: github.event_name != 'schedule'
+    # Only the shards the plan named are fanned; the shard run re-selects for itself, so a shard
+    # that disagrees with the plan prints its own selection rather than trusting the plan's.
+    if: github.event_name != 'schedule' && needs.plan.outputs.shards != '[]'
+    needs: [plan]
     timeout-minutes: 150
     runs-on: tenki-standard-medium-4c-8g
     strategy:
       fail-fast: false
       matrix:
-        shard: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
+        shard: ${{ fromJSON(needs.plan.outputs.shards) }}
     name: changed shard ${{ matrix.shard }}/12
     steps:
       - uses: actions/checkout@v6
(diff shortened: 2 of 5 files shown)


davidfarah2003:
Workflow semantics review for PR #1457 at commit 82fa8f11ca900da495fe3d358fa01c5b9d0fab44.

Scope: Equivalence between --list-shards <count> and --shard <index>/<count>.

1. Selector and Partitioning Equivalence
Both execution paths in scripts/mutation-reproof.mjs share the identical corpus loading, diff detection, metadata-only exclusion filtering, and fixture selection logic.
When --shard N/K is supplied, selected fixtures are filtered by:
  shardOf(fixture.path, K) === N
When --list-shards K is supplied, the plan output computes:
  [...new Set(selected.map(({ path }) => shardOf(path, K)))].sort((x, y) => x - y)
Because shardOf is a pure function over fixture path strings and modulus count K, the planned shards set contains index N if and only if --shard N/K selects at least one fixture from the exact same diff.

2. Refusal and UNMEASURED Failures
Selection and corpus validation precede sharding in both paths.
Dirty working trees, non-ancestor bases, empty corpus configs, malformed fixture JSON, and dangling fixtures (deleted guarded source files or missing suite files) fail closed with exit status 1 or 2 before shard dispatch.
If any fixture is dangling or unmeasurable across the diff, the planner exits non-zero immediately. In the workflow, this marks the plan job as failed (PLAN_RESULT=failure) and skips changed_shard; the aggregate changed gate under set -euo pipefail tests PLAN_RESULT=success and fails closed.

3. Empty and Non-Empty Fan-Out
When a diff intersects no mutation fixtures (for example, documentation changes), --list-shards prints an explicit empty list [] rather than producing no output. The workflow changed_shard matrix is skipped, and the aggregate changed job verifies both PLAN_RESULT=success and SHARD_RESULT=skipped, passing cleanly.

Residuals:
None.

Verdict: APPROVE

workflow semantics lens

davidfarah2003:
Review of the mutation-reproof gate at 82fa8f11ca900da495fe3d358fa01c5b9d0fab44.

The Gate `run:` block is:

```
test "$PLAN_RESULT" = success
if [ "$PLAN_SHARDS" = "[]" ]; then test "$SHARD_RESULT" = skipped; else test "$SHARD_RESULT" = success; fi
```

under `set -euo pipefail`. I extracted that block from `.github/workflows/mutation-reproof.yml` and drove the extracted shell with `PLAN_RESULT`, `PLAN_SHARDS`, and `SHARD_RESULT`. This is not a copy of the smoke table.

Ten-case table (same cells the smoke asserts):

- plan=success shards=[0,3] matrix=success: accept
- plan=success shards=[0,3] matrix=failure: reject
- plan=success shards=[0,3] matrix=cancelled: reject
- plan=success shards=[0,3] matrix=skipped: reject
- plan=success shards=[] matrix=skipped: accept
- plan=success shards=[] matrix=success: reject
- plan=failure shards=[0,3] matrix=success: reject
- plan=failure shards=[] matrix=skipped: reject
- plan=cancelled shards=[0,3] matrix=skipped: reject
- plan=skipped shards="" matrix=skipped: reject

All ten matched.

Extra states against the same extracted shell:

- Cancelled plan, empty shards, skipped matrix: reject. Cancelled plan, nonempty shards, matrix success: reject. The first `test` fails unless the plan itself is `success`.
- Empty plan with matrix cancelled, failure, or blank: reject. Nonempty plan with blank matrix: reject.
- Partial skip is not a GitHub job result. `needs.changed_shard.result` is one of success, failure, cancelled, skipped. With `fail-fast: false`, any shard failure is `failure` and the gate rejects. I still fed `success,skipped` and `success skipped`. Both reject.
- Malformed shards that are not exactly `[]`, with matrix success: accept (`null`, `0`, `true`, `{}`, `[0,]`, `[[0]]`, `{"shards":[0]}`, `"[]"`, leading or trailing space, a newline after `[0]`, empty string). The `else` branch only tests matrix=success. The same strings with matrix=skipped reject.

On this workflow that last class does not reach GitHub. A successful plan parses the selector last line as `{shards: number[]}` and writes `JSON.stringify` of that array, so `needs.plan.outputs.shards` is compact `[]` or `[0,2]` and so on. A failed or skipped plan leaves the output empty, but `PLAN_RESULT` is then not `success` and the first test already rejects. `fromJSON` of an empty or malformed output fails matrix expansion, so the pair (plan success, empty shards, matrix success) does not occur on the graph.

`--list-shards 12` on this SHA printed `{"shards":[0,2]}` as the last line. The plan job's `JSON.parse` / `Array.isArray` / `JSON.stringify` path yields `shards=[0,2]`.

Belonging smoke `pnpm smoke:mutation-reproof`: 95 passed, 0 failed. That includes the ten aggregate-shell cells, the plan-job / matrix / `always()` wiring checks, and the shard partition cells.

Residuals:

- Exact string `[]` is the only empty plan. Whitespace around it is treated as a live matrix. The plan job cannot emit that whitespace. Accept.
- Empty `PLAN_SHARDS` plus matrix success is accepted by the shell. The plan job cannot succeed without writing a JSON array, and an empty output would fail `fromJSON` rather than report matrix success. Accept.
- GitHub has no partial-skip result string. Mixed shards collapse to failure or success. Accept.

Verdict: APPROVE

gate-table lens

davidfarah2003:
Merged-tree and workflow review of this PR at 82fa8f11ca900da495fe3d358fa01c5b9d0fab44 (one commit over merge-base 1c32566fc1b837d59779a9a9e918c5574f93f88b; five files: mutation-reproof.yml, scripts/mutation-reproof.mjs, its smoke, the shard mutation config, and the changeset).

Interaction with #1451, re-verified at current main 36d1779519f01e9cc04195e88d94435d61c368b1 rather than the old base: main + this PR merges clean to d3dd2c1581280dd2f70c0732817fa5e7813f2b21; adding #1451 at 2a3a5da30de0a17bb812216fdc5f8d04191ea85b gives c828cbaac763d10e7498874fde0fc4e291c16a21, and the reverse order (main + #1451 first, eaf15e7ab5ddb3b398617f539446d74296c576a7, then this PR) lands on the same c828cbaa. No conflicts either way. Read as final text at the union tree: the workflow-level concurrency block sits above jobs: intact and the plan job is the first job, exactly as each side wrote it; the concurrency group governs the whole run including the new plan job, which is correct since a superseded run's plan is as disposable as its shards. The file parses as YAML with jobs plan/changed_shard/changed/full.

The empty-plan path: the plan emits `shards=[]` through JSON.stringify, changed_shard's if is `github.event_name != 'schedule' && needs.plan.outputs.shards != '[]'`, so an empty plan skips the matrix, and fromJSON('[]') is never evaluated for a skipped job. The gate shell, driven directly across the state table, accepts exactly [success, [], skipped] and [success, non-empty, success], and rejects failed/cancelled plans, non-empty-plan-with-skipped-matrix, and empty-plan-with-success-matrix. A skipped fan-out cannot read as a pass unless the plan itself succeeded and said [].

Selector equivalence, driven rather than inferred at this head: `--list-shards 12` for the diff against 1c32566fc selects shards [0,2]; iterating `--shard N/12` for every N and checking which report fixtures independently agrees on [0,2]. Same selector, same refusals: the ancestor check, dirty-tree check, malformed-fixture UNMEASURED refusals and the dangling-anchor check all run before the branch that prints the plan, so a plan that prints has passed every gate a shard run would have.

The force-push/first-parent concern: on a push where github.event.before is a commit the fresh clone does not contain, the selector refuses loudly (verified: a missing base exits with "not an ancestor ... refuse a comparison" rather than printing a plan), which fails the plan job and the gate; the 40-zero base falls back to HEAD~1 inside the same guarded step. The pre-existing fallback shape is unchanged by this PR; only the consumer of the selection changed.

Local suites at this head: mutation-reproof.smoke.ts 95/95 passed after a frozen-lockfile install. Runner labels tenki-standard-small-2c-4g and tenki-standard-medium-4c-8g are both in live use on already-landed workflows. Changeset carries the repo's empty-frontmatter convention for CI-only changes.

Verdict: APPROVE 82fa8f11ca900da495fe3d358fa01c5b9d0fab44. Residual: none blocking. The gate table accepts [success, malformed-non-empty, success] in the abstract, but the plan's JSON.stringify output path makes a non-empty non-JSON shards string unrepresentable on this graph, matching the panel's accepted disposition.

merged tree lens

Issue #1466, opened 2026-09-10
mutation-coverage: documented Node spawn forms are false-refused by the executes witness


davidfarah2003:
## What

The `executes` gradability witness recognises only some of the documented ways a suite launches a repository entrypoint. Documented Node forms are **refused**, so a genuine execution witness is rejected and the mutation cannot be graded.

## Reproduction

Throwaway fixture repo, tally-only commands, nothing real executed. Each fixture is otherwise **identical** to a positive control that grades — same declared `executes`, same real package build, same mutation.

| suite form | result |
| --- | --- |
| `spawnSync(process.execPath, [ENTRY])` (**positive control**) | GRADED |
| `execFileSync(process.execPath, [ENTRY])` | **REFUSED** |
| `spawnSync("node", [ENTRY])` | **REFUSED** |

Previously reported alongside these, from the same class: `cp.spawnSync(process.execPath, [ENTRY])`, `import { spawnSync as run }`, and a quoted `pnpm --filter '@cotal-ai/seat' build`.

The refusal message names the mutation as targeting a file the suite does not reach, which is the correct message for a suite that genuinely does not reach it and a misleading one here.

## Direction of failure

This is over-**strictness**. It fails toward refusing to grade, which is the safe direction for an audit tool and the opposite of the over-admission class in #1434. It is a real defect and it does not hide missing coverage; it discards real coverage.

## Corpus exposure: latent, not live

- `execFileSync` appears in **14** suite files; `spawnSync("node", …)` in **4**.
- Configs that declare `executes`/`assembles` **and** whose suite uses one of those forms: **0**.

So no coverage is being lost today. It becomes live the moment someone writes an `executes` config over one of those 18 suites.

## History

Originally reported as a blocking review finding on 2026-09-08 against a head that is no longer an ancestor of the branch, and not carried forward across the rewrite. Related: #1434.

davidfarah2003:
**Trigger condition, stated plainly**, because it is the useful part of this issue:

The false-refuse only bites a config that declares `executes` or `assembles` over one of the suites using `execFileSync(process.execPath, …)` or `spawnSync("node", …)`. **Zero such configs exist today.** It becomes live the moment someone writes one — and that person will read a refusal saying their mutation targets a file the suite does not reach, and conclude their *suite* is wrong.

An issue that names the trigger converts a confusing future refusal into a two-minute lookup.

[...]

Issue #1434, opened 2026-09-10
mutation-coverage: all three gradability witnesses accept a mention instead of an execution


davidfarah2003:
## What

`scripts/mutation-coverage.mjs` decides whether a mutation is gradable using three **textual** witnesses. All three accept a suite that merely *mentions* the right characters, without running or copying anything. A mutation the resolver would hide behind `dist` can therefore be reported as covered.

The script says this about itself, in the docblock above `referencesRoot`:

> That reference check is the same order of approximation as the `../src/` substring above — a textual witness for a data-flow fact

This issue is to close the class, not one route. Tightening any one witness leaves the other two, and the next fixture reaches for whichever is loosest.

## The three witnesses

| witness | rule | where |
| --- | --- | --- |
| `../src/` | same package root, and the suite source contains the substring `../src/` | `gradableFrom` |
| `assembles` | the mutated file is under a declared root, and the suite source references that root | `referencesRoot` |
| `invokesFile` | a spawn token, then the target's quoted basename anywhere within the next 500 characters | `invokesFile` |

## Reproduction

A fixture tree with a package, a script, and one suite per shape. Each config's `command` only prints a tally, so acceptance is the whole result.

- **A** — `"assembles": ["packages/seat"]`, target `packages/seat/src/impl.ts`, suite is a by-name `@cotal-ai/seat` import plus an unused `join(ROOT, "packages", "seat")`. **ACCEPTED.**
- **B** — target `scripts/direct.mjs`, suite is `spawnSync(process.execPath, ["-e", "void 0"]);` followed by an unused `join(ROOT, "scripts", "direct.mjs")`. **ACCEPTED.**
- **C** — target `packages/seat/src/impl.ts`, suite is in the same package, imports by name, and contains only `const note = "see ../src/impl.ts for details";`. **ACCEPTED.**

Negative controls discriminate: A without the root reference and B without the spawn token are both refused.

**C reproduces against `ffa1501748d` with none of #1378's code**, using the predicate as it ships today:

```js
const suite = suites.find((source) =>
  packageRoot(m.file) === packageRoot(source) && readFileSync(source, "utf8").includes("../src/"));
```

A `.includes()` substring test is satisfied by a comment. So the class predates the `assembles` and `invokesFile` witnesses; those added two more instances of it.

## Correction: the repo instance is LATENT, not live

An earlier revision of this issue claimed `bin/smoke/fixtures/mutation-fixtures.mutations.json` is
[...]

PR #1468, opened 2026-09-10
fix(mutation-coverage): grade aliased launchers and bound entry names

L4XB:
Refs #1466.

**What the witness does at head.** I rebuilt the issue's fixture shapes with the self-test's own harness (throwaway root, tally-only command, one mutation on `scripts/direct.mjs`) and ran `scripts/mutation-coverage.mjs` from `main` against each suite form:

| suite form | main |
| --- | --- |
| `spawnSync(process.execPath, [join(ROOT, "scripts", "direct.mjs")])` (control) | graded |
| `execFileSync(process.execPath, [join(ROOT, "scripts", "direct.mjs")])` | graded |
| `spawnSync("node", [join(ROOT, "scripts", "direct.mjs")])` | graded |
| `cp.spawnSync(process.execPath, [join(ROOT, "scripts", "direct.mjs")])` | graded |
| `import { spawnSync as run }` + `run(process.execPath, [join(ROOT, "scripts", "direct.mjs")])` | **refused** |
| `const ENTRY = join(ROOT, "scripts", "direct.mjs"); execFileSync(process.execPath, [ENTRY]);` | **refused** |

So the two headline forms of #1466 already grade at head (the regex in `invokesFile` lists `execFileSync`, and `"node"` is never inspected), but two spellings from the same class in the issue's history still are refused: an aliased launcher import, and a target bound to a name before the call, which is exactly the `[ENTRY]` shape the issue's table uses whenever `ENTRY` is declared above the call rather than inline. The refusal message then says the suite does not reach the file.

**Change** (`scripts/mutation-coverage.mjs`)

- `launcherNames()` adds `x as y` aliases from the suite's `child_process` import to the four canonical launcher names.
- `boundNames()` collects identifiers bound to a quoted mention of the file; such a name counts only when it sits inside a launcher's argument list (one level of nesting), so `const ENTRY = …; spawnSync(process.execPath, ["-e", "void 0"]); console.log(ENTRY)` stays refused.
- The existing quoted-basename-within-500-characters rule is unchanged, including its known over-admission (#1434, self-test cells 9/10).

**Self-test** (`scripts/mutation-coverage.selftest.mjs`): three graded cells (aliased launcher, bound entry, `"node"` by name) and one refused cell (bound name outside the argument list). `node scripts/mutation-coverage.selftest.mjs` → 14 passed. Changeset added with no package bump (scripts only, same shape as `attribution-check.md`).

### DIFF
--- a/.changeset/mutation-coverage-executes-witness-forms.md
+++ b/.changeset/mutation-coverage-executes-witness-forms.md
@@ -1,0 +1,4 @@
+---
+---
+
+`mutation-coverage`'s executes witness now recognises a launcher imported under an alias (`import { spawnSync as run }`) and a target bound to a name before the call (`const ENTRY = join(ROOT, "scripts", "x.mjs"); execFileSync(process.execPath, [ENTRY])`), so a suite that really runs the mutated script is graded instead of refused as not reaching it. A bound name only counts inside the launcher's argument list; a name that is merely printed after an unrelated spawn stays refused. Self-test cells cover both directions plus `spawnSync("node", …)`.
--- a/scripts/mutation-coverage.mjs
+++ b/scripts/mutation-coverage.mjs
@@ -101,11 +101,40 @@ const quoted = (s) => `["'\`]${s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}["'\`]`;
  */
 const referencesRoot = (suiteSource, root) =>
   new RegExp([quoted(root), root.split("/").map(quoted).join("\\s*,\\s*")].join("|")).test(suiteSource);
+const LAUNCHERS = ["spawnSync", "spawn", "execFileSync", "execFile"];
+/**
+ * The names a suite calls the child_process launchers by: the canonical four plus whatever a
+ * `{ spawnSync as run }` import binds them to. A member call (`cp.spawnSync(`) already matches
+ * the bare name.
+ */
+const launcherNames = (suiteSource) => {
+  const names = new Set(LAUNCHERS);
+  for (const m of suiteSource.matchAll(/import\s*\{([^}]*)\}\s*from\s*["'](?:node:)?child_process["']/g)) {
+    for (const part of m[1].split(",")) {
+      const alias = part.match(/^\s*(\w+)\s+as\s+(\w+)\s*$/);
+      if (alias !== null && LAUNCHERS.includes(alias[1])) names.add(alias[2]);
+    }
+  }
+  return [...names];
+};
+/** Identifiers the suite binds to a quoted mention of the file (`const ENTRY = join(ROOT, "scripts",
+ *  "x.mjs")`), so a later `spawnSync(process.execPath, [ENTRY])` still names the target. */
+const boundNames = (suiteSource, basename) =>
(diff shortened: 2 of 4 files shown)


davidfarah2003 (approved):
Independent review of PR #1468 at sha 6566b94702937ab62878523c55c63383edb97f83: APPROVE.

First-hand verification in clean worktree:
1. `git rev-parse HEAD`
   Verified exact sha: 6566b94702937ab62878523c55c63383edb97f83

2. `node scripts/mutation-coverage.selftest.mjs`
   14 passed, 0 failed.
   Verified direct script launcher, aliased launcher, bound entry const, node spawn by name, and negative controls.

3. `node scripts/mutation-proof.mjs --config scripts/mutations/mutation-coverage.json`
   All 6 mutations killed on their named assertions:
   - "a suite mutating its own directly executed source is rejected" -> killed on "a suite is gradable when it directly executes the file being mutated"
   - "a suite launching the exact mutated repo script is rejected" -> killed on "a suite is gradable when it launches the exact mutated script path"
   - "any quoted mutation path is treated as a direct execution witness" -> killed on "a quoted script path without an invocation is refused"
   - "launcher aliases are ignored" -> killed on "a launcher imported under an alias still witnesses the script it runs"
   - "bound entry names are ignored" -> killed on "a target bound to a const before the launcher call still witnesses it"
   - "a bound entry anywhere after a launcher is treated as its argument" -> killed on "a name bound to the target but never passed to a launcher is still refused"

Mechanism inspection:
Bound names are properly constrained to the argument list of child_process launcher calls, preserving rejection of unused bindings. Changeset is clean and accurate.

davidfarah2003 (approved):
Approve at 6566b94702937ab62878523c55c63383edb97f83 (PR head verified; graded first-hand in a detached worktree at exactly that sha, clean before and after).

Commands run and results:
- git rev-parse HEAD -> 6566b94702937ab62878523c55c63383edb97f83
- node scripts/mutation-coverage.selftest.mjs -> 14 passed, 0 failed, including the new cells: aliased launcher graded, bound-entry const graded, spawnSync("node", ...) graded, quoted-path-only refused, bound-name-outside-arg-list refused.
- node scripts/mutation-proof.mjs --config scripts/mutations/mutation-coverage.json -> all 6 mutations KILLED, each red on its named assertion, not collateral: quoted-path (7 marks vs baseline 14), launcher alias (4 marks), bound-entry (5 marks), unused-bound (8 marks). Tree restored byte-clean after the proof.

Independent checks beyond the required runs:
- Reproduced the defect at origin/main 981606262255fd8c56c9c250341144dd907d46b8: the aliased-launcher suite is false-refused ("none of [...] imports by source path"); at this head the same fixture is graded ("1 / 3 cells observed failing").
- Nesting control: a bound name passed one call deeper inside the launcher argument list is graded; the same shape with an unrelated argument and the name only printed after is refused.
- Ordering is sound: the quoted-path negative cell runs before bound-entry-unused, and the two mutations (collapsing invokesFile to referencesRoot, widening the argument-list regex) kill on distinct cells, so each guard is individually controlled.

Diff is exactly the four files claimed (scripts/mutation-coverage.mjs, its selftest, the mutation config, the changeset); no scope creep.

Issue #1473, opened 2026-09-11
mutation-reproof FATAL_VERDICTS has no base-vs-head, so a registered attach-reconnect survivor fails every selecting PR


davidfarah2003:
## What

`scripts/mutation-reproof.mjs` treats `SURVIVED` / `UNGRADABLE` / `WRONG-RED` / `ERROR` as fatal with no base-vs-head comparison. PRE-RED already compares. A registered survivor that already exists on main therefore fails every PR that causes that fixture to be selected, and the check rollup looks like a regression.

## Where

At `54ce14e7bb9f468309f300df30f4f6317292c2c8`:

- `scripts/mutation-reproof.mjs:314` — `FATAL_VERDICTS = {SURVIVED, UNGRADABLE, WRONG-RED, ERROR}`
- `scripts/mutation-reproof.mjs:526` — `if (verdicts.some((v) => FATAL_VERDICTS.has(v))) fatal.push(path);` with no base comparison
- `bin/smoke/mutations/attach-reconnect.json` mutation 10 (unref'd link deadline) is a registered survivor. The fixture's own `why` says it is kept so a run reports `SURVIVED` rather than a smaller set. #634 tracks the observing instrument. The recording commit `93ec1d92b` is an ancestor of main.

## Why main looks green

Main Mutation reproof [34548444753](https://github.com/Cotal-AI/Cotal/actions/runs/34548444753) at 54ce14e7 is SUCCESS. The plan log selected 1 of 389 fixtures: `scripts/mutations/mutation-coverage.json`. `attach-reconnect.json` was not selected. Changed-file fan-out only grades fixtures whose files changed. A SUCCESS rollup is therefore not evidence that `attach-reconnect.json` is healthy on main.

That is a second misattribution, independent of the missing FATAL_VERDICTS base-vs-head: "mutation green on main, red on the PR" is indistinguishable from "main never ran this fixture" unless the selection list is opened.

## What selects it

Any PR that touches `implementations/manager/smoke/attach-reconnect.smoke.ts` or `bin/smoke/mutations/attach-reconnect.json` selects the fixture, including PRs whose diff does not touch the surviving mutation in `implementations/cli/src/commands/agents.ts`.

## What not to do

Do not de-register mutation 10 to green the shard. That hides the coverage gap #634 names. The tooling fix is to attribute FATAL_VERDICTS the same way PRE-RED is attributed: an inherited `SURVIVED` already present at base is not a PR regression. The product fix remains #634.

[...]

PR #1383, opened 2026-09-08
fix(test): audit aggregate mutation coverage

davidfarah2003:
This PR's own coverage config was refused by this PR's own new pairing rule. That rule had zero tests and zero mutations covering it.

## Progress-marker summaries (c)

Main's `scripts/mutation-coverage.mjs` has no `progressPattern`. This PR adds the parser and type-checks the keys. Adoption is later (#1444).

Counted at head `07d769c1f` (rebased onto `5b2281cd`) with the tool's own glob (`git ls-files '*/mutations/*.json' '*.mutations.json'`), not a narrower filter. Glob counts are unchanged from `e652804c2`:

- 388 configs enumerated
- 65 carry `progressPattern`
- 32 carry `minTicks`
- 31 paired (both keys)
- 35 configs the dropped pairing rule would have refused, in both directions: 34 `progressPattern` without `minTicks`, plus 1 `minTicks` without `progressPattern` (`extensions/connector-jcode/smoke/mutations/jcode-readiness.json`)
- 4 carry all three keys (`progressPattern`, `minTicks`, `completionMarker`):
  - `extensions/connector-jcode/smoke/mutations/jcode-diagnostics.json`
  - `extensions/connector-jcode/smoke/mutations/jcode-no-self-update.json`
  - `packages/core/smoke/mutations/provision-stream-info-role-gate.json`
  - `packages/core/smoke/mutations/read-acl-stream-info-scope.json`

The dropped rule was symmetric (`(progressPattern === undefined) !== (minTicks === undefined)`). 35 is the operative number for this PR because the subject is that refusal, not a one-direction key distribution.

The progress path does not fire for any config that could be measured: two of the four triples are preempted by the tallied parser, one is refused before parseSummary is reached (missing `cell`), and the fourth is excluded because it invokes a live suite and is therefore undetermined rather than cleared.

The pairing *refusal* this PR briefly added had zero coverage: one reference in the tree, its own definition, no self-test cell, no mutation. That is why it is a diagnostic, not a hard refusal. Type checks stay (`progressPattern` a non-empty compilable regex; `minTicks` a positive integer). When no parser matched, UNPARSED now names both directions of that dropped rule, plus the two completion-marker holes: `progressPattern` present and `minTicks` absent; `minTicks` present and `progressPattern` absent, so the progress path cannot run at all because `minTicks` is only ever consumed by that path; `completionMarker` undeclared; or a declared marker not present on the final output line.

[...]

PR #1475, opened 2026-09-11
fix(ci): attribute mutation-reproof fatals by base-vs-head verdict

davidfarah2003:
## Summary
- Mutation reproof treated SURVIVED, UNGRADABLE, WRONG-RED, and ERROR as absolute findings with no base-vs-head comparison.
- A fixture whose named fatal verdict is unchanged from the base commit is inherited and does not fail the PR. A verdict the PR actually moves still fails. Unmeasurable base comparison still fails closed.
- Controls: inherited SURVIVED on a comment-only select is green; a PR that moves KILLED to SURVIVED stays red. A mutation that deletes the comparison reds the inherited-SURVIVED cell.

Refs #1473

## Test plan
- [ ] `pnpm smoke:mutation-reproof` green
- [ ] Mutation reproof at this exact head
- [ ] Tenki CI including ci-ok
- [ ] Confirm #1383 / #1435 / #1445 would no longer inherit the registered survivor as a PR red

### DIFF
--- a/bin/smoke/mutation-reproof.smoke.ts
+++ b/bin/smoke/mutation-reproof.smoke.ts
@@ -460,28 +460,33 @@ try {
     );
   }
 
-  // 1d. Duplicate mutation names are valid. Identity is file+find, not the displayed label. A PR
-  //     that turns a later same-name KILLED into ERROR must fail even though an earlier same-name
-  //     ERROR is inherited.
+  // 1d. Duplicate mutation names are valid. Identity is file+find+replace, not the displayed
+  //     label. A later same-name mutant that newly becomes ERROR must still fail; matching by
+  //     label would inherit an earlier same-name ERROR from a different replacement.
   {
     const { root, base, head } = makeSingle(
       (r) => {
-        writeFileSync(join(r, "dup.mjs"), "export const dup = () => 1;\nexport const extra = () => 0;\n");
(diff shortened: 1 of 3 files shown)


davidfarah2003:
FAIL at b1cc10121635e41ba42a3ee406df01496e932c2d.

CI 34582723072 SUCCESS. Mutation 34582723059 SUCCESS. Those runs, the 98/0 smoke, and the 26/26 mutation-proof kill set do not cover mutation-array identity.

First-hand: base [equivalent survivor, positive control kill] is SURVIVED+KILLED. Head only reverses those unchanged objects to KILLED+SURVIVED. Reproof compares by occurrence index in scripts/mutation-reproof.mjs, prints MUTATION REPROOF FAILED, and exits 1. Inserting a killed sibling before the unchanged survivor fails the same way. An unchanged fatal verdict is required to be inherited and non-fatal. Array position is not stable mutation identity.

davidfarah2003 (commented):
PASS for the contract repair at `b1cc10121635e41ba42a3ee406df01496e932c2d`.

Inherited fatal verdicts are attributed by base-versus-head occurrence identity, while newly introduced fatals still fail. The duplicate-name false-clearance reproducer now exits nonzero, the inherited `SURVIVED` control remains nonfatal, and the registered ANSI-strip and label-identity mutations are killed.

Acceptance evidence: CI run `34582723072` succeeded, including `ci-ok`; Mutation reproof run `34582723059` succeeded.

PR #1436, opened 2026-09-10
feat(auth)!: add hosted managed retirement authority

davidfarah2003:
## Scope

Compose remote manager maintenance, retained-agent validation and hosted terminal retirement through host-owned authority.

- Add a closed retirement phase. Derive owner and manager principal server-side, require current registration proof and serve epoch, and mint only a short-lived requester pinned to the exact managed lifecycle.
- Complete host release before requesting terminal retirement. Failed release retains the alias hold. Successful terminal confirmation permits fresh-lifecycle reuse without ignoring a newer hold.
- Validate retained actors through the typed registry-pinned HTTPS authority route. Signing records stay on the host. Bind every returned identity, lifecycle and credential envelope before adoption.
- Move manager goal-index enumeration to an owner-filtered host scanner. Remove participant records-consumer create, info and delete powers. Validate key/body agreement and delete the temporary scanner after use.
- Make the runtime authentication test wait for its actual request's consumer before publishing. Production runtime and permissions are unchanged.
- Update embedding documentation, generated docs, typed fixtures and fixed-group minor changesets. Required remote-authority callbacks change the embedding API.

Current source: `990827fc98a10123a827818b3906ffd613fb2da8`, tree `b21e0600e601589edb9dc525be48e0662d6eac92`. This preserves the previous public `876c70c7` ancestry and combines reviewed maintenance/retirement source with the runtime readiness correction. The stock activation and remote admin follow-ups described below are excluded.

## Current owner-native verification

The owner receipt binds the isolated Linux ARM64 guest run to exact source `990827fc98a10123a827818b3906ffd613fb2da8`, tree `b21e0600e601589edb9dc525be48e0662d6eac92`, and NATS 2.12.12. It records that all 2,155 tracked source blobs and executable modes matched before and after execution; the private submodule was explicitly excluded from that census. The complete service log independently carries the gate output and normal terminal status listed below.

| Gate | Result |
| --- | --- |
| Complete workspace build and typechecks | Passed |
| Documentation voice | 40 pages passed |
| Authenticated runtime and admission boundaries | 74 passed, 0 failed |
| Host-owned manager goal-index scanner | 8 passed, 0 failed |
| Participant executor permission refusals | 6 passed, 0 failed |
| Registry-pinned HTTPS retained validation and direct-Manager retirement | 43 passed, 0 failed |

[...]

PR #1479, opened 2026-09-11
fix(manager): renew the remote authority credential family before expiry

davidfarah2003:
## Scope

Renew the stock remote manager's credential family through its pinned host authority before the short-lived executor expires. Bind responses to the current request and identity family, preflight all five credentials, and install them before reconnecting their holders.

Compensate incomplete activation and renewal through durable, request-scoped cleanup intent. The real KV gate reports whether staging created a row or reused an existing one. Cleanup preserves prior active credentials, attempts every owned revoke, retains audit rows, and refuses new issuance while cleanup debt remains. Activation and renewal are serialized per instance. Missing rows with known creation ownership remain explicit debt.

The generic issuance-gate API retains its explicit-revocation behavior, including revocation after an identical restage. The concrete KV gate adds atomic ownership reporting; manager-specific wrappers use that result to limit rollback to request-created rows. Participant roots receive no host signing authority or ledger state.

This follows the lifecycle work merged in #1436.

## Verification

- Original executor-expiry failure reproduced after the real five-minute lifetime. Fixed stock path: early control 16/0, original-expiry-plus-five-seconds 16/0, clean retirement 15/0.
- Original compensation implementation failed the new same-clock regression by revoking the prior active serve row. The repaired suite passed 65/65 assertions against the shipped authority plane and a real broker.
- The unchanged auth session-adapter suite reproduced the intermediate shared-revoke regression at 80 passed, 1 failed. The compatibility repair passed 81/0. Same-clock activation and renewal rollback are covered by the compensation suite.
- Six compensation mutations failed at their named assertions. All seven positive controls passed 65/65. Source and compiled targets were restored and checked.
- Earlier manager binding and holder-order proof: 23 named mutations and 24 controls reporting 70/0. Disabling renewal and retaining the expired executor also failed their real-duration checks.
- Final integrated source passed documentation regeneration, all workspace builds, all workspace typechecks, and all package test scripts. Structural inventory and CI-registration gates passed. Mutation metadata checked 397 fixture files and 2,127 anchors with no missing or ambiguous targets.
- Independent process, cgroup, temporary-root and read-error checks were zero. Final proof logs contained no credential material.

The compensation suite controls the clock for its same-second collision cases. The stock expiry test waits through the original real lifetime. Focused CAS-loss cells control only gate collaborators while exercising the shipped compensation code and real auth KV.

## Limits

Missing, corrupt, or superseded-coordinate cleanup debt stays fail-closed pending operator reconciliation. This change does not promise exactly-once HTTP delivery. Completed CI on `2825505b02944ef9d17e0b62793446eb91e4dff9` reports 35 successful checks and 2 skips. The separate attribution policy check failed and still blocks merge. No package publication or deployment is included.


davidfarah2003:
Fixed the Linux and Windows unit failure in 48cda805ac97cb85f31246a8885a2b0c638422ec. The existing fixture used 1970 expiry timestamps; it now scopes and restores a deterministic clock, preserves the envelope checks, and covers rejection exactly at expiry. Production code is unchanged.

Owner Linux: manager typecheck and all eight package test scripts passed in 11.152s. The new five-cell expiry suite passed before and after changing the production comparison from <= to <; that mutation failed at the new boundary assertion. All 2179 tracked source blobs were restored and verified. Independent process, cgroup, temporary-root and read-error checks were zero. The three focused logs contained no credential material.

Auth partial-family cleanup and final combined acceptance remain open.

Issue #1432, opened 2026-09-10
AG-UI egress fence reads only .events[].type, so tool bytes on a sibling property publish


davidfarah2003:
The AG-UI egress fence reads `.events[].type`. Tool bytes carried on any other property of a frozen
frame part are not looked at and publish.

Measured on PR #1429, a frozen body of the shape

    { kind: "ag-ui.frame", protocol: "ag-ui/0.0.57", threadId, runId, epoch, seq,
      events: [{ type: "RUN_STARTED", ... }],
      recovery: { type: "TOOL_CALL_RESULT", content: "<bytes>" } }

publishes with the bytes intact. The `events` array is well-formed and clean, so the fence returns
`clean` and the sibling property rides along. This is **pre-existing** — it measures identical on the
pre-#1429 fence, which parsed the frame strictly and also never looked outside `events` — and #1429
neither introduces nor widens it.

The fence is event-**kind** control on a known property, not content classification over a body. It
answers "does this frame carry a forbidden event type" and cannot answer "does this frame carry tool
output anywhere". Those are different questions and the second one is not solved by extending the
first.

No shipped writer produces such a body today: `aguiFrame()` constructs the part and carries no
passthrough for extra properties. The exposure is that nothing enforces that — a future field on the
frame part, or any writer reaching `EventWal.beginSend` directly, lands outside the fence with no
gate and no failure.

**What a fix has to decide:** whether frozen frame parts get a closed shape enforced at the WAL
boundary (so an unknown property is refused rather than carried), or whether the fence stops reading
one property and starts classifying the body. The first is cheap and bounds the class; the second is
the general problem.

Relates to #695.

davidfarah2003:
Confirmed on isolated origin/main `981606262255fd8c56c9c250341144dd907d46b8` (fetched 2026-09-10T23:40Z). The AG-UI egress fence is still event-kind control on `.events[].type`. A well-formed frame with a forbidden kind on a sibling property still publishes.

## What still reproduces (read-only, no mesh)

Worktree: `/home/cotal/.cotal/jcode-scratch/lane-1432-triage/` detached at that SHA. No live fleet, no `cotal up`/`down`.

Mechanism on this SHA:

- `frozenBodyEgressVerdict` (`extensions/connector-core/src/agui.ts:1335`) calls `parseAguiFrame` then scans `frame.events` for `isForbiddenEgressEventType((e).type)` (`:1356`). Extra properties are never read.
[...]

PR #1501, opened 2026-09-12
fix(connector-core): enforce closed schema on AG-UI egress fence

davidfarah2003:
## Mechanism

The AG-UI egress fence validated only `.events[].type`, answering "does this frame carry a forbidden event type" but not "does this frame carry tool output anywhere." A well-formed envelope with tool bytes on a sibling property of the event, on an unknown top-level property, or nested inside an allowed event's unknown field passed the fence and published with the bytes intact.

## What now happens

The fence validates the whole egress envelope against a closed schema:

- **Frame level**: only known keys (`kind`, `protocol`, `threadId`, `runId`, `epoch`, `seq`, `events`) are allowed. Any extra property is refused.
- **Event level**: each event is checked against the known keys for its `type` (sourced from the AG-UI 0.0.57 schemas), plus the `cotal` extension key. Any extra property is refused.

An envelope with an unknown or extra property at either level receives a new `extra-property` verdict from `frozenBodyEgressVerdict`, which names the path of the offending property. `AguiEmitter.attempt()` halts with reason `egress-extra-property` and never forwards a refused envelope.

Well-formed envelopes (including those with `cotal` metadata, `rawEvent`, optional schema keys like `outcome` and `parentRunId`, and JSON-round-tripped WAL recovery frames) continue to pass.

Refs #1432

### DIFF
--- a/.changeset/agui-egress-closed-schema.md
+++ b/.changeset/agui-egress-closed-schema.md
@@ -1,0 +1,4 @@
+---
+---
+
+AG-UI egress fence now validates the whole frame envelope against a closed schema, refusing any unknown or extra property at the frame level or inside individual events and naming the path. Previously, the fence only checked `.events[].type`, so tool bytes on a sibling property of the event, on an unknown top-level property, or nested inside an allowed event's unknown field passed untouched.
--- a/bin/smoke/ci-suites.d/7747ede8c90e1bc42e5b9d92fd23fba29b92376f584b4c878bb73825813b1c98.txt
+++ b/bin/smoke/ci-suites.d/7747ede8c90e1bc42e5b9d92fd23fba29b92376f584b4c878bb73825813b1c98.txt
@@ -1,0 +1,2 @@
+# The egress fence validates the whole frame envelope against a closed schema (#1432).
+smoke:agui-egress-closed-schema
--- a/extensions/connector-core/smoke/agui-egress-closed-schema.smoke.ts
+++ b/extensions/connector-core/smoke/agui-egress-closed-schema.smoke.ts
@@ -1,0 +1,364 @@
+/**
+ * THE EGRESS FENCE MUST REFUSE UNKNOWN PROPERTIES ON A FRAME ENVELOPE.
+ *
+ * Issue #1432: the AG-UI egress fence checked only `.events[].type`, so tool bytes on a sibling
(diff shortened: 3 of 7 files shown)


davidfarah2003:
## Gate status at `09d888fe396ba04fbabe758a371855adda863fb3`

The `gate-inventory` repair worked: that red is gone and `attribution` is SUCCESS. One new red at this exact head, and it is **not inherited** (origin/main `13f29e17a` has zero failing check-runs).

`smoke (shard 0/4)` FAILED at `pnpm smoke:mutation-fixtures`: **MUTATION FIXTURES FAILED (2 anchors need repair)**. Both are in the new `extensions/connector-core/smoke/mutations/agui-egress-closed-schema.json`:

- **`[0] T-SCHEMA-BYPASS`** — ANCHOR SPANS A COMMENT, 3 comment lines in `find`, including `// Closed-schema check: refuse any property not in the known set.` and `// This is the fix for #1432: the fence previously checked only`
- **`[1] T-ATTEMPT-EXTRA`** — ANCHOR SPANS A COMMENT, 1 comment line in `find`: `// Re-scan to name the path. This is the error path; the cost is negligible.`

The checker states the repair, quoted rather than paraphrased:

> Anchor on code. A `find` that spans prose is disarmed by anyone tidying prose, and a comment-only commit cannot announce that it disabled a guard. INSTEAD OF REACHING FOR THE COMMENT: widen to a multi-line, CODE-ONLY window until it is unique. A bare code line that occurs twice makes mutation-proof refuse an ambiguous target, and the nearest comment is always unique — which is why this keeps happening. More code lines resolve the ambiguity just as well, and nothing a docs commit can touch is load-bearing afterwards.

Re-anchor both mutations on a multi-line **code-only** window wide enough to be unique. Do not delete the comments, and do not weaken the mutations to dodge the check. Verify with `pnpm smoke:mutation-fixtures` green first-hand, plus the touched smokes with every mutation KILLED, then one push; fetch and rebase onto `13f29e17a` first.

Rest of the head: 30 check-runs, that one failure, 4 in flight, `attribution` SUCCESS.

Once terminal green, seat the three-family review panel per the lane brief. This PR is gated only when every check-run at the exact head is success or skipped, mergeStateStatus is CLEAN, and three family APPROVE lines sit at that exact head.

Issue #1422, opened 2026-09-09
ci-suites-freeze reports green locally: its frozen-list cell is dropped without BASE/HEAD, and its every-cell-ran sentinel is computed from the same condition


davidfarah2003:
`bin/smoke/ci-suites-freeze.smoke.ts` drops the only cell that inspects the commit unless `BASE` and
`HEAD` are set, and reports a confident green without it. So the local run a contributor does before
pushing is structurally incapable of catching the thing the suite exists to catch.

Found by a lane manager while grading a PR that had appended to the frozen list; confirmed
independently here.

## Measured

Same worktree, same commit, run twice:

```
no env:          "every cell ran - 8 expected"  ->  CI SUITES FREEZE SMOKE OK (9 passed, 0 failed)
BASE+HEAD set:   "every cell ran - 9 expected"  ->  CI SUITES FREEZE SMOKE OK (10 passed, 0 failed)
                 plus the cell "the committed frozen list added no suite names versus base"
```

The mechanism is `ci-suites-freeze.smoke.ts:122`:

```ts
const base = process.env.BASE;
const head = process.env.HEAD;
const EXPECTED = base && head ? 9 : 8;
```

## Why this is worse than an unguarded suite

The suite already carries a counted-inventory sentinel, and its own text names this exact hazard:

> `every cell ran - N expected, so a cell that stops existing is not mistaken for one that passed`

That guard cannot fire here, because `EXPECTED` is computed from the same `base && head` that removes
the cell. **The sentinel adapts to the absence it was written to detect.**

The result reads as more trustworthy than a suite with no guard at all: a contributor sees eight
green cells plus an explicit every-cell-ran check and reasonably concludes the suite ran completely.
Worth adding, from the lane manager who found it: all eight of those self-tests run against fixtures
the suite constructs itself, so every one can pass while the single cell with a real input is the one
that is absent.

## Consequence in practice

A contributor appends a suite name to the frozen `bin/smoke/ci-suites.txt`, runs
`pnpm smoke:ci-suites-freeze` locally, sees `OK (9 passed, 0 failed)`, pushes, and CI rejects it. The
local verification could not have told them otherwise.

## Suggested fix

Make the counted-inventory sentinel independent of the condition it is guarding. Either the
frozen-list cell always runs and fails loudly when `BASE`/`HEAD` are absent, or `EXPECTED` is a
constant and the cell is reported as SKIPPED rather than silently not existing — a skip is visible in
[...]

PR #1503, opened 2026-09-12
ci-suites-freeze committed-list cell reds when BASE/HEAD are unset

davidfarah2003:
## Summary

A local `pnpm smoke:ci-suites-freeze` run used to skip the only cell that compares two committed frozen lists whenever `BASE` and `HEAD` were unset, then shrink its inventory sentinel to match that omission, and exit green. A contributor could append a suite name to `bin/smoke/ci-suites.txt`, see `OK (9 passed, 0 failed)` plus `every cell ran`, and still be rejected later by the real scan.

The committed-list cell now always runs. Missing or zero `BASE`/`HEAD` reds that cell by name instead of deleting it. The inventory count is a constant of cells that actually executed. A new control suite drives freeze under that dropping condition and asserts the named red. The shard wrapper re-supplies `BASE` and `HEAD` after `env -i` so CI still compares two revisions.

Refs #1422

## Test plan

- [ ] `pnpm smoke:ci-suites-freeze` without `BASE`/`HEAD` is red and names `the committed frozen list added no suite names versus base`
- [ ] `BASE`/`HEAD` set on two revisions with no frozen-list append is green, 10 cells
- [ ] `pnpm smoke:ci-suites-freeze-drop` is green
- [ ] `pnpm mutation-proof --config bin/smoke/mutations/ci-suites-freeze.json` is 3/3 KILLED
- [ ] `pnpm check:shard-stability <main> <head>` is STABLE

### DIFF
--- a/.changeset/ci-suites-freeze-inventory.md
+++ b/.changeset/ci-suites-freeze-inventory.md
@@ -1,0 +1,4 @@
+---
+---
+
+The ci-suites-freeze committed-list cell always runs. Missing BASE/HEAD reds that cell by name instead of shrinking the inventory to a false green.
--- a/bin/smoke/ci-suites-freeze-drop.smoke.ts
+++ b/bin/smoke/ci-suites-freeze-drop.smoke.ts
@@ -1,0 +1,58 @@
+/**
+ * The freeze suite's committed-list cell must redden by name when BASE and HEAD are absent.
+ * A green inventory that adapted to that absence is the defect in #1422.
+ *
+ * Run: pnpm smoke:ci-suites-freeze-drop
+ */
+import { spawnSync } from "node:child_process";
+import { fileURLToPath } from "node:url";
+
+const ROOT = fileURLToPath(new URL("../..", import.meta.url));
+const CELL = "the committed frozen list added no suite names versus base";
+
+let pass = 0;
+let fail = 0;
+const check = (name: string, cond: boolean, extra?: unknown) => {
+  if (cond) {
+    pass++;
+    console.log(`  ✓ ${name}`);
+  } else {
+    fail++;
+    console.log(`  ✗ FAIL: ${name}`, extra ?? "");
+  }
+};
+
+const env = { ...process.env };
+delete env.BASE;
+delete env.HEAD;
+
+const ran = spawnSync("pnpm", ["smoke:ci-suites-freeze"], {
+  cwd: ROOT,
+  encoding: "utf8",
+  env,
+  timeout: 60_000,
+  maxBuffer: 2 * 1024 * 1024,
+});
+const out = `${ran.stdout ?? ""}${ran.stderr ?? ""}`;
+const named = out.includes(`FAIL: ${CELL}`);
+
+check(
+  "ci-suites-freeze is red when BASE and HEAD are unset",
+  ran.status !== 0 && !ran.error,
+  ran.error ? String(ran.error) : `status=${ran.status}`,
+);
+check(
+  "the red names the committed frozen-list cell",
+  named,
+  named ? undefined : out.slice(-800),
+);
+
+const EXPECTED = 2;
+check(
+  `every cell ran - ${EXPECTED} expected, so a cell that stops existing is not mistaken for one that passed`,
+  pass + fail === EXPECTED,
+  `${pass + fail} cells reported`,
+);
+
+console.log(`CI SUITES FREEZE DROP SMOKE ${fail === 0 ? "OK" : "FAILED"} (${pass} passed, ${fail} failed)`);
+if (fail) process.exitCode = 1;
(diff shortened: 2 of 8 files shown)


davidfarah2003:
APPROVE at 5fbb97904ce2a30a19d8827b4fe2b3ead533b84b (head of this PR, verified identical to the graded worktree).

## What this change does

Two commits. First, `ci-suites-freeze` no longer shrinks its inventory when `BASE`/`HEAD` are absent: the committed-list cell always runs and reds by name ("BASE and HEAD were unset, so the committed frozen list was not compared"), with `EXPECTED` pinned at 9 instead of deriving 8-or-9 from the same predicate that gated the cell. The shard wrapper re-supplies `BASE`/`HEAD` after `env -i` (from `HEAD`/`GITHUB_SHA` and the event payload's `pull_request.base.sha`, falling back to `HEAD^` and a depth-1 fetch) so CI still reaches the real two-revision comparison. Second, the new `ci-suites-freeze-drop` control spawns the freeze suite with `BASE`/`HEAD` deleted and asserts the red names that exact cell; it strips `COTAL_` keys from the child env copy before doing so, satisfying the `suite-ambient-env` census. The mutations config gains the new suite plus two mutations: restoring the old omit-and-shrink shape, and passing the unset cell outright; the shard-stability verifier hash pin is updated to the new wrapper bytes (verified: `sha256sum` matches), and a changeset and a frozen-suite fragment entry ship with it.

## What I ran (all in the graded worktree at this SHA)

- `pnpm smoke:ci-suites-freeze-drop` — OK (3 cells).
- `env -u BASE -u HEAD pnpm smoke:ci-suites-freeze` — FAILED, exit 1, and the red names the committed frozen-list cell by name. This is the exact defect path from the issue.
- `BASE=<HEAD~1> HEAD=<HEAD> pnpm smoke:ci-suites-freeze` — OK (10 cells), so the comparison path still passes.
- `pnpm mutation-proof --config bin/smoke/mutations/ci-suites-freeze.json` — all 3 mutations KILLED, each red and named (no INCONCLUSIVE, no FAILED-at, no parser gap).
- `pnpm smoke:suite-ambient-env` — PASS, which confirms the `COTAL_` strip in the new spawn.
- `pnpm check:shard-stability <HEAD~1> <HEAD>` — STABLE (565→565 suites, 0 shard changes).
- `pnpm typecheck` — clean.
- `pnpm changeset status` — valid; this PR's changeset is the only new one (others predate the merge-base).
- `sha256sum bin/smoke/verify-shard-inputs.sh` — matches the updated verifier pin in `shard-stability.ts`.

## What I read

The two commits' full diffs; `suite-ambient-env.smoke.ts` to confirm the strip satisfies its census rule (strip-before-delete ordering is immaterial to the rule, which keys on any `COTAL_`-filtering of the spread copy); the mutations JSON including both new mutations' find/replace blocks; the required-CI status reported terminal green at this SHA.

## Notes

The changeset is release-note-only (empty package list), consistent with existing repo precedent like `attribution-check.md`. `EXPECTED = 2` in the drop control with a cell-count assertion matches the repo's every-cell-ran convention. One observation, not blocking: the wrapper's base-rev fallback to `HEAD^` means a genuinely-squashed single-commit merge into main would compare the merge commit against itself rather than red — acceptable for the CI event-path shape it serves.

davidfarah2003:
APPROVE 5fbb97904ce2a30a19d8827b4fe2b3ead533b84b

RAN
- `git rev-parse HEAD` in this worktree is `5fbb97904ce2a30a19d8827b4fe2b3ead533b84b`.
- `pnpm smoke:ci-suites-freeze` with BASE and HEAD unset exits 1. Output includes `FAIL: the committed frozen list added no suite names versus base` and `every cell ran - 9 expected`.
- `pnpm smoke:ci-suites-freeze-drop` exits 0: `CI SUITES FREEZE DROP SMOKE OK (3 passed, 0 failed)`. The child freeze run is red by that same cell name.
- A child spawned with the same copy/strip/unset construction as freeze-drop had parent `COTAL_REVIEW_LEAK` set and child `COTAL_*` / BASE / HEAD keys empty.
- `pnpm smoke:suite-ambient-env` passes and reports `bin/smoke/ci-suites-freeze-drop.smoke.ts - strips COTAL_ before the spread`.
- With BASE at HEAD~1 and HEAD at this SHA, freeze is `OK (10 passed, 0 failed)` and the committed-list cell is green.

READ
- The committed-list cell always runs. Missing or zero BASE/HEAD reds it by name. `EXPECTED` is the constant 9, so the inventory does not shrink when those vars are absent.
- freeze-drop copies `process.env`, deletes keys that start with `COTAL_`, then unsets BASE and HEAD before spawning `pnpm smoke:ci-suites-freeze`.

This grades that SHA only.

Issue #1205, opened 2026-09-02
attach: refuses on an open-mode mesh because the session redeem needs a seed open mode never has


Lanzelot1:
## What happens

On a mesh started with `cotal up --open --detach`, `cotal attach` refuses every seat:

```
❯ cotal attach --name console-implementer
✗ mesh attach needs this space's local seed to redeem the session grant, and the mesh resolved for "main" does not provide one.
  broker nats://127.0.0.1:4222
  resolved root /Users/user/Projects/cotal.ai
  A static-auth mesh keeps this space's seed under <root>/.cotal/auth. That root is what this command resolved and connected with, so if it is not the checkout holding this space's auth, re-register the mesh with `cotal meshes add main --server nats://127.0.0.1:4222 --root <dir>`.
```

The resolved root is the registered one and it is correct. The mesh is in open mode, so there is no seed under `.cotal/auth` by design (the directory holds only `manager-instance.<key>.json`). The session redeem needs a seed, open mode never has one, so attach cannot work on an open mesh at all.

## Why the message makes it worse

The error assumes a static-auth mesh and tells the operator to re-register the root. Following it leads nowhere: the root is already right, and `cotal meshes add main --server ... --root <dir>` with the same root changes nothing. Nothing in the message, or in `docs/cli.md`, says that open mode is not supported by attach, and nothing points at what does work (the manager's browser console, or `cotal input`).

## Steps

1. `cotal up --open --detach` in a fresh folder.
2. `cotal spawn default --detach --name x`.
3. `cotal attach --name x`.
4. `cotal meshes` shows the mesh registered as mode `open` with exactly the root the error names.

## Expected

One of:

- attach works on an open mesh with a bare session redeem, the way every other command connects bare in open mode; or
- attach refuses with "attach is not supported on an open mesh" and names the alternatives (the manager console URL from the manager log, `cotal input --name <seat>`).

Either way, the `--root` re-register hint should only appear when the resolved root actually lacks auth material that another root holds.

## Environment

cotal-ai 0.38.0, macOS, broker started by `cotal up --open --detach`, manager `pty` runtime.

davidfarah2003:
[...]

PR #1514, opened 2026-09-12
fix(cli): attach on an open-mode mesh, which never has a local seed

davidfarah2003:
> **SUPERSEDED FIGURES.** This body was written at an earlier head and frozen once CI began running against it, so it is left intact rather than rewritten. **Every count below is stale.** Corrected figures, all re-measured on the current head:
>
> | this body says | the tree says |
> |---|---|
> | seed census: **five** | **seven** (six deciding sites, plus the mode derivation in `localTarget`) |
> | `attachSessionMaterial` has **three accepting branches** | **two** accepting (`bare`, `mint`) against three refusing; four accepting branches across both graded decisions, ten cells |
> | suite **20 passed, 0 failed** | **30 passed, 0 failed** |
> | pre-fix control **8 passed, 12 failed** | **11 passed, 19 failed** |
> | mutations **7 of 7 KILLED** | **10 of 10 KILLED** |
> | branch-table cells **six (three accepting, three refusing)** | see the branch count above |
>
> The pinned comment is the figures of record; read it before this section. Four review rounds moved these numbers, and three of the corrections were to claims that overstated the coverage.

## What was wrong

On a mesh started with `cotal up --open`, `cotal attach` refused every seat:

```
✗ mesh attach needs this space's local seed to redeem the session grant, and the mesh resolved for "main" does not provide one.
```

An open-mode mesh has no local seed **by design**. There is no credential system on it at all, so
there is nothing to be missing. The redeem path minted a session-scoped caller credential from that
seed and read its absence as "the seed is gone", which is true on a static-auth mesh and meaningless
on an open one. The refusal therefore fired on exactly the configuration the feature exists to
serve, and its remedy told the operator to re-register a root the mesh had already resolved
correctly, which changes nothing.

## What happens now

Which contract redeems a grant is the **recorded mesh mode**, decided once and carried as a value.
An open mesh redeems over the same bare connection the control round trip already used, and mints
nothing. A static-auth mesh still mints from the seed at the root the mesh resolved to. A
static-auth mesh whose seed is genuinely absent still refuses, and now names restore-at-checkout
instead of a re-registration that would not help. A user-auth mesh still refuses loud, because
two-step user-mode redemption is not wired.

The seed is never invented or synthesised for an open mesh, and the seed requirement on a sealed
mesh is never relaxed.

## Every place the attach path reaches for a seed, and how each is covered

There are **five**. They are listed because a fix that covers four of them is a fix that comes back.

1. **The redeem decision** — `attachSessionMaterial` in `implementations/cli/src/commands/agents.ts`.
   Was `if (!spaceAuth) refuse`. Now an **exhaustive switch on the recorded mode**, with `user` named
   rather than reached by fallthrough and an unrecorded mode taking the conservative sealed arm.
   Covered by six branch-table cells (three accepting, three refusing) and by mutations 1, 2, 3, 4.
2. **The session dial** — the `dialerFor(...)` connect on the redeem path. Was
   `standaloneConnectOpts(creds ? … : …)`, a nullable-credential test at the call site. Now
   `redeemConnectOpts(link)`. Covered by the live open attach and by mutations 1 and 7.
3. **The hand-back dial** — `releaseAbandonedSession`, which returns a manager slot after a link
   dies. It had the same nullable test, so a future edit could have diverged the two. Both dials now
   go through the single `redeemConnectOpts` seam, which is the only place a `RedeemLink` becomes
   connect options. Covered by `smoke:attach-reconnect` and `smoke:attach-stdin`, whose anchors are
   re-pointed at the new shape, and by mutation 5.
4. **The connect-time mode check** — `connectOrThrow` in `packages/workspace/src/connect.ts`. A
   registered `auth` mesh whose seed is missing must not be re-classified as open by a credless
   probe. Covered by a live sealed-broker cell and by mutation 6.
5. **The carry-forward** — `resolveControlTarget` in `packages/workspace/src/control-target.ts`,
   which propagates the recorded mode out of the resolve. This is the seam through which a future
   caller would silently reintroduce the requirement: drop the mode and every open mesh takes the
   sealed arm again. Covered by mutation 7, which deletes exactly that propagation and reds the live
   open attach.

The structural part is that `Established` and `Abandoned` no longer carry `creds?: string`. They
carry `RedeemLink`, a closed union of `{ mode: "bare" }` and `{ mode: "session-caller"; creds }`.
An optional credential invites one question at every call site — "is the seed there?" — and
answering it with `!creds` **is** this bug: on an open mesh absence means bare, on a sealed mesh
absence means refuse, and one nullable field cannot carry both meanings. The compiler now makes
every consumer read the mode.

## Parser/validator branch accounting

`attachSessionMaterial` has **three accepting branches** (open → bare, sealed-with-seed → mint, and
the transport arm of `redeemConnectOpts` which accepts both link modes). It has **three refusing
cases paired to them**, each differing from its partner only by context:

| accepting | refusing partner | what differs |
|---|---|---|
| open, no seed → bare | sealed, no seed → refuse | only the mode; seed absent in both |
| open, seed present on disk → bare | user, seed present → refuse | only the mode; seed present in both |
| sealed, seed present → mint | unrecorded mode, no seed → refuse | only the mode |

Plus three connect-option cells: a bare link must carry no credential, a session-caller link must
carry the one it was handed, and the TLS requirement must survive both arms.

## Proof

Run first-hand on a clone at `origin/main` `275dea4e111f565daa232cf748ab5ab0597e87fa`.

- `pnpm smoke:attach-open-mode` — **20 passed, 0 failed**, against a real open nats-server, a real
  manager, a real supervised seat and the real binary, plus a second real **authed** broker for the
  sealed half.
- **Pre-fix control.** The same suite, unchanged, run against the product files as they stand on
  `origin/main`: **8 passed, 12 failed**, and the live open cell reds on the issue's own sentence
  (`mesh attach needs this space's local seed …`). A probe that only exercised the fixed shape would
  have passed before the fix and proved nothing.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/attach-open-mode.json` — **7 of 7
  KILLED**, each on its own named cell. Mutation 4 is the band-aid this change must not be: it
  rescues a seedless sealed mesh with the open path, and it is killed by the sealed-refusal cell.
  That is how the refusing cell is shown capable of failing **on this tree**, not merely somewhere.
- `pnpm typecheck` — clean, including `tsconfig.smoke.json`.
- `smoke:attach-auth-root` 26/26 and `smoke:attach-stdin` 83/83, the two suites whose mutation
  anchors moved under this change; both re-anchored and re-run rather than merely re-matched.
- `smoke:manager-stop-spare-guard`, `smoke:mutation-fixtures`, `smoke:ci-suites-freeze`,
  `smoke:ci-fragments`, `smoke:ci-declarations` — all green.

Refs #1205


davidfarah2003:
Reproduced the two CI reds locally at head `fa5013ab92212cf9621c7a29d4217bd84eca2ae8`, deterministic across 3 consecutive runs (`6 passed, 2 failed` every time, same two cases). These are not load flakes.

```
pnpm smoke:attach-open-mode
  ✗ FAIL: open path is not selectable for an authenticated registry entry
        "✗ no mesh running at nats://127.0.0.1:41657 (stale registry entry - removed) - run `cotal up`"
  ✗ FAIL: open-without-seed attaches
        "✗ mesh attach needs this space's local seed to redeem the session grant,
           and the mesh resolved for "openattach" does not provide one."
```

**The second failure is this PR's own headline claim.** The test named `open-without-seed attaches` is the one asserting the behavior in the title, and it is failing on the path the change is meant to enable, so the fix is not yet doing what the PR says.

The first failure looks like a distinct problem in the fixture rather than in the product code: the assertion expects the open path to be *rejected* for an authenticated entry, but the run tears the broker down first (`stale registry entry - removed`), so the case passes/fails on broker lifecycle rather than on path selection. Worth separating those two, since as written that case could go green for the wrong reason once the broker ordering is fixed.

Attribution, with a control, since a red alone does not establish whose fault it is:

- All six sibling PRs in the queue (#1501, #1502, #1503, #1504, #1506, #1513) run these same check names in this same hour and are green on `unit`, `windows-ok`, and `Windows / required`. So the infrastructure is healthy and these reds are this branch's own.
- My first control attempt was invalid and I am recording it: I checked `main` and saw `0 red`, but `main`'s push workflow does not run `unit` or `smoke (shard 0/4)` at all (it runs 14 checks, none of them these). A green main was therefore no evidence at all. The sibling-PR comparison above is the control that actually discriminates.
- My local typechecker also had to be repaired before I trusted it: the first control run on plain `origin/main` reported an error (`SecretStoreIdentity` missing from `@cotal-ai/core`) from a stale build. After `tsc -b packages/core`, control on main is `0 errors` and this branch is likewise `0 errors` in `packages/workspace`. So there is **no typecheck defect here**; the failure is purely behavioral.

One structural note: this branch is 11 commits behind `main`, and `git merge-tree` against current `main` gives **178 files differing** between the tested tree and the tree that would actually land. Whatever CI says here is not grading the merged result. Rebasing onto `dd6fea0a1` before the next CI round will make the run describe the artifact that ships.

The change itself is well-scoped (11 files, changeset present, two mutation configs, docs updated). The two failures look fixable without redesign.

davidfarah2003:
Two review findings landed, both reproduced first-hand before acting on them. Both were right, and both are now fixed in code rather than in prose.

**1. A sixth seed reach, not five.** `attach` resolves seat locality before it redeems anything (`attach` → `pinForTarget` → `locateSeat` → `scatterManager` → `askManagerScatterEp`), and that site decided open-vs-sealed with `!auth.creds && !auth.bearer` — the same inference from a nullable credential that this PR exists to remove. My enumeration missed it because I traced the redeem and stopped there.

It is safe today, and it is not safe by accident: that answer only selects whether a frozen instance's liveness is **probed**, so a wrong answer downgrades a row to `not-probed` and can never refuse an attach. Documenting that would have been the band-aid. Instead it is extracted into an exported `scatterProbeMaterial(auth, spaceAuth)` returning `bare` / `mint` / `as-is`, so a caller that ever turns this answer into a refusal has to pass through a function whose whole purpose is to say why it must not. It is graded in both directions:

- `ACCEPT` an open mesh probes bare, minting nothing → mutation 8 removes the bare arm, KILLED.
- `REFUSE` a sealed mesh with no seed **degrades** rather than refusing → mutation 9 reclassifies it as open at that site, KILLED. That is the dangerous direction: a refusal there would reintroduce #1205 one call before the redeem.

**2. The band-aid mutation's attribution was imprecise, and the correction is worth stating.** Applied alone, mutation 4 reddens the **in-process** sealed-refusal cell, not the live one, because `connectOrThrow`'s recorded-mode check (mutation 6's target) refuses a seedless sealed mesh one layer earlier and the live attach never reaches the redeem decision. Removing both guards together does redden the live sealed cell (measured 16 passed / 4 failed). So the sealed refusal is defended in depth by two independent guards and each is killed by a cell that names it. The mutation's note now says exactly that instead of implying a single live kill.

**Corrected accounting: six seed reaches, nine mutations, all KILLED.** The redeem decision, the session dial, the abandoned-session hand-back, the connect-time mode check, the mode carry-forward, and the scatter probe.

Re-measured at `681ff22261a20f85795dede7396bd2735a3ad473`, on a clone at fresh `origin/main`:

- `pnpm smoke:attach-open-mode` — **25 passed, 0 failed**, against a real open broker and a second real authed broker.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/attach-open-mode.json` — **9 of 9 KILLED**, each on its own named cell.
- Pre-fix control, now reverting all four product files to `origin/main` — **8 passed, 17 failed**, the live cell reding on the issue's own sentence.
- `pnpm typecheck` — clean, including the smoke project.

**A third finding landed after this comment was first posted, and it was also right.** Two numbers in the PR body were wrong in the PR's own favour, so both are corrected here. The body itself is frozen because CI is running against it; this comment is the correction of record.

**Correction 1 — the branch accounting was inflated.** The body's table said "four accepting branches and four paired refusing cases". Counted properly, per BRANCH of the decision rather than per cell, the two graded decisions have **four accepting branches in total, not eight**:

- `attachSessionMaterial`: two accepting branches, `bare` (mode `open`) and `mint` (mode `auth` with a seed), against three refusing cases: `auth` without a seed, `user` with a seed present, and an unrecorded mode.
- `scatterProbeMaterial`: two accepting branches, `bare` (no creds, no bearer) and `mint` (seed plus creds), against one non-accepting `as-is` outcome reached from two directions.

The `bare` branch of each decision is exercised by two cells rather than one, because the presence of a seed on disk must not change its answer. That is why the cell count (ten) exceeds the branch count (four); quoting the cell count as a branch count overstated the coverage, and the corrected table is four accepting branches, five refusing-or-degrading cases, ten cells.

**Correction 2 — the seed census was drawn at a self-serving granularity.** "Six places reach for a seed" counted only the sites that can turn a missing seed into a REFUSAL. Other sites do load seed material and were silently excluded rather than named. Stated at both granularities:

- **Decides the redeem (six, each graded):** the redeem decision `attachSessionMaterial`; the session dial; the abandoned-session hand-back; `connectOrThrow`'s mode check; the mode carry-forward in `resolveControlTarget`; the scatter's liveness probe.
- **Merely loads, cannot refuse (named, deliberately untouched):** `targetFromEntry` and `localTarget` in `mesh-target.ts`, which read composed trust during resolution via `loadTrustOrThrow`; and the two `loadSpaceAuth` calls inside `divergentCwdAnchor`, a report-only comparison documented and tested never to throw. Neither can produce the #1205 refusal, and this change deliberately leaves both alone.

Both errors were found by review, not by me, and neither changes a measured result: the suite, the mutations, the control and the typecheck are as reported above. What changes is that the coverage claim is now the one the tree actually supports.



---

**A fourth finding landed, and it is the most substantive of the four: my own census above was wrong in this PR's favour, so it is corrected here rather than left standing.**

The census filed `localTarget` under "merely loads, cannot refuse". That is false in a way that matters. `localTarget` (`packages/workspace/src/mesh-target.ts`) ends with `mode: auth ? "auth" : "open"`. It does not merely load a seed; it **synthesises the mode**, and that mode is the sole input to the redeem decision. The resolver therefore settles the redeem contract one step before the decision runs, which puts it squarely in the deciding tier.

The consequence was a real, ungraded behaviour change. Driving the actual resolver with the registry isolated behind an empty `meshes.json`:

- a genuinely unrecorded, seedless root resolves to `mode="open"`, `source=flag-server`, `auth` absent, and the redeem returns **`bare`**;
- on `origin/main` that same input reached `const auth = t.spaceAuth; if (!auth) return fatal`.

So the behaviour changed, and **no cell measured it**. The cell that claimed this ground hand-built `mode: undefined`, a shape this resolver path never produces: true about the function, false about the product.

`localTarget` is byte-identical to `origin/main`. The defect was in my two claims, not in that code, and it is not edited to make a claim true. The `bare` outcome is correct on its own terms: an unregistered root with no auth material genuinely is an open mesh, `connectOrThrow`'s guard is `mode === "auth"` and so cannot fire, and an authed broker refuses a bare connect itself.

**Fixed by grading the resolver instead of a hand-built value.** The suite now looks `resolveMeshTarget` up by name and drives it, with `COTAL_HOME` pointed at a fixture holding an empty registry so a mesh recorded on the developer's machine cannot answer in place of the fixture. New accepting cell: an unregistered seedless root resolves to `"open"` and redeems `bare`, which is this issue's own scenario. Paired refusing partner, same root plus real trust material on disk: the resolver synthesises `"auth"` and the redeem mints from that material instead of going bare. The hand-built case is kept and relabelled for what it actually tests, that the decision takes the conservative arm for a mode it cannot read. A tenth mutation attacks the derivation itself, flipping the synthesised mode to another valid member, and is KILLED on the new cell by name.

Two errors of my own along the way, recorded because both produced confident wrong answers before being caught:

- the first version of that mutation set the mode to `undefined`, which only failed the build. A mutation that does not compile is a WRONG-RED and proves nothing; it was replaced with a valid member.
- the first adapter fed the resolver's `MeshTarget` straight into the decision. `MeshTarget.auth` is the space's **trust material**; `ControlTarget.auth` is the **connection's credentials**, with trust carried in `spaceAuth`. Feeding one to the other read a trust chain as a bearer token and made the seeded partner look like a by-name refusal, which I briefly took for a finding about the product. It was an artifact of my adapter. The two shapes are now typed separately in the suite and adapted the way `resolveControlTarget` adapts them.

**Corrected census, stated at both granularities.**

- **Decides the redeem (seven, each graded):** the redeem decision; the session dial; the abandoned-session hand-back; `connectOrThrow`'s mode check; the mode carry-forward in `resolveControlTarget`; the scatter's liveness probe; and **the mode derivation in `localTarget`**, which this correction adds.
- **Merely loads, cannot refuse (named, deliberately untouched):** `targetFromEntry`'s composed-trust read, and the two `loadSpaceAuth` calls inside `divergentCwdAnchor`, a report-only comparison tested never to throw.

Re-measured at `bb49726d88e5b20e48a3819e4cfe2b62a698efd3`, rebased onto `main` at `3e5b694cf`. The product files are unchanged from the previously reported head; the difference is entirely in the instrument.

- `pnpm smoke:attach-open-mode` - **30 passed, 0 failed**, against a real open broker and a second real authed broker.
- `node scripts/mutation-proof.mjs --config bin/smoke/mutations/attach-open-mode.json` - **10 of 10 KILLED**, each on its own named cell.
- Pre-fix control, reverting the four product files to `origin/main` - **11 passed, 19 failed**, with the new resolver cell among the reds and the live cell reding on the issue's own sentence.
- `pnpm typecheck` - clean, including the smoke project.

The body's stated figures predate all of this and are frozen because CI ran against them; the numbers in this comment are the ones of record. Four review findings, four corrections, three of them to claims I made rather than to code I wrote.

Issue #1318, opened 2026-09-05
delivery exits on 'broker unreachable' when the real condition is local CPU starvation, so Plane-3 removes itself exactly when the host is busy


davidfarah2003:
The delivery daemon exits when it decides the broker is unreachable. Tonight it made that decision twice while the broker was demonstrably alive, and the actual condition was local CPU starvation on the daemon's own host.

## What was observed

netcup, 2026-09-05, in a window from roughly 22:37Z to 22:50Z. Load reached 311 (12 cores) from concurrent typechecks and an unrelated off-mesh workload. The delivery daemon exited twice inside that window and an external watchdog restarted it each time. (The same watchdog log also shows two restarts at 04:05Z and 04:10Z, eighteen hours earlier and unrelated to this; naming the window because "twice today" and "twice during the incident" are different claims.)

The two exits:

```
✗ delivery: broker unreachable for >15s — exiting (coupled to the broker)
✗ delivery: lost the lease (wrong last sequence: 0) — exiting so the holder is single
```

The broker never went anywhere:

```
nats-server pid 2168717   uptime 631446s (7.3 days), continuous, no restart across either exit
TCP <broker-host>:4222    listening throughout
manager pid 1742087       up, and logging against that same broker 77s after one of the exits
```

So the daemon exited citing a server that was serving the whole time, and Plane-3 was down for roughly nine minutes across the two events.

## Mechanism

`implementations/delivery/src/delivery.ts:392-413`:

```ts
const BROKER_GONE_MS = Number(process.env.COTAL_DELIVERY_BROKER_GONE_MS) || 15_000;
let lastReachable = Date.now();
const brokerWatch = setInterval(() => {
  if (stopping) return;
  void isReachable(server, { creds: latestCreds, ...(tls ? { tls: true } : {}) })
    .then((ok) => {
      if (ok) { lastReachable = Date.now(); return; }
      if (Date.now() - lastReachable > BROKER_GONE_MS) { /* exit */ }
    })
    .catch(() => {});
}, 2000);
```

**Three different conditions produce the same signal, and only one of them is "the broker is gone".**

1. The broker is actually down. This is the case the check was written for.
2. **The event loop is starved, so the interval callback does not run.** `lastReachable` then ages with *no probe having failed and no evidence having been gathered*. The first callback that does run, on a host still under load, is also the one most likely to return `false`, and it finds the window already blown.
[...]

Issue #837, opened 2026-08-24
cli: stale ready delivery lease makes ensureDelivery false-green


davidfarah2003:
## Summary

`ensureDelivery` can return success from a stale `ready:true` delivery lease even when no delivery daemon or `ctl.delivery` responder exists. A newly launched child can lose the lease CAS and exit, while the parent sees the old ready record and reports the control-plane refresh as successful.

## Environment

- `cotal-ai 0.28.2`
- static-auth space with a delivery lease record present
- no live delivery process
- `cotal describe delivery` returns `deadline-exceeded`

## Live reproduction

After a delivery daemon exited without releasing its lease, a same-root refresh ran:

```sh
cotal up --space main --server nats://broker.cotal.ai:4222
```

The command exited 0. The child log then showed:

```text
✗ delivery: a live lease already exists for shard 0 — another delivery daemon is running. Not binding.
```

Post-condition:

```text
cotal status             -> delivery: stale pidfile
cotal describe delivery  -> deadline-exceeded: no describe reply
ps -p <delivery.pid>      -> no process
```

The lease survived repeated launch attempts well beyond `DELIVERY_LEASE_TTL_MS`; regardless of why expiry was delayed, the operational false green is that a ready KV record was accepted as service readiness without a responder proof.

## Source seam

`implementations/cli/src/lib/delivery-proc.ts::ensureDelivery`:

1. launches a new child whenever the local PID state is not alive;
2. calls `waitForDeliveryLease`;
3. treats a ready lease as sufficient readiness.

A pre-existing ready record can therefore satisfy the wait even when the new child immediately exits on the lease CAS. The comment says readiness means the responder is bound, but that implication only holds for the process that wrote the record; it does not prove that process remains alive or answers now.

## Expected behavior

A refresh must not report delivery healthy unless the registered `delivery` service answers its control surface. A stale ready lease is diagnostic state, not proof of a live responder.

## Acceptance

- Regression cell seeds a `ready:true` lease with no `ctl.delivery` responder and no live local daemon.
- A launched child that loses the lease cannot cause `ensureDelivery` to return a healthy result.
- The caller either safely waits for expiry/reacquires or refuses with an exact stale-lease/no-responder diagnosis; it never prints a green control-plane result.
- Live-holder control: a real daemon with a ready lease and responding service is reused without disruption.
- Slow-start control: a new daemon may take the documented bounded boot window without a false negative.
- The repair does not delete or take over a lease while its holder is provably live.

PR #1524, opened 2026-09-12
Tell local CPU starvation from broker loss before the delivery daemon exits

davidfarah2003:
## What broke

The delivery daemon is deliberately coupled to the broker: if the broker is gone it exits, because a daemon that outlives its broker loops reconnect attempts forever and nothing recovers it. That coupling is right. The way it decided "the broker is gone" was not. It read elapsed wall-clock time, and a clock cannot tell **"the broker is gone"** from **"this process did not get scheduled."**

On the shared host, 2026-09-05, load 311 on 12 cores: Plane-3 removed itself twice while `nats-server` had been up continuously for 7.3 days. It went away exactly when the host was busiest, which is when messages queue up and operators are trying to coordinate.

Three separate mechanisms, all invisible to a clock:

1. The 2s poll's interval doesn't fire under starvation, so the window ages with **no probe having failed**.
2. A process that can't get scheduled can't complete a handshake, so `isReachable` returns `false` from a **perfectly live server**.
3. A rejected probe was swallowed by `.catch(() => {})`, so it neither refreshed the window nor evaluated anything, so it silently aged the daemon toward an exit it had gathered no evidence for.

The lease path had the same defect in a second form. Any renew error meant exit, so the incident's `wrong last sequence: 0`, where the key had **expired with nobody else holding it**, was indistinguishable from a real takeover. It ended the only holder there was.

## Which signals now separate the two conditions

The exit predicate no longer reads a clock. It reads evidence the daemon actually gathered, from four independent sources:

| Signal | What it establishes | Where it comes from |
|---|---|---|
| **Interval-gap lag** | This process was off the CPU for *N* of the window's seconds, so the window did not age against the broker | Gap between consecutive firings of the daemon's own timer, credited back |
| **Completed, credible negatives** | A probe *ran*, and refused within the time the server was actually given | Counted; time that merely passed is not |
| **Own open transport** | A fresh handshake that can't complete to an address this daemon is *currently connected to* says little about the server, but only **within the backstop** | The endpoint's own `transport` event |
| **Deadline attribution** | *Who* the probe's deadline was enforced against, this process or the server | Two rules, below |

Judging a single probe needs two rules, because starvation reaches a probe in two shapes:

- **The loud shape.** The answer arrives so far past its own deadline that the deadline plainly wasn't enforced against the server. This is what the triage captured directly: a `false` at **2554ms against a 1000ms budget**, with the broker answering immediately either side of it. Caught with no instrumentation at all.
- **The quiet shape**, which is what a busy host actually produces most of the time, and which is invisible to a clock. The process issues a connect, is taken off the CPU, and its deadline timer fires the instant it's scheduled again. Elapsed wall-clock looks like an ordinary prompt timeout; the server was given a fraction of its second. So each probe now watches a short timer's own lateness for its duration, and that off-runqueue time is subtracted before the refusal is judged, because a refusal is only evidence about the server if the server had the time the deadline promised it.

That subtraction is bounded at **half the budget**, and the bound is not cosmetic. Ordinary Node timer jitter on an idle host is 10 to 30ms, so an "any measured deschedule disqualifies the refusal" rule read **10 of 10 blackholed probes as starvation on a healthy host**. Measured, not predicted; the reviewer who found it is the reason the share exists.

The probe budget is the transport's own, asked of `defaultProbeTimeoutMs` rather than hardcoded: `ws(s)` brokers get 5s and tcp gets 1s. A daemon judging a ws probe against a 1s budget would classify the broker's honest 3s answer as a refusal.

On a starvation diagnosis the daemon reports **DEGRADED once per episode, naming which condition it is**, keeps serving, and clears when the broker answers again.

## What is deliberately unchanged

The window constant is the same. What changed is that the quantity compared against it is evidence rather than elapsed time.

Three guarantees are preserved **by construction**, not by tuning, because the failure mode this repair could introduce, a daemon that outlives a dead broker, or two daemons serving one shard, is worse than the defect it removes:

- **A genuinely dead broker still ends the daemon, on the same window and just as fast.** A dead port refuses in ~1ms, so it never reaches its budget and stays a plain negative *however starved the host is*. The daemon's own connection closes too, so the transport signal goes false and every bound applies again. The condition that suspends the exit is the same condition that makes the exit unnecessary.
- **The hard backstop is decided FIRST and unconditionally.** Both reviewers arrived at this independently, and one measured it: on the earlier head a daemon stayed up 75s past a SIGSTOP'd broker, because `transportConnected` outranked the bound. That flag is cached client socket state: it can stay open for minutes after a silent broker death, so it now defers nothing once the bound is reached. Inside the backstop it still carries.
- **A failed renew is a question, not a verdict.** The daemon re-reads the key: another daemon holds it → exit; merely expired → re-acquire atomically (the create arbitrates); broker unreachable → stay alive. What it does **not** do is keep serving while it works that out. See below.

## The review findings, and the regression this branch introduced

Two independent review families returned REQUEST_CHANGES. Every finding was verified first-hand before being fixed; three are folded into the sections above (jitter share, ws budget, backstop ordering). Two more are worth stating on their own, because both are defects this branch *created*:

**The departing daemon deleted its replacement's lease.** `shutdown()` released the lease on every exit path, and release was an unconditional KV delete, so on precisely the paths that matter, where another daemon has taken the shard, the departing process removed the **new holder's** row and left the shard unheld. A quieter version of the same outage, reached through the lease path instead of the broker-watch path. Release is now a compare-and-swap on the revision this endpoint last owned, and the takeover paths carry no revision, so a departing daemon offers nothing to release.

**The daemon served the shard while deciding whether the shard was still its own.** A compare-and-swap keeps one lease **row**, not one **server**. Across the ownership re-read and the atomic create that may follow it, the daemon was still consuming the fan-out durable, still running the inbox reader, and still answering `ctl.delivery`, so a replacement that won the shard in that window shared both durables with it. Pre-fix there was no such window, because a failed renew went straight to shutdown. Treating that failure as a question is right; asking the question while still serving is not, and it trades an availability bug for a correctness one.

The daemon now unbinds fan-out, the reader and both control responders **before** asking, withdraws its readiness claim while quiet, and re-arms only on proof: its own row on a re-read (at the **broker's** revision), or a won create. The re-read half of that sentence was **not true when it was first written**. The ownership comparison could never return `held`, so the only re-arm that could fire was the won create. A reviewer caught it; it is fixed and graded below. `unknown` stays quiet, because not being able to ask is not permission to keep acting. `taken` exits quiesced, and a reconnect cannot silently resume mid-question. Those quiet periods are recoverable under the daemon's own power: the evidence that ends them is the same evidence that proves they were unnecessary, which is what cell H exists to refuse.

Grading that path surfaced two more defects that reading it had not:

- A re-acquired lease was **never flipped back to ready**, so a recovered daemon served correctly while every `ensureDelivery` waiter in the space timed out against a permanently not-ready row.
- The ownership re-read returned the daemon's **cached** revision, under a comment asserting the record carries none. It does. A renew whose write landed with only its reply lost left that token permanently one behind, refusing every later CAS over a sequence the daemon had moved itself, the same misreading in a second costume.

## A third defect, found by review, in the repair itself

The ownership re-read compared the row against the wrong identity, and it did so in two
independent ways. Both are fixed here, each with a pre-fix control that fails without the fix.

**It compared across identity namespaces, which made `held` unreachable.** The comparison used
`idFromCreds(creds)`, the bare connection nkey `U…`, while the endpoint rewrites `card.id` in its
constructor to the wire principal dot-form `local.U…` and stamps *that* into the row. So the daemon
re-read its own row, failed to recognise itself, took the takeover branch, and **exited naming
itself** as the daemon that stole its shard. That branch drops the revision, so the release then
freed nothing: the shard was left with a not-ready row and no process. `held` is the one survivable
reading, so making it unreachable turned every recoverable renew failure into a permanent
withdrawal, which is this issue's own outage, re-entering through the path built to prevent it.

**Comparing on the principal alone is also wrong, and looks correct.** The daemon's cred is a file
on disk that every restart re-reads, so a replacement daemon authenticates as the same nkey and
writes the same `holder`. A displaced daemon would read its **successor's** row as its own, keep
serving a shard it had lost, and CAS-release the live holder's row on the way out, which is the split the
lease exists to prevent. Not hypothetical: cells F and G already run two daemons from one creds
file, because that is what the product does, and the principal-only version puts **two servers on
`ctl.delivery`** (G12). The row now carries an `incarnation` minted per endpoint instance, and
ownership requires both halves. A row with no incarnation predates the field and cannot be proven
ours, which is the safe reading; all four writers go through `encodeLease`, so no path can write a
holder without one.

**The same confusion, a third time, in code this PR did not otherwise touch.** The fresh-launch
readiness wait asked for the bare nkey too, so it could never match: every `cotal up` ran its full
8s timeout and then reported a ready daemon as not-ready. It failed in the safe direction, which is
why it survived. The #837 guarantee it exists to enforce was not in force at all, but defeated by
a comparison that never fires rather than one that accepts anything. Three call sites have now
reached for `idFromCreds` and been wrong, so the fix is not a corrected expression at the call site:
`deliveryLeaseHolderFor` states the rule once, beside the record it describes.

## A fourth defect, exposed by the third one's fix, reached from the public CLI

Fixing the readiness wait made it correct, and correct made it **fast**: `cotal up` now returns the
instant the row appears instead of running its full 8s timeout. That exposed a window the broken
wait had been hiding. The daemon CAS-creates its lease row early in start-up, then binds Plane-3,
flips the row ready, and awaits the membership feed and the timer writer, and its signal handlers
were registered only after all of it. **SIGTERM in between took Node's default action**: immediate
death, no release, the row claiming the shard with no process behind it until the 30s bucket TTL
expired, after which the next `cotal up` was refused with *"a live lease already exists"* and the
shard was unservable by anyone. This issue's outage, with no starvation and no broker fault at all.

CI found it, as `smoke:up-tls-routes` → `delivery-refresh-keeps-tls`. **I bisected it rather than
retried it**, because a 50% failure rate makes one green run worthless as evidence: merge base
untouched 4/0 route runs; this branch 3/3; this branch with the holder wait neutralised 4/0; this
branch with handlers armed at the acquire 6/0. So the branch caused it, the holder fix was the
specific cause, and reverting that fix would have restored an always-false comparison to hide a real
window, a band-aid, refused. A reviewer's deterministic repro agreed independently: SIGTERM at the
readiness edge killed the daemon 6/6 with the row stranded, the same signal 1500ms later released
cleanly 2/2.

Handlers are now armed **the statement after the acquire**, the first instant a row bearing this
daemon's name exists. Three hypotheses were tested and **disproved** before this one: shutdown
ordering, a handler window below `shutdown`'s definition, and a stale-revision release. The second
produced cells T1-T5 that **passed on the unfixed tree twice**, so they graded nothing and were
**deleted** rather than shipped. The third is real and is fixed here (shutdown re-reads and releases
at the broker's revision, since `markDeliveryLeaseReady` moves the row and an ordinary interleaving
is enough to leave the cached token one behind), but it was **not** this route's cause, and it is
described as what it is.

The W cells took three attempts to grade anything, and the reason is recorded in the fixture. The
first polled with `readLease` and passed on the unfixed tree; instrumenting it showed `readLease`
dials a **fresh NATS connection per call**, so a loop written as 1ms actually polls every 32ms and
the row's first sighting was already `ready:true`. The window opened and closed between two polls.
A KV watch would take the pushed create, but the daemon's cred is granted `kv.get` and not
consumer-create, and so is the space provisioner's; both were refused by the broker. The cells now
poll down **one reused connection**, and **W2 records which edge was caught** rather than assuming
it. M19 pins the placement, which no assertion about a return value can defend.

## A fifth defect, found by review, in the fourth one's fix

The process-level guards added above do not fire on the path that ships. `runCli` awaits the
command inside its OWN try/catch, so a rejection out of start-up is a HANDLED error: the CLI prints
one line and exits 1, neither `uncaughtException` nor `unhandledRejection` ever runs, and the lease
row keeps claiming the shard for the rest of the bucket TTL. A reviewer reproduced it against a
real broker on the public `bin/cotal.ts deliver` path by pre-creating a conflicting fan-out durable
so `startPlane3` rejects past the acquire: exit 1, row still present, replacement refused with "a
live lease already exists". I reproduced it first-hand before changing anything, and the control is
in the table above.

The release now hangs off the REJECTION rather than off a process event: `runDelivery` is a thin
wrapper owning a try/catch, and the body publishes its releaser only once the shard is actually
ours, so before the acquire there is nothing to give back and the catch re-throws untouched. The
fault stays a fault, since the original error is re-thrown after the release and the exit code is
unchanged, which cell X3 grades directly. M20 pins the catch body, which the process guards cannot
defend. Cells X1 to X6 use the reviewer's stimulus, a real broker-side durable conflict rather than
an injected throw, so they exercise a failure the product can actually meet.

**M12 SURVIVED this round, and that is reported rather than papered over.** Once shutdown re-reads
and releases at the broker's revision, the end-to-end cell can no longer feel whether that delete is
conditional, a real coverage gap, not a defect that went away, since the CAS is still the last
thing between a departing daemon and its replacement's row. It is re-anchored on the seam where the
claim is still decidable: a release arguing a stale revision must leave the row alone.

Controls, measured rather than argued:

| tree | result |
| --- | --- |
| namespace bug present | R4, R5, R6, R7 red (75 passed, 4 failed) |
| principal-only ownership | R9 red **and** F5, F6, G9, G10, G12 red (74 passed, 7 failed) |
| launcher derives the bare nkey | derivation cell red, refusing control still green (17 passed, 1 failed) |
| all three fixed | 87, 91 and 22 passed, 0 failed |
| handlers armed only at the end of start-up | W4-W7 red (84 passed, 4 failed), and `delivery-refresh-keeps-tls` 2/3 route runs red |
| handlers armed at the acquire | 93, 91 and 22 passed, 0 failed; 6/6 route runs green, and 3/3 again on the merge with current main |
| only the process-level guards, no catch-reachable release | X4, X5, X6 red (91 passed, 3 failed): row stranded, replacement refused |
| the release reachable from an ordinary catch | 94 passed, 0 failed |

## Branches and refusing cases

**Two suites, 184 cells, 20 mutations, every one KILLED in a single clean run. Plus the lease suite's 22 cells and 3 mutations, and the launcher fixture's 3, also all KILLED. 26 of 26 mutations killed, zero survivors, each on its own named cell.**

- `smoke:delivery-watchdog-evidence`: **91 cells**, pure, no broker required. Sections A to N. **Every ACCEPTING branch has at least one REFUSING case that differs in exactly one input.** Section N grades the quiesce gate specifically, including N7, which asserts the *shape* (exactly one of the four lease readings admits serving) so a widening cannot pass by adjusting individual answers.
- `smoke:delivery-starvation`: **93 cells**, live, against a real `nats-server` and real spawned daemons. **A** full SIGSTOP with the broker alive; **B** broker actually killed, must exit *and name why*; **C** new connections blackholed while the standing socket survives, must **not** exit; **D** the same blackhole with the transport **cut as well**, must exit promptly; **E** a real daemon on a **~10% CPU duty cycle**; **F** a handover, where the loser must not delete the winner's row; **G** the arbitration window; **H** recovery from a quiesced state with no rival; **R** the survivable renew failure: one daemon, no rival, the row's own bytes written back so the revision moves while the holder stays byte-identical, which is the renew whose write landed with its reply lost, plus the refusing case of a successor's row; **W** an ORDINARY stop inside the start-up window, the only non-SIGKILL stimulus in the fixture; **X** a start-up failure the CALLER catches, stimulated by a real broker-side durable conflict rather than an injected throw.

Cell **G** is worth describing precisely, because the obvious claim is false. An overlap **exists and cannot be abolished**: the instant the replacement binds, the frozen incumbent is still bound and cannot act, since demanding zero overlap would demand action from a descheduled process, and the only way to pass it is to weaken the stimulus. What is under test is **how the overlap ends**. It is observed from the broker's own view: parked pull requests per durable, and the per-connection `ctl.delivery` subscription count from CONNZ, so "two daemons were bound at once" is a count of processes, not a self-report.

The window it grades is a couple of broker round trips wide, so it is **held still rather than chased**: SIGSTOP freezes the loser the instant it announces going quiet, the two-responder overlap is read while the replacement is bound and the loser is frozen, and the quiesced state is then sampled at leisure before SIGCONT. Freezing a process does not unbind it, because the broker's answer is about the connection, not about whether the process is scheduled, so this changes nothing about what the daemon did. The first version raced instead, won locally and lost on CI, and a re-run would have been exactly the kind of green this lane refuses.

**G7e is the discriminating cell, and it exists because a weaker version of it survived a mutation.** `shutdown()` unbinds via `ep.stop()` and only then exits, so "stopped serving while still alive" is satisfied by the **pre-fix** behaviour too: with quiescing disabled entirely, every other G cell stayed green. What only quiescing can do is stop serving **before the ownership question has been answered at all**.

**Pre-fix control.** On an isolated clone at `c8fe5a8c9`, source untouched, with only the suites copied in: **48 passed, 24 failed**, including C6-C9, E5-E6, F6-F8 and the new G/H cells, with the shipped line `✗ delivery: broker unreachable for >2s, exiting (coupled to the broker)` while the broker was verified up throughout. Post-fix: **72 passed, 0 failed**. The pure suite cannot load there at all, which is the honest reading of a module that does not exist yet.

Several cells exist **because mutations survived**. My first live cell passed pre-fix and proved nothing, since SIGSTOP alone never produces a completed-false probe on loopback. Cells D, E, L6, L7 and G7e each exist because a mutation slipped past the suite as it stood. The in-probe measurement itself was found that way: cell E put a real daemon on a duty cycle and it **exited**, which is how the quiet shape of starvation got found at all.

One suite-level constraint worth recording: cells G and H **reuse** earlier cells' spaces. Eight spaces exceed one test broker's storage (each reserves a 4 GiB artifact Object Store) and the suite failed to provision before a single cell ran. Both cells delete the lease row explicitly, which is what the per-space isolation was for.

## Also fixed: a suite that was grading nothing

`smoke:delivery-broker-coupling` was carried as untriaged debt. It spawned the daemon without a `$SYS` observer cred and from a cwd whose root walk climbed out of the repo, so the daemon **refused during startup**, and that refusal satisfied its own "exits when the broker is gone" assertion. It reported green while its subject had been dead since startup and had never once been coupled to the broker it was asked to outlive. Now provisioned, pinned to a scratch workspace, required to **name** the reason it exited rather than merely to exit, and gated in CI rather than exempt. `EXPECTED_EXEMPTIONS` 26 → 25.

Both new suites are gated in CI as well.

A later style pass replaced every em and en dash in this branch's added lines, including the daemon's own runtime log strings, and that pass silently broke two cells: `H6` and `R6` do not match a literal, they match across a gap, and the dash sat inside the traversed span, so the subject test `/serving shard \d+ again/` stayed true while the attribution clause went dead and the suite would have printed 98/0 with that cell grading nothing. Both matchers now accept either separator (`again[,]? `), verified against all three `why` values the call sites actually pass, in the comma form and the pre-rewrite dash form, so they survive a revert in either direction rather than being a one-way ratchet.

## Filed separately

The pre-observation expiry gap (a lease can expire before the daemon's first ownership observation), and the ambiguity of a shared delivery credential across two daemons in one space. Both were raised in review, both are real, and neither is this issue.

Refs #1318


davidfarah2003:
## Review Verdict: REQUEST_CHANGES

- **PR**: Cotal-AI/Cotal #1524 ("Tell local CPU starvation from broker loss before the delivery daemon exits")
- **HEAD Commit**: `78489a325f097e8e4900be326dd23f63ad752e70`
- **Merge Base**: `f574a0179305b46425df9850807d55d43b11e1a5` with origin/main
- **Census**: 42 commits from 1 author (enumerated via `pulls/1524/commits`)
- **Review Range**: `f574a0179305b46425df9850807d55d43b11e1a5..78489a325f097e8e4900be326dd23f63ad752e70` (ahead 42, behind 0)

---

### 1. M14 Mis-Anchor: G9 (Green) vs G7d (Red) and Execution Mechanism

Mutation fixture `M14` (`bin/smoke/mutations/delivery-starvation.json`) targets `packages/core/src/endpoint.ts`:

```typescript
// Find:
async quiescePlane3(): Promise<void> {
  this.plane3Quiesced = true;

// Replace:
async quiescePlane3(): Promise<void> {
  this.plane3Quiesced = true;
  if (Date.now() > 0) return;
```

`M14` declares:
```json
"expectRed": "G9 and it stopped serving BEFORE it concluded it had lost the shard, not as part of exiting",
"cell": "G9 and it stopped serving BEFORE it concluded it had lost the shard, not as part of exiting"
```

#### Mechanism: Why G9 Stays Green While G7d Reds

In `implementations/delivery/src/delivery.ts`:
```typescript
try {
  await ep.quiescePlane3();
  if (revision !== undefined) {
    try { revision = await ep.markDeliveryLeaseNotReady(shard, revision); }
    catch { /* the row may have moved on; the ownership read below is what decides */ }
  }
  noteQuiesce();
}
catch (e) { console.error(`! delivery: could not quiesce Plane-3 while checking the lease (${(e as Error).message})`); }
const reading = await readOwnLease();
switch (leaseAction(reading)) {
  case "exit":
    console.error(`! delivery: lease is held by another daemon (${reading.holder ?? "unknown"} vs ours ${holder}) -- exiting so the holder is single`);
    return;
```

1. **Cell G9 stays GREEN**:
   `ep.quiescePlane3()` sets `this.plane3Quiesced = true` and returns immediately without error.
   `delivery.ts` continues directly to execute `noteQuiesce()`, which logs `! delivery: stopped serving shard` to stderr.
   Subsequently, `readOwnLease()` observes that another daemon holds the lease and logs `! delivery: lease is held by another daemon ... exiting so the holder is single`.
   Cell G9 evaluates transcript order in stderr:
   ```typescript
   const quiesceAt = incumbent.stderr.indexOf("stopped serving shard");
   const lostAt = incumbent.stderr.search(/taken shard|is held by/);
   check("G9 and it stopped serving BEFORE it concluded it had lost the shard, not as part of exiting",
     quiesceAt >= 0 && lostAt >= 0 && quiesceAt < lostAt, { quiesceAt, lostAt });
   ```
   Because `noteQuiesce()` executed sequentially before `readOwnLease()`, `quiesceAt < lostAt` remains true. G9 evaluates transcript order of log statements and stays **GREEN**.

2. **Cell G7d fails RED**:
   Because `ep.quiescePlane3()` returned early, Plane-3 subscriptions and control responders were never unbound from the broker. Active Plane-3 subscriptions remain at 2 and do not drop to 1 while the loser process is alive.
   Cell G7d evaluates broker state:
   ```typescript
   check("G7d the overlap ENDED while the loser was still alive, not by the loser exiting",
     overlapEndedAlive, { peakBound, answeredSubs, loserExited: incumbent.exited });
   ```
   `overlapEndedAlive` fails because subscriptions remained at 2 until process exit. G7d is the cell that catches the defect.

`M14` is mis-anchored: the fixture named a cell evaluating log output order rather than broker-level quiescence.

---

### 2. Control Kills: M21 and M22

To prove the test suites actively execute and fail on specific structural mutations, two control mutations were tested against `implementations/delivery/smoke/delivery-watchdog-evidence.smoke.ts`:

- **M21** (`implementations/delivery/src/watchdog.ts`):
  - Find: `lo = Math.min(lo, s); hi = Math.max(hi, e);`
  - Replace: `merged.push([s, e]);`
  - Result: **KILLED** on named cell `I3b` ("I3b one stall seen by BOTH instruments is charged ONCE, on the daemon's own timeline").
- **M22** (`implementations/delivery/src/watchdog.ts`):
  - Find: `if (e < lo || s > hi) { merged.push([s, e]); continue; }`
  - Replace: `if (false) { merged.push([s, e]); continue; }`
  - Result: **KILLED** on named cell `H11` ("H11 disjoint stalls across a window accumulate").

These confirmed kills establish that the test harness reaches the code under test and correctly detects mutations when cells are properly anchored.

---

### 3. G7e Prose Occurrences vs Zero Defining Lines

In `implementations/delivery/smoke/delivery-starvation.smoke.ts` (1430 lines), `G7e` appears in 7 prose comment lines:

- **Line 204**: `// line before the poll came round, and cell G7e reddened on a state that had genuinely existed.`
- **Line 838**: `// back afterwards. A daemon that only stopped serving by exiting fails G7d and G7e; a daemon that`
- **Line 860**: `// the sampled reading below meaningful. But nothing PASSES on it any more. G7e now grades the`
- **Line 870**: `// line, so the trigger never fires, the loop below falls through on its deadline, and G7e reds`
- **Line 889**: `// coin flip as the old G7e: losing the race does not mean the daemon failed to quiesce, only that`
- **Line 957**: `// G7e, RE-MEASURED AFTER A CI RED. The claim is unchanged; the way it is observed is not.`
- **Line 977**: `// G7e IS GONE, AND THE REASON MATTERS MORE THAN THE DELETION.`

**Lines defining a cell named G7e (check/it/test)**: **0**.

The comments contradict each other: lines 860 and 870 describe G7e in the present tense as actively grading and reddening, while line 977 notes that G7e is deleted. The file presents conflicting descriptions of whether G7e exists depending on which section is read.

---

### 4. Evaluation of Hook Comment at Lines 874-876

In `implementations/delivery/smoke/delivery-starvation.smoke.ts`, lines 874-876 read:

```typescript
    // Synchronous, inside the data event: SIGSTOP lands before this process yields, so the
    // daemon cannot get another slice in which to answer the ownership question.
    // A 20ms poll could not keep up under load -- measured, and it is why this hook exists at all.
```

This comment is graded as a **justification claim** rather than a factual architectural description. Ending with "measured, and it is why this hook exists at all" appeals to a settled experimental conclusion to justify test apparatus behavior. Asserting in prose that an approach was measured and necessary does not constitute verification that the hook is sound or that alternative poll mechanisms fail under current architectural conditions.

---

### 5. Suite Baselines and Denominators

- `implementations/delivery/smoke/delivery-watchdog-evidence.smoke.ts`:
  **106 passed, 0 failed (106/0)**
- `implementations/delivery/smoke/delivery-starvation.smoke.ts`:
  **99 passed, 0 failed (99/0)**

---

### 6. Scope of What Was NOT Measured

1. **Full live execution of delivery-starvation under M14 on a contended shared host**: Because machine load on the shared test host was elevated (load15 > 33 vs gate limit 25), live starvation smoke runs produce uninterpretable timing and lease-renewal failures (such as cell A8 failing when the test harness is descheduled). M14 was evaluated through static trace analysis of sequential call sites and verified broker-level observables rather than an uninterpretable contaminated live run.
2. **Multi-host network partitions**: The suite exercises in-process NATS servers and simulated local scheduler/CPU starvation (via SIGSTOP/SIGCONT and LoopLagMeter interval union accounting), not multi-host network interface drops or cross-DC routing splits.
3. **Hardware starvation exceeding the 40s backstop**: Starvation intervals exceeding `backstopMs` (40s) trigger the unconditional broker-loss exit by design and were not evaluated for indefinite survival.

davidfarah2003:
## Re-grade at `94d1bde848edba24dfd12647d233c50375a76d99`

Head verified two ways (board `headRefOid` and paginated `commits[-1]`, full 40-character equality), and confirmed a strict follow-on: the new commit's parent is exactly `78489a325f09`, the head I graded. **Every line number in my previous comment is stale.** Both files changed.

**Both findings I raised are addressed at source level. Reporting that first, because they were the blocking ones.**

### Finding 1, the M14 mis-anchor: fixed, and fixed at the anchor rather than around it

```
78489a325f09   expectRed: "G9 and it stopped serving BEFORE it concluded it had lost the shard ..."
94d1bde848ed   expectRed: "G7d the overlap ENDED while the loser was still alive ..."
```

The fixture now names the cell that actually reds. The commit subject says it plainly: "anchor M14 on what the daemon did, not on what it said."

### Finding 2, the dangling `G7e` citations: fixed

`G7e` rows in the suite went from 7 to 2, and both survivors are now about its removal rather than references to a live cell:

```
:891  // coin flip as the old G7e: losing the race does not mean the daemon failed to quiesce ...
:959  // G7e IS GONE, AND THE REASON MATTERS MORE THAN THE DELETION.
```

The prose no longer cites a cell that does not exist. `G7d` (3 rows) and `G9` (6 rows) are present as the known-positive control, so this is not a reader that stopped matching.

### One new problem, created by the fix itself

**The fixture and the source now contradict each other about what M14 proves, inside the same commit.**

The fixture says M14 reds `G7d`. Fifteen lines below that cell, the source says:

```
:954  // THE DISCRIMINATING ASSERTION. Unbinding inside `shutdown()` also happens while the process is
:955  // alive, so G7d alone is satisfied by the pre-fix behaviour - verified by disabling the quiesce
:956  // call and watching every G cell stay green.
```

And M14's mutation is precisely a disabled quiesce:

```json
"replace": "async quiescePlane3(): Promise<void> {\n  this.plane3Quiesced = true;\n  if (Date.now() > 0) return;"
```

**So the fixture asserts that neutering `quiescePlane3` reds `G7d`, while the comment reports that neutering `quiescePlane3` leaves every G cell green, verified by running it.** Both cannot be true.

I want to name the one difference rather than overstate this, because it is the difference that decides it. M14 **still sets `plane3Quiesced = true`** before returning; a disabled call site would leave that flag false. If any cell keys off the flag rather than off the observable unbind, the two are not the same experiment and the contradiction dissolves. **I have not run M14, so I cannot tell you which it is.** What I can say is that the fixture's `expectRed` is now a claim the file argues against in prose, and one run settles it.

If M14 does red `G7d`, the comment at `:954-956` is stale and should be corrected, because it tells a future reader that `G7d` is not discriminating. If M14 does not red `G7d`, the anchor has moved from one wrong cell to another and the fixture is still mis-anchored, just less obviously.

**This is the same class of defect as the original finding, which is why it is worth the paragraph: an assertion about which cell reds, not verified against a run.** The previous anchor named `G9` and reded `G7d`. The repair is right about where the red lands and now disagrees with the file's own account of whether that red means anything.

### Position

**REQUEST_CHANGES stands, on the single item above.** The two findings that blocked before are genuinely closed and I am not re-raising them. Run M14 at this head and either correct `:954-956` or re-anchor, and post the row the run produces rather than the conclusion.

PR #1525, opened 2026-09-12
fix(endpoint): refuse a creds generation there is nothing to renew from

### LINKED ISSUE #1523: an expired credential from a creds source re-arms the renewal timer at its 1s floor, retrying a dead source once a second
Found while reviewing the fix for #1411. **Pre-existing: measured identical before and after that change, so it is not a regression from it.**

`fetchFreshCreds` caches whatever the source returns and arms the next renewal from it:

```
this.currentCreds = creds;
this.armCredsRefresh(credsRenewalDelayMs(creds));
```

`credsRenewalDelayMs` returns `iat + 0.75*(exp-iat) - now`, which is **negative** for a credential that is already expired. `armCredsRefresh` clamps with `Math.max(1_000, delayMs)`. So a source returning already-expired material does not simply cache something stale, it schedules the next renewal at the 1s floor, which fetches again, which is expired again, which re-arms at 1s.

Measured, with a source returning a live credential on the first call and an already-expired one thereafter, counting source reads over 12 seconds:

```
fixed tree (#1411 branch):                    readsAtStart=1, readsAfter12s=6, delta=5
unfixed tree (main, product source untouched): readsAtStart=1, readsAfter12s=6, delta=5
```

Identical, which is what establishes this as pre-existing rather than introduced by #1411.

The wire is not affected: after #1411 every presentation routes through a checkpoint, so an expired cached credential can never reach the broker. The cost is load. A credential store having a bad day gets hammered once a second per endpoint, and `CREDS_RETRY_MS` (60s) — the backoff that exists for exactly this situation — is bypassed, because this path is not a *failed* fetch. It is a *successful* fetch of dead material, which no current branch treats as a failure.

Two candidate fixes, and the choice is a behavioural call rather than an obvious bug fix, which is why this is its own issue:

1. Treat a fetch that returns an already-expired credential as a failed renewal: emit the recoverable warning and back off to `CREDS_RETRY_MS`. Arguably the most honest reading, since the source did not deliver something usable.
2. Clamp the armed delay to `CREDS_RETRY_MS` whenever the computed delay is non-positive. Narrower, but leaves a successful-looking fetch of dead material unremarked.

Either wants a pre-fix control that reproduces the 1s cadence and fails on today's tree, plus a mutation fixture proving the assertion discriminates. Worth checking whether the bearer path (`bearerExpiryMs` / `BEARER_REFRESH_MARGIN_MS`) has the same shape.

L4XB:
Fixes #1523 with option 1, and the reason it is not a behavioural coin-flip is that **this repo already chose** for the sibling path.

## The precedent

`membership-feed.ts`, `adoptRwCreds`:

```ts
// An UNCHANGED generation past its own renewal point is a MISSED remint (the renewal owner has not
// re-signed yet), not a candidate to adopt. Adopting it would recommit the same cred, floor the next
// timer to 1s (delay <= 0), and reconnect EVERY second for the JWT's remaining 25% of life.
if (candidate === currentRwCreds && delay <= 0)
  throw new Error("reloadRwCreds: the rw source still holds the previous generation past its renewal point ...");
```

pinned by `membership-rw-renewal.smoke.ts` scenario 5 ("an unchanged source past 75% must NOT 1s-churn"). The endpoint's `delivery.creds` seam is the same seam with the same floor and never got the rule, so the two renewal paths answered a dead source differently. This makes them answer identically.

## The change

`CotalEndpoint.assertRenewableGeneration(candidate, current, delayMs)` refuses a generation there is nothing to renew from, before either commit path assigns the cache:

- `fetchFreshCreds` (the passive-backstop / pre-connect fetch, the one the issue quotes);
- `adoptFreshCreds` (the prove-then-adopt transaction shared by the 75% timer and the explicit `reloadCreds`), at the same point the bounded-window validation already sits, so a refusal after a successful preflight stays a true no-op on the cache.

A refusal routes through each caller's **existing** failure posture, so nothing new had to be invented: `renewCredsOnTimer` emits one recoverable warning and arms `CREDS_RETRY_MS` (60s) while the last renewable generation stays live to its expiry, and `reloadCreds` answers structured with nothing adopted.

Two conditions, named separately:

1. `candidate === current && delayMs <= 0` — the missed remint, exactly the membership feed's rule.
2. `exp * 1000 <= Date.now()` — an already-expired generation, refused even when the source keeps returning a *different* one each call, which condition 1 cannot see. This is the shape the issue's own measurement used.

An initial fetch is unaffected by condition 1 (`current` is undefined, so a usable generation past its renewal point is still adopted and connects); an expired one now fails loud at `start()` instead of dialling with dead material and reading the store on a 1s loop.

## Reproduced live before the fix

Per AGENTS.md, the repro came first. New scenario in `standing-renewal-auth.smoke.ts`: a source pinned to one 12s generation, reads counted across the 2.8s window past its 75% point.

| tree | reads in the churn window |
| --- | --- |
| before | **4** (baseline 1, so 3 extra: one per second) |
| after | **1** |

With the fix reverted the scenario fails on its own assertion:

```
✗ FAIL: an unchanged source past 75% did NOT busy-loop (<=1 renewal read in the churn window) { churnReads: 4, churnBaseline: 1 }
✗ FAIL: the refusal is reported once and names the un-re-signed generation []
STANDING RENEWAL SMOKE FAILED ❌  (13 passed, 2 failed)
```

And `mutation-proof` kills it on the named assertion:

```
KILLED  packages/core/src/endpoint.ts: if (candidate === current && delayMs <= 0)
  red, and named: an unchanged source past 75% did NOT busy-loop · 13 marks (baseline 15)
```

The second assertion is deliberate: "did not busy-loop" alone would also pass if the timer never fired at all, so the warning count pins that the refusal is what stopped it.

## Verified

Run with `nats-server` v2.14.6 on PATH:

- `standing-renewal-auth.smoke.ts` — 15/15 (was 13/13; the two new cells are mine)
- `membership-rw-renewal.smoke.ts` — 28/28, so the sibling path is untouched
- `cred-lifetime-auth.smoke.ts` — 18 checks, `anchor-renewal.smoke.ts` — 17 cells
- `tsc --noEmit` on `packages/core` clean, `check:docs-voice` 40 pages passed

## The bearer path: yes, same shape

The issue asks. `refreshBearer` arms `bearerExpiryMs(bearer) - Date.now() - BEARER_REFRESH_MARGIN_MS` (60s) and `armBearerRefresh` floors at 5s, so a source that keeps returning a bearer already inside its refresh margin re-reads every **5 seconds** with `BEARER_RETRY_MS` (15s) bypassed for the same reason: the fetch succeeded, the material is not usable. Identical defect, slower cadence.

I did not fix it here. It is a different auth mode with no comparable smoke seam in this file, so it needs its own repro to satisfy the reproduce-first rule rather than riding along on this one. Happy to open it as its own issue, or to add it here if you would rather have both in one change.

## Docs

`docs/embedding.md` already promised the daemon "fails loud rather than riding to expiry" on the delivery creds read. That was not quite true for an un-re-signed generation, which looped quietly. The page now says what happens: reported, retried in 60 seconds, current cred live to its expiry, store read once per retry.

AI disclosure: written by Claude (Anthropic) under my review and testing; no human reviewed it besides me.


L4XB:
CI was red on one thing and it was mine: `check:docsbundle` compares `extensions/connector-core/src/docs-bundle.generated.ts` against the pages, so the sentence I added to `docs/embedding.md` had to be regenerated into the bundle in the same change. Done in `78f8834a` (`pnpm gen:tooldocs && pnpm gen:docsbundle`, one line of the bundle changed), and `pnpm check:docsbundle` now exits 0 locally.

That was the only failing step: `unit` failed solely on the bundle diff, and `ci-ok` is the aggregate gate, so it reported `unit=failure` with `smoke=success live=success seat-pack=success seat-linux-arm64=success`. Mutation reproof and the live suite were green, including the new churn cells.

davidfarah2003:
Thanks for this. It conflicts with main as of a9c984939 (#1517 landed after you branched): `packages/core/src/endpoint.ts` only, your hunks at 803, 831 and 968.

What moved: #1517 deleted the old pre-dial block around 803-830 and put the expiry refusal into `private static presentableCreds` (now line 824), reached through `private credsForWire()` (844). Your `assertRenewableGeneration` guards adoption of a generation; that one guards presentation to the wire. They belong side by side rather than one replacing the other. The `Math.max(1_000, delayMs)` re-arm at 1092 is untouched by #1517, so the churn you are fixing is still live on main.

One more thing to carry over from #1523's triage: the fix needs a cell that goes red without it. `standing-renewal-auth.smoke.ts` on main stays 12/12 green with the arming path broken (measured with the repo's mutation-proof), so a passing suite is not evidence here. Your new churn cells in that file may already be that; if so, a note in the PR body saying which cell fails on main before the fix would let the review lean on it.

A rebase onto main and a push is all that's needed from your side. We won't push to your branch.

L4XB:
Rebased onto `3e5b694c` and force-pushed. The conflict was positional only, and it resolved the way you described: `presentableCreds` / `credsForWire` guard **presentation to the wire**, `assertRenewableGeneration` guards **adoption of a generation**, and both now sit in the class. Neither replaces the other, and the `Math.max(1_000, delayMs)` re-arm this PR is about is untouched by #1517.

`pnpm smoke:standing-renewal` is 15/15 on the rebased head, and `pnpm check:docsbundle` is clean (the regenerated bundle is byte-identical, so the docs commit still holds after #1517).

**Which cell fails on main before the fix.** Removing the two `assertRenewableGeneration` call sites and leaving the rest of the branch in place:

```
✗ FAIL: an unchanged source past 75% did NOT busy-loop (<=1 renewal read in the churn window)
        { churnReads: 4, churnBaseline: 2 }
✗ FAIL: the refusal is reported once and names the un-re-signed generation   []
STANDING RENEWAL SMOKE FAILED ❌  (13 passed, 2 failed)
```

The first is the one to lean on: it counts source reads inside a fixed window, so it grades the loop rather than the message. Four reads where the arming path is sound produces two.

You were right that a passing suite is not evidence, so I have shipped the fixture rather than only the note — `packages/core/smoke/mutations/standing-renewal-churn.json`, in the shape of `standing-renewal-expiry.json`. It disarms the same-generation arm with a comparison that can never hold (`delayMs <= Number.NEGATIVE_INFINITY`), which leaves the tree compiling and the already-expired arm above it untouched:

```
KILLED  the un-re-signed generation is adopted instead of refused
  red, and named: an unchanged source past 75% did NOT busy-loop · 13 marks (baseline 15)
```

L4XB:
CI caught something the rebase created, and it was mine: `smoke:mutation-fixtures` reported

```
packages/core/smoke/mutations/creds-supply-expiry.json
  ✗ [2] ANCHOR IS AMBIGUOUS — `find` matches 2x in packages/core/src/endpoint.ts
```

`#1517`'s `presentableCreds` and the `assertRenewableGeneration` this PR adds both open with the same line:

```ts
    if (typeof exp === "number" && exp * 1000 <= Date.now())
```

so the grace-window mutation's anchor stopped being unique the moment the two landed together. Pushed `6b4e19f`, which extends that anchor by the two lines that distinguish the presentation check:

```
    if (typeof exp === "number" && exp * 1000 <= Date.now())
      throw new Error(
        opts.renewable
```

Uniqueness alone is not the point, so I re-proved the config rather than just re-running the checker — all 5 still kill, each naming its own cell:

```
KILLED  the wire getter is unchecked again                               · 17 marks (baseline 23)
KILLED  the adoption preflight bypasses the checkpoint                   · 21
KILLED  the expiry comparison is loosened to a strict past-check         · 16
KILLED  the dial gate tests the cached cred for TRUTHINESS again         · 22
KILLED  the self-heal never runs after a credential refusal              · 20
```

`pnpm smoke:mutation-fixtures` is now `420 fixture file(s), 2243 anchor(s)`, 0 dead and 0 ambiguous.

I left both guards' text as it is rather than renaming one to dodge the collision: the shared opening line is shared because they check the same property at two different seams, which is the reason you gave for keeping both, and a fixture anchor is the right place to absorb that.

davidfarah2003 (changes requested):
Thanks for this, and the renewal latch work is solid. One blocking issue in the new generation guard, plus a coverage gap that hides it.

## The guard compares the envelope, not the credential

`packages/core/src/endpoint.ts:875`:

```ts
if (candidate === current && delayMs <= 0)
  throw new Error("the creds source still holds the previous generation ...");
```

`candidate` and `current` are whole creds files. The JWT inside them is what identifies a generation, and `jwtFromCreds` (`packages/core/src/identity.ts:68-70`) extracts it with `\s*` padding and a `.trim()`:

```ts
creds.match(/BEGIN NATS USER JWT-----\s*([\s\S]*?)\s*------END NATS USER JWT/)?.[1].trim()
```

So the envelope and the credential are not the same string, and any whitespace difference separates them:

| source returns | `candidate === current` | `credsFingerprint` equal |
| --- | --- | --- |
| byte-identical | true | true |
| one trailing newline | **false** | true |
| leading whitespace | **false** | true |
| blank line after the BEGIN marker | **false** | true |
| a genuinely different JWT | false | false |

The last row is the control: the fingerprint still separates a real generation change, so this is not a case of the comparison being too weak in general. It is specifically blind to representation.

This was confirmed at the real authentication boundary rather than against the regex alone. Using the version matched `@nats-io/jwt` and `@nats-io/nats-core` already in the tree, one valid credential and a copy with only outer whitespace added (961 vs 965 bytes) were passed through the public `parseCreds` and `credsAuthenticator`: both returned the same JWT, seed key and subject, and both authenticated. **NATS sees one generation where `:875` sees two different strings.**

## Why it matters

`EndpointOptions.creds` takes opaque file content and promises no canonical whitespace. When such a source re-serves the same credential past its renewal point, the guard is meant to refuse it. Instead `candidate === current` is false, the refusal is skipped, the stale generation is adopted with `delayMs <= 0`, and the 1 second renewal loop this PR exists to close comes back. The guard reports that it enforces generation identity while enforcing byte identity.

This is attributable to this change rather than pre-existing: `candidate === current` appears 0 times at the merge base and once at this head, with `credsFingerprint` present on both sides.

## Scope

Credential path only. `assertRenewableGeneration` is `private static` with two call sites, `:910` and `:1054`, both credential related. `refreshBearer` does not call it, so the bearer path is unaffected. The 2.0s value and the latch itself are untouched by this and read correctly.

## Suggested fix

`credsFingerprint` is already imported at `endpoint.ts:19`, so no new machinery is needed:

```ts
if (current !== undefined && credsFingerprint(candidate) === credsFingerprint(current) && delayMs <= 0)
  throw new Error(...);
```

Comparing the extracted JWT directly would work equally well. The point is to compare what identifies the generation rather than the bytes that carry it.

## The coverage gap that hides this

The declared mutations all pass, and that green is not evidence here. The fixture's source returns the exact same string every time, so it cannot express a case where two different strings carry one credential. A mutation suite that emits one representation cannot detect a representation dependent bug, which is why this survived a passing run.

Worth adding one cell whose source returns **the same credential with different whitespace**, expecting the stale generation to be refused. It should go red on the current code and green after the fix. Without it, the next reader sees a fully green suite over a case it structurally cannot reach.

Happy to look again once the comparison is on the JWT.

L4XB:
Fixed in `37b9609b`. You were right on both halves, and the second one is the part I would have missed on my own.

**The comparison.** `assertRenewableGeneration` now compares fingerprints:

```ts
if (current !== undefined && credsFingerprint(candidate) === credsFingerprint(current) && delayMs <= 0)
```

`credsFingerprint` was already imported. The comment at the call site now says why the envelope is the wrong thing to compare — no canonical whitespace from `EndpointOptions.creds`, `jwtFromCreds` padding with `\s*` and trimming, and the seed block riding along in the same file.

**The coverage gap.** A new cell drives a source that returns the same credential with a trailing newline after the first read, with an accept control asserting the two envelopes really do differ in bytes and agree in fingerprint. The fixture now carries **one mutation per conjunct**: the delay mutation was passing while the equality comparison went entirely ungraded, which is exactly how this survived a green run. Under the new byte-comparison mutation the churn cell stays green — its source emits one exact string, where the two equalities agree — and the reformatted cell is the one that reddens.

```
pnpm smoke:standing-renewal      19/19
node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/standing-renewal-churn.json
  KILLED  the un-re-signed generation is adopted instead of refused
    red, and named: an unchanged source past 75% did NOT busy-loop · 15 marks (baseline 19)
  KILLED  the refusal compares envelope bytes again
    red, and named: a reformatted envelope of the SAME generation is refused, so no 1s re-read follows · 18 marks (baseline 19)
  All 2 mutation(s) killed. The suite discriminates.
```

One thing worth recording, because it nearly made the new cell vacuous: **the reformatting has to be one the broker accepts.** My first attempt used CRLF line endings. Those differ in bytes, but the broker refuses that envelope, so under the mutation the adoption was blocked by the *preflight* for an unrelated reason and the suite stayed green — a test that would have passed on the unfixed code. The trailing newline passes the preflight, which is what lets the defect actually reach the churn.

Same defect, same file family: the membership feed's `adoptRwCreds` has it too (`candidate === currentRwCreds`), reported separately as #1563 and fixed in #1571, since it predates this branch and this PR does not touch that file.

`tsc --noEmit -p packages/core/tsconfig.json` is clean.

davidfarah2003:
Thanks for the update. The generation comparison at the guard is correct and the panel has verified it over several envelope shapes. Two things before this can merge, both in the test rather than the product change.

First, the new churn cell reddens under mutation by a margin of one read. Under the byte-comparison mutant the source returns the same reformatted envelope on every read after the first, so the mutant adopts once, then refuses on the next read because the bytes now match what it cached, and the loop stops. That leaves two reads in the window against a tolerance of one, and any jitter that lands the second read before the baseline capture makes the mutation survive. Across the review the equality mutation was run nine times and failed to produce a clean named red four times, and an ungradable result fails the reproof shard outright. The fix is to have the reformatted source return a fresh envelope on each call, for example an incrementing count of trailing newlines, so successive envelopes are byte-distinct but fingerprint-identical, the mutant cannot self-arrest, and the storm runs the full window. Re-anchoring the baseline alone does not remove the self-arrest. Please show stable kills across several runs rather than one, since a single green is what an intermittent cell produces more often than not.

Second, the branch is eleven commits behind main. That is what widened the reproof selection to fixtures this PR does not touch and pulled in an unrelated red in `creds-supply-expiry.json`. We reran that fixture on a clean clone and it killed 5 of 5, so it is not your defect. A rebase onto current main clears it.

L4XB:
Both done — `a39ce0cc` for the cell, `f874e012` for the rebase.

**The self-arrest.** You had the mechanism exactly right, and re-anchoring the baseline would not have touched it. The source served one repeated string, so the byte-comparison mutant adopted the reformatted envelope once, cached it, and then matched its own cache on the next read: the storm stopped at two reads against a tolerance of one, and any jitter that put the second read before the baseline capture erased the whole margin.

Each read now returns the same generation in different bytes:

```ts
const wsReformatted = (read: number) => wsCred + "\n".repeat(read);
const wsSource = async (): Promise<string> => {
  wsReads++;
  return wsReads === 1 ? wsCred : wsReformatted(wsReads);
};
```

The mutant can no longer recognise its own cache, so it churns the full window and the cell reds by the whole count rather than by one. The accept control now samples two different reads and asserts they differ from the original **and from each other**, so a future edit that collapses the source back to a constant reds there instead of silently restoring the self-arrest.

**Stable kills, five runs.** Not one green:

```
run 1  All 2 mutation(s) killed.
run 2  All 2 mutation(s) killed.
run 3  All 2 mutation(s) killed.
run 4  All 2 mutation(s) killed.
run 5  All 2 mutation(s) killed.
```

Each run kills both `the un-re-signed generation is adopted instead of refused` and `the refusal compares envelope bytes again` — the latter being the one that was ungradable four times in nine before. The suite itself is 19/19.

One operational note from doing this: an interrupted `mutation-proof` run left a mutated `endpoint.ts` in the worktree, and the next run refused with "working tree is dirty" rather than proceeding — the refusal did its job, but it is worth knowing that an aborted run needs a `git checkout --` before the next one.

**Rebase.** Merged current `main` (14 commits by then). `creds-supply-expiry.json` is not in the diff any more, so the reproof selection should narrow back to the fixture this PR owns.

Issue #1518, opened 2026-09-12
closure-gate rc 2/3 smoke cells are substring matches and cannot see a branch that exits 1


davidfarah2003:
Follow-up from PR #1502 (merged `b98e9da80`). Non-blocking there, deliberately not held for it, but it is a real coverage gap and it is the same shape as the bug #1286 existed to prevent.

**The gap.** In `bin/smoke/publish-closure-gate.smoke.ts`, the cells covering the UNSETTLED (rc 2) and NONE (rc 3) branches of the closure-gate step assert on substrings of the step's run script, roughly `"\$rc" -eq 2` and `"\$rc" -eq 3`. A substring match sees that the branch *exists*; it cannot see what the branch *body does*.

**Reproduced first-hand during review, not theorised.** Injecting `exit 1` into the rc-2 branch of `.github/workflows/changesets.yml` leaves `pnpm smoke:publish-closure-gate` green at 18/18. Two reviewers found this independently and one reproduced it by hand in a clone, then restored the tree.

**Why it matters.** rc 2 (UNSETTLED) and rc 3 (NONE) are the *quiet skip* paths: a version-PR push that publishes nothing must not fail the job. If either branch gained a failing exit, every no-publish push to main would red, and this suite would stay green while it happened. The committed workflow verifiably skips today, so nothing is broken right now; the suite just does not defend it.

**Suggested fix.** Assert on the branch *body*, not the presence of the comparison: parse the rc-2 and rc-3 branch bodies out of the step's run script and assert the absence of a failing exit, then add matching cells to `bin/smoke/mutations/publish-closure-gate.json` so a mutation that inserts `exit 1` into either branch is KILLED. `publish-closure-gate.json` currently has no cells for that branching at all.

**General note.** This is the second substring-assertion hole found in this one file today. The first was `run.includes("verify-publish-closure.mjs")`, which a diagnostic `echo` containing the same filename satisfied, letting the verifier invocation be deleted with the suite still green; that one is fixed and mutation-killed in #1502. A bare `includes()` on a workflow's run script is a weak assertion in general, because any comment, echo, or error message can satisfy it. Both cells should pin structure rather than text.

davidfarah2003:
[...]

PR #1527, opened 2026-09-12
test(closure-gate): assert the rc-2 and rc-3 branch bodies, not their comparisons

L4XB:
Refs #1518.

## Reproduced first, on a clean tree

Per AGENTS.md the repro comes before the fix. On `origin/main`, injecting `exit 1` into the rc-2 branch of `.github/workflows/changesets.yml`:

```
✓ every cell ran (17 before sentinel)
PUBLISH CLOSURE GATE SMOKE OK (18 passed, 0 failed)
SUITE COMPLETE
```

Green, exactly as the issue reports, while every no-publish push to `main` would have failed the job.

## The fix

The rc-2 and rc-3 cells substring-matched the branch comparison, which sees that the branch exists and nothing about what it does. They now parse the branch **body** out of the step's run script and assert the absence of a failing exit.

`rcBranchBody(run, code)` ends a branch at the next line indented the same as its own `if`/`elif`/`else` keyword, so anything indented deeper stays part of the body and a nested block cannot hide a line from the assertion. `else` is addressed as code `-1`, being the unexpected-rc arm.

**Two cells come first as the vacuity guard**, and that ordering is the point: "this body contains no failing exit" is worth nothing unless the same parser finds a failing exit where one exists. So the PARTIAL branch and the unexpected-rc branch are asserted to fail, with the same helper, before the two quiet skips are asserted not to. That is the hole the `includes("verify-publish-closure.mjs")` cell had in #1502 — an assertion that any echo could satisfy — and it is avoidable here only by making the positive direction part of the same family.

The unexpected-rc cell is new. The `else` arm also fails the job, it is the same class of property, and nothing covered it.

## Verified

Both injections are now caught, one cell each:

```
--- exit 1 in the rc-2 branch:
  ✗ FAIL: the closure-gate step handles exit code 2 (UNSETTLED) without failing
  PUBLISH CLOSURE GATE SMOKE FAILED (18 passed, 1 failed)
--- exit 1 in the rc-3 branch:
  ✗ FAIL: the closure-gate step handles exit code 3 (NONE) without failing
  PUBLISH CLOSURE GATE SMOKE FAILED (18 passed, 1 failed)
```

Four mutation cells added to `bin/smoke/mutations/publish-closure-gate.json`, one per branch, where it had none for this branching at all: an `exit 1` inserted into either quiet skip, and a removed `exit 1` in either failing branch.

```
KILLED  the UNSETTLED quiet skip starts failing the job
  red, and named: the closure-gate step handles exit code 2 (UNSETTLED) without failing
KILLED  the NONE quiet skip starts failing the job
  red, and named: the closure-gate step handles exit code 3 (NONE) without failing
KILLED  the PARTIAL branch stops failing the job
  red, and named: the closure-gate step branches on exit code 1 (PARTIAL) to fail the job
KILLED  an unexpected exit code stops failing the job
  red, and named: the closure-gate step fails the job on an unexpected exit code

All 10 mutation(s) killed. The suite discriminates.
```

`pnpm smoke:publish-closure-gate` → 19 passed, 0 failed. The cell-count sentinel moves 17 → 18 for the one new cell. `check:docs-voice` passes; no docs page describes this, so none changed.

## Not changed

The workflow itself. It skips correctly today, which is why the issue is a coverage gap and not a bug: the committed rc-2 and rc-3 branches carry only an `echo`, and the new cells now hold them there.

AI disclosure: written by Claude (Anthropic) under my review and testing; no human reviewed it besides me.

### DIFF
--- a/bin/smoke/mutations/publish-closure-gate.json
+++ b/bin/smoke/mutations/publish-closure-gate.json
@@ -10,7 +10,8 @@
   "proveWith": "pnpm mutation-proof --config bin/smoke/mutations/publish-closure-gate.json",
   "why": [
     "The release workflow used to check one package of twenty-two and cut a Release on that alone. A sibling that failed to publish was invisible. The closure gate now checks every package and the Release step depends on it.",
-    "A 200 without a body identifying the package at the requested version is no longer treated as presence. Each mutation below breaks one property the smoke is supposed to hold."
+    "A 200 without a body identifying the package at the requested version is no longer treated as presence. Each mutation below breaks one property the smoke is supposed to hold.",
+    "The rc-2 and rc-3 cells used to substring-match the branch comparison, which sees that the branch exists and nothing about what it does: an exit 1 added to either quiet-skip path left the suite green while every no-publish push to main would have failed (#1518). The cells now parse each branch body, and these mutations are what shows they can see into it."
   ],
   "mutations": [
     {
@@ -60,6 +61,38 @@
       "replace": "        if (body && body.version === version) continue;",
       "expectRed": "fake registry: one 200 naming a different package -> does NOT pass as published",
       "cell": "fake registry: one 200 naming a different package -> does NOT pass as published"
+    },
+    {
+      "name": "the UNSETTLED quiet skip starts failing the job",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"UNSETTLED — the registry has not converged. Skipping the Release.\"\n",
+      "replace": "            echo \"UNSETTLED — the registry has not converged. Skipping the Release.\"\n            exit 1\n",
+      "expectRed": "the closure-gate step handles exit code 2 (UNSETTLED) without failing",
+      "cell": "the closure-gate step handles exit code 2 (UNSETTLED) without failing"
+    },
+    {
+      "name": "the NONE quiet skip starts failing the job",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"NONE — this push published nothing. Skipping the Release.\"\n",
+      "replace": "            echo \"NONE — this push published nothing. Skipping the Release.\"\n            exit 1\n",
+      "expectRed": "the closure-gate step handles exit code 3 (NONE) without failing",
+      "cell": "the closure-gate step handles exit code 3 (NONE) without failing"
+    },
+    {
+      "name": "the PARTIAL branch stops failing the job",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"PARTIAL PUBLISH — failing the job.\"\n            exit 1\n",
+      "replace": "            echo \"PARTIAL PUBLISH — failing the job.\"\n",
+      "expectRed": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
+      "cell": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job"
+    },
+    {
+      "name": "an unexpected exit code stops failing the job",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"Unexpected exit code $rc from verify-publish-closure.mjs — failing the job.\"\n            exit 1\n",
+      "replace": "            echo \"Unexpected exit code $rc from verify-publish-closure.mjs — failing the job.\"\n",
+      "expectRed": "the closure-gate step fails the job on an unexpected exit code",
+      "cell": "the closure-gate step fails the job on an unexpected exit code"
     }
   ]
 }
--- a/bin/smoke/mutations/publish-closure-gate.json
+++ b/bin/smoke/mutations/publish-closure-gate.json
@@ -65,34 +65,52 @@
     {
       "name": "the UNSETTLED quiet skip starts failing the job",
       "file": ".github/workflows/changesets.yml",
-      "find": "            echo \"UNSETTLED — the registry has not converged. Skipping the Release.\"\n",
-      "replace": "            echo \"UNSETTLED — the registry has not converged. Skipping the Release.\"\n            exit 1\n",
+      "find": "            echo \"UNSETTLED \u2014 the registry has not converged. Skipping the Release.\"\n",
+      "replace": "            echo \"UNSETTLED \u2014 the registry has not converged. Skipping the Release.\"\n            exit 1\n",
       "expectRed": "the closure-gate step handles exit code 2 (UNSETTLED) without failing",
       "cell": "the closure-gate step handles exit code 2 (UNSETTLED) without failing"
     },
     {
       "name": "the NONE quiet skip starts failing the job",
       "file": ".github/workflows/changesets.yml",
-      "find": "            echo \"NONE — this push published nothing. Skipping the Release.\"\n",
-      "replace": "            echo \"NONE — this push published nothing. Skipping the Release.\"\n            exit 1\n",
+      "find": "            echo \"NONE \u2014 this push published nothing. Skipping the Release.\"\n",
+      "replace": "            echo \"NONE \u2014 this push published nothing. Skipping the Release.\"\n            exit 1\n",
       "expectRed": "the closure-gate step handles exit code 3 (NONE) without failing",
       "cell": "the closure-gate step handles exit code 3 (NONE) without failing"
     },
     {
       "name": "the PARTIAL branch stops failing the job",
       "file": ".github/workflows/changesets.yml",
-      "find": "            echo \"PARTIAL PUBLISH — failing the job.\"\n            exit 1\n",
-      "replace": "            echo \"PARTIAL PUBLISH — failing the job.\"\n",
+      "find": "            echo \"PARTIAL PUBLISH \u2014 failing the job.\"\n            exit 1\n",
+      "replace": "            echo \"PARTIAL PUBLISH \u2014 failing the job.\"\n",
       "expectRed": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
       "cell": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job"
     },
     {
       "name": "an unexpected exit code stops failing the job",
       "file": ".github/workflows/changesets.yml",
-      "find": "            echo \"Unexpected exit code $rc from verify-publish-closure.mjs — failing the job.\"\n            exit 1\n",
-      "replace": "            echo \"Unexpected exit code $rc from verify-publish-closure.mjs — failing the job.\"\n",
+      "find": "            echo \"Unexpected exit code $rc from verify-publish-closure.mjs \u2014 failing the job.\"\n            exit 1\n",
+      "replace": "            echo \"Unexpected exit code $rc from verify-publish-closure.mjs \u2014 failing the job.\"\n",
       "expectRed": "the closure-gate step fails the job on an unexpected exit code",
       "cell": "the closure-gate step fails the job on an unexpected exit code"
+    },
+    {
+      "name": "a separator-based failing exit is added to the UNSETTLED branch",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"UNSETTLED \u2014 the registry has not converged. Skipping the Release.\"",
+      "replace": "            echo \"UNSETTLED \u2014 the registry has not converged. Skipping the Release.\"; exit 1",
+      "expectRed": "the closure-gate step handles exit code 2 (UNSETTLED) without failing",
+      "cell": "the closure-gate step handles exit code 2 (UNSETTLED) without failing",
+      "note": "The direction that matters for this cell: it asserts the NEGATIVE, so a matcher that cannot read the failure certifies a job-failing branch as safe and the green is indistinguishable from a correct one. The line-anchored matcher this PR replaces stayed green here."
+    },
+    {
+      "name": "the PARTIAL branch exits 0 instead of failing",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"PARTIAL PUBLISH \u2014 failing the job.\"\n            exit 1",
+      "replace": "            echo \"PARTIAL PUBLISH \u2014 failing the job.\"\n            exit 0",
+      "expectRed": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
+      "cell": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
+      "note": "The accept direction, held as strict as before: a correct `exit 0` is never read as a failure, in any position. A widening that lost the non-zero lookahead would keep this cell green and would red a valid workflow elsewhere."
     }
   ]
 }
--- a/bin/smoke/mutations/publish-closure-gate.json
+++ b/bin/smoke/mutations/publish-closure-gate.json
@@ -111,6 +111,24 @@
       "expectRed": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
       "cell": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
       "note": "The accept direction, held as strict as before: a correct `exit 0` is never read as a failure, in any position. A widening that lost the non-zero lookahead would keep this cell green and would red a valid workflow elsewhere."
+    },
+    {
+      "name": "a reachable keyword-position failure is added to the UNSETTLED branch",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"UNSETTLED \u2014 the registry has not converged. Skipping the Release.\"",
+      "replace": "            echo \"UNSETTLED \u2014 the registry has not converged. Skipping the Release.\"\n            if true; then exit 1; fi",
+      "expectRed": "the closure-gate step handles exit code 2 (UNSETTLED) without failing",
+      "cell": "the closure-gate step handles exit code 2 (UNSETTLED) without failing",
+      "note": "bash gives this body rc=1, so the quiet skip becomes a job failure. A matcher that treats only punctuation as a command position reads it as safe: `; exit 1` matches and `; then exit 1` does not. Five more shapes hide behind the same gap (else, while, for, until, elif)."
+    },
+    {
+      "name": "the PARTIAL branch's exit is moved inside a string",
+      "file": ".github/workflows/changesets.yml",
+      "find": "            echo \"PARTIAL PUBLISH \u2014 failing the job.\"\n            exit 1",
+      "replace": "            echo \"PARTIAL PUBLISH; exit 1\"",
+      "expectRed": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
+      "cell": "the closure-gate step branches on exit code 1 (PARTIAL) to fail the job",
+      "note": "bash gives this body rc=0, so a partial publish stops failing the job and a Release is cut. It is the worse direction: the vacuity guard is what licenses the two negative cells, so a guard satisfied by a body that does not fail licenses nothing."
     }
   ]
 }
--- a/bin/smoke/publish-closure-gate.smoke.ts
+++ b/bin/smoke/publish-closure-gate.smoke.ts
@@ -104,45 +104,129 @@ function rcBranchBody(run: string, code: number): string | null {
   return body.join("\n");
 }
 
-/** Where a command can begin on a line: at the start, or after a separator or an opening group.
- *  Word characters are deliberately absent, so `echo "Unexpected exit code $rc"` is prose to this
- *  matcher rather than a command. */
-const COMMAND_START = String.raw`(?:^|[;&|(){}])\s*`;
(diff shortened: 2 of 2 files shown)


davidfarah2003:
Thanks for this. It is a real fix to a real defect, and the mutation config that ships with it is better than most internal PRs get. Review at `e15a0f09a4c938e09d0dc7d3193d29937918fb0d` came back **2 of 3 reviewers requesting changes**, on one specific point, with a bounded ask. Everything else cleared.

**What we verified and agreed on.** The suite reads 19 passed / 0 failed at this head, all 10 mutations in `bin/smoke/mutations/publish-closure-gate.json` are KILLED, each naming its own cell at 18 marks against a baseline of 19. We also confirmed the harness can report a survivor (a deliberate comment-only mutation reports one), and that the #1518 repro, a bare `exit 1` in the rc-2 arm, reds the right cell here and stays green on `main`. So the change is not cosmetic: it sees an input the old substring match provably cannot. `rcBranchBody` in particular resisted every attack we threw at it (nested blocks, `exit 7`, heredocs, deleted arms, reordered branches, a fifth branch, trailing whitespace), and where it cannot parse an opener it returns `null` and reds loudly rather than passing quietly. That is the right failure direction.

**The blocking point.** It is in `failsTheJob`, not in the branch slicing:

```
/^\s*exit\s+(?!0\b)\d+\s*$/m.test(body) || /^\s*(false|exit\s+1)\s*$/m.test(body)
```

Both patterns are line-anchored on `^…$`, so anything sharing a line with the exit is invisible. That matters more than a normal coverage gap because of how two of the four cells are written: lines 130 and 136 assert `!failsTheJob(...)`. When the matcher under-detects, those cells do not merely miss a defect; they affirmatively certify a job-failing branch as safe. A green there means "I did not recognise this", and it reads identically to "this is correct".

We confirmed 14 forms that evade the matcher while genuinely failing the release job, each checked end to end (suite reports 19/0, the step still exits non-zero): `exit $rc`, `exit 1 # comment`, `echo hi; exit 1`, `true && exit 1`, `exit 1 ;`, `exit 1 ;;`, `{ exit 1; }`, `(exit 1)`, `! :`, `eval "exit 1"`, `return 1`, `kill -TERM $$`, `/bin/false`, and `case x in x) false ;; esac`. Two of these are current house style in this repo: `|| { echo "..."; exit 1; }` appears in `ci.yml`, `mutation-reproof.yml` and `installer.yml`, and `exit $rc` is the natural way to forward a captured status.

One reviewer also found the inverse, which we think is the sharpest single case: replace PARTIAL's reachable `exit 1` with an unreachable decoy (`if false; then exit 1; fi`) and the vacuity guard at line 117 stays green while the job no longer fails on a real partial publish. The guard that exists to prove the parser can see a failing exit is itself satisfiable by text that never runs.

**The ask, and it is deliberately bounded.** We are not asking for a shell evaluator or for dynamic execution, the suite runs on Windows shards too, so executing bash is not portable here and we are not requesting it. Three things:

1. Widen `failsTheJob` so a failure command is recognised when it follows a command separator (`;`, `&&`, `||`, `(`) or is followed by an inline comment.
2. **Keep the accept direction exactly as strict as it is now.** This is the part we want to flag rather than assume: two reviewers independently drafted wider regexes off the cuff and *both* flagged a correct `exit 0` as a failure, which would red a valid workflow, strictly worse than the gap being fixed. Your current matcher has zero false positives across the benign bodies we tried. Please preserve the non-zero lookahead.
3. Add two mutation cells to `publish-closure-gate.json` covering both directions: one failure form that the current regex misses (an inline-comment or separator-based exit), and one benign `exit 0` control that must **not** be flagged.

For reference, a shape we tested in both directions detected 14 of 17 failure forms with 0 false positives on 6 benign bodies. Please treat that as a starting point rather than a specification, your own version with the mutation cells proving both directions is what matters.

Finally, a small thing: the body says `Fixes #1518`. Please change it to `Refs #1518` so the issue is closed deliberately after merge rather than automatically.

Nothing else is blocking. The head is clean and all checks are green, so once `failsTheJob` covers the demonstrated forms with the `exit 0` direction preserved and the two mutation cells are in, this is ready to land.

L4XB:
All three, plus the `Refs #1518` change in the body.

**1. `failsTheJob` reads a failure command anywhere on its line.** A failure command is recognised at the start of a line or after a separator or opening group (`;`, `&&`, `||`, `(`, `{`, `}`), and may be followed by a separator, a closing group, or an inline comment:

```ts
const COMMAND_START = String.raw`(?:^|[;&|(){}])\s*`;
const COMMAND_END   = String.raw`\s*(?:$|[;&|)}#])`;
const NONZERO_EXIT  = String.raw`exit\s+(?!0${COMMAND_END})\S+`;
```

plus `false` / `/bin/false`, `return <non-zero>`, `! :`, and `kill -SIG $$`. Word characters are deliberately absent from `COMMAND_START`, which is what keeps `echo "Unexpected exit code $rc ..."` prose rather than a command.

**2. The accept direction is unchanged.** The non-zero lookahead is preserved and now applies in every position, so `exit 0`, `exit 0 # fine`, `echo ok; exit 0`, `(exit 0)`, `{ exit 0; }` and `return 0` are all still benign. The one case I handled differently from the obvious widening is `eval "exit 1"`: matching the quote would also flag `echo "exit 1"`, so it is matched on the `eval` word instead.

Measured in both directions against your 14 forms plus the real branch bodies:

```
failing detected 22/22 · false positives 0/14
```

The 22 include all 14 you listed, and also `exit "$rc"`, `exit ${rc}`, `false || exit 1` and the actual PARTIAL body. The 14 benign include the four real branch bodies from `changesets.yml`, `echo "exit 1"` and a commented-out `# exit 1`.

**3. Two mutation cells, one per direction.** Both are in `bin/smoke/mutations/publish-closure-gate.json`:

```
KILLED  a separator-based failing exit is added to the UNSETTLED branch
  red, and named: the closure-gate step handles exit code 2 (UNSETTLED) without failing · 18 marks (baseline 19)
KILLED  the PARTIAL branch exits 0 instead of failing
  red, and named: the closure-gate step branches on exit code 1 (PARTIAL) to fail the job · 18 marks (baseline 19)
```

The first is the direction you identified as the sharp one: it appends `; exit 1` to the UNSETTLED echo, which the line-anchored matcher certified as safe. The second holds the accept direction, and a widening that dropped the lookahead would leave it green.

All 12 mutations in that config are KILLED, each naming its own cell at 18 against a baseline of 19. Suite is 19/0.

**On the unreachable decoy.** You are right that it is the sharpest case and I have not fixed it: `if false; then\n  exit 1\nfi` satisfies the vacuity guard while the job never fails, and seeing that needs the shell evaluator you explicitly did not ask for. This widening does not make it worse. What I have done is say so in the function's own docstring, so the next reader learns the limit from the code rather than from this thread. If you want it closed later, the cheap-and-honest version is probably a separate cell refusing a branch whose failure command sits inside a nested conditional at all, rather than trying to decide reachability.

davidfarah2003:
Re-reviewed at `8ae5c044905849772485b8ba8b6f21ae37261b29`. All three asks are addressed, the `Refs #1518` change is in, and I reproduced both of your headline figures exactly on an independent harness: `22/22` failing detected, `0/14` false positives. Every cell you listed is correct. The mutation config is real work: 12/12 killed, each naming its own cell, at marks 12, 16 and ten at 18 against a baseline of 19. Suite is 19 passed, 0 failed at the landing tree, and `mutation-proof --selftest` passes 46 checks, so the harness can fail.

I am still asking for changes, because the suite does not yet close #1518 and I can show the hole rather than assert it.

**The defects are entirely in rows your cell set does not contain.** Against real bash:

```
if true; then exit 1; fi              bash rc=1   matcher QUIET     false negative
if false; then : ; else exit 1; fi    bash rc=1   matcher QUIET     false negative
while true; do exit 1; done           bash rc=1   matcher QUIET     false negative
for i in 1; do exit 1; done           bash rc=1   matcher QUIET     false negative
until false; do exit 1; done          bash rc=1   matcher QUIET     false negative
elif [ x = x ]; then exit 1           bash rc=1   matcher QUIET     false negative
echo "a; exit 9"                      bash rc=0   matcher FAILING   false positive
exit 00                               bash rc=0   matcher FAILING   false positive
```

**Root cause is one token.** `COMMAND_START` is `(?:^|[;&|(){}])\s*`, which allows line start and punctuation. Shell keywords are command positions too. Holding everything else identical: `"; exit 1"` matches, while `"; then exit 1"`, `"; else exit 1"` and `"; do exit 1"` do not.

**#1518 is still reachable at this head.** Injected into the real committed quiet rc-2 arm of `changesets.yml`, with controls reading correctly in both directions (a bare `exit 1` reds the suite, a `:` no-op leaves it green):

```
if true; then exit 1; fi              job rc=1   suite 19/0 GREEN
if [ -n "$rc" ]; then exit 1; fi      job rc=1   suite 19/0 GREEN
while true; do exit 1; done           job rc=1   suite 19/0 GREEN
case x in x) exit 1;; esac            job rc=1   suite RED, caught
```

This is not the limitation your docstring discloses. You excluded UNREACHABLE failures, and `if false; then exit 1; fi` really is rc=0, so excluding it is correct. These rows are reachable and really do fail the job.

**The sharper finding is in the opposite direction, and it breaks the vacuity guard itself.** The guard is supposed to license the two negative cells. But the positive cells are wrong too. Replace the PARTIAL body with `echo "PARTIAL PUBLISH; exit 1"`:

```
bash rc=0, so a partial publish no longer fails the job and a Release is cut
matcher says failsTheJob = true
the cell "branches on exit code 1 (PARTIAL) to fail the job" stays GREEN
suite: 19 passed, 0 failed
control: deleting the exit 1 outright IS correctly caught
```

`exit 00` does the same. So a body that does not fail the job satisfies the guard, which means the guard cannot license the negative cells, and no mutation covers this direction.

**What would close this.** The one item from my last comment that is still missing is the item that would have caught every row above at authoring time: a self-test battery that calls `failsTheJob` on literal bodies and asserts both directions. All five current call sites are workflow-derived, so nothing tests the helper against a body you wrote by hand. The precedent is already in the repo as `mutation-proof --selftest`, it needs no bash and no Windows shard.

Seed it with the rows above, both directions, each asserting the bash exit status you can verify with `bash -c` in one line. Then fix whatever it reds. I am deliberately not handing you a regex: every candidate matcher proposed on this PR, including mine, looked correct to its author and was broken by a peer within minutes. The property to hold is that a shell keyword is a command position, and the battery is what proves you hold it.

One measured caution, since the obvious widening is not a free win. Adding `\b(?:then|else|do)\b` to `COMMAND_START` takes false negatives from 6 to 0 on a 21-row set, but takes false positives from 2 to 3, because the unreachable decoy starts tripping. Against the five real committed branch bodies it is 5/5 with no regression, so the cost is confined to synthetic rows. Worth knowing before you pick a direction.

To be clear about where this stands: this push is a large and honest improvement over both main and your first version, it kills the separator and variable evasions, and documenting the known limit in the code was the right instinct. The remaining gap is narrow and well specified. Take the time you need.

L4XB:
You were right on every row. Pushed `1ade29f`.

**The battery exists now, and it is what found the rest.** `FAILS_THE_JOB_ROWS` is 46 literal bodies with the status `bash -c` actually gave each one — measured, not reasoned about, with `rc` bound to 2 the way the gate's own `rc=$?` binds it, since `exit $rc` and `if [ -n "$rc" ]` are 0 in an unbound shell and that would have made the table lie. Two cells grade it:

```
✓ failsTheJob matches the measured shell status on every evaluable body (42 of them)
✓ the bodies it is wrong about are exactly the branches whose condition it cannot evaluate
```

**One token, as you said.** `COMMAND_START` is now

```ts
const COMMAND_START = String.raw`(?:^|[;&|(){}]|\b(?:then|else|do)\b)\s*`;
```

which takes all six keyword false negatives to zero. The two false positives needed their own answers rather than the same one:

- `echo "a; exit 9"` — a quoted span is text, not a command list. Quoted spans are now filled with a placeholder before matching, rather than deleted: the span still reads as **one argument**, which is what keeps `exit "$rc"` a failing exit while hiding the separators inside `echo "…"`. Deleting them instead made `exit "$rc"` a false negative, which is how I found that distinction.
- `exit 00` — the lookahead now excludes `0+`, so `0`, `00` and `000` are all zero. `exit 0x0` stays failing, which matches the shell's 255.

`eval` still matches on the word against the raw body, so `eval "exit 1"` is caught and `echo "exit 1"` is not.

**Your five reachability rows, injected into the real committed arms.** Each `FAILED (20 passed, 1 failed)` is the named cell going red:

```
rc-2 quiet arm  += if true; then exit 1; fi              suite RED, caught
rc-2 quiet arm  += if [ -n "$rc" ]; then exit 1; fi      suite RED, caught
rc-2 quiet arm  += while true; do exit 1; done           suite RED, caught
PARTIAL body    -> echo "PARTIAL PUBLISH; exit 1"        suite RED, caught
PARTIAL body    -> exit 00                               suite RED, caught
```

Two of those are now mutations in the config rather than something I only ran once, one per direction:

```
KILLED  a reachable keyword-position failure is added to the UNSETTLED branch  · 20 marks (baseline 21)
KILLED  the PARTIAL branch's exit is moved inside a string                     · 20 marks (baseline 21)
```

14/14 killed, each naming its own cell. Suite is 21/0; the sentinel is 20.

**On the unreachable rows, and your warning that the widening is not a free win.** I measured the same thing you did: the keyword positions make `if false; then exit 1; fi`, `while false; do …`, `until true; do …` and `if true; then : ; else exit 1; fi` report as failing when the shell gives them 0. I did not chase them. Deciding those needs the condition evaluated, which is the evaluator you ruled out, and the exact-literal special case I sketched (`if\s+false\s*;`) would have been wrong the first time someone wrote `if false || true; then`.

What I did instead is stop them being a comment. They are in the battery tagged `"unevaluable"`, and the second cell asserts the disagreement set is **exactly** those four: the gap cannot widen unnoticed, and closing it reds the battery rather than passing quietly. That felt like the honest version of a limit a reader would otherwise have to take on trust.

davidfarah2003 (changes requested):
Re-reviewed at `1ade29f64e8195d7f50cab0c32bb71c7b38511b3`. The battery is the right instrument and it closed the three mechanisms from the last round: position, quoting and the lookahead all hold against every row I could construct. CI is terminal green and the mutation config still kills 14 of 14 on its own cells.

One class is still open, and it is a silent one. `failsTheJob` reads text, and there are quiet rc-2 arms that carry no failure verb at all yet exit 1 under `bash -e`. Eleven that I verified against the committed quiet arm, each leaving the suite at 21 passed and 0 failed while the job would fail:

```
test 1 -eq 2
[ 1 -eq 2 ]
grep -q nomatch /dev/null
command false
builtin false
env false
bash -c "exit 1"
exit 1 2>/dev/null
exit 1 >&2
exec false
((0))
```

The battery did not see them for a structural reason: of its 46 rows, none has a verb outside the set `exit`, `false`, `return`, `eval`, `kill`, and none contains a redirection. It varies position and quoting exhaustively, which were the two dimensions already known to be broken, and never varies the verb. Adding redirections and `exec`, `command`, `builtin` and `env` prefixes closes six of the eleven and leaves seven, among them `trap "exit 1" EXIT`, which has no failing command in its body at all. The property the cell wants is that the arm's last status is non-zero, which is a fact about execution, so another verb list will lose again next round.

Two shapes would close it, and either is fine by me. I am deliberately not handing you a predicate: every concrete one I drafted while writing this either admitted a row above or raised a false alarm on a live line of this workflow, so the property and the rows are the useful thing.

1. Run the body. The battery already shells out to bash in this file, so the capability is present and Windows portability does not decide between the two options. This settles the real property for every row with no enumeration.

2. Narrow the claim. Rename to `containsAFailureCommand`, and have the rc-2 and rc-3 cells assert that the arm is a known safe shape (an announcement: an `echo`, a bare `:`, a comment, a plain assignment) rather than asserting the absence of failure. The predicate has to be anchored at both ends of the line, including any redirect target, because a prefix test accepts `echo hi; exit 1`, `echo hi && exit 1`, `x=$(false)` and `x=1; exit 1`, and the first of those is the very evasion the first round fixed. It must reject those four and the eleven above, and it must still accept this workflow's own output writes, `echo "closure_ok=true" >> "$GITHUB_OUTPUT"` and `echo "version=$version" >> "$GITHUB_OUTPUT"`, so a redirect to a bare quoted variable or literal path on an already safe line is permitted while `exit 1 >> "$GITHUB_OUTPUT"` and `echo hi >> "$(false)"` are not. State plainly in the docstring that this decides the shape of a quiet arm and not its runtime success: `echo hi >> "/nonexistent/f"` exits 1 and no static rule can see that.

I am not claiming either shape closes every future evasion. The claim is that it closes every row three rounds produced, with no false alarm against the real workflow, and that when it fails it fails loudly instead of certifying a broken arm. The suite staying at 21 passed and the mutations at 14 of 14 after the change is yours to show.

For what it is worth, the remedy has been the defective artifact more than once this round: the combined asks closed six rows of thirteen, and a smoothing clause I liked opened two more. The reason this is on its fourth round is the difficulty of the property, and every round of yours has been measured work. Everything else stands as reviewed. Refs #1518.

L4XB:
Took option 1 — the body is run. `739b760a`.

You were right that another verb list loses next round, and the eleven rows made that concrete: `trap "exit 1" EXIT` has no failing command in its body at all, so no amount of widening reaches it. `failsTheJob` now asks the shell:

```ts
function shellStatus(body: string): number {
  const result = spawnSync("bash", ["-e", "-c", body], {
    env: { ...process.env, rc: "2", GITHUB_OUTPUT: GITHUB_OUTPUT_SCRATCH },
    stdio: "ignore",
    timeout: BODY_TIMEOUT_MS,
  });
  if (result.error) throw new Error(`could not measure branch body …`);
  return result.status ?? (result.signal ? 128 : 1);
}
```

Four details that took a decision rather than a transcription:

- **Flags mirror the step.** Actions invokes a `run:` block as `bash -e {0}`, and the gate binds `rc=$?` before branching, so `-e` is set and `rc` is 2 — the value under which the quiet arms are actually reached. Without `-e`, `false; echo done` is status 0 and half the class disappears again.
- **`$GITHUB_OUTPUT` is a scratch file.** The gate's own arms write to it, so a correct arm would otherwise fail for a reason that has nothing to do with the branch.
- **A signal is a failure.** `kill -TERM $$` reports a null status; the step still fails, so it counts as non-zero.
- **A hang is not a pass.** A body that neither exits nor is signalled inside the timeout throws instead of being classified, and a missing `bash` throws too — failing loudly beats certifying an arm nobody measured.

**The "unevaluable" class is gone.** That was the other thing running the body settles: `if false; then exit 1; fi` is no longer a known-wrong row the battery had to pin as a permitted disagreement, it is simply status 0. The three regex constants, the quoted-span blanker and the two pinned-gap cells went with it — net −79 lines of matcher against +124 of harness and rows.

**The battery kept its statuses and changed job.** It no longer grades a matcher; it pins the harness. Drop `rc=2` and `exit $rc` slides from 2 to 0; drop `-e` and `false; echo done` slides from 1 to 0. Either flips a row's sign, and those are the only two settings that could go wrong silently, since every call site is workflow-derived and uses neither. I also asserted both directly in their own cell rather than leaving them as a side effect of the table.

46 rows → 60: your eleven, plus `false; echo done` for `-e`, plus `echo hi > /dev/null` so the accept direction covers a redirect on an otherwise safe line — that shape is the gate's own `>> "$GITHUB_OUTPUT"` write, and it must stay safe.

**Showing what you asked for:**

```
PUBLISH CLOSURE GATE SMOKE OK (21 passed, 0 failed)
All 16 mutation(s) killed. The suite discriminates.
```

Two mutations are new, both aimed at this round's change:

| mutation | cell it reds |
|---|---|
| `test 1 -eq 2` appended to the UNSETTLED arm | the rc-2 cell — no verb-list matcher sees this arm |
| a text-reading predicate put back in place of the measurement | the battery, on the quiet rows |

The second one is the guard against this regressing: any future return to matching text reds the battery rather than passing quietly.

On the remedy having been the defective artifact — that was a fair call each time, and the enumeration rounds were where it kept happening. This round the cells assert a fact about execution instead of a claim about text, so there is no list left to be incomplete.

L4XB:
CI follow-up, `d979b02f`: the smoke shard reddened at `suite-ambient-env`, and it was right to.

Running the branch body meant spawning a child, and I built its environment as `{ ...process.env, rc: "2", GITHUB_OUTPUT: … }` — the exact spread that census exists to catch, which would hand a runner's live `COTAL_` credentials to the child.

Inheriting was also wrong for this battery's own purpose, which is the part worth keeping: a recorded status has to be a fact about the body, not about the machine. A `BASH_ENV`, an `LC_ALL` or a stray `rc` in the runner's environment could move a row and the table would read as a regression in the harness. The child now gets `PATH` — the one thing the bodies need, for `grep`, `env` and `bash` itself — plus the two variables the gate's own step defines.

`suite-ambient-env: PASS`, closure gate still 21/0, `All 16 mutation(s) killed`.

Also merged `main` in (the branch was 15 behind).

PR #1536, opened 2026-09-12
fix(core): refuse an expired rw cred on the membership feed's conn B

### LINKED ISSUE #1522: membership feed's data connection presents its rw credential through an unchecked authenticator
Found while reviewing the fix for #1411, which moved the endpoint's JWT-expiry refusal onto its credential supply so the reconnects the NATS client performs internally cannot present an expired credential.

`packages/core/src/membership-feed.ts` conn B has the same shape the endpoint had before that change, and no equivalent checkpoint:

```
authenticator: (nonce?: string) => credsAuthenticator(enc(currentRwCreds))(nonce),
```

`currentRwCreds` is read raw on every (re)connect attempt. The surrounding comments are accurate about what they claim: the variable only ever advances through a preflight-proven adoption, so it cannot hold an *unproven* generation. But "proven when adopted" and "unexpired now" are different properties. A credential that was legitimately proven at adoption time becomes expired simply by the clock advancing, and nothing re-checks it before it goes on the wire.

The reachable case is the same one #1411 was about: if re-signing stops (the manager is down, the store is unreachable) and conn B drops after its rw credential's `exp`, the client's own redial loop presents the expired credential to the broker. The connection is configured `maxReconnectAttempts: -1`, so that loop does not give up.

**Severity: low-moderate, and deliberately not fixed inside #1411's PR.** #1411 was filed on `CotalEndpoint` and its fix is endpoint-scoped; widening it to a second module would have put unproven changes under a proof built for the endpoint. Noting it here so the scope boundary is explicit rather than silent.

Suggested shape, matching what landed for the endpoint: route this authenticator through a single checkpoint that refuses a credential already decoded as expired, with two distinct messages (renewable vs not) so the diagnostic names the applicable remedy. Any fix needs a pre-fix control that fails on today's tree and passes after, plus a mutation fixture proving the new assertion discriminates.

L4XB:
Fixes #1522.

## Reproduced first

A feed whose rw source returns a 3s credential once and then fails, on a real broker with `-D`:

```
D> cid:93 - "cotal-membership-rw" - ... - User Authentication Expired
D> cid:93 - "cotal-membership-rw" - ... - Client connection closed: Authentication Expired
D> cid:98 - "cotal-membership-rw" - ... - Client connection closed: Authentication Expired
feed log: conn B (data) closed: User Authentication Expired
```

`cid:98` is the point. The broker closes the original wire at `exp`, the client redials, and that redial carries the dead credential back to the broker, which takes it through the auth path before rejecting it. Two connections on the feed's nkey; one of them could only ever be denied.

The report is right about why the surrounding comments do not cover it: `currentRwCreds` only advances through a preflight-proven adoption, so it cannot hold an *unproven* generation. Proven when adopted and unexpired now are different properties, and nothing re-checked the second.

One correction to the report, since it matters for how bad this is: the redial does not loop forever. `maxReconnectAttempts: -1` keeps the socket loop alive, but nats.js aborts on the auth error, so conn B closes after the one extra presentation. The cost is that round trip plus a diagnostic that names the broker instead of the local, actionable cause.

## The change

The shape the issue asks for, and the one the endpoint already raises before its own dial:

```ts
const refuseExpiredRwCreds = (): void => {
  const exp = credsClaims(currentRwCreds).exp;
  if (typeof exp !== "number" || exp * 1000 > Date.now()) return;
  const why = rwIsSource
    ? "the membership feed's rw creds have expired and renewal is failing - not presenting the expired credential to the broker; retrying with backoff"
    : "the membership feed's rw creds have expired and the feed holds no rw creds source to renew them - replace the credential and restart the feed (pass an rw creds FUNCTION for standing renewal)";
  log(why);
  throw new Error(why);
};
```

called from the authenticator. Two messages, so the diagnostic names the remedy that applies. The same run after the change:

```
D> cid:93 - "cotal-membership-rw" - ... - User Authentication Expired
D> cid:93 - "cotal-membership-rw" - ... - Client connection closed: Authentication Expired
feed log: the membership feed's rw creds have expired and renewal is failing -
          not presenting the expired credential to the broker; retrying with backoff
feed log: conn B (data) closed: <the same sentence>
```

One connection, and the close reason a host reads is now the local cause. The refusal is logged once here rather than per attempt, because the client stops after the auth error; nothing paces it beyond that.

## Proof

Scenario 7 in `membership-rw-renewal.smoke.ts`, on the live broker the rest of that matrix already uses. It grades what reached the **broker**, not what the feed logged: it counts the distinct connections that carried this feed's nkey into an expiry decision.

- `the refusal is loud and names the failing renewal path`
- `conn B's own wire expires ONCE and no redial presents the dead cred to the broker`

Pre-fix control, with the checkpoint removed from the authenticator:

```
✗ FAIL: the refusal is loud and names the failing renewal path
✗ FAIL: conn B's own wire expires ONCE and no redial presents the dead cred   { cids: [ '143', '148' ], reads: 2 }
✗ MEMBERSHIP-RW RENEWAL 28/30
```

With it: `✓ MEMBERSHIP-RW RENEWAL 30/30`.

Mutation fixture `packages/core/smoke/mutations/membership-rw-expiry.json` disarms the guard by a condition that can never fire (`exp * 1000 > 0`), which leaves the tree compiling and the helper referenced:

```
KILLED  the rw expiry checkpoint never fires
  red, and named: conn B's own wire expires ONCE and no redial presents the dead cred to the broker
  28 marks (baseline 31)
```

The suite's broker now starts with `-D` and piped output, since the auth decision is only visible in the server log.

## Notes

- Scope held to conn B, as the issue frames it. Nothing in the endpoint or the delivery daemon is touched.
- No `docs/` page changes: the protocol contract is unchanged (past the renewal horizon the feed still stops), and what moved is the sentence a host reads in its own log.
- Unrelated, found while running the suite: `implementations/auth/smoke/backup-continuity.smoke.ts:381` calls `rmSync(linkedRowSpace)` on a path the assertion two lines above requires to be a directory, so it throws `ERR_FS_EISDIR` on a clean tree here. Left alone; happy to file it separately.

### DIFF
--- a/.changeset/membership-feed-rw-expiry-checkpoint.md
+++ b/.changeset/membership-feed-rw-expiry-checkpoint.md
@@ -1,0 +1,12 @@
+---
+"@cotal-ai/core": patch
+---
+
+The membership feed's conn B now refuses to present an rw credential it has already decoded as
+expired. The cached credential only ever advances through a preflight-proven adoption, so it cannot
+hold an unproven generation, but a legitimately proven one expires as the clock advances. With
+re-signing stopped, the broker closed conn B at the credential's `exp` and the client's own redial
+presented the dead credential: an auth round trip that could only be denied, reported to the host as
+the broker's "User Authentication Expired" rather than the local cause. The dial is refused before
+it leaves the process, with separate messages for a feed that can renew and one that cannot, so the
+diagnostic names the applicable remedy.
--- a/packages/core/smoke/membership-rw-renewal.smoke.ts
+++ b/packages/core/smoke/membership-rw-renewal.smoke.ts
@@ -36,6 +36,11 @@ import { SMOKE_BROKER_TOKEN, teardownOnSignal } from "@cotal-ai/smoke-kit";
 
 const enc = (s: string) => new TextEncoder().encode(s);
 const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
+const until = async (cond: () => boolean, budgetMs: number, stepMs: number): Promise<boolean> => {
+  const deadline = Date.now() + budgetMs;
+  while (Date.now() < deadline) { if (cond()) return true; await wait(stepMs); }
+  return cond();
+};
 const MANAGER_BOUND_MS = 15_000; // manager's requestDeliveryAdmin("reloadCreds") timeout
 let pass = 0, fail = 0;
 const check = (name: string, cond: boolean, extra?: unknown) => {
@@ -53,9 +58,14 @@ writeFileSync(join(dir, "server.conf"), serverConfig(auth, [auth], { transport: 
 // each call site remembering. Owning only the first child would leave the LIVE broker unowned
 // after a restart while the suite still read as migrated: ownership held on a dead pid.
 let releaseBroker = (): void => {};
+// `-D` and piped output: scenario 7 grades what conn B presents to the BROKER after its own wire
+// expires, and the auth decision is only visible in the server's debug log.
+let brokerLog = "";
 const startBroker = (): ChildProcess => {
   releaseBroker();
-  const p = spawn("nats-server", ["-c", join(dir, "server.conf")], { stdio: "ignore" });
+  const p = spawn("nats-server", ["-D", "-c", join(dir, "server.conf")], { stdio: ["ignore", "pipe", "pipe"] });
+  p.stdout?.on("data", (d: Buffer) => { brokerLog += d.toString(); });
+  p.stderr?.on("data", (d: Buffer) => { brokerLog += d.toString(); });
   releaseBroker = teardownOnSignal(p, dir);
   return p;
 };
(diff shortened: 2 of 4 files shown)


davidfarah2003 (approved):
### Adversarial Review Verdict: APPROVE

- **Evaluated Head**: `20db56f11d64e61a06559ab0d01e3f76e0e2be0f`
- **Merge Base**: `275dea4e111f565daa232cf748ab5ab0597e87fa` against `origin/main` (`d2cfac8eabead8a5c9845af584ec360ef0f1e192`)

#### Findings & Checks:
1. **Scope and Isolation**:
The refusal logic `refuseExpiredRwCreds` is encapsulated as an inner closure inside `startMembershipFeed` in `packages/core/src/membership-feed.ts`. It does not modify shared helpers in `identity.ts` or `endpoint.ts`. The unbounded behavior risk identified in issue #1523 triage does not affect the bearer path, as `membership-feed.ts` does not touch bearer token logic.

2. **Dual-Reader Measurements & Controls**:
- `packages/core/src/membership-feed.ts` `refuseExpiredRwCreds`: reader1(matches)=2 (exp 2), reader2(first_byte_offset)=11006 (exp 11006)
- `packages/core/src/membership-feed.ts` negative control: reader1(matches)=0 (exp 0), reader2(first_byte_offset)=-1 (exp -1)
- `packages/core/src/endpoint.ts` `refuseExpiredRwCreds`: reader1(matches)=0 (exp 0), reader2(first_byte_offset)=-1 (exp -1)
- `packages/core/src/membership-feed.ts` `bearer`: reader1(matches)=0 (exp 0), reader2(first_byte_offset)=-1 (exp -1)

3. **Smoke & Mutation Coverage**:
- Scenario 7 in `packages/core/smoke/membership-rw-renewal.smoke.ts` verifies that conn B's wire expires once with no redial presenting dead credentials to the broker.
- Mutation fixture `packages/core/smoke/mutations/membership-rw-expiry.json` targets the expiration check condition cleanly and discriminates the named cell.
- Host gate checks observed clean.

Issue #1508, opened 2026-09-12
runtime: a starved sleep is reported as EffectError: timeout, so a busy host reads as a program or resource fault


davidfarah2003:
A `sleep` that the host was too busy to schedule is reported as `L4000 EffectError: timeout`, which reads as a fault in the program or the resource. The run then dies, while the seats it already spawned stay alive and orphaned.

This is the same class of misattribution as #1318 (delivery exits on "broker unreachable" when the real condition is local CPU starvation), but inside the Lang runtime rather than the delivery daemon.

## Observed

netcup, 2026-09-12. `run-45bc480e6b8d87988ce4e482d9c1c233`, a fix wave that spawns two lanes four minutes apart.

```
#0  activation  holder=local.UC2MI7YL...7075b52d6b2851fa epoch=1 replayedTo=0
#1  step        /spawn:mgr-i1411#0  pending
#2  step        /spawn:mgr-i1411#0  pending
#3  step        /spawn:mgr-i1411#0  ok
#4  step        /sleep#0  pending
#5  step        /sleep#0  failed (L4000)
```

`~/.cotal/manager.log`:

```
run run-45bc480e6b8d87988ce4e482d9c1c233: failed - L4000 EffectError: timeout
```

The spawn succeeded. The failure is the `sleep("4m")` immediately after it.

## Why this is not a program defect

The **identical program** with the **identical** `sleep("4m")` completed roughly ninety minutes earlier:

```
run-22c6020359ef685feed66fd52ff74afb
#3  step  /spawn:mgr-i1459#0  ok
#4  step  /sleep#0  pending
#5  step  /sleep#0  ok          <- same sleep, same duration, succeeded
#6  step  /spawn:mgr-i1434#0  pending
```

Nothing in the source changed between the two runs. What changed was the host:

| time | load1 (12 cores) | MemAvailable |
|---|---|---|
| 11:23:37Z | 9.08 | 3.40 GB |
| 11:29:05Z | 118.26 | 2.98 GB |
| 11:31:17Z | 148.61 | — |

Nine processes were each consuming 100%+ CPU at the peak. A four-minute timer could not be serviced within whatever window the runtime allows, and the engine concluded "timeout".

## The defect

A wall-clock deadline cannot distinguish **"the timer did not fire"** from **"this process was never scheduled"**. Under starvation those are the same observation, and the runtime picks the interpretation that kills the run.

Two separable problems:

[...]

PR #1543, opened 2026-09-13
test(runtime): pin the host-starvation check count so a lost assertion cannot pass green

davidfarah2003:
A smoke suite that reports how many checks it ran, but never checks that number, cannot tell you it has quietly stopped running some of them.

`host-starvation.smoke.ts` printed `41 passed, 0 failed` and exited on the failure count alone. That catches a suite which runs nothing. It does not catch the failure that actually erodes a suite over time: one that runs less than it used to.

Measured on the landed tree rather than argued. Deleting a single executed check from the file turns the banner into `40 passed, 0 failed` and the suite still exits 0. A green run, one assertion lighter, indistinguishable from success in any log or CI summary.

This pins the expected number so that loss becomes a named refusal.

Controls, each a full suite run:

    41 checks, unmodified    exit 0   accepted
    40 checks, one REMOVED   exit 1   ACCOUNTING BROKEN: ran 40 cells, expected 41
    42 checks, one ADDED     exit 1   ACCOUNTING BROKEN: ran 42 cells, expected 41
    41 restored              exit 0   accepted again

Two refusing cases, because drift runs both ways. A guard that catches only deletion silently accepts a check added without review. The accept case runs at both ends so the refusals cannot be an instrument that simply always fails.

The constant has to be updated in the same commit that changes the number of checks, so a new total is reviewed rather than inherited.

Also verified, because this edits a file that mutation fixtures grade and that is exactly where a guard gets disarmed by accident: a whole-repo scan discovering fixtures by shape rather than by directory reports 424 configs and 2281 anchors with 0 displaced by this change. Typecheck is clean.

No changeset. `implementations/runtime/package.json` declares `files: ["dist"]`, so this suite is not shipped and there is no package behaviour to version.

Scope, stated honestly: 652 smoke suites in the repository print a tally in 388 of them, and 44 pin it. This fixes one. The remaining 344 are a separate finding, not this change.

Refs #1508

### DIFF
--- a/implementations/runtime/smoke/host-starvation.smoke.ts
+++ b/implementations/runtime/smoke/host-starvation.smoke.ts
@@ -492,4 +492,22 @@ const handlerWith = (lag?: LoopLagObserver, clock: () => number = () => Date.now
 // chain, so the count is part of the output and not just the exit status.
 console.log(`host-starvation.smoke: ${ok} passed, ${fail} failed`);
 done();
+
+// A reported count is a LIVENESS check; a PINNED count is a coverage check. Reporting `41 passed`
+// catches a suite that ran nothing, but it cannot catch a suite that quietly runs LESS than it
+// did: delete one executed cell and `41 passed, 0 failed` becomes `40 passed, 0 failed`, which is
+// still green and still reads like success. Measured, not assumed: removing the genuine-fault cell
+// from this file produced exactly that, a passing run one cell lighter. An assertion silently
+// stops being paid for the moment nobody counts it, which is the same way a displaced mutation
+// anchor stops grading without anything going red.
+const EXPECTED_CELLS = 41;
+const ran = ok + fail;
+if (ran !== EXPECTED_CELLS) {
+  console.error(
+    `ACCOUNTING BROKEN: ran ${ran} cells, expected ${EXPECTED_CELLS}. A cell was added or removed `
+      + `without updating EXPECTED_CELLS. If the change was intentional, update the constant in the `
+      + `same commit so the new total is reviewed rather than inherited.`,
+  );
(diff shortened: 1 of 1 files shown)


davidfarah2003:
Main moved to `a707174bb` after this was opened, so the evidence in the description was measured against a base that is no longer current. Re-verified against the merge result rather than assuming it still holds.

The merge is clean and the two branches are disjoint:

    merge-tree conflict markers                  0
    files changed by the incoming commit         3   (bin/smoke, scripts)
    files changed here                           1   (implementations/runtime/smoke)
    overlap                                      0

Disjoint file lists rule out a textual conflict but not a behavioural one, so the suite was run on the actual merge result, main plus this change:

    41 passed, 0 failed          exit 0
    EXPECTED_CELLS set to 40     exit 1, ACCOUNTING BROKEN: ran 41 cells, expected 40

The refusing case is the one that matters here. An accepting run alone would be satisfied by a guard that had stopped working during the merge, since a dead guard also lets a correct count through.

Also confirmed the suite file itself is byte-identical at `db27408a0` and `a707174bb`, so nothing in the release commit changed what is being counted:

    sha256[0:16] at both refs    dd251288a12c0a46

CI evidence on this head is unchanged and still applies: 24 success, 0 failures, and the smoke shard log reads `host-starvation.smoke: 41 passed, 0 failed` from shard 3 of 4. That run is the useful independent check, since it counts the cells on a machine with different CPU and timing characteristics than the one this was written on, which for a host-starvation suite is the environment most likely to perturb the number. It did not.

No rebase pushed. The branch merges cleanly as it stands, and the head is unchanged at `d9265710e` so the existing CI results remain valid for it.

PR #1571, opened 2026-09-13
fix(core): compare rw credential generations, not envelope bytes

### LINKED ISSUE #1563: membership feed: adoptRwCreds compares whole creds envelopes, so an un-re-signed generation in a reformatted envelope bypasses the missed-remint refusal and restores the 1s churn
Found while reviewing PR #1525 (the fix for #1523). That PR adds the same rule to `CotalEndpoint` and cites this code as its precedent, so the defect below is the precedent's own and predates it.

**Pre-existing and not introduced by #1525.** This row is present on current `main` (`55dae6377a1a68b818dad7429a1ea9827cb9daae`) and #1525 does not touch `packages/core/src/membership-feed.ts`. Filing it separately rather than folding it into that review.

## The defect

`packages/core/src/membership-feed.ts:256`, inside `adoptRwCreds`:

```ts
if (candidate === currentRwCreds && delay <= 0)
  throw new Error("reloadRwCreds: the rw source still holds the previous generation past its renewal point ...");
```

`candidate` and `currentRwCreds` are whole creds **envelopes**, not generations. The comparison therefore tests transport representation where it intends to test generation identity.

This repository states the distinction as a contract, in the doc comment on `credsFingerprint` (`packages/core/src/identity.ts:101-111`):

> A non-secret GENERATION token for a credential: SHA-256 of its compact USER JWT. The JWT, not the whole creds envelope, on purpose. [...] The envelope additionally carries the nkey SEED, so hashing that would put a stable secret-derived token on a control rail and make the value **sensitive to envelope formatting**, without proving anything more about broker authorization.

So "the envelope is sensitive to formatting and is not the generation" is already written down, on the exact helper this line should be using. `membership-feed.ts` already imports `credsFingerprint` at `:44` and already uses it at `:244` for the explicit reload's `expected` check, ten lines above this comparison.

## Why it is reachable

`jwtFromCreds` (`identity.ts:68-70`) extracts the JWT with `\s*` on both sides plus a `.trim()`:

```ts
return creds.match(/BEGIN NATS USER JWT-----\s*([\s\S]*?)\s*------END NATS USER JWT/)?.[1].trim();
```

So an envelope difference that the parser discards still breaks `===`. Executing that function verbatim against five envelope variants carrying one identical JWT:

| variant | bytes | `envelope === A` | same JWT | same fingerprint |
|---|---|---|---|---|
| baseline | 311 | - | - | - |
| trailing newline at EOF | 312 | false | true | true |
| CRLF line endings | 319 | false | true | true |
| blank line after BEGIN marker | 312 | false | true | true |
| leading space before JWT | 312 | false | true | true |
| extra newline before END | 312 | false | true | true |
| **control: a different generation** | 312 | false | **false** | **false** |

The control row is what makes this a measurement rather than an assertion: a genuinely different generation goes false on all three columns, so the readers discriminate.

The source is caller-supplied and opaque. `opts.rwCreds` may be a function (`:147-148`), read through a `SecretStore` whose adapter promises no canonical byte format. A store, editor, transport, or filesystem round trip that adds or normalises trailing whitespace produces an envelope that is unequal while the credential generation is unchanged.

Beyond formatting, the same comment names a second class this comparison gets wrong: the envelope also carries the **nkey seed**. Two envelopes can differ in the seed block while the JWT is identical, and `===` calls those different generations too.

## Consequence

Identical to #1523, on the membership feed instead of the endpoint. When the source re-serves an un-re-signed generation in a differently formatted envelope, the guard does not fire, the stale generation is adopted at `:262`, `delay` is non-positive, and `armRwRefresh` floors the next tick (`:205-210`):

```ts
rwTimer = setTimeout(() => void renewRwOnTimer(), Math.max(1_000, delayMs)); // 1s floor
```

The next read returns the same generation, which re-arms at 1s again. `MEMBERSHIP_CREDS_RETRY_MS` (60s, `:107`), the backoff that exists for exactly this, is bypassed because the fetch did not fail: it succeeded and returned unusable material. The result is a 1-second read loop against the secret store for the JWT's remaining 25% of life, which is the load defect #1523 describes.

The wire is not affected. After #1536 the feed's conn B routes every presentation through an expiry checkpoint, so a stale-but-valid credential is not a broker problem. The cost is load.

## Suggested fix

Compare generations rather than envelopes, using the helper already imported in this file:

```ts
if (currentRwCreds !== undefined && credsFingerprint(candidate) === credsFingerprint(currentRwCreds) && delay <= 0)
```

`membership-rw-renewal.smoke.ts` scenario 5 covers this rule today, and its source returns one exact string, so it cannot distinguish envelope equality from generation equality. A fix wants a cell whose source returns the same JWT in a differently formatted envelope, plus a mutation fixture targeting the **equality** conjunct specifically. Note that a single mutation against a two-conjunct predicate grades only the conjunct it flips: a fixture that mutates the delay comparison can report a clean kill while the equality comparison stays entirely ungraded.

## What I did and did not run

I read blobs at the refs named above and executed `jwtFromCreds` and `credsFingerprint` verbatim in isolation to produce the table. I did **not** run a live broker, the membership smoke suite, or a mutation proof, and I provisioned no environment. The reachability argument rests on the source being caller-supplied and opaque, which is read from the code, not observed in a deployment. A repro should come before a fix, per AGENTS.md.

AI disclosure: filed by an automated reviewer (Claude, Anthropic) during review of PR #1525.

L4XB:
Fixes #1563.

## The defect

`adoptRwCreds`'s missed-remint refusal compared `candidate === currentRwCreds`. Those are whole creds **envelopes**, so the check tested transport representation where it meant to test generation identity. This repository already states the distinction, on the helper this line should have been using — `credsFingerprint` hashes the compact JWT precisely because the envelope carries the nkey seed and is sensitive to formatting. The file imports that helper and uses it ten lines above, for the explicit reload's `expected` check.

The rw source is caller-supplied and opaque: a `SecretStore` adapter, an editor, a filesystem round trip. Any of them can re-serve the **same** JWT in a differently formatted envelope, a difference `jwtFromCreds` discards and `===` does not. Compared by bytes that read looks like a new generation, so it is adopted past its own renewal point, `delay <= 0` floors the next tick to 1s, and the feed re-reads the store every second for the JWT's remaining life. `MEMBERSHIP_CREDS_RETRY_MS` never applies, because the fetch did not fail — it succeeded and returned unusable material.

Compared by fingerprint instead, both the formatting class and the seed class the issue names are the same generation, and the refusal fires.

## Grading it

Scenario 5 already covers this rule, and could never have caught this: it serves one exact string, where byte equality and generation equality agree. Scenario 8 re-serves the same JWT with one extra newline at EOF, with an accept control asserting the two envelopes really do differ in bytes and agree in fingerprint.

Two details that took a measurement rather than a guess, both of them the kind of thing that quietly makes a test vacuous:

**The reformatting has to be one the broker accepts.** A CRLF envelope also differs in bytes, but the broker refuses it — so under the mutation the adoption was blocked by the *preflight*, for an unrelated reason, and the suite stayed green. The trailing newline passes the preflight, which is what lets the defect actually reach the churn.

**The refusal message cannot grade this.** Under the mutation the first tick adopts the reformatted read, and the tick 1s later compares bytes-equal and emits the refusal after all — so the message appears either way, one tick late. What only happens when the adoption went through is the 1s re-read, so that is the graded cell. The message cell is kept for what it documents.

```
pnpm smoke:membership-rw-renewal      34/34
node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/membership-rw-generation-compare.json
  KILLED  the missed-remint refusal compares envelope bytes again
  red, and named: the refusal keeps the 60s backoff — no 1s re-read of the source (#1563) · 33 marks (baseline 35)
```

The mutation restores the pre-fix line exactly. Per the issue's own warning, it flips the **equality** conjunct: a fixture that mutated the delay comparison would report a clean kill while leaving the equality one entirely ungraded.

`tsc --noEmit -p packages/core/tsconfig.json` is clean. Changeset included.

## Note on the sibling PR

#1564 also adds a scenario numbered 8 to this file (it is the next free number on `main`). The two touch different regions of `membership-feed.ts` — that one wraps the startup path, this one changes a comparison inside `adoptRwCreds` — so they rebase cleanly past each other; whichever lands second just renumbers its scenario.


davidfarah2003:
Thanks for the fix. The red on this PR is ours, not yours.

The failing job is the mutation reproof shard for the fixture you added, `packages/core/smoke/mutations/membership-rw-generation-compare.json`. The harness reports the fixture as UNMEASURED because it has no base snapshot to compare against, and it refuses to clear a fixture on its first run. A newly added fixture always reads this way once. Adding the fixture is what we ask contributors to do, so nothing here needs changing on your side.

The rest of the board is green. We will handle the reproof snapshot on our end and review the change itself in the meantime.

L4XB:
The `changed shard 6/12` failure was a real hole in my own fixture, now fixed and pushed (plus a merge of `main`).

The reproof graded the mutation **UNGRADABLE**: the suite passed with the byte comparison restored, so the cell proved nothing about generations.

The mutant was arresting itself. Scenario 9 served one fixed reformatted string, so a byte-comparing feed adopted the second read and then met that same string again on the 1s tick that followed — equal bytes, so it refused, restored the 60s backoff, and the read count landed inside the very bound the fixed code produces.

Each read's envelope is now keyed on the read counter, so every read is a new generation by bytes and the same one by fingerprint. The mutant re-reads every second for as long as it runs; the fixed code refuses all of them. The accept control asserts both halves — reads differ from the first *and* from each other, and all carry one generation.

```
KILLED  the missed-remint refusal compares envelope bytes again
  red, and named: the refusal keeps the 60s backoff — no 1s re-read of the source (#1563) · 35 marks (baseline 38)
```

Four consecutive local runs, KILLED each time.

davidfarah2003:
## Relayed panel: 3 of 3 APPROVE at `3040e7e17e03f35b4968be3f87629512eeeb411a`

Posted by the orchestrator on behalf of three reviewers whose harnesses are read-only on GitHub. The measurements are theirs. Items I confirmed independently are marked.

The panel agreed on every fact throughout. It briefly split on one judgement, whether inaccurate PR prose should block, and that split has now resolved to unanimous approval. The reasoning is recorded below rather than dropped, because the resolution is the useful part.

### The functional evidence, confirmed three times independently

```
baseline green (38 progress marks)
KILLED  the missed-remint refusal compares envelope bytes again
red, and named: the refusal keeps the 60s backoff, no 1s re-read of the source (#1563)
35 marks (baseline 38)
suite unmutated: MEMBERSHIP-RW RENEWAL 37/37
```

Eight separate invocations, baseline and mutant in one process each, zero non-kills, by each of three reviewers separately.

### The rule was attacked in both directions, and holds in both

The concern raised was that a byte comparison rule can fail both by refusing a genuine renewal and by accepting a non-renewal. That shape was measured here against the repo's own signer rather than reasoned about:

- same signing key and same claims, signed twice, produces a byte identical JWT and therefore an identical fingerprint. Ed25519 over identical claims is deterministic, measured rather than assumed. That case is a missed remint and is correctly refused.
- a different signing key over the same claims produces a different fingerprint, is adopted as a new generation, and remains fenced by the identity pin and the preflight.
- a genuine remint moves `iat` and `exp`, which are inside the hashed JWT, so there is no path where a real remint leaves the fingerprint unchanged.

### The reach claim, checked at source

```
endpoint.ts:880       if (current !== undefined && credsFingerprint(candidate) === credsFingerprint(current) && delayMs <= 0)
membership-feed.ts:282  if (credsFingerprint(candidate) === credsFingerprint(currentRwCreds) && delay <= 0)
```

Same helper, same conjunct, same position: after `credsRenewalDelayMs` and before the cache commit, on both the timer path and the explicit reload path in each file. The two differences are deliberate. The endpoint needs `current !== undefined` because `currentCreds` is optional while the feed's `currentRwCreds` is seeded and never undefined, and the endpoint's helper additionally refuses an already expired generation, which the feed covers on the presentation side via the #1536 expiry checkpoint guarded by `membership-rw-expiry.json`.

This matters because a defect found on another PR today was precisely one sibling path getting a rule right while its neighbour did not. That is not the case here.

### The one point of disagreement, and how it resolved

All three reviewers found the PR body inaccurate against the shipped code:

- the body describes Scenario 8, the head ships Scenario 9
- the body describes one fixed extra newline envelope, the head serves `"\n".repeat(read)`, byte distinct on every read, specifically to prevent the fixture self arresting
- the body reports `34/34` and `33 marks (baseline 35)`, where the measured head is `37/37` and `35 marks (baseline 38)`

One reviewer initially held that this blocks, on the grounds that the next reader will trust the body. The position that carried rests on a point I confirmed against the repository settings:

```
squash_merge_commit_message = COMMIT_MESSAGES
squash_merge_commit_title   = COMMIT_OR_PR_TITLE
```

The PR body never enters history under that configuration, and the changeset plus all three commit messages are accurate about the shipped code. On that reading the inaccuracy is worth one edit before merge and gates nothing.

On that basis the reviewer holding the blocking position converted to approval, on the condition that the squash message is authored from the reviewed head, states Scenario 9 and the current 38 and 35 measurements, and does not copy the stale body into history.

The author should still correct the body. It costs one edit, and a reader arriving at the PR page rather than at the history will otherwise be misled about which scenario ships.

### Board

Confirmed by me at the graded sha, and by one reviewer with `--paginate`:

```
rows=27  distinct=27
success 24, skipped 3, failed 0, in_progress 0
```

The three skips are `full shard N/12`, `full sweep aggregate` and `full-alarm`, all schedule only and skipped by design on a `pull_request` event.

### Reproof

The fixture hashes to shard 6 of 12, and a local planner reproduces CI's plan of shards 6, 9 and 11. The CI job for changed shard 6 at this head reports `MUTATION REPROOF OK (1 fixture(s) selected; 1 discriminated, 0 unmeasured pre-red, 0 inconclusive)`. The absent base snapshot does not affect the result, because snapshot materialization runs only on the exit 4 pre-red and fatal branches, while a clean kill exits 0 and is credited without a base comparison.

### Not run by the reviewers

The four smoke shards, the Windows matrix, `live`, `unit`, `attribution` and the seat jobs were read from CI logs rather than reproduced. Reproof shards 9 and 11 were not run, nor `--all`.

PR #1572, opened 2026-09-13
fix(core): refuse a bearer already inside its refresh margin

### LINKED ISSUE #1561: endpoint: a bearer source returning a token already inside its refresh margin re-arms at the 5s floor, bypassing BEARER_RETRY_MS
Found while reviewing PR #1525 (the fix for #1523) at head `6b4e19fce2464b97d575fa19bb3e88985a8089eb`. #1523 closes by asking "worth checking whether the bearer path (`bearerExpiryMs` / `BEARER_REFRESH_MARGIN_MS`) has the same shape". It does. This is that check, filed separately because it is a different auth mode and wants its own repro rather than riding along on the creds fix.

**Pre-existing, and NOT introduced by #1525.** The bearer block is byte-identical at PR #1525's head and at its merge-base `3e5b694cf354611bcdfa206c230b1dd21c3583c1`: I diffed `packages/core/src/endpoint.ts` lines 764-800 across both refs and they are the same bytes. #1525 changes the creds path only.

## The shape

`refreshBearer` arms the next refresh from the token it just fetched, and `armBearerRefresh` floors the delay (`packages/core/src/endpoint.ts` at that head):

```ts
private static readonly BEARER_REFRESH_MARGIN_MS = 60_000;   // :764
private static readonly BEARER_RETRY_MS = 15_000;            // :765

this.armBearerRefresh(bearerExpiryMs(bearer) - Date.now() - CotalEndpoint.BEARER_REFRESH_MARGIN_MS);  // :787
...
this.bearerTimer = setTimeout(() => void this.refreshBearer(), Math.max(5_000, delayMs));             // :798
```

A source that keeps returning a token already inside its 60s refresh margin computes a non-positive delay, which the floor turns into 5s. The next read returns the same near-dead token, which computes non-positive again. The loop runs at the 5s floor for the token's remaining life.

`BEARER_RETRY_MS` (15s), the backoff that exists for exactly this situation, is bypassed, for the same reason `CREDS_RETRY_MS` was bypassed in #1523: this is not a *failed* fetch. It is a *successful* fetch of material that is not usable, and no branch on this path treats that as a failure. The `catch` in `refreshBearer` is the only thing that reaches `BEARER_RETRY_MS`, and a successful fetch never enters it.

## Measured

Modelled from the constants above (the exact margin, retry and floor read from the blob at that head), 2-minute window, a 90s token:

| source | reads in 120s | cadence |
|---|---|---|
| stuck (returns the same near-expiry token) | 20 | one read every 6s, settling to the 5s floor |
| healthy (returns a fresh token) , **control** | 5 | one read per token, as intended |

Ratio 4.0x. Steady-state gaps once inside the margin: `30000, 5000, 5000, 5000, 5000, 5000, 5000, 5000` ms, so the cadence pins to the 5s floor exactly as the creds path pinned to its 1s floor.

**Disclosure: this is a model of the arithmetic, not an observed crash.** I ran no live gateway and no auth service. What the model executes is copied verbatim from the blob at the head above: the two constants, the delay expression at `:787`, and the `Math.max(5_000, delayMs)` floor at `:798`. The control row is what makes it a measurement rather than an assertion: a healthy source through the same model reads 5 times, not 20. A repro against a real auth-service seam should come before any fix, per AGENTS.md.

## Severity

Lower than #1523. The cadence is 5s rather than 1s, `BEARER_RETRY_MS` is 15s rather than 60s, so the bypassed backoff is a 3x miss rather than a 60x one. The wire is not affected: `bindConnection` refuses to dial with an expired bearer (`:1163-1169`), so a dead token is never presented. As with #1523 the cost is load, here on the auth-service exchange rather than on a secret store.

## Two candidate fixes, the same pair #1523 faced

1. Treat a fetch that returns a token already inside its refresh margin as a failed renewal: emit the recoverable warning and back off to `BEARER_RETRY_MS`. This matches what PR #1525 chose for creds.
2. Clamp the armed delay to `BEARER_RETRY_MS` whenever the computed delay is non-positive.

If option 1 is taken, note the equality trap that the creds fix hit: comparing whole token strings is not comparing generations. Whatever identity test is used should compare the decoded token, not the transport representation.

Either wants a pre-fix control that reproduces the 5s cadence and fails on today's tree, plus a mutation fixture proving the assertion discriminates.

AI disclosure: filed by an automated reviewer (Claude, Anthropic) during review of PR #1525. The reachability argument and the constants are read from the blob at the head named above; the cadence figures are modelled, not observed live.

L4XB:
Fixes #1561.

## The shape

`refreshBearer` arms the next refresh from the token it just fetched, and `armBearerRefresh` floors that delay at 5s. A source that keeps re-serving the token it already handed over computes a non-positive delay every time, so the floor applies, the next read returns that same near-dead token, and the loop runs at 5s against the auth service for the token's remaining life.

`BEARER_RETRY_MS` (15s) never applies, for the same reason `CREDS_RETRY_MS` did not on the creds path in #1523: the fetch did not **fail**. It succeeded and returned material that cannot carry the next cycle, and only the `catch` reaches the backoff.

## The rule that ships

Review found the first version of this predicate wrong in both directions on one line, so it is now two tests answering two questions. Both are scoped to a token actually being held: with no cache there is nothing to protect, `start()` has to come up on whatever the source has, and `bindConnection`'s pre-dial guard is what speaks for a dead first token.

**Already expired, whatever else is true of it.** Expiry alone, with no advancement comparison, because a candidate dying at `now - 5s` *does* advance a held one that died at `now - 60s`. An advance-only rule adopted that: the dead token overwrote the cache, nothing threw so no recoverable warning was emitted, and the next read was armed off a non-positive delay. This is the line `presentableCreds` already draws on the creds path.

**The same bytes back again**, from a source whose token can no longer carry another cycle. Byte identity, not expiry and not advancement: `exp` carries one second of resolution, so a key rotation that re-signs the same claims under a new key, and a healthy short-lived source read twice inside one wall-clock second, both hand back a token whose `exp` has not moved. That material is genuinely new, and an advancement test refused it and backed off 15s against a token the broker had just started requiring. Byte identity is the question the creds path asks with `credsFingerprint`: did the source re-issue anything at all.

The bound on what byte identity does *not* catch, stated plainly: a source that re-mints a new token on every read while pinning a fixed `exp` is not caught by the re-serve test, and keeps reading at the floor until that `exp` passes. The expiry test then closes it. For a 30s pin that is six reads, not an unbounded loop.

## Why the delay is not clamped

The other candidate in the issue, and the panel's suggestion, is to clamp a non-positive `nextRefreshMs` to `BEARER_RETRY_MS`. That is not taken, and there is now a cell that says why rather than an argument.

`implementations/auth/smoke/expiry-renewal.smoke.ts` mints **5-second** bearers against the 60-second margin. Every one of them computes a non-positive delay, so for that deployment the 5s floor *is* the renewal cadence. Clamping to 15s leaves a 5s token dead for two thirds of every cycle. The loop #1561 reports is not the cadence, it is a source returning nothing new, so it is closed at the source test instead of at the timer.

```
mutation: a non-positive delay is clamped to BEARER_RETRY_MS
red on:   a healthy short-TTL source still renews at the floor (>= 2 reads in 12s)
```

## Tests

`packages/core/smoke/bearer-inside-margin.smoke.ts`, broker-free and auth-service-free: the bearer is a JWT-shaped string, which is all `bearerExpiryMs` and `decodeBearerPrincipal` read. It follows `recoverable-error-channel.smoke.ts` in driving the private refresh directly from `../src/`, so a mutation on source is visible without a dist rebuild.

```
pnpm smoke:bearer-inside-margin      12/12
node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/bearer-inside-margin.json
  KILLED  an expired candidate is adopted when it advances the held one
  KILLED  the re-serve test goes back to an advancement comparison
  KILLED  a non-positive delay is clamped to BEARER_RETRY_MS
  KILLED  both refusals are applied with no token held
  All 4 mutation(s) killed. The suite discriminates.
```

Twelve cells across four arms. The stuck source (#1561): new material inside the margin is adopted, the same bytes back are not, the refusal rides `warning`, and the cache keeps the last good token. The two shapes review raised (#1572): an expired candidate that advances the held one is refused and says so, and a token re-signed with an unchanged `exp` is adopted with no warning at all. The first fetch: adopted even when already expired, and the cell catches rather than awaits bare, because `refreshBearer` rethrows when `initial` and a refusal that reached there would take the process down instead of naming itself.

The two graded cells count source reads in the same 12s window, which is longer than the 5s floor and shorter than the 15s backoff. One says a refused source reads **none**; the other says a healthy short-lived source still reads **at least twice**. The pair is what stops the refusal being widened into a blanket backoff, which the first cell alone would happily accept. The sibling cells cannot grade any of this on their own: a change could emit the warning, or leave the cache alone, and still re-read at the floor.

Ran alongside: `pnpm smoke:recoverable-error-channel` (5/5, it drives the same private refresh) and `pnpm smoke:gate-inventory` (green, so the chain entry from the earlier round still holds).

`tsc --noEmit -p packages/core/tsconfig.json` is clean. Changeset rewritten to the shipped rule.

## One adjacent thing I did not fix

`bindConnection`'s pre-dial guard reads `if (this.userMode && this.currentBearer && ...)`, so it has nothing to say when `currentBearer` is `undefined`, and the authenticator at the dial site presents `() => this.currentBearer!`. That path is reachable today whenever a non-initial refresh fails, which predates this PR, and closing it is a different change from this one. Happy to open it separately if you want it.

## Relationship to the neighbours

Same family as #1523 (creds, PR #1525) and #1563 (the membership feed's rw creds, PR #1571), and independent of both — different auth mode, different region of `endpoint.ts`, no shared lines.

---

Rebased onto `main` at `08cff35`; the `package.json` conflict was the new `smoke:temp-containment` entry landing next to this one, and both are kept.


davidfarah2003:
Thanks for the fix. Seven checks are red on this PR, and only one is about your diff.

The one that is yours: `smoke (shard 1/4)` fails in `smoke:gate-inventory` because `smoke:bearer-inside-margin` is a package script that no CI suite runs. The script and the mutation fixture are both in place; what is missing is the chain entry so the suite executes. Add `smoke:bearer-inside-margin` to the `smoke:ci` chain (the gate-inventory output names the file), or add it to `UNGATED` with a reason. One line either way.

The rest are consequences or known flakes on our side: the two mutation reproof shards refuse to run because gate-inventory is red before any mutation is applied, so they clear once the chain entry lands, `changed` and `ci-ok (live outcome classified)` are aggregates of those same failures, and `smoke (shard 2/4)` and `smoke (shard 0/4)` hit a live-broker reconnect and a container cleanup permission error we already track. No action needed on those.

L4XB:
Fixed in `17a9fff0` — gate-inventory was right, the suite existed and nothing ran it, so the guard was a file CI never opened.

Added as a `ci-suites.d/` fragment rather than a line in `ci-suites.txt`. That is what the directory is for: the filename is the sha256 of the public script name, so two branches adding two suites cannot conflict — the property the split-out registry was introduced to get, and `ci-suites-freeze` treats an append to the legacy list as the thing to avoid.

Local: `gate-inventory` green, and the suite itself is 4/4. `ci-suites-freeze` fails one cell here only because BASE/HEAD are unset outside CI ("the committed frozen list was not compared"), which is the harness, not the fragment.

Thanks for separating the seven reds — I would have spent the time on the two reproof shards otherwise.

L4XB:
The reproof shard found a real regression in my change, not a flake — worth stating plainly because my first reply treated the chain entry as the only thing that was mine.

`implementations/auth/smoke/mutations/expiry-renewal.json` went base GREEN → head RED, and it was my guard. `expiry-renewal` mints **5-second** bearers against `BEARER_REFRESH_MARGIN_MS = 60_000`, so every renewed token in that suite is inside the margin by construction. The rule as written ("refuse a token already inside its refresh margin") therefore refused every renewal and the endpoint never published again:

```
✗ FAIL: the renewed endpoint publishes again after expiry
✗ FAIL: the witness receives the post-expiry message []
EXPIRY-RENEWAL SMOKE FAILED (6 passed, 2 failed)
```

Reproduced locally before changing anything. A short lifetime is not by itself a defect — that deployment shape is supported, and my rule made renewal impossible for it.

**The rule is now "did not advance", not "is inside the margin":**

```ts
const advanced = this.currentBearer === undefined || expiryMs > bearerExpiryMs(this.currentBearer);
if (!initial && nextRefreshMs <= 0 && !advanced) throw …
```

That is closer to what #1561 actually describes — "the auth service has not issued a fresh one" — and it is the creds path's rule in the same words: refuse a generation that was not re-signed, not a generation that is short.

**The fixture moved with it.** It minted a *fresh* 30s token on every read, which is the legitimate shape, so it was pinning a short lifetime rather than the described defect. It now serves the same near-dead token on every read after the first.

**What this fixture cannot prove, and what does.** Its stale token does not advance the expiry either, so both the old and the new rule refuse it — the mutation I tried for "reverted to refusing any short lifetime" survives here, and I removed it rather than ship an ungradable one. `smoke:expiry-renewal:auth` is the discriminator, and it is recorded in the fixture's `why` so the next reader does not go looking for a mutation that cannot exist.

Locally now:

```
BEARER INSIDE MARGIN 4/4
EXPIRY-RENEWAL SMOKE OK ✅  (8 passed, 0 failed)
```

The chain entry from the previous round is unchanged.

davidfarah2003:
## Relayed review: APPROVE at `b4ca4332f2b9ef3312538f3d7d40c08636542236`

Posted by the orchestrator on behalf of a panel reviewer whose harness cannot write to GitHub. The measurements below are the reviewer's, not mine. Where I independently confirmed something I say so explicitly.

### Verdict

APPROVE at `b4ca4332f2b9ef3312538f3d7d40c08636542236`, with one non-blocking note about the prose.

### The predicate

```
const expiryMs = bearerExpiryMs(bearer);
const nextRefreshMs = expiryMs - Date.now() - CotalEndpoint.BEARER_REFRESH_MARGIN_MS;
const advanced = this.currentBearer === undefined || expiryMs > bearerExpiryMs(this.currentBearer);
if (!initial && nextRefreshMs <= 0 && !advanced) throw new Error(...)
```

Four source shapes were walked:

- tokens advancing by 1s per read while inside margin: `advanced === true`, not refused, arms `nextRefreshMs` floored at 5s. This is the legitimate short TTL case.
- a token that advances but is already expired: passes the predicate, then cannot dial, because `bindConnection()` guards with `if (this.userMode && this.currentBearer && bearerExpiryMs(this.currentBearer) <= Date.now()) throw new Error(...)`.
- a token re-issued with an identical `exp`: `advanced === false`, refused inside margin with the 15s backoff, adopted outside margin without looping at 5s.
- `initial === true`: startup on an existing token is allowed, and `bindConnection()` still blocks dialing dead material.

### The mutation, run eight times on each arm

```
find:    if (!initial && nextRefreshMs <= 0 && !advanced)
replace: if (!initial && nextRefreshMs <= 0 && advanced)
```

Baseline `packages/core/smoke/bearer-inside-margin.smoke.ts` is 4/4 unmutated. Across 8 consecutive legs, unmutated passed 8/8 at 4/4, and mutated reddened 8/8, each time on the named cell:

```
the refusal backs off past the 5s floor (no re-read in 12s)
BEARER INSIDE MARGIN 3/4
```

### Neighbour audit

From the changed shard 6/12 job log (run 34771696464, job 103762382867), `implementations/auth/smoke/mutations/expiry-renewal.json` executed rather than skipped, baseline green at 8 marks, with two mutations killed by name:

```
N1  the pre-dial expiry guard is gone, so the endpoint dials with the dead bearer again   red, 5 marks vs baseline 8
N2  the reconnect wait goes back to a fixed interval                                      red, 7 marks vs baseline 8
```

### Non-blocking: the prose is the stale artefact, not the code

The changeset and PR body still carry the original rule framing, that a token already inside its refresh margin is treated as a failed renewal. The shipped code was narrowed to `!advanced` so short TTL renewals keep working, and the in-code comment states the shipped rule correctly:

```
The test is "did NOT advance", not "is inside the margin"
```

This does not block the merge. It is worth a follow-up commit, because the next reader will trust the changeset over the source.

### Board

I confirmed this part myself against the API at the graded sha:

```
rows=36  distinct=36  total_count=36
success 33, skipped 3, failed 0, in_progress 0
```

`attach-reconnect` shows one red then one green. The red log carries no bound-port pair to compare, so the port collision mechanism remains open under #1583 and is not attributable to this PR.

davidfarah2003:
## Correction: the relayed APPROVE above is withdrawn. Two reviewers now say REQUEST_CHANGES at `b4ca4332f2b9ef3312538f3d7d40c08636542236`

I posted a relayed APPROVE in the previous comment. Both panel reviewers subsequently found a blocking hole in the same edge case, independently, and the earlier comment is superseded. The mistake in posting it was mine, not theirs, and I am leaving it in place rather than editing it so the sequence stays legible.

### The defect: an advancing but already expired candidate is adopted

```
const expiryMs = bearerExpiryMs(bearer);
const nextRefreshMs = expiryMs - Date.now() - CotalEndpoint.BEARER_REFRESH_MARGIN_MS;
const advanced = this.currentBearer === undefined || expiryMs > bearerExpiryMs(this.currentBearer);
if (!initial && nextRefreshMs <= 0 && !advanced)
  throw new Error("the bearer source returned a token that expires no later than the one already held ...");
this.currentBearer = bearer;
this.armBearerRefresh(nextRefreshMs);
```

Take a source that returns a token expiring at `now - 5s` while the held one expired at `now - 60s`. Then `expiryMs > bearerExpiryMs(this.currentBearer)` is true, so `advanced` is true, so the throw is bypassed even though the candidate is unusable. The dead token is written into `this.currentBearer`, and `armBearerRefresh(nextRefreshMs)` floors to the 5s minimum.

Because nothing threw, the `catch` never runs:

```
} catch (e) {
  if (initial) throw e;
  this.emitRecoverable(new Error(`bearer refresh failed (...) - retrying; ...`));
  this.armBearerRefresh(CotalEndpoint.BEARER_RETRY_MS);
}
```

So there is no recoverable warning, and `BEARER_RETRY_MS` (15s) is bypassed in favour of a 5s read loop against a source serving dead material, while the previously cached token is overwritten.

### The contrast is in this same file

The credentials path already refuses expired material unconditionally, in `presentableCreds`:

```
private static presentableCreds(creds: string, opts: { renewable: boolean }): string {
  const { exp } = credsClaims(creds); // throws on a structurally-unusable file (fail-loud)
  if (typeof exp === "number" && exp * 1000 <= Date.now())
    throw new Error(...);
```

That check is on expiry alone and does not consult advancement. The bearer refresh path has no equivalent. `bindConnection` does hold a pre-dial guard at `if (this.userMode && this.currentBearer && bearerExpiryMs(this.currentBearer) <= Date.now())`, which stops the dead cache from being dialed, but it does not turn the refresh into a failed renewal and it does not stop the 5s loop.

The suggested shape is an unconditional refusal of a candidate whose expiry is already past, independent of the advancement test. The design is the author's to choose.

### The fixture grades non-advancement, and has nothing for this case

Both reviewers ran the baseline and the `!advanced` to `advanced` mutant eight times each. All 8 mutant legs reddened on the named cell, and all 8 baselines were green. The cell that dies is:

```
the refusal backs off past the 5s floor (no re-read in 12s)
```

That cell grades the non-advancement rule and it grades it well. There is no cell anywhere in the fixture for an advancing candidate that is already expired, which is why a green board does not speak to this finding.

### The prose still describes a superseded rule

The PR body states that `no identity comparison is involved` and that `there is no previous token to compare it against`, while the shipped code compares the candidate `exp` against `currentBearer`'s. The changeset describes a token inside its margin being treated as a failed renewal, whereas an advancing inside margin token is adopted. This part does not block by itself, but it should be brought in line with the code in the same push as the fix.

### Board

```
rows=36  distinct=36  total_count=36
success 33, skipped 3, failed 0, in_progress 0
```

The board is terminal and fully green. It does not bear on the finding above, because no cell exercises that path.

`attach-reconnect` reddened four timing cells on loopback ports 37769 and 35697, then passed on the rerun with different ports 38111 and 35485, ending `PASS - 72 passed, 0 failed` across 139 smokes and 10368 cells. That port sensitivity is tracked separately under #1583 and is not attributable to this PR.

davidfarah2003:
## Panel complete: three of three reviewers say REQUEST_CHANGES at `b4ca4332f2b9ef3312538f3d7d40c08636542236`

Relayed by the orchestrator. The measurements below belong to the reviewers. Items I confirmed independently at source are marked as such. This supersedes the relayed APPROVE further up the thread, which I posted before the panel had finished, and which was withdrawn in the comment above.

The third reviewer adds a second hole that runs in the opposite direction from the first. The predicate is both too narrow and too wide, and the two defects share one line.

### Too narrow: an advancing but unusable token is adopted

Covered in the previous comment. A candidate expiring at `now - 5s` beats a held one that expired at `now - 60s`, so `advanced` is true, the throw is bypassed, and dead material is cached. Measured by the reviewer as adopted and re-read 5 times in 30 seconds while `bindConnection` refuses to dial with it.

### Too wide: a live renewal is refused because `exp` has one second of resolution

Confirmed at source:

```
function bearerExpiryMs(bearer: string): number {
  ...
  if (typeof claims.exp !== "number") throw new Error("user-mode bearer is missing a numeric exp claim");
  return claims.exp * 1000;
}
```

`exp` is a whole number of seconds, so a healthy short TTL source read twice inside the same wall clock second returns the same `exp`. That makes `advanced` false. If the token is inside the 60s margin, and for a short TTL source it always is, the refusal fires and arms a 15s backoff against a token that dies in 4s. A token re-signed with an identical `exp`, which is what key rotation produces, is refused for the same reason.

So the same line that lets unusable material through also rejects a genuine renewal. Both reviewers reached that conclusion from different starting points.

### The 5s floor is unconditional

Confirmed at source:

```
private static readonly BEARER_REFRESH_MARGIN_MS = 60_000;
...
this.bearerTimer = setTimeout(() => void this.refreshBearer(), Math.max(5_000, delayMs));
```

A non-positive `nextRefreshMs` does not reach `BEARER_RETRY_MS`. It reaches the 5s floor, because adoption happened on the success path and nothing threw. That is the read loop.

### Suggested shape

The reviewer's proposal, and the design remains the author's to choose: refuse an already expired candidate outright, in parity with the credentials rule, and clamp a non-positive `nextRefreshMs` to `BEARER_RETRY_MS` instead of letting the 5s floor take it. The second half also removes the same-second false refusal, since a healthy source that genuinely re-issued will be read again on the retry cadence rather than punished for the granularity of its `exp` claim.

### The fixture is sound, and silent on both holes

Three independent runs of the same experiment, eight legs each, baseline and mutant in one invocation:

```
mutation:  !advanced  ->  advanced
cell:      the refusal backs off past the 5s floor (no re-read in 12s)
result:    mutant red 8/8, baseline green 8/8, no flake, tree restored clean
```

That cell grades non-advancement and grades it well. Nothing in the fixture walks an advancing but unusable source, and nothing walks two reads inside one second. A cell for each belongs with the fix.

### Prose

The PR body states that `no identity comparison is involved` and that `there is no previous token to compare it against`, while the shipped line reads `bearerExpiryMs(this.currentBearer)`. The body also quotes a mutation name, `the inside-the-margin token is adopted`, that no longer exists; CI prints `a non-advancing token is adopted instead of refused`. The changeset describes a token inside its refresh margin being treated as a failed renewal, which is not the shipped rule, and the changeset is the artefact that reaches the changelog.

### Board

```
rows=36  distinct=36  total_count=36
success 33, skipped 3, failed 0, in_progress 0
```

Terminal and fully green, and it does not speak to either hole, because no cell exercises those paths.

The `attach-reconnect` red on the first smoke shard 1/4 attempt is foreign and resolved: the rerun is green at `PASS - 72 passed, 0 failed`, on a different runner with disjoint bound ports, and that suite has no bearer source at all. The port sensitivity itself is tracked under #1583.

L4XB:
Both holes are real and both are fixed at `dd4cc761`. Thank you for running the panel to the end rather than letting the first APPROVE stand.

**Too narrow.** Confirmed exactly as described: a candidate dying at `now - 5s` advances a held one that died at `now - 60s`, so the advance test passed it. Expiry is now its own test, on expiry alone, with no advancement comparison, in parity with `presentableCreds`.

**Too wide.** Also confirmed, and the fix is not the one I would have guessed from the symptom. Comparing `exp` is the wrong question in both directions, because `exp` has one second of resolution: a key rotation re-signs the same claims under a new key and produces a token that is genuinely new and does not advance. The re-serve test now asks byte identity instead, which is the question the creds path asks with `credsFingerprint`: did the source re-issue anything at all. That adopts the rotation and the same-second re-read, and still refuses the stuck source in #1561.

Both refusals are now scoped to a token actually being held, which is the set `!initial` described but named after what the rules depend on.

**On the clamp.** I did not take it, and I owe you a measurement rather than an opinion. `implementations/auth/smoke/expiry-renewal.smoke.ts` mints 5-second bearers against the 60-second margin, so every one of them computes a non-positive `nextRefreshMs` and for that deployment the 5s floor *is* the renewal cadence. Clamping to `BEARER_RETRY_MS` leaves a 5s token dead for two thirds of every cycle. It is in the fixture as a mutation so the reasoning is executable rather than prose:

```
mutation: a non-positive delay is clamped to BEARER_RETRY_MS
red on:   a healthy short-TTL source still renews at the floor (>= 2 reads in 12s)
```

The loop the issue reports is a source returning nothing new, so it is closed at the source test, not at the timer. If you would still rather have the clamp, say so and I will take it with the short-TTL arm re-cadenced instead, but the two cannot both stand as they are.

**Fixture.** 4 cells to 12, including the two you asked for. Two of them count source reads in the same 12s window in opposite directions: a refused source reads none, a healthy short-lived source reads at least twice. That pair is what stops this refusal being widened into a blanket backoff later, which the single cell you graded would have accepted.

```
pnpm smoke:bearer-inside-margin      12/12
node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/bearer-inside-margin.json
  KILLED  an expired candidate is adopted when it advances the held one
  KILLED  the re-serve test goes back to an advancement comparison
  KILLED  a non-positive delay is clamped to BEARER_RETRY_MS
  KILLED  both refusals are applied with no token held
  All 4 mutation(s) killed. The suite discriminates.
```

One note from building that: the fourth mutation first came back WRONG-RED, because `refreshBearer` rethrows when `initial` and the process died before the cell could name itself. The cell now catches, so "does not throw" is the half of the claim it always was. That is `76891a37`.

**Prose.** All three corrections are in. The changeset no longer says a token inside its refresh margin is a failed renewal; the PR body no longer claims no identity comparison is involved, and no longer quotes a mutation name that does not exist.

Also ran `pnpm smoke:recoverable-error-channel` (5/5, it drives the same private refresh) and `pnpm smoke:gate-inventory` (green, the chain entry from the first round still holds).

Last thing, and it is adjacent rather than mine to fix here: `bindConnection`'s pre-dial guard reads `if (this.userMode && this.currentBearer && ...)`, so it says nothing when `currentBearer` is `undefined`, and the dial site presents `() => this.currentBearer!`. That is reachable today whenever a non-initial refresh fails, so it predates this PR. Happy to open it separately.

davidfarah2003:
Panel complete: three independent reviews of `cc5c1f0ef437f4f2a5c70b0a44426382f2a79043`.

Two reviews approve and one requests changes, so this is recorded as **not green**.
All three agree the first round's blocking issue is **superseded rather than closed**: the expression it named no longer exists after the rewrite.

---

**Blocking (1):**
- Add the true refresh-margin equality cell and <= to < mutation, or equivalent proof. The inline rewrite preserves the exact boundary the first review named. With a same-file positive control, that mutation SURVIVES the shipped 12-cell suite, while a deterministic Date.now-controlled smoke kills it, so this is a real and readily closable coverage hole.

**Body defects (2):**
- The PR body's local verification claims are still not anchored to this head by a full SHA, artifact hash, or durable run/job URL. This includes 12/12, all four mutation kills, companion smoke results, and typecheck.
- The adjacent-finding paragraph says the undefined-currentBearer dial path is reachable whenever a non-initial refresh fails. Actual state disagrees: non-initial refresh is armed only after currentBearer was assigned, failure preserves it, initial failure rethrows before dial, and no code clears it.

**Non-blocking (3):**
- The 12-second refused-source cell proves no 5-second reread and code inspection shows a 15-second timer is armed, but no cell positively observes a retry after 15 seconds. Removing retry scheduling entirely would pass that cell.
- Byte identity necessarily accepts byte-different reserializations of materially identical fixed-exp claims, causing floor-cadence reads until expiry. The body bounds and discloses this, and no reliable credential-generation signal is available in the bearer, so it is not a blocker.
- Several of the 12 assertions are correlated observations of the same arm. The suite discriminates the main branches, but the mutation result alone does not establish cell independence.

**Not verified (2):**
- changeset status could not compute divergence because this lane has no local main ref; it exited 1. The changeset was therefore checked manually against the isolated four-file author diff.
- CI check state was not used because the provider-cancelled jobs are not evidence about this head.

---

### Review 2: `APPROVE`

**Blocking (0):**
_none_

**Body defects (4):**
- This is v1 defect [4]/[5]/[8]/[12] (all three legs raised it) and it is NOT fixed at head. The body pastes `12/12` and four `KILLED` lines as bare transcript with no head sha, no artifact hash, and no durable run id. I reproduced both numbers myself so the claims are TRUE, but the body still gives a reader nothing that anchors them to cc5c1f0ef rather than to b4ca4332, which is exactly the failure mode that produced v1 defects [1] and [10] (a pasted transcript that had gone stale against its own fixture).. Where: PR body lines 34-42 (the Tests fence) and lines 25-28 (the clamp fence); PR body updatedAt 2026-09-15T12:22:20Z, read with `gh pr view 1572 --repo Cotal-AI/Cotal --json body` at 2026-09-15T13:06Z. Note: Reported, not edited. Lowest-cost fix is one line naming the sha the transcript came from.
- The body says `Rebased onto main at 08cff35`. The head is NOT a rebase. `git cat-file -p HEAD` shows two parents, dd4cc761e and 08cff3533, i.e. a MERGE commit. The distinction is load-bearing for a reviewer: it is why the raw b4ca4332..cc5c1f0ef diff is 137 files / +19739 and why endpoint.ts appears to gain 254 lines, and a reader told 'rebased' will not expect that and may grade the merge's creds-renewal work as this PR's.. Where: PR body line 63 (the trailing `---` note). Note: New defect, not raised by the v1 panel (it did not exist at b4ca4332). Factual, mechanically checkable, and it misdescribes the head's own shape.
- The body's stated bound on byte identity is optimistic in one direction. It says a source that re-mints new bytes on a fixed `exp` 'keeps reading at the floor until that exp passes. The expiry test then closes it.' I measured this (q2-probe, fixed exp 12s out, new `sig` every read): reads land at 1ms, 5005ms, 10007ms, 15012ms. The pre-expiry floor loop is real as described, but 'closes it' overstates what happens next. After the exp passes the expiry refusal fires and warns, and the endpoint keeps reading at the 15s BEARER_RETRY_MS cadence indefinitely; it does not stop. The body's own 'six reads, not an unbounded loop' is therefore true only of the 5s phase, and the reader is left thinking the source is closed out when it is merely slowed 3x.. Where: PR body line 17. Note: The BEHAVIOUR is correct and is the pre-existing backoff path; this is a prose-accuracy defect, not a code defect.
- The changeset says such a fetch is handled 'with ONE recoverable warning and the normal backoff'. There is no 'one': the refusal is re-entered on every subsequent refresh, so a source that stays stuck emits a warning per 15s retry, not a single one. My q2-probe shows the refusal path re-arming and re-firing on a cadence. 'One' is the count a reader will budget log volume against.. Where: .changeset/bearer-inside-refresh-margin.md, the 'Refused: the byte-identical token' paragraph. Note: The changeset is the artefact that reaches the changelog, which is the author's own stated reason (commit dd4cc761) for correcting it. Verbal, one word.

**Non-blocking (4):**
- The `<= 0` margin conjunct in the re-serve guard is not graded by the suite at all, in either direction.. Evidence: Two mutation runs, each carrying a positive control (a sibling endpoint.ts mutation that was KILLED in the SAME run, so the suite provably reaches the file at runtime): (a) the packet's boundary flip `<= 0` -> `< 0` grades SURVIVED with my control attached (the packet's own run of it graded only UNGRADABLE because it carried no control; adding one upgrades the verdict, which the tool states explicitly); (b) my stronger probe deleting the conjunct outright, `bearer === held && <margin expr> <= 0` -> `bearer === held`, ALSO grades SURVIVED. Suite stays 12/12 under both.. Reading: So it is not merely the equality boundary that is untested. The entire margin conjunct can be removed and no cell notices. This is wider than the first round's blocking finding, and I am recording it as the more accurate statement of the gap.. Why not blocking: (b) survives because removing the conjunct only makes the refusal STRICTER (byte-identical token refused regardless of margin), and the suite has no cell exercising a byte-identical re-serve on a token with plenty of life left, which is the only input that distinguishes them. That is a coverage gap, not a defect: the shipped behaviour is correct at every input the suite does exercise, and I verified the shipped conjunct is the conservative direction. (a) survives because the two branches differ only when `expiryMs - now - 60000` is EXACTLY 0; `exp` has one-second resolution (bearerExpiryMs returns `claims.exp * 1000`) and the suite injects no clock, so hitting it is a sub-millisecond race. I confirm the packet's reasoning that this is a respectable reason to omit that particular mutation.
- The 12s cadence pair pins a FLOOR on the backoff, not the 15s value.. Evidence: Probe with control: replacing `armBearerRefresh(BEARER_RETRY_MS)` with `armBearerRefresh(600_000)` (10 minutes) SURVIVES, suite 12/12, while the control mutation in the same file was KILLED in the same run.. Reading: The refused-source cell asserts `stuckReads - baseline === 0` over 12s, which any delay >= 12s satisfies. So a future change that silently lengthened the failure backoff to minutes would pass. Cheap and genuinely non-blocking, since longer is the safe direction for a stuck auth service.
- The first round's diff finding that the 12s cell 'does not prove that a 15s retry is actually armed' is WRONG at this head, and I want that on the record rather than carried forward.. Evidence: I mutated away the arm entirely (`armBearerRefresh(BEARER_RETRY_MS)` -> no-op) and the cell went RED by name: `x FAIL: the re-serve refusal backs off past the 5s floor (no re-read in 12s) { stuckReads: 4, stuckBaseline: 3 }`, suite 11/12, exit 1. Confirmed twice: once under mutation-proof (KILLED, named) and once by hand on a mutated tree that I restored and re-verified clean.. Reading: Removing the arm does not silence the source: the 5s-floor path from the OTHER endpoint keeps running and the stuck arm's own state changes, so the cell catches a removed retry. That concern was reasonable against the v1 fixture and does not hold against the v2 one.
- The fixture's `expectRed` strings are not quoted in the body, only the mutation NAMES are.. Evidence: All four mutation names in the fixture appear verbatim in the body (I diffed them mechanically); three of the four `expectRed` cell names do not.. Reading: Not a defect. Noting it because the v1 name-mismatch defect was found exactly this way, and the name side is now clean.

**Not verified (6):**
- CI check state on this PR. The brief states every job on the self-hosted label since 11:44Z was cancelled by the provider; I did not re-run any workflow and did not read check state as evidence either way. No claim in this report rests on CI.
- implementations/auth/smoke/expiry-renewal.smoke.ts still has NOT been observed passing at this head. I attempted it twice and it dies at import with ERR_MODULE_NOT_FOUND, EXIT 1 (see ENV2). nats-server IS present, so this is a module-resolution fault in the shared install, not a missing dependency. What Q4 actually leans on from that file is `const TTL_SEC = 5` at line 50, which I read directly from the source at my head, and the clamp argument is additionally anchored by a mutation I reproduced. So the Q4 conclusion does not rest on this suite running, but I have not observed it green and do not claim to have.
- pnpm smoke:gate-inventory (body line 48) was not run. It is a repo-wide gate census whose cost is unbounded relative to this review's budget, and nothing in my verdict depends on it.
- A REAL BROKER and a REAL auth callout were never exercised. This is the honest residual limit of the acceptance evidence above: my entry-point probes use the real public API and the real refresh timer, which is why I grade them representative, but they dial a dead port and the bearer is a JWT-shaped string rather than one minted by a live callout. The live suite that WOULD close this (expiry-renewal) is blocked by ENV2. So: production driver reaching production code, yes, proven; production broker round trip, no.
- The author's comment text is quoted only from the packet copy L4XB-comment-2026-09-15T11-54-27Z.md, fetched 2026-09-15T12:49Z per the packet header, createdAt 2026-09-15T11:54:27Z. It is mutable prose on a live PR and I did not re-fetch it. The PR BODY I did re-read live (updatedAt 2026-09-15T12:22:20Z, read 2026-09-15T13:06Z), and all body defects above are graded against that read.
- Whether the `currentBearer!` non-null assertion at endpoint.ts:1244 is reachable through some path outside endpoint.ts. I established it is unreachable via the two in-file routes (initial rethrow, non-initial preserve) and that currentBearer is never cleared in this file. I did not audit callers in other packages.

---

### Review 3: `APPROVE`

**Blocking (0):**
_none_

**Body defects (2):**
- The PR body in §'Tests' line 58 states that 'refreshBearer rethrows when initial and a refusal that reached there would take the process down instead of naming itself', but refreshBearer only rethrows when initial && an exception is thrown; when no token is held and initial=true, the guard if (held !== undefined) does not throw and adoption succeeds normally.
- The PR body verification block (pnpm smoke:bearer-inside-margin 12/12, mutation-proof 4/4 killed, tsc clean) lacks durable run IDs or commit hash references tying the transcript to commit cc5c1f0ef437f4f2a5c70b0a44426382f2a79043.

**Non-blocking (2):**
- In packages/core/src/endpoint.ts at line 1207, bindConnection's pre-dial check `if (this.userMode && this.currentBearer && bearerExpiryMs(this.currentBearer) <= Date.now())` does not cover the case where userMode is true but currentBearer is undefined. If the initial refresh failed or was omitted, dialer options at line 1244 evaluate `bearer: this.userMode ? () => this.currentBearer! : undefined`, which will return undefined when evaluated by the authenticator. As noted in PR discussion, this issue predates PR #1572 and warrants a dedicated follow-up PR.
- The exact boundary condition expiryMs - Date.now() - CotalEndpoint.BEARER_REFRESH_MARGIN_MS === 0 cannot be reliably tested in automated suites without clock injection due to the sub-millisecond precision window versus 1-second JWT exp granularity. The first round's blocking request for an equality mutant is superseded by inlining into the byte-identity check and is practically untestable without synthetic clock control.

**Not verified (2):**
- Full CI suite execution on GitHub Actions cannot be verified as Tenki runner jobs were cancelled by the provider across the project.
- Behavior with external OIDC providers that produce differing signatures/byte-representations for identical claims during rotation was evaluated by inspection and test mock rather than a live external auth provider.

L4XB:
Follow-up in #1689.

- **Blocker from review 1.** A new helper pins `Date.now` for one refresh, the way `secret-fs.smoke.ts` does, and records the delay that refresh arms. The `<= 0` to `< 0` mutant is now KILLED on `a re-served token with zero margin left is refused (the margin edge is inclusive)`. Against this PR's 12-cell suite it SURVIVED, while the other four `endpoint.ts` mutants were KILLED in the same run.
- **Review 2, plus one mutant of my own.** Three more cells, each named by a new mutant:
  - A cached-token cell kills the deleted margin conjunct.
  - A cell that reads the armed delay (15000 ms) kills `armBearerRefresh(600_000)`.
  - A 1ms cell kills a threshold moved above zero (`<= 1`). Review 2 did not ask for this one. It shows that the zero-margin refusal comes from the margin and not from the pinned clock.

  8/8 KILLED at `5681bbaff4c2b8e45eaf07823b439f0b36731283`.
- **Two corrections to what I wrote here.**
  - Review 1 is right: the dial path with an undefined `currentBearer` is not reachable when a non-initial refresh fails. The initial refresh rethrows before the dial, and the timer is only armed once a token is held. So there is nothing to open separately.
  - "The expiry test then closes it. For a 30s pin that is six reads, not an unbounded loop" was also wrong. Six reads is only the 5s phase before `exp`. After `exp`, the source is read every 15s with an expiry warning each time, and that does not stop. With `exp` pinned 12s out, I measured reads at 0, 5002 and 10003 ms, then at 15004, 30006 and 45007 ms, each with a warning, until the probe stopped at 47s.
- **Fixed in #1689.** The changeset's "one recoverable warning" had the same error. A mutation note that repeated the unreachable-dial claim is corrected too.

Issue #1578, opened 2026-09-13
0.49.0 upgrade on an existing deployment: do static credentials survive, and what is the upgrade order for a split broker/manager topology?


NomadaDigital01:
**Deployment:** currently on 0.48.2. Split topology — one container runs the broker (`cotal up --space main --host 0.0.0.0`: nats + JetStream + delivery), a second host runs `cotal supervise` with 30 managed PTY agents, and one external peer joins over the LAN with a static credential.

0.49.0 is marked **"Breaking pre-1.0 authority change"**: a static credential becomes an issuance, with three new per-space stores, two new one-shot profiles (`issuer`, `run-admitter`), and `COTAL_ACCEPTED_TOKEN` added to every connector's spawn environment.

We could not find guidance on what that means for a deployment that already exists. There is no `UPGRADING.md`, `MIGRATION.md`, `CHANGELOG.md` or `docs/upgrade.md` in the repository, and across the 35k-character release body there is no mention of existing credentials, re-issuance, backward compatibility, or migration. The only occurrence of "legacy" is `run-start` being refused on the legacy rail.

Two questions before we upgrade a running fleet.

### 1. Do existing static credentials survive, or must the fleet be re-issued?

The release says a connected client "reads its generation from an issuer-written accepted row". Credentials minted under 0.48.2 have no such row, and the accepted store is described as write-once per key.

If existing credentials keep working, we upgrade and move on. If every agent must be re-issued, that is 30 issuances plus one external peer, and it changes the shape of the operation — we would want to know that before starting rather than halfway through.

### 2. In a split broker/manager topology, which side upgrades first?

Broker and manager are separate containers, so there is necessarily a window where the two run different versions. With a breaking authority change, does one order fail closed and the other not? Is there a supported sequence, or should the mesh be fully down for the upgrade?

### Minor, but it affects our plan

We are currently running `cotal deliver` as a standalone process, by hand, because the daemon did not bind at boot (#1576). Does standalone `cotal deliver` remain supported in 0.49.0, or should the upgrade assume delivery is only ever a child of `cotal up`?

Happy to test and report back from a 30-agent deployment if that is useful — we have filed most of our findings as issues here already.

NomadaDigital01:
We went ahead and upgraded rather than wait, with LVM-thin snapshots of both containers as a rollback path. Posting what we measured, since these are answers to our own questions and may save the next person the same evening.

[...]

Issue #457, opened 2026-08-16
Manager's own credential is not renewed at the 24h boundary; the resulting frozen issuance gate makes restart impossible


NomadaDigital01:
## Summary

On an auth-mode mesh, the manager's own NATS connection is rejected with `Authorization Violation` **exactly 24h after it starts** — the same moment it successfully re-signs the managed credentials of every agent it owns. The manager process stays alive but becomes unreachable (`no manager reachable on the ep rails`), so the whole control plane goes dark while the agent processes keep running blind.

Worse, recovery is not possible by restarting: the manager persists its instance identity on disk and reuses it, so every restart tries to reclaim an issuance gate that is now frozen, and fails. In our case the only way out was wiping the broker's JetStream store.

Two independent machines hit this on the same day, each exactly 24h after their manager started.

## Environment

- `cotal-ai` 0.18.0
- Node v22.23.1
- Debian container (LXC), manager under systemd
- Auth-mode mesh (`--mode auth`), broker on a separate host running `cotal up`
- ~18 agents, spawned with `--detach`

## Timeline (single machine, times are UTC)

```
Aug 14 18:17:26  manager starts
                 console: http://<host>:45007/#t=<redacted>

Aug 15 18:17:25  ! manager goal-writer connection closed: connection refused
Aug 15 18:17:25  ! manager session-ledger connection closed: connection refused
Aug 15 18:17:25  managed cred renewal claude-<agent>: re-signed for the same identity
                 (exp +86400s); the agent endpoint's source re-read adopts it
                 ... (repeats successfully for all 18 agents) ...
Aug 15 18:17:26  ! manager service endpoint connection closed: Authorization Violation
```

Exactly 24h, to the second. From that point `cotal ps` returns:

```
✗ no manager reachable on the ep rails
  (deadline-exceeded: no describe reply from manager within 10000ms)
```

The manager process is still alive (`systemctl is-active` → `active`, uptime 1d+), and all agent processes are alive too — they simply have no routing.

## Why restarting doesn't recover it

`~/.cotal/auth/manager-instance.<space-hex>.json` holds a persisted `instanceId` and is **not** rewritten on restart (file mtime stayed at the original start time across several restarts). So each restart reclaims the same identity, and hits:

```
✗ re-registration could not revoke + verify-evict the superseded serve family;
  the gate is left frozen for reconciliation, no new spec published (SPEC 13.1):
  the delivery daemon is not reachable on the ctl.delivery-admin rail
[...]

PR #1588, opened 2026-09-13
docs: an upgrade contract for existing deployments

davidfarah2003:
An operator with an existing deployment had no document to read before upgrading. There is no `UPGRADING.md`, `MIGRATION.md` or upgrade section anywhere in the repository, so the answers to "do my credentials survive" and "which side of a split topology moves first" existed only in the code. 0.49.0 changed how a credential's authority is recorded, and a deployment running 30 managed agents hit exactly that gap.

## What this adds

`docs/UPGRADING.md`, with the pre-1.0 upgrade contract stated in operator terms and the first release section, 0.48.2 to 0.49.0.

**What keeps working with no action:** existing agent credentials keep authenticating, the channel registry survives, standalone `cotal deliver` remains supported, and `cotal join --lifecycle-uid` is not new in 0.49.0 (it has been required alongside `--creds` since well before this release, and the pairing rule did not change).

**The one thing that does not migrate is renewal.** A credential minted before 0.49.0 carries no issuance, so the manager refuses to renew it and names the agent in its log. Managed credentials live 24 hours and are re-signed once they pass 75 percent of their life, so a fleet comes up fine after the upgrade and then stops one agent at a time, at each credential's own expiry, within about a day. The page says to respawn the managed agents deliberately as the last step rather than meet the same work spread over the following day.

The section also covers what the boot log's gate-repair lines do and do not mean (they are a single frozen registration being repaired, not a fleet-wide credential migration, and reading them as the latter is the most likely way to conclude a fleet is fine when it is not), the order for a split broker and manager topology, the shape of the outage window, what to snapshot first, and the commands end to end.

**Where an answer was reasoned rather than measured, the page says so.** Broker-first is named as measured on a live deployment; manager-first is named as reasoned from where the new stores are provisioned, because it has not been run end to end on a production split topology.

## What is not in this PR

A check that refuses a range carrying a breaking change with no new section on this page is proposed separately in #1598, with a design note stating what it proves and what it cannot. This PR is the page alone; the rule that a credential-shape or renewal change is breaking whatever the commit marker says is stated on the page for people to apply in review.

## Verification

- `pnpm typecheck` green. `check:docsbundle`, `smoke:docs`, `smoke:gate-inventory`, `smoke:ci-suites`, `smoke:ci-declarations` and `smoke:ci-suites-freeze` (with base and head set) all green.
- The new page is picked up by the docs bundle, so `cotal_docs` serves it at the shipped version.

Refs #1578


davidfarah2003:
Replacement head `85e0e8e1af1508d3100880d7c7d3ff40ee0f1fd2`, rebased onto current `main`. It carries two fixes to the check itself and eight corrections to the page. Reviewers who read the previous head should re-read rather than diff, because the page's instructions changed, not only its wording.

**What the check now asks.** Coverage is the release sections a range **adds**, base compared against head. The first version asked only whether the range touched the page, and accepted on the section count at head, so a breaking change beside a one-line typo fix passed as "covered by 8 sections" written for earlier releases. That predicate accrues coverage it never earned and weakens every time someone does the right thing. A heading must also be a `##` and name its release: coverage is claimed by a heading, so a heading naming no release claims every release and distinguishes none.

**Misuse and refusal are now distinct exits.** A mistyped ref previously threw out of the top level and exited 1, the refusal code, so a job with a bad base ref reported a missing upgrade section for a verdict never computed. Misuse exits 2. Both exit codes are graded by cells that run the real CLI in a subprocess, in both directions, so nothing can satisfy them by making every exit 2.

**What this check does not prove, stated plainly because it matters here.** It proves a section for the release was written. It cannot prove the section is correct. Eight factual defects were found in this page by reading it against the source, and this check would have passed every one of them. That is not an argument against it, and making it verify prose would be worse. It is the reason the page now says the same thing where a reader will meet it, and the reason a human review of the words is still the real gate.

**The page was re-derived against v0.49.0 rather than edited.** The corrections that change what an operator does:

- Managed credentials are re-signed once past **75%** of life, not at half life. Half-TTL is named in the manager as the old schedule and the cause of #457. The page was telling an operator planning a 24-hour window to expect renewal at 12 hours when it happens at 18.
- `cotal backup` is not the backup command and would have failed on the first line of the procedure; it is `cotal backup create <dir>`, and it is offline-only, requiring a `cotal down --preserve-state` cut. Step 0 backed up while the stack was still running, which cannot work, so the recipe is re-sequenced rather than reworded.
- `cotal channels` is likewise a usage error; it is `cotal channels list`. Found by executing the page, not by reading it.
- Stopping a 0.48.2 manager **ends its agent processes in either order**. Sparing agents is a capability handshake a 0.48.2 manager never publishes, so the new CLI either refuses or falls to a legacy path and signals anyway. The page previously claimed the processes keep running. The respawn is therefore mandatory recovery for this upgrade, not an optional pass.
- There is no broker-only mode, so `cotal up --detach` also starts a manager on the broker host. The recipe now waits for `✓ manager up` before stopping it, since a manager that has not finished starting cannot be stopped.
- An external peer's `cotal mint` credential is issued with **no expiry** and no manager renews it. The page told operators to re-mint these, which is third-party coordination for a problem those peers do not have.
- `--lifecycle-uid` predates the release the page named, so no release is claimed for it.
- The `authentication error` text is the broker's, not a Cotal command's, and is described by shape rather than quoted as ours.

**How the commands were verified.** Every `cotal` invocation in the page was executed against the real 0.49.0 CLI in an isolated sandbox, never against a live mesh. Isolation was proven rather than assumed: an early attempt resolved to a real deployment through the root walk and was discarded, and the final runs are pinned so the CLI reports the sandbox path. The two usage errors above were found this way. Commands that could not complete without a live mesh are recorded as parse-verified, and the page marks anything not measured as reasoned, per its own rule.

Replayed over `v0.48.2..v0.49.0` the check refuses and names both breaking commits; `v0.48.1..v0.48.2` and `v0.47.0..v0.47.1` pass with no output, which is the leg that distinguishes detecting breaking changes from reding on everything. Eight mutants, all killed, each reddening a named cell. One of them survived its first run because the predicate was computed inline where no cell could call it, which is why it is exported and graded directly.

Refs #1578

davidfarah2003:
**Head pointer correction.**

The comment above names `85e0e8e1af1508d3100880d7c7d3ff40ee0f1fd2`. That sha was accurate when it was posted and was superseded 23 minutes later by an amended commit. It is **not an ancestor** of the current head: the two are siblings off the same parent.

```
current head : 0e1882ac7fdf254a70c07a253cca7bfe7a96748e
named above  : 85e0e8e1af1508d3100880d7c7d3ff40ee0f1fd2   superseded, do not verify against it
merge-base --is-ancestor 85e0e8e1 0e1882ac  ->  NO
```

This matters beyond bookkeeping. The superseded sha predates the empty-subsection fix, so its gate **accepts** a release section whose entire body is subsection headings, which was a blocking review finding. Verified by reading both trees with the same reader, calibrated so its zero is meaningful:

```
                                 heading-skip guard   sectionsOf (live control)
0e1882ac (head)                          1                     2
85e0e8e1 (superseded)                    0                     2
```

The control is nonzero at both shas, so the zero at the superseded sha is a reading rather than a dead matcher.

A reviewer who fetches the named sha to check that fix will correctly find it absent and file a blocker against work that already addresses it. Verify against `0e1882ac7fdf254a70c07a253cca7bfe7a96748e`.

davidfarah2003:
The technical content on this head looks good, and three reviewers have now checked the sixth head's substance. One item is still open and it is in the PR description rather than the diff.

**The description still describes the first head.** I checked each claim against `8fd7079c77adf99e9b9a6eee458d40b47b670250` rather than taking a reviewer's word for it:

| description says | measured at this head |
| --- | --- |
| "Self-test: 16 cells" (line 43) | 36 cells, confirmed by the self-test's own output and by three seats |
| "five mutants, all killed" (line 44) | `scripts/mutations/upgrade-section-gate.json` contains **9** mutants |
| "no matching section on the page" (line 17) | the gate now skips headings of any depth, which is what made the hollow-section case unrepresentable |

The mutant count I took from the fixture on this exact ref, not from a test log.

I am not editing the description myself, since it is yours and a patch would rewrite what you wrote. Could you refresh it when convenient?

**Why this is worth the round trip rather than a shrug.** This PR is a document that tells operators how to upgrade, and every defect found across six heads has been the same shape: text asserting something the system does not do. Eight commands in the guide were reasoned and dead on execution, and one rollback step cited a snapshot that step 0 never takes. A description that misdescribes its own diff is that same defect one layer out. Someone quoting "16 cells" from this page in six months would be reading a measurement of a diff that was replaced five times.

One claim on line 7 I checked and it is **fine**, so please leave it: `cotal join --lifecycle-uid` "has been required alongside `--creds` since 0.45.0". That is a statement about 0.45.0 being where the pairing rule started, not a claim that this PR targets 0.45.0. It reads oddly next to the 0.49.0 material but it is accurate.

Nothing else is blocking from my side. The board at this head is clean with jobs still running, and no reviewer has raised a technical objection to the sixth head.

davidfarah2003:
Following up on my earlier note about the description, because the split changed what it needs to say.

At `fb768ee8293d2716bcfa9239ba9a8141c24d1b59` this PR is four files:

```
added     .changeset/upgrading-page.md
modified  docs/README.md
added     docs/UPGRADING.md
modified  extensions/connector-core/src/docs-bundle.generated.ts
```

The gate, its self-test, its mutation fixture and its suite fragment are no longer here. **Splitting them out was the right call** and I want that on the record rather than buried: the page answers a live report from a broken deployment, and a CI gate enforcing that future upgrades carry a section is a separate change with a different risk profile. Bundling them is why this took eight heads, with the page and the gate taking turns blocking each other.

The description now describes machinery that is not in the diff:

| description says | at this head |
| --- | --- |
| the section gate | not in this PR |
| "Self-test: 16 cells" | there is no self-test in this PR |
| "five mutants, all killed" | there is no mutation fixture in this PR |

One claim I checked and it is **correct**, so please leave it: `cotal join --lifecycle-uid` "has been required alongside `--creds` since 0.45.0". That is a statement about where the pairing rule started, not a claim about this PR's target.

I am not editing the description myself, since it is yours.

**Why this matters more here than on a normal PR.** Every defect found across eight heads has had one shape: text asserting something the system does not do. Nine commands in the guide were reasoned and dead on execution, and the last one was created by the fix for the one before it. A description claiming a gate that was deliberately removed is that same shape at the outermost layer.

The step 3a wait now reads only past the byte offset captured before the start, and the comment above it explains why that is the only correct point to capture it. That is the strongest thing in the diff: the page now teaches the failure mode it suffered, so the next person writing a wait against an append-only log has the reasoning in front of them.

Issue #1586, opened 2026-09-13
entryArray identifier resolution has no covering cell, and deleting it flips zero of 427 verdicts


davidfarah2003:
`entryArray` in the coverage resolver reassigns `node` from an identifier lookup, and no cell covers that branch. The line is reachable in principle and decides nothing in practice, so the gap is the missing cell rather than the three words.

## Measured

`scripts/mutation-coverage.mjs:915`:

```js
const entryArray = (node) => {
  if (ts.isIdentifier(node)) node = variables.get(node.text);
  if (!ts.isArrayLiteralExpression(node)) return false;
```

Two experiments, each with the outcome named before the run:

```
predicted: deleting the line fails the selftest
observed : selftest 106 passed, 0 failed        -> NO CELL COVERS IT

predicted: deleting the line flips corpus verdicts
observed : baseline refusals 142, sabotaged refusals 142, verdicts flipped 0 of 427
           tree restored, git status clean
```

The argument for keeping it was that 57 corpus call sites pass argv as an identifier rather than an inline array (370 use an inline array). That argument does not survive the experiment: every identifier-argv suite is already witnessed another way. Checked on one of the 57, `mutation-command-safety.smoke.ts` passes an identifier argv and its config is witnessed by a direct source import instead.

## Direction

Refuse-only. If the branch were removed and some future config did depend on it, the resolver would refuse that config rather than accept it. The line cannot manufacture coverage, so this is dead weight rather than a false accept.

## Why it is worth a cell rather than a deletion

The file's own thesis is that unexercised code is not covered code. A branch inside it that no cell exercises is the same shape one level up. Deleting three words removes the symptom and leaves the resolver with no test for identifier-argv resolution at all, which is a real capability that a future config could need.

## Remedy

Preferred: add a fixture whose spawn passes argv as an identifier bound to an array literal, with a named cell asserting the config resolves as gradable through that path, plus a mutation with `expectRed` so removing the reassignment reds by name.

Acceptable: delete the line, and record in the selftest that identifier-argv resolution is unsupported so a future contributor does not assume it works.

## Suggested cells

- a config whose spawn argv is an identifier resolves as gradable and names the array elements
- accept control: the existing inline-array config still resolves, unchanged
- refuse control: removing the identifier reassignment reds the new cell by name

## Note

[...]

Issue #1233, opened 2026-09-03
connector-jcode: when soft interrupts time out, the automatic queue has no fallback drain and mimics a wedge on a healthy seat


davidfarah2003:
## Summary

When a Jcode seat's soft interrupts time out, the connector-owned automatic queue has no
fallback drain and grows without bound. The seat stays healthy and productive throughout, and
`cotal ps` renders the growing queue in a way that is indistinguishable from a wedge.

This is not #1080 (a restart stranding DMs on a dead lifecycle) and not the active-turn stall
that PR #1079 fixed. It is the case where **PR #1079's own remedy fails**: that fix routes
directed traffic through Jcode soft interrupts and commits at the clean containing-turn
boundary, so when the soft interrupt never gets a reply, there is no second path and the queue
is stuck permanently.

## Measured

A Jcode orchestrator seat, 17.9h uptime, on grok-4.6.

Roster, twenty minutes apart:

```
inbound: 27 automatic queued, oldest 49512s
inbound: 27 automatic queued, oldest 49696s
```

The count is static and the age grows, which is the signature of a queue that is neither
draining nor being added to.

Its connector log carries the reason, five times:

```
[cotal-jcode] soft interrupt failed: timeout: no reply to soft_interrupt within 30000ms
```

**The seat is fine.** Asked directly, it replied in under a minute:

> Alive. cotal_inbox drained (no pull-only). 27 connector-managed automatics remain queued
> (oldest ~13.8h, last non-empty drain 04:34Z); this jcode seat cannot ack those via inbox.

Its own logs agree: 8 provider requests in the preceding 30 minutes, with real API calls of
249 messages and 50 tools. It reads DMs, drains `cotal_inbox`, and works.

So `cotal_inbox` and the automatic queue are separate channels, and draining the first does
nothing for the second. The seat has no mechanism to acknowledge connector-managed automatics
at all, which means the depth can only fall if a soft interrupt succeeds.

## Why this is worth fixing rather than documenting

The roster signal is actively misleading, and it misleads toward the destructive action.

An operator reading that seat saw four of the four things in the standard Jcode wedge
signature at once: a large and ageing automatic queue, repeated `soft interrupt failed`, no
`status=2xx` in the sampled log window, and only `Ambient runner: not time to run` repeating.
Every one of those was either structural or a sampling artifact. The seat was healthy and was
about thirty seconds from being renamed and restarted.

[...]

PR #1600, opened 2026-09-13
test(coverage): pin identifier-argv resolution where it actually lives

L4XB:
Refs #1586 — and a negative result on the remedy it asks for, which is the part worth reading first.

## The preferred remedy is not achievable as written

The issue asks for "a fixture whose spawn passes argv as an identifier … plus a mutation with `expectRed` so removing the reassignment reds by name". I built the fixture. It does not red when `:915` is removed, and the reason is measurable:

**`launchedPaths` already resolves identifier argv, and does it better.** Its resolver walks outward from the use site to the nearest enclosing scope that declares the name, with a comment giving exactly the reason `entryArray`'s version is wrong: *"A flat name->value map is not a binding, it is a name collision waiting to happen."* Every route that reaches `entryArray:915` has gone through `launchedPaths` first.

So `:915` is not merely uncovered — it is unreachable-in-effect, which is a stronger statement than the issue makes and consistent with both corpus sweeps (0 of 427, then 0 of 430).

## What this PR adds instead

One cell, pinning the capability **where it lives**:

```
✓ a spawn whose argv is an IDENTIFIER bound to an array witnesses the script it names
```

Verified both directions, which is the whole point:

| mutant | cell |
| --- | --- |
| `launchedPaths`'s `arrays.get` returns `undefined` | **reds**, with the config refused by name |
| `spawnsEntrypoint.entryArray:915` deleted | stays green |

The second row is the issue's own experiment, reproduced, and the first is what makes the cell worth having: identifier argv now has a test, so the 57 corpus suites that depend on it are no longer relying on a capability nothing checks.

## What I did not do, and why

**I did not delete `:915`.** Deleting product code on a corpus sweep I could not reproduce here, with no cell of my own covering it, is not a call this PR should make. The evidence is in the issue twice over — the report and your triage, refusal sets diffed per config path — so the decision is a one-line follow-up whenever you want it.

**I dropped a cell I wrote and could not stand behind.** I suspected `entryArray`'s flat map would accept a same-named array from another scope — the collision `launchedPaths` warns about — and wrote a decoy fixture for it through the `executes` path. The tool refuses that decoy, correctly. But when I weakened the scope check to check the cell discriminated, it **stayed green**: the refusal comes from the package/build gate, not from argv resolution at all. A cell whose name promises a property it does not grade is the defect this file exists to prevent, so it is not in the diff. Reporting it because the negative result is useful: there is no false-accept hazard here to justify the deletion on safety grounds — only dead weight, exactly as the issue says.

`mutation-coverage.selftest.mjs`: 107 passed, 0 failed.


davidfarah2003:
Your two-row table reproduces exactly, first-hand, at `a4bd9f37b25190ed83c9f5d47147b45b502f12ae`. One invocation, three runs against one built tree, each mutation confirmed applied by reading the file back so no green row can be a mutation that silently missed:

| run | applied | tally | new cell |
| --- | --- | --- | --- |
| baseline at head | false | `MUTATION-COVERAGE SELF-TEST: 107 passed, 0 failed` | green |
| `launchedPaths` `arrays.get` -> `undefined` | true | (no tally printed) | **red, by name** |
| `entryArray:915` deleted | true | `MUTATION-COVERAGE SELF-TEST: 107 passed, 0 failed` | green |

The red names the cell: `✗ a spawn whose argv is an IDENTIFIER bound to an array witnesses the script it names`. Both rows hold, and the cell pins a real capability that nothing else checked. **The diff is good and I am not asking you to change it.**

This comment is about two sentences in the body, not about the work.

## `:915` is reachable, and it decides real verdicts

The body says it is "not merely uncovered, it is unreachable-in-effect", and later "there is no false-accept hazard here to justify the deletion on safety grounds, only dead weight". I ran 13 config shapes against three tools side by side: **A** = `origin/main` verbatim (`git show`, byte-identical, sha256 checked), **B** = A with `:915` deleted and nothing else (1338 lines vs 1337, the single missing line is exactly `:915`), **C** = a candidate fix. A and B disagree on four shapes:

| shape | what the program does | A (main) | B (minus `:915`) |
| --- | --- | --- | --- |
| launcher's argv is its own **parameter**, unrelated same-named `args` elsewhere holds the entry | never launches it | **GRADED** | REFUSED |
| entry sits in argv **after `-e`** | node runs the eval program, never opens the file | **GRADED** | REFUSED |
| callee **spelled** `spawnProc`, imported from a local helper | not a subprocess launch | **GRADED** | REFUSED |
| argv is a **conditional** bound to a name | does launch it | **REFUSED** | GRADED |

Read the direction: on three of those four, main is **wrong with the line and right without it**, and all three are false **accepts**. `:915` is not dead weight; it is a false-accept path. A coverage grader that names an uncovered mutation as covered is the failure this file exists to prevent.

The fourth is the other direction and is not a refusal at all. At the head, that config prints:

```
REFUSED s11-conditional.json: Cannot read properties of undefined (reading 'kind')
```

`variables.get` returns `undefined` for a name it cannot resolve, and the next line dereferences it. That is a TypeError wearing a refusal's clothes: it exits non-zero and reads like a considered verdict.

## Why your negative result and this are both true

`launchedPaths` runs first and admits the honest shapes, so **on every shape your fixture covers, deleting `:915` changes nothing**, which is exactly what you measured. `:915` only becomes decisive on shapes `launchedPaths` **refuses** and `entryArray` then accepts anyway. Your fixture is an accepting shape; the hazard lives in the refusing ones. That is also why no existing cell and no corpus config catches it: both sweeps looked for a shape the first resolver already handles.

The mechanism is in the file, at the head (`scripts/mutation-coverage.mjs` is unchanged by this PR, same blob as main):

```
:894   const variables = new Map();
:915     if (ts.isIdentifier(node)) node = variables.get(node.text);
:930     if (ts.isVariableDeclaration(node) && ts.isIdentifier(node.name) && node.initializer) {
:931       variables.set(node.name.text, ...)
:939     ts.forEachChild(node, visit);
```

One flat `Map` keyed on the bare name, and `visit` recurses over the whole file with no scope boundary, so any `const x = [...]` anywhere binds at `:915`. That is precisely the collision `launchedPaths`'s own comment warns about, and you quoted it in this PR body, living one function away. And one line below:

```
:936   if (["spawn", "spawnSync", "spawnProc", "pty.spawn"].includes(callee)
```

the callee is matched on its **spelling**, so a locally imported `spawnProc` that never touches `child_process` satisfies it.

## Your dropped cell was one move away

Your diagnosis was right and I want to be specific about it, because it is the most useful paragraph in the PR: your decoy was refused by the **package/build gate before argv resolution was ever consulted**, so weakening the scope check could not change the verdict. The decoy never reached the code it was meant to exercise. Declaring `executes` with the target package actually built by the command satisfies that gate and leaves argv as the only open question . That is the single move between your negative result and a cell that bites. You built the right experiment and correctly refused to ship a cell that did not grade what its name promised.

## What I did not measure

The 13 shapes are hand-built configs, not corpus fixtures; I did not find a shape in the existing 430 that exercises this, and my sweep agrees with both of yours that no current config does. I did not run the full corpus against a fix under CI. I have not measured whether any real suite in this repo binds a launcher's argv through a parameter, so this is a live false-accept path with **no known present victim**, which is an argument about risk, not about anything currently miscounted.

So: the diff stands on its own and the two sentences above overstate what the experiment shows. `:915` should not be deleted on a dead-code rationale, because it is not dead.

---

**Correction to one number above, and it is the kind a reproducer would trip on.** I wrote that A and B are "1338 lines vs 1337". The real counts are **1337 and 1336**. The `1338` was my own counter, not the file: I split on newlines, which yields a trailing empty element on a newline-terminated file, so every count derived that way was inflated by one. `wc -l` and `awk 'END{print NR}'` both give 1337 for the unmutated file.

This matters because someone reproducing by line count would read "1337" as the already-mutated file, anchor on the unmutated one, and conclude the deletion never applied.

Nothing else moves: the **delta is 1 under either counter**, the single differing line is still exactly `if (ts.isIdentifier(node)) node = variables.get(node.text);`, and the unmutated tool is byte-identical (sha256) to `git show origin/main:scripts/mutation-coverage.mjs`. "B is A minus that one line and nothing else" is the only property the four-shape table depends on, and it holds.

Caught by a reviewer who compared my stated line count against their own checkout and saw that both claims could not be true.

---

**Two further corrections, one of them to my own comment above, and one that affects whether this merges.**

**1. CI is red at this head, and the cause is this diff's own new cell.** At `a4bd9f37b25190ed83c9f5d47147b45b502f12ae` the checks are now fully terminal: 18 success, 3 skipped, 4 failure. At the base `fe813fe390d3ef47ac54325f48be816ebaa84148` those same four checks are green and nothing is pending, so this is not a pre-existing red. The failing step is `Re-prove fixtures for changed guarded sources`:

```
WRONG-RED   a bound argument list is not resolved, so a real launch is missed
  exited 1 but never printed the expected failure: "a launch whose argument
  list is bound first still witnesses the script" - 8 marks (baseline 107)
1 of 50 mutation(s) did not produce a clean, named red.
MUTATION REPROOF FAILED (1 fixture(s)): scripts/mutations/mutation-coverage.json
```

The mechanism, verified link by link against the head bytes. `check()` calls `process.exit(1)` on its first failure, so it is not an accumulator. The pinned mutant in `scripts/mutations/mutation-coverage.json` deletes the `launchedPaths` identifier resolver at `mutation-coverage.mjs:239`, and its recorded `expectRed` is the cell at selftest line 507. The new cell is at line 320, and its fixture binds argv to an identifier, so it depends on exactly the resolver that mutant deletes. Under that mutant the run now reds at 320 and exits before reaching 507. The harness sees exit 1 without its expected sentinel and calls it a dirty red.

Worth naming: that mutant did exit non-zero. It exited non-zero for the wrong reason. This is the precise hazard of anchoring on an exit code instead of a printed reason.

The simplest repair keeps this a one-file diff: move the added block to after the cell at line 507. I have built that candidate and verified it is a pure move, but I have not yet run it, so I am not claiming it works. A reviewer is separately measuring how many of the other 49 pinned mutants could steal the abort the same way, which is the number that decides whether moving the block is a fix or only a relocation.

The `smoke (shard 3/4)` failure is a different matter and I am **not** attributing it to you. It fails in `pnpm smoke:jcode-queue-fallback` on soft-interrupt and queued-batch rows citing #1233, a subsystem that does not touch this file. That same suite passed 21 of 21 on an unrelated head about an hour earlier, so it is not permanently broken either. I record it as unrelated.

**2. A correction to my own comment above, and it cuts against what I credited you with.** I accepted your in-code claim that the cell "does NOT red when `:915` goes, which is the measurement rather than a claim about it". The table is right and it reproduces. The inference does not hold, and the reason is stronger than my "accepting shape versus refusing shape" framing.

The cell never executes `:915` at all. From the head bytes: `spawnsEntrypoint` has exactly one call site, inside `executedWitness` at `:1113`; `executedWitness` has exactly one call site, at `:1147`. In that same loop, `:1135` returns when `launchedPaths` plus `sourceReaches` admits the config. This config has no `executes` key, and CI supplies the missing half of the argument, because deleting the `launchedPaths` resolver is exactly what reds this cell, which proves the cell is admitted at `:1135` and returns there. So `:1147` is never evaluated for it and `:915` is unreachable.

A reviewer confirmed this directly by instrumenting both functions: zero trace lines for `identifier-argv`, with the same instrumentation printing two lines for the pre-existing `args-variable` cell as a known-positive control. In the whole selftest, `:915` fires exactly once, and not for the new cell.

So the green under deletion is not weak evidence about `:915`. It is no evidence. `:915` could be replaced with a `throw` and this cell would stay green. The comment claims a discrimination the cell cannot perform, and since the comment will be read as documentation by whoever touches this next, it is the part I would most want changed.

None of this touches the value of the diff. The capability is real, nothing pinned it, and pinning it is worth doing. The cell is a good `launchedPaths` witness. It is just not a `:915` witness, and I credited it as one.

**A correction to the correction immediately above, on the evidence rather than the conclusion.**

I wrote that CI supplies the missing half, because deleting the `launchedPaths` resolver is what reds this cell, and that this proves the config is admitted at `:1135`. That inference does not carry, and I would rather retract it than leave a chain that looks stronger than it is.

`launchedPaths` has two call sites, not one:

```
1121:  return launchedPaths(suite, source).some((entry) => entryPackages(entry).has(name));
1135:    if (launchedPaths(suite, source).some((entry) => sourceReaches(entry, mutation.file))) return;
```

and `:1121` sits inside `executedWitness` itself. So deleting the resolver proves only that the cell depends on it somewhere. It does not, on its own, say which call site admitted the config.

What actually pins it is a direct instrument rather than an inference: tracing which accept line admits each config yields `ACCEPT-AT-1135` for `identifier-argv` and `ACCEPT-AT-1147` for the pre-existing `args-variable`, identical with `:915` present and with `:915` deleted, alongside zero `spawnsEntrypoint` entry traces for `identifier-argv` against a known-positive of two for `args-variable`. That is measurement, not deduction.

The conclusion is unchanged and I have verified the static half myself: `spawnsEntrypoint` has exactly one call site, inside `executedWitness`, which itself has exactly one call site at `:1147`, below the `:1135` return. `:915` is unreachable for this cell. Only my reasoning for it was loose, and the distinction matters to anyone who later wants to attack the chain.

Caught by a reviewer reading my own argument more carefully than I did.

L4XB:
Closing this: the cell duplicates one that already exists, and it breaks the reproof by firing before it.

**It is a duplicate.** `argv-conditional` at `mutation-coverage.selftest.mjs:505` already pins identifier-bound argv resolution, under the name `a launch whose argument list is bound first still witnesses the script`, and `scripts/mutations/mutation-coverage.json` already backs it with the mutation `a bound argument list is not resolved, so a real launch is missed`. Both exercise the same line, `mutation-coverage.mjs:239`.

**It hides the canonical red.** Measured locally by deleting that line by hand and running the self-test both ways:

```
── on main ──
✗ a launch whose argument list is bound first still witnesses the script
     REFUSED argv-conditional.json: ...

── on this branch ──
✗ a spawn whose argv is an IDENTIFIER bound to an array witnesses the script it names
     REFUSED identifier-argv.json: ...
```

My cell runs earlier, so the self-test exits there and the expected cell never runs. That is exactly what CI reported: `WRONG-RED  a bound argument list is not resolved, so a real launch is missed — exited 1 but never printed the expected failure`, and then `1 of 50 mutation(s) did not produce a clean, named red`. Reproduced locally with `node scripts/mutation-proof.mjs --config scripts/mutations/mutation-coverage.json`, same verdict.

**#1586 stays open, and I now think it is right.** The line it names is a different one — `entryArray` at `:915`, `if (ts.isIdentifier(node)) node = variables.get(node.text)` — not the launcher-side `arrays.get` at `:239` that `argv-conditional` covers. My fixture was the plain identifier-bound-array shape that `:915` exists for, and it came back `graded=0 refused-with-reason=1`: the branch did not make it gradable. That is one more observation on the side of the issue's own measurement (deleting `:915` flips 0 of 427 verdicts), and against my attempt to write a cell for it.

Sorry for the red shard. Nothing from this branch is needed.

Issue #1585, opened 2026-09-13
generated .d.mts widens census state to string, so the drift cell cannot see a literal change


davidfarah2003:
The generated declaration widens the census `state` to `string`, so the five-member contract exists only in the implementation and the drift cell cannot see a literal change.

## Measured

`scripts/preflight-npm-publish.d.mts`, same generator, three call sites:

```ts
:24  export function classifyDirectPublishPermission(body: any, identity?: {...}):
       "refused:malformed-trust" | "refused:no-trusted-publisher" | "refused:no-github-publisher" | "createPackage" | "stage-only";

:40      state: string;      // preflightNpmPublish
:54      state: string;      // preflightFromRepository
```

Line 24 preserves the five-member literal union. Lines 40 and 54 emit bare `string` for the census result. The difference is not the emitter: `tsc` preserves a literal union as a function return type and widens an object-literal property. Line 24 is the accept control, since it comes from the same generator in the same file.

## Consequence

The drift cell at `bin/smoke/npm-publish-preflight.smoke.ts:627` compares the declaration against the contract. With `state: string` in the declaration, a change to any state literal leaves the declaration byte-identical, so that cell cannot red on it. The contract is enforced only where the implementation happens to be read.

## Fix

Narrow `state` to its literal union in the declaration, so the contract lives in the published type:

```ts
state: "all-present" | "nothing-to-publish" | "inconclusive" | "mixed" | "incomplete";
```

The change must regenerate the `.d.mts` via `pnpm gen:npm-publish-preflight-dts` in the same commit, or the existing skew cell reds instead.

## Suggested cells

- a renamed state literal reds the drift cell at `:627` on the declaration alone
- accept control: `classifyDirectPublishPermission` at `:24` keeps its union and its cell stays green
- the declaration and the generator agree after regeneration, so the skew cell stays green

Refs #1550

davidfarah2003:
## Disposition: `confirmed`

Measured at `fe813fe390d3ef47ac54325f48be816ebaa84148` (`origin/main` at the time of this pass), in a dedicated clone with its own `node_modules` (every `node_modules/@cotal-ai/*` symlink verified to resolve inside that clone before anything was measured).

### Mechanism, reproduced

`scripts/preflight-npm-publish.d.mts:40` and `scripts/preflight-npm-publish.d.mts:54` both emit `state: string` for the value returned by `preflightNpmPublish` and `preflightFromRepository`.

[...]

PR #1601, opened 2026-09-13
Pin the npm preflight census verdict in its published type

davidfarah2003:
The release preflight's published type said less than the preflight itself, so the verdict it returns was not pinned by anything a reader or a consumer could see.

## What was wrong

`scripts/preflight-npm-publish.d.mts` is generated from the module. At `:40` and `:54` the census result's `state` was emitted as bare `string`, while `:24` kept the literal union for `classifyDirectPublishPermission`. The union therefore existed only inside the implementation.

The cause is in the source, not the emitter. `tsc` preserves a literal union on a declared return type and widens a property assigned from an un-annotated object literal. Measured here on `origin/main` before any change, with the same generator on the same file:

- renaming either census verdict left the regenerated declaration **byte-identical**
- renaming a literal inside `classifyDirectPublishPermission`, which has a declared return type, **did** move it (`24c24`)
- a deliberately non-matching edit was graded void rather than green, so an unapplied edit could not be mistaken for the finding

Because the declaration and the module still agreed with each other byte for byte, the existing skew check stayed green on both renames. The contract the declaration is supposed to pin was not pinned.

## What now happens

The verdict is named on the preflight's own result shape, which is what carries it through the generator, and the declaration is regenerated with the repository's own generator command rather than edited. Both entry points now publish `state: NpmPublishPreflightState`, and the declaration exports `"nothing-to-publish" | "ready"`.

The release smoke now also grades the declaration directly and refuses one where either entry point's `state` is bare `string`, is missing a verdict, or admits a verdict the census can never return. A consumer that branches on a mistyped verdict is now a type error instead of silently compiling.

**The union is two members, not five.** The issue body proposed `"all-present" | "nothing-to-publish" | "inconclusive" | "mixed" | "incomplete"`. Measured in the module, `state` is only ever `"nothing-to-publish"` or `"ready"`; the other names are `registryVerdict` values, and `inconclusive`, `mixed` and `incomplete` all throw rather than return. Publishing the five-member union would have put three verdicts into the type that no call site can produce, so the two real ones are encoded and a mutation guards against widening the published union to admit `inconclusive`.

## Proof

Every proof here ran a pre-fix control, because a probe that only exercises the fixed shape would have passed before the change.

- **pre-fix control:** the new check carried back onto the pre-fix declaration reds, `SUITE COMPLETE: 65 passed, 2 failed`, exit 1.
- **accept control:** the same check against the regenerated declaration, `SUITE COMPLETE: 67 passed, 0 failed`, exit 0.
- **refusing cases:** the declaration grader has **7 accepting branches and 7 refusing cases**, one per branch rather than one per bug, plus an explicit accepting case. Each refusing fixture is derived from the real declaration by a single substitution, so it differs from an accepting input only in the one context under test, and each asserts the refusal names the entry point it came from, so a fault on one entry point cannot be reported as the other.
- **mutation fixtures:** three added, including one that widens `state` back to `string`. `node scripts/mutation-proof.mjs --config bin/smoke/mutations/npm-publish-preflight.json` reports **all 23 mutations killed**.
- **an honest negative result, kept:** a first attempt at the second entry point's mutation deleted its annotation instead of widening it, and graded `WRONG-RED`. That was correct: `preflightFromRepository` returns `preflightNpmPublish(...)` directly, so it inherits the union by inference and deleting the annotation widens nothing. The fixture now widens the declared return, which does reach the check. The annotation is kept on both entry points regardless, because that inference only holds while the body stays a direct return.
- `pnpm typecheck` exit 0.

`.d.mts` was never hand-edited; it is the generator's output, committed.

CI order is not a concern for this change: the generator has no CI step at all. The smoke regenerates the declaration in memory and compares, so the check cannot run before a generation it does not depend on.

Empty changeset: this is a repository script and test surface and ships in no package, matching the existing `npm-all-present-noop` changeset for the same area.

Refs #1585


davidfarah2003:
Two reviewers found the same class of defect in the checker this PR adds, and both were right. The check was green on a declaration it should have refused, so this round fixes the checker and adds a control for each way it was blind. The original fix to the declaration itself is unchanged.

**The checker read the first of several candidates in three places.** Each is a way the published type could be widened back to bare `string` with this suite fully green.

- The property regex took the **first** `state:` anywhere in the resolved block. Adding `census: { state: NpmPublishPreflightState }` above a top-level `state: string;` produced **zero refusals**. The read is now depth-aware and top-level only, and more than one top-level `state` is refused as ambiguous rather than resolved to whichever comes first.
- The result-block search took the **first** declaration of an entry point. A second `export function preflightFromRepository(...)` appended after the real one was never read, although it is the one a consumer resolves. Multiplicity is now refused by name, and the search is bounded to the entry point's own declaration, so a function whose result is not an object literal cannot be graded against the next function's result.
- `followAlias` treated `string` as an alias to look up, found no `export type string`, and refused at the undefined-alias branch. The two "widened back to string" fixtures therefore passed **for a reason unrelated to the widening they were named for**, which left `state is <t>, not a union of the census literals` reached by no fixture at all. That is the one branch whose message names the defect this check exists to catch. Type keywords now denote themselves, and the pre-fix control confirms the real defect lands there: reverting the module to a bare-string emit now reports `state is string, not a union of the census literals` on both entry points.

**A correction to this PR's own body.** It claimed 7 accepting branches with one refusing case each. Read out of the source, the grader has **9** refusal branches and the table had **13** fixtures covering **9** of 9. The earlier claim was not a count of anything; it was a description that stopped matching the code. Branch coverage is now asserted rather than described, and the enumerated branch set is itself checked against the grader's source, because a hand-maintained list of branches is the same kind of second copy that rots. Adding a tenth branch without extending the list reds the count cell; rewording a branch message reds the fixtures that name it.

**Controls.** Each new reader was reverted on its own, and reds exactly the fixtures that name it and nothing else: the depth-aware property read (2 cells), the multiplicity guard (1 cell), the type-keyword resolution (3 cells), the anti-rot count (1 cell). Restored, the suite is **78 passed, 0 failed**. The pre-fix control on the original defect still reds, regeneration diff is **0 bytes**, `pnpm smoke:mutation-fixtures` passes with **0** anchors spanning a comment, and `node scripts/mutation-proof.mjs --config bin/smoke/mutations/npm-publish-preflight.json` reports **all 23 killed**, each red for its named assertion.

The suite previously reported a passing contract check on a declaration whose published type was `string`. A green that a decoy walks through is worth less than a red, so it is the checker that changed, not the fixtures around it.

PR #1631, opened 2026-09-15
fix(scripts): run main() when a script is reached through a symlink

### LINKED ISSUE #1622: Operator-literal and attribution gates exit 0 without running when invoked through a symlink
Two CI gates exit 0 and print nothing when their script is reached through a symlink, because their entrypoint guard compares `import.meta.url` with the literal `process.argv[1]`. Through a symlink the two strings differ, `main()` never runs, and the process ends with exit code 0 having scanned nothing. A gate that reports clean without running is the dangerous direction for both of these.

Verified by execution on main `794e3d2ec6456838809f5096d8c7388fd1bf6c4a`:

| script | invoked directly | invoked through a symlink |
|---|---|---|
| `scripts/check-operator-literals.mjs --root <fixture repo>` | exit 2, 133 lines (refuses to grade without a range or event) | exit 0, 0 lines |
| `scripts/check-attribution.mjs --help` | exit 2, usage line | exit 0, 0 lines |

The guard in both files reads:

```js
if (import.meta.url === `file://${process.argv[1]}`) {
  process.exitCode = main(process.argv);
}
```

(`scripts/check-operator-literals.mjs:1650`; the same string in `scripts/check-attribution.mjs`.) The comparison also fails for any path that needs URL escaping and for a relative `argv[1]`, since `import.meta.url` is always absolute and percent-encoded.

Reviewers of #1614 flagged this as pre-existing and untouched by that PR; it is not a regression from #1613.

Fix direction:

- Compare resolved paths, not strings: `realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)`, or drop the guard in favour of a separate importable module and a thin executable entry that always calls `main`.
- Add a self-test cell that invokes the script through a symlink in a temp dir and requires the same exit code and output as the direct invocation, with the current guard kept as a registered mutant so the cell is proven live.
- Audit the other scripts under `scripts/` for the same guard string before closing; only the two above were executed here.

L4XB:
Closes #1622

Eight scripts under `scripts/` decided "was I run, or imported?" by comparing `process.argv[1]` against `import.meta.url` as strings. Node resolves the main module through its symlinks before it becomes `import.meta.url` and leaves `argv[1]` as the caller typed it, so through a link the two differ, `main()` never runs, and the process exits 0 having done nothing.

## It does not need a hand-made symlink

The issue reproduces it with `ln -s`. It also reproduces on an ordinary macOS checkout, because `/tmp` is a symlink to `private/tmp`, and that is the shape a reviewer will hit first. Measured on `08cff353`, the same checkout by two paths:

```
$ node /private/tmp/<wt>/scripts/check-attribution.mjs --selftest | tail -1
selftest: 14/14 fixtures graded as expected        $? = 2   (2 is --selftest's own pass code)

$ node /tmp/<wt>/scripts/check-attribution.mjs --selftest
                                                   $? = 0   no output at all
```

## Every guard under scripts/, measured

The new self-test run at the merge base `08cff353` is the audit the issue asked for. Direct invocation on the left, the same file through a symlink on the right:

| script | guard on main | direct | through a symlink |
| --- | --- | --- | --- |
| `check-attribution.mjs` | `` `file://${argv[1]}` `` | exit 2, usage line | **exit 0, nothing** |
| `check-operator-literals.mjs` | `` `file://${argv[1]}` `` | exit 2, `unknown argument` | **exit 0, nothing** |
| `live-job-conclusion.mjs` | `new URL(\`file://${argv[1]}\`)` | exit 1, `--result is required` | **exit 0, nothing** |
| `doc-binding.mjs` | `pathToFileURL(argv[1])` | exit 2, usage line | **exit 0, nothing** |
| `pr-head-gate.mjs` | `resolve(argv[1])` | exit 2, usage line | **exit 0, nothing** |
| `verify-publish-closure.mjs` | `resolve(argv[1])` | exit 2, usage line | **exit 0, nothing** |
| `preflight-npm-publish.mjs` | `pathToFileURL(argv[1])` | not spawned, see below | not spawned |
| `post-publish-install-probe.mjs` | `resolve(argv[1])` | not spawned, see below | not spawned |

`upgrade-section-gate.mjs` is the ninth and it is **already correct**: it resolves both sides inline and has a cell and a registered mutant of its own. It is exempted from the census by name rather than converted, because re-pointing another fixture's proof buys nothing here.

Three spellings, three ways to be wrong. The template literal also drops the percent-encoding `import.meta.url` always carries, so a path with a space or a `#` never matches; `new URL("file://" + p)` reads everything after a `#` as a fragment; `resolve()` and `pathToFileURL()` encode correctly and still compare an alias against a canonical path. **One claim in the issue did not reproduce:** a relative `argv[1]` is fine on Node 22+, because Node absolutizes `argv[1]` before the script sees it (checked on Node 26.8.2: `node ./probe.mjs` gives an absolute `argv[1]` and the string compare succeeds).

## The fix

`scripts/main-entry.mjs` asks the question once, comparing two real paths, with the `entry === self` fast path first so the common invocation costs no `stat`. `realpathSync` on **both** sides, not on the entry alone: under `--preserve-symlinks-main` it is `import.meta.url` that keeps the link.

## Cells

`scripts/entrypoint-guard.selftest.mjs`, registered as `smoke:entrypoint-guard` in the CI chain:

- **Parity, six cells.** Spawn the script by its real path and through a real symlink in a temp dir; require the same exit code, stdout and stderr. Each probe argv was measured by hand first so the suite reaches no network and writes nothing. The link is asserted to be a link (`lstat` plus `realpath`) before its result is read, so a cell that quietly stopped testing a symlink fails rather than passing forever.
- **Census, two cells.** `preflight-npm-publish.mjs` and `post-publish-install-probe.mjs` start their work at entry with no argv that refuses first, so no probe for them exists that does not pack a tarball or read the registry. They are graded on the source instead: a hand-rolled guard anywhere under `scripts/` is refused. A census reports "clean" on a healthy tree, which is also what a census that has stopped looking reports, so a planted control in **each** direction proves the reader can still say no.

At the merge base the suite is **1 passed, 7 failed**; on this branch it is **8 passed, 0 failed**.

## Mutants

`scripts/mutations/entrypoint-guard.json`, `node scripts/mutation-proof.mjs --config scripts/mutations/entrypoint-guard.json`, all five killed and each named its own cell:

| mutant | cell that reddened |
| --- | --- |
| `check-attribution.mjs` back to the concatenated URL string | `check-attribution.mjs answers the same through a symlink` |
| `check-operator-literals.mjs` back to the concatenated URL string | `check-operator-literals.mjs answers the same through a symlink` |
| `live-job-conclusion.mjs` back to `new URL()` | `live-job-conclusion.mjs answers the same through a symlink` |
| `isMainEntry` keeps the fast path and stops resolving | `verify-publish-closure.mjs answers the same through a symlink` |
| the census reader is blinded | `the census control refuses a planted hand-rolled guard...` |

The fourth is the one worth having. It leaves `entry === self` intact, so every direct invocation still answers correctly and only the symlink half reddens: that is what a plausible simplification of this helper looks like, and it is invisible to any cell that never uses a link.

## One coupling this created, and why it is one line

`check-operator-literals.celltamper.selftest.mjs` copies the scanner into a temp dir and spawns it 66 times. A copy needs the shared helper next to it, so the suite copies `main-entry.mjs` once alongside. Measured before the fix: **49/175 checks passed, 126 failed**, every copy dying on module resolution. After: **175/175**. Its docblock described the old guard and the two host conditions that used to disable a copy, so that paragraph is updated: `guardSafe` is kept, and the comment now says what it still buys, which is that a regression of the guard becomes one named refusal instead of 66 phantom reds.

## Everything run

| command | before | after |
| --- | --- | --- |
| `node scripts/entrypoint-guard.selftest.mjs` | 1 passed, 7 failed | **8 passed, 0 failed** |
| `node scripts/check-operator-literals.celltamper.selftest.mjs` | 175/175 | **175/175** |
| `node scripts/check-attribution.mjs --selftest` | 14/14 | **14/14** |
| `node scripts/check-operator-literals.mjs --selftest` | cells=65/65 PASS | **cells=65/65 PASS** |
| `pnpm smoke:gate-inventory` | OK | **OK** |
| `pnpm smoke:mutation-coverage` | 150 passed | **150 passed, 0 failed** |
| `pnpm smoke:mutation-fixtures` | OK | **OK** |
| `pnpm smoke:doc-binding` | 7 passed | **7 passed, 0 failed** |
| `pnpm smoke:live-job-conclusion` | 14 passed | **14 passed, 0 failed** |
| `pnpm smoke:verify-publish-closure` / `:npm-publish-preflight` / `:publish-closure-gate` / `:install-probe-ci` | pass | **pass** (npm-publish-preflight 80/80) |
| `pnpm changeset status` | ok | **ok** |
| `gen:publish-closure-dts`, `gen:npm-publish-preflight-dts`, `gen:ci-suites-dts`, `gen:reaper-dts` | - | **no diff** |

**Two suites are red here and are red at the merge base too, identically, and neither is touched by this change.** `smoke:pr-head-gate` fails one cell on both (`the shipped pr-head-gate command refuses a YAML merge key in on...`, 45 passed / 1 failed at `08cff353` and at this head). `smoke:post-publish-install-probe` fails on both with `ERR_MODULE_NOT_FOUND` for `@cotal-ai/workspace/dist`, which is an unbuilt workspace on this box rather than a code failure.

Node 26.8.2, macOS. A changeset with empty frontmatter is included: no published package changes.


davidfarah2003:
Cold read complete: one independent review of `c78990af6e74fb51db62ffb78a9a54d20463ada0`.

Relayed by the orchestrator. The measurements below belong to the reviewer.

Verdict `APPROVE`, no blocking issues. The reviewer reproduced the underlying defect on Linux with an explicit symlink against the merge-base source and confirmed the fix, and drove the guard census with positive controls. Four body defects and five non-blocking findings are recorded, the first non-blocking one significant: the parity cells grade agreement between the two invocation paths, not that `main()` ran, so a guard that never fires on either path still passes them. This is a single cold read, not the three-family gate.

---

**Blocking (0):**
none

**Body defects (4):**
- PR body, 'It does not need a hand-made symlink' section: 'node /private/tmp/<wt>/scripts/check-attribution.mjs --selftest | tail -1 -> selftest: 14/14 fixtures graded as expected $? = 2 (2 is --selftest's own pass code)'. --selftest's pass code is 0, not 2. scripts/check-attribution.mjs:291 reads `if (args.includes('--selftest')) return selftest() ? 0 : 2;` so 0 is the pass and 2 is the FAILURE code. Measured unpiped at this head: `node scripts/check-attribution.mjs --selftest >/dev/null 2>&1` exits 0. Measured at the merge base on the same file: also exits 0. The parenthetical teaches a future reader that 2 means pass for this flag, when 2 is what that flag returns when a fixture FAILS. A [...]
- PR body, 'Cells' and 'Mutants' sections, and scripts/mutations/entrypoint-guard.json 'guard': the suite grades that a script 'runs its main() and answers exactly as it does by its real path, rather than exiting 0 having done nothing'. The parity cells assert equality of (exit code, stdout, stderr) between the two invocations. They never assert that either invocation ran main(). I made isMainEntry always return false, so every gate exits 0 having scanned nothing by BOTH paths: parity holds and the suite reported 8 passed, 0 failed. See finding NB1 for the run and its control. The fixture's stated guard is stronger than what the cells assert. The 'rather than exiting 0 having done nothing' [...]
- PR body, 'Everything run' table: '`pnpm changeset status` | ok | **ok**'. Not reproducible in this lane: `pnpm changeset status` exits 1 with 'Failed to find where HEAD diverged from "main"'. `git rev-parse --verify main` fails here: the lane has no local `main` ref. Recorded as UNVERIFIED rather than contradicted. This is a lane artifact and I have no evidence against the authors' result. Listed so the figure is not carried as confirmed.
- PR body: 'Eight scripts under `scripts/` decided "was I run, or imported?" by comparing `process.argv[1]` against `import.meta.url` as strings.'. Accurate for scripts/, and the body is careful to scope itself to that directory. But the same defective guard is still live OUTSIDE scripts/ at this head: packages/seat/src/custodian.ts:481 carries the exact template-literal form, and examples/05-scale-showcase/src/{swarm.ts:129,observe.ts:216} carry it too. The census in the new suite reads only scripts/*.mjs, so nothing in this PR can ever see them. Not a false statement, but the body's framing ('Every guard under scripts/, measured') invites a reader to conclude the class of bug is now closed [...]

**Non-blocking (5):**
- The parity cell grades AGREEMENT between two paths, not that main() ran, so the exact defect of #1622 survives it. method: Reviewer-authored mutation fixture run through the repo's own scripts/mutation-proof.mjs, with a KILLED CONTROL in the same file and the same run, as required. control: CONTROL: isMainEntry stops resolving (the shipped fixture's mutation 4, byte-identical) => KILLED, red and named on 'verify-publish-closure.mjs answers the same through a symlink'. mutation-proof itself printed: 'positive control: another mutation in scripts/main-entry.mjs was KILLED in this run, so the suite provably reaches this file at runtime and this survivor is a real coverage gap'. probe: PROBE: [...]
- The census is line-local and literal, so a hand-rolled guard written over two lines evades it. method: Drove the shipped predicate's own logic, with GUARD_SHAPE parsed out of scripts/entrypoint-guard.selftest.mjs at runtime rather than retyped, against planted sources. Positive control included. control: 'if (import.meta.url === `file://${process.argv[1]}`) { main(); }' => CAUGHT. (This is the shape the shipped in-file control plants, and it works.) notHypothetical: scripts/upgrade-section-gate.mjs:892-894 is itself written in the two-line shape the census cannot see. It is exempted by name anyway, so nothing is wrong today, but it demonstrates that contributors in this repo already write [...]
- The identical defective guard is still live in shipped package code at this head, outside the census's reach. packages/seat/src/custodian.ts:481: `if (import.meta.url === `file://${process.argv[1]}`) {` verbatim, the exact template-literal form #1622 names. This is shipped code in @cotal-ai/seat, not a script. examples/05-scale-showcase/src/swarm.ts:129: same form (with an env-var escape hatch beside it) examples/05-scale-showcase/src/observe.ts:216: same form reachability: I did NOT demonstrate a live failure for custodian.ts. The launcher spawns it as join(packageRoot(),'dist','custodian.js') where packageRoot() derives from import.meta.url, which Node has already realpath-resolved, so [...]
- A sibling import makes the converted scripts newly dependent on main-entry.mjs being adjacent, which changes --preserve-symlinks-main from working to crashing. measured: A symlink to scripts/check-attribution.mjs placed in another directory, run with `node --preserve-symlinks-main`: BEFORE (merge-base copy) exit 2 with the usage line; AFTER (this head) exit 1 with ERR_MODULE_NOT_FOUND for main-entry.mjs resolved beside the LINK rather than beside the real file. Without the flag, the same link works correctly at this head (exit 2, usage line), because Node resolves the main module through its symlinks and the sibling import then resolves next to the real file. scope: Only under [...]
- scripts/main-entry.mjs ships without a .d.mts, unlike its converted siblings. scripts/ carries live-job-conclusion.d.mts, post-publish-install-probe.d.mts, preflight-npm-publish.d.mts, pr-head-gate.d.mts, verify-publish-closure.d.mts, and others; there is no scripts/main-entry.d.mts. The file is `// @ts-check`'d and fully JSDoc-annotated, and no gen:*-dts script targets it, so nothing is broken. Noted only as an inconsistency a maintainer may want to settle.

**Not verified (7):**
- The PR's headline reproduction is macOS-specific (/tmp -> private/tmp). This host is Linux, where /tmp is a real directory, so I could not reproduce THAT path. I reproduced the underlying defect directly instead, with an explicit symlink and the merge-base source, and it behaved exactly as the issue describes.
- `pnpm changeset status`: cannot run in this lane (no local `main` ref; changesets exits 1 looking for the divergence point). The PR's 'ok' is neither confirmed nor contradicted. See BD3.
- The PR's before/after figures for smoke:doc-binding, smoke:live-job-conclusion, smoke:verify-publish-closure, smoke:npm-publish-preflight (80/80), smoke:publish-closure-gate, smoke:install-probe-ci, smoke:mutation-coverage (150), smoke:mutation-fixtures, and the four gen:*-dts no-diff claims were not re-run here. Host is 96 percent full on root and suites are slow; I spent the budget on the two questions the packet says matter most (does the guard fire, and is the guard itself graded). Not evidence of a problem, just not independently confirmed by me.
- The PR's claim that the suite is '1 passed, 7 failed' at the merge base 08cff353 was not reproduced: the suite does not exist at that commit, so reproducing it means transplanting the new file onto the old tree, which grades a construction rather than either commit. I verified the equivalent fact directly instead: the merge-base check-attribution.mjs exits 0 silently through a symlink and exit 2 with a usage line directly.
- packages/seat/src/custodian.ts: no live failing invocation demonstrated. packages/seat/dist is not built in this lane. The static shape is certain; the runtime consequence is reasoned, and I have flagged it as such in NB3 rather than asserting it.
- smoke:pr-head-gate and smoke:post-publish-install-probe, which the PR reports as red at both the merge base and this head: not independently re-run. I therefore neither confirm nor dispute that they are equally red at both commits.
- I did not run the full CI chain (600 suites). I confirmed registration programmatically instead: readCiSuites() includes 'smoke:entrypoint-guard' (true) and a deliberately bogus name (false) as a control.

davidfarah2003:
Checked source head `c78990af6e74fb51db62ffb78a9a54d20463ada0` combined with base `5a34b2b70271dea5ea8d921da543397034c37e77`. Main has since advanced, so these results do not establish the result of landing on current main.

The production helper fixes the exercised entrypoint defect. Realpath comparison handles the tested symlink and escaped-path cases, the intended consumers adopt it, and the registered five-mutant fixture detects its covered reversions. The required change is in the new parity tests.

### Blocking test gap

In `scripts/entrypoint-guard.selftest.mjs`, the six parity cells compare exit status, stdout and stderr between direct and symlink invocation. They never require either invocation to execute `main()`. If both paths silently exit 0 with empty output, every comparison succeeds.

This was tested with matched controls:

- Disabling the `check-attribution.mjs` entry guard on both paths left the suite green at 8/8. Reverting that same file to its old symlink-broken guard was killed in the same mutation-proof run, on the named attribution parity cell.
- Disabling the shared helper for all real invocations also left the new suite green at 8/8. On that identical mutation in the same test tree, the existing operator-literal celltamper suite fell from 175/175 to 46/175 and exited 1.
- Under total disablement, the attribution selftest still exited 0 while producing no output. The unchanged invocation produced its expected selftest output, also with exit 0. Exit status alone therefore cannot establish that this check ran.

Please add a script-specific execution witness to each parity probe, such as its expected refusal status and expected diagnostic, while retaining the comparison between paths. Register a mutant that disables `main()` on both paths and require it to fail the intended cell. Keep a killed control in the same file and run. The fixture description should also state this coverage boundary until the new control closes it.

### Existing backstop and its limits

The repository is not wholly blind to total helper disablement. At the assessed tree, mutation-reproof selection included the operator-literal celltamper fixture, which detected it. That suite is reached through mutation reproof rather than as a regular smoke-chain entry. Its selection for a helper change is currently an incidental dependency; no assertion was found that preserves that selection relationship. This mitigates the immediate exposure, while leaving the new entrypoint suite unable to detect its own stated failure class.

### Other limits and prose corrections

- `preflight-npm-publish.mjs` and `post-publish-install-probe.mjs` have source-census coverage only. Their normal entrypoints were not run through symlinks because they can pack artifacts or contact a registry.
- The census is limited to top-level `scripts/*.mjs`, is line-local, and strips comments by opening token. It does not cover nested files, other extensions, multiline guard forms, or similar guards outside `scripts/`. Its planted positive and negative controls work, but do not widen that scope.
- `doc-binding.mjs` uses a `RUN_AS_CLI` assignment rather than the direct `if` shape targeted by the per-file reversion anchors. It has a parity cell and shared-helper mutation coverage, but no matching per-file reversion mutant.
- The helper compares realpath strings, not device/inode identity. Case-insensitive filesystem aliases were not exercised in these reviews. The Linux runs also do not reproduce the specific macOS `/tmp` alias claim or Windows junction behavior.
- The earlier finding about a symlink placed outside the script directory with `--preserve-symlinks-main` remains a separate consumer-import limitation: a sibling helper import can resolve beside the link. Helper-level path comparisons do not by themselves establish that every converted script imports successfully in that layout.
- The body's statement that `check-attribution.mjs --selftest` succeeds with exit 2 is incorrect. The source returns 0 for a successful selftest and 2 for failure. The no-argument usage refusal is a different invocation and does return 2.
- A review run of `smoke:pr-head-gate` on the assessed integration with its workspace built passed 46/0. The body's historical 45/1 result should stay tied to the exact tree and environment that produced it. Likewise, an unbuilt workspace module refusal is not evidence of a product failure.

The full smoke chain was not independently rerun by these reviews. The targeted runs and mutation evidence above do not substitute for fresh checks on a revised source head and the then-current integration base.

L4XB:
Follow-up for the blocking test gap: #1688.

- **Execution witness:** each of the six parity probes now also requires both runs to reach the script's own refusal: its exit code and a stderr line only its CLI body prints. The cross-path comparison stays. A cell whose runs both skip `main()` now fails with `ranByRealPath:false, ranThroughLink:false, same:true`.
- **Mutants:** the fixture has 17 instead of 5, and all 17 are killed.
  - Each probed guard is switched off by both paths, one file per mutation.
  - `isMainEntry` answers false for every invocation, with the existing fast-path mutation as its killed control in the same file and run.
  - Doc-binding, pr-head-gate and verify-publish-closure get exact per-file reversions.
  - Two mutants break one half of the witness each: a load failure at exit 1, and a dropped exit code.
  - Against the suite as merged, the seven both-path disables and the two witness-half mutants survive, each with a killed control.
- **Coverage boundary:** the fixture text now lists what stays uncovered: the census is line-local and top-level only, the two unprobed scripts rely on the census alone, realpath strings are compared rather than inodes, `--preserve-symlinks-main` (reproduced, loud, not changed), and the guards outside `scripts/`.
- **Corrections:** `--selftest` exits 0 on a pass, not 2; the suite header is corrected. The custodian guard is now at `custodian.ts:578`. Its compiled file exits 0 silently when its path contains a space, and the two example guards do the same through a `/tmp` path; details and the remaining out-of-scope sites are in the PR body.
- **One review point did not reproduce:** the census does flag `upgrade-section-gate.mjs:894` once its exemption is removed (line 894 carries `process.argv[1]` and `import.meta.url` on one line).

PR #1599, opened 2026-09-13
fix(mutation-proof): treat the working tree as shared, in space and in time

L4XB:
Fixes #1574
Refs #1581

Both defects, since they are one root cause: the tool treats the worktree as private when it is shared.

## In space — two proofs, one tree

A lock in `tmpdir()`, **not** in the repo: this tool refuses to start on a dirty tree, so a lockfile in the worktree would refuse itself. The second run refuses with exit 3 and names the live pid and its command line.

Keyed on `realpathSync`, not `resolve`. On macOS `/var/folders/…` and `/private/var/folders/…` are one directory, and a key that distinguishes them hands each run its own lock — the exact failure the lock exists to prevent. **My first version had that bug**; the self-test caught it by planting a lock the tool then failed to see.

## In time — a killed proof

`process.on("exit")` covers a normal return and an uncaught throw. `SIGINT` / `SIGTERM` / `SIGHUP` handlers cover the kill that skips it, and re-raise with the default disposition afterwards so the exit status still reports the signal rather than becoming a tidy 0 — a wrapper reading `$?` must not be told the run merely finished.

Re-measured at this head, three variants of the same tool, same fixture, same signal, kill gated on an observed applied mutation (probe: build a temp repo, run the tool, `SIGTERM` as soon as `git status` shows the mutation, then read the child's `exit` event and the tree):

```
with handlers + the yield   $? = 143 (SIGTERM)   dirty after: 0
handlers, no yield          $? = 1               dirty after: 0   M —
no handlers at all          $? = 143 (SIGTERM)   dirty after: 1   M src/impl.js
```

Row 3 is the defect this PR exists for: the mutant is still applied, and the next command in that tree is grading a file someone broke on purpose.

Row 2 is a defect the review found in **this branch** — see *The kill path was unreachable*, below. An earlier revision of this body reported row 1 as `exit 0`; that was never measured at any head and is replaced by the table above.

## The SIGKILL case, from #1581

No handler catches it, so a stale lock — one whose pid is gone — is **refused loudly** rather than cleared silently, and the refusal names the backups the dead run may have left so the target can be identified by hashing one. `--clear-lock` is the documented way out, and it refuses to clear a lock whose pid is still alive.

I did not touch the exit code. #1581's reporter retracted that claim themselves (`process.exit(3)` is correct; their `$?` was reading the tail of a `| sed` pipeline), and I re-measured it here: bare exit **3** through a refusal.

## Cells and mutants

[...]

PR #1632, opened 2026-09-15
test(mutation-proof): grade the breadcrumb recovery, and document --recover

### LINKED ISSUE #1607: mutation-proof leaves a live mutation in the working tree when it is killed
`scripts/mutation-proof.mjs` works by editing the target file in place, running
the suite, then restoring it. If the process is killed between those steps, the
mutation stays on disk.

This is easy to hit. The tool's own pinned config for `mutation-coverage` runs
50 mutations at roughly 70 seconds each, so a full proof takes about an hour. Any
supervisor timeout, CI step limit, or interrupted session lands in the middle of
a mutation.

Observed after a killed run:

    $ git status --porcelain
     M scripts/mutation-coverage.mjs

    $ git diff
    -  if (discovered && !executeDiscovered) {
    -    fencedDiscovered++;
    -    console.error(`FENCED ${path}: discovered configs are not executed; ...`);
    -    continue;
    -  }
    +  void discovered; void executeDiscovered;

That is a pinned mutation, not a partial write. The file is syntactically valid
and behaviourally sabotaged, which is the worst combination: nothing downstream
has a reason to complain.

Why it is worth fixing rather than documenting. The tool already refuses to start
on a dirty tree, with a message that says the tree has to be recoverable
independently of the tool. That check is the right instinct applied at the wrong
end: it protects the tree from the user, then leaves the tree dirty itself. A
subsequent run refuses to start, and the reason it gives points at the user's
uncommitted work rather than at the previous crash.

The failure mode that concerns me is a measurement taken against the leftover.
Anything run against that tree, a self-test, a probe, a typecheck, is running
against a deliberately broken build while every surface still reports a clean
checkout. A green result there is meaningless and there is no local signal saying
so. In one case here a probe was started against a tree a proof had mutated, and
the only reason it was caught was that `git status` disagreed with a `git status`
from ninety seconds earlier.

Suggested direction, in preference order:

1. Restore on signal. Install SIGINT/SIGTERM/SIGHUP handlers that run the same
   restore the happy path runs. This closes the common case, though not SIGKILL.

2. Leave a breadcrumb before mutating and check it on startup. Write the original
   contents, or the fact that a mutation is live, to a file next to the config
   before the first edit and remove it after the last restore. On startup, if the
   breadcrumb exists, say so and offer to restore, rather than reporting a dirty
   tree as if the user had dirtied it.

3. Mutate a copy. If the suite can be pointed at a scratch checkout, the source
   tree never changes and the whole class disappears. Larger change, and it may
   conflict with wanting the real entrypoint under test.

(2) seems like the best ratio: it is small, it survives SIGKILL, and it converts
a silent wrong state into a named one.

Related: the same run also shows a separate coupling issue, filed separately,
where the suite's abort-on-first-failure makes each mutation's result depend on
where its cell sits relative to every other cell.

L4XB:
Refs #1607

#1657 (`cba592c0b`) gave `mutation-proof` a breadcrumb that survives SIGKILL and a recovery that reads it. Nothing grades either. This adds the mutation config, the self-test cells its mutations redden, and the documentation for `--recover`. The recovery itself is unchanged.

## What it adds

- **`scripts/mutations/proof-breadcrumb.json`**: seven mutations, each a different way for the mechanism to be present and useless. All seven are killed, each on its own named cell (table below).
- **Five new self-test cells** in `scripts/mutation-proof.selftest.mjs`:
  - `a proof that completes leaves no breadcrumb behind`. It sits right after the first completed proof, and that position is load-bearing. The suite is fail-fast, and a leftover breadcrumb records the clean hash of the file that `a dirty tree is refused` later dirties, so that cell would redden first, for the wrong reason.
  - `...and a breadcrumb outside the tree names the file, its pre-mutation hash and the backup`
  - `...and recovery puts the timestamp back, not the time of the recovery`. It uses the same tolerance and the same "not now" test as `a restored file keeps its original mtime`, for the reason given there.
  - `a breadcrumb whose file already matches its baseline is cleared, and the proof runs`
  - `a live mutation with no backup is refused by naming the interrupted proof, not the dirty tree`
- **`--recover` is documented** in the usage text and in AGENTS.md. `main` already accepts it: the generic argument parser (`:259-264`) turns it into `a.recover`, the entry point exits after recovery (`:649`), and the SIGKILL cell already calls it (`selftest :957`). A second flag such as the `--restore-live` from this branch's first version would only duplicate it.
- **The SIGKILL cell no longer leaves its suite running.** The cell kills the proof, not the suite the proof launched, and `run()` starts that suite in a process group of its own. Its mutated branch was a bare `setInterval`, so on `main` every self-test run leaves one `node hung-suite.mjs` behind for good (measured: PPID 1, still alive after the run). A proof of this config runs that suite up to eight times. The suite now exits once the self-test writes a release file, which happens right after the kill. It cannot key on the parent dying: the kill usually lands before the proof has spawned the suite, so the suite starts already reparented. A 60 s timeout bounds the rest.

## Mutants

`node scripts/mutation-proof.mjs --config scripts/mutations/proof-breadcrumb.json`:

| mutant | verdict | cell that reddened |
| --- | --- | --- |
| the breadcrumb is never written | KILLED | `...and a breadcrumb outside the tree names the file, its pre-mutation hash and the backup` |
| the recovery never runs | KILLED | `next invocation recovers from breadcrumb` |
| a proof that restored its file leaves the breadcrumb behind | KILLED | `a proof that completes leaves no breadcrumb behind` |
| the recovery does not put the timestamp back | KILLED | `...and recovery puts the timestamp back, not the time of the recovery` |
| the recovery writes the current time | KILLED | `...and recovery puts the timestamp back, not the time of the recovery` |
| a matching breadcrumb is refused as live | KILLED | `a breadcrumb whose file already matches its baseline is cleared, and the proof runs` |
| an unrecoverable mutation is reported and then measured | KILLED | `a live mutation with no backup is refused by naming the interrupted proof, not the dirty tree` |

`All 7 mutation(s) killed. The suite discriminates.` Afterwards the tree was clean and no `hung-suite.mjs` was running.

**Not asserted, and recorded in the fixture's `why`:** that a backup which does not hash to the breadcrumb is refused before anything is written over the file. The landed recovery has no such check to mutate. See the review thread.

## Everything run

| command | result |
| --- | --- |
| `node scripts/mutation-proof.selftest.mjs` | 75 checks passed (70 on `main`), three runs, no process left behind |
| `node scripts/mutation-proof.mjs --config scripts/mutations/proof-breadcrumb.json` | 7/7 KILLED, each on its named cell |
| `node scripts/mutation-coverage.mjs --gradable-only scripts/mutations/proof-breadcrumb.json` | ACCEPTED |

Local runs used a short `TMPDIR`: on macOS the default one makes the existing `bystander.sock` cell fail with `listen EINVAL`, because the socket path is longer than `sun_path` allows. That happens on `main` too and is unrelated to this change.

## History

The branch was brought up to date by merging `main` instead of rebasing. The merge commit takes `main`'s tree byte for byte and drops this branch's first implementation, which #1657 superseded. The two commits after it are the whole change. They are identical to a rebase onto `4b676f30` (same tree).


L4XB:
CI was red on this and it was mine, not flake. Fixed in `cca058bc`.

The reproof harness refused the shard because `scripts/mutation-proof.selftest.mjs` was red *before any mutation* — which is the right refusal, since every mutation would then have graded KILLED for a reason unrelated to the mutation.

The failing cell was the one this PR adds:

```
✗ --restore-live puts the file back byte for byte, with its timestamp, and clears the record
  {"status":0,"sha":true,"mtime":false,"tree":""}
```

Bytes and record were fine; the assertion was `statSync(killImpl).mtimeMs === mtimeOriginal`.

**Cell 9, one screen earlier in the same file, already says why that is the wrong question:**

> Not an equality test. `statSync` surfaces a nanosecond timestamp as a millisecond Date and `utimesSync` can only write that precision back, so a faithful restore still lands up to 1ms off. The defect is not imprecision, it is the mtime becoming NOW.

My new cell contradicted that reasoning with a `===`, and whether the round trip survives depends on the filesystem underneath. It passed on APFS and failed on the runner. Visible in the new diagnostic: a restored mtime here reads `1789472607462.6423` — fractional milliseconds, which no equality against a whole-millisecond original can survive.

So the cell now asks cell 9's question: the restored time predates the restore, and it has not moved from the original by more than the write can account for.

**Still discriminating, measured rather than asserted.** Removing the `utimesSync` from the `--restore-live` path reddens it:

```
✗ --restore-live puts the file back byte for byte, with its timestamp, and clears the record
  {"status":0,"sha":true,"mtimeOriginal":1789472594258,"mtimeRestored":1789472607462.6423,
   "movedMs":13204.64,"restoreStartedMs":1789472607379,"tree":""}
```

13 seconds of drift — the mtime became NOW, which is the defect the cell exists for. The relaxation costs no coverage of it.

`node scripts/mutation-proof.selftest.mjs` is 60 checks green on the branch. The diagnostic now carries `mtimeOriginal`, `mtimeRestored` and `movedMs` instead of a bare `false`, so whoever next sees it red is told by how much.

One thing I could not do: reproduce the runner's failure locally. macOS/APFS round-trips the value fine, so the fix rests on cell 9's stated contract and on the mutation above, not on a local repro of the red.

No production change in this commit; the tool is untouched.

L4XB:
Heads-up on ordering, since these two are both mine and both land in the same file.

**#1599** (the lock and restore work, `fix/1574-proof-lock-and-restore`) and **#1632** (the killed-run breadcrumb, `fix/1607-proof-breadcrumb`) both touch `scripts/mutation-proof.mjs` and `scripts/mutation-proof.selftest.mjs`. `git merge-tree` between their current heads reports a conflict, so whichever you merge second will need a rebase.

They are complementary in substance and neither depends on the other: #1599 makes `releaseEverything()` fail closed and fixes the signal disposition, #1632 records a breadcrumb before the mutating write so a SIGKILLed run is a named state rather than a dirty tree. Merge them in whatever order suits you and I will rebase the loser the same day — just say which.

Overlapping files, for the record:

```
scripts/mutation-proof.mjs
scripts/mutation-proof.selftest.mjs
```

davidfarah2003:
Order: #1599 first, then #1632.

#1599 is further through review (its third revision is under panel now), so it lands first and #1632 rebases onto main once it does. No action needed on #1632 until #1599 merges; the rebase request will come on this thread with the merge sha.

davidfarah2003:
Half of this landed by another route while the pull request sat, so it now conflicts with `main` and the diff cannot be rendered. Here is exactly what changed and what still stands.

**Landed already** (`cba592c0b`, "fix(scripts): recover a tree left mutated by an interrupted proof", closing #1607): the breadcrumb itself. `main` writes a record before the mutation, scans the temp dir on the next run, matches on a tree key, restores from the backup only when the restored bytes hash to `shaBefore`, and puts the original mtime back with `utimesSync`. The selftest references it in four places. So the behaviour this pull request is titled for is present.

**Not on `main`, and still worth landing from here:**

1. `scripts/mutations/proof-breadcrumb.json`. There is no registered mutation fixture for the breadcrumb on `main` at all. The recovery code is there and nothing proves it is killed when it breaks. The six mutations in your fixture separate six ways the mechanism can be present and useless, including the one that rounds the restored timestamp, which a byte comparison cannot see and which makes `smoke:dist-freshness` call an untouched package stale. That argument is right and the coverage does not exist anywhere else.
2. `--restore-live`. `main` has no such flag. Recovery happens only as a side effect of the next proof run, so an operator who wants the tree back has no command to ask for it.

A rebase onto current `main` narrowed to those two, dropping the parts `cba592c0b` already covers, should land without further discussion. If the landed recovery path and yours disagree on any behaviour, say which and that is worth its own look rather than a silent reconciliation.

L4XB:
Narrowed at `8b0d37e6`. The diff against `main` is the mutation config, five self-test cells, the `--recover` documentation, and one fix to the self-test's own SIGKILL cell. The body has the table: 7/7 killed, each on its named cell.

On the two items:

1. **The fixture** is re-anchored on `cba592c0b`'s code. I dropped the mutation that "writes a backup it never checked" because the landed recovery has no such check (item 1 below). In its place are a pair for the timestamp (the call is gone / the call writes now), one for "the recovery never runs", and one for the refusal that stops a mutation it cannot put back.
2. **The restore command already exists on `main`.** It is `--recover`: `parseArgs` turns any `--flag` into `a[flag] = true` (`:264`), the entry point exits after recovery when it is set (`:649`), and the SIGKILL cell calls it (`selftest :957`). The usage text never mentioned it. So instead of a second flag, this documents `--recover` in the usage text and AGENTS.md.

**Where the landed recovery and my first version disagree**, named as asked (line numbers on `main` @ `4b676f30`). None of these is changed here:

1. **The recovery writes the backup before checking it.** `copyFileSync(backupPath, targetPath)` (`:136`) runs before `sha(targetPath) === shaBefore` (`:137`). If the backup is truncated or otherwise not the original, the mutant is replaced by a third state, and the refusal that follows says only that the file does not match its baseline. By then the record of what was mutated is gone. My version hashed the backup first and refused without writing. For a tracked file git still recovers either way; for a gitignored target it would not.
2. **A restore that did not verify still deletes its backup and breadcrumb.** In `proveOne`'s `restore()`, `rmSync(backup)` (`:448`) and `clearBreadcrumb` (`:450`) run whether or not `ok` is true. The caller's `RESTORE FAILED for <file> — backup at <path>` (`:489`) then names a file that was just deleted, and the next run has no breadcrumb for the broken file. My version kept both when `ok` was false.
3. **Automatic recovery vs refusal.** `main` restores on the next run; my version refused and asked for an explicit command. With the hash check from item 1 in place, automatic seems fine to me, so I am not proposing to change it.
4. **Timestamp precision.** `main` records `mtime.getTime()` (`:98`) and restores through a `Date`; mine recorded the raw `mtimeMs`. Truncation moves the time back by less than 1 ms and never forward, so `smoke:dist-freshness` ordering is unaffected. I am not proposing a change here either.

Items 1 and 2 are small and I can send them as their own pull request if you want them. Each would come with a cell and a mutation added to this config.

Issue #1621, opened 2026-09-14
run ps keeps printing running for a revoked run whose driver is gone


davidfarah2003:
`cotal run revoke <runId> --local --by <who> --reason <text>` prints the documented acknowledgement, and `cotal run ps --local` keeps listing the run as `running` afterwards. Observed on a run whose driver process had died in a broker outage: twenty minutes after the revoke the table still showed `running`, the original holder, and `journalHigh` 0, with a pending ask as the last journal entry.

## Chain

- `revokeRunAdmission` (`packages/core/src/run-admission.ts:30`) creates the marker `revoked.v1.<endpoint>.<runId>` and touches nothing else. That is the design in `docs/workflows.md` 170-175: the admission and the status record are never rewritten, and a host reads the marker before its next channel effect.
- The status record is written only by a driver: `note(req, "running", ...)` on start, `noteFinal` with `completed`, `released` or `failed` on exit (`implementations/runtime/src/run-driver.ts:644-690`). A driver that dies mid-run leaves `running` behind and nothing else writes the record.
- `run ps` (`implementations/runtime/src/run-command.ts:386-407`) walks the spec keys, reads the status record and prints `st.state`. It never reads the admission store, so a revoked run and a live one print the same.
- `resume` does read the marker (`run-command.ts:329`) and refuses, and boot reconcile leaves the run parked and names it in the manager log. The table is the one surface that never learns.

A run with a dead driver and a revocation marker therefore occupies a `running` row indefinitely, and an operator counting capacity from the table counts it as live.

## Proposed

1. `run ps` reads the admission store next to the status record and prints `revoked` whenever the marker exists, whatever the status record says. `journal` carries `by` and the reason.
2. A `running` status whose holder lease reads as missing or expired prints `running (holder lost)`. A lease read that throws prints plain `running`, since a failed read is absence of evidence.
3. No automatic terminal fact. A revoke must not fabricate `failed` or `released` for a run nobody drove there; the journal owns the facts and both changes above are display.
4. `revoke` prints, after its acknowledgement, what `ps` will now show.

## Verification

- Admit a run, write its status as `running` with a dead holder, revoke it: `ps` prints `revoked`, `journal` carries `by` and the reason. Control: the same run before the revoke prints `running`.
- Revoke a run whose driver is alive: the driver refuses at its next channel effect (existing behaviour) and `ps` prints `revoked` from the first read after the marker.
[...]

PR #1633, opened 2026-09-15
fix(runtime,core): run ps reads the revocation marker beside the run record

L4XB:
Refs #1621

Implements proposals **1**, **3** and **4** for the `--local` table. Proposal 2 is not implementable against the stores that exist today, and the hosted table needs a wider change; both are measured and reported at the bottom, so this does not close the issue on its own.

This head answers the panel review of `1d3b9099` and merges `main` three times: `91a3207c` resolved the conflicts, `a2477d78` brought in #1632 without any, and `c111b440` brings in #1682, #1687 and #1690, where the only conflict was the cell sentinel in `run-command.smoke.ts`. `e18bdc4a` then corrects three sentences in the fixture's `why`. File and line references are against `e18bdc4a`.

## Reproduced first, on `08cff353`, against a real broker

A run held open at a checkpoint, so the record genuinely reads `running` while the assertions run, then revoked. The acknowledgement below is one line because that is all the base prints. The second line, `run ps now lists <run> as revoked; ...`, is added by this PR (proposal 4) and does not exist at `08cff353`. The holder reads `cli-run-…` because the base predates `43a4281a`, which moved local holders to `cli_run_…` on `main`.

```
$ cotal run revoke run-41bf457d9ed3f4eaf4816be1f053d29c --local --by smoke-operator --reason "broker outage"
run run-41bf457d9ed3f4eaf4816be1f053d29c on manager: revoked by smoke-operator (broker outage); its hosts refuse the next channel effect, and it is not resumed or taken back

$ cotal run ps --local
RUN                                   ENDPOINT  STATE      HOLDER            JOURNAL  FORKED-FROM
run-41bf457d9ed3f4eaf4816be1f053d29c  manager   running    cli-run-c781a8eb  0        -
```

Neither surface is broken in a way anyone can see. The marker is correct, the record is correct, and the table that reads only one of them prints a live row for a run that will never run again.

## The chain, confirmed at this head

`revokeRunAdmission` writes the create-only marker and touches nothing else. The status record is written only by a driver, so a driver that dies mid-run leaves `running` behind with nothing left to write anything else. `resume` reads the marker and refuses (`run-command.ts:337-341`), boot reconcile parks the run and names it in the manager log, and `ps` walked the spec keys and printed `st.state`.

## The change

- **`ps` reads the marker next to each record** and prints `revoked` for a run that carries one, whatever the record says.
- **The revoker and the reason go under the table**, on stdout. The marker carries both and the columns carry neither. The issue expected `journal` to carry them; it does not, it renders journal records only and never reads the admission store.
- **A marker `ps` cannot read prints `unchecked` in the `STATE` column** (`run-command.ts:443`): a store it cannot reach, a marker version it does not know, bytes it cannot parse. The reason and the record's own state go to **stderr** (`run-command.ts:472`), every other row still prints, and the command **exits 1** once the table is out (`run-command.ts:473`).
- **No terminal fact.** A revoke writes no `failed` or `released`; a cell asserts the record still reads `running` after the revoke.
- **`revoke` says what the table will now show**, so the two surfaces cannot drift apart again without the person who caused it seeing it.
- **`revoke` refuses an empty `--by` or `--reason`** before it writes (`run-admission.ts:185`). The argument parser passes `--by ""` through as an empty string, so the CLI on `main` could write a permanent marker that printed as `revoked by  ()`.
- **`readRunRevocation`** reads the marker alone. `readRunAdmission` refuses a missing or unreadable admission, which is right for a host about to perform a channel effect and wrong for a listing, where it would hide every other row over one run that predates the store.

`run-operator` already holds `$JS.API.STREAM.MSG.GET.KV_<admission bucket>`, so no grant changes.

### Why `unchecked`, why stderr, why exit 1

This departs from the issue's own rule for proposal 2, which printed plain `running` when a read failed. A panel reviewer showed what that costs: `awk '{print $3}'` and `grep running` read the record's word for a run that may be revoked, and the process exits 0.

- **`unchecked`, not `running?`.** `grep running` matches `running?`, so a qualifier on the record's word still counts the run as live. `unchecked` is one token, so `awk '{print $3}'` reads it whole, and it is the word this PR's first head already used for this case in `docs/workflows.md`.
- **stderr for the reason line.** The line carries the record's own state for the operator (`its record reads running`). On stdout it put the run id beside `running` again, and `grep running` counted the run through it even with the column fixed. A mutant that moves the line back to stdout is killed on that cell.
- **Exit 1, after every row.** The listing is partial. A partial table is more use than none, which is why the listing reads the marker through its own helper, and the status is how a caller learns it is partial without parsing the column. `ls` and `find` exit non-zero for an entry they cannot read in the same way, and `cotal down` sets 1 on a partial teardown. The cost is real: a marker is never removed, so an older `cotal` facing a marker version it does not know exits 1 in that space until it is upgraded. That build cannot say whether the run is revoked, so 1 is the right answer there.

The revoked footnote stays on stdout. It is not an error, and it has no `running` in it, but it is not a table row either: `awk '{print $3}'` prints `by` for it. A script that reads the column should filter rows on the `run-` prefix.

## Through the built binary

`node bin/dist/cotal.js` against a static-auth broker on `nats-server v2.12.12`: two runs held at a checkpoint, one pure run, one of the held runs revoked, a version-2 marker written raw on the other, then an empty revoke of the pure run. `XDG_CONFIG_HOME` points at scratch and `COTAL_ALLOW_CHECKOUT_SEED=1`, as `bin/smoke/setup-pure-live.smoke.ts` does, since the binary refuses to reconcile the machine-wide seed store from a checkout.

At `91a3207c` (`run-command.ts` and `run-admission.ts` from there, both packages rebuilt):

```
$ cotal run revoke run-2d51…6004 --local --by "" --reason ""           [exit 0]
run run-2d51…6004 on manager: revoked by  (); its hosts refuse the next channel effect, and it is not resumed or taken back
run ps now lists run-2d51…6004 as revoked; the status record is left as its driver last wrote it

$ cotal run ps --local                                                  [exit 0]
RUN                                   ENDPOINT  STATE    HOLDER            JOURNAL  FORKED-FROM
run-2914…b057                         manager   running  cli_run_6e0d26ef  0        -
run-8f1d…0250                         manager   revoked  cli_run_2c12c89c  0        -
run-2d51…6004                         manager   revoked  cli_run_725ee47d  0        -
manager/run-8f1d…0250: revoked by probe-operator (broker outage); the status record is left as its driver last wrote it
manager/run-2d51…6004: revoked by  (); the status record is left as its driver last wrote it
manager/run-2914…b057: run revocation is malformed: revocation not checked, so this row's state is the run record's own
--- stderr: (empty)

$ cotal run ps --local 2>/dev/null | grep -c running
1
```

At `a2477d78`:

```
$ cotal run revoke run-0f43…4602 --local --by "" --reason ""           [exit 1]
--- stderr:
✗ run revocation of run-0f43…4602 names who revoked it and why; by and reason must be non-empty

$ cotal run ps --local                                                  [exit 1]
RUN                                   ENDPOINT  STATE      HOLDER            JOURNAL  FORKED-FROM
run-8546…8010                         manager   unchecked  cli_run_28d46565  0        -
run-35d6…7df9                         manager   revoked    cli_run_b0b79520  0        -
run-0f43…4602                         manager   completed  cli_run_400f68d3  0        -
manager/run-35d6…7df9: revoked by probe-operator (broker outage); the status record is left as its driver last wrote it
--- stderr:
manager/run-8546…8010: revocation marker could not be read (run revocation is malformed); its record reads running

$ cotal run ps --local 2>/dev/null | awk '{print $1, $3}'
RUN STATE
run-8546…8010 unchecked
run-35d6…7df9 revoked
run-0f43…4602 completed
manager/run-35d6…7df9: by

$ cotal run ps --local 2>/dev/null | grep -c running
0
```

Run ids are shortened here; the probe printed them in full. Before the marker was written, the same table at both heads printed both held runs as `running` with exit 0 and an empty stderr.

## Cells

`implementations/runtime/smoke/run-command.smoke.ts`, against a real broker: **45 cells, against 29 on `main`**.

- Block 9 (8 cells, the reported case). Its control is the half that matters: the same run, the same table, one read **before** the marker exists. It also asserts that a complete listing leaves the exit code alone.
- Block 10 (6 cells, the unchecked path). A version-2 marker is written raw on a run held at a checkpoint, so its record really reads `running`. The cells check the column, that no stdout line puts that run beside `running`, the stderr line, that every other row still prints, and the exit status.
- Block 11 (2 cells). The empty revoke is driven through the command, and the cells check that it is refused and that no marker was written.

`packages/core/smoke/endpoint-binding.smoke.ts`: **5 cells** on `readRunRevocation` and `revokeRunAdmission`, importing `../src`. No marker, a written marker, a version-2 marker, a truncated marker, and the empty revoke.

Red against the source before each change, with this head's cells kept and only the named source file swapped. The two `b086974f` rows and the head row were measured at this head; the two `91a3207c` rows were measured at `2243f4f0`, when the suite declared 40 cells, so their totals are five lower than the same runs would give now:

```
run-command.smoke.ts, run-command.ts from b086974f:      38 ok, 7 failed   (block 9's three, block 10's four)
run-command.smoke.ts, run-command.ts from 91a3207c:      36 ok, 4 failed   (block 10: column, stdout, stderr, exit)
run-command.smoke.ts, run-admission.ts from 91a3207c:    38 ok, 2 failed   (block 11, both)
endpoint-binding.smoke.ts, run-admission.ts from 91a3207c: 205 passed, 1 failed (the empty revoke)
endpoint-binding.smoke.ts, run-admission.ts from b086974f: does not load (`readRunRevocation` is not exported there)
this head:                                               45 ok, 0 failed / 206 passed, 0 failed
```

Some new cells were green at every base above, on purpose. Here is what grades each one:

- **Preconditions**, not claims: block 9's "open and held" and block 10's "its record reads running".
- **The before-revoke control** in block 9. No mutant names it. Mutant 2 happens to redden it.
- **The complete-listing exit control** in block 9, graded by the positive-control mutant.
- **`revoke acknowledges with the revoker and the reason`**, which asserts the acknowledgement line `main` already prints. No mutant names it.
- **`the run record is untouched`**, which has no mutant (below).
- **`still lists every other run`**, graded by mutant 10.
- **Block 11 against `origin/main`'s `run-command.ts`**: green there because the refusal is in core. The row above that swaps `run-admission.ts` shows both of its cells red.
- **The core cells.** `returns the marker a revoke wrote` and `throws on a marker version it does not know` are graded by the core fixture. `answers undefined for a run with no marker` and `throws on a truncated marker` are named by no mutant. The swallow mutant reddens the truncated cell as well, but it is named on the version-2 cell.

## Mutants

`implementations/runtime/smoke/mutations/ps-revocation.json`, **all 12 killed** in one run:

| mutant | cell that reddened |
| --- | --- |
| the table stops overriding the record's state | `ps prints revoked for it afterwards, whatever the record says` |
| the reader looks at the admission key, not the marker | the same cell, and seven others (below) |
| the revoker and the reason stop printing | `...with the revoker and the reason under the table...` |
| `revoke` stops saying what the table will show | `...and says what the table will now show...` |
| an unreadable marker keeps the record's state in the column | `ps prints unchecked in the STATE column of a held run...` |
| the failed read goes to stdout | `...and no stdout line names that run beside running...` |
| the failed read is not reported | `...gives the failed read and the record's own state on stderr` |
| a listing with an unchecked row exits 0 | `...and exits 1, because the listing is incomplete` |
| positive control: every listing exits 1 | `a listing whose every marker was read leaves the exit code alone` |
| one unreadable marker refuses the whole listing | `...still lists every other run, so one unreadable marker hides nothing` |
| the marker reader answers no marker when it could not look | `ps prints unchecked in the STATE column of a held run...` |
| the writer accepts an empty by and reason | `revoke refuses an empty --by and --reason` |

`packages/core/smoke/mutations/run-revocation-marker.json`, **all 3 killed**. It holds the three core mutants again, against the core cells: `returns the marker a revoke wrote`, `throws on a marker version it does not know`, and `refuses a marker with an empty by and reason`. Every mutant in both fixtures was killed, so no verdict needed `mutation-proof`'s same-file positive control. Each edited file still had more than one kill in its run: 9 in `run-command.ts` and 3 in `run-admission.ts` for the runtime fixture, and 3 in `run-admission.ts` for the core one. The one explicit positive control is the unconditional-exit mutant. It shows block 9's exit control can see a listing that always exits 1.

**What these mutants do not cover:**

- **No terminal fact** has a cell and no mutant. Nothing in this change writes the record, so there is no line to break, and the cell passes at every base.
- **Store unreachable** is not driven by any cell. It lands in the same `catch` as the version-2 marker, which mutants 5 to 8 and 10 grade, but no cell takes the store away.
- **The manager's `revoke`** (`run-hosting.ts:249`) gets the empty-field refusal from the same core function, and no cell sends it an empty revoke (`smoke:run-host-live` drives it only with a named revoker and reason).
- **Mutant 2 is not isolated.** Reading the admission key makes every row `unchecked`, so it reddens eight cells. The first is block 2's `with its state`, and block 9's before-revoke control is one of the others. It is killed on the claim it breaks, not on the first line it breaks. The fixture's note records this.
- **`mutation-coverage --gradable-only` refuses the runtime fixture**, because that suite imports `@cotal-ai/core` by name and the three core mutants are only visible through the build in their command. `mutation-proof` kills them all the same. On `main`, `run-host-isolation.json` is refused for the same reason, and a full `--gradable-only` pass at `4a243b92` refused 159 of the 449 fixtures it enumerated. The core fixture is accepted, which is why it exists.

**Two things the mutants measured that contradict the issue, both recorded in the fixture's `why`:**

1. The issue predicted that reading the admission key would make a live admission read as a revocation. It does not. `revocationSnapshot` is a closed-field validator, so an admission record fails it, the read throws, and `ps` prints the row as `unchecked`. Same cell, different mechanism.
2. Sharing one marker reader between `readRunAdmission` and `readRunRevocation` was the first shape written here, and it made that mutant **ungradable**. Pointing the shared reader at the admission key breaks every host read too, `resume` throws, and the suite dies before the cell runs. The host keeps its own two lines, with the reason (`run-admission.ts:213-219`).

## The empty-field refusal is at the writer only

`revocationSnapshot` still reads a marker with an empty `by` or `reason` as a revocation. A marker is never removed, so one already written that way stays. Refusing to read it would make that run's admission unreadable everywhere. `releaseConclave` (`mesh-handler.ts:2121-2126`) needs a readable admission for the cleanup a revoked run is still owed, and an unreadable one refuses it. Such a marker still prints `revoked by  ()`.

## Two things this does NOT do

**Proposal 2, `running (holder lost)`, has no lease to read.** The recorded holder is `cli_run_<hex>`, minted per invocation in `cliHolder()`, for a local drive, and `${manager endpoint id}.${takeoverId}` for a hosted one. The tree has two lease families, the delivery-shard lease (`leaseKey`) and the manager-instance lease (`managerLeaseKey`), and neither is keyed by either holder form. Reading "the holder lease" for a local run would find nothing for a healthy run just as surely as for a dead one, so the display would say `holder lost` for every local run alive or not. Making it work needs a per-run holder lease that does not exist: a record, a renew loop in the driver, a TTL, and a decision about who writes it.

**The manager-hosted `run ps` has the same defect.** `implementations/runtime/src/run-host.ts` `list()` (line 181) builds `RunListRow` from the record alone, so it still prints only the four recorded states while `--local` can also print `revoked` and `unchecked`. Fixing it needs an optional field on that served answer shape plus a manager credential that may read `KV_<admission bucket>`. It also cannot be driven from this suite, which runs `--local` against a bare broker. `docs/workflows.md` says the hosted table reads the record alone.

## Everything run

**Environment:** macOS (Darwin 25.6.0, arm64), Node v24.2.0 (CI pins 22, which this host does not have), pnpm 11.1.2, `nats-server v2.12.12` (the version CI pins). Everything below ran at `e18bdc4a` unless the row says otherwise.

| command | outcome |
| --- | --- |
| `pnpm smoke:runtime-run-command` | **45 ok, 0 failed** |
| the same, with the three pre-change sources above (run at `2243f4f0`, whose cells and other sources match this head) | 33/7, 36/4, 38/2 |
| `pnpm smoke:ep-binding` | **206 passed, 0 failed** |
| the same, with `run-admission.ts` from `91a3207c` (run at `fd2ba194`, same match) | 205 passed, 1 failed |
| `node scripts/mutation-proof.mjs --config implementations/runtime/smoke/mutations/ps-revocation.json` | **All 12 killed** |
| `node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/run-revocation-marker.json` | **All 3 killed** |
| `node scripts/mutation-coverage.mjs --gradable-only` on the core fixture | ACCEPTED |
| the same on the runtime fixture | REFUSED (by-name import, above) |
| `node scripts/mutation-reproof.mjs --base 5c63e4af --head c111b440` (the CI gate's selection) | **MUTATION REPROOF OK**: 8 fixtures selected, 8 discriminated, 0 unmeasured pre-red, 49 mutants killed |
| the built-binary probe above | as quoted |
| `pnpm smoke:runtime-run-command-usage` | 19 ok, 0 failed |
| `pnpm smoke:runtime-run-driver-auth` | 74 passed, 0 failed |
| `pnpm smoke:frozen-exports` | OK (11 passed; 22 arrays + 16 plain-objects scanned) |
| `pnpm smoke:mutation-fixtures` | OK (451 fixture files, 2656 anchors) |
| `pnpm check:docsbundle` | clean (41 pages pass `check:docs-voice`) |
| `pnpm typecheck` | clean (the runtime check includes `smoke/`; core's does not, so `endpoint-binding.smoke.ts` was also checked with a one-off `tsc -p` over core's config: clean) |
| `pnpm smoke:run-host-live` | 51 passed, 0 failed |
| `pnpm changeset status --since origin/main` | patch for the fixed group, no minor or major |
| `node scripts/check-attribution.mjs --range origin/main..HEAD` | clean, 11 commits |
| `pnpm smoke:gate-inventory` | **fails at this head**. `smoke:delivery-pidfile-ownership` and `smoke:reap-absent-refusal` are run by nothing, and `bin/smoke/ci-suites.txt` names `smoke:connection-status` twice. This branch changes nothing under `package.json`, `bin/`, `scripts/` or `.github/` relative to `main`: the inputs it fails on, `package.json` and `bin/smoke/ci-suites.txt`, are byte-identical to `main`'s, and a clean checkout of `main` reports the same three findings. Every open pull request fails `smoke (shard 1/4)` on it. |

Not run here: `smoke:mutation-coverage`, `smoke:mutation-reproof` and `smoke:mutation-reproof-full-partition`. They are self-tests of the mutation tooling, which this branch does not change. The full `smoke:ci` shard set was not run either.

`docs/workflows.md` gains the `run ps --local` paragraph under the revoke section, plus one sentence in the revoke paragraph about empty `--by` and `--reason`. The docs bundle is regenerated. A changeset names `@cotal-ai/core` and `@cotal-ai/runtime` at patch.


L4XB:
CI was red on `unit` / "Docs bundle in sync". Fixed in `1d3b9099`.

`check:docsbundle` runs the generators and then `git diff --exit-code` over `docs/mcp-tools.md` and `extensions/connector-core/src/docs-bundle.generated.ts`, so an edit to a page the bundle embeds has to be regenerated in the same commit. This branch added the `ps` revocation paragraph to `docs/workflows.md` and left the bundle holding the previous body.

One line moved, the "Workflow runs" entry, and it now carries exactly the paragraph the page carries:

```
+ "`ps` reads the marker beside each run record and prints `revoked` for a run that carries one,
   whatever state the record itself holds, with the revoker and ..."
```

`gen:tooldocs` reports the same 22 tools and leaves `docs/mcp-tools.md` byte-identical, so nothing about the tool surface moved — this is purely the embedded page body.

One note for anyone reproducing it: regenerating needs the workspace built first. `gen:tooldocs` imports `@cotal-ai/core/dist` through `tool-specs.ts` and dies on resolution without `pnpm --filter @cotal-ai/core --filter @cotal-ai/workspace --filter @cotal-ai/seat --filter @cotal-ai/lang build`. `check:docs-voice` passes 41 pages either way.

davidfarah2003:
Cold read complete: one independent review of `1d3b9099cdea17385c756c2540ddca58daad35e1`.

Relayed by the orchestrator. The measurements below belong to the reviewer.

Verdict `REQUEST_CHANGES`, two blocking items. The reviewer confirms the fix for the reported case end to end against a real broker and through the real binary: a revoked run whose record still reads running now prints revoked, with the revoker and reason under the table, and the shipped mutants die at named cells. The request rests on the instrument as a whole: when the marker read fails for any reason, the STATE column prints the record's own word with no in-column qualification and the caveat goes to a separate line after the table, so a script reading the column sees the word the issue was filed about and the process exits 0. The reviewer shows that this path is graded by nothing, with a killed positive control in the same file and run, while the body, the changeset and the workflows doc all describe the behaviour. Three body defects and four non-blocking notes follow.

---

**Blocking (2):**
- B1. The STATE column still prints a confident `running` for a run that IS revoked whenever the marker read fails; the caveat is out-of-band. This is the instrument question. `ps` is read by an operator scanning a column and by scripts doing `awk '{print $3}'` or `grep running`. When `readRunRevocation` throws, the row keeps the record's word with no in-column qualification and the only warning is a separate line printed after the table. For a run that is genuinely revoked, that column prints exactly the word issue #1621 was filed about, with the same confidence as a healthy row, and the process exits 0. probe 1 cases C/D/E/F, probe 2 cases J/K, probe 3 case M, probe 5 cases P and Q with [...]
- B2. The unchecked path is graded by nothing, proven with a killed positive control in the same file and the same run. The body lists 'A marker that could not be read is named as unchecked and the row keeps the state its record carries' among the change's bullets, the changeset states it, and docs/workflows.md documents it. Repo convention is that a test which passes with the fix reverted proves nothing, and here there is no test at all.

**Body defects (3):**
- D1. PR body, 'The change' bullet list and the 'Mutants' table. The body presents four mutants as the measurement of the change and lists 'A marker that could not be read is named as unchecked and the row keeps the state its record carries' as one of the shipped behaviours, without stating anywhere that this behaviour is graded by no cell. A reader of the body would reasonably conclude the mutant table covers the bullet list. probe mutation SURVIVED with a killed control in the same file and run (see B2).
- D2. PR body, 'Everything run' table. The table reports outcomes for eleven suites plus typecheck, docs-voice and changeset status, all as bare outcomes with no environment note beside them, while the environment line ('Node 26.8.2, nats-server 2.14.0, macOS') sits at the very bottom after the two 'does NOT do' sections. Several of those figures are the load-bearing evidence for the change.
- D3. PR body, 'Reproduced first, on 08cff353, against a real broker'. The quoted repro block shows the revoke acknowledgement WITHOUT the second line (`run ps now lists ... as revoked`), correctly, since that line does not exist at the base. But the block is introduced as the reproduction and the reader must notice the missing line themselves to understand which surface is being shown. A one-clause note would remove the ambiguity.

**Non-blocking (4):**
- N1. `revoked` is a display word outside the recorded state vocabulary. `RunState` is `"running" | "released" | "completed" | "failed"` (packages/core/src/run-record.ts:80). The STATE column can now print a fifth word that no record ever holds. This is deliberate and documented as display-only, and it is the right call versus fabricating a terminal fact. But any existing consumer that switches on that column now has an unhandled case, and the hosted `run ps` will keep printing only the four, so the two surfaces answer with different vocabularies until run-host.ts catches up.
- N2. A marker with empty `by` and `reason` renders as `revoked by ()`. `revocationSnapshot` accepts empty strings for both, unlike `provenanceSnapshot`'s operator branch which requires length > 0. Constructed and observed: `manager/run-69f58f2476ea796b1eb35f8c610182c4: revoked by (); the status record is left as its driver last wrote it`. The CLI itself requires --by and --reason, so this needs a non-CLI writer to reach; it is cosmetic rather than misleading.
- N3. One extra serial point read per row. `ps` now performs a leader-served `STREAM.MSG.GET` per row, sequentially inside the existing per-row loop, in addition to the record read. On this host one listing of a 6-row table took ~217ms. A space with many recorded runs pays a round trip per row. No grant change is needed (confirmed), and the reads are independent so they could be batched if it ever matters.
- N4. The exported `readRunRevocation` refusal contract is only measured through the ps cell. It is covered by smoke:frozen-exports as surface, and mutant 2 kills through it, but nothing asserts its throw-on-unreadable behaviour independently of the runtime table. If the ps cell moved, the core contract would be unmeasured.

**Not verified (7):**
- `pnpm changeset status` -> ok. Cannot be reproduced in this lane: it exits with `Failed to find where HEAD diverged from "main"`, because the lane is a detached review checkout on branch `review` with no local `main`. That is a lane artifact, not evidence against the claim. I neither confirm nor dispute it.
- smoke:runtime-run-driver-auth (74 passed), smoke:mutation-coverage (150 passed), smoke:mutation-reproof (143 passed), smoke:mutation-reproof-full-partition (4 passed, 438 fixtures), smoke:gate-inventory (OK). Not run. The shipped mutation fixture alone took 16m44s on this host and the full set would have taken the remaining budget. I prioritised the instrument questions and the coverage probe over re-confirming unrelated green suites. No reason to doubt them; simply not measured by me.
- The manager-hosted (non `--local`) `run ps`. I read `run-host.ts` `list()` to its closing brace and confirmed it builds RunListRow from the record alone, so the same defect is present there as the body says. I did not drive a manager-hosted listing end to end, which would need a manager process this suite does not stand up.
- The 'shared reader made the mutant ungradable' history in the fixture's `why` and in de20e9ec7. I verified the OUTCOME (the shipped split yields a clean named red for mutant 2) but did not reconstruct the rejected shared-reader shape to confirm it dies at `resume` with a stack trace.
- The store-unreachable variant of B1 and the malformed/truncated marker variants, through the REAL binary. B1 was confirmed end to end through the real `cotal` binary for the case that matters most (a held, genuinely `running` run, revoked, printing the literal word `running` at real exit status 0), plus the happy path and the scripted-reader pipeline. The store-loss variant and the other malformed-marker shapes were established in-process against a real broker and were NOT re-driven through the binary. They share the single code path that B1's binary run exercised, so I expect them to behave identically, but I did not observe that and do not claim it.
- Whether a real deployment reaches the store-unreachable state of B1 without operator error. I constructed it by removing the admission stream's on-disk files and restarting the broker. That is a real way a broker loses a stream, but I did not establish a naturally-occurring path to it. The version-2-marker and malformed-marker variants of B1 need no store loss at all, and the forward-compatibility variant only needs a newer revoker to exist, so B1 does not rest on this case alone.
- PR prose beyond the packet fetch time. The body and issue were read only from the packet, fetched at 2026-09-15T17:14:00Z. I did not re-fetch, so any later edit to the body is outside this review.

L4XB:
Thanks for the cold read. Everything below is at `e18bdc4a`, which merges `main` three times (the conflicts in `91a3207c`, #1632 without any in `a2477d78`, and #1682, #1687 and #1690 in `c111b440`, where only the cell sentinel conflicted) and then corrects three sentences in the fixture's `why`. I measured it on macOS with Node v24.2.0 and `nats-server v2.12.12`. The PR body is rewritten, and the full outputs are there.

**B1, the STATE column (fixed).** A row whose marker cannot be read now prints `unchecked` in the column, never the record's word. The reason and the record's state go to stderr: `manager/<run>: revocation marker could not be read (<reason>); its record reads running`. Every other row still prints, and the command exits 1 after the table.

- I picked `unchecked` over `running?` because `grep running` still matches `running?`. `unchecked` is also the word this PR's first head already used for this case on the workflows page.
- The reason line moved to stderr because it has to carry the record's state for the operator. On stdout, `grep running` counted the run through that line even with the column fixed.
- Exit 1 because the listing is partial, which is how `ls`, `find` and `cotal down` report a partial result. The cost is in the body: a marker is never removed, so an older build facing a newer marker exits 1 in that space until it is upgraded.

Through `node bin/dist/cotal.js`, with a held run carrying a version-2 marker:

- At `91a3207c`: `running`, exit 0, and `grep -c running` prints 1.
- At `a2477d78`, and unchanged by the later merges, which touch neither `ps` nor `revoke`: `unchecked`, exit 1, and `grep -c running` prints 0. `awk '{print $3}'` reads `unchecked`.

The reported case still works, and a revoked run prints `revoked` at both heads.

I did not drive the store-unreachable variant. It takes the same `catch` as the version-2 marker, but no cell removes the store.

**B2, grading the unchecked path (fixed).** Block 10 of `run-command.smoke.ts` writes a version-2 marker raw on a run held at a checkpoint. After its precondition cell it has five claim cells: the column, stdout, the stderr line, the other rows and the exit status. Block 9 gains a control that a complete listing leaves the exit code alone.

- `ps-revocation.json` gains six mutants, one per claim, plus an unconditional-exit mutant as the positive control for that control. All 12 were killed in one run, with 9 kills in `run-command.ts` and 3 in `run-admission.ts`.
- With `run-command.ts` from `91a3207c`, the suite reads 36 ok and 4 failed, and the four are block 10's column, stdout, stderr and exit claims. That run was measured at `2243f4f0`, when the suite declared 40 cells; with `run-command.ts` from current `main` at this head it is 38 ok and 7 failed.

**D1 (fixed).** The body now has a "What these mutants do not cover" list. It names the no-terminal-fact cell (no mutant), the store-unreachable variant (no cell) and the manager's revoke path (no cell).

- It also says mutant 2 is no longer isolated. It reddens eight cells, block 2's `with its state` first.
- It says `mutation-coverage --gradable-only` refuses the runtime fixture, because the suite imports core by name. On `main`, `run-host-isolation.json` is refused for the same reason.

**D2 (fixed).** The environment line now sits directly above the "Everything run" table.

**D3 (fixed).** The repro block now says the second acknowledgement line is added by this PR and does not exist at `08cff353`. It also says why the holder there reads `cli-run-`: `main` has used `cli_run_` since `43a4281a`.

**N1 (not changed).** The column now has two display words outside `RunState`: `revoked` and `unchecked`. The hosted `run ps` still prints only the four recorded states. `docs/workflows.md` says the hosted table reads the record alone, and the body keeps the hosted fix out of scope.

**N2 (fixed, at the writer).** I have to correct the premise. The empty marker is reachable through the CLI.

- `parseCommandArgs` hands `--by "" --reason ""` through as `""`, and `revoke` only checked for `undefined`.
- At `91a3207c` the built binary accepted it with exit 0 and printed `revoked by  ()` in both the acknowledgement and the table.
- `revokeRunAdmission` now refuses an empty `by` or `reason` before the create (`run-admission.ts:185`). That covers the CLI and the manager's `revoke`. Through the binary it now exits 1 and writes no marker.

The reader stays tolerant on purpose. A marker is permanent, and refusing to read an existing one would make that run's admission unreadable, which also refuses the conclave cleanup a revoked run is still owed (`mesh-handler.ts:2121-2126`). So a marker written before this change still prints `revoked by  ()`.

**N3 (not done).** The listing still makes one serial leader read per row, and I did not measure it.

**N4 (addressed).** `endpoint-binding.smoke.ts` imports `../src` and now has four `readRunRevocation` cells and one for the empty revoke:

- no marker answers undefined
- a written marker is returned
- a version-2 marker throws
- a truncated marker throws

The new `run-revocation-marker.json` holds the three core mutants against those cells, and all 3 were killed. `mutation-coverage --gradable-only` accepts that fixture. No mutant names the no-marker and truncated cells.

**Your not-verified list.**

- `pnpm changeset status --since origin/main` gives patch for the fixed group, with no minor and no major. A bare `changeset status` would compare against this clone's stale local `main`.
- `smoke:mutation-coverage`, `smoke:mutation-reproof` and `smoke:mutation-reproof-full-partition` were not run. They grade the mutation tooling, which this branch does not change.
- `smoke:gate-inventory` fails at this head: two suites that nothing runs, and one duplicate entry in `ci-suites.txt`. This branch touches none of them, and the body lists them. I did not run it on `main`.
- I ran the reproof gate the way CI selects it: `mutation-reproof.mjs --base ee64b03c --head a2477d78`. It reports `MUTATION REPROOF OK` for all 8 fixtures it picks for this diff (49 mutants). `smoke:run-host-live`, which drives the manager's `revoke`, passes 51 of 51.

Issue #1576, opened 2026-09-13
Delivery daemon is a hard dependency for spawn, retirement and join, but its absence has no surface of its own


NomadaDigital01:
**Environment:** Cotal 0.48.2. Split broker/manager topology: a broker container running `cotal up --space main --host 0.0.0.0` (nats + JetStream + delivery), and a separate host running `cotal supervise` with 30 managed PTY agents.

## What happens

When the delivery daemon is not bound, agent lifecycle operations fail and peers cannot join — but nothing reports the daemon as the cause except a single line inside one individual PTY, on one operation, after the fact.

## Reproduced

After a cold boot (power outage), the broker logged this once, at info level, and continued:

```
• delivery daemon not yet ready (responder not bound) - boot durable joins will reconcile when it is
```

That reconciliation never happened; the daemon never bound. `cotal up` (pid 87) had exactly one child for the next 22 hours:

```
   87  22:25:12  node /bin/cotal up --space main --host 0.0.0.0
  291  22:25:08  nats-server -c /root/.cotal/auth/server.conf
```

Over those 22 hours, **18 of 30 agents drifted out of the mesh**: present in `cotal ps`, absent from `cotal endpoints`, marked `mesh offline`. Six hours after boot the count was 0; by hour 22 it was 18. Throughout, `systemctl is-active` reported `active`, because the daemon is a child of `cotal up` and `cotal up` stayed alive.

Attempting to respawn a single agent surfaced the real cause for the first time:

```
✗ the name "worker-seo" is reserved pending retirement: its previous agent's despawn started
that lifecycle's teardown ... (last attempt: the static retirement did not complete (the delivery
daemon is not reachable on the ctl.delivery-admin rail (no responders:
'cotal.main.ctl.delivery-admin.local.UDAE...'); a restart on an auth mesh cannot verify-evict the
superseded serve family principal "local.UCGL..." without the liveness oracle. Start the delivery
daemon (`cotal up` runs it) and retry — eviction is never skipped (SPEC 13.1)); the name stays held
```

Starting the daemon by hand (`cotal deliver --space main`) and retrying the same spawn succeeded immediately, and the agent rejoined the mesh within 40 seconds.

A second deployment hit the same dependency from the peer side. A `cotal join` with valid, non-expired creds fails with:

```
! NATS permission denied: cannot publish "$JS.API.CONSUMER.INFO.DM_main.dm_local-<sub>-<rand>"
[...]

PR #1635, opened 2026-09-15
feat(core): let a credentialed peer ask which plane is broken (#1577)

### LINKED ISSUE #1577: No peer-readable liveness: a participant cannot determine whether the manager or the delivery daemon is alive
**Environment:** Cotal 0.48.2, 30-agent deployment, split broker/manager topology, plus one external peer joining over the LAN.

## What happens

The control subjects that would answer *"is the manager alive?"* — e.g. `cotal.main.ctl.manager.local.<sub>` — are owner-only. A peer holding valid credentials cannot ask. When a join or a send fails, the peer cannot distinguish a credential problem from a dead manager, an unbound delivery daemon, or a broker that is perfectly fine.

The practical consequence is that every failure presents as a credential failure, because that is the only hypothesis the peer can form.

## Why it matters, with evidence

Over one week operating a 30-agent deployment, **six independent surfaces reported success over a failure**:

| Surface | Reported | Reality |
|---|---|---|
| `systemctl is-active` | `active` | supervised child was dead |
| `cotal ps` | 30 agents running | 22 had lost mesh presence |
| `cotal input` | `✓ sent N bytes` | text never dispatched (#1333) |
| `tmux send-keys` | exit 0 | nothing dispatched |
| operator `pgrep` | daemon alive | matched its own command line |
| third-party MCP | "posted to #general" | its join was broken |

In every case, diagnosis cost hours rather than minutes. In every case the missing piece was the same: **no component can be asked whether it is alive by anyone who is not its owner.** The operator is left inferring liveness from side effects, and each inference is its own opportunity to be wrong — as the `pgrep` line above shows, that reasoning fails in both directions.

## What we would want

A read-only liveness surface that a credentialed peer may query — in the shape of the Synadia micro protocol's `$SRV.INFO` — reporting at minimum whether the manager and the delivery daemon have bound responders for a given space.

It does not need to expose state, ownership, or any roster. Presence alone would collapse most of the diagnosis time above, because it would let a failing participant answer the first question an operator always asks: *which plane is broken?*

davidfarah2003:
Fixes #1577. Refs #1594, #1576.

## What this fixes

The subjects that answer "is the manager alive" are owner-only, so a peer holding valid credentials cannot ask. When its join or its send fails, the peer cannot tell a credential problem from a dead manager, an unbound delivery daemon, or a broker that is fine. Every failure presents as a credential failure, because that is the only hypothesis the peer can form.

The reporter ran a 30-agent deployment for a week and recorded six independent surfaces that each reported success over a failure, including a `pgrep` that matched its own command line and so failed in both directions. In every case the missing piece was the same: nothing could be asked whether it was alive by anyone who did not own it.

## What changed

**A read-only liveness rail, `live.<plane>.<owner>.<actor>`.** A peer publishes a presence probe on its own principal's form and learns whether the manager and the delivery daemon have bound responders for the space. The shape follows the micro protocol's `$SRV.INFO` (well-known, read-only, presence-only request and reply) but rides a Cotal subject inside the space. The literal `$SRV` tree sits outside per-space account isolation, outside every grant builder, and outside the prefix subject audit, so it is not used.

**Its own tier, on purpose.** Not `ctl`, because `ctl.<service>` is only the delivery daemon since 1d and answering manager liveness there would reintroduce the manager control rail that change deleted. Not `ep`, because the endpoint rails carry the typed command surface with its authz and capability grammar, and a presence probe must not acquire that reach. The plane rides the subject rather than the payload so a credential can be scoped to one plane by the broker.

**Presence is the whole answer.** The wire reply is exactly `{plane, responder}` with `responder` one of `bound`, `unbound`, `stale`, `unknown`. No holder, no pid, no workspace root, no instance id, no runtime, no roster. This is why the probe is a request rather than a lease read grant: the manager lease row carries the operator's resolved workspace path and a pid, so the responder reduces the row to one enum and the row never crosses the wire. A peer gains no read of either lease bucket, cannot probe under another principal's identity slots, and cannot subscribe a responder's serve filter.

**Unknown stays first class.** The #1594 classifier moves from the CLI into `@cotal-ai/core` (`packages/core/src/liveness.ts`) and the CLI module aliases it, so all three surfaces grade by one rule instead of a copy that could drift. Only the broker's own no-responders answer becomes `unbound`; a timeout, a refusal and an unreadable reply are `unknown`, because each is a failure to find out. A reply is never counted as health merely for having arrived.

**Responders.** The manager serves its plane from whether its service handle is bound; the delivery daemon serves its plane from the readiness lease through the shipped classifier, queue-grouped so one probe reaches one instance. Each reply names the responder that answered by an opaque per-bind token, so a caller that probes more than once can tell two responders apart from one that changed state; a single probe still samples one instance and cannot report a split by itself. `Endpoint.probeLiveness(plane)` is the peer-side call; `serveLiveness` and `serveDeliveryLiveness` are the responder side.

Files: new `packages/core/src/liveness.ts`, `packages/core/smoke/liveness-peer.smoke.ts`, its mutation fixture and a changeset (`@cotal-ai/core` minor, `@cotal-ai/cli` patch); edits to `subjects.ts`, `provision.ts`, `endpoint.ts`, `index.ts`, the manager, the delivery daemon, the CLI responder lib and `package.json`.

## How it is graded

`smoke:liveness-peer`, 58 cells against a real auth broker, 0 failed at `7ae2134b`. Bound and unbound are the same probe under opposite conditions one run apart, so a surface that answered identically at every input could not pass both groups.

| axis | cells |
|---|---|
| manager bound / unbound | B1 / A1, B4 |
| delivery bound / unbound | B3 / A2, B2 (absent lease), B2b (claimed but not ready) |
| unknown, per plane | C3 manager, C1 delivery |
| unknown as a rule | C4 timeout, C5 refusal, C6 malformed, C9 no verdict |
| a non-owner can read presence | D1, plus every A and B cell is a peer probe |
| a non-owner cannot read more | D2 identity slots, D3 serve filter, D4b reply lane, D5/D6 manager lease bucket, D7 delivery lease write, D8 exactly two keys on the frame, D9 closed plane set |
| the remote-manager credential answers the plane it binds, and only the supervisor actor does | E6 to E9 |
| the rejection notice cannot end an embedder that attached no `error` listener | F1 to F5, driven in a child process with nothing attached |
| two responders are distinguishable across repeated probes, one responder yields one token | G1 to G5 |

Each refusal sits beside a positive control on the same credential in the same run (D1, D4, D4c, D10), because a broken fixture is refused everything too.

21 mutants, 21 killed at `7ae2134b`, each red naming its predicted cell: a timeout graded as unbound (C4), a no-verdict reply promoted to bound (C9), a not-ready lease graded bound (B2b), a throwing state reader answering bound (C1), a wrong-plane reply adopted (D11), an unrecognised responder value accepted (D12), a responder answering any reply target (D4b), a responder reporting a plane it was not asked about (B1), the probe subject no longer pinning the caller (D2), the agent granted the serve filter (D3), the supervisor reply grant widened off `.reply.` (E3), the supervisor granted the delivery serve filter (E2), a plane outside the closed set answered (D9), the answer carrying the lease holder (D8). Baseline 37 marks, each mutant drops to 32 to 36, so each kills a specific cell rather than the suite.

Two survivors found during development and closed in the second commit: an absent-lease read that never entered the not-ready branch (now B2b), and a bound-reply guard whose rejecting branch never ran because every cell sent a well-formed reply target. The first fix for the second was itself non-discriminating (an attacker cannot subscribe a lane it does not own, so it heard nothing either way); the third commit replaces it with a second agent listening on its own reply subtree and a control that fires a frame into that lane.

Also run at this head: the six suites that consume the moved classifier (`delivery-responder-visibility` 26/26, `delivery-boot-honesty` 15/15, `component-health`, `join-denial-guidance` 8/8, `doctor-superseded` 14/14, `endpoint-warning-consumers`), all 135 non-live core smokes, `suite-ambient-env`, `mutation-fixtures`, `ci-fragments`, `gate-inventory` (600 suites), docs voice, and the docs bundle regeneration, which is a no-op. Shard stability is stable on the merged tree against main.

## Second head

Three findings from the first review, each graded at this head.

The reply guard's rejection notice moves to the endpoint's non-fatal warning channel. It was emitted on the `error` channel, and an `error` emitted with no listener attached ends the process, so a peer whose probe named a reply target outside its own subtree ended any embedder that had not attached one, and the plane was then reported unbound. The guard still drops the frame and keeps serving. Graded by F1 to F5 in a child process with nothing attached on `error`, because survival cannot be graded from inside the process making the claim; F4 and F5 send the same child a well-formed probe so F1 is not a child that does nothing.

The remote-manager credential now carries the manager plane's serve subscription and the bounded reply row, pinned to the supervisor actor that runs `start()`. Without them the bind was denied and a peer received the broker's no-responders answer, graded `unbound`, for a manager that was running. E6 to E9, including that the executor actor carries neither row.

The changeset lists `@cotal-ai/manager` and `@cotal-ai/delivery` alongside core and cli. The `standaloneConnectOpts` pin in `required-arg-seam` moves from 156/116 to 158/118 in this commit, for the two connections the new cells open under `smoke/`.

## Follow-up

This PR ships the rail, the grants, the two responders and the peer-side probe. The operator surfaces that consume it are a separate change: `cotal doctor` and `cotal status` asking both planes and printing one line each, and `cotal join` probing on a refusal so the refusal names the dead plane as a measured fact rather than a plausible cause. `livenessLine` and `livenessConsequence` in core are the text those surfaces will print, so the wording is graded here and the consumers only route it.

## Not verified

- The full `pnpm check` chain and `pnpm smoke:ci` as composed commands. The gates were run individually.
- The CLI, delivery and manager smoke directories as a full sweep. A sweep exceeded a 600 s budget and produced no per-suite data; the six consumer suites above were run instead, and the rest are unknown, not assumed green.
- Every `:live` suite.
- The manager's real responder end to end. The manager-plane cells run a fixture responder standing in for `serviceServe`; the delivery-plane cells run the shipped responder. The one-line manager call typechecks and sits after the lease and before registration, but no real manager was booted and probed.
- The `stale` state at probe level through a real foreign holder. The classifier path is covered by the #1594 suite.
- A split across instances from a single probe. Two responders are stood up in G1 to G4 and the split is read across repeated probes; one probe reports one responder and says nothing about the others.


davidfarah2003:
Panel complete: three independent reviews of `bd7edf4f70d033a6dec1e7b34c5ee758d1587222`.

All three reviews request changes, so this is recorded as **not green**. Each reproduced the suite baseline and the full mutation proof on its own clone. Two findings recur across the reviews: the reply guard in the new responder emits `error` on an emitter that installs no listener, so an embedder that never attached one exits on a probe with an out-of-subtree reply target and the silence is then graded `unbound`; and queue-grouped responders across coexisting manager instances answer one at a time, so disagreeing instances produce a random verdict with no field that says others exist. The reviews differ on which of these is blocking. One review also found that the remote-manager credential lacks the liveness serve and reply grants, so a remote manager's peers are told `unbound` while it is bound.

---

**Blocking (1):**
- A hostile probe kills any embedder that did not attach an `error` listener, and the resulting silence is then graded as `unbound`, a positive health claim. The sender-bound reply guard rejects a probe whose reply target is outside the sender's own `.reply.` subtree by calling `this.emit("error", ...)`. `CotalEndpoint extends EventEmitter` with no `captureRejections`, no default `error` listener registered in the constructor, and no other mitigation. Node's EventEmitter throws when `error` is emitted with no [...]

**Body defects (3):**
- BD1. This understates a known limitation as an untested detail. Exactly one responder DOES answer; the unstated consequence is that when instances disagree the peer gets one of them at random with no field disclosing that others exist. My measurement at this head: 10/10 split across 20 probes with two disagreeing responders, against a unanimous 20/20 single-responder control. A maintainer publishing this list would tell readers a check was skipped, when the substantive fact is a scoped behaviour worth naming. [...]
- BD2. Not a defect in the figure, which I confirm, but a caveat the body does not carry and a maintainer may want: the '37 cells' and '14 mutants, 14 killed' figures are both accurate AT THIS HEAD, verified by me at bd7edf4f. The body's own commit history shows the earlier commit message claiming '34 cells ... 13 mutants'. Since the body is prose that outlives the head, consider pinning the figures to the sha that produced them. I am flagging it because the packet's rule is never to carry a figure across a sha that [...]
- BD3. Recorded so the absence is a checked fact rather than an unexamined silence.

**Non-blocking (5):**
- `livenessLine` and `livenessConsequence` ship with no test at all, while the body says their wording is graded here Evidence: `grep -rn 'livenessLine|livenessConsequence' --include=*.ts` over the repo returns only the definitions in src and the generated dist declarations. No smoke cell, no consumer, no reference anywhere in the 37-cell suite.
- `stale` is reachable in the wire contract but is exercised at no point in the suite. The manager responder can only ever return `bound` or `unbound` (it grades `serviceServe?.handle`), so `stale` on that plane is unreachable by construction today. Only the delivery responder can produce it. Evidence: `grep -n 'stale' packages/core/smoke/liveness-peer.smoke.ts` returns nothing (exit 1). `responderFromLease` can return `stale`, `parseLivenessAnswer` accepts it on the wire, and `livenessLine` renders a distinct [...]
- The manager's real responder is never booted and probed end to end Evidence: The manager-plane cells use a fixture responder standing in for `serviceServe`. The one shipped line typechecks (full `pnpm typecheck` passes at this head) and sits after lease acquisition and before registration as its comment claims, which I verified by reading the surrounding code.
- The changeset's package selection looks right but `pnpm changeset status` could not confirm the plan in my lane Evidence: `.changeset/peer-readable-liveness.md` declares `@cotal-ai/core` minor and `@cotal-ai/cli` patch. `pnpm changeset status` fails in my clone with 'Failed to find where HEAD diverged from "main"' because my lane has no `main` ref; the origin remote is a local path clone. This is an environment limitation in my lane, NOT a defect in the change.
- One design claim is stronger than the code strictly guarantees, though harmlessly. The docblock says the plane rides the subject so 'a credential can be granted "may ask about delivery" and nothing else, enforced by the broker'. That is true of the subject grammar and is a good reason for the design, but `permissionsFor` currently mints BOTH planes for every agent (`...LIVENESS_PLANES.map(...)`), so no shipped credential is actually scoped to one plane. The capability is real; the deployment of it is uniform.

**Not verified (9):**
- The manager's real liveness responder end to end. No real manager process was booted and probed; the manager-plane cells use a fixture responder. Whether `serviceServe?.handle` is falsy in EVERY state in which the manager is not serving is therefore unverified, and if it can remain truthy after the instance stands down, the manager plane could answer `bound` over a dead manager. I state this as an open risk, not as a finding, because I did not test it.
- Whether the B1 crash is reachable on the MANAGER plane specifically. I reproduced it on the delivery plane, where I could mint a `delivery` cred and bind the shipped responder directly. The crashing branch is in the shared `serveLiveness` and is plane-independent by inspection, and the manager binds the same method at manager.ts:1421, so I am confident it generalises, but I did not execute it against a real manager.
- `stale` end to end on the wire through a real foreign lease holder. Not exercised by this suite (grep for 'stale' in the smoke returns nothing); the body says the #1594 suite covers the classifier path and I did not run that suite.
- The full `pnpm check` chain and `pnpm smoke:ci` as composed commands. `smoke:ci` needs a broker and a long budget; I ran the targeted suite, the full mutation proof and a full typecheck instead. The body itself discloses these as not verified.
- The six consumer suites the body reports (delivery-responder-visibility 26/26, delivery-boot-honesty 15/15, component-health, join-denial-guidance 8/8, doctor-superseded 14/14, endpoint-warning-consumers) and the '135 non-live core smokes'. I did not re-run them, so those figures are the writer's claims and I neither confirm nor dispute them.
- `pnpm changeset status`. It cannot run in my lane: no `main` ref (origin is a local path clone), so the tool errors on 'Failed to find where HEAD diverged from main'. This is a lane limitation, not a defect in the change. A tool that cannot run is a refusal, not a pass and not a zero.
- The three CI checks still pending at my read time (windows-ok, changed, ci-ok). I did not and will not re-run them.
- Long-run or concurrency behaviour of the probe under load, reconnect, or credential renewal. Not exercised.
- Whether any OTHER pre-existing `emit("error")` site in endpoint.ts is peer-triggerable. I counted 31 at the base commit and confirmed the idiom predates this PR, but I audited only the liveness path's reachability. Out of scope for this head, and possibly worth its own look.

---

**Blocking (1):**
- Queue group causes random and nondeterministic answers when multiple manager instances coexist. serveLiveness subscribes to `livenessServeFilter(this.space, plane)` with `{ queue: 'live.' + plane }`. When multiple manager instances run in a space (which is valid and supported in Cotal via distinct instance IDs), requests are load-balanced across instances. If one instance is bound and another is unbound/degraded, successive probes return contradictory results at random without indicating the existence of multiple [...]

**Body defects (1):**
- PR body references unmerged commit history and out-of-date commit structure. The PR body under 'How it is graded' (fetched at 2026-09-15T11:44:17Z) states: 'Two survivors found during development and closed in the second commit... the third commit replaces it with a second agent listening on its own reply subtree'. On head bd7edf4f70d033a6dec1e7b34c5ee758d1587222, the PR consists of four commits (971fbd926, f6666d8d1, fead21465, bd7edf4f7), where bd7edf4f is a fourth commit that was not reflected in the narrative.

**Non-blocking (4):**
- Uncaught 'error' event risk in serveLiveness for third-party embedders. serveLiveness emits an 'error' event on CotalEndpoint when a sender specifies an out-of-subtree reply target. If an embedder using CotalEndpoint does not register an 'error' event listener before starting the endpoint, Node.js throws an unhandled error, crashing the process on receipt of a malformed probe.
- Wire protocol and subjects undocumented in SPEC.md and docs/. The new `live.<plane>.<owner>.<actor>` subject tier and liveness protocol are not documented in SPEC.md or docs/. SPEC.md is the normative wire standard; omitting normative wire subjects prevents independent conformant implementations.
- Unexercised production CLI consumers. probeLiveness, livenessLine, and livenessConsequence have no callers in implementations/cli (such as cotal doctor or cotal status); the surface is currently only accessible via the programmatic TypeScript API.
- D4b mutation fixture description text drift. In packages/core/smoke/fixtures/liveness-peer.mutations.json, the cell description for D4b reads 'CELL D4b: a probe whose reply target is ANOTHER peer's lane is NOT answered (no confused deputy)', whereas in liveness-peer.smoke.ts the assertion label reads 'delivers NOTHING into that peer's lane'. While expectRed ('CELL D4b') matches and kills the mutant, the text label is slightly drifted.

**Not verified (4):**
- Full pnpm smoke:ci chain and composed pnpm check commands across the entire repository (individual targeted smoke suites, typecheck, build, and mutation proofs were run).
- Live end-to-end execution of a fully booted production space manager servicing real peer liveness probes (tested via unit/smoke test harness and serviceServe mocks).
- Real foreign holder transition to 'stale' state driven over the wire at the probe level.
- Live-cluster multi-broker tests (:live suites).

---

### Review 3: `REQUEST_CHANGES`

**Blocking (2):**
- Remote-manager deployments publish a false `unbound` manager verdict. Manager.start() unconditionally calls serveLiveness("manager", ...), but the remote-manager credential has neither the manager liveness serve subscription nor the bounded liveness reply grant. On a real auth broker, the subscription was denied and an ordinary agent probe returned {plane:"manager", responder:"unbound"} while the remote manager endpoint was running and its readState returned bound. Evidence: The reproduction emitted permission [...]
- Disagreeing manager responders return contradictory random verdicts. All manager instances join the same live.manager queue group, but each reply contains only its own bound/unbound state. A single request therefore samples one arbitrary instance and provides no indication that other instances exist or disagree. Evidence: On a real broker using the shipped CotalEndpoint serveLiveness and probeLiveness functions, two responders returning bound and unbound produced 23 bound and 17 unbound answers across 40 identical [...]

**Body defects (1):**
- The body says a peer learns whether the manager has a bound responder and that several instances are queue-grouped so they answer once. It does not disclose that disagreeing instances yield a random contradictory verdict, and the broad claim is false for the supported remote-manager composition, whose responder subscription is denied and whose peers receive unbound. Update the body after the code is corrected, or state the actual limitations if any remain.

**Non-blocking (3):**
- serveLiveness is not safe by construction without an EventEmitter error listener. The sender-bound reply guard emits the special EventEmitter `error` event. CotalEndpoint installs no default listener. An embedder using the exported serveLiveness API without attaching its own listener exits when an authenticated probe supplies an out-of-subtree reply target. Evidence: A real-broker positive control with an error listener logged the rejection and survived. The same exported API with no listener exited 1 with Node's [...]
- The new wire rail is absent from the normative protocol documentation. The change adds cotal.<space>.live.<plane>.<owner>.<actor>, reply grammar, authorization semantics, and reply schema, but SPEC.md and docs contain no live.<plane> entry. A controlled git grep found established chat wire subjects in SPEC.md and docs but no liveness rail.
- The changeset omits changed manager and delivery packages from its frontmatter. The PR changes first-party shipped behavior in @cotal-ai/manager and @cotal-ai/delivery, while .changeset/peer-readable-liveness.md names only @cotal-ai/core and @cotal-ai/cli. The fixed group makes `pnpm changeset status --since 08cff353...` include the lockstep packages, but the repository rule still requires every package with a first-party change to be listed.

**Not verified (6):**
- Hosted CI state was not used and no workflow was re-run. The provider cancellations are not a verdict on this head.
- The full composed pnpm smoke:ci and full pnpm check chains were not run.
- The full CLI, manager, delivery, and every :live smoke sweep were not run.
- A complete local Manager process and complete delivery daemon were not booted together and probed end to end. The remote-manager authorization defect was reproduced with the real minted profile, real CotalEndpoint APIs, and a real auth broker.
- The stale reply was not exercised through a real foreign-holder daemon at probe level.
- No GitHub write, CI re-run, push, comment, review, label, or body edit was performed.

davidfarah2003:
Panel complete: three independent reviews of `21320292ec27a5df4d504352a7578cd72b9cdbbb`.

Second panel on this branch (the first, on bd7edf4f7, returned three REQUEST_CHANGES). This head is the empty tree-identical commit over 7ae2134b5 (tree `89d2a11785aff2524a0264ab6111429ce5b76093`). Two reviews return APPROVE and one returns REQUEST_CHANGES. All three measured the five fixes named in the commit message as holding at this head on their own clones: the reply-guard rejection notice moved to the recoverable channel, the per-bind responder token on the liveness reply, the supervisor-pinned serve filter and reply grant on the remote-manager credential profile, the peer-liveness suite at 58 of 58 across repeated runs, and the full mutation fixture caught on its named cells. The dissent and two of the approving reviews' non-blocking items land on the same point: the new liveness wire subject family and the reply schema, including the responder token this change adds, are absent from the normative SPEC.md and the generated schema. One review also reports a stale exact-field sentence in the changeset and one reports a misstatement of the mutation bookkeeping in the commit message. Merge basis: not met on the normative document gap, carried by one blocking and two non-blocking findings; the revision that adds the wire subject and reply schema to SPEC.md and corrects the changeset sentence goes back to the requesting review for re-grade.

---

### Review 1: `APPROVE`

All five items in the commit message do what they say at this head, verified by my own observation and not by reading the writer's report. I reproduced the writer's central claim for item 1 independently three times: the change that puts the rejection notice back on the `error` channel is caught, and it is caught on the cell the fixture names. The full 21-entry mutation fixture was driven to completion three clean times with every entry caught on its own named cell. The peer-liveness suite ran eleven times, 58/58 every time, with no flake. I went past the two writer-stated limitations rather [...]

**Blocking (0):** none

- BD1. The commit message misstates the mutation bookkeeping in both halves of one sentence.

**Non-blocking (4):**
- NB1. The new wire subject family and the new reply field are absent from SPEC.md and the generated schema. The branch adds a peer-visible request/reply exchange on subjects live.<plane>.<owner>.<actor> with a .reply. leaf, built by livenessSubject, livenessServeFilter and livenessReplyGrant in packages/core/src/subjects.ts. The commit under review adds a second field, instance, to the reply body. The repository's own guidance treats SPEC.md as the normative wire contract and asks that a wire change update it in the same change.
- NB2. The exported operator-facing renderer has no caller and does not render the field this commit adds. packages/core/src/liveness.ts exports livenessLine, which turns a LivenessAnswer into an operator-readable line, and packages/core/src/index.ts re-exports the module, so it is public API. Its switch covers the four responder states and never mentions instance.
- NB3. The peer-liveness suite reports a broker leak that did not occur. The suite's teardown path checks whether the spawned broker process is still present before cleaning its directory, and appears to do so before the process has been reaped, so it prints a warning on a normal successful exit.
- NB4. The split-detection property offers a caller no bound and no API affordance for the repetition it requires. probeLiveness returns a single answer. Detecting a responder split requires the caller to loop, compare instance tokens across answers, and choose a loop count, with no guidance from the type, the docs or the changeset, and no guarantee from the queue group about distribution.

**Not verified (13):**
- I did NOT run pnpm smoke:ci, the full 152-cell shard-0 CI gate. The brief states it halts at a cell that spawns docker, and neither docker nor podman is installed on this host. I treat that as an instrument refusal and every cell after it as unverified rather than passed. I therefore cannot speak to the shard as a whole.
- I did NOT run artifact-store or smoke:delivery-starvation. The brief attributes their failure here to insufficient storage on a root disk at 96 percent. I did not attempt them and did not run the pristine-base comparison that would confirm the attribution, so I am relying on the brief for that attribution and not confirming it myself.
- I did NOT boot a remote manager through the typed authority protocol end to end. My live harness mints the remote-manager credential directly via mintCreds and calls serveLiveness on an endpoint, which exercises the grant and the responder over a real broker but not the typed supervise/issuance protocol that would issue that credential in production. So the grant and the responder are proven live; the issuance path that hands them out is not.
- I did NOT run the full Manager.start() path against a real broker, so I verified by reading rather than by execution that line 1421's serveLiveness call is reached unconditionally in a real manager boot.
- I did NOT test behaviour across a broker reconnect, a responder restart mid-probe, or a multi-server or clustered broker. My split measurement is against a single local nats-server v2.14.0, so the queue distribution I measured may not characterise other deployments.
- I did NOT test the delivery plane's responder under a real delivery daemon; the delivery-plane cells I exercised are the suite's own.
- I did NOT verify the PR body on GitHub. This is an in-house head and I made no GitHub read or write of any kind, so first-panel findings about PR body text could not be re-checked at their source. I graded the commit message and the changeset, which are in the packet and in the tree.
- I did NOT independently confirm the writer's or the first panel's measurements at their own shas. Where I cite a reference file I say so and give my own observation at this head alongside it.
- I did NOT run the other two lanes' work or look at it, per the independence requirement.
- Invocation error, not a red. It printed its own reason and named the missing variables. I re-ran it correctly and it passed 10/10 three times. pnpm smoke:ci-suites-freeze without BASE and HEAD.
- Not run. The brief states these fail on this host for insufficient storage (root disk at 96 percent) and that smoke:ci halts at a cell requiring docker or podman, neither of which is installed here. A status-null probe for an absent tool is an instrument refusal, not a red, and every cell after it in that shard is unverified rather than passed. I did not run them and I do not count them either way. smoke:ci, artifact-store, smoke:delivery-starvation.
- My harness killed it, not the suite. It left a mutant in the tree, which I restored with git and then re-ran the pass in the foreground to completion. My instrument fault, not a defect in the head, and not counted as a suite failure. My own background mutation pass terminated at exit 143.
- Correct behaviour by the head, not failures. A 16-character instance id was refused as not a valid lifecycle token, and an agent credential minted without a lifecycleUid was refused citing SPEC 13.1. I fixed my harness. Recording them because they are incidental evidence that the minting path really validates its inputs. Two refusals from the head's own validation while I was building my live harness.

---

The five requested fixes behave as stated, the new cells are stable across repeated green and mutation runs, and the two writer-stated limitations are accurately scoped. I request changes because the new liveness subject and reply schema, including the responder token added here, remain absent from Cotal’s normative SPEC.md and protocol docs. The changeset also contains one stale exact-field sentence.

**Blocking (1):**
- B1. SPEC.md; docs/; packages/core/src/subjects.ts; packages/core/src/liveness.ts; packages/core/src/endpoint.ts. The peer-readable liveness wire protocol is missing from the normative specification. The branch introduces cotal.<space>.live.<plane>.<owner>.<actor>, bounded reply subtrees, a LivenessAnswer wire payload including the new opaque instance token, queue-group response semantics, and credential grants. SPEC.md and docs contain none of this. Path existence was checked first; a liveness-wire grep returned exit 1, while a chat-subject positive control returned matches with exit 0. Cotal is a wire standard and SPEC.md is normative. Leaving the protocol defined only by the reference TypeScript implementation prevents an independent conforming client from implementing or validating this surface and violates the repository rule to update SPEC.md for wire changes in the same change.

- P1. The release note says the reply carries exactly the plane and one responder verdict, but the same change now sends a third instance token. The later paragraph explains the token correctly, so the earlier exact-field sentence is stale and internally inconsistent. .changeset/peer-readable-liveness.md:27.

**Non-blocking (0):** none

**Not verified (9):**
- A fully booted remote manager using authority-issued supervisor credentials answering a live peer liveness probe. The grant rows and production call path were independently traced and mutation-tested, but this end-to-end composition was not run.
- Live multi-broker and unrelated acceptance surfaces not otherwise named in this record.
- The same 456 suites after smoke:the harness were unverified in each of three smoke:ci runs. The shard explicitly marked them NEVER RAN every time.
- pnpm smoke:delivery-injected-tenancy-admission. All three invocations refused before suite execution with ERR_MODULE_NOT_FOUND for implementations/delivery/node_modules/@cotal-ai/core/dist/index.js because the lane had not yet been built. Not counted as suite reds.
- pnpm smoke:delivery-starvation. All three invocations refused before suite execution with the same missing generated @cotal-ai/core/dist/index.js. Not counted as suite reds.
- pnpm smoke:manager-service-authority. Invocation refused before cells with ERR_MODULE_NOT_FOUND for implementations/auth/dist/index.js because build artifacts were absent.
- command -v docker / podman. Both commands exited 1 with no output. The tools are absent; this is an instrument refusal, not a failed container probe.
- third mutation-proof loop under outer harness timeout. The outer 600-second task timeout interrupted the third loop after two complete successful runs. Temporary mutations were restored byte-for-byte from HEAD and the lane was verified clean before a replacement third run.
- pnpm smoke:ci. Every run built successfully and started 144 suites, then smoke:the harness asserted over a child status:null because neither docker nor podman is installed. Per the packet this is an instrument refusal, not a head failure. Each run explicitly marked the same 456 later suites NEVER RAN.

---

### Review 3: `APPROVE`

All five claimed items in commit 7ae2134b5 (with empty tree-identical commit 21320292e) behave exactly as claimed and hold under independent observation. The reply guard rejection notice is moved to emitRecoverable (the warning channel), preventing unhandled EventEmitter crashes for embedders without error listeners (verified via dedicated child process in CELL F1-F5). The manager liveness reply carries an opaque per-bind responder token (randomUUID) disambiguating disagreeing coexisting manager instances across repeated probes (CELL G1-G8). The remote-manager credential profile carries the [...]

**Blocking (0):** none

**Non-blocking (1):**
- NB-1. SPEC.md / docs wire subject documentation. While core implementation and permissions are fully defined and tested, normative documentation in SPEC.md for the live.<plane> subject hierarchy remains a candidate for docs enrichment in follow-up work.

**Not verified (5):**
- End-to-end execution of a full booted remote manager daemon responding to peer liveness over typed authority protocol (disclosed limitation).
- Multi-host physical cluster networking across WAN (tested against local multi-client broker instances).
- Full composed 600-suite pnpm smoke:ci chain (stopped by known host docker tool absence on the harness cell; targeted shards and suites verified).
- pnpm smoke:ci-suites-freeze without BASE/HEAD. Requires BASE and HEAD environment variables; prints explanation when unset. With BASE and HEAD provided, runs 10 passed, 0 failed, exit 0.
- pnpm smoke:delivery-injected-tenancy-admission with TMPDIR inside lane. Intentionally tests workstation root matching; fails as a positive control when TMPDIR is inside lane cwd, passes 8/8 when TMPDIR is outside (/tmp).

PR #1636, opened 2026-09-15
fix(manager): many managers per space, on the same device or different ones

davidfarah2003:
Refs #773.

A space has many managers, on the same device or different ones, and a seat can be spawned on any of them. That is the core proposition. Since the #773 fix landed (b00f3c1d2) a manager refused to start unless its SecretStore was the one the delivery daemon reloads from, which on an auth-mode space meant one manager per space, on the daemon's host, from the daemon's root. Measured at 0.49.0 on a second machine: `cotal supervise --space <s>` refuses at construction naming both roots.

## What changed

The manager classifies the daemon's reload store instead of refusing it.

- **shared**: this manager's store is the daemon's. It is the daemon-cred renewal owner and remints as before, with fingerprint-only adoption. Unchanged.
- **foreign**: the daemon reloads from a store this manager does not write. The manager starts, serves seats, never remints daemon creds (a generation written where the daemon cannot read it was the original #773 defect), and writes a renewal record saying renewal is owned elsewhere, naming both stores. The note is logged once per distinct pair, not every pass.
- **absent**: no delivery daemon holds this space's lease. The manager determines that itself by reading the delivery lease row, a KV key whose only writer is the delivery credential, under a read-only grant of its own; it does not read it from the rail's outcome, because a NATS client builds its typed no-responder error from the reply message itself (an empty payload with a 503 status header) and a responder that does not hold the lease can send those bytes. A rail that yields nothing while a lease is live is refused and the refusal names the lease; a lease row that cannot be read is undetermined. Neither remints. When the row is absent the pass proceeds writing this store, and a later daemon on a foreign store moves the manager to foreign on the next pass (see the v3 note below).
- A hung, denied, or unreadable challenge skips the daemon remint and records the reason. It never remints on an unproven relation. Every other duty of the pass still runs.

A foreign classification skips the daemon remint and nothing else. The credentials a manager owns outright (its managed-agent statics, its endpoint-serve credential, its goal-writer, its hosted runs, its session ledger) are renewed by no other process, and this pass is their only driver, so a foreign manager renews them as a shared one does.

The store answer is bound to the process that reloads. The delivery-admin rail is queue-grouped, so any holder of a delivery credential can answer the challenge while only the lease holder reloads the standing creds. `reloadStoreIdentity` answers with the store identity, the responder's identity, and its own lease claim. The manager does not take that claim on trust: it reads the lease row's recorded holder itself and requires the answerer to be that principal, keeping the reply's claim as a cross-check that must agree. An honest non-holder is refused by its own admission; a responder that asserts the lease it does not hold is refused by the row. Managers (the supervisor actor and the remote-manager profile) gain a read-only keyed grant on the delivery bucket for those two reads, with no publish row, so the credential that verifies the row can never write it. `parseDaemonStoreAnswer` refuses an unknown top-level key and names it, matching the closed parser it documented.

`cotal doctor auth` renders the owner-elsewhere record as information, not as a problem. `RenewalRecord` gains an optional `renewalOwner` field. `divergentSecretStoreRefusal` stays exported for the delivery daemon's own `--creds`-vs-cwd check, which is unchanged.

## What is proved

`bin/smoke/manager-two-root-renewal.smoke.ts` is re-inverted. Phase 1 (manager rooted at A, daemon reloading from B, same space, real broker) requires that the manager starts, that its renewal record says owned elsewhere and names both roots, that results are empty and no adoption was requested, and that both roots' delivery and membership creds still hold the original generation. The control phase over a unified root still requires broker-accepted adoption of both components.

The suite has 54 cells at this head. The 21 added in v2 and the 12 added in v3 grade six guards, each with a positive control in the same run, and each was proved by disabling the guard in the built artifact the suite loads and confirming the named cells red:

- a foreign manager still reaches the system-cred expiry warning and the managed-static renewal scan, and still remints nothing into the foreign store (restoring the early return reds 2 cells);
- a pass whose classification could not reach a verdict still reaches the system-cred expiry warning and the managed-static renewal scan, remints nothing, and leaves both roots at the original generation (four (a2) cells; restoring the early return in the catch reds them, which is the mutation that survived at the v2 head);
- a peer body quoting `no responders` on the rail is never classified as absence, and a peer reply of empty payload plus a 503 status header, the bytes the client turns into its typed no-responder error, is never classified as absence either; each responder is required to win the queue at least once so the sweep measured something, and the refusal names the live lease (taking absence from the rail outcome again reds the (b1) cells);
- a responder that replies with a raw lease claim of true while another process holds the lease is refused, and the refusal names the answerer and the recorded holder (dropping the holder comparison reds the (b2) cells);
- a lease-less responder naming this manager's own store is refused, and the refusal names the lease (dropping the lease binding reds 2 cells);
- a different generation of the same daemon identity presented as `expected` is refused as a generation mismatch, with no digest in the refusal (disabling the expected-generation guard reds 3 cells; this is the mutant that survived at the previous head).

`packages/core/smoke/secret-store-identity.smoke.ts` has 29 cells, eleven of them on the closed parser for the daemon's answer including extra-key refusals with a positive control. `packages/core/smoke/delivery-lease-grant.smoke.ts` (24 cells) covers the manager's new lease read and its confinement to reads. The four manager smoke fixtures that stood in for the daemon now answer with the bound shape, and the boot helpers used by four other suites acquire the lease they claim, which was found by grepping each run for a skipped remint rather than trusting exit status. `smoke:suite-ambient-env`, `smoke:mutation-fixtures`, `smoke:ci-fragments`, `smoke:gate-inventory`, `smoke:shard-stability`, `check:docs-voice`, the docs bundle check, and `typecheck` pass at this head.

## What is not here

The lease row is read for shard 0, named as a constant with its reason, which matches every current caller and the single-shard delivery layout this repository ships; a multi-shard layout must carry the shard in the challenge. The holder comparison is an equality on the wire principal recorded in the lease row: it proves the answerer presents the holder's principal, and the broker's authentication of that principal is what makes the presentation mean anything. A responder that wins the queue can still deny a determination by answering, which makes the pass refuse and remint nothing; that is the fail-closed direction and is unchanged. Renewal ownership is still "the manager rooted where the daemon reads". A lease-elected owner that carries re-signed bytes over the delivery-admin rail, so the daemon's host need not run a manager at all, is the next step and has its own design record. `cotal up` on a participant machine for a hand-registered remote space still refuses; `cotal supervise --space <s>` is the command that starts a manager there today.

## v2 note

Head `bea07f879d10de54ffe9f0818b0e9cf4f537515c` replaces `ecbee4b1bcceb731e246f817052f241b2609fea4`. The first head had one defect reached by four instruments: the foreign early return sat above every own-credential duty, so a foreign manager renewed nothing it owned and died at its own credentials' expiry, invisible for a day on a quarter-TTL tick. It also decided absence by matching `no responders` or `503` over an error string, which a parse failure on a peer body quoting those words satisfied, and it accepted the first reply on a queue-grouped rail without binding it to the lease holder. The divergent-generation refusal the first body claimed was not asserted by any cell. All four are fixed or asserted at this head as described above, and the changeset and the embedding page state that every manager of one space must serve identical contracts.

## v3 note

Head `29fa7737783d4fc62f84afbd9a2e94c573a3627a` replaces `bea07f879d10de54ffe9f0818b0e9cf4f537515c`. The v2 head decided absence from the rail's typed no-responder error, which two reviews showed the client synthesises from the reply's own bytes, so a responder could still produce a determination of absence and reach the branch that remints. It also checked that the daemon's lease claim was a boolean rather than that it was true of the answerer. Both are closed above by reading the delivery lease row, a fact no reply on any rail participates in, on both the unanswered and the answered path. The changeset, `SPEC.md` around the `reloadStoreIdentity` paragraph, and the embedding page no longer say a divergent manager is refused before the remint, and they no longer describe the typed error as the broker's own statement; each now gives the three-field reply shape and states where absence and the holder come from.

One test defect was found in review at this head's predecessor and fixed here: the (b1) header arm kept only the first three refusal reasons of its sweep while the earlier text-body responder was still bound to the same queue group, so its message assertion measured capture order and failed on roughly one run in three with the guard holding. The earlier responder now leaves the queue group before the next arm binds, every refusal reason across each sweep is kept, and the message-shape cells speak only to what a refusal said. The suite ran green sixteen times in a row in the lane and six times in a row on a second machine after the fix.

Relation to #1642, which reaches the same product decision: that head keeps the string predicate on the error message, which a parse failure on a peer body quoting `no responders` satisfies, and carries no lease binding on the answered path, so rebasing onto it would reopen both findings above. This branch keeps everything #1642 is for (a foreign manager starts, serves seats, remints no daemon credential, records that renewal is owned elsewhere). Two smaller differences, its renewal-record choice on a non-owner and its realpath resolution of suite roots, are reconcilable at merge time.


davidfarah2003:
Panel complete: three independent reviews of `bea07f879d10de54ffe9f0818b0e9cf4f537515c`.

Two of three reviews request changes, so this is recorded as **not green**. All three reproduced the 42-cell baseline on their own clone and killed the four v2 guards by disabling them in the built artifact with same-run positive controls, so B3 and the divergent-generation refusal are held to be fixed and graded. The two requesting reviews independently reproduced the same residual defect on the suite's own live rig: the typed no-responder signal the new absence predicate accepts is constructed by the NATS client from the reply message itself (empty payload plus a 503 status header), so a responder bound to the queue-grouped delivery-admin rail can still produce a determination of absence and the foreign manager then remints into its own store. One of those reviews also found that the lease-holder binding trusts a responder-controlled boolean, so a non-owner asserting it can still be classified as shared. Both flag that the body, the changeset and the source comment state the absence guarantee more strongly than the predicate tests.

---

**Blocking (1):**
- B1 is fixed in instance, not in kind: the typed no-responder signal is synthesized client-side from the reply, so a bound responder can still produce a determination of absence, and the absent branch remints. The delivery-admin rail is queue-grouped and `requestDeliveryAdmin` issues a `noMux` request with an explicit reply subject. The NATS client constructs `NoRespondersError` when the message that arrives on that reply subject has an empty payload and a 503 status header (nats.js:325-326: `if (msg && msg.data?.length === 0 && msg.headers?.code === 503)`), then wraps it in a `RequestError` whose `isNoResponders()` is true. That condition is [...]

**Body defects (2):**
- The body, the changeset, and the source comment each assert that absence now comes only from the broker itself, which is not what the predicate tests. The body says `the broker itself reported no bound delivery-admin responder, as a typed no-responder signal` and `Absence is a determination; text on an error body is not one`. The changeset says `Absence now comes only from the broker's own typed no-responder signal`. The source comment says `Only NoRespondersError (or a RequestError whose isNoResponders() is true) is the broker's own statement that nothing was bound`. In this client the typed error is constructed locally from the reply [...]
- The body's proof claims for the four v2 guards are accurate as far as they go, and I confirmed them independently. Recorded here as a positive finding rather than a defect, because a re-review should say plainly what held. The body claims 42 cells, four guards each with a same-run positive control, each proved by disabling the guard in the built artifact. My own runs reproduce 42/0 and kill all four named mutants at the artifact the suite loads. The one inaccuracy in the surrounding prose is the absence claim recorded above.

**Non-blocking (4):**
- The (b1) cell sweeps a single reply shape, so a passing cell does not establish the property its name asserts. The cell is named `a peer body quoting "no responders" is NEVER classified as absence`, which is precise about its input, but the guard it stands for is the class `a peer cannot manufacture a determination of absence`. The rogue responder replies only with a raw body. Adding a second arm on the same rogue (the empty-payload plus 503 status header shape) is a small change to an existing loop and is what turned an untested class into a reproduced defect for me in about a minute of rig time.
- On a shared classification, a pass writes the renewal record twice and the first write is not observable. When `classifyDaemonSecretStore` throws, the catch writes a record carrying `adoption:{ok:false,error}`. `relation` is then undefined, so neither the foreign branch nor the remint branch runs and that record survives, which is the intended behaviour. On the normal shared path the remint branch writes a second record at the end of the same pass. That is fine. What is worth noting is only that the failure record and the success record share one file with no generation marker, so an operator reading `renewal.json` cannot distinguish `this [...]
- The lease claim in the daemon's answer reads shard 0 unconditionally. `readDeliveryLeaseEntry(0)` is hardcoded. This is correct for what ships: `implementations/delivery/src/delivery.ts:326-329` hard-rejects `shards !== 1 || shard !== 0`, so shard 0 is the only reachable shard today. The PR body discloses this limitation explicitly in `What is not here`. Recorded as non-blocking and already-disclosed rather than as a finding.
- A lease read that throws is answered as holdsDeliveryLease:false, which the caller cannot distinguish from an honest non-holder. The catch sets `holdsDeliveryLease = false`. The comment argues this is the fail-closed direction, and for the manager's decision it is: both readings send the caller to the refusal, nothing remints. I agree with the direction. The residual is diagnostic only. A genuine daemon whose KV read is failing (an expired credential, a broker hiccup) reports itself to every manager as a non-holder, and the manager's refusal text says the answerer does not hold the lease, which names the wrong fault and would send an [...]

**Not verified (4):**
- I established the 503-header path against the broker configuration this repository's own suite stands up (`serverConfig` plus `nats-server -c`, and separately a bare unauthenticated server). I did not survey whether any nats-server configuration strips or rejects a client-published 503 status header on a reply subject, which would narrow the exposure. My claim is scoped to the shipped and tested configuration.
- I ran the nine touched suites named in my brief. I did not run `pnpm smoke:ci` in full, nor `typecheck` or `check:docs-voice`, so the body's claims about those specific gates are unverified by me rather than disputed. `smoke:delivery-starvation` is documented as failing on this host for storage reasons and I did not run it.
- I read the PR metadata and body with `gh` read-only. I did not evaluate the check rollup, so I make no claim about CI colour at this head. I performed no GitHub writes of any kind: no push, comment, review, label, or check re-run.
- another review's v1 non-blocking note about two same-store managers reminting concurrently without an election is unchanged by this head. I did not test it and take no position on it.

---

**Blocking (2):**
- A non-owning delivery-admin responder can still forge the typed absence signal and make a foreign manager write daemon credentials. The v2 predicate distinguishes parse failures from NATS no-responder errors, but it still assumes the typed error can originate only from the broker. In @nats-io/nats-core 3.4.0, request(noMux:true) turns any empty reply carrying status 503 into NoRespondersError, then RequestError.isNoResponders() returns true. A delivery credential can subscribe to the queue-grouped admin request rail and publish to every admin reply subtree. It can therefore answer with m.respond(new Uint8Array(0), { headers: headers(503, 'No [...]
- The new lease-holder binding trusts the responder's self-asserted boolean, so a non-owning responder can still force shared classification. The shipped Plane-3 handler computes holdsDeliveryLease from the live KV row, but the manager receives only JSON and checks only answer.holdsDeliveryLease. parseDaemonStoreAnswer validates that the field is boolean and the responder string is nonblank. It does not authenticate either field. The manager never compares answer.responder with a broker-authenticated publisher identity, and the request/reply transport supplies no such identity binding. The delivery credential already has read access to the [...]

**Body defects (3):**
- The body says absence comes from the broker's typed no-responder signal and that text on an error body is not a determination. The latter is true, but an ordinary responder reply with empty payload plus status 503 is transformed by the client into that same typed signal. The body therefore overstates that typed absence is broker-only.
- The body says the store answer is bound to the process that reloads and that the response carries a live lease claim. Actual state is a responder-controlled boolean plus responder-controlled string. The manager validates shape, not provenance, and a non-owner asserting true is accepted.
- The body says the 42-cell suite grades the lease binding. Its B2 cell uses the shipped handler, which honestly reports false for a non-owner. It does not test a raw non-owner reply asserting true, so it does not grade the security property stated in the body.

**Non-blocking (3):**
- The new own-credential fallthrough after a failed classification is not graded by the 42-cell suite. V2 removes the return in the classifyDaemonSecretStore catch so a hung, denied, unreadable, or lease-less answer skips only the daemon remint and still runs own duties. Restoring that return in implementations/manager/dist/manager.js produced 42/0. In the same mutation harness run, the known foreign-branch early-return control produced 40/2 on the named own-duty cells. The reader is valid, but this new failure-path behavior survives it. This is a grading gap rather than a separate observed production failure because the shipped control flow [...]
- parseDaemonStoreAnswer does not implement the closed-parser discipline claimed around the new core behavior. parseSecretStoreIdentity rejects extra keys, but parseDaemonStoreAnswer accepts unknown top-level keys. The smoke has no extra-key cell. This did not create the two blockers above, but it is a mismatch between the new parser's documentation and actual behavior.
- The normative SPEC description of reloadStoreIdentity is stale relative to the new response shape. SPEC.md around 4153-4162 describes the store identity challenge but not responder, holdsDeliveryLease, or how those fields are verified. It also still says a divergent manager is refused before remint, while this PR changes that behavior to foreign classification and startup. The changeset and PR body are updated, but the normative protocol text was not.

**Not verified (3):**
- registration-executor-resume did not complete on this host. It failed during setupSpaceStreams with a JetStream TimeoutError before reaching the touched manager scenario. I do not use it as evidence for or against the change.
- Four GitHub checks were still in progress at the final read: CI smoke shard 3/4 and Mutation reproof changed shards 0/12, 9/12, and 11/12. There were no completed failing checks. One earlier attribution run was cancelled, while a later attribution run succeeded.
- I did not run smoke:freeslot-barrier:live or smoke:int2-revoke:live. The directly relevant live suite, parser suite, boot-self-heal, lifecycle-e2e, manager-reconcile-startup, and manager-restart-live were run.

---

### Review 3: `APPROVE`

**Blocking (0):** none

**Body defects (0):** none

**Non-blocking (0):** none

**Not verified (3):**
- pnpm smoke:delivery-starvation: skipped due to known host filesystem space constraint (insufficient storage resources available for 4 GiB JetStream reservation; host disk 96% full).
- Live multi-host physical networking across separate machines (tested via independent local multi-root processes and simulated daemon rails).
- Full pnpm smoke:ci suite end-to-end (targeted touched suites, typecheck, docs-voice, and mutation proofs were executed directly).

davidfarah2003:
Panel complete: three independent reviews of `0c798486149ef7c605dbac008b8e8716a46e5a1e`.

This is the re-review of the head that answers the first panel (tree `6d8cdeb66e0e0c0d28c29d062608fb09efaca73c`; the sha moved by an empty commit to rebuild the merge ref, so the code is the same bytes as the prior head). One review approves and two request changes. All three grade the first panel's absence finding as answered: absence is now determined from a requester-side read of the delivery lease row and no longer from the rail's reply outcome, and none could forge it. All three grade the foreign-manager renewal finding as answered: the early return above the manager's own credential duties is gone, and reinstating it reddens a named cell. Both reviews that checked the divergent-generation refusal find it restored with a positive control. All three report the suites green on repeated runs, the mutation tooling killing the guards, and the docs bundle regenerating byte-identical.

The split is on one finding, the lease-holder binding. The head replaced trusting the reply's lease boolean with reading the lease row locally and requiring the reply's responder field to equal the recorded holder. The approving review grades that as answered, since a distinct-principal responder is refused by name and dropping the comparison is mutation-killed. The two requesting reviews grade it as not answered or partly answered for the same reason: only the holder side of the comparison is measured locally, the responder side is a string the answerer supplies, and the lease row's incarnation is discarded by the read. One review drove a rogue holding an ordinary delivery credential that read the lease row itself, learned the holder's id, echoed it, and was accepted by the shipped classification against a real daemon. The other drove a raw responder using a copied standing delivery credential and produced shared verdicts in two of six sweeps, with the other four refused on a prefix mismatch it reads as a formatting-sensitive comparison and not a proof of provenance. Both add that the prose in the spec, the embedding doc and the changeset says the answerer is required to be the holder, which is stronger than the code enforces.

The other findings are non-blocking in the reviews that raise them: the delivery shard is pinned to zero; the lease-holder read conflates no live row with a failed read and the refusal text names only the second; three delivery fixtures were left lease-less and now exercise the undetermined branch; and the reply's lease boolean cross-check adds nothing against a dishonest responder.

---

Three of the first panel's four blocking findings are answered and I drove the evidence myself: absence is now determined from the delivery lease row rather than from the rail's outcome and I could not forge it; the foreign early return above the manager's own credential duties is gone, and reinstating either it or its failed-classification twin reddens a named cell; the same-identity divergent-generation refusal is back with a positive control. The fourth is not answered. The head replaces 'trust the reply's lease boolean' with 'read the lease row and require answer.responder to equal the [...]

**First-panel findings, disposition (4):**
- panel1-another review B1 / panel1-another review #1+#2: a peer on the queue-grouped delivery-admin rail can forge ABSENCE (first by quoting 'no responders' in a body, then by the typed no-responder error), putting a foreign manager on the remint path. ANSWERED. Absence no longer reads the rail's outcome at all. railAnsweredNothing only routes; the determination is ep.deliveryShardHeld(0), a KV point read of the lease row under the manager's own credential, and only `held === false` yields absent. `true` and `undefined` both throw and nothing reminted. I killed a reviewer-authored mutation (R1) that restores the rail-outcome reading: red on the named cell '(b1) a peer reply of EMPTY PAYLOAD + 503 STATUS HEADER is NEVER classified as absence', 57 marks of baseline 58. The suite's own (b1) arms drive both forgery shapes through the shipped classification against a real broker and real daemon, each with a [...]
- panel1-another review B2 / panel1-another review #2: a non-owning responder on the queue-grouped rail can flip a manager to SHARED, because the classification trusts the answer's content with nothing binding it to the process that reloads. NOT ANSWERED. The head adds a binding, and the binding is defeated by copying. The check is `holder !== answer.responder -> refuse`, where holder is read locally but responder is a string the reply supplies. The `delivery` profile that permits binding the rail also grants STREAM.MSG.GET on the delivery bucket, so a rogue reads the recorded holder and echoes it. My rig (real broker, real daemon rooted at B, real Manager rooted at A, shipped classifyDaemonSecretStore and shipped renewDaemonCreds) measured, 3 runs: honest control pure `foreign` 6/6 each run; own-id rogue 0 shared with 5-8 refusals naming the id mismatch; echoed-holder rogue 8-9 SHARED out of 16 with ZERO [...]
- panel1-another review foreign-manager-abandons-own-credentials / panel1-another review B3: the foreign early return sat above warnOnSystemCredExpiry, the managed-static scan, endpoint-serve, goal-writer, hosted runs and the session ledger, so a foreign manager renewed nothing it owned. ANSWERED. The early return is gone. Both the foreign branch and the failed-classification branch now fall through to every own-credential duty. Two reviewer-authored mutations confirm the suite notices: R4 reinstates the foreign return and is KILLED on '(a) a FOREIGN manager still reaches warnOnSystemCredExpiry' (56 marks of 58); R3 reinstates the return on the failed-classification path and is KILLED on '(a2) a pass whose classification FAILED still reaches warnOnSystemCredExpiry' (55 of 58). The suite's (a) and (a2) arms instrument Manager.prototype with DELEGATING wrappers so the shipped bodies still run.
- panel1-another review #3: the re-inverted smoke no longer guards the delivery same-identity divergent-generation refusal; a mutation disabling endpoint.ts's expected-generation check SURVIVED. ANSWERED. The (c) arm is restored with four cells plus a positive control: '(c) the rig really presents a DIFFERENT generation of the SAME identity (not a different nkey)', '(c) CONTROL: the daemon ACCEPTS the generation that is actually on disk', '(c) a DIFFERENT generation under the SAME daemon identity is REFUSED (nothing adopted)', '(c) ...and the refusal is the GENERATION mismatch, not the identity pin or a timeout'. All green 3/3. The suite also waits 1100ms before re-signing so a second-granular JWT iat cannot reproduce the first generation, which is the failure mode that would have made this arm vacuous. I did not independently re-run another review's exact endpoint.ts mutation; I take the restored cells and [...]

**Blocking (1):**
- another review-R17-B1. The lease-holder binding is satisfied by an ECHO: a non-holder that copies the recorded holder's id into `responder` is accepted, and a foreign manager then remints into a store the daemon never reads. implementations/manager/src/manager.ts classifyDaemonSecretStore, the `if (holder !== answer.responder)` check (~line 1599); packages/core/src/secret-store.ts DaemonStoreAnswer.responder; packages/core/src/provision.ts deliveryPermissions. The head replaced 'trust the reply's holdsDeliveryLease boolean' with 'read the lease row and require answer.responder to equal the recorded holder'. Both sides of that comparison are now measured, but only one of them locally: `holder` is read from the KV row, while `responder` is still a plain string the ANSWERER put in its own reply. The delivery-admin rail is queue-grouped and the `delivery` profile permits binding it. That same profile [...]

- another review-R17-D1. docs/embedding.md lines 204-208. This sentence survives from the previous round and states the rule this head deliberately abandoned. Absence is no longer determined from any signal on the rail, the broker's own or otherwise: the whole point of the change is that the no-responder signal is NOT sufficient, because a responder can manufacture it, and absence now requires the lease row to show no live holder. The paragraph immediately above it in the same file, added by this same change, says so correctly ('it determines an absent daemon from that same row rather than from the rail's outcome'). The page therefore states both rules a few lines apart. The stale one names the exact mechanism the first panel's B1 finding was about.
- another review-R17-D2. SPEC.md, the reloadStoreIdentity paragraph; .changeset/many-managers-per-space.md; implementations/manager/src/manager.ts classifyDaemonSecretStore doc comment. The requester verifies one half of the comparison and takes the other half from the reply. `responder` is a string the answerer chose; the row supplies only the value it is compared against. The manager's own comment goes further and says 'the binding that decides is the one measured here', and the changeset says the manager 'requires the answerer to BE that principal'. Requiring an answerer to be a principal and requiring its reply to name that principal are different requirements, and the head implements the second. I measured a non-holder satisfying it (see blocking finding another review-R17-B1).
- another review-R17-D3. implementations/manager/smoke/boot-self-heal-gate.smoke.ts, manager-restart-live.smoke.ts, registration-executor-resume.smoke.ts. The fixture claims the lease and does not hold it. The comment describes the intended fixture, not the one in the file. See another review-R17-N2.

**Non-blocking (3):**
- another review-R17-N1. `deliveryLeaseHolder` conflates 'no live lease row' with 'the read failed', and the manager's refusal text names only the second. packages/core/src/endpoint.ts deliveryLeaseHolder (~3672); implementations/manager/src/manager.ts, the `holder === undefined` refusal (~1595). deliveryLeaseHolder returns `(await readDeliveryLeaseEntry(shard))?.info.holder` inside a try/catch returning undefined. `undefined` therefore means either (a) the read threw, or (b) the read succeeded and there is no live row, or (c) a row exists with no `holder` field. The manager renders all of these as 'this manager could not read the delivery lease row', which is only case (a). Its sibling deliveryShardHeld keeps the two apart correctly by returning false for the empty row and undefined only on a throw, and its own doc comment insists `undefined` must never be read as absence. The holder reader does [...]
- another review-R17-N2. Three delivery fixtures were left lease-less while five others were updated, so those suites now exercise the undetermined branch rather than the shared-owner path they were written for. implementations/manager/smoke/boot-self-heal-gate.smoke.ts, manager-restart-live.smoke.ts, registration-executor-resume.smoke.ts. This head updated the delivery fixtures so their reloadStoreIdentity answers carry {identity, responder, holdsDeliveryLease:true}. Three of them assert the lease claim without acquiring the lease, and the manager's new binding compares the claim against the lease row it reads itself. With no row, the holder read is undefined and every pass refuses. _boot-delivery.ts and lifecycle-e2e.smoke.ts were given real acquireDeliveryLease/markDeliveryLeaseReady calls; these three were given only the reply-shape update and a comment claiming the fixture 'stands in for the [...]
- another review-R17-N3. The `holdsDeliveryLease` cross-check is described as adding safety, but contributes none against a non-honest responder. packages/core/src/secret-store.ts DaemonStoreAnswer; implementations/manager/src/manager.ts, the `!answer.holdsDeliveryLease` refusal before the holder comparison. The code and the changeset both say the claim is 'kept as a cross-check that must agree, so an honest non-holder is still refused by its own admission, but the binding that decides is measured locally'. That is accurate as far as it goes. The part left unsaid is that the local binding compares the row's holder to a string the same untrusted reply supplied, so for any responder willing to set holdsDeliveryLease true, neither test is an obstacle. The two checks are not independent: both are satisfied by one reply the answerer composed.

**Not verified (14):**
- pnpm smoke:ci. Not run. The brief documents it stopping at the harness on an absent docker/podman, which makes every cell after that point unverified rather than passed. I make no claim about it.
- pnpm check (the full gate). Not run; it chains dozens of :live suites, several of which the brief documents as failing here for insufficient storage (root disk resource-constrained).
- artifact-store and smoke:delivery-starvation. Not run; the brief names them as environment-fault reds on this host and I did not want to attribute a storage failure to the head.
- The upgrade-section gate's actual verdict on this range. Refused on a shallow clone, three times. I reasoned about the gate from its source instead and say so in answer 7; I did not observe it grade this branch.
- pnpm changeset status. Refused for the same shallow/no-main reason. I verified the changeset's package list against the four packages with first-party changes and against .changeset/config.json's fixed group by reading, not by running the tool.
- another review's original endpoint.ts expected-generation mutation. I did not re-drive it. I graded the (c) arm's restoration from its cells and its positive control, and from R2's kill on the adjacent (b2) property, rather than re-running the exact mutation that previously SURVIVED.
- Multi-shard behaviour. DELIVERY_SHARD is hardcoded to 0 with a comment saying runDelivery rejects any other layout. I read that claim in the comment and did not exercise a multi-shard deployment.
- The remote-manager profile end to end. I verified its new lease-read grant through delivery-lease-grant's permission matrix, not by running a remote manager against a broker.
- Whether the echo attack is reachable through any path that does not already hold a `delivery`-profile credential. I did not attempt to obtain one by other means, and `cotal mint` excludes that profile.
- Windows or macOS behaviour. Linux only.
- The writer's own mutation fixture. I deliberately wrote my own four mutations rather than re-running the writer's six, so my kills are independent; I did not verify the writer's fixture reproduces its reported results.
- node scripts/upgrade-section-gate.mjs --base 08cff3533d08aedaad3faa500cc265f5948a110c. my lane is a shallow clone (git rev-parse --is-shallow-repository -> true); the tool names its own refusal and I take it at its word rather than reading exit 2 as a defect in the head.
- pnpm changeset status. same shallow/no-main-ref cause; no `main` ref exists in the lane (git merge-base HEAD main -> 'fatal: Not a valid object name main').
- pnpm smoke:ci. the brief states it stops at the harness on a cell that spawns docker, and neither docker nor podman is installed here; a status-null probe for an absent tool is an instrument refusal and every cell after it is unverified. I did not run it and claim nothing from it either way.

---

The first-panel findings are substantially answered and their new cells discriminate under mutation. One blocking provenance defect remains: the manager compares a reply-declared responder principal with the lease row principal, but does not bind the reply to the lease incarnation. A process with a copied standing delivery credential can therefore answer as the lease-holder principal and produce shared classification without holding the lease. The public prose says the answerer is required to be the holder, which is stronger than the code enforces.

**First-panel findings, disposition (5):**
- Forgeable absence from a peer-controlled rail outcome. answered. classifyDaemonSecretStore routes typed no-responder outcomes to deliveryShardHeld(0), and absent requires a successful lease-row read showing no live row. The 503-header and phrase-echo cells passed 3/3; the header path is live broker coverage.
- A non-owning responder can claim shared and authorize a wrong-store remint. partly answered. Distinct-principal honest and lying responders are refused, and deleting the principal comparison is mutation-killed. However, the lease row stores {holder, incarnation}, while the manager reads only holder and compares it to answer.responder, a value parsed from reply.data. A copied standing delivery credential uses the same holder principal from a different endpoint incarnation. The live probe observed 22 false shared classifications across runs 2 and 3. Thus the original distinct-principal form is closed but responder-process provenance is not.
- A foreign or indeterminate classification abandons manager-owned renewal duties. answered. The early returns are gone. Both foreign and forced-classification-failure cells reach warnOnSystemCredExpiry and the managed-static scan. Restoring the failure-path early return was mutation-killed on the named a2 assertion.
- The two-root smoke did not guard same-identity divergent-generation refusal. answered. The suite now presents a real second generation for the same nkey, has a positive current-generation control, and requires a generation-mismatch refusal. It passed 3/3.
- Body omitted contract-homogeneous deployment requirement. answered. The changeset and docs/embedding.md now state that manager classes must be contract-homogeneous and describe mixed-class census/refusal behavior.

**Blocking (1):**
- responder-principal-is-not-process-provenance. A copied delivery credential can answer as the lease-holder principal and authorize shared classification without holding the lease incarnation. The lease row already carries holder and a random per-endpoint incarnation, and ownsDeliveryLease requires both. deliveryLeaseHolder discards incarnation and returns only holder. reloadStoreIdentity sends responder:this.card.id inside reply data. Manager parses that untrusted field and accepts when holder === answer.responder. A second process authenticated with a copied standing delivery credential has the same card.id but a different incarnation, so it can assert holdsDeliveryLease:true and name the manager store as shared. In an external TMPDIR copy of the real 54-cell broker/Manager smoke, a raw responder using the live daemon credential produced 10 shared verdicts in one 24-request sweep and [...]

- prose-overstates-responder-binding. SPEC.md, docs/embedding.md, and the changeset say the manager requires the answerer to be the recorded lease holder. The implementation compares only the principal and ignores the row incarnation, so the statement is not honored for a copied credential used by another process.

**Non-blocking (1):**
- changeset-status-refusal. pnpm changeset status could not grade because the detached shallow checkout has no synchronized main divergence point. instrument refusal, not a branch failure.

**Not verified (8):**
- I did not complete smoke:ci. The first pass timed out after 600 seconds during smoke:delivery-renewal, before the harness. Later cells are unverified, and no second or third full pass was attempted.
- I did not obtain a valid production upgrade-section range verdict because the lane is shallow.
- I did not complete a live wrong-store file write in the copied-principal probe. I did reproduce the load-bearing false shared classification 22 times; eight shipped renewal attempts in each of three later runs happened to select the honest/refusing path and wrote nothing.
- I did not test macOS or non-Linux platforms.
- Refusal not counted as a failure. pnpm changeset status: changesets could not find divergence from main in this detached shallow lane.
- Refusal not counted as a failure. production upgrade-section gate: exit 2 because the repository is shallow; nothing was graded.
- Refusal not counted as a failure. Initial same-principal external probes failed before execution due missing ESM/package resolution; these were instrument refusals and were repaired before evidence was counted.
- Refusal not counted as a failure. The renewal-record persistence probe failed before the scenario because its external copy lacked node_modules/.bin/tsx; no finding was drawn from it.

---

### Review 3: `APPROVE`

All three blocking findings from the first panel (B1 absence forgeability, B2 queue-group spoofability/lease unbinding, and B3 foreign manager dropping own-credential renewals) are thoroughly and robustly answered. Absence is now determined from a requester-side KV read of the delivery lease row rather than the rail's reply outcome. Lease holding is verified against the lease row holder rather than the responder's claim. Own-credential duties fall through and execute regardless of foreign classification or rail classification failure. All suites pass cleanly across repeated runs, docs bundle [...]

**First-panel findings, disposition (3):**
- B1: Absence was forgeable via string matching on error text or reply synthesis (NoResponders/503 status header). answered. implementations/manager/src/manager.ts no longer uses the rail reply or error text/type to determine absence. railAnsweredNothing() only gates checking the lease row via ep.deliveryShardHeld(DELIVERY_SHARD). If held === false, absence is determined. If held === true or undefined, it throws and skips remint. The supervisor profile in packages/core/src/provision.ts was granted read-only point-get access (STREAM.INFO and STREAM.MSG.GET) on KV_${deliveryBucket(space)} without write permission. bin/smoke/manager-two-root-renewal.smoke.ts now covers both string body 'no responders' and empty payload + 503 status header responses across queue sweeps, verifying that neither is ever classified as absence and neither remints into root A or pushes to root B.
- B2: Lease claim on queue-grouped rail was unauthenticated / trusted any responder's claim. answered. CotalEndpoint now includes responder (wire ID) and holdsDeliveryLease in reloadStoreIdentity replies. In manager.ts classifyDaemonSecretStore(), the manager reads the lease row holder itself via ep.deliveryLeaseHolder(DELIVERY_SHARD) under its own credential and verifies that holder === answer.responder, and also cross-checks answer.holdsDeliveryLease === true. A non-holder claiming true is refused by name. Covered in manager-two-root-renewal.smoke.ts section (b2) with both honest non-holder and lying non-holder responders asserting holdsDeliveryLease: true, verified killed by mutation when holder comparison is dropped.
- B3: Foreign manager stopped renewing its own service credentials and managed agents. answered. In implementations/manager/src/manager.ts renewDaemonCreds(), the foreign branch no longer executes an early return. Instead, it records the owner-elsewhere note and falls through to execute warnOnSystemCredExpiry(), the managed-static renewal loop (renewManagedStaticCred), serviceServe endpoint-serve renewal, goal-writer renewal, runHosting renewal, and session-ledger renewal. Similarly, when classification throws (e.g. timeout, unreadable reply, or lease check failure), the catch block writes the audit record and falls through instead of returning early. Covered in manager-two-root-renewal.smoke.ts sections (a) and (a2), and verified killed by mutation when restoring the return in the catch block.

**Blocking (0):** none

**Non-blocking (1):**
- delivery-shard-pinned-to-zero. DELIVERY_SHARD is hardcoded to 0 in implementations/manager/src/manager.ts. Line 908 hardcodes DELIVERY_SHARD = 0. This matches runDelivery in implementations/delivery/src/delivery.ts, which rejects shards !== 1 || shard !== 0. If Cotal later supports multi-shard delivery daemons, the reloadStoreIdentity wire request will need to pass the target shardIndex rather than assuming shard 0.

**Not verified (5):**
- smoke:ci full run across all shards was not run due to missing docker/podman on the host for the harness and disk storage limits on artifact store
- Physical multi-host network deployment across separate physical machines (exercised locally via isolated multi-root processes and independent brokers)
- Refusal not counted as a failure. upgrade-section-gate --range with git fetch --unshallow needed for full git history on shallow clone (exit code 2 is misuse/refusal, self-test passed 93/93).
- Refusal not counted as a failure. docker/podman absent in CI environment noted in WHAT-TO-GRADE.
- Refusal not counted as a failure. smoke:delivery-starvation / artifact-store JetStream capacity constraint due to host 96% disk capacity.

davidfarah2003:
Closing this in favour of two heads that split what it carried.

The many-managers change shipped as #1670 (`4ab8b4ba5`, 2026-09-17), which elects a single
daemon-credential renewal owner through a per-space lease. Verified on main at `b086974ff`:
`bin/smoke/manager-two-root-renewal.smoke.ts` exits 0 with 29 named assertions reached, 29 passing,
including the cells proving that two managers sharing one store both start and that exactly one owns
renewal across the lease TTL.

What this branch carried that #1670 did not is the binding on the store-identity challenge: the
delivery-admin rail is queue-grouped, so the reply has to name the answering process, and the
manager has to verify that process is the delivery lease holder rather than take its word for it.
That part is now open as #1695 against current main, at 6 files and +201/-5 rather than 21 files and
+1028/-113, with the binding proved load-bearing by an inverted mutant and the new permission proved
by a production-only revert. It tracks #1694.

Two things here are deliberately not carried into #1695. Determining daemon absence from the lease
row instead of from a rail outcome is a real finding and its own change; it is recorded in #1694.
The rename of `daemonStoreRelation` and its return type belonged to this branch's larger shape and
is unnecessary for the binding alone.

This head has also been conflicting against main since 2026-09-16, with main 54 commits past its
base, so there is nothing here worth rebasing that #1695 does not already carry.

PR #1640, opened 2026-09-15
fix(core,cli,runtime): scope an unanswered endpoint verdict to the rail the request rode

### LINKED ISSUE #1630: cli: a 0.49.0 client against a 0.48.2 manager reports "no manager answered" and suggests --local; the cause is the versioned ep.v1 rail and it is never named
## What happens

A 0.49.0 CLI talking to a 0.48.2 manager gets this from every `cotal run` command that needs the manager (`ps`, `journal`, `answer`, `start`):

```
no manager answered on the endpoint rails (deadline-exceeded: no describe reply from manager within 10000ms); is a manager running for this mesh? A run can still be driven from this terminal with --local
```

Exit 1, on stderr, after about 13.5 s against the stated 10 s budget. A manager is running, idle on the roster, starting runs and renewing credentials the whole time. `cotal ps` works from the same shell, and `cotal run ps --local` works. The same command through the 0.48.2 CLI on the same host, same shell, same credentials, returns rc 0.

Measured 2026-09-15 between 10:07Z and 10:20Z, ten CLI attempts from a managed seat plus six from a bare shell, all identical. Cross-check that isolates the CLI version:

| command | CLI 0.48.2 | CLI 0.49.0 |
|---|---|---|
| `cotal ps` | rc 0 | rc 0 |
| `cotal run ps` | rc 0, 173 rows | rc 1, describe refusal |
| `cotal run journal <live run>` | rc 0, 16 lines | rc 1, describe refusal |
| `cotal run ps --local` | rc 0 | rc 0 |

The manager process has only the 0.48.2 package tree mapped. The 0.48.2 seat credentials in play were issued 4.4 h earlier with 19.6 h left, so this is not the 24 h boundary. The broker logged no publish violation for the refused describes, so nothing was denied; the request went to a subject nobody serves.

## Why

#1395 (36d177951, in 0.49.0) moved the run commands' manager describe onto the versioned `ep.v1` rail: `epPlaneTokens(caller)` emits the `v1` form whenever the caller carries an issued generation, and the workspace's minted control caller now does (`epCaller = { owner, actor, uid, generation: issued.generation }` in 0.49.0; the 0.48.2 form has no generation). A 0.48.2 manager serves the unversioned `ep` subjects only, so the describe is never delivered and the client waits out its deadline. The class rail that `cotal ps` uses did not change, which is why that command masks the skew.

## Two defects

1. **The diagnosis is false.** The text tells the operator that no manager is running and offers `--local`, which drives the run from the caller's own process and names the caller as its answerer. A worker seat on this mesh correctly refused to take that path and left its run answer unsubmitted; a less careful one would have forged the answerer identity on the tool's own advice. An unanswered describe on a versioned subject should say that no manager serves this rail generation and name the generation, and the `--local` hint should not appear when a manager is visibly registered.

2. **There is no negotiation or preflight.** The CLI has the manager's registered instance record (it can read the version from the same store `cotal ps` reads) and could either fall back to the unversioned rail for a pre-`v1` manager or refuse up front with the two versions in the message. Today an operator who upgrades the CLI before restarting the manager, which is the documented order for a host upgrade, loses every `cotal run` verb until the manager restart, with no message that says so.

## What would close this

- The refusal names the rail generation it tried and the manager generation it found, or says the manager's generation is unknown; the `--local` hint is dropped when a manager instance is registered.
- Either a documented compatibility rule (CLI N with manager N-1 is refused by name) or a fallback describe on the unversioned rail when the registered manager predates `v1`.
- A cell that runs a 0.49.0-shape client against a serve loop that only subscribes the unversioned rail and asserts the named refusal rather than the deadline text, plus a mutant that restores the deadline text.

Tested at `main` 08cff3533 with the 0.49.0 and 0.48.2 published CLIs.

L4XB:
Closes #1630

## What was wrong

A 0.49.0 CLI carries an issued generation, so `epPlaneTokens` puts every request on the versioned
`ep.v1` rail. SPEC 13.15 makes that rail a separate subject space from the legacy `ep` rail and
requires an endpoint to serve both, so a 0.48.2 manager subscribes `ep` alone and never receives the
describe. The describe then waited out its whole budget and the CLI turned that silence into a
verdict about the mesh:

```
no manager answered on the endpoint rails (deadline-exceeded: no describe reply from manager within 10000ms); is a manager running for this mesh? A run can still be driven from this terminal with --local
```

Both halves of that sentence are wrong in this state, and the second half is the expensive one.
`--local` drives the run in the calling process and names the caller as the run's answerer, so an
operator who follows the tool's own advice submits an answer under the wrong identity. The reporter's
worker seat declined and left its answer unsubmitted.

No fallback describe on the unversioned rail is added, and the reason is not only the repo's
no-fallbacks rule. `epRequestGrantRows` mints request rows under `plane(caller)`, so an issued
credential holds rows on `ep.v1` only: a legacy-rail describe from this caller would be refused by
the broker at publish, not answered. Naming the skew is the only thing the client can do from here.

## The change

- `EP_UNANSWERED` carries the `ep` plane the request was published on, and `describeEndpoint` names
  it in its own refusal (`… within 10000ms on the ep.v1 rail`). `unansweredRail(e)` reads it back, so
  a renderer branches on the marker rather than on the message.
- `epRailFailure` scopes the verdict to that rail. On `ep.v1` it says nothing answered *on that
  rail*, states that the two rails are disjoint at the broker and that an endpoint must serve both,
  and **names both causes the silence is consistent with**: no manager running, or one older than
  the rail. This side cannot tell them apart, because the service registry records no package
  version, so it asserts neither.
- `cotal run`'s hosted verbs do the same and drop the question and the `--local` suggestion there.
  `--local` is the remedy that follows from asserting one cause, and it names the caller as the
  run's own answerer.
- On the legacy rail every message is byte-identical to before: there is no second rail its silence
  could be hiding a manager on.

## Callers that read the sentence instead of the marker

Renaming an operator-facing sentence is only safe if nothing keys on it, so every consumer of the
old headline was swept. Two did key on it, and both are fixed here rather than worked around:

- **`cotal update`** (`reportRunningManager`) returned "no manager is running" on
  `status.unanswered && /no manager reachable/i.test(status.error)`. It runs under an issued control
  caller, so it threw where it used to report "none". It now keys on `unanswered` alone, which is
  the marker `epRailFailure` derives from core's answer provenance. The string test was redundant
  when written: that call passes no `--on`, so an unanswered reply can only be the unpinned verdict.
- **`bin/smoke/control-transport-dial.smoke.ts`** decided whether a dial reached the ep rails by
  matching the headline. Its `models` cells ride an issued caller on `ep.v1` and its `ps` cells ride
  the scatter, which re-mints a caller with no generation and stays on the legacy rail, so one
  identical dial produced two headlines. It now keys on the describe deadline, which is what that
  rig is built around (its own note: it "intentionally has no manager so it can use the
  endpoint-rails deadline as the discriminator"), is common to both rails, and proves more.

The two `VERDICT` regexes used for negative assertions gained the rail-scoped spelling, so a
reworded verdict cannot satisfy "no reachability verdict is stated" by failing to match.
`component-health`, `manager-singleton-live`, `cli-on-instance-live` and `persona-announce` were
checked and left alone: they either drive `ps`, which stays on the legacy rail, or assert the
headline's absence, where a rename cannot produce a false pass.

There is no behaviour change for legacy-only managers. The rails are disjoint at the broker and the
grant rows are disjoint with them, so an issued caller could not reach a legacy-only manager before
this change either. Only the sentence changed.

## Evidence

Live, against a real `nats-server`, with a serve loop subscribed to the unversioned rail only (what
a 0.48.2 manager is) and two callers that differ in nothing but the issued generation.

Before:

```
0.48.2 client (no generation): 2ms  code=unavailable  unanswered=false
  message: describe(manager) failed: answered by the 0.48.2 manager
0.49.0 client (issued generation): 2002ms  code=deadline-exceeded  unanswered=true
  message: no describe reply from manager within 2000ms
  the run path then prints: no manager answered on the endpoint rails (…); is a manager running for
  this mesh? A run can still be driven from this terminal with --local
```

After, the same fixture, promoted into `implementations/cli/smoke/ep-rail-failure.smoke.ts`:

```
✓ live: the ep-only serve loop ANSWERS a legacy caller (the control: this manager is serving)
✓ live: the same loop never sees an ISSUED caller's describe, and the refusal names the rail it rode
✓ live: the CLI renders the ep.v1 skew as a named refusal, not as 'no manager reachable'
EP-RAIL-FAILURE POLARITY SMOKE OK ✅  (31 passed, 0 failed)
```

`run-command-usage` grades the hosted-verb sentence off the wire, both rails and the unmarked case:

```
✓ legacy rail: the question and the --local remedy stand (one rail, and nothing answered on it)
✓ versioned rail: the refusal names ep.v1 and states what the silence does not establish
✓ versioned rail: neither the false question nor the --local remedy that follows from it is printed
✓ an unanswered failure whose producer recorded no rail keeps the legacy wording
run-command-usage: 19 ok, 0 failed
```

Two mutation fixtures restore the pre-fix behaviour, one half at a time, and all three mutants are
killed:

```
KILLED  the renderer stops reading the rail and returns to the unscoped verdict
KILLED  the describe deadline goes back to the rail-blind text and marker
KILLED  the refusal returns to one rail-blind sentence, with the question and the --local remedy on every rail
```

`control-transport-dial` was red at this head before the predicate fix and green at the base; it is
5/5 now, and its mutation fixture grades again (2/2 killed) instead of refusing on a red baseline.

Also run green at this head: `smoke:ep-invoke`, `smoke:ep-verbs`, `smoke:ep-serve`, `smoke:ep-grants`,
`smoke:ep-publish-denial:auth`, `smoke:manager-invoke-verdict`, `smoke:legacy-manager-report`,
`smoke:cli-kernel`, plus a re-proof of the existing mutation fixtures that guard the sources this
touches (all mutants still killed). `pnpm typecheck` and `pnpm check:docsbundle` are clean, and the regenerated
`docs-bundle.generated.ts` is in the same commit as the `docs/cli.md` edit.

## Not done here

The describe still spends its full budget before failing. `epCall` publishes with the reserved
no-responders sentinel and learns from the broker in milliseconds that nothing serves a rail;
`describeEndpoint` publishes with no reply-to and can only infer silence. Wiring the sentinel into
the describe would make this refusal affirmative and fast, but it interacts with the startup retry
window that exists so a responder registering after the first publish is still reached, so it is a
separate change rather than a rider on this one.

File and line references are against `main` @ `08cff3533`.


davidfarah2003:
Cold read complete: one independent review of `952898af336cebaef8000e33b4ac609e95a56b5f`.

Relayed by the orchestrator. The measurements below belong to the reviewer. Two further reviews follow; this one is posted early because its body defect is actionable now.

### Review 1: `APPROVE`

**Blocking (0):**
_none_

**Body defects (2):**
- BD1 (material), PR #1640 body, section 'The change', last bullet; and .changeset/ep-rail-scoped-unanswered-verdict.md, final paragraph of the summary. The core describe deadline message changed for every caller. Base 08cff3533 packages/core/src/endpoint-invoke.ts:163 produced `no describe reply from ${endpoint} within ${deadlineMs}ms`. Head 952898af line 168 produces `no describe reply from ${endpoint} within ${deadlineMs}ms on the ${rail} rail`, with no conditional. Running the real describeEndpoint + epRailFailure + unansweredManagerRefusal against a live nats-server for an unissued caller yields: core 'no describe reply from manager within 1500ms on the ep rail'; cotal ps 'no manager [...]
- BD2 (minor), PR #1640 body, 'Evidence' section. The 'Before:' block reports 2ms and 2002ms timings measured pre-fix. They are not attributable to this head and are not reproducible from it. Presented under 'Evidence' without a sha attached, they read as measurements of the shipped change.

**Non-blocking (3):**
- NB1 (low), implementations/runtime/src/run-command.ts unansweredManagerRefusal(), and implementations/cli/src/lib/control.ts epRailFailure() skew clause. The versioned-rail message asserts the CAUSE ('a manager older than ep.v1 serves ep only and cannot answer here') in a message that was produced from silence alone, and the client cannot distinguish a version-skewed manager from a manager that is simply not running. Measured: On a current auth mesh at THIS head, with the manager stopped and no version skew present anywhere, the real binary printed the full 'a manager older than ep.v1 serves [...]
- NB2 (informational), issue #1630, 'What would close this'. The issue lists a fallback describe on the unversioned rail as one acceptable remedy. The PR deliberately does not do that and explains why, and the explanation is correct against the code and SPEC 13.15 (an issued credential holds no legacy-rail row, so such a describe would be refused at publish). Flagged only so a later reader does not mistake the divergence for an oversight.
- NB3 (informational), implementations/cli/src/lib/control.ts, unanswered branch. The CLI's skew tail is appended after the closing parenthesis with a single leading space and no sentence-ending period on the preceding clause, producing '...on the ep.v1 rail) The ep.v1 and legacy ep rails are disjoint...'. The runtime's sibling message places a period there ('...rail). The ep.v1 and...'). Cosmetic inconsistency between two messages that otherwise deliberately mirror each other. Measured: Observed in the real epRailFailure output captured from the live run.

**Not verified (6):**
- The PR body's claim of a 're-proof of the eight existing mutation fixtures that guard the sources this touches (all mutants still killed)'. I located eight fixtures referencing the touched sources but did not run them; that is a long serial run I did not complete within this review. Neither confirmed nor disputed.
- The body's pre-fix 'Before:' timings (2ms / 2002ms). Measured at a different sha; I did not reproduce them and do not carry them across.
- `pnpm changeset status` did not run in my lane: changesets failed with 'Failed to find where HEAD diverged from main' because this lane has no main ref. This is an environment fault in my lane, not a PR defect and not a verdict. I read the changeset file directly instead and confirmed its package list against the actual diff.
- CI at head 952898af336cebaef8000e33b4ac609e95a56b5f was still in flight when I read it (read-only, via gh api check-runs; I ran nothing). Two reads several minutes apart both returned 15 success, 3 skipped, 12 pending, and ZERO failure/cancelled/timed_out. Green so far includes unit, plan, attribution, Windows/required, windows-ok, seat native (linux-arm64) and three of the twelve changed shards. The remaining [...]
- Whether any downstream consumer outside this repository parses the core describe deadline string. Inside the repository I checked: the only assertions on that text are prefix regexes in implementations/manager/smoke/cli-on-instance-live.smoke.ts (which still match, since the new words are appended) and hand-built fixture strings in three smokes. No exact-equality assertion on the full sentence exists in-tree.
- Behaviour on a genuine 0.48.2 manager binary. I reproduced the skew's SHAPE with an ep-only serve loop (the shipped smoke's method) and reproduced the operator-visible sentences with a real stopped manager, but I did not install a 0.48.2 manager to observe the true mixed-version mesh.

**Questions answered:**
- noFallbackDescribeOnTheOtherRail: VERIFIED TRUE. No fallback describe is issued, and the body's stated reason holds against the code.
- epUnansweredFieldGradedByACellThatRedsUnderAMutant: YES, and I proved it for the FIELD specifically, not merely for the prose that happens to contain the same word.
- localAdviceGoneFromTheUnansweredPathAndOnlyThere: YES on both halves. It is gone from the versioned-rail unanswered path, it SURVIVES on the legacy unanswered path and at every other site, and I confirmed both by running the real binary rather than by reading only.
- doTheTwoMutationFixturesDiscriminate: YES. All three mutants across the two fixtures were KILLED, each reddening the cell the fixture named in advance. Both were run, not read.
- changesetOrBodyOverstatement: YES, one substantive overstatement, repeated in both the PR body and the changeset, plus one smaller unverifiable claim. Reported, never edited.

davidfarah2003:
CI at `952898af336cebaef8000e33b4ac609e95a56b5f` is red on two jobs for one cause, and it reproduces deterministically off the runner (2 of 2 runs on a separate machine at this head; green at the base `712dfadf`).

`pnpm smoke:control-transport-dial` fails cells A and C:

```
✗ FAIL: A: models reaches the manager over ws:// through askManagerEp - {"status":1,"out":"✗ no manager answered on the ep.v1 rail (deadline-exceeded: no describe reply from manager within 10000ms on the ep.v1 rail) The ep.v1 and legacy ep rails are disjoint at the broker (SPEC 13.15) ..."}
✗ FAIL: C: models still reaches the ep rails over nats:// - (same text)
control transport dial: 3 passed, 2 failed
```

The suite's `reachedRails` predicate in `bin/smoke/control-transport-dial.smoke.ts` accepts a refusal only when it matches `no manager reachable on the ep rails` (or the two registry texts). This change rewrites that refusal in `implementations/cli/src/lib/control.ts` to `no manager answered on the ep.v1 rail (...)` whenever the describe rode the versioned rail, which is the case for every issued caller, including the one the suite mints. Cells B and D still pass because `ps` goes through the scatter path, whose text is unchanged. So the cells are reporting the new wording, not a transport regression: the CLI did reach the rails over both ws:// and nats://.

The mutation reproof job hits the same thing one step earlier: it selects `bin/smoke/mutations/control-transport-dial.json` because `control.ts` changed, and refuses it because its command is red before any mutation, which the tool reports as a green-at-base, red-at-head transition.

Two ways to close it, either is fine from the maintainer side: teach the predicate the rail-scoped text (a match on `no manager answered on the ep\.v1 rail` alongside the legacy one), or have cells A and C assert the scoped verdict explicitly, which is the property this change adds and would make the suite grade it. Either way the fixture needs its baseline re-proved at the new head so the reproof job can run its mutants.

davidfarah2003:
Panel complete: three independent reviews of `952898af336cebaef8000e33b4ac609e95a56b5f`.

Two reviews approve and one requests changes, so this is recorded as **not green** at this head. The requesting review's single blocking item is the control-transport-dial predicate that CI reddened on, which it classifies as a head-caused test expectation regression rather than a production behaviour regression. All three reviews measured the same body defect independently: the claim that every legacy-rail message is byte-identical is false, because the core describe deadline text now carries "on the ep rail" for a legacy caller; two also note the changeset repeats it and one notes the existing-fixture re-proof claim is overstated while that fixture's baseline is red. Each review reproduced the two new mutation fixtures killing their mutants and the touched suites green. The head has since moved to `31f005e38b1c759efec1b2ca725dc0c637fc9a1c`; these measurements are for the head named above, and the new head gets its own panel.

---

### Review 1: `APPROVE`

**Blocking (0):** none

**Body defects (2):**
- The core describe deadline message changed for every caller. Base 08cff3533 packages/core/src/endpoint-invoke.ts:163 produced `no describe reply from ${endpoint} within ${deadlineMs}ms`. Head 952898af line 168 produces `no describe reply from ${endpoint} within ${deadlineMs}ms on the ${rail} rail`, with no conditional. Running the real describeEndpoint + epRailFailure + unansweredManagerRefusal against a live nats-server for an unissued caller yields: core 'no describe reply from manager within 1500ms on the ep rail'; cotal ps 'no manager reachable on the ep rails (deadline-exceeded: no describe reply from manager within 1500ms on the ep [...]
- The 'Before:' block reports 2ms and 2002ms timings measured pre-fix. They are not attributable to this head and are not reproducible from it. Presented under 'Evidence' without a sha attached, they read as measurements of the shipped change. The body does close with 'File and line references are against main @ 08cff3533', which partly covers this, but the timings sit above that line and are not obviously in its scope.

**Non-blocking (3):**
- The versioned-rail message asserts the CAUSE ('a manager older than ep.v1 serves ep only and cannot answer here') in a message that was produced from silence alone, and the client cannot distinguish a version-skewed manager from a manager that is simply not running. On a current auth mesh at THIS head, with the manager stopped and no version skew present anywhere, the real binary printed the full 'a manager older than ep.v1 serves ep only and cannot answer here' sentence. In that run there was no old manager. The message's load-bearing clause, 'This does not establish that no manager is running', remains true in both states, and the remedy [...]
- The issue lists a fallback describe on the unversioned rail as one acceptable remedy. The PR deliberately does not do that and explains why, and the explanation is correct against the code and SPEC 13.15 (an issued credential holds no legacy-rail row, so such a describe would be refused at publish). Flagged only so a later reader does not mistake the divergence for an oversight.
- The CLI's skew tail is appended after the closing parenthesis with a single leading space and no sentence-ending period on the preceding clause, producing '...on the ep.v1 rail) The ep.v1 and legacy ep rails are disjoint...'. The runtime's sibling message places a period there ('...rail). The ep.v1 and...'). Cosmetic inconsistency between two messages that otherwise deliberately mirror each other. Observed in the real epRailFailure output captured from the live run.

**Not verified (6):**
- The PR body's claim of a 're-proof of the eight existing mutation fixtures that guard the sources this touches (all mutants still killed)'. I located eight fixtures referencing the touched sources but did not run them; that is a long serial run I did not complete within this review. Neither confirmed nor disputed.
- The body's pre-fix 'Before:' timings (2ms / 2002ms). Measured at a different sha; I did not reproduce them and do not carry them across.
- `pnpm changeset status` did not run in my lane: changesets failed with 'Failed to find where HEAD diverged from main' because this lane has no main ref. This is an environment fault in my lane, not a PR defect and not a verdict. I read the changeset file directly instead and confirmed its package list against the actual diff.
- CI at head 952898af336cebaef8000e33b4ac609e95a56b5f was still in flight when I read it (read-only, via gh api check-runs; I ran nothing). Two reads several minutes apart both returned 15 success, 3 skipped, 12 pending, and ZERO failure/cancelled/timed_out. Green so far includes unit, plan, attribution, Windows/required, windows-ok, seat native (linux-arm64) and three of the twelve changed shards. The remaining twelve (most smoke shards, live, seat native linux-x64) had not concluded. I did not re-run any workflow; a re-run is a write.
- Whether any downstream consumer outside this repository parses the core describe deadline string. Inside the repository I checked: the only assertions on that text are prefix regexes in implementations/manager/smoke/cli-on-instance-live.smoke.ts (which still match, since the new words are appended) and hand-built fixture strings in three smokes. No exact-equality assertion on the full sentence exists in-tree.
- Behaviour on a genuine 0.48.2 manager binary. I reproduced the skew's SHAPE with an ep-only serve loop (the shipped smoke's method) and reproduced the operator-visible sentences with a real stopped manager, but I did not install a 0.48.2 manager to observe the true mixed-version mesh.

---

**Blocking (1):**
- Update the existing control-transport-dial acceptance predicate for the new scoped refusal. At the pinned head, `pnpm smoke:control-transport-dial` reached the intended CLI endpoint path but failed cells A and C because `reachedRails()` in bin/smoke/control-transport-dial.smoke.ts accepts only the old `no manager reachable on the ep rails` wording (or registry variants). The actual output was the new `no manager answered on the ep.v1 rail ...` refusal. I reproduced 3 passed, 2 failed at 2026-09-15T16:50:21Z to 16:51:42Z. GitHub smoke shard 2 independently recorded the same A and C failures and 3/2 result at the same head. Mutation reproof [...]

**Body defects (2):**
- On the legacy rail every message is byte-identical to before. False as written in the PR body fetched at 2026-09-15T15:40:31Z and still present in the current body read at 2026-09-15T17:07:54Z. `describeEndpoint` now appends `on the ep rail` to the core deadline message for a legacy caller. The CLI and hosted-run outer renderer preserve their prior legacy wording, but the nested core message is not byte-identical.
- Re-proof of the eight existing mutation fixtures that guard the sources this touches (all mutants still killed). Overstated at the pinned head. GitHub mutation reproof shard 9 refused the existing control-transport-dial fixture because its baseline is head-red, and the aggregate check is red. Locally I reproduced that baseline red. The PR's two new mutation fixtures do kill all three new mutants, but the blanket existing-fixture claim is not the actual current state.

**Non-blocking (1):**
- Changeset repeats the legacy unchanged overstatement. .changeset/ep-rail-scoped-unanswered-verdict.md says `On the legacy rail every message is unchanged`. The core describe deadline text changed from `within <ms>ms` to `within <ms>ms on the ep rail`. Narrow the statement to the outer CLI and hosted-run reachability wording if that is what is intended.

**Not verified (2):**
- I could not obtain a trustworthy local base run of `smoke:control-transport-dial` or `smoke:runtime-run-driver-auth` after 17:11Z because the host was again under heavy whole-box pressure: load was about 28 to 32, swap was full, and even core-pinned runs timed out during JetStream setup before the suites' grading cells. The first private base archive attempt also used head-built package dist and is not accepted as a base measurement. I therefore do not claim a local base result.
- I did not complete the full repository smoke suite. The focused changed-path checks, typecheck, docs bundle, real entry point, and new mutation fixtures were run.

---

### Review 3: `APPROVE`

**Blocking (0):** none

**Body defects (2):**
- PR body (fetched 2026-09-15T15:40:31Z) states: 'On the legacy rail every message is byte-identical to before: there is no second rail its silence could be hiding a manager on.' and changeset states: 'On the legacy rail every message is unchanged...'. In reality, packages/core/src/endpoint-invoke.ts line 168 now appends ' on the ${rail} rail' (' on the ep rail' for legacy callers) to the describe rejection message. Because this message forms ${detail}, the rendered legacy refusal in both cli (epRailFailure) and runtime (unansweredManagerRefusal) now includes ' on the ep rail', which is not byte-identical or unchanged compared to the pre-fix [...]
- PR body implies that core smoke suites verify the new functionality, but no test within packages/core/smoke asserts the new rail property on EP_UNANSWERED or tests unansweredRail(); coverage for these exists only in implementations/cli and implementations/runtime suites.

**Non-blocking (1):**
- changeset lists @cotal-ai/connector-core with a minor bump, although connector-core only had docs-bundle.generated.ts regenerated without any first-party functional code changes. However, this is harmless given the lockstep fixed versioning group.

**Not verified (0):** none

davidfarah2003:
Panel complete: three independent reviews of `31f005e38b1c759efec1b2ca725dc0c637fc9a1c`.

All three reviews approve with zero blocking items, so this head is recorded as **green** on review. The review that requested changes at the previous head confirmed its blocking item resolved here, and another review confirmed that resolution cell by cell rather than inheriting it. One body defect survives in all three reviews: the claim that every legacy-rail message is byte-identical to before is false at this head, because the core describe deadline text now carries the rail name for a legacy caller; two reviews note the changeset repeats it. One review adds that the production line changed by this head is not graded by any cell, shown with a killed control in the same run, and that the dial predicate is now satisfiable by a describe that reached the broker but was refused at publish. These are recorded as body and coverage notes, not blockers.

---

### Review 1: `APPROVE`

**Blocking (0):** none

**Body defects (2):**
- BD1. PR #1640 body, section 'The change', final bullet; and .changeset/ep-rail-scoped-unanswered-verdict.md, final paragraph of the summary. False. packages/core/src/endpoint-invoke.ts:168 builds the describe deadline as `no describe reply from ${endpoint} within ${deadlineMs}ms on the ${rail} rail` with no branch guarding the suffix, and attaches `rail` to the EP_UNANSWERED detail unconditionally. `rail` is `epPlaneTokens(caller).join('.')` (endpoint-subjects.ts:198-199), which is 'ep' for a legacy caller, so a legacy caller's message gains ' on the ep rail' too. STILL LIVE at this head.
- BD2. The 'Before:' block presents timings (2ms for the legacy client, 2002ms for the issued one) under a heading that reads as evidence for the shipped change. They were measured pre-fix at another sha. I did not reproduce them and I do not carry them across. The body does close with 'File and line references are against main @ 08cff3533', which covers the line references but not the timings. PR #1640 body, 'Evidence' section, the 'Before:' block. STILL LIVE at this head. Minor and partly self-disclosed. Flagged for completeness, not as a reason to withhold approval.

**Non-blocking (4):**
- NB1. The production line changed by commit 31f005e3 is not graded by any cell, proven with a killed control in the same file and the same run. implementations/cli/src/commands/update.ts:236, `if (status.unanswered) return "none";` inside reportRunningManager. Three reasons, and I weighed them rather than assumed them. (a) The gap is pre-existing, not introduced: the string test this commit removed was equally ungraded at 952898af and at the base, so the PR does not make coverage worse. (b) The seam ITSELF is now pinned: commit 31f005e3 added a cell to ep-rail-failure asserting `unanswered` is true on both rails while the sentences differ, [...]
- NB2. The dial predicate is now satisfiable by a describe that reached the broker but was refused at publish. bin/smoke/control-transport-dial.smoke.ts:139-141.
- NB3. The prior panel's NB about an asserted cause is resolved, and resolved in the right direction. At 952898af the versioned message asserted the skew as the cause ('a manager older than ep.v1 serves ep only ... this is not evidence that no manager is running. Check the manager's version.'), which sent an operator whose manager was simply down to check a version. At this head both the CLI (control.ts:262) and the runtime (run-command.ts:534) name both causes and assert neither, and direct the operator to check whether a manager is running FIRST and its version only if it is. That is the correct ordering: the cheaper and more likely check [...]
- NB4. run-refusal-rail.json carries an incidental whitespace reformat alongside its substantive edit. The delta reformats the `suite` array from one line to three and appends a trailing newline inside both the `find` and `replace` strings, beyond the wording change the commit is about. I confirmed the fixture still functions: I ran it at this head and its single mutant was KILLED, red and named on 'versioned rail: neither the false question nor the --local remedy that follows from it is printed', 17 marks against a baseline of 19. Noise in the diff, not a defect.

**Not verified (5):**
- I reproduced the skew's SHAPE the way the shipped smoke does, with a serve loop subscribed to the unversioned rail only, and I reproduced the two rails' real messages live against a real nats-server. I did not install a 0.48.2 manager. The rail-disjointness the fix rests on is SPEC 13.15 and is visible in epPlaneTokens, so I am confident in the mechanism, but I did not observe the true mixed-version mesh. Behaviour against a genuine 0.48.2 manager binary on a real mixed-version mesh.
- I did not read it, and I ran nothing on GitHub. A CI re-run is a write and I made none. My greens are my own lane's, listed in evidence.suitesRun. CI status at this head.
- I ran the changed-path suites, the three shipped mutation fixtures the PR claims, the dial fixture, typecheck, check:docsbundle, and four of the body's other claimed-green suites. I did not run every shard. smoke:delivery-starvation is documented as failing on this host for storage reasons and I did not run it; delivery-boot-honesty I deliberately did not hand-run, per the brief. The full repository smoke suite.
- Not run: this lane has no `main` ref for changesets to diverge from, which is a property of my lane and not a PR defect. I read .changeset/ep-rail-scoped-unanswered-verdict.md directly and checked its package list against git diff --stat 08cff3533..31f005e3. The four listed packages (core, cli, runtime, connector-core) all have changed files in the diff; connector-core's is only the regenerated docs bundle, which the lockstep `fixed` group would version anyway. pnpm changeset status.
- Inside the tree I swept it: the only assertions are prefix regexes in implementations/manager/smoke/cli-on-instance-live.smoke.ts, which still match because the new words are appended, and hand-built fixture strings in three smokes. I cannot speak for out-of-tree consumers. Whether any consumer OUTSIDE this repository parses the core describe deadline string.

---

### Review 2: `APPROVE`

**Blocking (0):** none

**Body defects (1):**
- On the legacy rail every message is byte-identical to before. False in the PR body fetched at 2026-09-15T17:50:53Z. packages/core/src/endpoint-invoke.ts unconditionally changes a legacy describe deadline from `no describe reply from <endpoint> within <ms>ms` to the same text plus `on the ep rail`. The outer CLI and hosted-run legacy renderings retain their prior branch, but the nested core message bytes do not.

**Non-blocking (2):**
- The changeset repeats the false legacy-message claim. .changeset/ep-rail-scoped-unanswered-verdict.md says `On the legacy rail every message is unchanged`, but describeEndpoint changes the legacy core deadline text by appending `on the ep rail`.
- The new cotal update marker decision has no discriminating regression cell. In a reviewer-only mutation run with a green baseline, restoring the exact pre-fix predicate `status.unanswered && /no manager reachable/i.test(status.error ?? "")` in implementations/cli/src/commands/update.ts SURVIVED `smoke:cli-kernel` plus `smoke:ep-rail-failure`. In the same file and same run, removing the existing legacy-update stop was KILLED at `legacy report refuses before extension replay can mutate packages`, proving the harness reached update.ts. The added ep-rail-failure cell proves only that ManagerReply.unanswered is rail-independent. It does not [...]

**Not verified (3):**
- I did not run the full smoke:ci suite or every existing mutation fixture. I ran the required control-transport-dial suite and its shipped mutation fixture, both rail mutation fixtures, isolated reviewer mutations for the structured field and two-cause wording, typecheck, docs-bundle validation, cli-kernel, runtime-run-command-usage, ep-rail-failure, and bind-fence-live.
- I did not execute the expensive legacy-packaged-manager acceptance fixture. I read its real packaged update workflow through completion and ran the focused command kernel instead.
- Plain `pnpm changeset status` could not resolve a missing local `main` branch. `pnpm changeset status --since 08cff3533d08aedaad3faa500cc265f5948a110c --verbose` passed and showed the intended minor fixed-group release.

---

### Review 3: `APPROVE`

**Blocking (0):** none

**Body defects (1):**
- On the legacy rail every message is byte-identical to before: there is no second rail its silence could be hiding a manager on. In the PR body fetched at 2026-09-15T17:50:53Z, the author states: 'On the legacy rail every message is byte-identical to before: there is no second rail its silence could be hiding a manager on.' Similarly, .changeset/ep-rail-scoped-unanswered-verdict.md states: 'On the legacy rail every message is unchanged: there is no second rail its silence could be hiding a manager on.' This claim is inaccurate: packages/core/src/endpoint-invoke.ts line 168 unconditionally appends ' on the ${rail} rail' to the describe [...]

**Non-blocking (2):**
- Changeset repeats the legacy unchanged overstatement. .changeset/ep-rail-scoped-unanswered-verdict.md asserts 'On the legacy rail every message is unchanged'. The nested core describe deadline error gained ' on the ep rail'. This is an additive clause in operator-facing diagnostics and does not break callers, but the changeset wording overstates byte identity.
- Lockstep minor changeset bump includes connector-core without runtime changes. The changeset lists @cotal-ai/connector-core with a minor bump, but the only change in connector-core was the regenerated docs-bundle.generated.ts. This is consistent with repository lockstep versioning and harmless.

**Not verified (0):** none

davidfarah2003:
Merge basis for `31f005e38b1c759efec1b2ca725dc0c637fc9a1c`.

Three independent reviews at this head approve with no blocking items, and the review that requested changes at the previous head confirmed its item resolved here. CI at this head is terminal green. The one finding all three reviews share is a body sentence: on the legacy rail the core describe deadline text now carries the rail name, so "every message is byte-identical to before" overstates it, and the changeset repeats that line. That is a description defect and the squash message states the legacy change as it is. No code change was asked for.

Merging by squash closes #1630, which this change fixes.

PR #1641, opened 2026-09-15
fix(cli): refuse to re-exec a detached daemon from an entry that is not the CLI

### LINKED ISSUE #1629: delivery-boot-honesty smoke re-execs itself through startManagerDetached on a persistent host
## What happened

One hand run of `implementations/cli/smoke/delivery-boot-honesty.smoke.ts` on a persistent Linux host (checkout at 049e265fd, the #1594 head as merged) produced 970 further runs of the same suite over 4.7 hours, about 206 an hour, until the chain was killed by pid. Each generation was a detached process reparented to init, with its working directory in its own `cotal-boot-honesty-root-*` scratch directory under the temp dir and its stdout on `.cotal/manager.<spaceKey>.log` inside that directory. Each log held the complete 15-cell output of the suite followed by the sentinel.

## Why

The suite calls `ensureControlPlane` (its cell "ensureControlPlane PROPAGATES responderBound=false to its caller"). On main that function runs `ensureDelivery` and then `ensureManager`, and `ensureManager` reaches `startManagerDetached`, which builds the child's argv with `selfArgv()`:

```ts
return [process.execPath, ...process.execArgv, process.argv[1]];
```

Under tsx, `process.argv[1]` is the smoke file, so the detached "manager" is the smoke itself with `supervise --space boot-honesty ...` appended to an argv the smoke never reads. It runs the whole suite, reaches the same call, and spawns the next generation. The suite's own comment says the manager "is NOT started here" and "may well fail in this fixture"; on this host it did not fail.

The process tree at the moment of capture, one generation per line, parent pid on the left:

```
tsx (ppid 1, cwd <tmp>/cotal-boot-honesty-root-x5kq2k)   the previous generation
  tsx (cwd <tmp>/cotal-boot-honesty-root-xLYreT)         this generation, stdout -> .cotal/manager.<hex>.log (deleted)
    nats-server -c <tmp>/cotal-boot-honesty-root-xLYreT/server.conf
    node .../delivery-responder-holder.mjs nats://127.0.0.1:<port> boot-honesty claim ...
```

A single bounded run on macOS at e6a831e36 (main plus #1626) did not chain: the parent exited 0 with 15 cells green and no survivor under the private temp dir, and the scratch root's `.cotal` held `auth`, `delivery.pid` and the delivery-aware marker but no manager pid or log. Which condition makes `ensureManager` spawn on one host and not the other is not established here.

## Why it matters

Any host that outlives one run of this suite becomes a fork chain with a delay: about 1.5 of load for as long as nobody notices, plus one nats-server and one holder per generation for the generation's lifetime. On an ephemeral CI runner the chain dies with the job, but it competes with every later suite in the same shard while the job lasts, which is a plausible contributor to the shard timing reds this repo attributes to "runner timing".

## Ask

1. A suite that calls `ensureControlPlane` or `ensureManager` in a fixture must not be able to re-exec itself. Either the fixture passes a flag that makes `ensureManager` refuse to spawn (and the cell asserts the refusal is reported, not swallowed), or `startManagerDetached` refuses when `process.argv[1]` is not the CLI entry it expects (a `.smoke.ts` path is never a valid `supervise` entry).
2. A cell that reds when the suite's own pid tree grows after the suite exits: count processes whose cwd is under the suite's scratch root after teardown and assert zero.
3. `smoke-kit`'s `teardownOnSignal` cannot own a child that `startManagerDetached` unrefs; the guard has to be at the spawn site.

## Evidence

- 970 scratch roots created between 23:33Z and 02:15Z, 206 per hour, all owned by the same user, every one empty after its generation was killed.
- Every surviving generation's stdout fd resolved to a deleted `manager.<hex>.log` inside its own scratch root, and the log tail read `DELIVERY-BOOT-HONESTY SMOKE OK (15 passed, 0 failed)` followed by the sentinel and `smoke broker teardown: pid <n> did not exit before path cleanup`.
- Killed by explicit pid (census by cwd prefix), three TERM/KILL rounds; no new generation in the ten minutes after.

L4XB:
Closes #1629

## What was wrong

`selfArgv()` builds the argv every detached re-exec is spawned with:

```ts
return [process.execPath, ...process.execArgv, process.argv[1]];
```

Each caller appends a cotal subcommand to it (`supervise`, `deliver`, `ext add`) and detaches the
child, so the entry has to be the file that reads a subcommand. It took `process.argv[1]` on trust.
Under tsx that is whatever file was run, so a process started from anything but the `cotal` entry
spawned a child that re-ran THAT file with the subcommand appended, which the file does not read.

`implementations/cli/smoke/delivery-boot-honesty.smoke.ts` calls `ensureControlPlane` in a fixture,
so its detached "manager" was the suite. The copy ran to the same call and spawned the next: 970
generations over 4.7 hours on a persistent host, each with a nats-server and a delivery holder.
The suite's own teardown cannot reach those children, because `startManagerDetached` unrefs on
purpose; that is why the guard is at the site that builds the argv rather than in the smoke harness.

## The change

`selfArgv()` refuses unless `basename(process.argv[1])` is the CLI's composition root. Three shapes
are accepted, and the third is the one that made an extension-only rule wrong:

| shape | `process.argv[1]` |
|---|---|
| source checkout (`pnpm cotal`) | `<repo>/bin/cotal.ts` |
| published install / npx | `…/cotal-ai/dist/cotal.js` |
| `npm i -g cotal-ai` | `<prefix>/bin/cotal`, a **symlink**, no extension |

The third is what an installed binary actually runs under, and `seed.smoke.ts` already covers it.
A guard that demanded an extension would refuse every global install's manager start while still
passing the headline cell, so both directions have a mutation.

The refusal is a throw, not a skipped spawn: a re-exec that silently does not happen reports a
healthy control plane over nothing. It names the entry, what a child spawned from it would run, and
the remedy.

The manager, delivery and auth starters now build the argv **before** opening the daemon logfile, so
a refused start leaves the mesh root as it found it instead of creating `manager.<key>.log` and
leaking its descriptor.

`delivery-boot-honesty` no longer reaches a spawn at all. It records a live manager (a real `cotal
up` path: a second `up` on a live root), which is what `ensureManager` short-circuits on, so the
`responderBound` propagation cell the suite exists for is graded with nothing re-execed. It also
grades the refusal itself and asserts the fixture root stays empty.

## Evidence

A bounded reproduction, from a file that is not the `cotal` entry, calling the real
`startManagerDetached` once against a scratch root. The child prints one line and exits, so exactly
one generation exists.

Before:

```
PARENT: startManagerDetached returned pid 91571
PARENT: manager log: "CHILD: I am …/zz-reexec-1629.mts pid 91571, argv tail=[\"supervise\",\"--space\",\"reprospace\",\"--server\",\"nats://127.0.0.1:4222\"]"
PARENT: manager log exists after the call? true
```

The detached "manager" is the calling file, re-run with `supervise …` appended. After:

```
PARENT: startManagerDetached REFUSED: refusing to re-exec this process as `cotal`: it was started from …/zz-reexec-1629.mts, which is not the CLI's own entry (`bin/cotal.ts` in a checkout, `dist/cotal.js` in an install).
PARENT: manager log exists after the call? false
```

The suite the issue names, at this head, spawning nothing:

```
✓ FIXTURE: the recorded manager reads ALIVE, so the composition needs no spawn
✓ CELL: ensureControlPlane PROPAGATES responderBound=false to its caller (`cotal up`)
✓ CELL: the composition launched NOTHING into the fixture root (no detached manager log)
✓ CELL: with no manager recorded, the spawn REFUSES rather than re-execing this suite
✓ CELL: and the refusal names this suite as the entry it will not re-exec
✓ CELL: the refused spawn left no manager log behind either
DELIVERY-BOOT-HONESTY SMOKE OK ✅  (20 passed, 0 failed)
```

`install.smoke.ts` pins the three accepted shapes and the refusals off the wire, and all three
mutants are killed:

```
KILLED  the guard is removed and any entry builds a re-exec argv again
KILLED  the entry must carry an extension, so a global install's bin symlink is refused
KILLED  the log is created before the argv is validated, so a refused start still leaves it behind
```

The two suites that reach a real re-exec are unaffected, which is the point of the scoping:
`delivery-explicit-space` already sets `process.argv[1]` to `bin/cotal.ts` before calling
`ensureControlPlane` (7/7 green), and `seed.smoke.ts` drives the CLI as a child, including the
global-install symlink shape (71/71 green). `pnpm smoke:install`, `pnpm smoke:split-topology-docs`
and `pnpm --filter @cotal-ai/workspace test` are green, and `pnpm --filter @cotal-ai/cli typecheck`
is clean.

## Not done here

The issue also asks for a cell that reds when the suite's pid tree grows after teardown. What is
here instead is deterministic and cheaper: the detached starter opens `manager.<key>.log` before it
spawns, so the log's absence after the call is evidence no child was launched, with no race against
the spawn. A process census would have to sample after an unbounded delay to say the same thing.

`ensureManager` gained no refusal flag. With the guard at the argv builder, a fixture cannot re-exec
itself whether or not it remembers to pass one, and the flag would only cover the callers that
thought to.

File and line references are against `main` @ `08cff3533`.


davidfarah2003:
Panel complete: three independent reviews of `5504eaff59a25f57c654238532af3a2827fd0e94`.

Each review graded the merge of this head onto main in its own lane (merge tree `e760f02c3897153b070cb540f3b6b13ff7846ba0`), so the figures include the main-side change that landed while this pull request was open. One review approves with zero findings: the named reproduction set passes completely in its lane (install smoke twice, the delivery boot honesty suite, both new mutation fixtures with every mutant killed). Two reviews request changes, on different primary grounds, and neither disputes that the guard closes the reported spawn chain at the site that builds the argv; one review drove the chain to form with the guard reverted in a built copy and saw it refused at the head.

The first request-changes review reports a defect the change introduces in two callers it touched but did not correct. The pull request says the callers compute the argv before they open the daemon logfile so a refused start leaves the mesh root as it found it. Measured at this head that holds for the manager and delivery starters and not for the auth starter or the connector seed: the auth starter claims its pid slot, pre-populated with the launcher's own pid, before the guard throws, so a refused start leaves a pidfile that reads alive, a same-process retry then reports a healthy auth service over no service, and teardown signals whichever process holds that pid; the seed writes its pending child marker before the guard and the clear after the spawn is never reached, so a refusal wedges the next boot. The review records that its second finding was a hypothesis in its first revision and was then driven to a live failure with a control run before it was raised to blocking.

The second request-changes review reports that the entry test compares only the basename of the process's entry file against the CLI's names, so an unrelated script named cotal.ts or cotal.js in any directory is accepted and can reproduce the same detached recursion; it showed this with the repository's own mutation tool by replacing the install smoke's negative input with a sibling path named cotal.js, which turned the named cell red. It asks for a check that establishes the entry is one of the real CLI entry shapes, including the resolved global symlink, plus a shipped negative cell for that path and mutation grading of the false-accept direction. It records the auth starter's pidfile ordering as a non-blocking finding, so the two request-changes reviews agree on that mechanism and differ on which item blocks.

The census cell red on the auth smoke suite at these merge trees is inherited from main's tip at the time of the merge and is not attributed to this head; it is fixed on main since. What would change the verdicts, in the reviewers' words, is under each review below.

---

**Blocking (2):**
- F1. The auth starter claims its pid slot BEFORE the guard, so a refused start leaves a live-reading auth-service pidfile in the mesh root. implementations/cli/src/lib/auth-proc.ts, startAuthServiceDetached. The function calls reclaimDeadLegacyPid(space) and then claimAuthPidSlot(space) BEFORE entering the try that contains selfArgv(). claimAuthPidSlot publishes the slot atomically and PRE-POPULATED with the LAUNCHER's own pid (writeSync(fd, String(process.pid))). Only afterwards does selfArgv() throw. The finally closes the fd, but nothing removes the published pidfile, so the refusal leaves a record naming a process that is alive and is not an auth service. Moving the selfArgv() call above openSync fixed the LOG, which is what the PR measured, but the pidfile is claimed earlier still and was not moved. Driving the real ensureAuthService from a non-cotal entry: it THREW the expected [...]
- F3. A refusal inside the connector seed leaves a PENDING child marker that wedges the next boot. implementations/cli/src/seed/reconcile.ts, seedOne. seedOne writes a crash cursor (phase 'copy', then 'add'), stages the payload, and writes a PENDING child marker via writePendingChildMarker(nonce), all BEFORE `const [bin, ...argv] = selfArgv();`. clearChildMarker() sits AFTER the spawnSync, so when the guard throws, the marker is never cleared. This is the same before/after ordering the PR corrected in manager-proc.ts and delivery-proc.ts and did not correct here. Isolated with XDG_CONFIG_HOME pointed at a scratch dir and COTAL_ALLOW_CHECKOUT_SEED=1, from an out-of-lane package root so the lane tree stayed clean. reconcileSeededConnectors() from a non-cotal entry THREW the expected refusal and left the seed dir holding reconcile.cursor.json (phase 'add', package 'claude') AND [...]

**Non-blocking (4):**
- F2. The side-effects fixture leaves dist/ compiled from its mutant. Fully described under q4. Reproduced 3 times; once it made the next run of the same fixture refuse with a red baseline on the PR's own new cell. Remedy is a one-line afterRestore.
- F6. A refused manager or delivery start still runs the pre-upgrade reclaim first, so the refusal is not side-effect-free. implementations/cli/src/lib/manager-proc.ts startManagerDetached, and implementations/cli/src/lib/delivery-proc.ts startDeliveryDetached. Both call reclaimDeadPreUpgradeRecord(...) BEFORE `const [node, ...self] = selfArgv();`. Seeded a root with the three pre-upgrade spellings (manager.pid, manager.delivery-aware, delivery.pid) holding dead pids: a refused manager start deleted two and threw, a refused delivery start deleted the third and threw. Root went from three records to zero. So the strict reading of 'leaves the mesh root exactly as it found it' is not satisfied.
- F4. The refusal's remedy line omits the fourth guarded entry. The message names ensureManager, ensureDelivery and ensureControlPlane. ensureAuthService is also behind the guard and is not named.
- F5. The changeset and commit message overstate the invariant. Both say the manager, delivery AND auth starters leave the mesh root exactly as they found it on a refusal. Measured, that holds for two of the three. Reported as a finding; not edited.

**Not verified (3):**
- pnpm changeset status. 'Failed to find where HEAD diverged from main'. The lane is a detached merge with no local main branch. Invocation limitation of the lane, not a defect.
- node scripts/check-attribution.mjs with no arguments. printed its usage line and required --range. Re-invoked correctly over the branch range, which passed.
- Inherited from main, not attributed to this head. implementations/auth/smoke/issued-profile-census.smoke.ts INHERITED FROM MAIN. Not a defect in this head.

---

**Blocking (1):**
- The basename allowlist still accepts non-CLI entries and leaves the repeated-spawn condition possible. implementations/cli/src/lib/self-exec.ts:14,34-42. selfArgv reads process.argv[1], takes only basename(entry), and returns the re-exec argv whenever that basename matches /^cotal(?:\.(?:ts|js|mjs|cjs))?$/. It does not resolve the path or establish that the file is bin/cotal.ts, the installed package's dist/cotal.js, or a symlink to that installed entry. Consequently /srv/fixture/cotal.ts, <tmp> and an unrelated executable named cotal pass. If such a file reaches ensureManager, ensureDelivery, or ensureControlPlane on load, the detached child re-runs that same non-CLI file with a subcommand appended. The condition described in issue 1629 therefore remains for this allowed-basename class.

**Non-blocking (2):**
- A refused auth-service re-exec still mutates the mesh root before validating the entry. implementations/cli/src/lib/auth-proc.ts:112-122,152-181; .changeset/refuse-reexec-from-a-non-cli-entry.md. startAuthServiceDetached calls claimAuthPidSlot before selfArgv. claimAuthPidSlot creates a temp file, writes the launcher pid, atomically links it as the canonical auth pidfile, removes the temp name, and returns the open descriptor. selfArgv then throws for a non-CLI entry, and finally closes the descriptor without removing the published pidfile. The refused process remains alive, so a subsequent auth start adopts that pidfile as a live holder instead of retrying the daemon launch.
- The changeset includes incident narrative far beyond the user-facing release summary. .changeset/refuse-reexec-from-a-non-cli-entry.md.

**Not verified (3):**
- I did not run the full smoke:ci suite. I ran the packet-prescribed suites and fixtures, build, typecheck, the inherited census cell, the real source-entry detached-start smoke, and auth pid-slot smoke.
- I did not run the networked seed-tarball live suite. I inspected its shipped use of the actual npm .bin/cotal symlink, while the supplied install smoke directly graded the bare symlink argv shape and killed the extension-only mutant.
- The repository exposes no public direct test seam for startAuthServiceDetached. The auth refusal pidfile aftermath is established by source ordering plus the shipped auth-pidfile observation that claimAuthPidSlot immediately publishes the launcher's live pid. No bespoke harness was created.

---

### Review 3: `APPROVE`

**Blocking (0):** none

**Non-blocking (0):** none

**Not verified (1):**
- .

davidfarah2003:
Merge basis for `5504eaff59a25f57c654238532af3a2827fd0e94`: not met. Two of three independent reviews request changes and one approves; the request-changes items are the auth starter claiming its pid slot before the entry guard can refuse (a refused start leaves a live-reading pidfile), the connector seed writing its pending child marker before the guard (a refusal wedges the next boot), and the entry test comparing only the basename so an unrelated file named cotal.ts or cotal.js is accepted. What would change each verdict is stated in the reviewers' own words above. This head stays open for the contributor; nothing here is pushed to the branch by anyone else.

L4XB:
Thanks for the three reviews. #1641 was merged before I had addressed them, so the follow-up is #1691. Every item was reproduced first on `438c6296` and is fixed in #1691. Each code defect has a cell that is red at that base and a mutant killed on it. F2 is checked by the `dist/` hash, and F5 is a text fix.

- **Entry test by basename (review 2).** `selfArgv()` now resolves `process.argv[1]` through symlinks. It accepts the entry only if the resolved file is the `cotal-ai` manifest's `bin.cotal` or the `cotal.ts` beside that manifest.
  - Reproduced: a foreign `<scratch>/foreign/cotal.ts` was re-execed as its own `supervise` child.
  - The install smoke now has negative cells for foreign `cotal.ts`, `cotal.js` and bare `cotal` files, an impostor manifest, a non-bin file inside the package, a missing path, and a symlink to a foreign file.
  - Eight mutants grade both directions: false accept and false refuse.
- **F1, auth pid slot before the guard.** Reproduced: after the refusal, the slot held the launcher's pid and `authServiceUp` read true. The retry started nothing, and teardown would have sent SIGTERM to the launcher's pid (the probe recorded signals instead of sending them).
  - One nuance: the retry "reports healthy" only because my provider `ready()` was a stub that answers at once. A real provider would wait there with no daemon behind it.
  - The check now runs before the reclaim and the claim.
- **F3, seed journal before the guard.** Reproduced, including through the real entry: `tsx bin/cotal.ts ext list` exited 1 with "a connector seed may be mid-flight".
  - The check now runs first. The same `ext list` then seeds all seven extensions.
- **F6, pre-upgrade reclaim before the guard.** Reproduced: three dead records went to zero. The check now runs first in both starters, and each record kind has its own cell.
- **F2, mutant `dist/`.** Reproduced as a stale `dist/lib/manager-proc.js` after the proof, with a clean source tree. `afterRestore` now rebuilds, and the hash returns to the clean build's.
  - I did not see the red-baseline run you describe. The suite command rebuilds before its baseline, so on my runs the stale `dist/` only lasted until the next build.
- **F4.** The remedy line names `ensureAuthService`.
- **F5.** The #1641 changeset still sat unreleased in `.changeset/`. It now claims only what #1641 fixed, that each starter builds the argv before it opens its log. It no longer says a refusal leaves the root as it found it, and the narrative is gone. The new changeset states the rest.

`smoke:gate-inventory` fails on that branch for reasons it also has at `438c6296`: two suites are run by nothing, and one suite is listed twice in `ci-suites.txt`. All measurements were on Node 24.2.0; I have no Node 22 here.

PR #1447, opened 2026-09-10
fix(manager): prove one SecretStore authority before daemon credential renewal

davidfarah2003:
Refs #773.

The manager re-signs the standing daemon credentials through **its own** `SecretStore` and then asks the delivery daemon to reload from **the daemon's own** store, sending only the expected fingerprints — no credential bytes and no shared-store coordinate. When those two stores are not the same authority (the stock cross-host composition), the manager writes filesystem A, the daemon re-reads filesystem B, and every adoption is refused on the generation compare. Nothing refuses the divergent composition at startup, so the failure is delayed until the 24-hour JWT window is already burning.

## The change

One store authority, proven at construction time rather than discovered at renewal time.

- `packages/core/src/secret-store.ts` gains `SecretStoreIdentity` — a way for a process to *name* the store it treats as the standing-daemon credential authority. It names the store, never a secret. Two filesystem stores agree only when they resolve the same directory; an injected (hosted) store is identified by an operator-supplied coordinate and is never guessed from a local root. There is no fallback between the two shapes, and `parseSecretStoreIdentity` refuses unknown fields, extra keys, blank values and mixed shapes rather than guessing.
- `implementations/manager/src/manager.ts` challenges the daemon's `reloadStoreIdentity` on the delivery-admin rail **before the first remint**, and refuses when the two stores diverge. An injected store additionally requires `COTAL_SECRET_STORE`, so a hosted manager cannot silently fall back to a local root.
- `implementations/delivery/src/delivery.ts` answers that challenge with the identity of the store its reload actually resolves.
- Fingerprint-only `reloadCreds` is **kept**. The fingerprint was never the defect; the divergent read was. Once both sides provably name one authority, a fingerprint is sufficient proof of generation.

The refusal names **both** stores. A refusal that declines to start without naming them is the same defect wearing a different error, so that is asserted on the message text rather than on the fact of refusing.

## Reading the test signal — please read this before judging CI

`bin/smoke/manager-two-root-renewal.smoke.ts` is an **inverted** suite. It was landed by #1185 (merged 2026-08-31) as a deliberate reproduction of this issue, and its cells encoded the **defective** behaviour as the expectation, labelled `#773 unfixed:`. **Those cells were green precisely because the bug was present.** This PR flips them to the fixed expectations, so:

- the divergent composition is now refused at start, with both roots in the message;
- the refused start **never remints** — root A's credential still holds its original generation, so there is no half-applied renewal to reason about;
- root B's delivery and membership credentials are untouched;
[...]

PR #1642, opened 2026-09-15
fix(manager): let a space carry a manager on more than one workspace root

### LINKED ISSUE #1634: A manager on any second workspace root is refused at start since #1447, so a space cannot have more than one manager
## What happens

A space is meant to be served by more than one manager, on the same device or on different devices. Since #1447 (b00f3c1d2, merged 2026-09-11, closing #773) that is impossible on an auth-mode space.

`Manager.start()` calls `assertDaemonSharesSecretStore()` (`implementations/manager/src/manager.ts`, the call at the start of the renewal block and again inside every `renewDaemonCreds()` pass). That method asks the space's delivery daemon which SecretStore it reloads from and refuses unless `sameSecretStoreIdentity()` holds (`packages/core/src/secret-store.ts`). For a filesystem store the identity is the resolved workspace root. Two roots on two machines never compare equal, and two roots on one machine do not either.

So on any auth-mode space whose delivery daemon runs from one root, a manager started from any other root refuses at construction. A participant machine cannot run a manager for a space it joined; a second manager on the same machine from a different checkout cannot either.

## Reproduction

On 0.49.0, with a broker and a delivery daemon running for an auth-mode space from workspace root A:

```
cotal supervise --space <space>        # from workspace root B, same broker, same creds
```

refuses, naming both roots. `cotal up --detach` from root B for the same hand-registered space refuses as well and points at `meshes rm`.

The repository asserts the refusal as correct behaviour: `bin/smoke/manager-two-root-renewal.smoke.ts` stands up two roots and grades "daemon B outlives the refused start (refusal is a manager construction error, not a daemon death)" green, and grades the unified-root path as the only one that adopts.

## Why this is a regression and not a fix

#773 diagnosed the real defect: the `reloadCreds` request carries only fingerprints, "not the re-signed credentials or a shared-store coordinate", so a manager renewing one store while the daemon reloads another had every adoption refused. #1447 closed that issue by forbidding the composition instead of carrying the bytes or a coordinate. The failure mode is gone because the deployment is gone.

## What is required

1. A manager on any host, with its own workspace root, joins an auth-mode space and serves seats.
2. Daemon credential renewal has exactly one owner per space at a time, chosen by lease rather than by being the only manager. A manager that does not hold the renewal lease never remints daemon credentials and is never challenged on store identity. The store-identity proof from #1447 stays, scoped to the lease holder.
3. The renewal owner hands the daemon the re-signed credential bytes over the delivery-admin rail (authenticated and ownership-checked), or a shared coordinate, so the daemon's store and the owner's store no longer need to be one directory. Fingerprint-only reload stays valid when the identities agree; byte-carrying reload is the cross-host path.
4. `cotal up` on a participant machine for a hand-registered remote space either starts a manager for that space or says exactly which command does.
5. Two managers on one device with different roots is the same case and must work.
6. `bin/smoke/manager-two-root-renewal.smoke.ts` is re-inverted: the two-root composition is accepted, the non-owner never remints, the owner's byte-carrying reload is adopted, and the divergent-fingerprint refusal is kept for the same-identity path.

## Out of scope

Hosted injected stores keep the coordinate path unchanged. No change to seat credentials or agent-kind stores.

L4XB:
Closes #1634.

`Manager.start()` throws on any auth-mode space whose delivery daemon reloads from a workspace root
other than the manager's own, naming both roots. A participant machine cannot run a manager for a
space it joined, and a second manager on one machine from a different checkout cannot start either,
so an auth space is capped at one manager.

That cap contradicts a decision the code states in two places. `packages/core/src/lease.ts:58`
records that P2 item 3 "demoted the per-space singleton to per-instance liveness, so a second
manager's create no longer THROWS", and `implementations/manager/src/manager.ts:1384` says a second
manager in a second workspace root "has a distinct id ⇒ a distinct key ⇒ it coexists". The store
check reinstates the singleton for every space with auth.

## What #1447 guarantees, and why it is worth keeping

#773 was not a false alarm and this change does not undo it. `reloadCreds` transports the expected
per-component fingerprint and nothing else: no credential bytes, no shared-store coordinate. So a
manager that re-signs into a store the delivery daemon never reads has every adoption refused, for
as long as the daemon's 24 hour JWT window lasts, with both files structurally healthy the whole
time. #1447 added the `reloadStoreIdentity` challenge to make that composition impossible to build.

The property is therefore: **no manager remints a daemon credential into a store the daemon does not
read.** Path equality is not a weak proxy for that property. Two unshared filesystem roots really are
two stores, and the fingerprint really cannot match across them, so a looser identity would be a
false same-store proof of the kind #1447's own description says it had to fight off three times. The
identity test is right. What is wrong is who it is applied to.

## The fix

The store identity selects **which manager owns the daemon's credential renewal** instead of deciding
whether a manager may run. The daemon names one store, so at most one manager can match it, and only
that manager remints. A manager on any other root starts, serves its own agents, renews its own
supervisor, goal-writer, session-ledger, hosted-run and managed-agent credentials, and remints no
daemon credential.

- `assertDaemonSharesSecretStore` (`implementations/manager/src/manager.ts:1488`) returns
  `shared` / `absent` / `foreign` rather than throwing on divergence.
- The daemon half of the renewal pass moves into `remintDaemonCredsIfOwner`, gated on that verdict.
  The rest of `renewDaemonCreds` is this manager's own and runs on every manager of the space. On
  `main` a divergence throw at the top of the pass also skipped the manager's own serve, goal-writer
  and session-ledger renewals, because the whole pass is one `try`.
- A non-owner writes no `renewal.json`. Nothing was renewed there, and a record saying so renders in
  `cotal doctor auth` as a pass that happened. It logs the notice on **every** pass instead, naming
  both stores and stating that it wrote nothing, so an operator whose space has no manager on the
  daemon's own root keeps being told rather than being told once at a start that no longer fails.
- `divergentSecretStoreRefusal` becomes `foreignDaemonSecretStoreNotice`. It still names both stores;
  a function called `Refusal` that no longer refuses would be worse than the message it carries.

Refusing that start bought nothing operationally. It renewed nothing: the daemon's own root is just
as unrenewed with the second manager dead as with it running. It only removed the second manager.

**Fail-closed is unchanged everywhere it was load bearing.** A delivery-admin rail that hangs (a
timeout, deliberately not treated as absence), a daemon that will not name its store, and an
unparseable identity all still end the start. An unbound daemon is still not a named store, so remint
proceeds into this manager's store, and a daemon that binds later on a foreign store stops the remint
from the next pass rather than being certified by the earlier absence. Nothing in
`sameSecretStoreIdentity` (`packages/core/src/secret-store.ts:71`), in the daemon's answer, or in the
delivery-side cwd guard changed.

## Evidence

`bin/smoke/manager-two-root-renewal.smoke.ts` is the graded suite: a real `Manager`, a real
`tsx bin/cotal.ts deliver` daemon, a real authed broker, two unshared roots. The **same suite file**
was run against the merge base and against this head.

| | merge base `712dfadf` | this branch |
|---|---|---|
| `smoke:manager-two-root-renewal` | 21 passed, **2 failed** | **23 passed, 0 failed** |

The two cells that fail at the base are the two this change exists for:

```
✗ FAIL: #1634: a manager on a SECOND workspace root starts against a daemon rooted elsewhere -
  {"startRefusal":"daemon credential renewal cannot be constructed across two SecretStores: the
   manager remints through <rootA> while the delivery daemon reloads from <rootB>. ..."}
✗ FAIL: #1634: it reports that it is not the renewal owner, naming both roots and its own inaction
```

**The #773 guard cells are unchanged and pass on both sides.** They are not made green by this
change; they assert what they always asserted:

```
✓ #773: the non-owner never reminted (no manager renewal record, no write into A)
✓ #773: root A's delivery cred still holds the ORIGINAL generation (remint did not run)
✓ #773: root B's delivery cred still holds the ORIGINAL generation
✓ #773: root B's membership rw cred still holds the ORIGINAL generation
```

Two cells are new on the other side of the question, and they are what proves one owner per space is
selected rather than merely being the only survivor. A **second** manager, co-rooted with the daemon,
starts while the non-owner is already live in the same space, and its pass adopts:

```
✓ #1634: the co-rooted manager starts too, so ONE space carries TWO managers on TWO roots
✓ #1634: the OWNER's pass adopts (adoption.ok:true) with the non-owner live in the same space
✓ #1634: root B's delivery cred now holds the owner's re-signed generation
✓ #773: root A is STILL the original generation after the owner's full adopted pass
```

Both managers register their service endpoints and deregister cleanly. The unified-root control
phase is untouched and still adopts, broker-accepted and pinned to each daemon's own nkey.

### Every suite that names the sources this touches, at this head

`implementations/manager/src/manager.ts` and `packages/core/src/secret-store.ts`. All green:

| suite | result |
|---|---|
| `smoke:secret-store-identity` | 16 passed |
| `smoke:reload-store-identity` | 32 passed |
| `smoke:secret-store-seam` (injected path) | 9 passed, 0 failed |
| `smoke:delivery-renewal` | 18 passed, 0 failed |
| `smoke:manager-supervise-restart` | 21 passed, 0 failed |
| `smoke:run-host-live` | 51 passed, 0 failed |
| `smoke:turn-relay-auth` | 13 passed, 0 failed |
| `smoke:boot-self-heal-gate` | 12 passed, 0 failed |
| `smoke:manager-restart-live` | 7 passed, 0 failed |
| `smoke:lifecycle-e2e` | 30 passed, 0 failed |
| `smoke:manager-reconcile-startup` | 25 passed, 0 failed |

`smoke:manager-supervise-restart` and `smoke:run-host-live` are the two suites #1447 had to repair by
unbinding the daemon around a user-mode cell. They still unbind, and they still pass; this change did
not need them touched.

No suite in the tree asserts the old refusal text, so the rename cannot silently pass a stale
assertion. `pnpm typecheck` is clean across every package, including the smoke configs.

### Two results that are the same before and after, stated rather than omitted

`pnpm smoke:adoption-passive-preflight` reports `0/4` on the machine this was measured on, and it
reports `0/4` at the clean merge base as well. `pnpm mutation-coverage` exits non-zero with
`enumerated=437 examined=437 graded=0 refused-with-reason=151` at both, byte-identical, against files
this change does not touch. Neither is caused by this change and neither is claimed as passing.

### One rig fix in the suite

The suite's roots are now `realpathSync`ed. A store identity is compared as a path, and the daemon's
side of that comparison arrives through `process.cwd()`, which is already resolved. On a platform
whose tmpdir is reached through a symlink the unresolved `mkdtemp` path makes one directory look like
two, and the daemon refuses at boot on the delivery-side cwd guard before any cell can grade
anything. It changes nothing about what is graded: the roots are still distinct directories.

Noticed while doing that and **not** fixed here: `assertUninjectedCredsSharesCwdRoot` compares a
`resolve`d `--creds` path against a `findCotalRoot()` that starts from an already-resolved
`process.cwd()`, so a `--creds` path whose prefix is a symlink refuses at start on a genuinely
single-root composition. That is #1447's cwd guard, not the manager challenge, and it deserves its
own issue rather than a drive-by.

## What this does not do

Requirements 1, 2 and 5 of the issue are what this delivers, and 2 in a narrower shape than asked.

- **Not a negotiated lease.** The issue asks for renewal ownership "chosen by lease". Here it is
  derived from the store identity the daemon already publishes. That is a single deterministic owner
  per space with no new distributed state, and it is sufficient while reload is fingerprint-only,
  because only the co-rooted manager can produce bytes the daemon will accept. A negotiated lease
  becomes necessary once requirement 3 lands and several managers are capable of renewing.
- **No byte-carrying reload (requirement 3).** Putting a re-signed `.creds` file, which contains a
  private nkey seed, onto the delivery-admin rail is a security decision for maintainers, and #773's
  own acceptance list requires that no credential or seed lands in an unscoped message. Until it
  exists, a space whose daemon runs on a root with no manager on it still has no renewal owner. That
  is unchanged by this PR, and the per-pass notice is what makes it visible.
- **No `cotal up` guidance change (requirement 4).**
- **Requirement 6 partially.** The suite is re-inverted for what this change makes true. The
  divergent-fingerprint refusal on the same-identity path is untouched, and the "non-owner never
  remints" assertion is now graded from the non-owner's side rather than from a start that failed.
- Requirement 1 is graded as "starts, registers its service endpoint, holds its own instance lease
  alongside another manager". A spawned seat on the second root is not graded by this suite.

`SPEC.md` and `docs/embedding.md` are updated in the same change, and the docs bundle is regenerated
(`pnpm check:docsbundle` is clean). `docs/embedding.md` already said the manager "refuses to remint"
(`docs/embedding.md:232`); it was `SPEC.md:4161` and the code that read as a refusal to run. A
changeset is included, `@cotal-ai/core` and `@cotal-ai/manager` at minor.

File and line references are against `main` @ `712dfadfe1e4593810614303170aa876ad3a0d30`.


davidfarah2003:
Panel complete: three independent reviews of `8e346f558625799d00b814836d7e970b1c9726d0`.

All three reviews request changes, reached without seeing one another, and they converge on the same two mechanisms. Each review reproduced both on this head against a real authenticated broker with the shipped built manager, with positive controls in the same run.

**B1, the absence verdict is reachable from a reply.** The predicate at `implementations/manager/src/manager.ts:903-907` tests an error message string against a regex, and it is consumed at line 1510 inside the store challenge. The JSON parse of the reply happens inside the awaited request, so a parse failure throws into the same catch as a transport absence. A bound responder that answers with the literal text `no responders`, or with an empty payload carrying a 503 status header, is therefore classified as absent. The orchestrator ran the positive control itself: parsing that literal text throws a message that tests true against the production regex. At the merge base the absent branch was one of two ways to reach a daemon remint from a manager that is not co-rooted with the daemon; on this head the foreign verdict is deliberately non-fatal, so the absent branch is the only remaining way, and it contradicts the reasoning that only the co-rooted manager can produce bytes the daemon accepts. The predicate is unchanged since the merge base; what this head changes is its consequence. One review also drove the case where a genuine daemon is bound and answering honestly with a foreign root while a second holder of the same delivery profile joins the same queue group the shipped serve path uses and answers with an empty payload and a 503 header, run twice: the control arm with the daemon alone never reminted across nine passes in either run, and the arm with both bound reminted on three of nine passes in one run and two of nine in the other, wrote the renewal record and changed the delivery credential. Because the rail is queue-grouped the challenge is load-balanced between the two members, so the outcome is per pass rather than absolute. The same review states the boundary of the finding in its own words: it minted the responder's credential from the space's own signer, which it calls the strongest assumption in the finding; it established positively that a supervisor-profile credential is refused by the broker at subscribe time, so the broker grant is a real gate and an ordinary agent cannot reach this rail; it did not establish how a second holder of the delivery profile arises in a real deployment, and the severity of B1 depends on that question.

**B2, the reply carries no provenance.** The store identity comparison is structural only: kind, then a trimmed root path or a coordinate string. The rail is queue-grouped, so any bound delivery-credential responder can take the request, and the manager accepts the first reply with no identity check, no lease check and no proof that the responder is the daemon this manager provisioned. A reply naming the manager's own root selects the remint. The orchestrator adds one nuance that must travel with the finding: the serve path does carry two guards three lines away, a sender-bound reply guard and a from-id authenticity guard. Both authenticate the request direction, who may ask. Neither authenticates the reply direction, whether the thing that answered is the daemon whose store identity is about to be accepted. B2 is about the reply direction and is not addressed by either guard.

The body defects and non-blocking items below are the reviewers' own. Suite runs at this head, where a review repeated them, are listed under each review.

---

**Blocking (1):**
- B1-absence-branch-reachable-from-a-reply-on-this-head. A bound responder on the delivery-admin rail can produce, by mechanism, the condition the manager reads as 'no daemon is bound', and the manager then performs a full daemon remint into its own store while not co-rooted with the daemon. requestDeliveryAdmin issues a NATS request with noMux:true and an explicit reply subject. In @nats-io/nats-core 3.4.0 the noMux single-request path checks the received reply message and, when its payload length is 0 and its header code is 503, replaces the message with a NoRespondersError whose text is "no responders: '<subject>'". That text is produced [...]

**Body defects (3):**
- BD1-suite-table-omits-the-environment-sensitivity-of-one-listed-suite. The body's table lists `smoke:reload-store-identity | 32 passed` under 'Every suite that names the sources this touches, at this head'.
- BD2-fail-closed-claim-is-true-as-written-but-carries-an-unstated-assumption. The body states: "**Fail-closed is unchanged everywhere it was load bearing.** A delivery-admin rail that hangs (a timeout, deliberately not treated as absence), a daemon that will not name its store, and an unparseable identity all still end the start." The changeset and the source comment at implementations/manager/src/manager.ts:1498-1501 make the same claim.
- BD3-source-comment-asserts-a-property-the-shipped-tests-do-not-grade. implementations/manager/src/manager.ts:1489-1492 states the challenge "is a proof, not a poll: the daemon names ONE store, so `shared` holds for at most one manager and only that manager may remint".

**Non-blocking (3):**
- N1-start-challenges-the-rail-twice.
- N2-the-absent-verdict-has-no-grading-cell.
- N3-suite-environment-sensitivity-worth-a-rig-note.

**Not verified (7):**
- Whether B1 is reachable in a realistic deployment rather than in a harness.
- Whether the equivalent of #1636 v2 B2 (a responder-controlled lease claim trusted by shape) exists here.
- Whether an equivalent of #1636 v1 B1 (a malformed reply whose error text quotes the matched phrase) exists here.
- The PR body's full suite table.
- The PR body's merge-base comparison (21 passed / 2 failed at the base).
- pnpm changeset status.
- Host-constrained suites.

---

**Blocking (2):**
- B1_ERROR_TEXT_FORGES_ABSENCE. A delivery-admin responder can make a foreign manager treat an unreadable reply as absence and remint into the wrong store. requestDeliveryAdmin ends by parsing the response as JSON. When a bound responder sends the literal body 'no responders', JSON parsing throws an error whose message quotes that responder-controlled text. assertDaemonSharesSecretStore catches the error, passes only e.message to isAbsentDeliveryAdmin, and the /no responders/ regex returns absent. remintDaemonCredsIfOwner treats absent as permission to re-sign daemon credentials in this manager's store. No readable store relation was [...]
- B2_REPLY_HAS_NO_LEASE_PROVENANCE. Renewal ownership trusts whichever delivery-admin responder answers, without binding the answer to the delivery lease holder. Delivery credentials may subscribe to the wildcard delivery-admin request rail and publish to its reply subtrees. serveControl queue-groups responders by service. The client accepts the first ControlReply, parses only {kind, root|coordinate}, and compares it with the manager's store. There is no responder identity or lease-holder evidence in the reply, and no independent lease read. A responder that does not own the live delivery daemon can therefore name the manager's root and select [...]

**Body defects (4):**
- BODY_FAIL_CLOSED_OVERSTATEMENT. The PR body captured at 2026-09-15T18:36:32Z says an unparseable identity still ends the start and that fail-closed behavior is unchanged. On this head, an unreadable non-JSON response containing 'no responders' is classified absent and startup/remint proceeds.
- BODY_NO_REMINT_OVERSTATEMENT. The body and changeset state that a manager on another root remints no daemon credential and that only the matching manager remints. This is true for an honest foreign identity reply, but false when another delivery-admin responder returns crafted error text, an empty 503 reply, or the manager's root.
- BODY_OWNERSHIP_PROOF_OVERSTATEMENT. The body describes store identity as selecting a single deterministic renewal owner. Actual state is first queue response wins, with no binding to the delivery lease holder. The shipped evidence exercises honest responders only.
- BODY_ISSUE_CLOSURE_SCOPE. The body does disclose that issue requirements 3 and 4 are not implemented and requirement 6 is partial, while the PR still says `Closes #1634`. That closure claim overstates completion of the issue's explicit acceptance list.

**Non-blocking (1):**
- NB_SINGLE_ROOT_CELL_LABEL. The new `#1634: the co-rooted manager starts too` cell is `ok(..., true)` after an awaited start. A thrown start still reds the scenario, so it is not a false green, but the named cell itself carries no discriminating condition. The owner adoption and generation cells provide the real assertions.

**Not verified (5):**
- I did not exercise cross-host participant setup or `cotal up` guidance. The body explicitly says no guidance change was made.
- I did not run artifact-store, delivery-starvation, or hand-run delivery-boot-honesty. The packet identifies the first two as host-resource-fault-prone on this host and forbids the third hand run.
- `pnpm changeset status` could not determine divergence because this lane lacks a local `main` ref at the expected commit. I manually confirmed the feature changes touch @cotal-ai/core and @cotal-ai/manager and the changeset lists both as minor. This command limitation is not treated as a head defect.
- The first reload-store-identity attempt used an external TMPDIR under a parent containing `.cotal` and failed its fixture assertion that a non-.cotal path names no workstation. I corrected the instrument to a short private /var/tmp directory outside the lane and any .cotal ancestor. Two fresh runs then passed 32/32. The initial red was an instrument/setup fault, not a product finding.
- I did not independently replay the merge-base 21/2 comparison claimed in the body.

---

### Review 3: `REQUEST_CHANGES`

**Blocking (2):**
- B1_UNSOUND_ABSENCE_PREDICATE. isAbsentDeliveryAdmin regex on error string is satisfiable by peer replies, leading to unauthorized credential reminting into wrong store. isAbsentDeliveryAdmin tests the error message string via `/no responders|\b503\b/i`. When `ep.requestDeliveryAdmin('reloadStoreIdentity')` executes, the response payload is parsed via `m.json<ControlReply>()`. In Node.js / V8, if a responder sends a non-JSON body such as the text 'no responders', `JSON.parse` throws `Unexpected token 'o', "no responders" is not valid JSON`. This thrown message matches `/no responders/i`, causing the manager to classify the outcome as [...]
- B2_PROVENANCE_NOT_AUTHENTICATED. Manager trusts unauthenticated secret store identity claim from any delivery-admin responder. The `reloadStoreIdentity` request is dispatched over the queue-grouped delivery-admin rail (`ctl.delivery-admin.*.*`, queue: `service`). Any entity holding a delivery credential can bind to this rail. The manager parses `reply.data` via `parseSecretStoreIdentity(reply.data)` and performs a purely lexical/structural comparison `sameSecretStoreIdentity(this.secretStoreIdentity, daemon)`. It checks only shape and path equality, verifying neither the identity/lease of the responder nor whether the answering process [...]

**Body defects (1):**
- BD1_FAIL_CLOSED_OVERSTATEMENT.

**Non-blocking (2):**
- NB1_UNGUARDED_DELIVERY_GENERATION_MISMATCH. Smoke suite does not grade delivery generation mismatch refusal on identical identity. While manager-two-root-renewal tests that root A and root B generations remain distinct and owner renewal adopts, it lacks an explicit cell testing rejection when a delivery daemon re-reads an unexpected generation under the same identity.
- NB2_CONCURRENT_SHARED_MANAGERS_RACE. Multiple managers sharing an injected store coordinate lack an election or mutual exclusion lock for daemon renewal. While two managers on the exact same filesystem workspace root cannot run concurrently due to manager instance ID collisions, two distinct managers injecting the same store coordinate can both classify as shared and race during remint and reloadCreds.

**Not verified (2):**
- UV1_MULTI_SHARD_DELIVERY. Testing was conducted on single-shard configurations as supported by current implementation. Multi-shard behavior was not evaluated.
- UV2_HOST_RESOURCE_CONSTRAINED_SUITES. Heavy starvation suites (e.g. smoke:delivery-starvation) were skipped in accordance with packet guidance regarding host disk and environment constraints.

davidfarah2003:
Re-reviewed at `8e346f558`. Two independent reviews reach the same blocking conclusion, each with its own reproduction, so this needs a change before it can land.

**The single-renewal-owner premise is not enforced.** `assertDaemonSharesSecretStore` returns `shared` for any manager whose store identity equals the daemon's, and `sameSecretStoreIdentity` in `packages/core/src/secret-store.ts` is plain equality: trimmed path equality for `fs`, string equality for `injected`. It carries no holder, no instance id, no epoch and no tiebreak, so N managers satisfy it at once. Confirmed on both identity kinds.

**That reintroduces the adoption failure #773 exists to prevent.** `reloadCreds` carries a fingerprint and no bytes, so the daemon re-reads the file and compares against the caller's expected generation, and refuses with `reloadCreds: re-read credential generation did not match the expected re-signed generation ... nothing adopted`. Two owners remint on independent half-TTL timers with no ordering between them, so one owner's write lands between the other's re-sign and its adoption. Removing the startup refusal without putting an owner election in its place makes that reachable.

**The new suite does not cover the premise it rests on.** `bin/smoke/manager-two-root-renewal.smoke.ts` has a cell for the non-owner and a cell for the owner, and no cell for two managers that both match the daemon's store. The sentence the safety argument depends on is asserted in three documents and tested nowhere. Whatever shape the fix takes, a cell in that shape belongs in this suite.

Two things bound the severity and are worth saying plainly. The `#1634` correction itself is real and well covered: the non-owner starts, writes no `renewal.json` into its own root, leaves both roots' credential bytes at the original generation, and the daemon survives its start. The fail-closed paths the changeset claims to preserve are genuinely preserved, including a hung delivery-admin rail, an unreadable identity, and a daemon that will not name its store.

The remedy the notice offers an operator ("pass both processes the same store") is also narrower than it reads: when both managers take that shared root as their `workspaceRoot` they derive the same persisted manager instance identity, and the second is refused on that path instead.

What would unblock this: an explicit single-owner election over the store rather than an equality test, plus the two-matching-managers cell. Happy to look again once that is in.

davidfarah2003:
The basis for this pull request changed today, so the earlier panel note from 2026-09-15 has been removed rather than left to mislead, and this replaces it.

The single election that both reviews blocked this head for has now landed on `main` (PR 1670, closing #1634). It decides the daemon-renewal owner by two conditions rather than one: the daemon must reload from this manager's store, and this manager must hold a per-space renewal lease held as a CAS row with the same heartbeat as the per-instance liveness lease.

The reason for the second condition is the thing measurable here. `sameSecretStoreIdentity` is pure equality with no holder and no tiebreak, so every manager sharing one store passes it. The gate on that change reverted the production files and ran the suite: with identity as the only test, both managers recorded ownership and each reported two remint results. The named cell reads "exactly ONE of them owns daemon renewal ACROSS the lease TTL (the lease breaks the tie, store identity cannot)".

That is not a criticism of the approach here. The shared/foreign classification in this pull request is the same shape and it was right; it was the half that identity alone cannot finish.

What that means for this branch:

1. The classification work overlaps with what is now on `main` and will conflict. A rebase will show most of the manager change already present.
2. What is still not on `main` is item 3 of the issue, the byte-carrying `reloadCreds` so the owner's store and the daemon's store need not be one directory. The merged change starts and serves a cross-host manager and has it remint nothing, which is safe and incomplete. If you want to carry this forward, that is the piece with no owner.

Before rebasing, one thing worth knowing: our own attempt at the cross-host binding is blocked on a finding you would hit too. A process holding a copied standing delivery credential can read the lease row, echo the holder id in the reply's responder field, and be classified as sharing the store. It was reproduced against a real broker and daemon. Any design where the answerer supplies the principal it is checked against has that hole.

L4XB:
Thanks for the two write-ups, and for spelling out why identity alone could not finish it. #1670 is the election this branch was missing, so I am closing this rather than rebasing it onto a change that already carries most of it.

On item 3, one thought on the hole you describe, in case it is useful whoever picks it up. The echo only matters if the reply decides something the answerer can profit from. Two parts of a byte-carrying reload can avoid trusting the answerer at all:

- **What is sent.** A re-sign needs only the daemon's public nkey, and a user JWT is not a secret without the seed. If the owner signs for the delivery nkey it holds on record (from provisioning or the registry, never from the reply) and sends the JWT alone, an answerer without that seed receives nothing it can connect with. An answerer that has the seed already holds the credential; what is left is that each renewal extends such a copy, which only rotating the nkey per renewal (one successor per key, recorded next to the lease) would close.
- **What is believed.** "Adopted" can be a signature over the owner's nonce and the expected fingerprint, verified against that same recorded key, instead of a responder field. Something that cannot sign as the daemon then cannot produce a false adoption record, which on a cross-host daemon would otherwise hide a daemon that dies at expiry.

If that shape looks right to you I am happy to draft it as a separate pull request. If your own attempt is further along, I will leave it.

PR #1643, opened 2026-09-15
fix(core): stop a run's own replay durable from failing one of its steps

### LINKED ISSUE #1623: runtime: a parallel branch fails with RunJournalReplayRaced on the run's own takeover consumer
A parallel panel leg inside a manager-hosted Cotal Lang run failed with a `RunJournalReplayRaced` fault while the run itself was healthy and its two sibling legs settled normally seconds before and after.

Observed 2026-09-14 20:55:58Z on a `parallel` of three `ask` steps, each backed by its own spawned seat, 40 s after the second sibling settled:

```
L4000 handler-fault: run run-<id>: the replay consumer wfj_run-<id>_<takeover> had already delivered 17 records to someone else. Another driver is replaying this run; retry the takeover rather than resuming from a partial prefix.
```

The durable named in the fault is the run's own takeover consumer: the manager log for the run reads `started on endpoint manager (epoch 1, takeover <same id>)`, and no second driver ever took the run over (one activation, `completed in 14 step(s)` right after). Operator-side journal reads mint their own takeover ids (`newTakeoverId`) and the only ones against this run happened five minutes later, so the contender was inside the driver.

What the code allows: `replayRunJournal` (packages/core/src/run-journal.ts:387) creates a durable named by `(space, run, takeoverId)`, reads, and deletes it in `finally`. Two replays under the same takeover id therefore share one durable; `consumers.add` returns the existing one, `assertReplayConsumerFresh` sees a nonzero delivered count and throws `RunJournalReplayRaced`, and the first replay's `delete` tears down the second's fetch. `RunScopeAuthority` serializes its own replays through a promise chain (implementations/runtime/src/run-scope-authority.ts:49), but other in-driver replays under `req.lease.takeoverId` do not join that chain (`noRecordToResume` in run-driver.ts:184, and any caller that hands the lease id to `status` in run-host.ts:157), so a settle on one parallel branch overlapping a replay on another can collide.

Effect: a completed reviewer verdict was lost as a step failure that is not retried at the ask level. The recovery in `resumeRunTakeover` (run-journal.ts:775) treats the race as "a lost round, not a lost run" and retries, but the ask path surfaces it as L4000 and the branch fails.

Fix direction:

- Route every in-driver replay under the lease takeover id through the scope authority's serialized reader, or give each replay call its own consumer suffix so concurrent branches never share a durable.
- Treat `RunJournalReplayRaced` inside a live driver as retryable at the step, the way the takeover path already does, instead of failing the ask.
- Add a runtime cell: a `parallel` of three asks settling within the same second must not fail any branch; register the shared-durable behaviour as the mutant.

Not the #1619 poll-timeout family: this leg was at 149 s of work, not at a deadline.

L4XB:
Closes #1623

## What the fault was

A `parallel` branch failed with `RunJournalReplayRaced` naming the run's own takeover consumer, on a
run with one activation, no second driver, and sibling branches that settled either side of it. The
message told the operator another driver was replaying the run and to retry the takeover. There was
no other driver.

Reproduced on a live broker, message and signature both:

```
leftover: drained=17 delivered=17 ackPending=0
replay 1 -> RunJournalReplayRaced: run run-l: the replay consumer wfj_run-l_deadbeefdeadbeef had
            already delivered 17 records to someone else. Another driver is replaying this run;
            retry the takeover rather than resuming from a partial prefix.
replay 2 -> ok(17)
replay 3 -> ok(17)
```

One fault, then healthy, because the refusing replay deletes the offending durable on its way out.
That is the shape the report describes: one branch lost, the run completed.

## Who can attach to `wfj_<runId>_<takeoverId>`

The name is minted in one place, `runJournalConsumerConfig` in `packages/core/src/endpoint-binding.ts`
(`durable_name: wfj_${runId}_${takeoverId}`), and created in one place, `replayRunJournal` in
`packages/core/src/run-journal.ts`. No other shipped file calls `jsm.consumers.add` on the WFJ
stream. So the enumeration is the six callers of `replayRunJournal`:

| caller | takeover id it uses | connection |
| --- | --- | --- |
| `activateRun`, inside its replay-then-activate loop (`run-journal.ts`) | the drive's lease id | driver |
| `noRecordToResume` (`implementations/runtime/src/run-driver.ts`) | the drive's lease id | driver |
| the reader behind `createRunScopeAuthority` (`implementations/runtime/src/run-scope-authority.ts`) | the drive's lease id | mediator |
| `cotalLangRunHost.status` (`implementations/runtime/src/run-host.ts`) | whatever the caller hands it | caller's |
| `locateOpenCheckpoint` via `.locate` and `resolveCheckpoint` (`implementations/runtime/src/resolve-checkpoint.ts`) | whatever the caller hands it, or a fresh one when omitted | caller's |
| `cotal run journal`, `answer`, `ps` (`implementations/runtime/src/run-command.ts`) | a fresh `newTakeoverId()` per verb | operator |

The operator side does mint its own, on every path: `RunHosting.withOperator` in
`implementations/manager/src/run-hosting.ts` opens a one-shot `run-operator` connection with
`newTakeoverId()`, and `run answer` mints a second one for the write half. That part of the issue
holds.

Three of the six hold the drive's lease id, and only one of them serializes. `RunScopeAuthority`
chains the reads it makes itself; `activateRun`, `noRecordToResume`, and a `status` or `locate`
handed the lease id join no chain, and the first two are on the driver's connection while the
authority is on the mediator's. Nothing about the durable name enforces anything across those.

## The two ways it breaks, measured

**A durable can be left behind.** A replay is interrupted by whatever ends its connection, and the
delete in the `finally` is the part that does not land. Closing the connection at a sweep of instants
during a 17-record replay, on a live broker:

```
close after  1ms -> replay ClosedConnectionError    durable LEFT delivered=0  ackPending=0
close after  2ms -> replay ClosedConnectionError    durable LEFT delivered=17 ackPending=17
close after  3ms -> replay ClosedConnectionError    durable LEFT delivered=17 ackPending=11
close after  4ms -> replay ok(17)                   durable gone
...
close after 40ms -> replay RequestError             durable LEFT delivered=0  ackPending=0
```

A drive keeps one takeover id for its whole life, and its connection is a standing one that
reconnects without bound (`maxReconnectAttempts: -1`), so the next replay after such a blip meets its
own leftover and reads it as another driver's half-fed consumer.

**Two replays can share it.** Eight replays fired together on one takeover id, on base:

```
JetStreamStatusError, RunJournalReplayRaced, RunJournalReplayRaced, ok(0),
ok(0), ok(0), JetStreamStatusError, JetStreamStatusError
```

`ok(0)` is the bad one, because a caller cannot see it: a silently empty replay reads back as a run
with no history. A `TypeError` out of the client also turned up across repeated runs.

## What I could not reproduce, said plainly

I could not get the driver's own effect path to overlap on its own. Driving a `parallel` of three
pauses through the hosted composition performed 64 replays under one takeover id in 14 seconds, all
of them serialized by `RunScopeAuthority`, and instrumenting `replayRunJournal` to report an overlap
on a durable found none, in that probe or across `smoke:run-host-live`. `activateRun` and
`noRecordToResume` both run before the first authority read. So of the two mechanisms above, the
leftover is the one reachable through the shipped manager today; the concurrent case is reachable
from the interface rather than from a caller in this tree.

That is not a reason to fix only the leftover. Removing a leftover means deleting a durable of that
name, and doing that while a sibling replay in the process is using it is the tear-down the
per-takeover name was introduced to prevent. The serialization is what makes the repair safe, so
both go in together.

## The fix

`replayRunJournal` now holds the two properties its own header already claims.

- Replays on one durable run one at a time in the process, keyed by `<stream>/<durable>` rather than
  by any one caller. The lock belongs to the name because the name is what is shared: a caller
  cannot opt out and a new caller cannot forget to join. Different runs and different takeovers
  never queue behind each other.
- A durable of that name that comes back holding a tail is deleted and remade before the read. With
  the serialization in place it can only be this process's own leftover.

The freshness guard is unchanged and is not widened. `assertReplayConsumerFresh` still refuses on the
consumer that actually gets read, and that consumer is now a remade one; if it also comes back
holding a tail then a reader this process cannot account for is on the name, which is what
`RunJournalReplayRaced` says.

I did not take the issue's second suggestion, retrying `RunJournalReplayRaced` at the step. It is the
right instinct and it is what `activateRun` does, but at the step the state that would justify a
retry is not visible; inside the replay it is, so the recovery lives there and the step never sees a
fault it cannot act on.

The other alternative, a per-replay consumer suffix, is not available: a consumer name is one subject
token that no grant pattern covers in part, which is why the id is chosen when the credential rows
are minted rather than by the driver.

## Evidence

New cell in `packages/core/smoke/run-journal.smoke.ts`, on the merge base `712dfadf`:

```
$ pnpm smoke:run-journal
  ✗ FAIL: two replays fired together on ONE takeover id both read the whole run JetStreamStatusError, ok(17)
  ✗ FAIL: eight of them on one id read the whole run too, none of them a silent empty prefix JetStreamStatusError,RunJournalReplayRaced,RunJournalReplayRaced,ok(0),ok(0),ok(0),JetStreamStatusError,JetStreamStatusError
  ✗ FAIL: and the next replay on that same takeover id reads the whole run rather than reporting a rival RunJournalReplayRaced
run-journal.smoke: 123 passed, 3 failed
```

Three consecutive base runs gave the same three failures. On the branch, three consecutive runs:

```
$ pnpm smoke:run-journal
run-journal.smoke: 126 passed, 0 failed
```

The leftover in that cell is produced through the delete path the suite already exercises, not by
hand: a `jsm` whose delete refuses with an unnamed error, which is the same shape the connection
close above produces.

`pnpm mutation-proof --config packages/core/smoke/mutations/run-journal-replay-sharing.json`:

```
baseline green (0 progress marks)
KILLED       replays under one takeover id run concurrently again, sharing one durable
  red, and named: eight of them on one id read the whole run too, none of them a silent empty prefix
KILLED       a durable an interrupted replay left behind is inherited instead of remade
  red, and named: and the next replay on that same takeover id reads the whole run rather than reporting a rival
KILLED       the freshness guard stops refusing, so a tail is read as a run
  red, and named: a durable that outlives the replay's own delete is still refused by name
```

Suites run on the branch, all green: `smoke:run-journal` (126), `smoke:run-journal-auth` (8),
`smoke:ep-journal` (115), `smoke:runtime-journal-store` (8), `smoke:runtime-run-driver` (87),
`smoke:runtime-run-driver-auth` (74), `smoke:runtime-run-command` (23), `smoke:runtime-fork` (110),
`smoke:runtime-migrate` (68), `smoke:runtime-mesh-checkpoint` (42), `smoke:runtime-mesh-ask` (42),
`smoke:run-host-live` (51), `smoke:lang-spawn-live` (37), `smoke:lang-supervise-live` (11),
`smoke:docs`. Plus `pnpm typecheck`, `pnpm build`, `pnpm check:docsbundle` and
`pnpm check:docs-voice`.

`pnpm test` fails on this machine, and it fails identically at `712dfadf` with the same change absent:
`examples/06-feed-agent` on a fixture publish and `implementations/auth` on
`Path is a directory: .../managed-actors`. Both look like macOS environment rather than this change.
Everything above was measured against `nats-server v2.12.12`, the version CI pins.

## Scope

- `packages/core/src/run-journal.ts`: the fix, plus `dropReplayConsumer` extracted because the delete
  is now made from two places with the same rule.
- `packages/core/smoke/run-journal.smoke.ts`: cell 5m2, one takeover id read many times.
- `packages/core/smoke/mutations/run-journal-replay-sharing.json`: the fixture, three mutations.
- `docs/workflows.md` and the regenerated `extensions/connector-core/src/docs-bundle.generated.ts`.
- `.changeset/replay-durable-not-shared.md`.

`SPEC.md` is untouched: the wire contract is unchanged. A replay still creates a per-takeover durable
filtered to the run's subject, reads it, deletes it, and still refuses to resume from a tail.

One correction to the issue text: the recovery it cites as `resumeRunTakeover` at
`run-journal.ts:775` is the retry loop inside `activateRun`, which begins at `run-journal.ts:745`.
There is no `resumeRunTakeover` in the tree. The point it makes is right.

File and line references are against `main` @ `712dfadf`.


davidfarah2003:
Panel complete: three independent reviews of `49b416bd60ec59c13c4573c684d31676113729ef`.

Each review graded the merge of this head onto main in its own lane (merge tree `3f39229725b0298586fd592ac123306b4a42fa3e`), so the figures include the main-side change that landed while this pull request was open; the census cell the packet warned about as inherited red was already fixed on main by the time the lanes were cut, and all three reviews report it green. Two reviews approve and one requests changes. All three reproduced the three failure modes the pull request names at the merge base and found them healed at this head, and all three report the new mutation fixture killing every mutant.

The two approving reviews differ in depth. One approves with zero findings on the named reproduction set. The other drove the primitive and the recovery path directly and records four non-blocking items: the serializer has no timeout, so a body that never settles blocks its key; the leftover recovery deletes a consumer another process is actively reading if two processes ever share a durable name, which it graded as unreachable in shipped code after tracing every place a takeover id is minted (64 random bits minted in the process that uses it, never read back, and ignored by the two wire-reachable handlers when a caller supplies one), disclosed in the code, changeset and docs, and worse at the base; a pre-existing predicate accepts coerced code shapes; and a security smoke on a separate package reports a red rather than a refusal when docker is absent. It also records that the full smoke gate stopped at that docker cell on the review host, so the cells after it are unverified rather than passed.

The request-changes review reports the same cross-process mechanism as its one blocking item and grades it differently: with the lock process-local and the consumer name broker-global, a second process can adopt the first process's live consumer as a leftover and delete it under the first process's fetch, and its own two-process trials at this head produced torn fetches and silent empty replays on a seventeen-record journal. Its objection is to the prose as much as the code: the changeset says a non-fresh durable can only be this process's own leftover, and the docs describe deletion as recovery of an interrupted earlier read, so the residual cross-process risk is asserted away rather than bounded. It also records that the changeset names only the core package while the branch changes a generated file in the connector-core package.

So the panel agrees on the mechanism and divides on whether the shipped code can reach it and whether the prose is honest about the remainder. What would change each verdict, in the reviewers' words, is under each review below.

---

### Review 1: `APPROVE`

I reproduced the three measurements the change rests on by driving them, not by reading them. At the merge base all three appear: eight overlapping replays on one takeover id return torn fetches plus a silent ok(0) that reads as a run with no history, a 1ms stagger returns a mixture including ok(0), and a replay whose delete is refused leaves its durable at delivered=17 so the next replay on that id reports RunJournalReplayRaced against a rival that does not exist. At this head all three are healed across five repetitions, and the guard that should still refuse still refuses. The suite is [...]

**Blocking (0):** none

- packet WHAT-TO-GRADE.md. States the head is '2 commits behind main and 4 ahead'. I measure 3 behind and 4 ahead in my lane. Most likely main advanced by one commit, #1645, between the packet being written and the lane being cut. Reported, not edited.
- packet WHAT-TO-GRADE.md and my brief. Both describe an inherited red in implementations/auth/smoke/issued-profile-census.smoke.ts. That suite is green in my lane on three runs, because the fix for it, #1645, is one of my merge parents. The instruction was sound when written and is stale for this lane. Reported, not edited.
- my brief. Says to expect six OK lines from sha256sum -c MANIFEST.sha256. The manifest contains five entries and does not list itself, so the check prints five OK lines and exits 0. All five verified and the manifest's own digest matched. Reported, not edited.

**Non-blocking (4):**
- NB-1. The serializer has no timeout, so a body that never settles blocks its key indefinitely. exclusively chains each caller onto the previous promise with no upper bound on how long a body may take. I drove this: behind a body that never settles, the next caller on that key never ran. In shipped use the body is bounded by the fetch's 15-second expiry and by the request-reply timeouts on add and delete, so reaching it would need a broker client that neither resolves nor rejects. Recording it as a property of the primitive worth knowing, not as a defect to fix here. Probe scenario 7: 'behind a never-settling body, the next caller ran: false'.
- NB-2. The leftover recovery can delete a consumer another process is actively reading, if the two ever share a name. The recovery treats any non-fresh consumer of the name as this process's own leftover. I constructed a live foreign reader holding a tail on that name and the replay deleted it and returned ok(17), three times out of three. The premise that makes this safe is that no other process can hold the name. I confirmed that premise twice over: every takeover id is 64 random bits minted inside the process that uses it, and, separately, the two wire-reachable handlers ignore a caller-supplied takeoverId entirely, taking the id only from withOperator, whose scope type has no such field. So the premise holds under adversarial input and not merely under well-behaved callers. The code, changeset and docs all state the in-process scope, so it is disclosed rather than hidden. The base [...]
- NB-3. isConsumerNotFound accepts some coerced code shapes. The predicate uses Number(err.code) === 10014, so a string '10014' or a single-element array [10014] also satisfies it, and a name of ConsumerNotFoundError alone satisfies it with no code. No real broker error I produced takes those shapes. The predicate is pre-existing code that this PR moves rather than writes. Real errors: ConsumerNotFoundError/10014 true, StreamNotFoundError/10059 false, storage 10023 false, generic Error false, undefined false, null false.
- NB-4. smoke:jcode-security reports a red rather than a refusal when docker is absent. The suite guards for Windows and for running as root but not for docker being installed, so on a host without docker the spawn returns status null and the cell asserts false. This is a pre-existing gap in that suite and is unrelated to this PR, which touches no file in that package. which docker exits 1; the assertion prints {"status":null}.

**Not verified (14):**
- I did not sever a live TCP connection 2ms into a replay, which is the literal mechanism the PR describes for the leftover durable. I produced the same end state by the mechanism the code names, a delete that fails for a reason the replay does not swallow, and observed delivered=17. The end state matches; the specific trigger is untested.
- I did not reproduce the original operator-visible symptom, a parallel branch failing with RunJournalReplayRaced on a real multi-agent workflow run. I verified the mechanism beneath it and that live runtime suites reach the changed code, but I did not run the end-to-end scenario from issue #1623.
- I did not run the full CI gate to completion. It stopped at a missing docker binary, and 455 of 599 planned smokes never started, including the ones outside the run-journal neighbourhood. I compensated for this PR's own area by direct runs, but I have not observed the whole gate green on this host.
- I did not test on an authenticated mesh with real minted credentials. All my broker work used an unauthenticated local nats-server, so I have not exercised the grant rows that pin the durable name, only the naming and consumer logic.
- I did not test the behaviour under JetStream storage exhaustion, which the environment notes warn can make a legitimate consumer operation fail. I observed no storage refusal in any run I made, so I have nothing to report on that axis either way.
- I did not measure whether the serialization introduces a throughput cost for a drive that replays hundreds of times under one takeover, since the replays it serializes were previously racing rather than running in parallel usefully. No latency comparison was made.
- I did not review the other four commits' interaction with anything outside packages/core, beyond confirming the diff touches only six files.
- I did not inspect the .internal submodule or any plan record, and so cannot speak to whether a plan row was updated.
- My cross-process finding rests on how takeover ids are produced and on which request fields the handlers actually read. I audited shipped source under packages, implementations and extensions, and I confirmed the two wire-reachable run handlers ignore a caller-supplied takeoverId. I did not audit example composition roots or deployment configurations for a path that might supply a fixed takeover id.
- extensions/connector-jcode/smoke/jcode-security.smoke.ts, reached via pnpm smoke:ci. docker. REFUSAL. A command that is not found is not a zero and not a failing suite. This stopped the CI shard and is the sole reason pnpm smoke:ci exited 1. which docker exits 1; docker --version prints 'command not found'; the failing cell reports {"status":null}, meaning the child never ran.
- pnpm changeset status. INVOCATION ARTIFACT of a detached merge lane, not a defect in the changeset. The changeset file itself is well formed and declares @cotal-ai/core as a patch. 'Failed to find where HEAD diverged from main'; git rev-parse main resolves to 6ab195b865af58a87ccd1d77f1314f0882537541, so the ref exists.
- pnpm check:docsbundle, first attempt. MISSING BUILD in my lane. Passed with exit 0 after pnpm build. ERR_MODULE_NOT_FOUND on extensions/connector-core/node_modules/@cotal-ai/core/dist/index.js.
- pnpm test, first attempt. MY OWN INSTRUMENT. Passed end to end with a shorter TMPDIR still outside the lane. listen EINVAL on a 118-byte unix socket path under my TMPDIR, against the 107-byte sun_path limit.
- Inherited red, not attributed to this head. My merge parent 5e23b1d56ba5b35376be9422fff7d41d6b3667ae is PR #1645, the separate fix for that cell, and git merge-base --is-ancestor confirms it is an ancestor of my HEAD. The red was fixed on main before my lane was cut, so there was no inherited red to misattribute.

---

The in-process queue is sound and the new tests discriminate, but leftover recovery is unsafe across processes. A second process can adopt a live same-name consumer, classify it as a leftover, and delete it out from under the first process. Live-broker trials at this head produced torn fetches and silent empty replays. The changeset and docs do not state this residual risk and instead claim the non-fresh consumer can only be this process's own leftover.

**Blocking (1):**
- Leftover recovery deletes a live replay owned by another process. packages/core/src/run-journal.ts:406-419; packages/core/src/run-journal.ts:463-479; .changeset/replay-durable-not-shared.md:33-39; docs/workflows.md:215-219. exclusively is process-local, but the consumer name is broker-global. A second process can receive the first process's consumer from add, see it as non-fresh, and delete it because there is no local map owner. The delete tears down the first process's fetch. A near-simultaneous add can also let both readers proceed before freshness reflects delivery. At this head, distinct-process tests produced JetStreamStatusError, ConsumerNotFoundError, RunJournalReplayIncomplete, and silent ok(0) returns for a 17-record journal. The silent empty outcome is data-loss behavior at the API boundary, and the recovery path actively disrupts a live peer. The PR prose asserts the unsafe [...]

**Body and prose defects (2):**
- .changeset/replay-durable-not-shared.md. The statement that a non-fresh durable can only be this process's own leftover is false because the lock is process-local and the durable is broker-global.
- docs/workflows.md. The page describes deletion as recovery of an interrupted earlier read and assigns unaccounted ownership only after deletion survives. It does not disclose that the first deletion can target a live reader in another process or that silent empty replay remains possible.

**Non-blocking (1):**
- Changeset package list omits the changed connector-core package. The branch changes extensions/connector-core/src/docs-bundle.generated.ts, but the changeset frontmatter lists only @cotal-ai/core. The fixed version group still schedules lockstep versions, so this is release-note attribution rather than a version-skew blocker.

**Not verified (9):**
- I did not run the full pnpm smoke:ci or pnpm check matrix.
- I did not run artifact-store or smoke:delivery-starvation because the packet identifies host storage pressure as an environmental source of red and they are outside this PR's focused reproduction set.
- I did not reproduce the original issue through the exact full parallel Cotal Lang program and manager UI. I exercised the built core replay API on a live broker, plus runtime-driver and authenticated runtime smoke suites.
- I did not construct two complete authenticated manager daemons that legitimately share one takeover credential. The cross-process measurement used separate Node processes and NATS connections calling the built replayRunJournal entry point with the same run and takeover durable.
- The .internal submodule was not initialized in this lane, so I could not read .internal guidance or STATUS.md. I did not initialize or modify it during this independent review.
- I did not write to GitHub, push any branch, or edit contributor prose.
- pnpm changeset status. invocation/context refusal. The lane has no local branch named main, so changesets could not find divergence. Re-running with --since and the packet's branch merge base exited 0.
- two mutation-proof invocations accidentally started concurrently. review-instrument interference, not a product result. One process saw its target already mutated and the other lost its temporary restore file, leaving the deliberate mutant in the lane. I restored only packages/core/src/run-journal.ts from the recorded merge HEAD, verified identical SHA-256, then ran two additional sequential clean passes. Together with the earlier clean pass, the reported mutation result is three valid sequential runs.
- Inherited red, not attributed to this head. implementations/auth/smoke/issued-profile-census.smoke.ts, versioned-rail subject-token census cell 60 checks passed, 0 failed The merge lane includes main commit 5e23b1d56, which fixes that inherited cell. The contributor branch itself adds zero changes to the census and endpoint subject-builder files. (not observed in lane).

---

### Review 3: `APPROVE`

PR #1643 correctly and minimally addresses issue #1623. The three reported failure modes on live JetStream (torn fetch from overlapping delete, silent empty prefix/replay on 0ms collision, and leftover durable holding delivered sequence on interrupted connection) were independently reproduced and confirmed. The in-process `exclusively` promise chain correctly serializes concurrent replays without unhandled rejection leaks or state leaks. Leftover durable recovery safely remakes un-fresh consumers while preserving the guard against active rival readers. All 3 mutations in [...]

**Blocking (0):** none

**Non-blocking (0):** none

**Not verified (3):**
- Full multi-tenant container deployment and long-running soak test under 24-hour continuous load.
- Refusal not counted as a failure. Foreground timeout on recursive pnpm test cancelled after 120s due to broad scope across all examples and packages; individual smoke targets and build/typecheck used instead.
- Inherited red, not attributed to this head. None observed in the merged lane. In commit 5e23b1d56 on main ('keep the versioned rail subject token out of source comments (#1645)'), the comment token in runtime that caused the census check red was fixed prior to merge. Running `node --import tsx implementations/auth/smoke/issued-profile-census.smoke.ts` passed all 60 checks.

davidfarah2003:
Merge basis for `49b416bd60ec59c13c4573c684d31676113729ef`: not met. Two of three independent reviews approve and one requests changes; the request-changes item is that the leftover recovery treats any non-fresh consumer of the durable name as this process's own leftover, so a second process on the same name can delete a live reader's consumer under its fetch, and that the changeset and docs state the inference as safe rather than bounding the cross-process remainder. One approving review reproduced the same deletion and graded the shared-name state as unreachable in shipped code; that disagreement is left to the reviewers' own words above. Under the standing rule an unanswered request for changes gets no new panel; a revision addressing it goes back to the requesting review for re-grade. This head stays open for the contributor; nothing here is pushed to the branch by anyone else.

davidfarah2003:
Re-reviewed at `1073a3a90`. The earlier note on this pull request graded `49b416bd6` and no longer applies, so it has been removed rather than left to mislead; this replaces it. Two independent reviews of the current head reach the same conclusion.

Issue #1623 asks for three things. This head does the first and not the other two.

**1. Serialise in-driver replays, or give each replay its own consumer suffix.** Implemented, but in-process only. `exclusively()` is backed by a module-level `Map` in `packages/core/src/run-journal.ts`, so two node processes hold two maps and two drivers sharing a takeover id share the durable exactly as before. Measured directly: four processes, two replays each on one takeover id, ten trials, 66 of 80 replays failed and three returned a silent `ok(0)` empty history. A guard that holds inside one driver does not answer the case the issue describes.

**2. Treat the race inside a live driver as retryable at the step.** Not implemented. `RunJournalReplayRaced` has no references outside `packages/core`. The in-driver replay sites (`run-driver.ts:184`, `run-host.ts:157`, `run-command.ts:432`, `resolve-checkpoint.ts:211`, `run-scope-authority.ts:18`) have no retry for it. Only `activateRun` retries it, and that is the takeover path the issue asks the step path to imitate.

**3. A runtime cell where a parallel of three asks settling within the same second fails no branch.** Not implemented. Nothing under `implementations/runtime` is touched. The new cell lives in the core run-journal smoke and calls `replayRunJournal` by hand rather than driving a parallel of three asks, and the new mutation names the core suite. So nothing yet proves a real parallel reaches the shared durable.

Two things here are good and should not be lost in a rework. The leftover-durable repair is a real second bug, found and fixed: an interrupted replay leaves the durable at `delivered=17` and every later replay on that takeover id then reports a phantom rival. The reverted suite reproduces it as a named failure and the fix closes it. The revert gate on this head is also sound: restoring only `packages/core/src/run-journal.ts` to the merge base with the tests kept gives exit 1, zero module-resolution errors, and three named assertions failing; restored, all 126 assertions pass over ten consecutive runs and the registered mutations are all killed.

What would unblock this: make the serialisation hold across processes (or give each replay its own consumer suffix, which the issue offers as the simpler alternative), retry the race at the step in the runtime driver, and put the parallel-of-three cell under `implementations/runtime` with the shared-durable behaviour as its mutant.

L4XB:
Thanks — the three points are clear, and the leftover-durable repair and the revert gate are noted as keepers.

Before I rework point 1 I want to check one constraint with you, because it decides which of the two directions is actually open.

**The per-replay suffix cannot be granted as the credentials are minted today.** `runJournalConsumerConfig` names the durable `wfj_<run>_<takeover>`, and the rows `consumeCreateRow` / `consumeBindRows` / `consumeDeleteRow` pin that name exactly. A consumer name is one subject token, so no pattern covers a suffix inside it (`wfj_<run>_<takeover>_*` is a literal, as the comment on `takeoverId` says). A replay that invented `wfj_<run>_<takeover>_<n>` would be refused by the broker. Making the suffix work means minting more than one name with the lease — a fixed pool of, say, four pinned names that replays take in turn — and `consumers.add` returning an existing consumer on a name collision means the pool still needs the freshness check to pick a free slot, and two processes can still pick the same slot at the same instant.

**Cross-process serialization needs a primitive the driver's credential does not hold either** — a KV lock row or similar — so it is also a grant change.

Which leads to the question: **is there a production path where two processes hold the same takeover id?** A takeover id is minted with the lease for one driver, and operator reads mint their own (`newTakeoverId`). The four-process measurement exercises two drivers sharing one lease's credential, which I read the lease model as ruling out; every sharing I can find (`noRecordToResume`, `RunHost.status` with the lease id, `activateRun`) runs on the driver's own connection, in its own process, which is what `exclusively` covers. If you know of a cross-process sharer, I will take whichever grant change you prefer (a small pinned pool, or a lock row); if not, I would rather state the in-process scope in the code than widen the credential for a configuration the lease model excludes.

**Points 2 and 3 do not depend on that answer**, so I am doing them now: retrying `RunJournalReplayRaced` at the step in the runtime driver's in-drive replay sites, the way `activateRun` already does, and a cell under `implementations/runtime` that drives a real `parallel` of three `ask` steps settling within the same second, with the shared-durable behaviour registered as its mutant.

L4XB:
Points 2 and 3 are in #1692.

**Point 2.** A live drive now replays again when a read loses a round to another reader. It uses `activateRun`'s bound: three rounds with no pause between them, and the race is raised unchanged after the third. This applies to the two reads a drive makes under its lease id: the step reads behind `RunScopeAuthority` and the no-record diagnostic. `status`, `locate` and `cotal run journal` each run under a takeover id minted for that one read (the manager and the CLI mint a fresh one per call), so they still report the race on the first read. The PR has the per-site table.

**Point 3.** `smoke:runtime-run-host-replay` drives the manager's run host through a `parallel` of three asks, all answered within the same second, in two variants. In one, `RunHost.status` reads the drive's own takeover id. In the other, a second connection takes a round off the drive's durable. No branch fails in either. Each retry site is killed on a named cell, and so is the bound. Reverting #1643's lock is also killed, with a core control that proves the rebuilt core is loaded. With core from before #1643, the status variant fails a branch in three runs out of three, and in one of those runs a status read came back with an empty journal.

One finding: in a hosted drive, the authority's own chain already orders the step reads. Reverting the core lock alone is therefore only visible when a second reader uses the lease id, such as `status`. Removing the chain while the core lock is in place survives (a same-file control was killed), so the fixture records that under `redundant`.

One limit worth naming: only the race is retried. A reader in another process can also tear the fetch, leave a replay incomplete, or return an empty replay, and those stay with point 1.

Point 1 is still open on the grant question above.

PR #1644, opened 2026-09-15
fix(cli): report a partial ps census instead of exiting 0 on half a space

### LINKED ISSUE #1637: cotal ps drops every seat of a manager instance that refuses the pinned contract, and exits 0
## What happens

In a space served by two manager instances that serve different command contracts (for example two versions of the manager), `cotal ps` sometimes lists only one instance's seats, prints the other instance as a single refusal line, and exits 0.

Measured against two live managers, eight consecutive calls: when instance B answered the initial `describe`, instance A was rendered as

```
manager <A>  pinned digests sha256:.../sha256:... do not match the served contract sha256:.../sha256:...; a member that cannot honor a pinned digest rejects, never coerces (SPEC 13.7)
manager <B>  1 agent
  <seat>  ...
```

and the five seats hosted by A were absent from the output. When A answered the describe, the roles swapped and A's five seats were listed while B's one seat was absent. Which instance answers the describe is whichever wins the queue, so the output flips between calls. Exit status was 0 in every case. Roughly one call in four reported a one-agent mesh while six seats were running.

## Why

The client pins the contract digests from one `describe` reply and fans the list command to every registered instance. `bindContract` (`packages/core/src/endpoint-serve.ts`) on the instance whose served contract differs refuses with `contract-mismatch`, which is the specified behaviour for that instance. The ps renderer's multi-manager branch (`implementations/cli/src/commands/agents.ts`, the `for (const inst of instances)` loop) prints `inst.error` as that instance's header line and continues. Nothing counts the instances that contributed no rows, nothing says the census is partial, and the exit status does not change.

The refusal is correct. The rendering is not: a refusal that hides every seat behind it is indistinguishable from an instance with no seats, and the caller has no signal that the list is incomplete.

## What is required

1. When any instance answers the fan-out with an error, `cotal ps` says the census is partial, names how many instances did not contribute rows, and exits non-zero (or a documented distinct code) so a script cannot read it as a full listing.
2. A contract mismatch between instances of one space is a condition the operator should see once, plainly: the two instances serve different contracts, with both digest pairs, rather than a SPEC citation on one row that moves between calls.
3. Either the client describes each instance and pins per instance before fanning out, or the requirement that all managers of one space serve the same contract is stated as a deployment rule and checked where a second manager joins.

## Reproduction

Two managers for one space from two different builds, at least one seat under each, then `cotal ps` repeated eight to ten times and the per-instance seat counts compared with the roster.

## Related

#1634 (many managers per space). #1577 (surfaces that report success over a missing component); this is the same failure class on the command plane.

L4XB:
Closes #1637

## What is wrong

`cotal ps` on a static or open mesh is a class scatter: it resolves the manager service once,
which pins the `ps` command contract from whichever instance answers the `describe`, and fans the
request to every registered instance. An instance serving a different contract cannot honor that
pin, so it rejects (`contract-mismatch`, SPEC 13.7), and the seats it hosts never appear.

The refusal is right. The rendering was not, in two ways, and the second is the dangerous one:

1. The refusal was printed as that instance's header line and nothing else. A refusal that hides a
   whole roster is indistinguishable from an instance with no seats, and which instance refuses
   follows whichever one wins the describe queue, so the listing swaps between calls.
2. The command returned normally. Nothing counted the instances that contributed no rows, nothing
   said the census was partial, and the exit status was 0 either way. The exit status is the only
   half of this output a script reads, and it said the listing was the space.

The multi-manager branch of `ps` (`implementations/cli/src/commands/agents.ts`, the
`for (const inst of instances)` loop) prints `inst.error` and continues, then the function ends.
The refusal itself comes from `bindContract` in `packages/core/src/endpoint-serve.ts`, which is
correct and unchanged here.

### Measured, before the change

A fixture with two instances of one endpoint whose `ps` output contracts differ (five seats under
one, one seat under the other), a real broker, real `serveEndpoint` on both, and the real CLI:

```
$ cotal ps --space <space>
manager aaaaaaaaaaaaaaaaaaaaaaaaaa  5 agents
  aaa-seat-0  claude · managed  running 1s  idle
  aaa-seat-1  claude · managed  running 1s  idle
  aaa-seat-2  claude · managed  running 1s  idle
  aaa-seat-3  claude · managed  running 1s  idle
  aaa-seat-4  claude · managed  running 1s  idle
manager bbbbbbbbbbbbbbbbbbbbbbbbbb  pinned digests sha256:1db0472a…/sha256:cd36fef9… do not match
the served contract sha256:1db0472a…/sha256:02d0816c…; a member that cannot honor a pinned digest
rejects, never coerces (SPEC 13.7)

EXIT STATUS: 0
```

Six seats are running. One instance's worth is listed. Nothing on stderr, and status 0.

## What this changes

`ps` counts the instances that did not report their seats and, when there is at least one, names
how many of how many and exits non-zero.

**A silence and a refusal count the same way.** Both leave the same hole: the seats behind either
are equally absent from the listing, and the operator's question ("is this every seat in the
space?") has one answer for both. This does widen the fix past the reported symptom, deliberately:
a registered instance that gave no answer within the deadline also produced a listing that was not
the space, and it also exited 0. Giving those two conditions different exit statuses would mean a
script reading an incomplete census as complete in one of them. An instance that answers with no
seats has reported, so a healthy multi-manager space is unchanged.

**The exit code is 1, not a new one.** `docs/cli.md` already states the rule for the user-auth
branch of this same command ("A manager that does not answer fails the command outright (exit
non-zero)"), and that path exits 1 through `failIfNotOk`. This is the same condition on the other
branch, so it takes the same status rather than inventing a code for it.

**The contract split is stated once, plainly.** When a refusal carries `contract-mismatch`, the
notice adds one line naming the pinned digest pair, the instance whose `describe` supplied it, and
the deployment rule behind it. Naming the instance is what makes the swap legible: the same command
run twice pins a different contract and hides the other half of the fleet. The per-instance rows
keep the manager's own verbatim message, so both digest pairs stay where the reader can see them
and no server wording is rewritten by the client.

**Decided on the code, never on the message.** `ScatterInstanceReply` now carries each refusal's
structured error code beside its operator wording, and the split is keyed on that. The message is
copy; a renderer matched to its prose would go quiet the day the wording improved.

**Nothing is hidden and `--json` stays machine-readable.** The rows print as before. The notice
goes to stderr in both presentations, so `--json` stdout remains one JSON object per seat per line.

### After the change, same fixture

```
$ cotal ps --space <space>
manager aaaaaaaaaaaaaaaaaaaaaaaaaa  5 agents
  aaa-seat-0  claude · managed  running 1s  idle
  … 4 more …
manager bbbbbbbbbbbbbbbbbbbbbbbbbb  pinned digests … do not match the served contract …

# stderr
✗ partial listing: 1 of 2 manager instances did not report its seats, so any seat they host is
  missing here
  ↳ the managers of this space do not all serve the same command contract. This listing pinned
    aaaaaaaaaaaaaaaaaaaaaaaaaa's sha256:1db0472a…/sha256:cd36fef9… from its describe, and a manager
    that cannot honor a pinned digest rejects rather than coerces (SPEC 13.7) - so which half of
    the fleet you see follows whichever instance answers the describe. Run one build per space, or
    read a single instance with --on <instance>.

EXIT STATUS: 1
```

Repeating the call swaps which instance answers the describe, and the notice follows it. Both runs
now exit 1.

## What this does not change

The issue's third requirement offers two ways to keep a space coherent: describe each instance and
pin per instance before fanning out, or state the same-contract requirement as a deployment rule
and check it where a second manager joins. This PR does the first half of the second option: the
rule is now stated in `docs/cli.md` beside the `ps` mode split, and the command reports when it is
violated. The join-time check is not here. It belongs to the manager's registration path rather
than to the CLI, and #1634 owns many-managers-per-space; a maintainer may want it filed separately
rather than folded into a rendering fix. I have used the closing keyword because the reported
defect (dropped seats, exit 0) is fixed; reopen or spin the join-time check out if you would rather
track it under this number.

Per-instance describe-and-pin was not taken. It would replace one `describe` and one `all`-rail
publish with N of each, and it would make an incoherent deployment quietly work rather than say so,
which is the opposite of what SPEC 13.7 asks a member to do.

One sibling is deliberately untouched. `locateSeat` (used by `stop` and `attach`) counts an
instance that answered with a refusal among its "reachable" instances, so `stop --name X` in a
split space can report `no managed agent "X" on any of the 2 reachable manager instance(s)` when
one of the two was never able to look. That is the same defect class on the verbs that act. It
already exits non-zero, so it is a wrong message rather than a false success, and it is out of the
scope this issue names.

## Evidence

Everything below was executed. `nats-server` v2.14.7, node v26.8.2, macOS arm64.

**The condition is reproduced live, not simulated.** The fixture registers two instances of one
endpoint with genuinely different `ps` contracts and serves both through `serveEndpoint`; the
refusal comes out of the serve boundary's own §13.7 binding.

### Base vs branch, on the real command

The same fixture script (it imports nothing this branch adds, so it compiles in both trees) run in
a worktree at the merge base and in the branch worktree, each after its own `pnpm build`. The build
matters: the CLI resolves `@cotal-ai/cli` through `dist`, so a comparison without it grades two
stale builds.

| | `cotal ps` | `cotal ps --json` |
|---|---|---|
| merge base `531581b9` | exit **0**, no stderr | exit **0**, no notice |
| branch | exit **1**, partial-listing notice | exit **1**, stdout still pure JSONL |

Base run, verbatim tail:

```
manager bbbbbbbbbbbbbbbbbbbbbbbbbb  pinned digests … do not match the served contract …
--- stderr ---

EXIT STATUS: 0
```

Branch run, verbatim tail:

```
--- stderr ---
✗ partial listing: 1 of 2 manager instances did not report its seats, so any seat they host is missing here
  ↳ the managers of this space do not all serve the same command contract. This listing pinned …

EXIT STATUS: 1
```

### The suite

`pnpm smoke:ps-partial-census`, 24 cells, all green on the branch:

```
1. the space: two instances on different contracts, plus one that answers nothing
  ✓ the freeze names all three registered instances (six seats run under the two that serve)
  ✓ the describe that supplied the pin was answered by ONE of the two serving instances
  ✓ every registered instance is in the result, silent one included (SPEC 13.5 pin 3)
  ✓ the instance on the other contract REFUSED, and the refusal is the serve boundary's own
  ✓ ...naming both digest pairs, which is why the CLI never has to parse this message
  ✓ THE DEFECT'S INPUT: six seats are running and the rows carry one instance's worth
  ✓ ...and the silent registration contributed no rows either
2. the census over those replies
  ✓ one instance of three reported its seats
  ✓ ...so two did not, the refusal and the silence counted the same way
  ✓ the refusal is read as a CONTRACT SPLIT, off the structured code
  ✓ a partial census produces a notice
  ✓ ...that names how many instances did not report
  ✓ ...and a second line naming the split, the pinned digests, and the instance they came from
3. controls: what must NOT produce a notice, and what must not claim a split
  ✓ every instance reporting is a COMPLETE census
  ✓ ...and produces no notice, so a healthy multi-manager space stays quiet
  ✓ an instance that answers with NO seats has reported: zero rows is an answer, not a hole
  ✓ a refusal under another code still makes the listing partial
  ✓ ...but does NOT claim a contract split, so the split line keys on the code and not on prose
  ✓ ...and a refusal whose MESSAGE reads like a mismatch but carries no code is not one either
4. and the command itself, run against this space
  ✓ `cotal ps` over this space EXITS NON-ZERO: a partial census is not a listing of the space
  ✓ ...while still printing every seat the instance that answered could enumerate
  ✓ ...and saying on stderr that the listing is partial
  ✓ ...and naming the contract split once, with the pin, off the code the scatter carried
  ✓ --json exits the same way and keeps stdout one JSON object per seat, notice on stderr

ps partial census smoke: 24 passed, 0 failed
```

At the merge base the suite cannot load at all (`psCensus` does not exist yet), which is a weak
discriminator, so the base-vs-branch table above is measured on the real command instead and the
suite's own discrimination is proven by mutation.

### Mutation proof

`node scripts/mutation-proof.mjs --config implementations/cli/smoke/mutations/ps-partial-census.json`

```
KILLED       the partial census exits 0 again
  red, and named: EXITS NON-ZERO: a partial census is not a listing of the space · 22 marks (baseline 24)
KILLED       a refusing instance is counted as one that reported
  red, and named: the refusal and the silence counted the same way · 12 marks (baseline 24)
KILLED       the notice is withheld from a partial census
  red, and named: a partial census produces a notice · 17 marks (baseline 24)
KILLED       the contract split is decided on the message instead of the code
  red, and named: a refusal whose MESSAGE reads like a mismatch but carries no code is not one either · 23 marks (baseline 24)
KILLED       the scatter stops carrying the refusal code to the renderer
  red, and named: naming the contract split once, with the pin, off the code the scatter carried · 23 marks (baseline 24)

```

The first mutation restores the shipped exit 0 and is killed by the cell that runs the real
command, so the status is graded through the command rather than through a helper.

### Other suites

Every suite that imports the two files this touches, plus the registry gates:

```
pnpm smoke:ps-mesh-column          23 passed, 0 failed
pnpm smoke:ps-provenance            7 passed, 0 failed
pnpm smoke:ep-rail-failure         25 passed, 0 failed
pnpm smoke:attach-open-mode        30 passed, 0 failed
pnpm smoke:scatter-pinned-probe    21 passed, 0 failed, 21 expected
pnpm smoke:cli-on-instance         23 passed, 0 failed
pnpm smoke:gate-inventory          GATE INVENTORY OK
pnpm smoke:ci-fragments             8 passed, 0 failed
pnpm smoke:ci-suites-freeze        10 passed, 0 failed  (with BASE/HEAD set)
pnpm typecheck                     clean across all packages
pnpm check:docs-voice              41 pages passed
node scripts/check-attribution.mjs --selftest && --range   14/14 fixtures, clean
node scripts/check-operator-literals.mjs --selftest && --range   65/65 cells, clean
node scripts/upgrade-section-gate.mjs --range   OK: no breaking commits in range
pnpm changeset status              @cotal-ai/cli at minor, plan verified
```

`server-resolution:live` and `attach-reconnect` also import these modules and were not run here;
they are live suites that need a full local stack.

## Files

- `implementations/cli/src/commands/agents.ts` — `psCensus` / `psCensusNotice`, and the notice plus
  exit at the end of `ps`'s multi-manager branch.
- `implementations/cli/src/lib/control.ts` — `ScatterInstanceReply.code` (the refusal's structured
  code) and `ScatterReply.pinned` (the contract this scatter pinned, and the instance whose
  describe supplied it), both read off values the scatter already had.
- `docs/cli.md` — the partial-listing rule and exit status, and the same-contract deployment rule.
- `implementations/cli/smoke/ps-partial-census.smoke.ts` + `mutations/ps-partial-census.json`,
  registered in `package.json` and `bin/smoke/ci-suites.d/`.
- `.changeset/ps-partial-census.md` — `@cotal-ai/cli` minor.

File and symbol references are against `main` @ `531581b9`.


davidfarah2003:
The `unit` job on 0226b884a fails at the last step, `pnpm check:docsbundle`, not in a test. The PR edits `docs/cli.md`, and the generated docs bundle in `extensions/connector-core/src/docs-bundle.generated.ts` still carries the previous `docs/cli.md` body, so the `git diff --exit-code` inside that script exits 1.

Regenerating at this head changes exactly one line in that file:

```
pnpm gen:tooldocs && pnpm gen:docsbundle
git add extensions/connector-core/src/docs-bundle.generated.ts
```

Commit that file with the branch and the job goes green. Every other job on this head is green or still running.

L4XB:
Rebased onto `main`. Two conflicts, both from #1640 landing:

`docs/cli.md` — #1640 added a top-level paragraph on endpoint-rail scoping at the same spot where this branch continues the indented `ps` list item. Unrelated topics, so both are kept: the partial-census paragraph stays inside the `ps` item, the rail paragraph follows it as its own top-level block.

`extensions/connector-core/src/docs-bundle.generated.ts` — generated, so it was regenerated with `pnpm gen:docsbundle` rather than hand-merged.

`check-operator-literals.mjs` reports `findings=0/2353_files status=clean`, and `check-attribution.mjs --range origin/main..HEAD` reports clean over both commits.

L4XB:
`changed shard 10/12` is red on `attach-reconnect`, and I have measured it rather than assuming. I do **not** think this is settled, so here is everything, including the part that argues against me.

## What CI saw

The reproof selected `bin/smoke/mutations/attach-reconnect.json` — correctly; this PR touches `implementations/cli/src/lib/control.ts` — and refused to grade it because the baseline is red at head:

```
✗ ...without re-establishing after the press, so this is the backoff path and not a reconnect
    {"reconnectsAtPress":0,"now":1,...}
✗ ...and it gives the shell back on the press, not when the rung it was waiting on expires
    {"pressToExitMs":27779,"boundaryInMs":26728,"spareMs":-1051}
FAIL - 70 passed, 2 failed
```

## What I measured locally

Three runs, each tree with its own `pnpm install --frozen-lockfile` and its own full `pnpm build` (the first attempt graded a stale `dist` from the rebase and died on `does not provide an export named 'unansweredRail'`, so this matters):

| tree | run | result |
|---|---|---|
| base `fc6f0b1d` | 1 | `PASS - 72 passed, 0 failed` |
| head `4f8ce28f` | 1 | `PASS - 72 passed, 0 failed` |
| head `4f8ce28f` | 2 | `PASS - 72 passed, 0 failed` |

Both cells that failed in CI pass on both sides, and there is not a single `✗` in any of the three logs. Same head SHA CI ran (`4f8ce28f1`).

## Why I am not calling it flake

`spareMs: -1051` is not a near miss against the 10 s margin — the shell came back **after** the next attempt was already due. The comment above that assertion records the original defect as *"27.0s measured with the next attempt 26.9s away"*, and CI measured 27.779 s against 26.728 s. That is the defect's own signature, not a slow-runner artifact, and a relative boundary does not protect against it if the press-to-exit path itself is delayed.

Against that: the harness declined to attribute it either, in the same log —

```
PRE-RED TRANSITION UNMEASURED (1 fixture(s)) — base comparison was absent or could not run; refusing to clear:
  attach-reconnect.json -> transition: UNMEASURED (root PRE-RED was not reproduced in the clean
  head snapshot; root execution-state contamination detected) -> head RED
```

It tried to reproduce the red on a clean head snapshot and could not.

And the diff is hard to connect to it. The `control.ts` change is two additive fields on the scatter reply — `ScatterInstanceReply.code` and `ScatterReply.pinned` — read off values the scatter already had. Nothing in it touches the attach loop, the detach press, or any timer.

## What I cannot rule out

I am on **macOS arm64** and the shard runs on **Linux**, so a platform- or load-dependent path is exactly what three green local runs would miss. This fixture was also on my "not run locally" list when I opened the PR, precisely because it needs a full stack — so CI is the first place it ran against this change at all.

A re-run of that shard would discriminate: if it goes green, it was contention on the runner; if it reds again with a similar `spareMs`, it is reproducible on Linux and I will chase it there rather than argue from macOS. I cannot trigger one myself — happy to push an empty commit if you would rather have a fresh run than a re-run of the same one.

davidfarah2003:
Panel complete: three independent reviews of `4f8ce28f15260495e6d2c62dca3e3c5a67cf60c0`.

Each review graded the merge of this head onto main in its own lane (merge tree `55dff7bac0e1d17a1795e5c3bbb69b56d4c6a2f4`). Two reviews approve and one requests changes. All three reproduced the defect at the merge base and the fix at the merge tree with a real broker and the real command (one instance's worth listed and exit 0 before; every enumerable seat listed, the partial-listing notice on stderr and exit 1 after), all three report the 24-cell suite green on repeated runs and the fixture killing every mutant, and all three confirm the generated docs bundle regenerates byte-identical. Two reviews also ran the real command as a subprocess with real pipes, which the pull request's own in-process suite does not, and found stdout under `--json` stays pure JSON lines.

The request-changes review reports one mechanism the two approving reviews did not probe: the scatter projection in `control.ts` is built from valid replies plus missing instances only, while core also classifies invalid and churned replies and marks the scatter incomplete for them. A registered instance whose only answer is an invalid frame is counted as having responded, so it is neither a reply nor missing, falls out of the projected list entirely, and the census counts it as nothing. Its three real-broker runs with one valid manager and one instance sending an on-time invalid frame gave exit 0, the valid manager's seats only, and empty stderr. The same review grades the `Closes #1637` keyword as blocking because the issue's third requirement (a join-time same-contract check or per-instance pinning) is stated in the body as not implemented. One approving review records the closing keyword as a non-blocking concern; the other grades closing as sound.

The two approving reviews graded the detection as sound over every shape the scatter returns; the request-changes review found a shape the scatter returns that the projection drops. That is the disagreement, and each review's own reproduction is below.

Findings the reviews share, graded non-blocking by those that raise them: the notice reads "N of M manager instances did not report its seats" for every count above one, and the `locateSeat` sibling counts a refusing instance as reachable and is reasonably left outside this change. One review adds that exit 1 on this command now covers a hard scatter failure, an auth refusal and a partial census, with no in-tree consumer reading it. Two reviews hit the build-prerequisite refusal the packet describes and attribute it to the repository rather than to this change; one never saw it because its lane already carried built output.

---

### Review 1: `APPROVE`

The change does what it claims, and I established that by driving it rather than by reading it. I reproduced the defect at the merge base with my own fixture (six seats running, one instance's worth listed, exit 0, stderr empty, three runs) and the fix at the merge tree (exit 1, the partial-listing notice on stderr, every enumerable seat still printed, three runs), with the two runs differing only in implementations/cli/src because both worktrees shared one installed node_modules. I also closed the gap the packet itself flags: the PR's suite calls ps() in process with console and process.exit [...]

**Blocking (0):** none

- The commit message and the changeset both describe the notice as naming 'how many of how many did not [report]'. The shipped string's possessive does not agree in number with its subject, so the rendered sentence is not grammatical for any count above one. The description is accurate; the string is not.
- Minor wording divergence between the commit body and the changeset covering the same facts ('hides a whole roster' versus 'hides every seat behind it'; 'swaps between calls' versus 'the output swapped between calls'). Harmless, noted because the two texts are otherwise near-identical.
- The PR body's central measurement claim, 'six seats running, one printed, exit 0', is CORRECT: I reproduced it independently at the merge base on three runs. I record this as a body claim I confirmed rather than a defect.

**Non-blocking (5):**
- The shipped notice reads 'N of M manager instances did not report ITS seats' for every M above one. The noun is pluralized and the possessive is not. Operator-visible.
- psCensus tests 'error === undefined' while the renderer tests truthiness of the same field; they disagree on the empty string. Not reachable through the shipped scatter today, so latent.
- The split cell asserts digest membership rather than order, so a reversed input/output pair survives. Confirmed SURVIVED with a same-file positive control.
- 'Closes #1637' would close an issue that still carries an unimplemented requirement (the join-time same-contract check) and a partially met one (both digest pairs in the plain line).
- Exit 1 is now overloaded on ps: it means a hard scatter failure, a failIfNotOk refusal, and a partial census. No in-tree consumer reads ps's status, so nothing breaks today, but a distinct code would have been machine-separable. The issue's own text permits either.

**Not verified (16):**
- I did NOT observe the ERR_MODULE_NOT_FOUND build refusal the packet describes. My lane already had packages/core/dist and packages/workspace/dist present before I touched anything, so the fresh-lane condition did not exist here. I neither confirm nor contradict that claim.
- I did NOT run the real binary as a subprocess at the MERGE BASE. Doing so would have required building the whole monorepo at the merge base, overwriting the lane's dist. I drove the merge base in process instead, which is sufficient to establish exit 0 there, and I ran the real binary only at the merge tree.
- I did NOT run 'pnpm smoke:ci' or any full CI shard. The packet states it stops at the harness on a docker-spawning cell and that neither docker nor podman is installed, and I did not test that claim.
- I did NOT run artifact-store or smoke:delivery-starvation, which the packet attributes to the resource-constrained root disk.
- I did NOT test a multi-manager space with more than three instances, nor a real two-build manager fleet outside a fixture.
- I did NOT probe the pathological stdout size at which process.exit could truncate a pending pipe write. I tested 2000 JSON lines (about 460 KB) through a pipe and to a file with no loss, on Linux only.
- I did NOT test any non-Linux platform.
- I did NOT verify the user-auth (bearer) branch of ps, which returns before the census is computed. The new exit applies only to the static/open class-scatter branch; a bearer-mode caller in a split space still gets no census signal. I note this as untested behaviour rather than as a finding.
- I did NOT test the --on <instance> branch, which also returns before the census.
- I did NOT execute any 'cotal run ...' workflow command; nothing in this review required one.
- I did NOT verify the contributor's fork state or any GitHub-side CI result.
- Refusal not counted as a failure. 'pnpm upgrade-section-gate' with no arguments printed a usage line and exited 2. The script documents exit 2 as misuse. INSTRUMENT REFUSAL, not a red. Re-run correctly with --self-test (95 cells passed) and --range (exit 0).
- Refusal not counted as a failure. My first subprocess attempt at the real binary failed with ERR_MODULE_NOT_FOUND for the package 'tsx', resolved from the fixture's cwd rather than from the repo. INSTRUMENT REFUSAL of my own making, not a result about the PR. Fixed with an absolute loader path.
- Refusal not counted as a failure. The second and third attempts failed with ERR_MODULE_NOT_FOUND for @cotal-ai/cli/dist/index.js and then @cotal-ai/manager/dist/index.js, because my scratch worktree had no built monorepo. Again my own instrument, not a red. Resolved by running 'pnpm build' in the lane (exit 0) and pointing the subprocess at the lane's bin/cotal.ts.
- Refusal not counted as a failure. My first mutation run of my own reversed-digest mutant returned 'did not produce a clean, named red' rather than SURVIVED. That is the tool refusing to draw a conclusion it had not earned, not a result. I added the same-file control it asked for and only then recorded SURVIVED.
- Refusal not counted as a failure. My first positive control for check:docsbundle was invalid: I perturbed the GENERATED file, which the gate regenerates before diffing, so it passed and proved nothing. I discarded it and ran the correct control (perturb the SOURCE page), which exited 1.

---

The reported contract-split and silence cases now fail closed with usable stderr and JSONL separation, and the named suite is stable and mutation-sensitive. Request changes because control.ts drops core scatter invalid and churn classifications before psCensus sees them. A real registered instance can answer with an invalid output frame, contribute no seats, vanish from the projected instance list, and make real cotal ps exit 0 with no partial notice. The issue also explicitly requires a join-time same-contract check or per-instance pinning, while this head implements neither complete option [...]

**Blocking (2):**
- Partiality detection drops invalid and churned registered slots. askManagerScatterEp builds ScatterReply.instances only from EpScatterResult.replies and EpScatterResult.missing. Core classifies malformed output replies in result.invalid and wrong-epoch or registration-advanced replies in result.churn. Those frozen registered slots are omitted, so psCensus totals only the surviving projection and can return silent=0 even when core reports complete=false. Three real-broker runs with one valid manager and one registered instance sending an on-time invalid ps output produced real cotal ps status 0, only the valid manager's two seats, and empty stderr. A positive-controlled core projection run showed valid complete=true and silent=0, invalid complete=false with one invalid entry but projected total=0 and silent=0, and wrong-epoch churn complete=false with one churn entry but projected [...]
- PR closes issue #1637 without completing requirement 3. Issue requirement 3 offers per-instance describe and pinning, or the same-contract deployment rule plus a second-manager join check. This head states the rule but omits the join-time check and does not implement per-instance pinning. Authenticated read-only issue retrieval shows requirement 3. The PR body says, 'The join-time check is not here' and calls this only 'the first half of the second option' while retaining 'Closes #1637'.

**Body and prose defects (4):**
- The PR retains Closes #1637 while acknowledging the issue's join-time same-contract requirement is not implemented.
- The PR evidence names merge base 531581b9, but the graded branch merge base is fc6f0b1da8093996eaa53f7fa567a3fac122647d.
- The body does not clearly distinguish its separate base-compatible fixture from the new 24-cell suite that imports psCensus.
- The body generalizes from refusal and silence to every instance that does not report seats, but invalid and churned scatter slots contradict that claim.

**Non-blocking (3):**
- locateSeat counts refusals as reachable. Three split-contract runs for each of stop and attach exited 1 but reported an absent seat across 2 reachable instances even though only one instance could answer ps.
- Partial-listing pronoun disagrees with plural noun. Real CLI output: '2 of 3 manager instances did not report its seats'.
- Mutation fixture build-prerequisite prose is inaccurate. Before builds, pnpm smoke:ps-partial-census refused with ERR_MODULE_NOT_FOUND for @cotal-ai/core/dist/index.js. Core and workspace builds made it runnable.

**Not verified (12):**
- The full smoke:ci matrix was not run.
- Docker and podman dependent cells were not run; neither runtime is installed on this host according to the packet.
- server-resolution:live and attach-reconnect were not run.
- No macOS or Windows platform run was performed.
- No GitHub write, PR comment, fork push, or reviewed-head edit was performed.
- The absent join-time same-contract check could not be tested as a feature because this head does not implement it.
- pnpm smoke:ps-partial-census before builds. ERR_MODULE_NOT_FOUND for core dist; documented build prerequisite, not a suite red.
- pnpm upgrade-section-gate --base fc6f0b1d in shallow clone. misuse exit 2, explicitly printed 'Nothing was graded'; rerun after git fetch --unshallow passed.
- pnpm changeset status in shallow clone. could not find divergence from main; rerun after full history passed.
- web fetch of the issue and initial gh repository guess. anonymous URL returned 404 and the first repository name was wrong; authenticated read-only gh retrieval from Cotal-AI/Cotal then succeeded.
- reviewer mutation attempt while reviewer harnesses were present. mutation-proof correctly refused a dirty tree; no result was counted.
- early custom harness launches before package-local dependency setup. module-resolution and CJS top-level-await setup refusals; only the corrected ESM package-context runs were counted.

---

### Review 3: `APPROVE`

PR #1644 cleanly and soundly fixes issue #1637. A partial census in cotal ps now prints a clear, structured notice to stderr (with detailed pinned contract diagnosis when the refusal is a contract-mismatch) and exits 1 instead of 0, while keeping stdout clean for --json consumers. All 24 test cells pass across multiple runs, all 5 mutants are killed, and docs/changeset are accurate.

**Blocking (0):** none

**Non-blocking (0):** none

**Not verified (4):**
- smoke:ci shard stopping at the harness (requires docker/podman which is not installed on this host)
- JetStream disk storage tests (omitted due to root filesystem storage constraints at 96% capacity)
- pnpm smoke:ps-partial-census (fresh lane before build). Repository convention: tsx packages require building dist prerequisites for monorepo peer dependencies before running test scripts.
- pnpm upgrade-section-gate --base fc6f0b1da8093996eaa53f7fa567a3fac122647d. Repository clone is shallow (depth-1), which triggers intentional exit 2 guard in upgrade-section-gate.

davidfarah2003:
Merge basis for `4f8ce28f15260495e6d2c62dca3e3c5a67cf60c0`: not met. Two of three independent reviews approve and one requests changes with two items: the scatter projection in `control.ts` omits instances whose only reply was an invalid frame, so a registered instance that answers with malformed output falls out of the census and the command exits 0 with no notice, reproduced three times with a real broker; and the closing keyword would close #1637 while its third requirement (a join-time same-contract check or per-instance pinning) is stated in the body as not implemented. The two approving reviews graded detection over the shapes they drove and did not drive an invalid frame, so the disagreement is about which shapes were driven, and it is left to the reviewers' own words above. Under the standing rule an unanswered request for changes gets no new panel; a revision addressing it goes back to the requesting review for re-grade. This head stays open for the contributor; nothing here is pushed to the branch by anyone else.

L4XB:
`changed shard 10/12` is green on the re-run — 30 checks passing, nothing red.

So the `attach-reconnect` baseline failure was contention on the runner, not this diff. I'd left that open rather than calling it, because `spareMs: -1051` carried the original defect's own signature and three green local runs on macOS could not rule out a Linux-only path. The re-run settles it in the direction the harness's own `PRE-RED TRANSITION UNMEASURED … root execution-state contamination detected` pointed.

Nothing was changed to achieve that; the head is the same `4f8ce28f` the failing run used.

davidfarah2003:
This branch no longer merges with `main`. Two files conflict, and neither is your production code.

```
CONFLICT (content): extensions/connector-core/src/docs-bundle.generated.ts
CONFLICT (content): package.json
```

The bundle is generated from `docs/workflows.md`, and #1633 regenerated it from the same page
after this branch was cut. Rebase onto `main`, then re-run the generator and commit its output
rather than resolving that file by hand.

`package.json` conflicts on the smoke script list, where both sides appended an entry. Both
entries should survive; keep them in the order the rebase gives you.

I have not pushed to your branch.

Separately, a generated artifact that is checked in will conflict between any two branches that
touch the page it is generated from, whether or not the edits overlap. Three open pull requests
are currently sitting on that one file. That is a repository-shape problem and is being tracked
apart from this PR.

PR #1645, opened 2026-09-15
fix(core,cli,runtime): keep the versioned rail's subject token out of source comments

davidfarah2003:
## What was wrong

`implementations/auth/smoke/issued-profile-census.smoke.ts` has a cell that scans every shipped source under `packages/*/src`, `implementations/*/src`, `extensions/*/src` and `bin` for the versioned rail's subject token and requires that only core's four subject and grant builders spell it, so a second unpinned way onto the rail cannot appear unnoticed. #1640 wrote that token into comments in five other shipped files, and the cell now fails on `main`:

```
FAIL: the issued rail is spelled only by core's endpoint subject and grant builders
+   'implementations/cli/src/commands/update.ts',
+   'implementations/cli/src/lib/control.ts',
+   'implementations/runtime/src/run-command.ts',
+   'packages/core/src/endpoint-error.ts',
+   'packages/core/src/endpoint-invoke.ts',
```

The suite is not on a CI shard, so `main` is green. It is the baseline of `implementations/auth/smoke/mutations/issued-profile-census-prototype.json`, and the reproof at the #1640 head did not select that fixture because the diff touched none of its suite paths. The next pull request that touched `packages/core/src/provision.ts` selected it, and the reproof refused to clear a baseline that was red at both base and head.

## The change

The comments in those five files say "the versioned rail" or "the versioned plane" where they spelled the token: eight comment lines in seven comment blocks, no code change.

## Evidence

The changeset is a patch: comment text only, no behaviour change.

At this head on Linux:

```
tsx implementations/auth/smoke/issued-profile-census.smoke.ts   -> OK (was FAIL at fc6f0b1da on the cell above)
node scripts/mutation-proof.mjs --config implementations/auth/smoke/mutations/issued-profile-census-prototype.json
  -> All 26 mutation(s) killed. The suite discriminates.
pnpm typecheck   -> exit 0
pnpm check:docs-voice   -> exit 0
```

## Not done here

The census cell can only fail a pull request whose diff selects its fixture for the reproof. A shipped-source scan that guards every file is worth running on every merge tree; putting the census suites on a CI shard is a separate change to the frozen suite list.

Refs #1640.


davidfarah2003:
Panel complete: three independent reviews of `10546445b741d170b417d8a2d9db2f8ff11a6953`.

All three reviews approve with zero blocking items, so this head is recorded as **green** on review. Each review ran the census cell at main and at this head and saw the transition itself (the cell red on main naming exactly the five files this head touches, green at head across repeated runs), read every changed line to confirm no executable line moves, and reran the mutation fixture with the repository's own tooling (26 of 26 mutants killed on each clean run). The corpus question in the brief (two files under the smoke tree of bin still spell the token) was settled by reading the scanner: it refuses any directory named smoke by name, the exclusion is documented in the scanner itself, and the published package's file list does not ship that subtree, so those files are outside the corpus and cannot be a second way onto the rail. Two reviews state that widening the cell's allowed set would have been the wrong remedy because the set's value is that it is closed and small. One review went further than reading to establish that no executable line changes: it compiled the three changed packages with comments stripped at this head and again with main's source content and found the emitted trees byte-identical across 222 files, with a positive control showing the same comparison detects a one-token change.

Two prose findings are recorded, neither blocking. One review found that the rewritten comment at `implementations/cli/src/commands/update.ts:233` keeps backticks around a sentence the program never emits, since the headline interpolates the rail token and the repository's own rail-failure suite pins that literal; the same review verified the emitted string with that suite twice. That comment is corrected in a follow-up pull request rather than by amending this head. Another review found the changeset and commit prose say "five comments" where the diff changes eight comment lines in seven comment blocks across five files; the pull request body and the squash message carry the corrected count, and the changeset text is left as is because it ships only into changelog prose. One review notes, as follow-on work outside this change's scope, that the census cell is still on no CI shard, so the same silent red can recur.

One review's first pass ran its lane checks and proofs in another reviewer's lane at the same head and byte-identical tree; it re-ran both arms and the mutation fixture inside its own lane before its verdict was accepted, and its report records the correction.

---

### Review 1: `APPROVE`

Approve with one non-blocking prose defect reported below (finding F1). The stale quotation is introduced by this head's rewrite, though it is confined to comment text and changes no behavior. I report it and did not edit it.

**Blocking (0):** none

**Non-blocking (3):**
- F1. The rewritten comment presents a quoted operator string that the program never emits. implementations/cli/src/commands/update.ts. 233. The backticks mark this as a verbatim quotation of the headline the caller saw. The headline is built in implementations/cli/src/lib/control.ts line 269 as `no manager answered on the ${versioned} rail (${detail})${skew}`, where `versioned` is the rail value off the unanswered marker. That value comes from epPlaneTokens(caller).join(".") in packages/core/src/endpoint-verbs.ts, which yields the literal string "ep.v1" for an issued caller. endpoint-verbs.ts line 251 is the only producer that sets the rail [...]
- F2. Four rewritten lines now exceed the surrounding comment width. The substitutions are longer than the tokens they replace and the lines were not re-wrapped. Measured lengths: control.ts:233 = 110, control.ts:249 = 107, endpoint-error.ts:81 = 111, endpoint-error.ts:92 = 111, endpoint-invoke.ts:116 = 112, update.ts:233 = 106. The 95th-percentile comment line in each of those files is 101 to 102 characters, so these sit just past the file's own prevailing width. Mentioned for completeness, not as a change request.
- F3. The cell remains outside every CI shard, so this class of red can recur silently. The suite has no smoke:* script entry in package.json, and the documented way to run it is the direct tsx invocation named in packages/core/smoke/prototypes/README.md. I verified its absence from the CI chain at the source of truth rather than at the runner: bin/smoke/shard.mjs composes its list from the frozen bin/smoke/ci-suites.txt plus the one-file fragments under bin/smoke/ci-suites.d/, and grepping both for issued-profile-census returns nothing. This matches the brief's and the commit message's account of why main was green while the cell was red.

**Not verified (0):** none

---

### Review 2: `APPROVE`

**Blocking (0):** none

- P1_COMMENT_COUNT. The prose says five comments spelled the token and later calls the replacements the five comments. The diff changes five source files, but those spellings occupy eight changed comment lines across seven separate comment blocks: one in update.ts, two in control.ts, one in run-command.ts, two in endpoint-error.ts, and one in endpoint-invoke.ts whose block spans two changed lines. This does not affect code, the census property, package versioning, or the reproduced red-to-green transition. It is a small factual count error in prose about a prose-only change, so I report it but do not withhold approval. Say comments in five [...]

**Non-blocking (0):** none

**Not verified (2):**
- The full repository smoke and CI shard matrix was not run because the grading brief asks for the shipped census cell and mutation fixture, and documents unrelated storage-sensitive suites on this host.
- GitHub state was not read or changed. No comment, review, label, CI re-run, push, or other GitHub write was performed.

---

### Review 3: `APPROVE`

**Blocking (0):** none

**Non-blocking (0):** none

**Not verified (0):** none

davidfarah2003:
Merge basis for `10546445b741d170b417d8a2d9db2f8ff11a6953`: three independent reviews, all approve, zero blocking items; CI green on every required check including the mutation reproof, which now measures the census fixture green against main from this head. The one comment-accuracy finding (a rewritten comment in the update command that quotes a headline the program never emits) is corrected in follow-up pull request #1647 rather than by amending this head, so that the reviewed head merges as reviewed and the four pull requests blocked on the census baseline can re-run their reproofs.

Issue #1616, opened 2026-09-14
Durable workflow spawn cannot select an existing absolute working directory


davidfarah2003:
## Current behavior

Cotal Lang accepts a logical worktree on a durable spawn:

```js
const builder = await spawn("builder", { worktree: "change-42" })
```

The worktree value is a logical ownership identifier. The runtime uses it to prevent concurrent agents from claiming the same logical tree. It is not forwarded as the child process working directory.

A physical directory cannot currently be expressed on the workflow spawn effect. `cwd` is rejected as an unknown option, and the manager uses its workspace root when the spawn request omits a directory. As a result, a durable run cannot place a new seat in an existing prepared checkout through the workflow interface.

This is an absent capability. It is not a regression and not a security exploit.

## Requested contract

Add a physical placement option, proposed as `cwd`, without changing the meaning of logical `worktree`.

The complete contract should include:

1. `cwd` names an existing absolute directory on the serving manager's host. The feature does not create a clone, select a branch, or install dependencies.
2. A request carrying a host local path is pinned to the intended manager through explicit target manager affinity. It must not be accepted by an arbitrary manager.
3. The resolved directory participates in the durable spawn step hash and idempotent spawn identity. Changing it during replay produces a migration divergence rather than moving a live seat.
4. A resumed run recovers the same seat and directory without starting a second process.
5. A missing directory, malformed value, or unavailable target manager is refused explicitly. There is no fallback to the manager workspace root and no partial writer remains.
6. Existing programs that use only logical `worktree` retain their current behavior and exclusivity guarantees.
7. Physical aliases of one writable checkout cannot bypass the single writer rule. The normalization and identity policy must be specified and tested.

## Acceptance behavior

- A durable run spawns a seat into a prepared checkout that differs from the manager workspace root. The child process directory and repository revision match the requested directory.
- Replaying the accepted step returns the same lifecycle and directory without another process.
- Editing `cwd` causes a deterministic divergence.
- Invalid or unreachable placement refuses without a fallback seat or duplicate spawn.
- Concurrent logical reservations remain exclusive.
- A valid mutant drops directory forwarding while leaving the program parseable. The seat then starts in the wrong directory and the placement assertion fails.
- Integration tests use isolated brokers and own every child on success, refusal, cancellation, and timeout.

## Related issues

- #741 covers directory configuration in an agent file and silent ignored frontmatter.
- #383 covers routing host local paths when multiple managers can answer.

[...]

PR #1646, opened 2026-09-15
fix(runtime): re-issue a not-executed bind refusal instead of failing the run

### LINKED ISSUE #1638: lang: an unpinned spawn inside a durable run fails the run when the class queue picks the other manager instance
## What happens

In a space served by two manager instances, a durable Cotal Lang run whose program contains a `spawn` effect with no `placement` fails at its first spawn, and the failed run stays on the run table. Each per-command refusal is honest and atomic; composed into a run they destroy it.

Measured against two live managers (A and B), four consecutive `cotal run start` attempts of the same program:

1. refused: the request reached B, the caller was bound to A, nothing ran
2. refused, identical
3. refused with the polarity inverted: reached A, caller bound to B. The same client, which had never addressed B, now considered itself bound to it
4. `started run <id> on the manager`

The started run has three journal records: activation, `spawn` pending, `spawn` settled failed with

```
L4000 spawn(<persona>) was refused by the manager endpoint: this request reached manager instance <B>, but the caller bound to <A>; the class queue chose a different member and "spawn" WAS NOT RUN - no effect of it exists here. Re-resolve and re-issue, or address one instance (SPEC 13.2)
```

`run ps` shows the run as `failed`. `run journal` on it was refused by the same class-queue choice (reached A, bound to B, rc 1), so the journal had to be read by raw stream replay. No seat was created by any attempt.

Two related faces of the same choice, observed in the same session:

- `cotal agent stop <seat>` returned "no managed agent on any of the 2 reachable manager instances" twice for a seat that a three-call union of `cotal ps` still listed as running; the third stop succeeded. A single stop result therefore cannot be read as "already gone".
- `cotal run status` and `cotal run journal` return the same 13.2 refusal at random, so a run cannot be reliably observed either.

## Why

A run binds to the instance that answered its start (`this.binding`), but a `spawn` effect that names no `placement` resolves the manager through `managerService`, the class-anycast resolution (`implementations/runtime/src/mesh-handler.ts`, `manager()`), and `invokeCommand` rides the class `one` queue. The serving instance's bind check (`packages/core/src/endpoint-serve.ts`, the `b.instanceId !== identity.instanceId` refusal) correctly refuses when the queue delivered the request to the other member. The handler maps that refusal to `L4000` as the effect's own failure (mesh-handler.ts, the `reply.reply.ok === false` branch), and an uncaught `L4000` ends the run.

The per-command contract holds: nothing was executed and the refusal says so. The run-level contract does not: a `not-executed` refusal whose remedy is "re-resolve and re-issue" is turned into a terminal failure instead of being re-issued, and the run id and its journal are consumed.

The bind check compares instance ids only, so this does not depend on the two instances serving different contracts. By the code, two managers of the same build for one space would produce the same outcome for every unpinned spawn.

`placement: { endpoint, instanceId }` on `spawn` (#1628) pins a single effect and sidesteps the queue, and `cotal spawn --detach --on <instance>` pins a CLI spawn. `cotal run start`, `run status`, `run journal`, `run answer`, and `agent stop` have no way to name an instance, and a program written before #1628 has no placement at all.

## What is required

1. A `not-executed` bind refusal (13.2) on an effect inside a run is re-resolved and re-issued by the handler, bounded, rather than raised as `L4000`. The refusal already says no effect exists, which is exactly the condition under which a re-issue is safe.
2. `run start`, `run status`, `run journal`, `run answer`, and `agent stop` either pin to the instance the run or seat is bound to, or accept an explicit instance, so that observing and stopping do not depend on which member the queue picks.
3. `agent stop` reports "not found" only after every reachable instance has answered for itself; a refusal from one instance is reported as unknown, not as absent.
4. A run that dies this way is recorded as `not-executed` at run level, or the failed run is cleared, so that retries do not accumulate failed runs.

## Reproduction

Two managers for one space, a program whose first effect is an unpinned `spawn`, `cotal run start` repeated until one start succeeds, then `cotal run ps` and a raw journal read of the started run.

## Addendum: the pinned-spawn workaround covers placement only

Placing the seats with `cotal spawn --detach --on <instance>` works: every seat lands on the named instance first time. The durable half then fails on its own rule. An ask-only program (no `spawn` effect, so the class queue cannot fail it) that addresses one of those seats by a correct `<name>#<lifecycleUid>` handle is refused with

```
L4000 ask(...) addresses <name>#<lifecycleUid>, which is not in this run's roster
```

A run's roster contains only the agents that run spawned. So as Lang stands the two halves exclude each other: pin the seat and the run cannot ask it; let the run spawn it and the spawn rides the class queue. Four more failed runs were spent establishing this. Requirement 2 above should therefore also cover `ask` on an adopted or pre-placed agent, or `run start` should accept a placement that every `spawn` in the program inherits.

## Addendum 2: the not-found face is on `cotal input` too, so every managed-agent command needs a pin

`cotal input --name <seat>` without `--on` answered "no managed agent <seat> on any of the 2 reachable manager instance(s) in this space" for two seats that a three-call union of `cotal ps` listed as running on the same instance the whole time. The same command with `--on <instance>` succeeded first time for both. The message is a definite negative that names the instance count, which is the shape a reader believes. Requirement 2 should therefore cover every command that addresses a managed agent, and a "not found" answer should be reported only when every reachable instance answered for itself.

`cotal spawn --detach` also submits no first turn, so a pinned seat needs a separate pinned `input` to start work. That is not a defect in itself, but it means the workaround is three pinned commands per seat rather than one.

## Addendum 3: a spawn that the manager never finishes provisioning can be neither stopped nor respawned

A pinned `cotal spawn --detach --on <instance>` for one seat name returned "launch status uncertain: no process exit and no mesh presence within 180s". The manager's log has exactly one line for that name, `static retirement <name> (<uid>): timeout`, with no eviction precursor; every other retirement timeout in that log (five of 5714 retirement lines) was preceded by the manager's "verify-evicting orphan seat principal" step. `cotal agent stop <name>` refuses with `durable state contradicts absence [static-slot slotPhase=provisioning ... consistency=ordered-not-atomic]`, and a second `spawn` under the same name is refused because the slot exists. The name is therefore held by a slot the manager will not retire and will not reuse, and the only way forward was a different name.

Requirement 3: a slot that has stayed in `provisioning` past the launch deadline is retired by the manager on its own (the same timeout that produced the log line should also free the name), and `agent stop` on a provisioning slot should succeed as a cancellation rather than refuse.

`cotal input` also takes `--name` and `--text` as flags. A positional call fails with a usage error that prints the `--on` help line, so it reads exactly like the not-found fault above; a retry loop cannot tell them apart. Worth a distinct exit code.

## Addendum 4: a name stays reserved after a launch that exited on its own, and a lost first input is indistinguishable from an idle seat

Three more faults on the same pinned-spawn path, all on one seat name within two minutes.

1. A pinned `cotal spawn --detach --on <instance>` launched a seat whose harness exited during startup ("host startup failed (unknown)"; the harness log shows no provider request and a SIGTERM about 30 s in). The manager reaped it as "its own process exited and this manager did not stop it". The launch failure was real and the census agreed, so this one is a plain failed spawn, not an uncertain one.

2. The retry under the same name was refused: "the name <seat> is reserved pending retirement ... the static retirement did not complete". The static retirement had stalled on the delivery-admin rail ("the delivery daemon is not reachable ... no responders"), so the name stayed reserved indefinitely and the only way forward was a different name. That is the second time today a name has been stranded this way (the first was the provisioning case in Addendum 3, which also required a renamed respawn).

Requirement 4: a seat whose process exited is retired even when the delivery rail does not answer; the retirement must not depend on a responder that the same outage takes down. Until then the failure message should say which step stalled and that the name will not free itself, so an operator does not retry on a timer.

3. The pinned `cotal input --on <instance> --name <seat> --text <order>` for a freshly spawned seat printed "sent N bytes" and the seat's journal never received it: the seat sat at two provider requests for six minutes with no user row until a second identical send landed. A detached spawn submits no first turn, so a seat whose first input is lost looks exactly like a seat that was never briefed, and nothing on the sending side distinguishes the two. Separately, a `cotal ps --on <instance>` during the same window exited 1 with "unavailable: the contract-store read for <id> failed (a failed observation is never absence ...)". That refusal is the right shape; it is listed here only because it appeared on the same path and a retry loop that treats exit 1 as "no seats" would have concluded the fleet was empty.

Requirement 5: `cotal input` reports delivery only when the target seat has accepted the turn (a journal write or an equivalent acknowledgement), not when the bytes left the CLI; a send that the seat did not accept must fail loudly so the sender can resend once instead of polling a journal.

L4XB:
Refs #1638. This change addresses its first requirement. The remaining requirements and addenda stay open.

A durable run whose program contains a `spawn` with no `placement` fails at that spawn whenever the
class queue hands the submission to a manager instance other than the one the run resolved against.
This takes the first of the issue's four requirements: the refusal is re-issued rather than raised
as the effect's own failure. Requirements 2 to 4 and the three addenda are not in this change; the
last section says what I found about them.

## The mechanism

Two independent trips through the same anycast queue.

1. The run resolves the manager through `resolveService` -> `describeEndpoint`. With no `instanceId`
   the describe rides the class rail: `packages/core/src/endpoint-invoke.ts:113` builds
   `{ mode: "one" }`. Whichever member wins that queue becomes `ResolvedService.responder`.
2. `invokeCommand` then sends the command on the class rail as well, carrying
   `bind: { instanceId: responder.instanceId, epoch }` (`endpoint-invoke.ts:406`). That is a second,
   independent draw.
3. When the two draws disagree, the serving instance refuses at the pre-effect seam:
   `assertBoundIncarnation` (`packages/core/src/endpoint-serve.ts:140`, the
   `b.instanceId !== identity.instanceId` branch at `:156`) raises `failed-precondition` carrying
   `ai.cotal.ep.bind-refused` and `outcome: "not-executed"`.
4. The mesh handler maps that reply to the effect's own failure at
   `implementations/runtime/src/mesh-handler.ts:1339`, which becomes `L4000` at `:1346`, and an
   uncaught `L4000` ends the run.

Every step is correct on its own. The composition is not: a refusal whose stated remedy is
"re-resolve and re-issue" is turned into a terminal run failure, and the run id and journal are
spent.

The same handle is used for `turn` as well (`mesh-handler.ts:245`, `manager()`), so a run that
survives its spawn meets the same draw at its first turn.

## Why the caller's binding flipped (item 3 in the report)

This is the report's strangest fact and it has a mechanical answer: `bind` is not something the
client chooses.

`bind` is `ResolvedService.responder`, which is whichever instance answered that attempt's own
`describe`, and the describe is class-anycast too (`endpoint-invoke.ts:113`). So "the caller bound to
B" never meant the client addressed B. It meant B won the describe queue on that attempt.

Attempts 1 and 2 drew A for the describe and B for the invoke. Attempt 3 drew B for the describe and
A for the invoke. Nothing in the client changed between them, and nothing could have: each
`cotal run start` is a fresh process with a fresh resolve, so there is no client state that could
carry a binding from one attempt to the next. The polarity inversion is two independent coin flips
landing the other way, not a client that learned a new binding.

That also explains the `run status` and `run journal` refusals arriving "at random", and the
`agent stop` that failed twice and then worked: same two draws, same disagreement rate.

## The change

`MeshHandler.invokeManager` performs the manager calls a run makes, and a refusal the responder
marked as pre-effect is re-issued instead of returned:

- the reply is a refusal, and `replyRefusedBeforeEffect` holds (the `ai.cotal.ep.bind-refused`
  marker together with `outcome: "not-executed"`, which is the responder's own statement that no
  effect of the command exists),
- the stale class handle is dropped so the re-issue re-describes rather than rebinding the
  incarnation that was just refused,
- the call goes out again, up to `BIND_SPLIT_REISSUES` times, after which the refusal surfaces
  unchanged and still states that the command did not run.

This is the licence `Endpoint.invokeService` already re-issues on (`packages/core/src/endpoint.ts:2202`),
applied to the calls a run performs rather than only to the ones a client makes. The re-issue is a
first attempt and not a second, so nothing can be duplicated by it. It covers `spawn`, `turn`, the
relay a paused `ask` or `checkpoint` submits, and the `despawn` a cancelled spawn discharges with.

A handle pinned to one instance is never repaired: it addresses that incarnation by name, so a
refusal from it is that instance answering about itself, and re-resolving onto the class rail would
be the anycast fallback #1616 removed.

### What this does not do

The re-issue draws the same queue, so it converges rather than eliminating. A space of `m` managers
still splits `(m-1)/m` of the time per attempt; nine attempts leave two managers a 1-in-512 residual
where the unrepaired refusal was 1-in-2.

The deterministic fix is to address one instance, and the run's caller cannot: `run-driver-grants.ts`
mints instance-rail rows only when the program named a placement (`:137`, over the three
`PLACEMENT_COMMANDS` at `:72`), and its describe row is class-only
(`epDescribeAllGrantRow`, `packages/core/src/endpoint-grants.ts:317`, which is
`ep.one.*.describe...`). Pinning an unplaced run's spawn would need that grant widened, which is the
issue's requirement 2 and a wider change than this one.

## Evidence

Two live manager instances serving one endpoint, in `mesh-spawn.smoke.ts`, which already stands up a
second instance for #1616's matrix.

**Forced arm, deterministic on both tips.** Which member wins the queue is the run's luck, so an arm
that waits for a split can decline to grade. The handler's memoized class handle is rewritten to bind
an incarnation nothing is serving, which is the same device `describe-split-duplicate-effect.smoke.ts`
uses, so the refusal is guaranteed. Only one instance is serving in that arm, so the re-issue's
re-resolve can land nowhere else.

Base (`implementations/runtime/src/mesh-handler.ts` at the merge base, the new cells added):

```
• 3f — an unpinned spawn survives a not-executed bind refusal
  ! the unpinned spawn rejected: spawn(builder) was refused by the manager endpoint: this request reached manager instance mmmmmmmmmmmmmmmmmmmmmmmmmm, but the caller bound to zzzzzzzzzzzzzzzzzzzzzzzzzz;
  ✗ FAIL: an unpinned spawn refused as not-executed is re-issued and returns a handle, instead of failing the run { agent: undefined }
  ✗ FAIL: the repair submitted once and allocated one seat: a re-issue after a not-executed refusal is not a second attempt { allocations: 0, submissions: 0 }
     12 unpinned spawns, two managers serving: 5 returned a handle, 7 did not; 4 landed on A, 1 on B
mesh-spawn.smoke: 68 passed, 2 failed
```

Head:

```
• 3f — an unpinned spawn survives a not-executed bind refusal
     12 unpinned spawns, two managers serving: 12 returned a handle, 0 did not; 5 landed on A, 7 on B
mesh-spawn.smoke: 70 passed, 0 failed
```

The natural arm is the second line: twelve unpinned spawns through the ordinary entry point with both
instances live, nothing rewritten. Its assertions are the conservation that holds on any correct tip
(one submission per handle returned, nothing submitted by a refused attempt) plus a positive control
that the queue really did spread across both members, because a tip where all twelve landed on one
member would otherwise score green while proving nothing. The handle counts are what the repair
moves: 5 of 12 at base, 12 of 12 at head, repeated three times at head with the same result.

The two trees are separate worktrees with separate builds, each resolving `@cotal-ai/core` to its own
`dist` and reading its own `implementations/runtime/src/mesh-handler.ts`:

```
wt-1638-base:  invokeManager present: 0   BIND_SPLIT_REISSUES: (absent)
               core -> .../wt-1638-base/packages/core/dist/index.js
wt-1638:       invokeManager present: 1   const BIND_SPLIT_REISSUES = 8
               core -> .../wt-1638/packages/core/dist/index.js
```

**Every suite that names a source this diff touches**, base and head, on a loopback broker:

| suite | base | head |
| --- | --- | --- |
| `smoke:runtime-mesh-spawn` | 1 (the two new cells) | 0 |
| `smoke:runtime-mesh-turn` | 0 | 0 |
| `smoke:runtime-mesh-ask` | 0 | 0 |
| `smoke:runtime-mesh-replied` | 0 | 0 |
| `smoke:runtime-mesh-wait` | 0 | 0 |
| `smoke:runtime-mesh-conclave` | 0 | 0 |
| `smoke:runtime-mesh-monitor` | 0 | 0 |
| `smoke:runtime-mesh-notify` | 0 | 0 |
| `smoke:runtime-mesh-sleep` | 0 | 0 |
| `smoke:runtime-mesh-checkpoint` | 0 | 0 |
| `smoke:runtime-mesh-worktree` | 0 | 0 |
| `smoke:runtime-mesh-seam` | 0 | 0 |
| `smoke:runtime-placement-grants` | 0 | 0 |
| `smoke:runtime-host-starvation` | 0 | 0 |
| `smoke:runtime-run-driver` | 0 | 0 |
| `smoke:runtime-run-driver-auth` | 0 | 0 |
| `smoke:runtime-run-command` | 0 | 0 |
| `smoke:runtime-run-command-usage` | 0 | 0 |
| `smoke:runtime-migrate` | 0 | 0 |
| `smoke:runtime-fork` | 0 | 0 |
| `smoke:runtime-spawn-policy` | 0 | 0 |
| `smoke:lang-options` | 0 | 0 |

**Mutation fixtures.** Two mutants were added to `mesh-spawn.json`, both named against the forced
cell so the red is deterministic: the refusal returned to the effect instead of re-issued, and the
re-issue keeping the handle that carried the refused bind instead of re-describing. The second is
the one a probabilistic arm cannot grade, because a re-issue that never re-resolves still converges
whenever the queue cooperates.

`mutation-reproof --list-shards` selects 17 fixtures for this diff. All 17 were re-proved serially,
**186 of 186 mutations KILLED, no survivors and no unmeasured verdicts**, every fixture reporting
`baseline green` before it mutated anything:

```
mesh-spawn 23/23   mesh-turn 23/23   mesh-ask 22/22   mesh-replied 12/12   mesh-worktree 13/13
host-starvation 14/14   run-host-isolation 19/19   mesh-monitor 11/11   cancel-discharge 8/8
mesh-conclave 8/8   wait-position 7/7   resume-attach 6/6   fire-taken 5/5   wait-rearm 4/4
mesh-notify 4/4   spawn-policy 4/4   mesh-seam 3/3
```

## The rest of the issue

**Requirement 4, whether the failed run stays on the run table deliberately.** Deliberately. The
driver classifies its exits: `RunSuperseded`, `RunReleased` and `RunHeld` record `released`, and
everything else records `failed` (`implementations/runtime/src/run-driver.ts:690`, "a program that
FAILED is a fact about the program, and the run record is where a reader looks for it"). An `L4000`
is a program failure by that classification, so the row says `failed` and is doing what it was built
to do. What the requirement is asking for already has a name here: `EffectRefused` settles the entry
`refused` and halts with `RunHeld` (L5025), which is "nothing was attempted, a resume performs the
step live". Wiring an exhausted bind refusal to that disposition would satisfy requirement 4 exactly,
but no shipped handler raises `EffectRefused` today (only the lang smokes and the engine bridge), and
the only existing code for it is L5016, "effect not durable on this host", which is not what a class
split is. That looked like a language-disposition decision to make deliberately rather than inside
this fix.

**Requirements 2 and 3, and addenda 2 and 3**, are CLI and manager work on a different surface
(`run start/status/journal/answer`, `agent stop`, `input`, and a slot stuck in `provisioning`). They
are untouched here.

**Addendum 1** notes that `ask` on a pre-placed seat is refused because a run's roster holds only the
agents it spawned. With this change the other half of that deadlock is gone: an unpinned spawn inside
a run no longer has to be pinned to survive, so the program can spawn the seat and ask it. The roster
rule itself is unchanged.

One adjacent thing worth recording: in `turn`, a serve-boundary `expired` is mapped to L4002, "the
agent went down" (`mesh-handler.ts`, the `refusal` helper). A bind refusal for the same instance at
another epoch also carries `expired`, so that mapping can report an agent as down when the truth is
an incarnation mismatch. This change makes that case rarer by absorbing it, and does not fix the
mapping.

File and line references are against `main` @ `712dfadfe1e4593810614303170aa876ad3a0d30`.


davidfarah2003:
Checked source head `cd41bb2eb7bfdf98ad2f6c25313f805906c344c5`.

This change re-resolves and reissues manager calls after a marked bind refusal confirms that no effect executed. Explicitly pinned requests remain pinned. The remaining control, observation, retirement and failed-run requirements in #1638 remain open.

The integrated tree was built and its runtime mesh-spawn suite was executed independently with isolated brokers. Every repetition passed, and each recorded broker was absent afterward. The source tree stayed unchanged.

| Integrated-tree validation | Result |
| --- | --- |
| Runtime mesh-spawn repetitions | 3 |
| Assertions per repetition | 70 passed, 0 failed |
| Recorded broker processes remaining | 0 |

The forced-refusal tests and mutation checks cover reissuing a refused call and discarding its stale binding. The fixture endpoints used by these tests do not establish that all production multi-manager workflows work. The aggregate mutation total in the PR description is not the basis for this conclusion.

Remaining limitations:
- Retry limits apply per manager call and can extend existing timing expectations. Exhaustion and re-describe failure still need permanent regression coverage.
- Runtime recovery lacks the recovery counter or event used by the core path, and its handling of unexpected refusal codes differs.
- The natural dispatch control depends on queue distribution and can fail statistically.
- Workflow documentation should state the numeric retry bound and cover all affected call sites.
- The changeset omits the package containing the modified generated documentation bundle.

The description now uses `Refs #1638`. Landing is restricted to an explicit replacement squash message without an issue-closing keyword. The landed message, parent, tree and ancestry must pass post-merge checks, and #1638 must remain open.

PR #1688, opened 2026-09-17
test(scripts): require each entry-guard parity probe to reach main()

L4XB:
Refs #1622. Follow-up to #1631.

#1631 merged with a blocking test gap from its review. The six parity cells in `scripts/entrypoint-guard.selftest.mjs` compared the direct run with the symlinked run, and nothing else. Two runs that both skip `main()` agree perfectly: exit 0, nothing on either stream. So the suite could not see the failure it exists for.

This branch closes that gap and records what the suite still does not cover. It is test-only: no script, helper or published package changes.

## The gap, reproduced on main `438c6296`

A scratch config run through `node scripts/mutation-proof.mjs`, with a killed control in the same file and the same run:

| mutant | verdict on main |
| --- | --- |
| `check-attribution.mjs` back to its old string guard (control) | KILLED |
| `check-attribution.mjs`: `if (isMainEntry(import.meta.url)) {` -> `if (false) {` | **SURVIVED**, suite 8 passed / 0 failed, positive control noted |
| `isMainEntry` keeps the fast path and stops resolving (shipped mutation 4, control) | KILLED |
| `isMainEntry`: `if (!entry) return false;` -> `return false;` | **SURVIVED**, suite 8 passed / 0 failed, positive control noted |

With the helper switched off, `check-attribution.mjs --selftest` exits 0 with 0 bytes on stdout and stderr. The unchanged file also exits 0, printing `selftest: 34/34 fixtures graded as expected`. The exit code alone cannot show that `main()` ran.

## The change

**Execution witness.** Each probe now names the exit code of its script's refusal and one whole stderr line that only the script's CLI body prints. `doc-binding.mjs` and `pr-head-gate.mjs` have no function called `main`; for them, the body inside the guard plays that part.

| script | argv | exit | stderr line |
| --- | --- | --- | --- |
| `check-attribution.mjs` | none | 2 | `usage: check-attribution.mjs --range <base>..<head> [--event <event.json>] \| --selftest` |
| `check-operator-literals.mjs` | `--help` | 2 | `operator literal check: unknown argument: --help` |
| `doc-binding.mjs` | none | 2 | `usage: pnpm doc-binding <ref> \| pnpm doc-binding --self-test` |
| `live-job-conclusion.mjs` | none | 1 | `--result is required` |
| `pr-head-gate.mjs` | none | 2 | `usage: pnpm pr-head-gate <pull-request-number>` |
| `verify-publish-closure.mjs` | none | 2 | `usage: verify-publish-closure.mjs <version> [--json]` |

A parity cell now passes only if all of these hold:
- the link is a link;
- both runs reach the witness;
- both runs agree on exit code, stdout and stderr.

The cross-path comparison stays. The cell is renamed to `<script> runs main() by its real path and through a symlink, and answers the same`. When it fails, it prints `ranByRealPath`, `ranThroughLink`, `same` and the expected witness. With the helper switched off, every parity cell now fails like this:

```
FAIL check-attribution.mjs runs main() by its real path and through a symlink, and answers the same: {"linked":true,"ranByRealPath":false,"ranThroughLink":false,"same":true,"expected":{"exit":2,"err":"usage: check-attribution.mjs --range <base>..<head> [--event <event.json>] | --selftest"},"direct":{"exit":0,"out":"","err":""},"symlink":{"exit":0,"out":"","err":""}}
```

`"same":true` is the gap #1631 left open.

Both halves of the witness are needed. A guard that runs `main()` but drops its return value prints the usage line at exit 0, and only the exit code catches that. `live-job-conclusion.mjs` refuses with 1, which is also Node's exit code when the script cannot load at all, and only the stderr line catches that.

**Fixture.** `scripts/mutations/entrypoint-guard.json` goes from 5 to 17 mutations. Each per-file mutation names the parity cell of the script it edits. The two helper mutations (4 and 12) each name one parity cell, and mutation 5 names the census control.

| # | mutant | old suite (main) | this branch |
| --- | --- | --- | --- |
| 1-3 | existing per-file reversions | KILLED | KILLED |
| 4 | existing: `isMainEntry` keeps the fast path, stops resolving | KILLED | KILLED |
| 5 | existing: census reader blinded | KILLED | KILLED |
| 6-11 | each probed script's guard off by both paths (`if (false) {`, `RUN_AS_CLI = false`) | **SURVIVED** x6 | KILLED |
| 12 | `isMainEntry` answers false for every invocation; mutation 4 is its killed control in the same file and run | **SURVIVED** | KILLED |
| 13 | `doc-binding.mjs` back to `pathToFileURL(argv[1])` (its guard is a `RUN_AS_CLI` assignment, which the if-shaped anchors never matched) | KILLED | KILLED |
| 14-15 | `pr-head-gate.mjs`, `verify-publish-closure.mjs` back to `resolve(argv[1])` | KILLED | KILLED |
| 16 | `live-job-conclusion.mjs` imports a missing helper: exit 1 and identical output on both paths | **SURVIVED** | KILLED |
| 17 | `check-attribution.mjs` runs `main()` but drops its exit code: usage line at exit 0 on both paths | **SURVIVED** | KILLED |

"Old suite" means the selftest as merged in #1631 (`2de1248f`, unchanged on main since), with the old cell names; the survivors each carry mutation-proof's positive-control note. Mutation 12 names the attribution cell; the helper mutant reddens all six parity cells. Mutations 6-11 exist so that each cell's witness is graded on its own.

Each half of the witness was also run alone:

| `reachedMain` checks only | mutation 16 | mutation 17 |
| --- | --- | --- |
| the exit code | SURVIVED | KILLED |
| the stderr line | KILLED | SURVIVED |

In both runs, the same-file controls were KILLED.

**Coverage boundary**, now stated in the fixture's `why` text:
- The census reads only top-level `scripts/*.mjs`, one line at a time, and drops a line by its opening comment token. It cannot see a guard that reads `process.argv[1]` on one line and compares it on the next, a nested file, another extension, or anything outside `scripts/`.
- `post-publish-install-probe.mjs` and `preflight-npm-publish.mjs` are held by that census alone.
- `isMainEntry` compares realpath strings, not device and inode. Case-insensitive aliases and Windows junctions are not exercised.
- `--preserve-symlinks-main` (below) is not covered.
- The same hand-rolled guard is still live outside `scripts/`.

**Header corrections** in the suite:
- It said `check-attribution.mjs --selftest` exited 2 by its real path. It exits 0 on a pass (`check-attribution.mjs:444` returns 2 only on failure). The measured refusal with exit 2 is the no-argument usage line.
- The census cell was named `every entry-point guard under scripts/ ...`. It is now `every one-line entry-point guard in a top-level scripts/*.mjs goes through isMainEntry`.

## RED before, GREEN after

- **Pre-#1631 sources:** the six probed scripts restored from `4b676f30` (the parent of the #1631 merge commit), tests kept: **1 passed, 7 failed**. Every parity cell showed `ranByRealPath:true, ranThroughLink:false`, and the census listed the six old guards. Each is byte-identical to the guard line that mutations 1-3 and 13-15 put back. Mutation 13 also re-adds the `pathToFileURL` import on the line above it.
- **This branch:** **8 passed, 0 failed**. `mutation-proof` on the fixture: **all 17 killed**.

## Reproduced, not changed

**`node --preserve-symlinks-main` with the link in another directory.** Node resolves `./main-entry.mjs` beside the link:

| script | `4b676f30` | this head |
| --- | --- | --- |
| check-attribution | exit 2, usage | exit 1, `ERR_MODULE_NOT_FOUND` for `main-entry.mjs` |
| check-operator-literals | exit 2, unknown argument | exit 1, `ERR_MODULE_NOT_FOUND` |
| live-job-conclusion | exit 1, `--result is required` | exit 1, `ERR_MODULE_NOT_FOUND` |
| verify-publish-closure | exit 2, usage | exit 1, `ERR_MODULE_NOT_FOUND` |
| doc-binding, pr-head-gate | exit 1, `ERR_MODULE_NOT_FOUND` (`typescript`, `yaml`) | same |

This is how Node documents that flag, the failure is loud, and nothing in the repo passes the flag. Without the flag the same link works (exit 2, usage line). I found no fix that is clearly right, so the case is recorded as a boundary rather than patched.

**`scripts/main-entry.d.mts` is not added.** Nothing would consume it: there is no `allowJs`, no TS file imports `main-entry.mjs`, and the sibling `.d.mts` files do not reference it.

**Census and `upgrade-section-gate.mjs`.** The earlier review said this file is written in a shape the census cannot see. Measured with the suite's own predicate: with its exemption removed, the census flags `upgrade-section-gate.mjs:894`. A real two-line guard (`const entry = process.argv[1];` followed by the comparison) is invisible, and that is what the boundary text records.

## Out of scope: the same guard outside `scripts/`

Re-read on main `438c6296`; the lines are unchanged at `ee64b03c`. Nothing here reads or changes these files.
- **`packages/seat/src/custodian.ts:578`** (moved from 481 by #1666): `` import.meta.url === `file://${process.argv[1]}` ``. Measured on the compiled `dist/custodian.js` with empty stdin:
  - by the repo path: exit 1, `custodian requires a JSON launch on stdin`;
  - a copy under a real directory whose path contains a space: **exit 0, no output**;
  - the repo file through the `/tmp` spelling: **exit 0, no output**.

  `launchSeat` builds `argv[1]` from the launcher's own resolved `import.meta.url` (`launcher.ts:26-35`, `:167`), so the symlink trigger should not be reachable there under default flags. The encoding trigger is reachable. Its consequence would be `custodian exited before ready` from `launcher.ts:191-198`; that is reasoned, not run, because the launcher is Linux-only and this host is macOS.
- **`examples/05-scale-showcase/src/observe.ts:216`:**
  - real path: exit 1, NATS connection error;
  - absolute `/tmp` spelling: **exit 0, no output**;
  - relative `pnpm serve` from a `/tmp`-spelled cwd: still runs, because `process.cwd()` is physical;
  - relative run from a copy under a path with a space: **exit 0, no output**.
- **`examples/05-scale-showcase/src/swarm.ts:129`:** real path runs (`connect FAILED ... connection refused`); absolute `/tmp` spelling: **exit 0, no output**.
- **Same class, not measured.** These use the `resolve` / `pathToFileURL` / `fileURLToPath` forms, which break on a symlink only:
  - `bin/smoke/gen-ci-suites-dts.mts:56`
  - `bin/smoke/gen-npm-publish-preflight-dts.mts:46`
  - `bin/smoke/gen-publish-closure-dts.mts:55`
  - `bin/smoke/gen-reaper-dts.mts:303`
  - `examples/06-feed-agent/src/pump.ts:399`
  - `extensions/connector-hermes/src/launch.ts:312`
  - `packages/seat/scripts/assert-shipped-natives.mjs:64`

## Everything run

macOS (Darwin 25.6.0, arm64), Node 24.2.0 (CI pins 22; no Node 22 on this host), pnpm 11.1.2.

| command | result |
| --- | --- |
| `node scripts/entrypoint-guard.selftest.mjs` at main | 8 passed, 0 failed (blind to both-path disablement, see above) |
| same, six sources from `4b676f30` | **1 passed, 7 failed** |
| same, this head | **8 passed, 0 failed** |
| `node scripts/mutation-proof.mjs --config scripts/mutations/entrypoint-guard.json` (rebased on `cd9c8dda`) | **All 17 mutation(s) killed** |
| `node scripts/mutation-coverage.mjs --gradable-only scripts/mutations/entrypoint-guard.json` | ACCEPTED, graded=1, refused-with-reason=0 |
| `node scripts/mutation-reproof.mjs --base origin/main --head HEAD` (on the tree merged with `ee64b03c`) | 1 fixture selected from 449, 17 killed, MUTATION REPROOF OK |
| `node scripts/mutation-coverage.selftest.mjs` | 170 passed, 0 failed |
| `pnpm smoke:mutation-fixtures` | OK |
| `node scripts/check-operator-literals.celltamper.selftest.mjs` | 175/175 |
| `node scripts/check-attribution.mjs --selftest` | 34/34, exit 0 |
| `node scripts/check-operator-literals.mjs --range origin/main..HEAD` | clean, 0 findings |
| `pnpm changeset status --since=origin/main` | no packages bumped (empty-frontmatter changeset) |

Not run:
- the full smoke chain;
- `pnpm typecheck` (no TypeScript changed);
- docs checks (no docs changed);
- Linux and Windows.


davidfarah2003:
Verified independently before merging, on macOS arm64 under Node 26.7.0 with the workspace dependencies linked in.

- selftest at this head: 8 passed, 0 failed
- `mutation-proof` over the 17-mutation fixture: all 17 killed
- the counter-check that makes those numbers mean anything: I applied mutation 17 by hand (`check-attribution.mjs` runs `main()` but drops its exit code) and ran both suites against the same mutated tree. Main’s suite: `OK (8 passed, 0 failed)`. This branch’s suite: `FAILED (7 passed, 1 failed)`. The gap you describe is exactly that, reproduced here.
- `smoke:mutation-fixtures`: OK

One incidental confirmation of the design. Run in a git worktree with no `node_modules`, the `doc-binding.mjs` and `pr-head-gate.mjs` cells fail at this head and pass on main, because those two scripts cannot load `typescript` and `yaml` at all and so never reach `main()`, while main’s cells see two identical failures and call that agreement. That is the `"same":true` hole doing precisely what you said it does, and the new witness catching it. Worth knowing that this suite now depends on the deps being installed, which reads as a stricter test rather than a flaky one.

Thanks for the measurement tables and for stating the coverage boundary rather than implying it was covered. The `--preserve-symlinks-main` case and the guards outside `scripts/` being recorded instead of quietly patched is the right call.

PR #1689, opened 2026-09-17
test(core): grade the bearer refresh-margin edge on a pinned clock

L4XB:
Refs #1561. Follow-up to #1572.

The round-2 panel on #1572 left one blocker and two coverage notes. On `main`, no cell grades the margin conjunct of the re-serve guard (`packages/core/src/endpoint.ts:822`): flipping `<= 0` to `< 0` survives, and so does deleting the conjunct. The 12s cadence cell also bounds the 15s retry only from below. This PR adds the equality cell the blocker asked for and cells for both notes. It also corrects two merged sentences that do not match the code. No production source changes.

## How the cells reach the edge

The two sides of `<= 0` differ only when `exp - now - 60s` is 0. `exp` is whole seconds, so on a live clock that is a one-millisecond race. The new cells pin `Date.now` for one `refreshBearer()` call, the way `packages/core/smoke/secret-fs.smoke.ts:57-69` already pins it. The same window wraps `setTimeout`, so a cell can read the delay the refresh arms. A `finally` restores both before any timer can run, because the refresh awaits only the source, which resolves as a microtask. The helper is `refreshAt` (`bearer-inside-margin.smoke.ts:165`). No clock seam is added to `endpoint.ts`.

## What it adds

Four cells in `packages/core/smoke/bearer-inside-margin.smoke.ts` (12 to 16):

- `a re-served token with zero margin left is refused (the margin edge is inclusive)` (`:187`). The held token is re-served at `now = exp*1000 - 60000`.
- `that refusal arms the retry at 15s (BEARER_RETRY_MS), not at the 5s floor` (`:192`). That refresh arms one timer, at 15000 ms.
- `the same token with 1ms of margin left is adopted, with no warning` (`:203`). This is the other side of the edge: same token, `now` one millisecond earlier. There is no warning and one timer at the 5s floor. It also shows that the refusal above comes from the margin, not from the pinning.
- `a re-served token with most of its life left is adopted, with no warning` (`:215`). A caching source returns the same bytes for a token with about 10 minutes left. This is the input review 2 found missing.

Four mutations in `packages/core/smoke/mutations/bearer-inside-margin.json`, each naming its cell.

## Mutants

"Before" means the 12-cell suite as it is on `main`, run against the same eight mutations. The suite is a copy of `origin/main:packages/core/smoke/bearer-inside-margin.smoke.ts` whose only change is an import pointing at this checkout's `src`, so it loads the same `endpoint.ts` the mutations edit. The scratch config differs from this PR's in two places: its `command` runs that copy, and the four new entries name `the re-serve refusal backs off past the 5s floor (no re-read in 12s)` as `expectRed` and `cell`, because their own cells do not exist on `main`. The four existing `endpoint.ts` mutants were KILLED in that same run, which is the positive control the tool requires before it grades a survivor.

| mutant (`endpoint.ts`) | before | after | cells red after (16 cells, measured by hand) |
| --- | --- | --- | --- |
| `:822` `<= 0` to `< 0` | SURVIVED, control present | KILLED | zero-margin refusal; retry at 15s (`[ 5000 ]`) |
| `:822` `<= 0` to `<= 1` | SURVIVED, control present | KILLED | 1ms margin adopted |
| `:822` margin conjunct deleted (`if (bearer === held)`) | SURVIVED, control present | KILLED | 1ms margin adopted; most-of-life adopted |
| `:835` `armBearerRefresh(600_000)` | SURVIVED, control present | KILLED | retry at 15s (`[ 600000 ]`) |
| the four existing mutants | KILLED | KILLED | unchanged |

Not registered, measured once: deleting the `:835` retry arm outright reddens `that refusal arms the retry at 15s` with `[]` (no timer armed). It also reddens the 12s cell (`stuckReads: 4, stuckBaseline: 3`), as review 2 measured. So a cell now observes the retry being armed, which review 1 noted was missing.

Against the pre-#1572 source (`endpoint.ts` restored from `2de1248f41dd5040537c23d4bf7cb2cee88496ce`, tests kept), the zero-margin cell and the 15s cell are RED (`[]` warnings, `[ 5000 ]`). The 1ms cell and the most-of-life cell are GREEN there by construction, because that source never refused anything. Those two guard against over-refusal, and the `<= 1` and conjunct-deletion mutants grade them.

Scope: like the rest of the suite, these cells call the private `refreshBearer()` directly. The timer path runs the same method with `initial = false`. A probe that drives it through real timers read a stuck source at 5002, 20004, 35005 and 50007 ms, which is the 15s cadence the new cell pins.

## Text corrections

I checked every file #1572 touched, plus `docs/`, `SPEC.md` and `spec/`, for the claims the panel listed.

- **Changeset warning count.** `.changeset/bearer-inside-refresh-margin.md` said a re-served token is refused "with one recoverable warning". I measured a source that serves one healthy token, then the same near-dead token forever (it expires at +26510 ms). Reads came at 1, 1, 5002, 20004, 35005 and 50007 ms. Warnings came at 5003 (re-served), 20004 (re-served), 35006 (expired) and 50007 (expired). The changeset now says there is a warning on every retry, and that once the held token expires the expiry refusal takes over at the same cadence.
- **Mutant note on the dial.** The note on the `held !== undefined || true` mutant said the authenticator would present `undefined`. I measured `start()` against a source whose first token has already expired, with servers pointed at a closed port:
  - On `main`, it throws the pre-dial guard (`this endpoint's user bearer has expired and renewal through the auth exchange is failing ...`) and keeps the dead token held.
  - Under the mutant, it throws `the bearer source returned a token that has already expired - nothing adopted` with `currentBearer` undefined, after one read, before the dial.
  - The note now says that.
- **Claims not in the tree.** "The expiry test then closes it", "six reads" and "reachable whenever a non-initial refresh fails" appear only in #1572's PR body and thread. For the record, I ran a source that re-mints new bytes on an `exp` pinned 12s out, for 47s. Reads came at 0, 5002, 10003, 15004, 30006 and 45007 ms, with an expiry warning at each of the last three. The 5s phase ends at `exp`, so "six reads" described only that phase of a 30s pin. The 15s retries, each with a warning, continue until the probe stops.

An empty changeset covers the test additions. The edit to the core changeset stays in its pending patch.

## Everything run

Head `5681bbaff4c2b8e45eaf07823b439f0b36731283`: 3 commits on `main` at `ee64b03cc44ed11025024f58222b18004951c4c2`. Environment: macOS (Darwin 25.6.0), Node v24.2.0, pnpm 11.1.2. CI pins Node 22, and this host has no Node 22. The suite needs no broker.

| command | result |
| --- | --- |
| `pnpm smoke:bearer-inside-margin` | 16/16, five consecutive runs |
| same, with `endpoint.ts` from `2de1248f41dd5040537c23d4bf7cb2cee88496ce` | 10/16; the two new refusal cells RED as above |
| `node scripts/mutation-proof.mjs --config packages/core/smoke/mutations/bearer-inside-margin.json` | 8/8 KILLED, each on its named cell; tree clean after |
| same 8 mutants on `main`'s 12-cell suite | 4 KILLED, 4 SURVIVED with the positive-control note |
| `node scripts/mutation-reproof.mjs --base origin/main --head HEAD --shard 0/12` (short `TMPDIR`) | `MUTATION REPROOF OK (1 fixture(s) selected; 1 discriminated, 0 inherited pre-red, 0 attributable pre-red, 0 unmeasured pre-red, 0 inconclusive)`; `--list-shards 12` puts it on shard 0 only |
| `node scripts/mutation-coverage.mjs --gradable-only packages/core/smoke/mutations/bearer-inside-margin.json` | ACCEPTED |
| `pnpm smoke:mutation-fixtures` | OK |
| `pnpm smoke:recoverable-error-channel` | 5/5 |
| `tsc` on the smoke file (repo `tsconfig.base.json`, strict) | clean |
| `pnpm changeset status` | exit 0 |
| `pnpm smoke:gate-inventory` | red, and red in the same way on an archive of `main` at `ee64b03cc44ed11025024f58222b18004951c4c2` (see below) |

The gate-inventory failures on both trees: `smoke:delivery-pidfile-ownership` and `smoke:reap-absent-refusal` are ungated, and `smoke:connection-status` is listed twice. This branch touches neither `package.json` nor `ci-suites*`.

Not run: `smoke:expiry-renewal:auth`. It needs a broker and imports `@cotal-ai/core` through `dist/`, and this checkout has no build. This branch changes no source, and this suite still kills the clamp mutant that `smoke:expiry-renewal:auth` motivated.


davidfarah2003:
This branch no longer merges with `main`, and the conflict is mechanical rather than anything
about your change.

```
CONFLICT (modify/delete): .changeset/bearer-inside-refresh-margin.md
  deleted in origin/main and modified in this branch
```

That file was on `main` at your merge base, and the release commit `b086974ff`
(`chore(release): version packages`, #1559) consumed and deleted it, which is what a release does
to every pending changeset. Your branch edits it, so git cannot decide on its own.

Take `main`'s deletion. The entry it described has already shipped in the changelog, so
re-adding the file would double it. Your own changeset, `.changeset/bearer-refresh-margin-edge.md`, is a different filename and
survives the rebase untouched, so your release note is not lost.

No production file of yours is in the conflict. I have not pushed to your branch.

L4XB:
Merged `main` in `537a30c9`, taking `main`'s deletion of `.changeset/bearer-inside-refresh-margin.md` as you described; it's a merge because the branch can't be force-pushed. `.changeset/bearer-refresh-margin-edge.md` is untouched, and the diff against `main` is now just that changeset and the two `bearer-inside-margin` files. On the merged head, `bearer-inside-margin.smoke.ts` reports 16/16, and `check:docsbundle` and `check-operator-literals --range` are clean.

PR #1692, opened 2026-09-17
fix(runtime): replay a drive's own journal again when a read loses a round

L4XB:
Refs #1623. Follow-up to #1643.

The re-review of #1643 at `1073a3a90` found that it did the first of the issue's three points and not the other two. This PR does points 2 and 3: a live drive now retries `RunJournalReplayRaced` at the step, and a runtime cell drives a real `parallel` of three `ask` steps. Point 1, serialising replays across processes, is waiting on the grant-model question in the #1643 thread and is not part of this change.

## Point 2: the race is retried at the step

`activateRun` treats `RunJournalReplayRaced` as a lost round. It allows `Math.max(1, opts.attempts ?? 3)` rounds (`packages/core/src/run-journal.ts:836`) and `continue`s on the race (`:846`-`:848`). The only pause between rounds is `beforeRetry` (`:840`), and the driver's call (`implementations/runtime/src/run-driver.ts:511`) passes only `onReplayed`, so a lost round is replayed at once. `replayOwnJournal` (`implementations/runtime/src/journal-store.ts:50`) copies that: up to `OWN_REPLAY_ATTEMPTS = 3` replays (`:39`), no pause, only the race is caught, and the race is raised unchanged when the third round is lost too (`:61`).

I checked each `replayRunJournal` call in the runtime and decided per site:

| site | takeover id it reads under | decision |
| --- | --- | --- |
| reader behind `createRunScopeAuthority` (`run-scope-authority.ts:19`): every effect dispatch, the bind check, pause and wait authorization, and each poll of a parked `ask` | the drive's lease id | **retried**. This is the read whose race the interpreter recorded as the step's `L4000`. |
| the same reader in `cotal run --local`'s own drive (`run-command.ts:360`) | the CLI attempt's lease id | **retried**, through the same call |
| `noRecordToResume` (`run-driver.ts:183`) | the drive's lease id | **retried**. It is a read on the takeover path before activation, the same kind as `activateRun`'s. Before this change, a race there escaped `drive()` as a raw throw, and the host recorded the attempt `failed` instead of `released` with its diagnostic. |
| `cotalLangRunHost.status` (`run-host.ts:171`) | the caller's id; the manager passes a fresh `newTakeoverId()` for each call (`implementations/manager/src/run-hosting.ts:636`, used at `:332`) | not retried |
| `replayRunEntries` behind `locate` and `resolveCheckpoint` (`resolve-checkpoint.ts:211`) | the caller's id, or a fresh one when none is given (`:121`); the manager passes a fresh one (`run-hosting.ts:308`) | not retried |
| `cotal run journal` (`run-command.ts:433`) | a fresh `newTakeoverId()` for each verb (`run-command.ts:714`) | not retried |

Only `RunJournalReplayRaced` is retried. A reader outside this process can also tear a fetch (`JetStreamStatusError: consumer deleted`), leave `RunJournalReplayIncomplete`, or return an empty replay. None of those is this race, and none of them is retried here; they belong to the cross-process case the issue's point 1 describes.

The last three sites are operator reads, not reads inside a live drive. Each one mints its own 64-bit id for a single call. A race there means an unexplained reader holds a name that was issued to nobody else. Such a race should be reported on the first read, not replayed away, and the caller can run the command again.

## Point 3: the runtime cell

`implementations/runtime/smoke/run-host-replay.smoke.ts` (`pnpm smoke:runtime-run-host-replay`, registered as a CI fragment) drives `cotalLangRunHost.drive`, the entry point the manager uses. It runs on a real `nats-server` with the manager's connection split: a driver connection and a separate mediator connection, both under one lease. The program spawns three seats and runs `parallel({ dev, rev, qa })` over three `ask` steps. The suite answers all three at once through `resolveCheckpoint`, requires the three answers to land within one second, and then checks that every branch settled `ok`.

- **Block 1.** Two loops call `RunHost.status` under the drive's own takeover id, on the driver's connection, for as long as the drive runs. This is the reader that #1643 named as outside `RunScopeAuthority`'s chain. Cells: no branch fails; every status read returned the run from its activation, with no torn read and no empty one; the reads spanned the settle (a later read saw more rows than the first), so the cell cannot pass unless the status reads ran while the drive was settling its branches. No shipped caller hands `status` a drive's lease id: the manager mints a fresh `newTakeoverId()` for each operator read (`implementations/manager/src/run-hosting.ts:636`). Block 1 drives the reader that `packages/core/src/run-journal.ts:406` describes, not a path the manager takes today.
- **Block 2.** A second connection takes one record off the drive's durable right after the drive creates it, twice, so the next step read loses one round. Cells: that round was really lost; no branch fails.
- **Block 3.** This block uses block 2's run and lease through `createRunScopeAuthority(...).journal()`. After two lost rounds, the third replay reads the run. With seven creations planted, the read fails with `RunJournalReplayRaced` and exactly one creation is left over, so it made three rounds and not four. A refused consumer create is raised on the first read.
- **Block 4.** A journal that was activated but never pinned is resumed through the host, with one lost round planted on the driver's connection. The drive still ends `released` with the no-record diagnostic.

The second connection stands in for a reader outside this process's lock, which is point 1's case, limited to one lost round. At this head I found no in-process path by which a step read loses a round: the core lock orders every read of a durable, and a leftover durable is remade. So blocks 2 to 4 use a planted reader, and block 1 checks the in-process lock with a real one. Blocks 1, 2 and 4 go through the manager's entry point. Block 3 builds the authority by hand from block 2's lease.

The suite imports core by path from `packages/core/dist/index.js`, the same file the runtime's `@cotal-ai/core` resolves to. `mutation-coverage --gradable-only` only accepts a core mutation for a runtime suite when it can see that the suite loads core's built output. Block 3's `instanceof RunJournalReplayRaced` passing shows that the suite and the drive share one copy of core.

## One thing the cell showed

In a hosted drive, `RunScopeAuthority`'s own promise chain already orders the step reads. So reverting the core per-durable lock alone leaves a hosted `parallel` green, and block 1 kills that mutant only because it adds a second reader under the lease id. The opposite also holds: with the core lock in place, removing the authority's chain changes nothing any cell can see. I measured it with a same-file control and recorded it under `redundant` in the fixture rather than registering a survivor:

In three runs of each, `mutation-proof` killed RH5 every time, and the removal of the authority's own read chain survived every time. The run counts are my tally over separate runs, not a line the tool prints.

With both removed and core rebuilt, three runs out of three failed both panels before any ask parked. In block 1 the branches recorded L4000 `consumer deleted` (two runs) or an incomplete replay (one run).

## Red before, green after

The three runtime sources restored to `origin/main` with the suite kept. The suite loads without module errors. Three runs out of three gave this result:

```
• 2: a parallel of three asks, with one of the drive's step reads losing a round
  ✓ the three answers landed within the same second, again
  ✓ the other reader really took a round off the drive's durable
  ✗ FAIL: no branch fails when one of the drive's step reads loses a round to another reader {
  outcome: { status: 'failed', error: { name: 'EffectError', code: 'L4000',
    message: 'run panel-raced: the replay consumer wfj_panel-raced_9eb125941e578bd5 had already delivered 1 records to someone else. Another driver is replaying this run; retry the takeover rather than resuming from a partial prefix.' } },
  branches: [ 'dev-verdict:failed L4000 run panel-raced: the replay consumer ...', 'rev-verdict:cancelled', 'qa-verdict:cancelled' ]
• 3: a step read replays at most three times
  ✗ FAIL: a step read that loses two rounds replays a third time and reads the run
  ✗ FAIL: a step read that loses every round fails with the race after three replays, not a fourth
  ✓ a step read that fails for any other reason raises that failure without replaying again
• 4: a journal with no run record is still named after a lost round
  ✓ the resume's diagnostic read really lost a round
  ✗ FAIL: and it still releases the run naming a journal that was activated and never pinned {
  status: 'failed', error: { name: 'RunJournalReplayRaced', message: 'run unpinned: the replay consumer wfj_unpinned_158f3420ccba8ed5 had already delivered 1 records to someone else. ...' } }
run-host-replay.smoke: 8 passed, 4 failed
```

At head, 12 passed and 0 failed: three direct runs, plus the green baseline that every proof run below starts with. An earlier revision of the suite (11 cells, before the scope cell) was also green in 10 consecutive runs.

Block 1 checks #1643's lock, not this change, so it is green on both sides above. Against the core before #1643 (`packages/core/src/run-journal.ts` from `438c6296^`, core rebuilt, runtime at head), it failed in three runs out of three:

```
• 1: a parallel of three asks, with the drive's takeover id also read by status
  ✓ the three answers landed within the same second
  ✗ FAIL: no branch fails while status reads the drive's own takeover id {
  outcome: { status: 'failed', error: { name: 'EffectError', message: 'consumer deleted', code: 'L4000' } },
  branches: [ 'dev-verdict:failed L4000 consumer deleted', 'rev-verdict:cancelled', 'qa-verdict:cancelled' ]
  ✗ FAIL: and every one of those status reads returned the run from its activation, none torn and none empty {
  reads: 4, broken: [ 'JetStreamStatusError: consumer deleted',
    { runId: 'panel-shared', ..., journal: [] },
    'RunJournalReplayIncomplete: run panel-shared: the journal replay delivered 11 of 18 records; ...' ]
```

The `journal: []` is the silent empty replay #1643 describes, returned through `RunHost.status`. Block 3 also fails on that older core, because its creation counts assume the leftover repair: without the repair a lost round costs one creation instead of two.

## Mutation proof

`pnpm mutation-proof --config implementations/runtime/smoke/mutations/run-host-replay.json`:

```
KILLED  RH1: a step read raises the race on its first lost round again
        red, and named: no branch fails when one of the drive's step reads loses a round to another reader
KILLED  RH2: the replay bound grows by one round
        red, and named: a step read that loses every round fails with the race after three replays, not a fourth
KILLED  RH3: the replay bound shrinks by one round
        red, and named: a step read that loses two rounds replays a third time and reads the run
KILLED  RH4: every failure of a step read is replayed, not only the race
        red, and named: a step read that fails for any other reason raises that failure without replaying again
KILLED  RH5: the step reads go back to the unretried replay
        red, and named: no branch fails when one of the drive's step reads loses a round to another reader
KILLED  RH6: the no-record diagnostic goes back to the unretried replay
        red, and named: and it still releases the run naming a journal that was activated and never pinned
KILLED  RH7: replays under one takeover id run concurrently again, sharing one durable
        red, and named: no branch fails while status reads the drive's own takeover id
KILLED  RH8: CONTROL (expect KILLED): the freshness guard stops refusing, so a tail is read as the run
        red, and named: a step read that loses every round fails with the race after three replays, not a fourth
All 8 mutation(s) killed. The suite discriminates.
```

RH7 is the shared-durable mutant, which reverts #1643's lock. RH8 is its positive control in the same file: it shows that the suite reads a rebuilt core and not a stale `dist`. Both RH7 and RH8 build core before the run and rebuild it after the restore. RH7 was killed in every run I made: five runs of the full config (two of them inside `mutation-reproof`) and three runs of RH7 alone. RH5 and RH6 swap each call site back to the unretried core replay under the same local name, which is the source as it was before this change.

## Everything run

Environment: macOS 26.6.2 (Darwin 25.6.0, arm64), Node v24.2.0, pnpm 11.1.2, `nats-server v2.12.12` (the version CI pins). Every broker suite ran with a short `TMPDIR` under `/tmp`, because the macOS default is too long for the socket paths some suites create.

| command | result |
| --- | --- |
| `pnpm smoke:runtime-run-host-replay` | 12 passed, 0 failed, 3 runs of 3 |
| same, runtime sources at `origin/main` | 8 passed, 4 failed, 3 runs of 3 (above) |
| same, core `run-journal.ts` at `438c6296^` | red on block 1 in 3 runs of 3 (above) |
| `node scripts/mutation-proof.mjs --config implementations/runtime/smoke/mutations/run-host-replay.json` | 8/8 KILLED, each on its named cell |
| `node scripts/mutation-coverage.mjs --gradable-only implementations/runtime/smoke/mutations/run-host-replay.json` | ACCEPTED |
| `node scripts/mutation-reproof.mjs --base 5c63e4af --head HEAD` | this fixture's 8 mutations KILLED. The run ends `PRE-RED TRANSITION UNMEASURED` for `package-json-duplicate-keys` and `mesh-spawn`: both are red at the base and at the head, without a stable failure signature (below) |
| every other `smoke:runtime-*` suite, plus `smoke:run-journal` | all green except `smoke:runtime-mesh-spawn` (see below) |
| `pnpm typecheck` (build plus every package) | clean |
| `pnpm check:docsbundle` (includes `check:docs-voice`) | clean |
| `pnpm changeset status --since origin/main` | `@cotal-ai/runtime` and `@cotal-ai/connector-core` patch, with the fixed group |
| `pnpm smoke:ci-fragments`, `smoke:broker-migration`, `smoke:dist-freshness`, `smoke:core-boundary` | green |
| `BASE=<origin/main sha> HEAD=<head sha> pnpm smoke:ci-suites-freeze` | 10 passed, 0 failed |
| `pnpm smoke:gate-inventory` | red, the same three findings as the base (below) |
| `pnpm check:shard-stability <base sha> <head sha>` | STABLE: 611 to 612 suites, added 1, no frozen legacy suite changed index, no pre-existing suite changed runner |

Green runtime suites, with passed-cell counts: `runtime-contract` 5, `runtime-fork` 110, `runtime-host-starvation` 58, `runtime-journal-store` 8, `runtime-mesh-ask` 42, `runtime-mesh-checkpoint` 42, `runtime-mesh-conclave` 39, `runtime-mesh-monitor` 25, `runtime-mesh-notify` 51, `runtime-mesh-replied` 23, `runtime-mesh-seam` 16, `runtime-mesh-sleep` 20, `runtime-mesh-turn` 49, `runtime-mesh-wait` 62, `runtime-mesh-worktree` 20, `runtime-migrate` 68, `runtime-placement-grants` 12, `runtime-run-command` 24, `runtime-run-command-usage` 19, `runtime-run-driver` 87, `runtime-run-driver-auth` 74, `runtime-spawn-policy` 14, `run-journal` 126.

Red on this host, and red without this change:

- `smoke:runtime-mesh-spawn` stops at `mesh-spawn.smoke.ts:689` with `ERR_FS_EISDIR: Path is a directory: .../sp-alias-*/prepared`. That line is `rmSync(aliasLink, { force: true })` on a symlink to a directory. The same error stops it at head with `TMPDIR` under `/tmp` and under `/private/tmp`, and with the three runtime sources at base.
- `smoke:gate-inventory` reports `smoke:delivery-pidfile-ownership` and `smoke:reap-absent-refusal` as ungated, and `smoke:connection-status` as listed twice in `ci-suites.txt`. A clean checkout of `origin/main` reports the same three findings, with and without `CI=true`. The suite added here is gated through its fragment and is not among them.

One note on the reproof tool. I first passed `--base origin/main`. The tool clones the checkout with `--shared` and checks out that name inside the clone, where `origin/main` points to this repository's local `main` branch (`794e3d2e` here), not to the ref that was resolved when the diff was selected. On that run, both red suites above were reported as green at base and red at head. With the base given as a SHA, the tool finds both suites red at base too, and reports the transition as UNMEASURED because the failure signatures differ. For `smoke:gate-inventory`, the only line that differs is the count (668 scripts and 641 reached at head, 667 and 640 at base; the extra one is the suite added here), and the findings are identical. For `smoke:runtime-mesh-spawn`, my own base and head transcripts are identical once the random temp names are masked. I cannot see which line differed in the tool's snapshots.

## Not verified

- An authenticated mesh. Every broker in this suite is open. `smoke:runtime-run-driver-auth` covers the retried reader on an enforcing broker, and it is green.
- A reader in a separate process. The planted reader is a second connection in the same process, which is outside the lock just as another process would be, limited to one lost round.
- The full `pnpm smoke:ci` chain and `pnpm test`.
- The `.internal` submodule is not initialized in this checkout, so I could not check it for a related plan row.

## Scope

- `implementations/runtime/src/journal-store.ts`: `replayOwnJournal` and `OWN_REPLAY_ATTEMPTS`.
- `implementations/runtime/src/run-scope-authority.ts`, `implementations/runtime/src/run-driver.ts`: the two lease-id reads now call it.
- `implementations/runtime/smoke/run-host-replay.smoke.ts`, `implementations/runtime/smoke/mutations/run-host-replay.json`, the `package.json` script, and the `bin/smoke/ci-suites.d` fragment.
- `docs/workflows.md`, with the regenerated `extensions/connector-core/src/docs-bundle.generated.ts`.
- `.changeset/replay-race-retried-at-step.md`.

`packages/core` is unchanged. Its comments already say the lock is per process (`run-journal.ts:406`, "in THIS process"; `:466`, "no live replay in this process"), so I left them as they are while point 1 is open. `SPEC.md` is unchanged, since the wire contract does not change.

File and line references are against this branch, based on `main` @ `5c63e4af`.


davidfarah2003:
Verified this head (`cb01b363a`) independently. Everything declared holds.

| gate | result |
| --- | --- |
| runtime typecheck | clean |
| `smoke:runtime-run-host-replay` | 12 passed, 0 failed |
| `smoke:runtime-run-driver-auth` | 74 passed, 0 failed |
| `smoke:mutation-fixtures` | 450 fixtures, 2649 anchors, 0 dead / ambiguous / missing |

I also hand-applied two mutants from `run-host-replay.json` rather than taking the fixture on
faith. Removing the retry entirely takes the suite to 8 passed, 4 failed. Retrying on any error
instead of only `RunJournalReplayRaced` takes it to 11 passed, 1 failed, and the cell that catches
it is the one that says a read failing for another reason raises without replaying again. The
fixture discriminates in both directions, which is the part that is easy to get wrong.

No broker survived any run.

## One thing blocks the merge

The branch no longer merges with `main`. Exactly one file conflicts:

```
CONFLICT (content): Merge conflict in extensions/connector-core/src/docs-bundle.generated.ts
```

`docs/workflows.md` itself auto-merges cleanly. The conflict is only in the generated bundle,
because #1633 regenerated it from the same page after this branch was cut. So the resolution is
mechanical: rebase onto `main`, then re-run the generator and commit its output. Nothing in your
production change is involved, and I have deliberately not pushed to your branch.

Worth flagging separately: a generated artifact that is checked in will conflict between any two
branches that touch the page it is generated from, regardless of whether the edits overlap. That
is a repository-shape problem rather than anything about this PR.

Once it merges cleanly it is good to land on the evidence above.

L4XB:
Merged `main` in `682ed098`; this branch can't be force-pushed, so it's a merge rather than a rebase. As you described, the only conflict was `extensions/connector-core/src/docs-bundle.generated.ts`. I regenerated it from the merged `docs/workflows.md` with `pnpm gen:tooldocs && pnpm gen:docsbundle` instead of picking a side, and `pnpm check:docsbundle` passes on the clean tree.

Re-ran your gates on the merged head, with nats-server v2.12.12 (the version CI pins):

| gate | result |
| --- | --- |
| runtime typecheck | clean |
| `smoke:runtime-run-host-replay` | 12 passed, 0 failed |
| `smoke:runtime-run-driver-auth` | 74 passed, 0 failed |
| `check-operator-literals --range` | clean |

`smoke:mutation-fixtures` currently reports 8 dead anchors (`control-target.ts`, three connector files, `spawn.ts`, `manager.ts`, `commands.ts`). The same 8 are dead on `main` at `4900d22d` alone, so they come from `main`, not from this merge; this PR's own `run-host-replay.json` anchors are all present.

PR #1691, opened 2026-09-17
fix(cli): judge the re-exec entry by its resolved file and refuse before writing any state

L4XB:
Refs #1629. Follow-up to #1641.

#1641 was merged while two of the three panel reviews still requested changes. This branch fixes each item those reviews measured.

## What #1641 left open, reproduced at `438c6296`

Each probe below is driven from a file that is not the cotal entry, against a scratch root. The seed probe also uses a scratch `XDG_CONFIG_HOME` with `COTAL_ALLOW_CHECKOUT_SEED=1`.

**Any file named `cotal.ts`, `cotal.js` or `cotal` passed the guard.** The check was a regex on `basename(process.argv[1])` (`implementations/cli/src/lib/self-exec.ts:14`, `:36`). A file at `<scratch>/foreign/cotal.ts` calls `startManagerDetached` on load:

```
PARENT: startManagerDetached returned pid 11621
PARENT: manager log: "CHILD: I am <scratch>/foreign/cotal.ts pid 11621, argv tail=[\"supervise\",\"--space\"]"
```

The detached "manager" is that same file again, which is the #1629 shape. The fixture exits as soon as it sees `supervise`, so only one generation ever exists.

**The auth starter claimed its pid slot before the guard.** `startAuthServiceDetached` ran `reclaimDeadLegacyPid`, then `claimAuthPidSlot`, and only then called `selfArgv()` inside the `try` (`auth-proc.ts:153-160`). The claim publishes the slot already holding the launcher's pid (`auth-proc.ts:119`). The probe calls the real `ensureAuthService` twice, then `stopAuthService` with a signal function that records instead of sending:

```
before: auth-service.f1%20space.pid
first call: THREW: refusing to re-exec this process as `cotal`: ...
after first: auth-service.6631207370616365.pid
  auth-service.6631207370616365.pid = "8831" (this process is 8831)
authServiceUp: true
retry: RETURNED {"endpoints":"reported"}
provider ready() calls: 1
teardown would signal: [[8831,"SIGTERM"]]
```

The pre-hex record was deleted as well. In this probe that record was an empty husk, which `reclaimDeadLegacyPid` removes as a pre-protocol file; it held no dead pid. The suite cell below seeds a dead pid instead, and that record is deleted at the base too. The retry started nothing, because the record read alive, and went straight to the provider's readiness wait. That `ready()` is a stub that answers at once. A real provider would wait there for a daemon nobody started.

[...]

Issue #1554, opened 2026-09-13
manager: an absent reap outcome is rendered as "already forgotten" at two sites, so a live seat can be retired silently


davidfarah2003:
An absent reap outcome is a statement about ADDRESSABILITY. Two sites in the manager render it as a statement about LIVENESS, and both render it as success. The reaper's own contract is correct; the two rendering sites drop its discipline.

## The contract

`implementations/manager/src/runtime/index.ts:50-52`, verbatim at `origin/main`:

```
/** Reap an orphaned custody by reference, or refuse by name when this runtime does not custody its
 *  own processes. Absent means REFUSE, never "assume gone": the lifecycle stays held rather than
 *  retiring over a live seat. */
```

The refusal that IS implemented is the non-custodial throw at `:55`. The absent case is not refused anywhere.

## The source of absent

`packages/seat/src/reap.ts:80-83`:

```
  try {
    rec = readRecord(path);
  } catch (e) {
    if ((e as NodeJS.ErrnoException).code === "ENOENT") return { outcome: "absent" };
```

A missing `record.json` yields `outcome: "absent"`. That is a claim that the record could not be read, not a claim that the process is gone.

## The two sinks

`implementations/manager/src/manager.ts:7130`:

```
disposal = `the spawned seat was reaped (${evidence.outcome === "absent" ? "already forgotten by the runtime" : evidence.detail})`;
```

`implementations/manager/src/manager.ts:7158`:

```
console.error(`static retirement ${a.name}: orphan seat process ${evidence.outcome === "absent" ? `already forgotten by runtime ...
```

Both convert an unreadable record into "already forgotten" and return normally, so cleanup proceeds and the alias is freed.

## Accept control, expected value stated before measuring

Expected: `requireRuntimeReap(` appears exactly 2 times in `manager.ts`, and the set of `outcome === "absent"` renderings is the SAME 2 lines, so the census is exhaustive rather than a sample.

```
requireRuntimeReap( callsites in manager.ts : 2      (expected 2)
outcome === "absent" renderings             : 2      lines 7130 and 7158
REFUSE control, requireRuntimeZzqq          : 0
```

Both readers were evaluated over one captured blob at `origin/main` = `d2cfac8eabead8a5c9845af584ec360ef0f1e192`.

## Three upstream causes of a missing record, only one of which is benign

A. The seat genuinely exited and its record was cleaned up. Benign, and this is the case the message describes.

[...]

Issue #1528, opened 2026-09-12
delivery: the daemon never writes its own pidfile, so a stale record satisfies the fail-closed liveness guard


davidfarah2003:
Split out of #1526, where it was measured as a distinct mechanism rather than a footnote. That issue is about a session reporting `ready` over a queue that never drains; this one is about the delivery daemon's liveness record, which is a different component with a different fix.

## What

Nothing writes `delivery.<space>.pid` except the launcher. `implementations/delivery/src/delivery.ts` contains no pidfile handling at all — the record is written only by `startDeliveryDetached` (`implementations/cli/src/lib/delivery-proc.ts:150-153`). A daemon started by any other path leaves whatever is on disk untouched, and readers believe it.

## Measured on netcup at `275dea4e111f565daa232cf748ab5ab0597e87fa`

Using the product's own resolver (`localProcessPath` over `DELIVERY_PIDFILE`) and probe (`probeLiveness`), not a re-implementation:

```
candidate .cotal/delivery.6e6574637570.pid   exists=false
candidate .cotal/delivery.netcup.pid          exists=false
candidate .cotal/delivery.pid                 exists=true   ← resolved (pre-segmentation read shim)

record names pid 1670039   →  probeLiveness: dead     (record four days stale)
actually running: pid 3303858  `cotal deliver --space netcup …`  →  probeLiveness: alive
```

The daemon was relaunched again during the measurement window under another new pid, with the record still naming the original dead one.

## Why it is worth its own fix: it clears a fail-closed guard

`down` decides what to stop from that same record. Asking its own predicates (`implementations/cli/src/commands/down.ts:305-332`):

| | this host | control: record names a live pid | control: no record |
|---|---|---|---|
| `processRecorded` | true | true | false |
| `processAlive` | **false** | true | false |
| `mayBeRunning` | **false** | true | false |

The controls matter: the instrument is fully capable of answering `alive` and `absent` correctly. It is the ownership that is missing, not the probe.

`mayBeRunning` is documented at `down.ts:316-323` as the guard that must fail **closed** so that `cotal down nats` cannot pull the broker out from under a live dependant. Here it returns false while the daemon is live, because the record it fails closed on names a corpse. A stale record does not merely under-report; it satisfies the proof-of-death the guard requires.

## Not a duplicate of #1238

[...]

Issue #1526, opened 2026-09-12
a session can report state=ready with a transport up while its delivery queue never drains, and cotal_reconnect does not clear it


davidfarah2003:
Hit first-hand while orchestrating the issue pipeline. This is a liveness defect in the delivery path that presents as a healthy session, which is the worst shape for it.

## Symptom

`cotal_connection_status` reported, over a span of more than an hour:

```
state: ready              connected: true        transportConnected: true
stopping: false
bufferedCount: 96         automaticCount: 96
oldestAutomaticAt:        2026-09-12T13:38:17Z   (>2h old at time of reading)
lastDrainedAt:            2026-09-12T14:28:07Z   (75 min before the reading)
```

Every health field is green. The transport is connected, the bind is complete, the session is not shutting down. And **96 messages had not been delivered**, with nothing drained for 75 minutes.

`cotal_inbox` consistently returned `No pull-only messages. 96 connector-managed automatic messages are still queued.` So the messages are known to exist, are counted, and are named in the response, but cannot be read.

## Reconnect does not clear it

`cotal_reconnect` returned `Reconnected ✓`. After it:

```
bufferedCount: 96         lastDrainedAt: 2026-09-12T14:28:07Z   (unchanged)
```

Same count, same drain timestamp. **A reconnect that reports success while changing nothing is a second defect layered on the first**, because it is the documented manual recovery path for exactly this situation and it gives a false all-clear.

## Why this is worse than a disconnect

A disconnected session is obvious and every consumer handles it. This session reports `ready` and is therefore trusted. In my case the practical consequence was that I spent roughly forty minutes attributing a stall to the wrong cause: I could see peers' work was not reaching me, the roster showed reviewer seats sitting `idle`, and I publicly pressed several of them to do work they may well have already done and reported. **The status surface actively pointed away from the fault.**

There is a related symptom worth checking in the same area: joining a channel with replay enabled reports `Backfilled N earlier messages into your inbox`, and those N messages then never appear in `cotal_inbox`. I saw this with N=6 on two separate channels. It may share a root cause with the above, since both involve messages that are counted but not deliverable.

## Suggested handling

1. `state` should not read `ready` when the delivery queue has made no progress for a configurable interval while non-empty. A distinct state (`stalled`, say) that names `bufferedCount` and `lastDrainedAt` would make the fault self-describing rather than invisible.
[...]

PR #1693, opened 2026-09-17
fix(ci): gate two suites nothing ran and drop a duplicate chain entry

L4XB:
## The problem

`pnpm smoke:gate-inventory` is the second entry in the `smoke:ci` chain, so every pull request's `smoke (shard 1/4)` job stops there with exit 1 about a minute in, and the rest of that shard never runs:

```
✗ FAIL: 2 suite(s) exist but nothing runs them, and they are not in UNGATED:
      smoke:delivery-pidfile-ownership
      smoke:reap-absent-refusal
✗ FAIL: bin/smoke/ci-suites.txt names 1 suite(s) twice: smoke:connection-status
GATE INVENTORY FAILED ❌
✗ shard 1/4 FAILED at: pnpm smoke:gate-inventory (exit 1)
NEVER RAN — 143 of 144 planned smoke(s) in this shard's partition did not start
```

That is from #1689's job (`105297955036`, head `5681bbaf`, started 16:59:05Z, failed 17:00:23Z). Shard 1/4 fails the same way in 1m3s to 1m9s on #1691, #1633 and #1692. It reproduces on a clean checkout of `origin/main` at `b086974f`, so it is main's state rather than any branch's change.

## Where the three findings come from

All three arrived on 2026-09-17. Established with `git log -S` on a full clone plus parent-versus-commit greps:

| finding | commit | what it did |
| --- | --- | --- |
| `smoke:reap-absent-refusal` run by nothing | `83ab007f` — *fix(manager): refuse a reap that cannot prove the seat gone (#1554) (#1671)* | added the script to `package.json` (absent in `83ab007f^`), touched no `bin/smoke/` file |
| `smoke:delivery-pidfile-ownership` run by nothing | `44cdcc2e` — *fix(delivery): have the daemon write and remove its own liveness record (#1528) (#1674)* | the same, no registry entry |
| `smoke:connection-status` named twice | `baed5d14` — *fix(connector-core): measure queue progress at the head (#1526) (#1677)* | added a `ci-suites.d/` fragment for a suite already on the frozen list |

The first copy of `smoke:connection-status` entered `bin/smoke/ci-suites.txt` in `4e802614` (2026-08-29, *feat(connector-core): report connection status*) at parse index 429 of 437, so the fragment is the second copy, and the chain ran that suite on two runners: the legacy index put it on shard 1, the fragment hash on shard 2.

## The change

Three files, all under `bin/smoke/ci-suites.d/`:

- **added** `00881ec2…118b.txt` → `smoke:reap-absent-refusal`
- **added** `c1d098d8…bb29.txt` → `smoke:delivery-pidfile-ownership`
- **deleted** `d67ed066…d079.txt`, the duplicate `smoke:connection-status`

Each filename is `sha256(<public suite name>)`, one suite per file with a `#` comment naming the issue the suite proves, as `bin/smoke/ci-suites.d/README.md` prescribes and as the neighbouring fragments are written.

**Gated rather than exempted**, because both suites prove behaviour that is already shipped on main and both pass here, and `gate-inventory.smoke.ts:38` says that adding a line to `UNGATED` is a decision rather than a fix: if a suite proves shipped behaviour, gate it.

```
pnpm smoke:reap-absent-refusal          REAP-ABSENT-REFUSAL OK ✅ (7 passed, 0 failed)
pnpm smoke:delivery-pidfile-ownership   DELIVERY-PIDFILE-OWNERSHIP SMOKE OK ✅ (14 passed, 0 failed)
pnpm smoke:connection-status            PASS: 23 passed, 0 failed
```

**The duplicate was removed on the fragment side, not on the `ci-suites.txt` side**, and that choice is measured rather than preferred:

- Deleting the fragment: `pnpm check:shard-stability b086974f 0065ef6d` → `STABLE`, 0 of 437 frozen indices changed, 0 of 611 pre-existing suites on a different runner. The suite is not dropped from a shard: it keeps shard 1 through its frozen index, and shard 1's partition stays 144 suites.
- Deleting the line instead: on a synthetic commit built with `commit-tree` (so no worktree was touched), the same tool reports `RESHARD=1`, 7 of 436 frozen indices changed and 8 of 611 pre-existing suites moving runner, including `smoke:connection-status` itself.

## Everything run

| command | result |
| --- | --- |
| `pnpm smoke:gate-inventory` | **GATE INVENTORY OK ✅** (exit 0; it fails on `main`) |
| `pnpm smoke:ci-fragments` | 8 passed, 0 failed |
| `pnpm check:shard-stability b086974f 0065ef6d` | STABLE, 0 of 437 frozen indices moved, 0 of 611 suites changed runner |
| `pnpm smoke:reap-absent-refusal` | 7 passed, 0 failed |
| `pnpm smoke:delivery-pidfile-ownership` | 14 passed, 0 failed |
| `pnpm smoke:connection-status` | 23 passed, 0 failed |
| `node scripts/check-operator-literals.mjs --range origin/main..HEAD` | `findings=0/2376_files status=clean` |

Environment: macOS (Darwin 25.6.0, arm64), Node v24.2.0 (CI pins 22, which this host does not have), pnpm 11.1.2, `nats-server v2.12.12`, a short `TMPDIR` because the default one on this host is too long for a unix socket.

## Not verified

- The 10 `UNGATED` entries the inventory reports as untriaged debt are untouched here; this change only clears the two hard failures and the duplicate.
- Whether CI's own Node 22 changes any of the three suite results. Both newly gated suites are broker-free or use `nats-server` directly, so nothing here looks version-sensitive, but I could not run Node 22 on this host.
- `.internal` is not initialised in this checkout, so the plan-record requirement in `AGENTS.md` could not be checked.


davidfarah2003:
This landed already, as #1700, and that was my duplication rather than anything wrong with this PR. Yours was open from 2026-09-17 and I opened mine on 2026-09-18 without checking for an existing one first. Apologies for the wasted work.

The two changes are the same three files with the same effect: gate `smoke:reap-absent-refusal` and `smoke:delivery-pidfile-ownership`, and drop the duplicate `smoke:connection-status` fragment. Only the comment prose differs, and both numberings are right, yours naming the issue and mine the pull request that added the script without its fragment.

Closing as already on main. Nothing here needs changing.

Flagging one thing in passing, since your fragment comment gets it exactly right: `smoke:reap-absent-refusal` guards that an `absent` reap refuses instead of rendering as a disposal. On main right now the ordinary despawn path hits that refusal and skips its footprint teardown, because a clean despawn consumes the custody record before static retirement reaps the same reference. Written up with the evidence in #1706.

